Issue #921, opened 2015-02-19
Improve code coverage for System.ComponentModel.Annotations


stephentoub:
According to the current code coverage report, this library is only at 75%.  Our general goal is closer to 90%, though this will depend on the library.  We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

ghuntley:
This looks like a nice and easy/simple starting point for contribution - Bagsies. Would a pull-request per annotation linked to this issue be acceptable?

MattWhilden:
@ghuntley is going to go ham on this thing.

karelz:
@ghuntley are you still working on it?

danmosemsft:
It's now at an awesome 95%! Closing
https://ci.dot.net/job/dotnet_corefx/job/master/job/code_coverage_windows/Code_Coverage_Report/

PR #1619, opened 2015-05-03
added EditableAttribute construction test

ghuntley:
+5 covered - from 0% to 100% coverage.

progresses issue #921

[x] When the value passed to EditableAttribute's ctor is false?
[ ] When AllowInitialValue is changed, making sure AllowEdit isn't also changed?

	# Ignore - WIP notes.

	The actual intent vs implementation of
	https://github.com/dotnet/corefx/blob/master/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/EditableAttribute.cs

	It makes minimal sense that public bool AllowEdit { get; private set; } has a
	private setter and public bool AllowInitialValue { get; set; } does not.

	InitialValue imho should be private/immutable whilst AllowEdit should be
	mutable.

	ellismg 23 minutes ago :

	@ghuntley if I had to wager a guess, I would think the intention is that you
	must always set a value for AllowEdit, but you don't for AllowInitialValue
	(since it inherits the value from the .ctor. I don't think people expected
	folks to new up these attributes at runtime and start calling the setters on
	them, instead just set their values in metadata But that's just my guess from
	looking at the surface area and docs. Looking into the internal history, it
	seems like previously we had an "ImmutableAttribute" which had the
	AllowInitialValue property (but not AllowEdit) and it was replaced with
	EditableAttribute. Jeff Handley may have been responsible for the creation of
	this type.

	waynd: @ghuntley the problem with making AllowInitialValue’s setter private is that
	you can no longer do [Editable(true, AllowInitialValue = false)] i dont think
	the intent for usage is to ever change AllowInitialValue after creation, but to
	allow it to be set by name on application, it has to have a public setter
	otherwise you need to add a new ctor with two parameters — which may be a
	better solution, but can also be an anti pattern. easy to miss the difference
	between [Editable(true, false)] and [Editable(false, true)]

### DIFF
--- a/src/System.ComponentModel.Annotations/tests/EditableAttributeTests.cs
+++ b/src/System.ComponentModel.Annotations/tests/EditableAttributeTests.cs
@@ -1,0 +1,17 @@
+﻿using Xunit;
+
+namespace System.ComponentModel.DataAnnotations
+{
+    public class EditableAttributeTests
+    {
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public static void Can_construct_and_both_AllowEdit_and_AllowInitialValue_are_set(bool value)
+        {
+            var attribute = new EditableAttribute(value);
+            Assert.Equal(value, attribute.AllowEdit);
+            Assert.Equal(value, attribute.AllowInitialValue);
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/System.ComponentModel.Annotations/tests/EditableAttributeTests.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace System.ComponentModel.DataAnnotations
+{
+    public class EditableAttributeTests
+    {
+        [Fact]
+        public static void Can_construct_and_both_AllowEdit_and_AllowInitialValue_are_set()
+        {
+            var attribute = new EditableAttribute(true);
+            Assert.Equal(true, attribute.AllowEdit);
+            Assert.Equal(true, attribute.AllowInitialValue);
```
stephentoub: What about:
- When the value passed to EditableAttribute's ctor is false?
- When AllowInitialValue is changed, making sure AllowEdit isn't also changed?

ellismg:
I would add the second test case and squash the commits.

stephentoub:
Replaced by #2005

PR #1620, opened 2015-05-04
exclude Visual Studio 2015 `.vs` working directory

ghuntley:
"The .vs directory is the new location where per-user solution-specific items are saved. Previously there was no directory where this information could be placed, so the IDE kept adding files alongside the
solution. It's finally a solution to a problem that's been here for a very long time." - @sharwell

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,10 @@ bld/
 [Oo]bj/
 msbuild.log
 
-# Roslyn stuff
+# Visual Studio 2015
+.vs/
+
+# Visual Studio 2015 Pre-CTP6
 *.sln.ide
 *.ide/
 


.gitignore
```diff
@@ -21,6 +21,9 @@ bld/
 [Oo]bj/
 msbuild.log
 
+# Visual Studio 2015
+.vs/
+
 # Roslyn stuff
```
sharwell: :bulb: The definition above *replaces* the items currently listed under `Roslyn stuff`. Visual Studio "14" CTPs used to create a folder **{SolutionName}.sln.ide**, but starting with CTP 6 it creates a folder just called **.vs**.

stephentoub:
LGTM

.gitignore
nguerrera: Good point. Our documentation still suggests you can use VS 2015 Preview. We may be able to bump that requirement, but I'm not sure I would  tie it to this PR. 

How about just changing "Roslyn stuff" to "Visual Studio 2015 Pre-CTP6" so that it's clear when Pre-CTP6 is long gone that this can be deleted?

nguerrera:
LGTM apart from the small comment above -- Can you change "Roslyn stuff" to "Visual Studio 2015 Pre-CTP6."?

.gitignore
ghuntley: Fixed, just waiting for the build to go green.

ghuntley:
Done.

PR #1636, opened 2015-05-05
Move all packages.config to project.json

ericstj:
DNX and BuildTools now support all of the nuget v2 conventions that we depend on so we can
move all remaining nuget v2 packages to the project.json file and eliminate packages.config.

This change also adds "frameworks" to the project.json so that dnu restore can calculate
applicable assets.  I am not yet using "runtimes" in the project.json since we are still resolving
our runtime assets manually (eg: build tools copies known files from packages).  A future
change will move us to using our runtime packages and add runtimes sections to projects
as appropriate.


ericstj:
@weshaggard @stephentoub @ellismg can one of you give this a quick review?  It's going to get stale  pretty quickly.

ellismg:
Did a quick glance over the diffs, LGTM,

PR #1639, opened 2015-05-06
Remove trick for blocking usage of var

nguerrera:
We decided not to keep these around, but I just spotted one
incoming in a PR and decided to grep for others. There was
just one, which this removes.

cc @stephentoub @AlekseyTs


weshaggard:
LGTM

stephentoub:
Thank you. LGTM.

PR #1643, opened 2015-05-06
Fix incorrect documentation about parallel dispose

nguerrera:
Previously, PEReader allowed mutliple calls to Dispose in parallel, but
that was removed because the caller must not Dispose the PEReader in
parallel to any other non-Dispose operation and the caller's
synchronization required to guarantee that can also guarantee that
Dispose is not called in parallel.


nguerrera:
cc @tmat @stephentoub

stephentoub:
LGTM.

Issue #1556, opened 2015-04-28
NullReferenceException in FileSystemWatcher in CI


stephentoub:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
     at System.IO.FileSystemWatcher.get_IsHandleInvalid()
     at System.IO.FileSystemWatcher.Monitor(Byte[] buffer)
     at System.IO.FileSystemWatcher.CompletionStatusChanged(UInt32 errorCode, UInt32 numBytes, NativeOverlapped* overlappedPointer)
     at System.Threading._IOCompletionCallback.PerformIOCompletionCallback(UInt32 errorCode, UInt32 numBytes, NativeOverlapped* pOVERLAP)

d:\j\workspace\dotnet_corefx_windows_debug_prtest\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00040\lib\tests.targets(101,5): warning MSB3073: The command "corerun.exe xunit.console.netcore.exe System.IO.FileSystem.Watcher.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests  -notrait category=OuterLoop -notrait category=failing " exited with code -1073741819. [d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.IO.FileSystem.Watcher\tests\System.IO.FileSystem.Watcher.Tests.csproj]
d:\j\workspace\dotnet_corefx_windows_debug_prtest\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00040\lib\tests.targets(112,5): error : One or more tests failed while running tests from 'System.IO.FileSystem.Watcher.Tests' please check log for details! [d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.IO.FileSystem.Watcher\tests\System.IO.FileSystem.Watcher.Tests.csproj]
```
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_debug_prtest/1052/console

krwq:
cc: @sokket

chcosta:
@stephentoub, have you happened to notice this occurring frequently or is this an infrequent occurrence?

@ericstj, in FileSystemWatcher.Win32.cs

56         private void StopRaisingEvents() 
57         { 
58             // If we're not attached, do nothing. 
59             if (IsHandleInvalid) 
60             { 
61                 return; 
62             } 
63 
64             // Close directory handle 
[...]

PR #1649, opened 2015-05-06
Fix NullReferenceException in FileSystemWatcher.

chcosta:
This change is to address the NullReferenceException reported in Issue #1556.  See that issue for additional details.  

I was able to repro this by creating a unit test which creates two threads.

Thread 1: Create and delete a file
Thread 2: Enable and disable raising events

I then ran thread 1 and thread 2 using the same FileSystemWatcher for 30 seconds.  On average, it took less than 10 seconds for a NullReferenceException to fire.  

I have a local fix for this, which is to remove the "_directoryHandle=null" statement.  Not setting _directoryHandle to null means the GC won't collect it.  Whenever we start raising events, we initialize "_directoryHandle" to refer to the directory being monitored.  An alternative solution would be to use a lock, but that is a bit heavy handed and not necessary unless setting _directoryHandle to null is a requirement.


stephentoub:
Thanks for tracking it down.

PR #1662, opened 2015-05-07
Improve Path test coverage

stephentoub:
Adding some more tests for System.IO.Path, in particular for its functions that access OS functionality.

(Also adding one more test case for Stopwatch, for one of its members that didn't have any coverage.)


ellismg:
LGTM

PR #1666, opened 2015-05-07
Remove a few unnecessary dictionary lookups

stephentoub:
I was considering writing a Roslyn analyzer to look for unnecessary dictionary lookups, and searched across corefx for patterns like:
```
if (dict.ContainsKey(key))
    Use(dict[key]);
```
and
```
if (dict.ContainsKey(key))
    dict.Remove(key);
```
where there's an extra dictionary lookup that can be removed trivially. I only found a few such occurrences, not enough to warrant spending time on an analyzer at present.  But since fixing those I did find is quick and easy, this commit does so.


weshaggard:
LGTM

ellismg:
LGTM

PR #1674, opened 2015-05-08
Use Thread.Sleep in System.Threading's Helpers.Sleep on CoreCLR

stephentoub:
Helpers.Sleep (used by ReaderWriterLockSlim) is currently coded to call WaitOne on a static WaitHandle as an approximation for Thread.Sleep.  This is a holdover from when Thread.Sleep was unavailable.  Now that it is available, it's better to simply use it, rather than constructing and caching a never-set ManualResetEvent purely for the purposes of waiting on it to put the thread to sleep.


mellinoe:
LGTM

PR #1675, opened 2015-05-08
Remove dead code from System.Threading

stephentoub:
These types are internal and aren't being used at all in the assembly.


stephentoub:
@mellinoe, please let me know if there's some nuance I'm missing about why these are here.

mellinoe:
These are most likely used in the .NET Native build. We can remove them from the project file but let's leave the source code for now.

stephentoub:
Ok, thanks.  As you and I discussed, it's strange then that these files from the .NET Native implementation are here but others from it are not; I'll close out this PR, and in the other one where I was doing some cleanup, I'll just remove them from the .csproj... then you can clean them up as is appropriate in TFS.

PR #1684, opened 2015-05-08
Fix IImmutableList documentation issue

govert:
Remove incorrect part of description for IImmutableList.Add

The documentation for IImmutableList.Add incorrectly stated that the method would return the current list if the item being added is already present in the list. This is not correct for an IImmutableList (but would be correct for an ImmutableSet). The implementation of this method in ImmutableList<T>.Add is correct, and in fact has a contract that ensures the resulting list has an additional element added.


stephentoub:
LGTM.

PR #1686, opened 2015-05-09
Add nuget/myget version badges to library components

stephentoub:
Experimenting with including badges for each library, highlighting the latest published dev version of the component and providing a link to where it can be downloaded.

"View" the README.md as rendered markdown to see all of the badges, courtesy of shields.io.  The majority of these are from the dotnet-core myget gallery; when components weren't available there, they were usually on NuGet.  The remaining few that show up as "invalid" don't yet have a published package, but as soon as they're published to myget, these badges will be updated automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,206 +45,86 @@ Want to chat with other members of the CoreFX community?
 
 ## .NET Core Library Components
 
-The repo contains the following components. More libraries are coming soon. ['Watch'](https://github.com/dotnet/corefx/subscription) the repo to be notified.
-
-* **System.Collections**. Provides classes that define generic collections, which allow developers to create 
-  strongly-typed collections.
-
-* **System.Collections.Concurrent**.  Provides a set of thread-safe collection types, instances of which may be used 
-  concurrently from multiple threads.
-
-* **System.Collections.Immutable**. Provides a set of immutable collection types that make it easy to keep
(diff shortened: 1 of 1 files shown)


stephentoub:
@richlander, @jhendrixMSFT, @ericstj, thoughts?

mellinoe:
I like it. I think it reads better/easier than our current mass of bullet points, and the package badge things are pretty nifty as well.

Petermarcu:
I like it. Not sure why the build failed with this.

stephentoub:
Thanks, guys.

> Not sure why the build failed with this.

All Windows builds in CI appear to be failing due to #1685.

weshaggard:
I like the badges as well bug I'm curious about the links. Perhaps an orthogonal question but when are our builds going to point to the dotnet-core myget feed like you are linking from the badges. This is likely a question for @jhendrixMSFT.

stephentoub:
> but I'm curious about the links

I couldn't find any way to get a link to the latest version without hardcoding it into the link. If there is a way, we should change the links to use it.

jhendrixMSFT:
CoreFx builds are now pulling from the dotnet-core repro, see https://github.com/dotnet/corefx/commit/08c0d8628ef621fbd44d7aaa85b3239616a6b108.

Issue #1182, opened 2015-03-19
System.Linq performance improvement suggestions


ikopylov:
With Linq-to-Objects it is quite common practice to perform a series of transformations, and then materialize sequence to a concrete collection type by calling _ToArray(), ToList(), ToDictionary()_. These operations would work much faster if they knew the number of elements in the sequence.

Currently System.Linq.Enumerable has special treatment for _ICollection\<T\>_ interface only. 
I suppose that additional support for _IReadOnlyCollection\<T\>_ can improve performance in some cases, because through it we can figure out the total number of elements.

Another problem with System.Linq is that in many cases the information about the number of elements is lost. One of the most common example:
```C#
List<int> source = ...;
List<int> transformed = source.Select(o => o + 1).ToList();
```

Obviously _Select_ does not change the number of elements in the sequence, but _ToList_ method cannot take advantage of that. Information is lost.
In this particular scenario, it would be great for _Select_ to return some _SelectIterator_ instance, which implements _IReadOnlyCollection\<T\>_, and thereby passes the number of elements to the subsequent methods.

__Steps to measure performance gain__:

1. Find in _System.Linq.Tests\Performance_ the test for the method you've changed;
2. Uncomment ```[Fact]``` attribute above that method;
3. Build test project by Visual Studio in Release mode;
4. Go to the folder with tests binaries: ```bin\tests\Windows_NT.AnyCPU.Release\System.Linq.Tests\aspnetcore50\```;
5. Open command prompt (cmd.exe or PowerShell);
6. Run command: ```CoreRun.exe xunit.console.netcore.exe System.Linq.Tests.dll -parallel none -trait "Perf=true"```;
7. Wait for results that will be printed right in the command window;
8. Don't forget to run tests with different collection sizes. This can be done by varying _elementCount_ argument of _Measure_ method.

__Casting to ```IReadOnlyCollection<T>```__ is a slow operation so it is not a good idea to check if this interface implemented. The performance drop will be most noticeable on the small collections.

__Tasks__:

- [ ] ```Select```: add iterators for ```List<T>```, ```Array[T]```, ```ICollection<T>```;
- [ ] ```ToArray```, ```ToDictionary```: add special support for ```ICollection<T>``` (and, very carefully, for ```IReadOnlyCollection<T>```) to get the initial capacity;
- [ ] ```ToList``` (???): add special support for ```IReadOnlyCollection<T>``` to get the initial capacity (separated as it will affect _System.Collections.Generic_);
[...]

PR #1688, opened 2015-05-09
LINQ Range, Repeat unit tests

ikopylov:
Unit tests for ```Range``` and ```Repeat``` methods from ```System.Linq```. This PR is a part of work on #1182 issue.

### DIFF
--- a/src/System.Linq/tests/RangeRepeatTests.cs
+++ b/src/System.Linq/tests/RangeRepeatTests.cs
@@ -1,0 +1,193 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class RangeRepeatTests
+    {
+        #region Range
+
+        [Fact]
+        public void Range_ProduceCorrectSequence()
+        {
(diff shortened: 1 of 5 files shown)


weshaggard:
cc @VSadov

src/System.Linq/tests/RangeRepeatTests.cs
```diff
@@ -0,0 +1,193 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class RangeRepeatTests
+    {
+        #region Range
```
stephentoub: Why not split Range and Repeat each into their own files?

stephentoub:
Other than this being in one file rather than two, LGTM.

src/System.Linq/tests/RangeRepeatTests.cs
ikopylov: My thoughts were: it is good to put these tests in a single file, because the test cases are very similar. But it is not a problem to split them into separate files. I'll do it.

stephentoub:
@dotnet-bot test this please

VSadov:
LGTM

PR #1692, opened 2015-05-10
LINQ ToArray, ToList, ToDictionary unit tests

ikopylov:
Unit tests for ```ToArray```, ```ToList``` and ```ToDictionary``` methods from ```System.Linq```. This PR is a part of work on #1182 issue.


stephentoub:
cc: @VSadov

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+
+namespace System.Linq.Tests
+{
+    public class ToArrayToListToDictionaryTests
```
stephentoub: Can we split this into multiple classes/files, one per each of the To* methods?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+    {
+        #region =============  Helper classes =============
+
+        private class EnumerableCollectionTest<T> : IEnumerable<T>
```
stephentoub: Nit, but could you rename these helper classes to end with the thing they are, e.g. TestEnumerable, TestReadOnlyCollection, TestCollection, etc., rather than EnumerableCollectionTest, ReadOnlyCollectionTest, etc.?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            {
+                for (long i = 0; i < MaxSize; i++) yield return (byte)1;
+            }
+            IEnumerator IEnumerable.GetEnumerator() { return this.GetEnumerator(); }
+        }
+
+        /// <summary>
+        /// Emulation of async collection change
+        /// </summary>
+        private class AsyncCollectionTest : CollectionTest<int>
```
stephentoub: I don't understand the "Async" naming here.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            CollectionTest<int> source = new CollectionTest<int>(new int[] { 1, 2, 3, 4 });
+            var resultArray = source.ToArray();
+
+            Assert.Equal(source, resultArray);
+            Assert.Equal(1, source.CopyToTouched);
+        }
+
+
+        // Currently not passed
+        //[Fact]
```
stephentoub: I'd suggest leaving [Fact] here but adding [ActiveIssue] with the appropriate issue number, probably 1561.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+
+            Assert.True(resultArray.Length >= 4);
+            Assert.Equal(1, resultArray[0]);
+            Assert.Equal(2, resultArray[0]);
+            Assert.Equal(3, resultArray[0]);
+            Assert.Equal(4, resultArray[0]);
+        }
+
+
+        //[Fact]
```
stephentoub: Should this be [Fact, OuterLoop]?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        public void ToArray_FailOnExtremelyLargeCollection()
+        {
+            LargeSequenceTest largeSeq = new LargeSequenceTest();
+            Assert.ThrowsAny<OverflowException>(() => { largeSeq.ToArray(); });
```
stephentoub: Could also be an OOM, right?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            CollectionTest<int> source = new CollectionTest<int>(new int[] { 1, 2, 3, 4 });
+            var resultList = source.ToList();
+
+            Assert.Equal(source, resultList);
+            Assert.Equal(1, source.CopyToTouched);
+        }
+
+
+        // Currently not passed
+        //[Fact]
```
stephentoub: Same comment about [Fact, ActiveIssue(...)]

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
```diff
@@ -0,0 +1,469 @@
...
+            Func<int, int> keySelector = key => key;
+            Func<int, int> valueSelector = value =>
+            {
+                if (value == 1)
+                    throw new InvalidOperationException();
+                return value;
+            };
+
+            Assert.Throws<InvalidOperationException>(() => source.ToDictionary(keySelector, valueSelector));
+        }
```
stephentoub: Should there be another test for the case where the key selector returns an invalid key, e.g. null?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: These tests use the same set of helper classes. So it will require to put helper classes in separate files too. Is it OK?

stephentoub: I think that's fine.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: Agreed

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: Ok. I'll split them.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: This class helps to verify that ```ToArray``` and ```ToList``` can work with Concurrent collections. It adds a new element to the sequence each time the ```Count``` property read.
Probably you are right: 'Async' is not a good naming. Any suggestion how to rename it?

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: I didn't know about ```ActiveIssue``` attribute. I'll add it.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: Yes. You are right.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: This may be the case on 32-bit machines when there's no continuous block of 1Gb virtual memory. It is unlikely that such situation can occur during test running. But it is not a problem to handle it too.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
ikopylov: Ok. I will add tests for that case. 
But it is rather tests for ```Dictionary<TKey, TValue>``` class, not for ```ToDictionary``` method.

stephentoub: > But it is rather tests for Dictionary<TKey, TValue> class, not for ToDictionary method.

It's ```Dictionary<TKey, TValue>``` that currently does the argument validation, but that assumes that ```ToDictionary``` appropriately passes such values through, which of course a faulty implementation might not, so it's still good to test it via ```ToDictionary```.

src/System.Linq/tests/ToArrayToListToDictionaryTests.cs
Maxwe11: rather

    [Fact]
    [OuterLoop]

there is no rule in corefx guidelines but https://github.com/dotnet/codeformatter/pull/64

stephentoub:
@dotnet-bot test this please

VSadov:
LGTM

PR #1693, opened 2015-05-11
Fixing race condition in non-generic collections tests

### LINKED ISSUE #1504: Some Collections.NonGeneric tests have race conditions and ocassionally failing
( http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_debug_prtest/963/console )
```
  Xunit.Sdk.EqualException: Assert.Equal() Failure
            Î“Ã¥Ã´ (pos 0)
  Expected: en-US
  Actual:   cs-CZ
            Î“Ã¥Ã¦ (pos 0)
     at Xunit.Assert.Equal(String expected, String actual, Boolean ignoreCase, Boolean ignoreLineEndingDifferences, Boolean ignoreWhiteSpaceDifferences)
     at Xunit.Assert.Equal(String expected, String actual)
     at Comparer_DefaultInvariant.runTest() in d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Collections.NonGeneric\tests\Comparer\Comparer_DefaultInvariant.cs:line 52
  Fail! iCountErrors==1
     Comparer_DefaultInvariant.ExecuteComparer_DefaultInvariant [FAIL]
        Assert.True() Failure
```

The tests are setting:
```csharp
CultureInfo.DefaultThreadCurrentCulture = culture;
CultureInfo.DefaultThreadCurrentUICulture = culture;
```
Which may fail when run on different threads.

Maxwe11:
Fixes #1504

### DIFF
--- a/src/System.Collections.NonGeneric/tests/Comparer/Comparer_DefaultInvariant.cs
+++ b/src/System.Collections.NonGeneric/tests/Comparer/Comparer_DefaultInvariant.cs
@@ -96,7 +96,6 @@ public class Comparer_DefaultInvariant
 
 
     [Fact]
-    [ActiveIssue(1504)]
     [ActiveIssue(846, PlatformID.AnyUnix)]
     public static void ExecuteComparer_DefaultInvariant()
     {
--- a/src/System.Collections.NonGeneric/tests/XunitAssemblyAttributes.cs
+++ b/src/System.Collections.NonGeneric/tests/XunitAssemblyAttributes.cs
@@ -1,0 +1,7 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+//Some Collections.NonGeneric tests have race conditions and ocassionally failing (e.g. see issue #1504)
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly, DisableTestParallelization = true, MaxParallelThreads = 1)]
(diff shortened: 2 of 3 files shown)


src/System.Collections.NonGeneric/tests/Comparer/Comparer_DefaultInvariant.cs
```diff
@@ -7,6 +7,7 @@
 using System.Globalization;
 using Xunit;
 
+[Collection("DefaultThreadCurrentCulture")]
```
stephentoub: I think I'd rather see parallelism disabled for this whole test suite rather than using a named collection like this.  None of these tests were written to be run in parallel, and I worry this is just one of potentially many such concurrency issues.  I think it'd make sense to disable parallelism for the library until it can be entirely rewritten (which it should be at some point).

Maxwe11: > I think it'd make sense to disable parallelism for the library until it can be entirely rewritten (which it should be at some point).

If tests will be rewritten they should also will rely on `CultureInfo.DefaultThreadCurrentCulture` and also will require deterministic execution like `[Collection]` does.

>None of these tests were written to be run in parallel, and I worry this is just one of potentially many such concurrency issues.

I think we shouldn't worry about that until it appears, but if it appear it'd be helpful for rewriting.

Maxwe11:
@stephentoub updated

src/System.Collections.NonGeneric/tests/XunitAssemblyAttributes.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly, DisableTestParallelization = true, MaxParallelThreads = 1)]
```
stephentoub: Could you add a comment here explaining why this is being used?

stephentoub:
LGTM.  Thanks for the fix.

krwq:
:+1:

PR #1694, opened 2015-05-11
Update comments on Safe{File/Pipe}Handle.ReleaseHandle

### LINKED ISSUE #1689: SafePipeHandle.ReleaseHandle() violates contract on Unix without justification.
The contract for ReleaseHandle ([per MSDN](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.safehandle.releasehandle.aspx)) is to only return false on catastrophic failure. However as coded, on Unix it will return false for an `EINTR`, which is hardly catastrophic. Either it should retry on EINTR, or it should have the comment expanded to explain why we should consider that a catastrophic error.

stephentoub:
Add a better comment to indicate why we don't retry on an interrupted close syscall.

(And since I was changing this function anyway, I made SafePipeHandle's ReleaseHandle identical to that of SafeFileHandle, adding the same assert used by SafeFileHandle.ReleaseHandle.)

Fixes #1689

### DIFF
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
@@ -69,9 +69,9 @@ namespace Microsoft.Win32.SafeHandles
         [System.Security.SecurityCritical]
         protected override bool ReleaseHandle()
         {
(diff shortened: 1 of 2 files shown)


stephentoub:
@dotnet-bot test this please

ellismg:
LGTM

stephentoub:
Thanks, Matt.

PR #1695, opened 2015-05-11
Removes some tests that are now bogus

KrzysztofCwalina:
Last week I changed the implementation of encoders such that they now don't guarantee that their Default property always returns the same instance of the default encoder. Access to the property’s backing field used to be synchronized using memory berries; I now removed these synchronizations. This change was made to simplify code and improve performance of the common scenario: retrieving the default encoder, at the cost of having to create (and collect) more than one instance of the encoder in rare cases (once in a blue moon).
This change just remove tests that test behavior that we don’t guarantee anymore.

### DIFF
--- a/src/System.Text.Encodings.Web/tests/HtmlEncoderTests.cs
+++ b/src/System.Text.Encodings.Web/tests/HtmlEncoderTests.cs
@@ -71,17 +71,6 @@ namespace Microsoft.Framework.WebEncoders
             }
         }
 
-        [Fact]
-        public void Default_ReturnsSingletonInstance()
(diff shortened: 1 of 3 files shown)


stephentoub:
LGTM.

KrzysztofCwalina:
@dotnet-bot test it please

stephentoub:
@KrzysztofCwalina, as with every other Windows build in CI right now, it's failing because of #1685... retesting won't help at the moment :wink:

stephentoub:
@dotnet-bot test this please

PR #1698, opened 2015-05-11
Enable DispatchProxy to proxy internal types

### LINKED ISSUE #1696: System.Reflection.DispatchProxy fails to generate proxy for internal types
When System.Reflection.DispatchProxy is asked to create a proxy for an internal type, it fails during Reflection Emit because the dynamic assembly cannot access the internal type.

roncain:
System.Reflection.DispatchProxy failed to generate proxies
for internal types because the dynamic assembly could not
access them.

The fix was to add an attribute to the dynamic assembly to identify
each assembly containing internal types used by proxy generation.

Fix #1696


roncain:
@gkhanna79 @jkotas @weshaggard @nguerrera

jkotas:
LGTM

nguerrera:
Looks good to me too.

PR #1704, opened 2015-05-11
Small API cleanup

KrzysztofCwalina:
This is last set of small changes before I will submit the API for an API review.

I changed the order of parameters to Encode to make the APIs work better with intellisense, i.e.
I made the parameter order as consistent as possible.

It would be even better if Encode that takes TextWriter, and Encode that returns string were not simply overloads,
but we can discuss this at the API review.


KrzysztofCwalina:
@dotnet-bot test this please

stephentoub:
LGTM

Eilon:
:shipit: from me too. I like the new parameter order.

PR #1713, opened 2015-05-12
Add "frameworks" node to project.json files that missed them

akoeplinger:
This caused an issue when building on Linux as the projects would build against DNX,4.5.1 instead of DNXCore,5.0

### DIFF
--- a/src/Microsoft.VisualBasic/src/project.json
+++ b/src/Microsoft.VisualBasic/src/project.json
@@ -20,5 +20,8 @@
     "System.Text.Encoding": "4.0.10-beta-22809",
     "System.Threading": "4.0.10-beta-22809",
     "System.Threading.Tasks": "4.0.10-beta-22809"
+  },
+  "frameworks": {
+    "dnxcore50": {}
   }
 }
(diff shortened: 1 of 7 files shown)


stephentoub:
LGTM. @ericstj?

ericstj:
LGTM, not sure how I missed these.  Maybe came in after initial change?  Thanks for fixing them!

akoeplinger:
@ericstj :) I actually just saw that System.Security.Principal is also still using packages.config, so I guess that one was simply missed:

- https://github.com/dotnet/corefx/blob/e8f04cede01a78889e842e29b50a0deea84b261c/src/System.Security.Principal/tests/packages.config
- https://github.com/dotnet/corefx/blob/e8f04cede01a78889e842e29b50a0deea84b261c/src/System.Security.Principal.Windows/tests/packages.config

ericstj:
I see, commit is dated before my change to move everyone to project.json but was merged after my PR.  We need to watch for things like this coming in.  I have a tool that I wrote to convert.  I will run it again on the repo and fix everything up.  That might be better then this.

akoeplinger:
Closing in favor of #1717.

PR #1601, opened 2015-05-01
Workarounds for building with Roslyn/MSBuild on Unix.

JeremyKuhne:
Allows a successful compile with xplat MSBuild branch and Roslyn on Mono.

Adds the Roslyn package and UseRoslynCompiler property to use Roslyn.
Disable PDB generation (NYI), skip packaging, disables ECMA delay signing.
Also temporarily disables two test projects and VB.
Fixes casing for a few source file includes.

PR #1715, opened 2015-05-12
Remove mono workarounds in mono.targets that are no longer required

akoeplinger:
@JeremyKuhne these are part of the fixes to get you a working MSBuild+corefx on Mono, I'll write up a comment in https://github.com/dotnet/corefx/pull/1601#issuecomment-101079839 after I validated everything on a fresh VM again.

While they did cause issues when building on Mono 4.1, I'm not sure if you want to keep them around for 3.12 and make this conditional that's why I marked this as WIP.

### DIFF
--- a/mono.targets
+++ b/mono.targets
@@ -24,47 +24,4 @@
   <Target Name="RunTestsForProject">
     <Message Importance="high" Text="Skipping test run pending fix." />
   </Target>
-
-  <!--
-    Mono Microsoft.Portable.Core.targets do not have the condition that the Windows targets do.
-    Copied to override and added the condition.
-
-    Note that this issue is fixed in https://github.com/mono/mono/pull/1464.
-  -->
-  <Target
-    Name="ImplicitlyExpandTargetFramework"
-    Condition="'$(ImplicitlyExpandTargetFramework)' == 'true'"
-    DependsOnTargets="$(ImplicitlyExpandTargetFrameworkDependsOn)">
-
-    <ItemGroup>
-      <ReferenceAssemblyPaths Include="$(_TargetFrameworkDirectories)"/>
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
If we can get working on v4 I'd prefer to just remove these. I'm not having much luck finding a combo that doesn't crash, however, so I'm eager to see your writeup.

Btw, I've got another minor change to allow overriding the reference folder in MSBuild: https://github.com/JeremyKuhne/msbuild/commit/bff24ae68ca9ba3db479d4bb83e704c5d330aef1

If I can get a clean build on a CI with this I'll put a PR in for it.

mono.targets
```diff
@@ -12,59 +12,4 @@
...
-  <!--
-    Skipping this temporarily to focus on the failures in building the product assemblies. The error we get is as follows:
-
-     RunTestsForProject:
-      mono corerun.exe xunit.console.netcore.exe System.Xml.XPath.XmlDocument.Tests.dll  -xml testResults.xml  -notrait category=OuterLoop -notrait category=failing 
-      Cannot open assembly 'corerun.exe': No such file or directory.
-
-    Note that MSBuild will shell launch as opposed to mono launch if the target file doesn't end in .exe.
-  -->
-  <Target Name="RunTestsForProject">
```
JeremyKuhne: This is the equivalent of the skiptests bit so it can't come out yet.

akoeplinger: Ah, sorry, looks like I was a bit overzealous ripping the workarounds out ;)

akoeplinger: PR updated.

JeremyKuhne:
LGTM, @weshaggard ?

weshaggard:
LGTM.

PR #1720, opened 2015-05-13
Improve code coverage for System.ComponentModel.EventBasedAsync

Maxwe11:
from 75% to 87%

### DIFF
--- a/src/System.ComponentModel.EventBasedAsync/tests/AsyncCompletedEventArgsTests.cs
+++ b/src/System.ComponentModel.EventBasedAsync/tests/AsyncCompletedEventArgsTests.cs
@@ -1,0 +1,68 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace System.ComponentModel.EventBasedAsync.Tests
+{
+    // We want to test protected method AsyncCompletedEventArgs.RaiseExceptionIfNecessary
+    // so we should make derived class for access it.
+    public class AsyncCompletedEventArgsTests : AsyncCompletedEventArgs
+    {
+        public AsyncCompletedEventArgsTests()
+            : base(null, false, null)
+        {
+        }
+
+        private AsyncCompletedEventArgsTests(Exception error, bool cancelled, object userState)
+            : base(error, cancelled, userState)
+        {
+        }
+
+        public static IEnumerable<object[]> TestInput
+        {
+            get
+            {
+                return new[]
+                {
+                    new object[] { null, false, (Type)null },
+                    new object[] { null, true, typeof(OperationCanceledException) },
+                    // dummy exceptions
+                    new object[] { new FormatException(), false, typeof(FormatException) },
+                    new object[] { new DllNotFoundException(), true, typeof(OperationCanceledException) }
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData("TestInput")]
+        public static void CtorTest(Exception expectedException, bool expectedCancelled, object expectedState)
+        {
+            var target = new AsyncCompletedEventArgsTests(expectedException, expectedCancelled, expectedState);
+            Assert.Equal(expectedException, target.Error);
+            Assert.Equal(expectedCancelled, target.Cancelled);
+            Assert.Equal(expectedState, target.UserState);
+        }
+
+        [Theory]
+        [MemberData("TestInput")]
+        public static void RaiseExceptionIfNecessaryTest(Exception expectedError, bool cancelled, Type expectedExceptionType)
+        {
+            var target = new AsyncCompletedEventArgsTests(expectedError, cancelled, null);
+
+            if (expectedExceptionType == null) // if null should NOT throw
+            {
+                target.RaiseExceptionIfNecessary();
+            }
+            else
+            {
+                Exception error = Assert.Throws(expectedExceptionType, () => target.RaiseExceptionIfNecessary());
+                if (expectedError != null && !cancelled)
+                {
+                    Assert.Same(expectedError, error);
+                }
+            }
+        }
+    }
+}
--- a/src/System.ComponentModel.EventBasedAsync/tests/AsyncOperationManagerTests.cs
+++ b/src/System.ComponentModel.EventBasedAsync/tests/AsyncOperationManagerTests.cs
@@ -1,0 +1,51 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+using Xunit;
+
+namespace System.ComponentModel.EventBasedAsync.Tests
+{
+    public class AsyncOperationManagerTests
+    {
+        [Fact]
+        public static void SetNullSynchronizationContext()
+        {
+            var originalContext = SynchronizationContext.Current;
+
+            try
+            {
+                var oldContext = AsyncOperationManager.SynchronizationContext;
+                Assert.NotNull(oldContext);
+
+                AsyncOperationManager.SynchronizationContext = null;
+                var newContext = AsyncOperationManager.SynchronizationContext;
+                Assert.NotNull(newContext);
+                Assert.NotSame(oldContext, newContext);
+            }
+            finally
+            {
+                SynchronizationContext.SetSynchronizationContext(originalContext);
+            }
+        }
+
+        [Fact]
+        public static void SetNonNullSynchronizationContext()
+        {
+            var originalContext = SynchronizationContext.Current;
+
+            try
+            {
+                Assert.NotNull(AsyncOperationManager.SynchronizationContext);
+
+                var testContext = new SynchronizationContext();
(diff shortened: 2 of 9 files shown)


src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -133,10 +133,10 @@ private void DoWorkExpectCancel(object sender, DoWorkEventArgs e)
                 return;
             }
 
-            // we want to wait for cancellation - wiat max 100 * timeoutShort millisec
+            // we want to wait for cancellation - wiat max 100 * TimeoutShort millisec
             for (int i = 0; i < 1000; i++)
```
Maxwe11: :exclamation: Comment does not match with code. What should be fixed?

src/System.ComponentModel.EventBasedAsync/tests/AsyncCompletedEventArgsTests.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace System.ComponentModel.EventBasedAsync.Tests
+{
+    public class AsyncCompletedEventArgsTests : AsyncCompletedEventArgs
```
stephentoub: Seems strange to have the tests class derive from the event args class.  Why was it done this way?

Maxwe11: because method `RaiseExceptionIfNecessary(...)` which we want to test (below) is protected

src/System.ComponentModel.EventBasedAsync/tests/AsyncOperationManagerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        public void SetNonNullSynchronizationContext()
+        {
+            Assert.NotNull(AsyncOperationManager.SynchronizationContext);
+
+            var testContext = new SynchronizationContext();
+            AsyncOperationManager.SynchronizationContext = testContext;
+            Assert.Same(testContext, AsyncOperationManager.SynchronizationContext);
+        }
+
+        public void Dispose()
```
stephentoub: Could you move this up to be after the ctor?

src/System.ComponentModel.EventBasedAsync/tests/AsyncOperationManagerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            SynchronizationContext.SetSynchronizationContext(_originalContext);
```
stephentoub: Does xunit guarantee/document that it'll run the ctor, then all of the facts, and then dispose all on the same thread and with no other facts from other tests interleaved on that thread in between the ctor and dispose?  If not, this usage of synchronization context could be problematic, and it'd be better to do the setup and teardown within each fact.

Maxwe11: will fix

src/System.ComponentModel.EventBasedAsync/tests/AsyncOperationManagerTests.cs
Maxwe11: will fix

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
stephentoub: I'm guessing the author intended 1000 to be correct, as in the code, as that would make this timeoutShort seconds total.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -148,45 +148,96 @@ private void DoWorkExpectCancel(object sender, DoWorkEventArgs e)
...
+                throw new TestException(expectedExceptionMsg);
+            };
+
+            bw.RunWorkerCompleted += (sender, e) =>
+            {
+                TestException ex = Assert.Throws<TestException>(() => e.Result);
+                Assert.Equal(expectedExceptionMsg, ex.Message);
+            };
+
+            bw.RunWorkerAsync(expectedArgument);
```
stephentoub: Shouldn't we block or otherwise prevent the fact method from completing until the test has completed?

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -148,45 +148,96 @@ private void DoWorkExpectCancel(object sender, DoWorkEventArgs e)
...
+            Assert.False(bw.WorkerSupportsCancellation);
+            Assert.False(bw.CancellationPending);
         }
 
-        private void DoWorkWithException(object sender, DoWorkEventArgs e)
+        [Fact]
+        public void RunWorkerAsyncTwice()
+        {
+            var bw = new BackgroundWorker();
+            bw.DoWork += (sender, e) => Wait(3 * TimeoutShort);
```
stephentoub: If we want to do more waiting like this in inner loop tests, we should probably significantly decrease the timeout.  As is, this one test will take a second, and we're striving to have each inner loop test suite take no more than a few seconds total (we're not there yet, obviously).

stephentoub: What's the value of the waiting here?  If the goal of the test is to invoke RunWorkerAsync while it's already running, then this kind of timing issue can lead to random failures (plus it can result in the tests taking longer than is needed), and it'd be better to use synchronization to ensure you get the interleaving you want, e.g.
```
var bw = new BackgroundWorker();
Barrier barrier = new Barrier();
b.DoWork += (sender, e) => { barrier.SignalAndWait(); barrier.SignalAndWait(); };
bw.RunWorkerAsync();
barrier.SignalAndWait();
Assert.Throws<InvalidOperationException>(() => bw.RunWorkerAsync());
barrier.SignalAndWait();
```

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
stephentoub: Note that because these tests are running on xunit's SynchronizationContext, and because BackgroundWorker will post back its completion event to run on the original SynchronizationContext, care will need to be taken with such blocking.  Either you'd want to make the test be async and await something, or if you explicitly blocked you'd want to ensure you were on a null SynchronizationContext for the body of the method, as you did in other tests.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -148,45 +148,96 @@ private void DoWorkExpectCancel(object sender, DoWorkEventArgs e)
...
+        }
+
+        [Fact]
+        public void TestCancelAsyncWithoutCancellationSupport()
         {
-            Assert.Equal("Exception", e.Argument);
-            throw new TestException("Exception from DoWork");
+            var bw = new BackgroundWorker() { WorkerSupportsCancellation = false };
+            bw.DoWork += (sender, e) => Wait(3 * TimeoutShort);
+            Assert.Throws<InvalidOperationException>(() => bw.CancelAsync());
```
stephentoub: See previous comment; this test has a race condition due to the timeout, and we should instead use synchronization to ensure we get the interleaving we want.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
Maxwe11: Actually I don't know. Original test didn't.

src/System.ComponentModel.EventBasedAsync/tests/AsyncCompletedEventArgsTests.cs
stephentoub: Ah.  Makes sense, but consider adding a comment to that effect, as it looks strange.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
stephentoub: Yes, please. I didn't see the original test, but if we don't, the test could theoretically not finish by the time the tests were shutdown.  It's possible xunit handles this by hooking some additional members in SynchronizationContext, but even if it does, failures would be difficult to debug.

Maxwe11: got it. will try fix

stephentoub:
Thanks for doing this, @Maxwe11.  Overall, it looks good, but it'd be good to address some of the comments I left before we merge this in, in particular the ones about timeouts.

Maxwe11:
updated

stephentoub:
Thanks for the fixes. LGTM.

stephentoub:
Thanks for the fixes. LGTM.

PR #1735, opened 2015-05-13
Fix ImmutableArray<T>.Equals handling of variant types

### LINKED ISSUE #1731: ImmutableArrayTest.Covariance doesn't work as intended
This issue affects the following tests in System.Collections.Immutable:

* `ImmutableArrayTest.Covariance`
* `ImmutableArrayTest.CovarianceImplicit`
* `ImmutableArrayTest.CastUpReference`

These three tests contain calls to `Assert.Equal` for immutable arrays with different generic arguments. For example:

```csharp
ImmutableArray<string> derivedImmutable = ImmutableArray.Create("a", "b", "c");
ImmutableArray<object> baseImmutable = ImmutableArray<object>.CastUp(derivedImmutable);
Assert.Equal(derivedImmutable, baseImmutable);
```

The problem is the call to `Assert.Equal` resolves to `Assert.Equal<object>(IEnumerable<object>, IEnumerable<object>)` instead of `Assert.Equal<object>(object, object)`. This particular overload of `Assert.Equal` does not use the default equality comparer for the input arguments. Rather, it enumerates over each of the input enumerables checking for sequence equality. Unfortunately, this also undermines the purpose of the test by never calling `ImmutableArray<T>.Equals`.

### LINKED ISSUE #1733: ImmutableArray<T>.Equals(object) does not handle variance
The tests mentioned in #1731 suggest that `ImmutableArray<string>.Empty.Equals(ImmutableArray<object>.Empty)` should return `true`. Currently it does not.

The current implementation of `ImmutableArray<T>.Equals(object)` is the following:

```csharp
if (obj is ImmutableArray<T>)
{
    return this.Equals((ImmutableArray<T>)obj);
}

return false;
```

To handle variance, this method would need to be extended with a fallback to using `IImmutableArray`:

```csharp
if (obj is ImmutableArray<T>)
{
    return this.Equals((ImmutableArray<T>)obj);
}

IImmutableArray other = obj as IImmutableArray;
if (other != null)
{
    return this.array == other.Array;
}

return false;
```

sharwell:
This commit updates the implementation of `ImmutableArray<T>.Equals(object)` to use `IImmutableArray` instead of `ImmutableArray<T>`. This allows the underlying arrays to be compared by reference as instances of `System.Array`, even if the `ImmutableArray<T>` instances wrapping them use a different type arguments `T`.

Fixes #1731
Fixes #1733

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -1006,9 +1006,10 @@ namespace System.Collections.Immutable
         [Pure]
         public override bool Equals(object obj)
         {
-            if (obj is ImmutableArray<T>)
+            IImmutableArray other = obj as IImmutableArray;
+            if (other != null)
             {
-                return this.Equals((ImmutableArray<T>)obj);
+                return this.array == other.Array;
             }
 
             return false;
--- a/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
+++ b/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
@@ -251,7 +251,8 @@ namespace System.Collections.Immutable.Test
             ImmutableArray<string> derivedImmutable = ImmutableArray.Create("a", "b", "c");
             ImmutableArray<object> baseImmutable = derivedImmutable.As<object>();
             Assert.False(baseImmutable.IsDefault);
-            Assert.Equal(derivedImmutable, baseImmutable);
+            // Must cast to object or the IEnumerable<object> overload of Equals would be used
+            Assert.Equal((object)derivedImmutable, baseImmutable, EqualityComparer<object>.Default);
 
             // Make sure we can reverse that, as a means to verify the underlying array is the same instance.
             ImmutableArray<string> derivedImmutable2 = baseImmutable.As<string>();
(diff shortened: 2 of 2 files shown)


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -1006,9 +1006,10 @@ public override int GetHashCode()
         [Pure]
         public override bool Equals(object obj)
         {
-            if (obj is ImmutableArray<T>)
+            IImmutableArray other = obj as IImmutableArray;
```
sharwell: I removed the check for `ImmutableArray<T>` altogether since `obj` is already boxed on entering this method.

src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
```diff
@@ -251,7 +251,8 @@ public void Covariance()
             ImmutableArray<string> derivedImmutable = ImmutableArray.Create("a", "b", "c");
             ImmutableArray<object> baseImmutable = derivedImmutable.As<object>();
             Assert.False(baseImmutable.IsDefault);
-            Assert.Equal(derivedImmutable, baseImmutable);
+            // Must cast to object or the IEnumerable<object> overload of Equals would be used
+            Assert.Equal((object)derivedImmutable, baseImmutable);
```
stephentoub: I think you're going to need to change these Equals to something like:
```C#
Assert.Equal((object)derivedImmutable, baseImmutable, EqualityComparer<object>.Default);
```
Otherwise, the default comparer that xunit uses will I believe compare the concrete types and fail the comparison as a result.

sharwell: Seems you are correct. I was using this during my tests because the first thing I tried was this:

```csharp
Assert.Equal(derivedImmutable, baseImmutable, EqualityComparer<object>.Default);
```

This did not resolve the problem (it resolves to `Equal<T>(IEnumerable<T>, IEnumerable<T>, IEqualityComparer<T>`). I didn't realize that both the cast and the explicit comparer were the reason it started to work and removed it. Back now though. :smile:

stephentoub:
LGTM.  Thanks for the fix.  @AArnott?

sharwell:
BTW, if you decide that the behavior I mentioned in #1733 is *by design*, then I'll submit a different fix for #1731 that compares the underlying arrays directly instead of assuming `ImmutableArray<T>.Equals` will provide the answer.

AArnott:
:+1:

VSadov:
Nice!

PR #1736, opened 2015-05-13
Make {Process/ProcessThread].PriorityBoostEnabled consistent on Unix

### LINKED ISSUE #1726: Make Process.PriorityBoostEnabled and ProcessThread.PriorityBoostEnabled consistent on Unix
On Unix, Process.set_PriorityBoostEnabled is a nop, whereas ProcessThread.set_PriorityBoostEnabled throws PlatformNotSupportedException.  Whichever behavior we choose (probably nop in this case), we should be consistent across these two setters.

stephentoub:
This property isn't supported on Unix, but how we don't support it currently differs between Process and ProcessThread: the former makes it a nop, whereas the latter makes it throw an exception.  This commit just makes the behavior consistent, having both be a nop, since ignoring this has limited impact on the consuming application.

Fixes #1726.

### DIFF
--- a/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessThread.Unix.cs
+++ b/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessThread.Unix.cs
@@ -30,7 +30,7 @@ namespace System.Diagnostics
         private bool PriorityBoostEnabledCore
         {
             get { return false; }
-            set { throw new PlatformNotSupportedException(); }
+            set { } // Nop
         }
 
         /// <summary>


stephentoub:
cc: @sokket

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessThread.Unix.cs
```diff
@@ -30,7 +30,7 @@ public void ResetIdealProcessor()
         private bool PriorityBoostEnabledCore
         {
             get { return false; }
-            set { throw new PlatformNotSupportedException(); }
```
sokket: Are any tests expecting this to throw?

sokket:
LGTM

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessThread.Unix.cs
stephentoub: Fortunately or unfortunately, depending on how you look at it, no ;)  The Process library is an area where we have a fairly significant test gap.

stephentoub:
Thanks, Jon.

PR #1737, opened 2015-05-13
Add System.Security.Cryptography.Hashing.Algorithms Source and Test

bartonjs:
Resolves https://github.com/dotnet/corefx/issues/836.

### DIFF
--- a/src/Common/src/Interop/Windows/BCrypt/Cng.cs
+++ b/src/Common/src/Interop/Windows/BCrypt/Cng.cs
@@ -1,0 +1,394 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text;
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Runtime.InteropServices;
+
+namespace Internal.NativeCrypto
+{
+    internal static partial class BCryptNative
+    {
+        /// <summary>
+        ///     Well known algorithm names
+        /// </summary>
+        internal static class AlgorithmName
+        {
+            public const string ECDHP256 = "ECDH_P256";         // BCRYPT_ECDH_P256_ALGORITHM
+            public const string ECDHP384 = "ECDH_P384";         // BCRYPT_ECDH_P384_ALGORITHM
+            public const string ECDHP521 = "ECDH_P521";         // BCRYPT_ECDH_P521_ALGORITHM
+            public const string ECDsaP256 = "ECDSA_P256";       // BCRYPT_ECDSA_P256_ALGORITHM
+            public const string ECDsaP384 = "ECDSA_P384";       // BCRYPT_ECDSA_P384_ALGORITHM
+            public const string ECDsaP521 = "ECDSA_P521";       // BCRYPT_ECDSA_P521_ALGORITHM
+            public const string MD5 = "MD5";                    // BCRYPT_MD5_ALGORITHM
+            public const string Sha1 = "SHA1";                  // BCRYPT_SHA1_ALGORITHM
+            public const string Sha256 = "SHA256";              // BCRYPT_SHA256_ALGORITHM
+            public const string Sha384 = "SHA384";              // BCRYPT_SHA384_ALGORITHM
+            public const string Sha512 = "SHA512";              // BCRYPT_SHA512_ALGORITHM
+        }
+
+        /// <summary>
+        ///     Magic numbers identifying blob types
+        /// </summary>
+        internal enum KeyBlobMagicNumber {
+            ECDHPublicP256 = 0x314B4345,                        // BCRYPT_ECDH_PUBLIC_P256_MAGIC
+            ECDHPublicP384 = 0x334B4345,                        // BCRYPT_ECDH_PUBLIC_P384_MAGIC
+            ECDHPublicP521 = 0x354B4345,                        // BCRYPT_ECDH_PUBLIC_P521_MAGIC
+            ECDsaPublicP256 = 0x31534345,                       // BCRYPT_ECDSA_PUBLIC_P256_MAGIC
+            ECDsaPublicP384 = 0x33534345,                       // BCRYPT_ECDSA_PUBLIC_P384_MAGIC
+            ECDsaPublicP521 = 0x35534345                        // BCRYPT_ECDSA_PUBLIC_P521_MAGIC
+        }
+
+        internal static class KeyDerivationFunction
+        {
+            public const string Hash = "HASH";                  // BCRYPT_KDF_HASH
+            public const string Hmac = "HMAC";                  // BCRYPT_KDF_HMAC
+            public const string Tls = "TLS_PRF";                // BCRYPT_KDF_TLS_PRF
+        }
+    }
+
+    //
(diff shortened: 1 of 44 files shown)


src/Common/src/Interop/Windows/BCrypt/Cng.cs
```diff
@@ -0,0 +1,394 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using global::System;
```
stephentoub: Can we clean up these "global::"s?  It's ok to do so in a separate commit.

src/Common/src/Interop/Windows/BCrypt/Cng.cs
```diff
@@ -0,0 +1,394 @@
...
+            public const string Hash = "HASH";                  // BCRYPT_KDF_HASH
+            public const string Hmac = "HMAC";                  // BCRYPT_KDF_HMAC
+            public const string Tls = "TLS_PRF";                // BCRYPT_KDF_TLS_PRF
+        }
+    }
+
+    //
+    // Interop layer around Windows CNG api.
+    //
+    internal static partial class Cng
```
stephentoub: In general there's probably some clean up to do in these interop files.  I'd suggest just opening an issue to deal with it separately.

src/System.Security.Cryptography.Hashing.Algorithms/tests/MD5Tests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            return MD5.Create();
+        }
+
+        // Test cases are defined in RFC 1321, section A.5
+
+        [Fact]
+        public void MD5_Rfc1321_1()
+        {
+            Verify("", "d41d8cd98f00b204e9800998ecf8427e");
```
stephentoub: As I mentioned offline, we could consider using [Theory]/[InlineData] for these, but I understand we're currently limited with the internal test harness.

src/System.Security.Cryptography.Hashing.Algorithms/tests/System.Security.Cryptography.Hashing.Algorithms.Tests.csproj
```diff
@@ -0,0 +1,47 @@
...
+    <RootNamespace>System.Security.Cryptography.Hashing.Algorithms.Tests</RootNamespace>
+  </PropertyGroup>
+  <!-- Default configurations to help VS understand the configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- Compile tests against the contract, but copy our local-built implementation for testing -->
+    <ProjectReference Include="..\src\System.Security.Cryptography.Hashing.Algorithms.csproj">
```
stephentoub: This is the mechanism used for partial facades, which this isn't, right?  We should be able to use a normal P2P reference without the rest of the below gook.

stephentoub:
Other than the structure of the P2P reference in the tests project, which I think can be updated to match how P2P references are done in other non-partial facades, LGTM.

src/Common/src/Interop/Windows/BCrypt/Cng.cs
bartonjs: Fixed, squashed commit.

src/Common/src/Interop/Windows/BCrypt/Cng.cs
bartonjs: Created https://github.com/dotnet/corefx/issues/1739 to track the cleanup.

src/System.Security.Cryptography.Hashing.Algorithms/tests/System.Security.Cryptography.Hashing.Algorithms.Tests.csproj
bartonjs: Must have picked a bad example at random.  Changed the reference style to match System.Xml.XmlDocument and System.Diagnostics.FileVersionInfo. (Squashed commit)

stephentoub:
Thanks. LGTM.

src/System.Security.Cryptography.Hashing.Algorithms/src/System.Security.Cryptography.Hashing.Algorithms.csproj
```diff
@@ -0,0 +1,78 @@
...
+    <CLSCompliant>false</CLSCompliant>
+  </PropertyGroup>
+  <!-- Help VS understand available configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU' " />
+  <!-- References are resolved from packages.config -->
```
akoeplinger: ~~packages.config~~ project.json. Should probably also move it to the respective ItemGroup below.

src/System.Security.Cryptography.Hashing.Algorithms/tests/System.Security.Cryptography.Hashing.Algorithms.Tests.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <Compile Include="HmacTests.cs" />
+    <Compile Include="MD5Tests.cs" />
+    <Compile Include="Rfc2202HmacTests.cs" />
+    <Compile Include="Rfc4231HmacTests.cs" />
+    <Compile Include="Sha1Tests.cs" />
+    <Compile Include="Sha256Tests.cs" />
+    <Compile Include="Sha384Tests.cs" />
+    <Compile Include="Sha512Tests.cs" />
+  </ItemGroup>
+  <!-- References are resolved from packages.config -->
```
akoeplinger: ~~packages.config~~ project.json

stephentoub:
Jeremy, ideally we'll get the Linux CI machines configured correctly for this asap, but if you don't want to block and would prefer to check in, you could temporarily suppress the Linux tests by adding a line like this to the test project:
https://github.com/dotnet/corefx/blob/master/src/Microsoft.Win32.Registry/tests/Microsoft.Win32.Registry.Tests.csproj#L11

src/System.Security.Cryptography.Hashing.Algorithms/src/System.Security.Cryptography.Hashing.Algorithms.csproj
bartonjs: Did a check across the repo and it's hit-or-miss on whether the comment was there at all (the location here is where VS would normally write in the references), so I went ahead and just removed the lines in both files.

PR #1740, opened 2015-05-13
Implement Debug.ShowAssertDialog on Unix

### LINKED ISSUE #1724: Implement Debug.ShowAssertDialog on Unix
Failing asserts currently throw NotImplementedException due to this not being implemented.

stephentoub:
Provides a slightly better ShowAssertDialog implementation.  While we will probably want to do something better once we have better debugger integration, for now this avoids throwing an exception and instead ensures that the assert information is written out to stderr, making assertion failures visible at the console and in CI logs.

Fixes #1724.


stephentoub:
cc: @ellismg

ellismg:
LGTM

stephentoub:
Thanks, Matt.

PR #1743, opened 2015-05-14
Cleanup AsyncStreamReader

### LINKED ISSUE #1734: Cleanup/fix Process's AsyncStreamReader
This implementation has a variety of issues, some worse than others:
- It invokes the user's delegate while holding a lock.  In certain circumstances this could lead to deadlock.
- Calling CancelOperation() doesn't actually cancel: it prevents the delegate from being invoked again, but it continues to read asynchronously until the end of the stream.
- It's got a bunch of dead code: its IDisposable implementation, Close, CurrentEncoding, BaseStream, etc.
- Its async reading implementation could be simplified using async/await.

Maxwe11:
fix #1734
first attempt :smile: 
will reword  commit messages in case if attempt is viable

### DIFF
--- a/src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
+++ b/src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
@@ -13,6 +13,7 @@
 
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -40,10 +39,8 @@ internal sealed class AsyncStreamReader : IDisposable
         private readonly Process _process;
 
         // Delegate to call user function.
-        private Action<string> _userCallBack;
+        private readonly Action<string> _userCallBack;
 
-        // Internal Cancel operation
-        private bool _cancelOperation;
         private ManualResetEvent _eofEvent;
```
stephentoub: Consider changing this to be a ManualResetEventSlim.  With the usage patterns here, it won't allocate a ManualResetEvent such that the impact of not disposing the event will be minimized. In general an MRES is lighter weight than an MRE, hence the "Slim".

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -83,82 +78,28 @@ internal AsyncStreamReader(Process process, Stream stream, Action<string> callba
...
         }
 
-        internal void CancelOperation()
-        {
-            _cancelOperation = true;
-        }
-
         // This is the async callback function. Only one thread could/should call this.
-        private void ReadBuffer(Task<int> t)
+        private async void ReadBuffer(CancellationToken token)
```
stephentoub: I think this function needs to main changes:
- The body should be wrapped in a Task.Run(async delegate { ... }).  That's because this call to ReadBuffer could be happening on the caller's thread, and without the ContinueWith forcing subsequent reads to run asynchronously, it's theoretically possible that this could become a long-running synchronous call.  With that, you also don't need the _eofEvent at all... you just store the scheduled Task, and the WaitForExit method just waits on the Task.
- The recursive call to ReadBuffer should be removed.  Instead, this function should just contain a loop: read, process, read, process, etc., and at some point either because 0 bytes were read or an exception occurred, the loop is exited and the ending code runs to flush the message queue.

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -215,9 +156,7 @@ private void ReadBuffer(Task<int> t)
         // immediately followed by a line feed. The resulting string does not
         // contain the terminating carriage return and/or line feed. The returned
         // value is null if the end of the input stream has been reached.
-        //
-
-        private void GetLinesFromStringBuilder()
+        private void GetLinesFromStringBuilder(CancellationToken token)
```
stephentoub: Could we rename this method?  I find the existing name to be less than explanatory.  How about MoveLinesFromStringBuilderToMessageQueue?

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -280,10 +219,10 @@ private void GetLinesFromStringBuilder()
                 _currentLinePos = 0;
             }
 
-            FlushMessageQueue();
+            FlushMessageQueue(token);
         }
 
-        private void FlushMessageQueue()
+        private void FlushMessageQueue(CancellationToken token)
```
stephentoub: Rather than passing the CancellationToken around everywhere, what would you think about passing it just into the ctor and storing it as a readonly instance field?

src/System.Diagnostics.Process/src/System/Diagnostics/Process.cs
```diff
@@ -72,8 +72,10 @@ public partial class Process : IDisposable
         public event DataReceivedEventHandler ErrorDataReceived;
 
         // Abstract the stream details
-        internal AsyncStreamReader _output;
-        internal AsyncStreamReader _error;
+        private AsyncStreamReader _output;
+        private CancellationTokenSource _outputCts;
```
stephentoub: As an alternative to having these two fields here, you could instead just add the _cts field to the AsyncStreamReader and construct the cts in the AsyncStreamReader's ctor, then expose a Cancel() method that just calls _cts.Cancel.  That saves the duplication here, and addresses my previous comment about passing around the CancellationToken.

stephentoub:
Thanks, @Maxwe11.  Good start.  I'll take another pass through once you've addressed some of the initial feedback.

@pallavit, what do you think?

Maxwe11:
updated

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -19,126 +19,65 @@
...
-            if (_cancelOperation)
-            {
-                _cancelOperation = false;
-            }
-
             if (_sb == null)
             {
                 _sb = new StringBuilder(DefaultBufferSize);
-                _stream.ReadAsync(_byteBuffer, 0, _byteBuffer.Length).ContinueWith(ReadBuffer, CancellationToken.None, TaskContinuationOptions.None, TaskScheduler.Default);
+                _readToBufferTask = Task.Run(new Action(ReadBuffer), _cts.Token);
```
stephentoub: I don't think we want to pass the token to the task like this.  Doing so could cause the task to end in the Canceled state, which would then cause the Wait on it in WaitUntilEOF to throw an exception, which I believe would then propagate out of WaitForExit.  Better to just not make the task cancelable.

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -148,76 +87,67 @@ internal void BeginReadLine()
 
         internal void CancelOperation()
         {
-            _cancelOperation = true;
+            _cts.Cancel();
         }
 
         // This is the async callback function. Only one thread could/should call this.
-        private void ReadBuffer(Task<int> t)
+        private async void ReadBuffer()
```
stephentoub: This should be "async Task" rather than "async void".

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -148,76 +87,67 @@ internal void BeginReadLine()
...
                 }
-            }
-            else
-            {
+
+                if (byteLen == 0)
+                {
+                    // We're at EOF, we won't start the next iteration so we process current buffer content,
+                    // flush message queue and return
+                    lock (_messageQueue)
```
stephentoub: You might disagree, but I was thinking it'd be cleaner to just move all of this to after the while loop.  then the catch blocks would just "break;", in the try above we'd also break if byteLen == 0, and the logic around getting chars / appending / moving lines would also be moved into the try.

Maxwe11: >it'd be cleaner to just move all of this to after the while loop. then the catch blocks would just "break;", in the try above we'd also break if byteLen == 0

agree

>and the logic around getting chars / appending / moving lines would also be moved into the try.

disagree, because `MoveLinesFromStringBuilderToMessageQueue()` calls up `FlushMessageQueue()` and that in turn calls user's delegate which could throw.

Maxwe11: I moved call `FlushMessageQueue()` from `MoveLinesFromStringBuilderToMessageQueue()` and put it directly in `ReadBuffer()` after `MoveLinesFromStringBuilderToMessageQueue()`

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -148,76 +88,64 @@ internal void BeginReadLine()
...
-                finally
+                catch (OperationCanceledException)
                 {
-                    _eofEvent.Set();
+                    // We should consume any OperationCanceledException from child read here
+                    // so that we don't crash the parent with an unhandled exc
+                    break; // Treat this as EOF
                 }
+
+                FlushMessageQueue();
```
stephentoub: Why wouldn't this be inside the try as well?

stephentoub: There is now a subtle behavioral difference between the old and new implementations.  In the old implementation, any exceptions from within the callback would be eaten.  In the new implementation, any exceptions from within the callback will end up getting propagated out of WaitForEOF when the task is waited on. And both the old and the new behaviors are different from desktop, where any exception from the user callback will likely go unhandled on an arbitrary thread pool thread.  @pallavit, we should decide what the right behavior here is; if we want to match the previous .NET Core behavior, that's easy, we just wrap a try/catch around the body.  If we want to match the desktop behavior, we'd likely need to queue a work item that just re-throws the exception.

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -318,11 +250,10 @@ private void FlushMessageQueue()
         // We will lose some information if we don't do this.
         internal void WaitUtilEOF()
         {
-            if (_eofEvent != null)
+            if (_readToBufferTask != null)
             {
-                _eofEvent.WaitOne();
-                _eofEvent.Dispose();
-                _eofEvent = null;
+                _readToBufferTask.Wait();
```
stephentoub: I'd suggest changing this to be .GetAwaiter().GetResult() instead of .Wait().  The only difference here is around what exception propagates out if the task is faulted, and that way if we do decide to let the exception from the user callback propagate out here, it'll be the exact same exception object rather than being wrapped in an aggregate.

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
Maxwe11: Got it.
Will wait for your decision.

pallavit: The current .NET Core behavior is an oversight and not intentional. I don't think we should swallow the exception and should propagate it via a thread pool thread (desktop behavior). This pattern seems more intuitive. @stephentoub Thoughts?

Maxwe11:
@dotnet-bot test this please

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -292,37 +223,68 @@ private void FlushMessageQueue()
...
+            try
+            {
+                _userCallBack(argument);
+            }
+            catch (Exception e)
+            {
+                if (rethowInNewthread)
+                {
+                    ExceptionDispatchInfo edi = ExceptionDispatchInfo.Capture(e);
+                    Task.Run(() => edi.Throw());
```
stephentoub: This will need to be ThreadPool.QueueUserWorkItem rather than Task.Run, as Task.Run will store the exception into the returned task (which is ignored).

Maxwe11: fixed

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -13,211 +13,147 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace System.Diagnostics
 {
-    internal sealed class AsyncStreamReader : IDisposable
+    internal sealed class AsyncStreamReader
     {
-        internal const int DefaultBufferSize = 1024;  // Byte buffer size
+        private const int DefaultBufferSize = 1024;  // Byte buffer size
```
whoisj: Out of curiosity, and the PR author likely isn't the best person to answer this, why is this buffer so incredibly tiny? Given that a page is generally 4096 bytes, shouldn't it at least be a page in size.

Given [TN I/O Request Size](https://technet.microsoft.com/en-us/library/cc938632.aspx) article, the ideal size for reading/writing to IO is generally 64 KB. Wouldn't increasing the size of this buffer improve performance in general?

Am I missing something? (likely)

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -13,211 +13,147 @@
...
         internal void BeginReadLine()
         {
-            if (_cancelOperation)
-            {
-                _cancelOperation = false;
-            }
-
             if (_sb == null)
             {
                 _sb = new StringBuilder(DefaultBufferSize);
```
whoisj: The constant has a comment stating that it is intended to be the buffer size in bytes, yet a StringBuilder is being allocated with this size. Given that sizeof(char) == 2, shouldn't this be `DefaultBufferSize / sizeof(char)`?

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -279,11 +215,9 @@ private void GetLinesFromStringBuilder()
                 _sb.Length = 0;
```
whoisj: The `StringBuilder` class exposes a `Clear()` method that is the equivollent of setting `Length` to zero, but is clearer in its intention.

whoisj:
Added a few nits, don't let those block these good changes.

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
```diff
@@ -13,211 +13,147 @@
...
-            if (_cancelOperation)
-            {
-                _cancelOperation = false;
-            }
-
             if (_sb == null)
             {
                 _sb = new StringBuilder(DefaultBufferSize);
-                _stream.ReadAsync(_byteBuffer, 0, _byteBuffer.Length).ContinueWith(ReadBuffer, CancellationToken.None, TaskContinuationOptions.None, TaskScheduler.Default);
+                _readToBufferTask = Task.Run(() => ReadBuffer());
```
stephentoub: We could avoid a few allocations here by changing this to be:
```
_readToBufferTask = Task.Factory.StartNew(s => ((AsyncStreamReader)s).ReadBuffer(),
    this, CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
```
That'll avoid the closure allocation and associated delegate allocation per BeginReadLine call.

stephentoub:
Looking good.  Left one more relatively minor comment.

Maxwe11:
@dotnet-bot test this please

stephentoub:
Thanks, @Maxwe11. @pallavit, look good to you?

src/System.Diagnostics.Process/src/System/Diagnostics/AsyncStreamReader.cs
pallavit: I do not have any historical context on this but in general since the code is shared across different platforms I have seen us being conservative in buffer sizes. Do you know how much of a performance gain are we likely to see? We can surely consider making the changes depending on the gain.

pallavit:
@Maxwe11 Thanks for doing this. LGTM. Stephen let me know if you have any further comments before I merge.

PR #1746, opened 2015-05-14
Add System.Security.Cryptography.Encryption Source and Test

bartonjs:
Resolves https://github.com/dotnet/corefx/issues/834

### DIFF
--- a/src/System.Security.Cryptography.Encryption/System.Security.Cryptography.Encryption.sln
+++ b/src/System.Security.Cryptography.Encryption/System.Security.Cryptography.Encryption.sln
@@ -1,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.22911.2
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Security.Cryptography.Encryption", "src\System.Security.Cryptography.Encryption.csproj", "{D04A73AE-E418-4ACD-A132-7688435BE8B5}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Security.Cryptography.Encryption.Tests", "tests\System.Security.Cryptography.Encryption.Tests.csproj", "{54C20799-0F9C-4921-81E0-B3BEB73B451A}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{D04A73AE-E418-4ACD-A132-7688435BE8B5}.Debug|Any CPU.ActiveCfg = Windows_Debug|Any CPU
+		{D04A73AE-E418-4ACD-A132-7688435BE8B5}.Debug|Any CPU.Build.0 = Windows_Debug|Any CPU
+		{D04A73AE-E418-4ACD-A132-7688435BE8B5}.Release|Any CPU.ActiveCfg = Windows_Release|Any CPU
+		{D04A73AE-E418-4ACD-A132-7688435BE8B5}.Release|Any CPU.Build.0 = Windows_Release|Any CPU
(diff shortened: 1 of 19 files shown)


src/System.Security.Cryptography.Encryption/System.Security.Cryptography.Encryption.sln
```diff
@@ -0,0 +1,28 @@
...
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Security.Cryptography.Encryption.Tests", "tests\System.Security.Cryptography.Encryption.Tests.csproj", "{54C20799-0F9C-4921-81E0-B3BEB73B451A}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{D04A73AE-E418-4ACD-A132-7688435BE8B5}.Debug|Any CPU.ActiveCfg = Windows_Debug|Any CPU
```
stephentoub: Can we add configs for the other OSes, too? e.g. https://github.com/dotnet/corefx/blob/master/src/System.Console/System.Console.sln#L26-L49

stephentoub: Or, actually, does this library have any platform-specific code such that it needs platform-specific builds?  If not, we can just reduce this down to Debug and Release, and similarly remove the configs from the .csproj file.

src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -0,0 +1,456 @@
...
+namespace System.Security.Cryptography
+{
+    [System.Runtime.InteropServices.ComVisible(true)]
+    public class CryptoStream : Stream, IDisposable
+    {
+        // Member veriables
+        private readonly Stream _stream;
+        private readonly ICryptoTransform _Transform;
+        private byte[] _InputBuffer;  // read from _stream before _Transform
+        private int _InputBufferIndex = 0;
```
stephentoub: Nit: we don't need to initialize these to their default values (or if we want to for some reason, we should be consistent)

src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -0,0 +1,456 @@
...
+
+namespace System.Security.Cryptography
+{
+    [System.Runtime.InteropServices.ComVisible(true)]
+    public class CryptoStream : Stream, IDisposable
+    {
+        // Member veriables
+        private readonly Stream _stream;
+        private readonly ICryptoTransform _Transform;
+        private byte[] _InputBuffer;  // read from _stream before _Transform
```
stephentoub: Some of these, in particular ones that begin with 'i', start with a capital.

src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -0,0 +1,456 @@
...
+            else
+            {
+                Buffer.BlockCopy(_OutputBuffer, 0, buffer, currentOutputIndex, _OutputBufferIndex);
+                bytesToDeliver -= _OutputBufferIndex;
+                _OutputBufferIndex = 0;
+                return (count - bytesToDeliver);
+            }
+        }
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```
stephentoub: Why do we override this if we then just delegate to the base?

src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/SymmetricAlgorithm.cs
```diff
@@ -0,0 +1,207 @@
...
+            }
+        }
+
+        public virtual KeySizes[] LegalKeySizes
+        {
+            get
+            {
+                // Desktop compat: Unless derived classes set the protected field "LegalKeySizesValue" to a non-null value, a NullReferenceException is what you get.
+                // In the Win8P profile, the "LegalKeySizesValue" field has been removed. So derived classes must override this property for the class to be any of any use.
+                throw new NullReferenceException();
```
stephentoub: Oh my.

stephentoub:
A few minor comments (which could be addressed as necessary separately), but otherwise LGTM.

PR #1747, opened 2015-05-14
Complete System.IO.Pipes tests

### LINKED ISSUE #1142: Improve code coverage for System.IO.Pipes
According to the current code coverage report, this library is only at ~25%. Our general goal is closer to 90%, though this will depend on the library. We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

bpschoch:
Added more functional testing
Added/reworked parameter checking throws testing
One test ServerCloneTests(), currently hangs but is marked to skip
   until issue is resolved

Fix #1142

### DIFF
--- a/src/System.IO.Pipes/tests/AnonymousPipesSimpleTest.cs
+++ b/src/System.IO.Pipes/tests/AnonymousPipesSimpleTest.cs
@@ -15,7 +15,7 @@ public class AnonymousPipesSimpleTest
         using (AnonymousPipeServerStream server = new AnonymousPipeServerStream(PipeDirection.Out))
         {
             Assert.True(server.IsConnected);
-            using (AnonymousPipeClientStream client = new AnonymousPipeClientStream(PipeDirection.In, server.ClientSafePipeHandle))
+            using (AnonymousPipeClientStream client = new AnonymousPipeClientStream(PipeDirection.In, server.GetClientHandleAsString()))
             {
                 Assert.True(server.IsConnected);
                 Assert.True(client.IsConnected);
@@ -27,6 +27,57 @@ public class AnonymousPipesSimpleTest
                 Assert.Equal(1, client.Read(received, 0, 1));
                 Assert.Equal(sent[0], received[0]);
             }
+            Assert.Throws<System.IO.IOException>(() => server.WriteByte(5));
+        }
+    }
+
+    [Fact]
+    public static void ServerSendsByteClientReceivesServerClone()
+    {
+        using (AnonymousPipeServerStream serverBase = new AnonymousPipeServerStream(PipeDirection.Out))
+        {
+            using (AnonymousPipeServerStream server = new AnonymousPipeServerStream(PipeDirection.Out, serverBase.SafePipeHandle, serverBase.ClientSafePipeHandle))
+            {
+                Assert.True(server.IsConnected);
+                using (AnonymousPipeClientStream client = new AnonymousPipeClientStream(PipeDirection.In, server.GetClientHandleAsString()))
+                {
+                    Assert.True(server.IsConnected);
+                    Assert.True(client.IsConnected);
+
+                    byte[] sent = new byte[] { 123 };
+                    byte[] received = new byte[] { 0 };
+                    server.Write(sent, 0, 1);
+
+                    Assert.Equal(1, client.Read(received, 0, 1));
+                    Assert.Equal(sent[0], received[0]);
+                }
+                Assert.Throws<System.IO.IOException>(() => server.WriteByte(5));
+            }
+        }
+    }
+
+    [Fact]
+    public static void ServerSendsByteClientReceivesAsync()
+    {
+        using (AnonymousPipeServerStream server = new AnonymousPipeServerStream(PipeDirection.Out))
+        {
+            Assert.True(server.IsConnected);
+            using (AnonymousPipeClientStream client = new AnonymousPipeClientStream(server.GetClientHandleAsString()))
+            {
+                Assert.True(server.IsConnected);
+                Assert.True(client.IsConnected);
+
+                byte[] sent = new byte[] { 123 };
+                byte[] received = new byte[] { 0 };
+                Task writeTask = server.WriteAsync(sent, 0, 1);
+                writeTask.Wait();
+
+                Task<int> readTask = client.ReadAsync(received, 0, 1);
+                readTask.Wait();
+
+                Assert.Equal(1, readTask.Result);
+                Assert.Equal(sent[0], received[0]);
+            }
         }
     }
 
@@ -36,18 +87,25 @@ public class AnonymousPipesSimpleTest
         using (AnonymousPipeServerStream server = new AnonymousPipeServerStream(PipeDirection.In))
         {
             Assert.True(server.IsConnected);
+            server.ReadMode = PipeTransmissionMode.Byte;
+
             using (AnonymousPipeClientStream client = new AnonymousPipeClientStream(PipeDirection.Out, server.ClientSafePipeHandle))
             {
                 Assert.True(server.IsConnected);
                 Assert.True(client.IsConnected);
+                client.ReadMode = PipeTransmissionMode.Byte;
 
                 byte[] sent = new byte[] { 123 };
                 byte[] received = new byte[] { 0 };
                 client.Write(sent, 0, 1);
 
+                server.DisposeLocalCopyOfClientHandle();
+
                 Assert.Equal(1, server.Read(received, 0, 1));
                 Assert.Equal(sent[0], received[0]);
             }
+            // not sure why the following isn't thrown because pipe is broken
+            //Assert.Throws<System.IO.IOException>(() => server.ReadByte());
         }
     }
 
--- a/src/System.IO.Pipes/tests/NamedPipesSimpleTest.cs
+++ b/src/System.IO.Pipes/tests/NamedPipesSimpleTest.cs
@@ -8,56 +8,171 @@ using Xunit;
 
 public class NamedPipesSimpleTest
 {
-    [Fact]
-    public static void ServerSendsByteClientReceives()
+    static byte[] ReadBytes(PipeStream pipeStream, int length)
     {
-        using (NamedPipeServerStream server = new NamedPipeServerStream("foo", PipeDirection.Out))
+        Assert.True(pipeStream.IsConnected);
+
+        byte[] buffer = new byte[length];
+        Assert.True(length > 0);
+
+        buffer[0] = (byte)pipeStream.ReadByte();
+        if (length > 1)
         {
-            byte[] sent = new byte[] { 123 };
-            byte[] received = new byte[] { 0 };
-            Task t = Task.Run(() =>
-            {
-                using (NamedPipeClientStream client = new NamedPipeClientStream(".", "foo", PipeDirection.In))
-                {
-                    client.Connect();
-                    Assert.True(client.IsConnected);
+            int len = pipeStream.Read(buffer, 1, length - 1);
+            Assert.Equal(length - 1, len);
+        }
 
-                    Assert.Equal(1, client.Read(received, 0, 1));
-                }
-            });
-            server.WaitForConnection();
-            Assert.True(server.IsConnected);
+        return buffer;
+    }
 
-            server.Write(sent, 0, 1);
+    static void WriteBytes(PipeStream pipeStream, byte[] buffer)
+    {
+        Assert.True(pipeStream.IsConnected);
+        Assert.True(buffer.Length > 0);
 
-            t.Wait();
-            Assert.Equal(sent[0], received[0]);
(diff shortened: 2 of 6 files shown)


Maxwe11:
>One test ServerCloneTests(), currently hangs but is marked to skip until issue is resolved

What issue do mean?

bpschoch:
NamedPipesSimpleTests.cs line 395 (ServerCloneTests()) which tests cloning a NamedPipeServerStream ends up hanging.  Perhaps it's a lack of understanding on the clone capabilities or perhaps it's a real bug.  Didn't know how to proceed further on it.

stephentoub:
There was one unrelated failure on Windows in the CI build, but a bunch of relevant failures on Linux.  Some of these could be product issues, of course, but it'd be good to understand the nature of each failure: for product bugs, we can suppress the test on Unix with [ActiveIssue(..., PlatformID.AnyUnix], but we should fix any test bugs (if a test itself is Windows-specific, it can be attributed with [PlatformSpecific(PlatformID.Windows)]).
```
13:30:02 Starting:    System.IO.FileSystem.Tests
13:30:02 [39;49m[31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerAfterDisconnectThrows [FAIL]
13:30:02 [37m      Assert.Throws() Failure
13:30:02       Expected: typeof(System.InvalidOperationException)
13:30:02       Actual:   (No exception was thrown)
13:30:02 [30m      Stack Trace:
13:30:02 [37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerAfterDisconnectThrows()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerWhenDisplosedThrows [FAIL]
13:30:03 [37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.ObjectDisposedException)
13:30:03       Actual:   typeof(System.InvalidOperationException): Pipe handle has not been set.  Did your PipeStream implementation call InitializeHandle?
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.NamedPipeServerStream.CheckDisconnectOperations()
13:30:03             at System.IO.Pipes.NamedPipeServerStream.Disconnect()
13:30:03             at System.IO.Pipes.Tests.NamedPipesThrowsTests.<>c__DisplayClass64.<ServerWhenDisplosedThrows>b__5a()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerVerifyMultipleServerInconsistantType [FAIL]
13:30:03 [37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.UnauthorizedAccessException)
13:30:03       Actual:   (No exception was thrown)
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerVerifyMultipleServerInconsistantType()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerVerifyMultipleServerOverflow [FAIL]
13:30:03 [37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.IO.IOException)
13:30:03       Actual:   (No exception was thrown)
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerVerifyMultipleServerOverflow()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerNotConnectedThrows [FAIL]
13:30:03 [37m      System.InvalidOperationException : Pipe handle has not been set.  Did your PipeStream implementation call InitializeHandle?
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.PipeStream.CheckPipePropertyOperations()
13:30:03             at System.IO.Pipes.PipeStream.get_TransmissionMode()
13:30:03             at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerNotConnectedThrows()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ClientDisconnectedPipeThrows [FAIL]
13:30:03 [37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.IO.IOException)
13:30:03       Actual:   (No exception was thrown)
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ClientDisconnectedPipeThrows()
13:30:03 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerDisconnectedPipeThrows [FAIL]
13:30:03 [37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.IO.IOException)
13:30:03       Actual:   (No exception was thrown)
13:30:03 [30m      Stack Trace:
13:30:03 [37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerDisconnectedPipeThrows()
13:30:03 Finished:    System.IO.Pipes.Tests
13:30:03 [37m
13:30:03 === TEST EXECUTION SUMMARY ===
13:30:03 [37m   System.IO.Pipes.Tests  Total: 97, Errors: 0, Failed: 7, Skipped: 1, Time: 1.014s
```

bpschoch:
ok I will look at these.  On the surface they appear to be issue relating to what happens on  a pipe when the other side shuts down the pipe.  It appears that unix may not detect the shutdown.  So this begins to get in a gray area of 'is that behavior part of the contract' or not.  Is there a general overall approach with these kinds of behaviors?

krwq:
General assumption is that correct behavior is the behavior of full .NET running on Windows. If the behavior is implementable on other platform it should behave exactly the same (unless you strongly believe that it is wrong then start a discussion). If not then we usually start an API compat discussion to get some advice what to do in this case (i.e. allow difference on different platforms or change both platforms or some other solution). In this particular case without investigating I believe it might be a product bug on linux although it might be as well a test bug where you accidentally don't close a handle (i.e. if there is inherited process it might inherit a handle and then you need to close it on both processes)

src/System.IO.Pipes/tests/NamedPipesSimpleTest.cs
```diff
@@ -224,4 +339,134 @@ public static void ClientPInvokeChecks()
...
+                Assert.True(server.IsMessageComplete);
+                Assert.Equal(msg1.Length, len3 + len4);
+                Assert.Equal(msg1, received3);
+
+                string userName = server.GetImpersonationUserName();    // not sure what to test here that will work in all cases
+            }
+        }
+    }
+
+    [Fact(Skip = "test hangs right now")]
```
stephentoub: If you believe this is a product bug, you can open an issue for it on GitHub, and then use:
[ActiveIssue(TheIssueNumberHere)]

src/System.IO.Pipes/tests/AnonymousPipesSimpleTest.cs
```diff
@@ -36,18 +87,25 @@ public static void ClientSendsByteServerReceives()
         using (AnonymousPipeServerStream server = new AnonymousPipeServerStream(PipeDirection.In))
         {
             Assert.True(server.IsConnected);
+            server.ReadMode = PipeTransmissionMode.Byte;    // cover that path
```
stephentoub: I'm not quite understanding this "cover" comments.  Were these lines just thrown in to boost code coverage?  if no such tests exist, it'd be better to have tests that not only execute this code but verify that the right things happened, e.g. that after setting ReadMode getting it returns the same value, that it has the appropriate impact on the rest of the library (if that's testable), that setting it to an invalid value yields the appropriate behavior, etc.  You might already have such tests, I was just thrown by the comment.

stephentoub:
> ok I will look at these.

Thanks.  If you can narrow things down quickly, please do so.  Otherwise, if these 7 tests are the only failures, I'd be ok with you opening an issue and tagging these as ActiveIssues on Unix, in the name of making forward progress and being able to check these tests in.  Then we can separately investigate those issues.

> If you still believe it is a product bug you may just open an issue or do a short investigation by checking searching if this is doable on linux (and see how) and then peek at the implementation and see if we are doing it correctly

Agreed.  There are a few places where we know we have some differences.  One's I'm aware of at the moment (some of these can be rectified, some of them will likely remain):
- Buffer sizes: we currently ignore requests to set a particular buffer size and we throw when trying to access the in/out buffer size (we should be able to support this, though, at least on Linux)
- Handle inheritability: without pipe2 on OSX, we don't support changing handle inheritability
- Connect: the timeout is ignored currently
- NumberOfServerInstances: we max out at 1, and throw from NumberOfServerInstances
- TransmissionMode: we only support Bytes
- PipeDirection.InOut: Windows and Unix pipes have subtly different behavior here we can't easily gloss over, in particular around when blocking happens
- Impersonation: not supported
- Cross-machine: not supported
- WaitForPipeDrain: not supported

> Complete System.IO.Pipes tests

Thanks very much for writing these.  Other than the aforementioned issue of understanding why these few tests are failing on Linux, LGTM.

bpschoch:
I'm going to have to setup a unix vm to really get to the bottom of these.  Also it doesn't help that the stacktrace doesn't show the line numbers of the failures (is that a bug? or forthcoming feature in unix?)
So I agree that in order to move forward it's best to do as you say, but.....

"I'd be ok with you opening an issue and tagging these as ActiveIssues on Unix, in the name of making forward progress and being able to check these tests in."

What are the mechanics to make them active issues on Unix?  Do I separate and duplicate the tests out and mark the duplicates as platform specific to unix along with the ActiveIssue tag?

On the 'coverage' I'll look at that in more detail also.  A lot of how pipes works (and the testing I'm doing) is not all methods/properties work (not fault) in all situations so that calls to a methods/properties and their not faulting is a valid test (vs faulting in other cases)

stephentoub:
> Also it doesn't help that the stacktrace doesn't show the line numbers of the failures (is that a bug? or forthcoming feature in unix?)

@janvorli?

> What are the mechanics to make them active issues on Unix? 

Add the attribute:
[ActiveIssue(123, PlatformID.AnyUnix)]
where 123 is the GitHub issue number.

> "faulting"

As @krwq noted, we are striving to have the behavior be as consistent as possible across implementations, and that includes for exceptional situations.  If an exception is being thrown on Windows but not on Unix, that's most likely a product bug; similarly if an exception is being thrown in both cases but of different types.  The more interesting cases are typically when an exception occurs on Unix and everything runs successfully on Windows, or when the results are different, or when one deadlocks and the other doesn't, etc.

bpschoch:
I added the issues and marked the tests and now it looks like it builds ok.

These tests now basically cover basically as much as can be covered without forcing error conditions (not sure how to do that).  Also there appears to be dead code e.g. private BeginRead()/EndRead() for async operations but code has path for non-async operations.

Also I took a 'gray box'  vs 'black box' approach in developing these tests in that I know that  i/o operations for both anonymous and named pipes go to a common PipeStream abstract class so for example the throws for i/o operations only appear on the anonymous classes and the various connect/disconnect tests appear on the named classes.  (from a black box perspective, they should be done in both and also on both the client/server sides)  Is this the correct approach?

Going forward, there are still probably some nooks and crannies that could be covered but I believe we're at the 90% plus level.

Where should I go next?  Do you need the UNIX issues debugged?  I've got some time as I'm in between jobs right now.

stephentoub:
> I added the issues and marked the tests and now it looks like it builds ok.

LGTM.  Thanks!

> Is this the correct approach?

Not ideal, but ok for now.  You could consider refactoring these into facts in a base class that both the named and anonymous test classes derive from, and add some sort of virtual method that gets an instance of the pipe; that virtual method can then be used by the base class' facts, and we'd get testing for both, in case in the future the implementations ever diverged.  If you want to do that, it can be done in a separate PR.

> I believe we're at the 90% plus level.

Excellent.  And while the number is nice, the more important thing is that it's helped to find some bugs.

> Where should I go next? Do you need the UNIX issues debugged?

Thanks!  There's plenty of stuff to do.  Feel free to look through the open issues, and in particular things marked as up-for-grabs.  If you'd like to investigate the Linux failures, that's good, too, otherwise I can.

stephentoub:
@bpschoch, a recent change to the project.json files made this non-mergeable in the GitHub UI.  Can you rebase your PR with the latest in master, and also fix up your addition to the test's project.json file?  It should now be using the latest version, notated with an asterisk... see https://github.com/dotnet/corefx/pull/1771 for examples.

bpschoch:
done

For the interim, I think i'm 'piped out' and will look at some of the other issues for now.

stephentoub:
Completely understand ;)

bpschoch:
i'll be working on doing more work on pipes in the background including making it more 'black box' by setting up a base class as you suggested for common tests and going back and re-examining all the states and making sure that all conditions are tested for all states for all 4 pipe types: Anonymous client/server and Named client/server.

Meanwhile, I've been able to complete some more unit tests for a few other parts of the library but I'm having trouble finding another good one to attack.  Let me know if you have any suggestions of what to attack next.

PR #1750, opened 2015-05-14
Multiple fixes to SecureString

stephentoub:
Fixes for a variety of issues (more details in the commit descriptions):
1. There's a bug that results in an assert + argument exception when the SecureString shrinks and then re-grows.
2. SecureString's handling of OOM conditions is inconsistent.
3. Lots of SafeBuffers are being left to finalization to clean up. 
4. The assembly has an unused resource string.

### DIFF
--- a/src/System.Security.SecureString/src/System/Security/SafeBSTRHandle.cs
+++ b/src/System.Security.SecureString/src/System/Security/SafeBSTRHandle.cs
@@ -11,16 +11,24 @@ namespace System.Security
     {
         internal SafeBSTRHandle() : base(true) { }
 
-        internal static SafeBSTRHandle Allocate(string src, uint len)
+        internal static SafeBSTRHandle Allocate(string src, uint lenInChars)
         {
-            SafeBSTRHandle bstr = Interop.OleAut32.SysAllocStringLen(src, len);
-            bstr.Initialize(len * sizeof(char));
+            SafeBSTRHandle bstr = Interop.OleAut32.SysAllocStringLen(src, lenInChars);
+            if (bstr.IsInvalid) // SysAllocStringLen returns a NULL ptr when there's insufficient memory
+            {
+                throw new OutOfMemoryException();
+            }
+            bstr.Initialize(lenInChars * sizeof(char));
             return bstr;
         }
 
         internal static SafeBSTRHandle Allocate(IntPtr src, uint lenInBytes)
         {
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @pallavit

src/System.Security.SecureString/src/System/Security/SafeBSTRHandle.cs
```diff
@@ -11,16 +11,24 @@ internal sealed class SafeBSTRHandle : SafeBuffer
...
+            {
+                throw new OutOfMemoryException();
+            }
+            bstr.Initialize(lenInChars * sizeof(char));
             return bstr;
         }
 
         internal static SafeBSTRHandle Allocate(IntPtr src, uint lenInBytes)
         {
             SafeBSTRHandle bstr = Interop.OleAut32.SysAllocStringLen(src, lenInBytes / sizeof(char));
```
morganbr: Not your change, but it looks like this is rounding down, so we might under-allocate

src/System.Security.SecureString/src/System/Security/SecureString.Windows.cs
```diff
@@ -103,9 +103,15 @@ private unsafe void InsertAtCore(int index, char c)
             }
             finally
             {
-                ProtectMemory(decryptedBuffer);
-                if (bufferPtr != null)
-                    decryptedBuffer.ReleasePointer();
+                try
+                {
+                    if (bufferPtr != null)
+                        decryptedBuffer.ReleasePointer();
```
morganbr: When do you expect ReleasePointer to fail?

stephentoub: I was probably just being overly protective.  It could fail if decryptedBuffer was already disposed, but then the ProtectMemory call would also fail.  It could fail if decryptedBuffer had elsewhere had ReleasePointer erroneously called to decrement the value, but that would indicate a bug elsewhere in the SecureString implementation.  I'll remove the extra try/finally.

src/System.Security.SecureString/src/System/Security/SafeBSTRHandle.cs
stephentoub: I'll add an assert to verify that it always represents a number of characters, i.e. that lenInBytes % sizeof(char) == 0.

PR #1752, opened 2015-05-14
Update build tools version.

JeremyKuhne:
Takes fix to allow use of newer DNU package.

@weshaggard, @ericstj, @akoeplinger


ericstj:
LGTM if it builds :+1:

PR #1756, opened 2015-05-15
Implement generic interfaces on Regex collections

### LINKED ISSUE #271: Regex collections should implement generic collection interfaces
`CaptureCollection`, `GroupCollection`, and `MatchCollection` currently only implement the non-generic `ICollection` interface. These collections should implement the generic collection interfaces to better interoperate with more modern APIs, such as LINQ. Since these collections are already indexable, they should implement  `IList<T>` and `IReadOnlyList<T>`, as well as the non-generic `IList` (to be consistent with the generic interfaces).

## Rationale and Usage

This is certainly a nice-to-have, but it *is* a long-standing [request](https://connect.microsoft.com/VisualStudio/feedback/details/469250/capturecollection-groupcollection-and-matchcollection-do-not-implement-icollection-t-and-ienumerable-t) that [developers](http://stackoverflow.com/q/7274422) [still](http://jimlynn.wordpress.com/2009/11/04/why-cant-i-use-linq-on-a-capturecollection/) [ask](http://stackoverflow.com/q/16379433) [about](http://stackoverflow.com/q/4108956). Implementing the generic interfaces will allow these collections to be used more easily with LINQ and interoperate better with more modern framework and library APIs.

For example, to use these collections with LINQ right now you have to know about and remember to use `Enumerable.Cast<TSource>()` to cast the non-generic `IEnumerable` into an `IEnumerable<T>`:

```c#
var captures =
    from capture in match.Groups.Cast<Group>().Last().Captures.Cast<Capture>()
    select capture.Value;
```

With these changes you'd no longer have to do that:

```c#
var captures =
    from capture in match.Groups.Last().Captures
    select capture.Value;
```

Plus, in the above example, you'd get a performance improvement when using `Enumerable.Last<TSource>.()` as its implementation has a fast-path for collections that implement `IList<T>`.

## Proposed API

```c#
// DebuggerDisplay and DebuggerTypeProxy added
[DebuggerDisplay("Count = {Count}")]
[DebuggerTypeProxy(typeof(RegexCollectionDebuggerProxy<Capture>))]
// Previously only implemented ICollection
public class CaptureCollection : IList<Capture>, IReadOnlyList<Capture>, IList
{
    // Existing members
    public int Count { get; }
    public Capture this[int i] { get; }
    public IEnumerator GetEnumerator();
    object ICollection.SyncRoot { get; }
    bool ICollection.IsSynchronized { get; }
    void ICollection.CopyTo(Array array, int arrayIndex);

    // Proposed members
    public void CopyTo(Capture[] array, int arrayIndex);
    IEnumerator<Capture> IEnumerable<Capture>.GetEnumerator();
    int IList<Capture>.IndexOf(Capture item);
    void IList<Capture>.Insert(int index, Capture item);
    void IList<Capture>.RemoveAt(int index);
    Capture IList<Capture>.this[int index] { get; set; }
    void ICollection<Capture>.Add(Capture item);
    void ICollection<Capture>.Clear();
    bool ICollection<Capture>.Contains(Capture item);
    bool ICollection<Capture>.IsReadOnly { get; }
    bool ICollection<Capture>.Remove(Capture item);
    int IList.Add(object value);
    void IList.Clear();
    bool IList.Contains(object value);
    int IList.IndexOf(object value);
    IList.Insert(int index, object value);
    bool IList.IsFixedSize { get; }
    bool IList.IsReadOnly { get; }
    void IList.Remove(object value);
    void IList.RemoveAt(int index);
    object IList.this[int index] { get; set; }
}

// DebuggerDisplay and DebuggerTypeProxy added
[DebuggerDisplay("Count = {Count}")]
[DebuggerTypeProxy(typeof(RegexCollectionDebuggerProxy<Group>))]
// Previously only implemented ICollection
public class GroupCollection : IList<Group>, IReadOnlyList<Group>, IList
{
    // Existing members
    public int Count { get; }
    public Group this[int groupnum] { get; }
    public Group this[String groupname] { get; }
    public IEnumerator GetEnumerator();
    object ICollection.SyncRoot { get; }
    bool ICollection.IsSynchronized { get; }
    void ICollection.CopyTo(Array array, int arrayIndex);

    // Proposed members
    public void CopyTo(Group[] array, int arrayIndex);
    IEnumerator<Group> IEnumerable<Group>.GetEnumerator();
    int IList<Group>.IndexOf(Group item);
    void IList<Group>.Insert(int index, Group item);
    void IList<Group>.RemoveAt(int index);
    Group IList<Group>.this[int index] { get; set; }
    void ICollection<Group>.Add(Group item);
    void ICollection<Group>.Clear();
    bool ICollection<Group>.Contains(Group item);
    bool ICollection<Group>.IsReadOnly { get; }
    bool ICollection<Group>.Remove(Group item);
    int IList.Add(object value);
    void IList.Clear();
    bool IList.Contains(object value);
    int IList.IndexOf(object value);
    IList.Insert(int index, object value);
    bool IList.IsFixedSize { get; }
    bool IList.IsReadOnly { get; }
    void IList.Remove(object value);
    void IList.RemoveAt(int index);
    object IList.this[int index] { get; set; }
}

// DebuggerDisplay and DebuggerTypeProxy added
[DebuggerDisplay("Count = {Count}")]
[DebuggerTypeProxy(typeof(RegexCollectionDebuggerProxy<Match>))]
// Previously only implemented ICollection
public class MatchCollection : IList<Match>, IReadOnlyList<Match>, IList
{
    // Existing members
    public int Count { get; }
    public virtual Match this[int i] { get; }
    public IEnumerator GetEnumerator();
    object ICollection.SyncRoot { get; }
    bool ICollection.IsSynchronized { get; }
    void ICollection.CopyTo(Array array, int arrayIndex);

    // Proposed members
    public void CopyTo(Match[] array, int arrayIndex);
    IEnumerator<Match> IEnumerable<Match>.GetEnumerator();
    int IList<Match>.IndexOf(Match item);
    void IList<Match>.Insert(int index, Match item);
    void IList<Match>.RemoveAt(int index);
    Match IList<Match>.this[int index] { get; set; }
    void ICollection<Match>.Add(Match item);
    void ICollection<Match>.Clear();
    bool ICollection<Match>.Contains(Match item);
    bool ICollection<Match>.IsReadOnly { get; }
    bool ICollection<Match>.Remove(Match item);
    int IList.Add(object value);
    void IList.Clear();
    bool IList.Contains(object value);
    int IList.IndexOf(object value);
    IList.Insert(int index, object value);
    bool IList.IsFixedSize { get; }
    bool IList.IsReadOnly { get; }
    void IList.Remove(object value);
    void IList.RemoveAt(int index);
    object IList.this[int index] { get; set; }
}

```

## Details

  * There was some discussion as to whether only the read-only interfaces should be implemented, or both the read-only and mutable interfaces. The consensus is to implement both the read-only and mutable interfaces. This is consistent with other collections in the framework. The mutable interfaces are implemented as read-only: mutable members are implemented explicitly and throw `NotSupportedException` (like `ReadOnlyCollection<T>`).

  * There was an open question as to whether the non-generic `IList` should be implemented as well. These collections are indexable and if  `IList<T>` and `IReadOnlyList<T>` are being implemented, `IList` should be implemented as well. This does add several more members, but they are all implemented explicitly so they don't add any new public members to intellisense, and the implementations are very straightforward.

  * `ICollection<T>.CopyTo` is implemented implicitly (public).

  * All other new members are implemented explicitly (non-public):

    * Mutable members are implemented explicitly because these collections are read-only and the mutable members throw `NotSupportedException` (like `ReadOnlyCollection<T>`).

    * `IList` members are implemented explicitly to hide non-generic members from intellisense.

    * `IList<T>.IndexOf` and `ICollection<T>.Contains` are implemented explicitly because these methods aren't very useful for these collections and should not be visible in intellisense by default. They're not useful because an implementation using `EqualityComparer<T>.Default` (consistent with other collections) will search the collection using reference equality due to the fact that `Capture`, `Group`, and `Match` do not implement `IEquatable<T>` and do not override `Equals()` and `GetHashCode()`. Further, these types do not have public constructors -- they are created internally by the regex engine, making it very unlikely that you'd want to search for an item in a collection "A" that was obtained from collection "B".

    * `IEnumerable<T>.GetEnumerator()` must be implemented explicitly because the non-generic `IEnumerable.GetEnumerator()` is already implemented implicitly and we can't overload on return type. This also precludes returning a `struct` `Enumerator` (for better enumeration performance) because changing the return type of the existing method would be a binary breaking change. As a result, you'll still have to specify the type when using `foreach` (e.g. `foreach (Capture capture in captures)`); you won't be able to use `var` (e.g. `foreach (var capture in captures)`), unfortunately.

## Open Questions

 * Should `GroupCollection` implement `IDictionary<string, Group>`, `IReadOnlyDictionary<string, Group>`, and `IDictionary`? `GroupCollection` already has a string indexer. Is it worth implementing the dictionary interfaces as part of this? Personally, I'm leaning toward "no" because there isn't a compelling scenario for the dictionary interfaces, and they can always be added in the future when needed.

## Pull Request

A PR with the proposed changes is available: #1756

## Updates

 * Edited this description to make it more of a *speclet*, based on the discussion below and the proposed API Review process.
 * Some improvements based on feedback from @sharwell.
 * Fixed existing members.
 * Added `IList`. These collections are indexable and it would be strange if `IList<T>` and `IReadOnlyList<T>` were implemented alongside `ICollection` but without `IList`.
 * Added `DebuggerDisplay` and `DebuggerTypeProxy` attributes.
 * Made `ICollection<T>.CopyTo` implicit (public).

justinvp:
Implement `IList<T>`, `IReadOnlyList<T>`, and `IList` on the Regex collections.

Fixes #271

### DIFF
--- a/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
+++ b/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
@@ -5,6 +5,7 @@
 // contained in a compiled Regex.
 
 using System.Collections;
+using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace System.Text.RegularExpressions
@@ -18,7 +19,9 @@ namespace System.Text.RegularExpressions
     /// Represents a sequence of capture substrings. The object is used
     /// to return the set of captures done by a single capturing group.
     /// </summary>
-    public class CaptureCollection : ICollection
+    [DebuggerDisplay("Count = {Count}")]
+    [DebuggerTypeProxy(typeof(RegexCollectionDebuggerProxy<Capture>))]
+    public class CaptureCollection : IList<Capture>, IReadOnlyList<Capture>, IList
     {
         private readonly Group _group;
         private readonly int _capcount;
@@ -47,9 +50,33 @@ namespace System.Text.RegularExpressions
         }
 
         /// <summary>
+        /// Copies all the elements of the collection to the given array
+        /// beginning at the given index.
+        /// </summary>
+        public void CopyTo(Capture[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
+            if (arrayIndex < 0 || arrayIndex > array.Length)
+                throw new ArgumentOutOfRangeException("arrayIndex");
+            if (array.Length - arrayIndex < Count)
+                throw new ArgumentException(SR.Arg_ArrayPlusOffTooSmall);
+
+            for (int i = arrayIndex, j = 0; j < Count; i++, j++)
+            {
+                array[i] = this[j];
+            }
+        }
+
+        /// <summary>
         /// Provides an enumerator in the same order as Item[].
         /// </summary>
         public IEnumerator GetEnumerator()
+        {
+            return new Enumerator(this);
+        }
+
+        IEnumerator<Capture> IEnumerable<Capture>.GetEnumerator()
         {
             return new Enumerator(this);
         }
@@ -78,6 +105,109 @@ namespace System.Text.RegularExpressions
             return _captures[i];
         }
 
+        int IList<Capture>.IndexOf(Capture item)
+        {
+            var comparer = EqualityComparer<Capture>.Default;
+            for (int i = 0; i < Count; i++)
+            {
+                if (comparer.Equals(this[i], item))
+                    return i;
+            }
+            return -1;
+        }
+
+        void IList<Capture>.Insert(int index, Capture item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList<Capture>.RemoveAt(int index)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        Capture IList<Capture>.this[int index]
+        {
+            get { return this[index]; }
+            set { throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection); }
+        }
+
+        void ICollection<Capture>.Add(Capture item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void ICollection<Capture>.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool ICollection<Capture>.Contains(Capture item)
+        {
+            return ((IList<Capture>)this).IndexOf(item) >= 0;
+        }
+
+        bool ICollection<Capture>.IsReadOnly
+        {
+            get { return true; }
+        }
+
+        bool ICollection<Capture>.Remove(Capture item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        int IList.Add(object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool IList.Contains(object value)
+        {
+            return value is Capture && ((ICollection<Capture>)this).Contains((Capture)value);
+        }
+
+        int IList.IndexOf(object value)
+        {
+            return value is Capture ? ((IList<Capture>)this).IndexOf((Capture)value) : -1;
+        }
+
+        void IList.Insert(int index, object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool IList.IsFixedSize
+        {
+            get { return true; }
+        }
+
+        bool IList.IsReadOnly
+        {
+            get { return true; }
+        }
+
+        void IList.Remove(object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList.RemoveAt(int index)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        object IList.this[int index]
+        {
+            get { return this[index]; }
+            set { throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection); }
+        }
+
         bool ICollection.IsSynchronized
         {
             get { return false; }
@@ -99,7 +229,7 @@ namespace System.Text.RegularExpressions
             }
         }
 
-        private class Enumerator : IEnumerator
+        private class Enumerator : IEnumerator<Capture>
         {
             private readonly CaptureCollection _collection;
             private int _index;
@@ -143,6 +273,10 @@ namespace System.Text.RegularExpressions
             void IEnumerator.Reset()
             {
                 _index = -1;
+            }
+
+            void IDisposable.Dispose()
+            {
             }
         }
     }
--- a/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCollectionDebuggerProxy.cs
+++ b/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCollectionDebuggerProxy.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Text.RegularExpressions
+{
+    internal sealed class RegexCollectionDebuggerProxy<T>
+    {
+        private readonly ICollection<T> _collection;
+
+        public RegexCollectionDebuggerProxy(ICollection<T> collection)
+        {
+            if (collection == null)
+                throw new ArgumentNullException("collection");
+
+            _collection = collection;
+        }
+
+        [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
+        public T[] Items
+        {
+            get
+            {
+                T[] items = new T[_collection.Count];
+                _collection.CopyTo(items, 0);
+                return items;
+            }
+        }
+    }
+}
--- a/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexGroupCollection.cs
+++ b/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexGroupCollection.cs
@@ -14,7 +14,9 @@ namespace System.Text.RegularExpressions
     /// Represents a sequence of capture substrings. The object is used
     /// to return the set of captures done by a single capturing group.
     /// </summary>
-    public class GroupCollection : ICollection
+    [DebuggerDisplay("Count = {Count}")]
+    [DebuggerTypeProxy(typeof(RegexCollectionDebuggerProxy<Group>))]
+    public class GroupCollection : IList<Group>, IReadOnlyList<Group>, IList
     {
         private readonly Match _match;
         private readonly Dictionary<int, int> _captureMap;
@@ -53,9 +55,33 @@ namespace System.Text.RegularExpressions
         }
 
         /// <summary>
+        /// Copies all the elements of the collection to the given array
+        /// beginning at the given index.
+        /// </summary>
+        public void CopyTo(Group[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
+            if (arrayIndex < 0 || arrayIndex > array.Length)
+                throw new ArgumentOutOfRangeException("arrayIndex");
+            if (array.Length - arrayIndex < Count)
+                throw new ArgumentException(SR.Arg_ArrayPlusOffTooSmall);
+
+            for (int i = arrayIndex, j = 0; j < Count; i++, j++)
+            {
+                array[i] = this[j];
+            }
+        }
+
+        /// <summary>
         /// Provides an enumerator in the same order as Item[].
         /// </summary>
         public IEnumerator GetEnumerator()
+        {
+            return new Enumerator(this);
+        }
+
+        IEnumerator<Group> IEnumerable<Group>.GetEnumerator()
         {
             return new Enumerator(this);
         }
@@ -101,6 +127,109 @@ namespace System.Text.RegularExpressions
             return _groups[groupnum - 1];
         }
 
+        int IList<Group>.IndexOf(Group item)
+        {
+            var comparer = EqualityComparer<Group>.Default;
+            for (int i = 0; i < Count; i++)
+            {
+                if (comparer.Equals(this[i], item))
+                    return i;
+            }
+            return -1;
+        }
+
+        void IList<Group>.Insert(int index, Group item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList<Group>.RemoveAt(int index)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        Group IList<Group>.this[int index]
+        {
+            get { return this[index]; }
+            set { throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection); }
+        }
+
+        void ICollection<Group>.Add(Group item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void ICollection<Group>.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool ICollection<Group>.Contains(Group item)
+        {
+            return ((IList<Group>)this).IndexOf(item) >= 0;
+        }
+
+        bool ICollection<Group>.IsReadOnly
+        {
+            get { return true; }
+        }
+
+        bool ICollection<Group>.Remove(Group item)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        int IList.Add(object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool IList.Contains(object value)
+        {
+            return value is Group && ((ICollection<Group>)this).Contains((Group)value);
+        }
+
+        int IList.IndexOf(object value)
+        {
+            return value is Group ? ((IList<Group>)this).IndexOf((Group)value) : -1;
+        }
+
+        void IList.Insert(int index, object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool IList.IsFixedSize
+        {
+            get { return true; }
+        }
+
+        bool IList.IsReadOnly
+        {
+            get { return true; }
+        }
+
+        void IList.Remove(object value)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        void IList.RemoveAt(int index)
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        object IList.this[int index]
+        {
+            get { return this[index]; }
+            set { throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection); }
+        }
+
         bool ICollection.IsSynchronized
         {
             get { return false; }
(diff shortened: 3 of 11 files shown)


src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
```diff
@@ -47,6 +50,25 @@ public int Count
...
 
         /// <summary>
+        /// Copies all the elements of the collection to the given array
+        /// beginning at the given index.
+        /// </summary>
+        public void CopyTo(Capture[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
+            if (arrayIndex < 0 || arrayIndex > array.Length)
```
KrzysztofCwalina: shouldn't it be arrayIndex >= array.Length?

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexGroupCollection.cs
```diff
@@ -53,6 +55,25 @@ public int Count
...
 
         /// <summary>
+        /// Copies all the elements of the collection to the given array
+        /// beginning at the given index.
+        /// </summary>
+        public void CopyTo(Group[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("array");
+            if (arrayIndex < 0 || arrayIndex > array.Length)
```
KrzysztofCwalina: Same.

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
justinvp: I originally had it like that, but then the behavior differs from `List<T>` and `Dictionary<TKey, TValue>`, which throws `ArgumentException` in that case instead of `ArgumentOutOfRangeException`.

See https://github.com/dotnet/coreclr/blob/ef1e2ab328087c61a6878c1e84f4fc5d710aebce/src/mscorlib/src/System/Collections/Generic/Dictionary.cs#L818

justinvp: There's actually even a behavioral difference between `List<T>.CopyTo` and `Dictionary<TKey, TValue>.CopyTo`.

The following tests pass:

```c#
ICollection<int> list = new List<int> { 0, 1 };
int[] listArray = new int[2];
Assert.Throws<ArgumentOutOfRangeException>(() => list.CopyTo(listArray, -1));
Assert.Throws<ArgumentException>(() => list.CopyTo(listArray, 2));
Assert.Throws<ArgumentException>(() => list.CopyTo(listArray, 3));

ICollection<KeyValuePair<int, int>> dictionary = new Dictionary<int, int> { { 0, 0 }, { 1, 1 } };
KeyValuePair<int, int>[] dictionaryArray = new KeyValuePair<int, int>[2];
Assert.Throws<ArgumentOutOfRangeException>(() => dictionary.CopyTo(dictionaryArray, -1));
Assert.Throws<ArgumentException>(() => dictionary.CopyTo(dictionaryArray, 2));
Assert.Throws<ArgumentOutOfRangeException>(() => dictionary.CopyTo(dictionaryArray, 3));
```

Notice the last case is `ArgumentException` for `List<T>` but `ArgumentOutOfRangeException` for `Dictionary<TKey, TValue>`

Right now `CaptureCollection` and `GroupCollection` are consistent with `Dictionary<TKey, TValue>` whereas `MatchCollection` is consistent with `List<T>` (because `MatchCollection` uses `List<T>.CopyTo` internally).

Maybe it'd be better if I remove the `arrayIndex > array.Length` check altogether from `CaptureCollection` and `GroupCollection`, so the exceptions are consistent with `MatchCollection` and `List<T>`. Or not worry about it too much because it's all `ArgumentException` anyway?

KrzysztofCwalina: Since we are already not very consistent, I would do either what you suggested (to simplify the code) or what I suggested (follow current convention and implemented correctly). I would not try to be consistent with incorrectly implemented range checks ;-)

Moreover, it would be great to have a helper for common range checks, so that people don't make similar mistakes (to those made in the existing collections). Something that can be called along the lines of:
Precondition.RequireFit(source : this, destination : array, destinationIndex : arrayIndex);

KrzysztofCwalina:
@dotnet-bot test this please

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
justinvp: After investigating how other collections implement `CopyTo`, I think we should leave the range checks in the PR as-is as this is how other collections implement `CopyTo`.

For example, all of the following cases **do not** throw:

```c#
new int[0].CopyTo(new int[10], 10);
new List<int>().CopyTo(new int[10], 10);
new LinkedList<int>().CopyTo(new int[10], 10);
new Queue<int>().CopyTo(new int[10], 10);
new Stack<int>().CopyTo(new int[10], 10);
new HashSet<int>().CopyTo(new int[10], 10);
new SortedSet<int>().CopyTo(new int[10], 10);
((ICollection<KeyValuePair<int, int>>)new Dictionary<int, int>()).CopyTo(new KeyValuePair<int,int>[10], 10);
((ICollection<KeyValuePair<int, int>>)new SortedDictionary<int, int>()).CopyTo(new KeyValuePair<int, int>[10], 10);
new ConcurrentStack<int>().CopyTo(new int[10], 10);
new ConcurrentQueue<int>().CopyTo(new int[10], 10);
new ConcurrentBag<int>().CopyTo(new int[10], 10);
((ICollection<KeyValuePair<int, int>>)new ConcurrentDictionary<int, int>()).CopyTo(new KeyValuePair<int, int>[10], 10);
ImmutableArray.Create<int>().CopyTo(new int[10], 10);
ImmutableList.Create<int>().CopyTo(new int[10], 10);
((ICollection<int>)ImmutableHashSet.Create<int>()).CopyTo(new int[10], 10);
((ICollection<int>)ImmutableSortedSet.Create<int>()).CopyTo(new int[10], 10);
((ICollection<KeyValuePair<int, int>>)ImmutableDictionary.Create<int, int>()).CopyTo(new KeyValuePair<int, int>[10], 10);
((ICollection<KeyValuePair<int, int>>)ImmutableSortedDictionary.Create<int, int>()).CopyTo(new KeyValuePair<int, int>[10], 10);
```

If I change the range check from `arrayIndex > array.Length` to `arrayIndex >= array.Length`, then the regex collections would throw in such cases and be inconsistent with all of these collections.

Maxwe11:
@dotnet-bot test this please

Maxwe11:
@dotnet-bot test this please

stephentoub:
@weshaggard / @mmitche, it seems like all PRs against the future branch are failing in CI, I believe for issues that were already fixed in master, and I'm wondering if merging master to future would help...?

stephentoub:
@dotnet-bot test this please

stephentoub:
@justinvp, master was just merged to future, in part because there were some fixes in master from a while ago that I believe were causing your PR's CIs to fail.  There are now merge conflicts with your change.  Can you fix it up appropriately and recommit/push?

justinvp:
Fixed. The conflict was due to the move from `packages.config` to `project.json`.

After running a build, the `project.lock.json` file becomes modified. Should these modifications to `project.lock.json` be checked-in and included in the PR?

After the changes to `project.json` and running `msbuild \t:BuildAndTest` from the test directory, it builds and the tests pass, but I get a build warning:

> "C:\dev\corefx\src\System.Text.RegularExpressions\tests\System.Text.RegularExpressions.Tests.csproj" (BuildAndTest target) (1) ->(ResolveAssemblyReferences target) ->  C:\Program Files (x86)\MSBuild\12.0\bin\Microsoft.Common.CurrentVersion.targets(1697,5): warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [C:\dev\corefx\src\System.Text.RegularExpressions\tests\System.Text.RegularExpressions.Tests.csproj]

I'm not sure where in the build log to look for the reference conflicts it is warning about, or why it's warning about them.

I see the same warning for System.Security.Principal.Windows.Tests.csproj when running `msbuild \t:BuildAndTest` for it out of the future branch (without any of the changes from this PR):

> "C:\dev\corefx\src\System.Security.Principal.Windows\tests\System.Security.Principal.Windows.Tests.csproj" (BuildAndTest target) (1) ->(ResolveAssemblyReferences target) ->  C:\Program Files (x86)\MSBuild\12.0\bin\Microsoft.Common.CurrentVersion.targets(1697,5): warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [C:\dev\corefx\src\System.Security.Principal.Windows\tests\System.Security.Principal.Windows.Tests.csproj]

stephentoub:
@ericstj, the answers to @justinvp's questions are "the changes to project.lock.json should be checked in" and "the warning is expected", correct?

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
```diff
@@ -78,6 +105,115 @@ private Capture GetCapture(int i)
...
+        }
+
+        void ICollection<Capture>.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool ICollection<Capture>.Contains(Capture item)
+        {
+            var comparer = EqualityComparer<Capture>.Default;
```
stephentoub: Rather than duplicate logic, did you consider just having the implementation be:
```C#
return ((IList<Capture>)this).IndexOf(item) >= 0;
```
?

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCollectionDebuggerProxy.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Text.RegularExpressions
+{
+    internal sealed class RegexCollectionDebuggerProxy<T>
```
stephentoub: @KrzysztofCwalina / @terrajobst , lots of assemblies end up having a debugger proxy like this.  Have we considered adding public ```CollectionDebuggerProxy<T>``` for any ```ICollection<T>``` implementation to System.Collections.dll?  Then any assembly implementing an ```ICollection<T>``` could just use that one public implementation rather than crafting their own and effectively duplicating the same implementation over and over.

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexGroupCollection.cs
```diff
@@ -101,6 +127,115 @@ private Group GetGroupImpl(int groupnum)
...
+        }
+
+        void ICollection<Group>.Clear()
+        {
+            throw new NotSupportedException(SR.NotSupported_ReadOnlyCollection);
+        }
+
+        bool ICollection<Group>.Contains(Group item)
+        {
+            var comparer = EqualityComparer<Group>.Default;
```
stephentoub: Same question about implementing this in terms of IndexOf.  (I'll refrain from commenting on this kind of thing if it crops up again.)

src/System.Text.RegularExpressions/tests/CaptureCollectionTests.cs
```diff
@@ -1,73 +1,271 @@
...
+            Assert.Throws<NotSupportedException>(() => list.Remove(default(Capture)));
+            Assert.Throws<NotSupportedException>(() => list.RemoveAt(0));
+            Assert.Throws<NotSupportedException>(() => list[0] = default(Capture));
+        }
+
+        [Fact]
+        public static void DebuggerAttributeTests()
+        {
+            DebuggerAttributes.ValidateDebuggerDisplayReferences(CreateCollection());
+            DebuggerAttributes.ValidateDebuggerTypeProxyProperties(CreateCollection());
```
stephentoub: Nice :)

stephentoub:
A few minor comments, but otherwise LGTM.  Thanks for doing this, Justin.

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCaptureCollection.cs
justinvp: Fixed throughout.

justinvp:
I updated the PR to include `project.lock.json` and changed `Contains` to be implemented in terms of `IndexOf`.

Thanks for the review, @KrzysztofCwalina and @stephentoub!

KrzysztofCwalina:
Yeah. Thanks for the contribution. I will start new rounds of tests and merge when they pass.

KrzysztofCwalina:
@dotnet-bot test this please

justinvp:
Unrelated test failure on linux:

> 09:18:55    System.IO.Compression.Test.zip_InvalidParametersAndStrangeFiles.StrangeFiles4 [FAIL]
09:18:55       Assert.Equal() Failure
09:18:55                 ��� (pos 0)
09:18:55       Expected: second.txt
09:18:55       Actual:   notempty\second.txt
09:18:55                 ��� (pos 0)
09:18:55       Stack Trace:
09:18:55             at System.IO.Compression.Test.ZipTest.IsZipSameAsDir(Stream archiveFile, String directory, ZipArchiveMode mode, Boolean dontRequireExplicit, Boolean dontCheckTimes)
09:18:55             at System.IO.Compression.Test.zip_InvalidParametersAndStrangeFiles.<StrangeFiles4>d__58.MoveNext()
09:18:55          --- End of stack trace from previous location where exception was thrown ---
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
09:18:55          --- End of stack trace from previous location where exception was thrown ---
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
09:18:55          --- End of stack trace from previous location where exception was thrown ---
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
09:18:55             at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
09:19:04 Finished:    System.IO.Compression.Tests
09:19:04 
09:19:04 === TEST EXECUTION SUMMARY ===
09:19:04    System.IO.Compression.Tests  Total: 75, Errors: 0, Failed: 1, Skipped: 0, Time: 10.114s

Looks like https://github.com/dotnet/corefx/issues/1904

https://github.com/dotnet/corefx/pull/1905 needs to be merged into future.

stephentoub:
> #1905 needs to be merged into future.

Yup, that should fix it.

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCollectionDebuggerProxy.cs
weshaggard: This is a good question for some folks on the debugger/EE side. @tmat @cston Is there another option to provide these debugger proxies in a more general fashion?

KrzysztofCwalina:
@dotnet-bot test this please

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCollectionDebuggerProxy.cs
tmat: I think it would be reasonable for the EE to handle ICollection<T> automatically without the need for proxies, if it doesn't do so already. However, if the logic is not present in older EEs removing the proxy would degrade the experience (might be ok if it would only affects Dev10 and prior).

I think it is possibly to specify a proxy as an assembly-level attribute, which specifies the target type. Not sure if it is possible to target an interface though, probably not. BTW, I'm not positive but I think that the current debugger proxy architecture requires the target type to be serializable.

stephentoub:
@justinvp, the future branch is now up-to-date with master.

justinvp:
@stephentoub, thanks! Does @dotnet-bot accept commands from me?...

justinvp:
@dotnet-bot test this please

stephentoub:
Apparently :smile:

stephentoub:
@dotnet-bot test this please

stephentoub:
Thanks for the contribution, Justin!

PR #1757, opened 2015-05-15
Add tests for CryptoStream et al

stephentoub:
This adds more tests for System.Security.Cryptography.Encryption and System.Security.Cryptography.Hashing.Algorithms, with the primary addition being for CryptoStream that wasn't previously exercised at all.

Code coverage:
- System.Security.Cryptography.Encryption: 31.7% => 91.7%
- System.Security.Cryptography.Hashing.Algorithms: 66.1% => 73.4%

There are two changes to the product code as part of this:
- Stored the int hr value passed to CryptographicException into its HResult property.
- Removed some unused resource strings.

### DIFF
--- a/src/System.Security.Cryptography.Encryption/tests/CryptoStream.cs
+++ b/src/System.Security.Cryptography.Encryption/tests/CryptoStream.cs
@@ -1,0 +1,226 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using System.Text;
+using Xunit;
+
+namespace System.Security.Cryptography.Encryption.Tests.Asymmetric
+{
+    public static class CryptoStreamTests
+    {
+        [Fact]
+        public static void Ctor()
+        {
+            var transform = new IdentityTransform(1, 1, true);
+            Assert.Throws<ArgumentException>(() => new CryptoStream(new MemoryStream(), transform, (CryptoStreamMode)12345));
+            Assert.Throws<ArgumentException>(() => new CryptoStream(new MemoryStream(new byte[0], writable: false), transform, CryptoStreamMode.Write));
+            Assert.Throws<ArgumentException>(() => new CryptoStream(new CryptoStream(new MemoryStream(new byte[0]), transform, CryptoStreamMode.Write), transform, CryptoStreamMode.Read));
+        }
+
+        [Theory]
+        [InlineData(64, 64, true)]
+        [InlineData(64, 128, true)]
+        [InlineData(128, 64, true)]
+        [InlineData(1, 1, true)]
+        [InlineData(37, 24, true)]
+        [InlineData(128, 3, true)]
+        [InlineData(8192, 64, true)]
+        [InlineData(64, 64, false)]
+        public static void Roundtrip(int inputBlockSize, int outputBlockSize, bool canTransformMultipleBlocks)
+        {
+            ICryptoTransform encryptor = new IdentityTransform(inputBlockSize, outputBlockSize, canTransformMultipleBlocks);
+            ICryptoTransform decryptor = new IdentityTransform(inputBlockSize, outputBlockSize, canTransformMultipleBlocks);
+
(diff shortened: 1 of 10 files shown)


stephentoub:
cc: @bartonjs, @nguerrera

stephentoub:
FYI, the new HMAC asserts are passing on Windows but failing on Linux. I'll investigate.

src/System.Security.Cryptography.Encryption/tests/CryptoStream.cs
```diff
@@ -0,0 +1,216 @@
...
+
+                using (StreamReader reader = new StreamReader(decryptStream))
+                {
+                    Assert.Equal(
+                        LoremText + LoremText + LoremText + LoremText,
+                        reader.ReadToEnd());
+                }
+            }
+
+            stream = new MemoryStream(stream.ToArray()); // CryptoStream.Dispose disposes the stream
```
bartonjs: Can you add a comment to pre-suggest to the reader that we're doing the read again so that it can test async?  I was fixated more on figuring out if the previous one had tried doing anything destructive through the assert sequence.

bartonjs:
@stephentoub It looks like the Unix provider doesn't copy the key back out of OpenSSL after initializing.  Cases 6 and 7 use a too-large key so that you have to do key=HASH(key).

The assert should be that hmac.Key.Length <= hmac.HashSize (bits to bytes as necessary).  ([RFC2104](https://tools.ietf.org/html/rfc2104) section 2: Applications that use keys longer than B bytes will first hash the key using H and then use the resultant L byte string as the actual key to HMAC.).  The less-than technically is addressed in the algorithm sequence, not the key selection, so asserting that they're equal might be overly strict.

src/System.Security.Cryptography.Hashing.Algorithms/tests/HmacTests.cs
```diff
@@ -31,7 +31,14 @@ protected HmacTests(byte[][] testKeys, byte[][] testData)
 
             using (HMAC hmac = Create())
             {
-                hmac.Key = _testKeys[testCaseId];
+                byte[] key = _testKeys[testCaseId];
+                hmac.Key = key;
+                Assert.NotSame(key, hmac.Key);
+                Assert.NotSame(hmac.Key, hmac.Key);
```
bartonjs: This makes me wonder if we should also have a test that assigns Key to an array, changes the local array, and asserts that they are no longer Equal.  (The NotSame test here shows that each get_Key returns a copy, and my musing is showing that set_Key takes a copy, instead of maintaining the reference)

src/System.Security.Cryptography.Encryption/tests/CryptoStream.cs
stephentoub: Sure, will do.

src/System.Security.Cryptography.Hashing.Algorithms/tests/HmacTests.cs
stephentoub: Isn't that what the first NotSame assert above effectively does?

bartonjs: I'm suggesting guarding against
```
public byte[] Key
{
    get { return _key.CloneByteArray(); }
    set { _key = value; }
}
```

key and hmac.get_Key() will not ReferenceEquals, but hmac is still holding a reference to the original array, so it's subject to mutation errors.

It'd be a very strange implementation, but a lot of tests are about making sure we didn't implement things strangely :).

stephentoub: Ah, sure, ok, will do.

stephentoub:
@bartonjs, as you and I discussed, there's a bug in both the Windows and Linux implementation when compared to the desktop implementation.  I've removed the assert for now, and we can add back a similar one when we separately fix that bug (which is that the getter for the Key is not returning the actual key when a new key needed to be computed).

stephentoub:
@dotnet-bot test this please

PR #972, opened 2015-02-20
Fix for Issue #587 - Untangle FileStream's async implementation

dpaoliello:
Fix for Issue #587 - Untangle FileStream's async implementation

This removes the need for the old Begin\End methods in Win32FileStream by modifying the wrapper around the Overlapped (Win32FileStream.FileStreamAsyncResult) to use a TaskCompletionSource, thus allowing the core Read\Write methods to return Tasks directly.

It also fixes a potential race condition where Begin* may be registering the cancellation at the same time that End* is trying to dispose it (if the user called the End* method from the callback that they provided to the Begin* method).

PR #1758, opened 2015-05-15
Fix canceled async I/O tasks to be Canceled instead of Faulted

stephentoub:
When FileStream async I/O completes due to cancellation, Windows gives back an error code of ERROR_OPERATION_ABORTED.  FileStream is translating this to an OperationCanceledException and then faulting the task with that exception, rather than canceling the task.  This commit changes it to cancel the task so that it ends in the Canceled rather than Faulted state.

This fixes a regression introduced by #972.

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -1212,7 +1212,7 @@ namespace System.IO
 
             // Create and store async stream class library specific data in the async result
 
-            FileStreamCompletionSource completionSource = new FileStreamCompletionSource(numBufferedBytesRead, bytes, _handle);
+            FileStreamCompletionSource completionSource = new FileStreamCompletionSource(numBufferedBytesRead, bytes, _handle, cancellationToken);
             NativeOverlapped* intOverlapped = completionSource.Overlapped;
 
             // Calculate position in the file we should be at after the read is done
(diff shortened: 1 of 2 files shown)


stephentoub:
@ericstj, I noted that this fixes a regression... in truth it mostly fixes it.  The difference is whether the right CancellationToken is passed to TrySetCanceled.  Previously the CancellationToken passed to the async operation is then passed to TrySetCanceled, such that it's stored into the task, such that any OperationCanceledException thrown from an await includes that token.  This change doesn't include that token.  I went back and forth on whether to do so or not.  In the name of being 100% correct, we should include it.  However, doing so would require us to add an extra reference-sized field to the async object in order to store the token, and it's very rare to actually hit this, so not exactly pay-for-play.  Thoughts?  Should I pass the token through or not?

ericstj:
True, but this seems like the one case where it is interesting to know the cancellation token, right?  I'm not so worried about a pointer sized field on something that is already tied to an I/O operation.  That said, if you want to be clever you can change private readonly SafeFileHandle _handle; to an object, and then store the cancellation token here right before you call CancelIoEx :wink:

stephentoub:
> if you want to be clever you can change private readonly SafeFileHandle _handle; to an object, and then store the cancellation token here right before you call CancelIoEx 

That would box the token.

> True, but this seems like the one case where it is interesting to know the cancellation token, right? I'm not so worried about a pointer sized field on something that is already tied to an I/O operation. 

Ok, I'll pass it through. Thanks for keeping me honest ;)

stephentoub:
Updated.

ericstj:
LGTM :+1:

PR #1759, opened 2015-05-15
Fix SymmetricAlgorithm.Key overflow bug

### LINKED ISSUE #1755: Potential overflow in key size validation in System.Security.Cryptography.SymmetricAlgorithm
The current implementation of `SymmetricAlgorithm.Key` checks the size of the key with unchecked `key.Length * 8` which could overflow.

Thus a byte array with a size of `536870928` would be approved as a 128 bit key.

A first fix could be adding a checked expression for the multiplication.

As a later solution I propose a `public bool ValidKey(byte[])` method as every implementer of `SymmetricAlgorithm` will need their own key size checkes in implementations of `CreateDecryptor(byte[], byte[])` and `CreateEncryptor(byte[], byte[])`.

The same bug is present in the full .NET framework since at least 2012 when I reported it on Connect.
There the different implementations can not handle that unexpected array: 
- `AesManaged`: `IndexOutOfRangeException`
- `AesCryptoServiceProvider`: `OutOfMemoryException`
- ...

stephentoub:
Check that the key's size in bits doesn't overflow Int32.MaxValue, or else we could potentially be allowing in a gigantic key that happens to overflow to a valid key size.

Fixes #1755.

### DIFF
--- a/src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/SymmetricAlgorithm.cs
+++ b/src/System.Security.Cryptography.Encryption/src/System/Security/Cryptography/SymmetricAlgorithm.cs
@@ -70,11 +70,12 @@ namespace System.Security.Cryptography
                 if (value == null)
                     throw new ArgumentNullException("value");
 
-                if (!ValidKeySize(value.Length * 8))
+                long bitLength = value.Length * 8L;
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @bartonjs, @nguerrera

bartonjs:
LGTM

stephentoub:
@dotnet-bot test this please

PR #1760, opened 2015-05-15
Remove outdated comment about packages.config from csproj's

akoeplinger:
Since the move to project.json we no longer use packages.config.

It's currently hit and miss whether a csproj contains the comment about packages.config (86 out of 170 files had them) so I decided to just delete the comment instead of fixing it to project.json (personally I don't think it provides that much value).


JeremyKuhne:
I agree that the comment is better placed in the docs. @weshaggard?

weshaggard:
cc @ericstj
I'm fine with removing comment. When I initially put those there it was to help folks realize they cannot add a reference directly in the project but that isn't going to matter much as we get more tooling in place and folks get used to the new model.

ericstj:
LGTM

PR #1777, opened 2015-05-16
Add initial version of a build.sh script for Linux/OSX

akoeplinger:
~~Focus is on Linux right now, will be adapted in the future.~~

@JeremyKuhne I think this should simplify the Unix build instructions :) we assume MSBuild is available in the same parent folder as corefx. The script also uses the `ReferenceAssemblyRoot` env var you added to MSBuild to avoid symlinking the PCL folder.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+__scriptpath=$(cd "$(dirname "$0")"; pwd -P)
+__msbuildpath=$__scriptpath/../msbuild/bin/Unix/Debug-MONO/MSBuild.exe
+__referenceassemblyroot=/usr/lib/mono/xbuild-frameworks
+__monoversion=$(mono --version | grep "version 4.[1-9]")
+
+if [ $? -ne 0 ]; then
+    echo "Mono 4.1 or later is required to build corefx. Please see https://github.com/dotnet/corefx/wiki/Building-On-Unix for more details."
+    exit 1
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
Cool, looks good. I'll give it a shot sometime this weekend.

weshaggard:
@dotnet-bot test this please

build.sh
```diff
@@ -0,0 +1,22 @@
...
+__reference_assembly_root=/usr/lib/mono/xbuild-frameworks
+__mono_ver=$(mono --version | grep "version 4.[1-9]")
+
+if [ $? -ne 0 ]; then
+    echo "Mono 4.1 or later is required to build corefx."
+    exit 1
+fi
+
+if [ ! -e "$__reference_assembly_root/.NETPortable" ]; then
+    echo "PCL reference assemblies not found."
```
weshaggard: Can you add a pointer to instructions on the wiki for how to set these up?

akoeplinger:
@weshaggard fixed.

I've pushed a slightly improved version that more closely resembles what build.cmd does (i.e. minimal verbosity in the console logger, additional parameters passed through to msbuild, etc). This should also work on OSX, but I haven't tested there yet.

JeremyKuhne:
Validated, thanks Alex!

akoeplinger:
Thank you :)

Issue #1143, opened 2015-03-15
Improve code coverage for System.Linq


stephentoub:
According to the current code coverage report, this library is only at ~24%. Our general goal is closer to 90%, though this will depend on the library. We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

rafidka:
@stephentoub, I am interested in working on this task, but I think I need some more explanation about what I need to do. For example, do I need to go over all methods and see which ones don't have unit tests? If you just give me some guideline then I can take a look and get back to you on whether I can do it or not.

stephentoub:
@rafidka, thanks for your interest.  I've put up some information on a [[Code Coverage]] wiki page.  Let us know if that doesn't address your questions.

stephentoub:
@AlekseyTs, are more tests forthcoming for System.Linq, or is what's up there now the full available set?

Clockwork-Muse:
@rafidka 
I've been working on the Linq.Parallel tests, would any of that be applicable/transferrable/helpful?  I've got tests for all the individual operators (`All`, `Where`, `Select`, etc)... although I think I need to "label" sources so it's easier to tell which one is failing (ie, array versus enumerable input source).

From the look of what's out there (System.Linq/tests), you're going to need to actually _write_ all the tests.  The repository doesn't appear to have anything really useful (it's checking with an empty source only, and expecting success states only).

rafidka:
Thanks for your reply @stephentoub. I just wanted to have a look at the source code, but I couldn't even find the System.Linq project! Or am I missing something?

rafidka:
@Clockwork-Muse yeah, I am sure it will be beneficial to have a look at them. Thanks!

Maxwe11:
@rafidka https://github.com/dotnet/corefx/tree/master/src/System.Linq

rafidka:
Yeah, sorry my bad, I realized I forgot to sync. OK then, I will have a look at it today.

rafidka:
@stephentoub I had a look at the code and I think I can do it. However, do we have a deadline here for getting this done? I don't think this will need months definitely, but just wanted to make sure there is no tight deadline for it.

Also, I noticed that System.Linq lacks documentation and it needs some refactoring (everything is in one big file of around 3000 lines, additionally ReSharper shows 178 warnings in Enumerable.cs). Should I add a task for documentation and another for refactoring?

stephentoub:
[...]

PR #1779, opened 2015-05-16
Enumerable.Sum Tests

MarcinJuraszek:
Unit tests for `Enumerable.Sum()` and `Enumerable.Sum(selector)`, part of https://github.com/dotnet/corefx/issues/1143.

### DIFF
--- a/src/System.Linq/tests/SumTests.cs
+++ b/src/System.Linq/tests/SumTests.cs
@@ -1,0 +1,443 @@
+﻿using System.Collections.Generic;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class SumTests
+    {
+        #region SourceIsNull - ArgumentNullExceptionThrown
+
+        [Fact]
+        public void SumOfInt_SourceIsNull_ArgumentNullExceptionThrown()
+        {
+            IEnumerable<int> sourceInt = null;
+            Assert.Throws<ArgumentNullException>(() => sourceInt.Sum());
+            Assert.Throws<ArgumentNullException>(() => sourceInt.Sum(x => x));
+        }
+
+        [Fact]
+        public void SumOfNullableOfInt_SourceIsNull_ArgumentNullExceptionThrown()
+        {
+            IEnumerable<int?> sourceNullableInt = null;
+            Assert.Throws<ArgumentNullException>(() => sourceNullableInt.Sum());
+            Assert.Throws<ArgumentNullException>(() => sourceNullableInt.Sum(x => x));
+        }
+
+        [Fact]
+        public void SumOfLong_SourceIsNull_ArgumentNullExceptionThrown()
+        {
+            IEnumerable<long> sourceLong = null;
+            Assert.Throws<ArgumentNullException>(() => sourceLong.Sum());
+            Assert.Throws<ArgumentNullException>(() => sourceLong.Sum(x => x));
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


src/System.Linq/src/System/Linq/Enumerable.Sum.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        public static int? Sum(this IEnumerable<int?> source)
+        {
+            if (source == null) throw Error.ArgumentNull("source");
+            int sum = 0;
+            checked
+            {
+                foreach (int? v in source)
+                {
+                    if (v != null) sum += v.GetValueOrDefault();
```
JohanLarsson: Why nullcheck here?

spzSource: Agree. This is redundant code because GetValueOrDefault is used.

MarcinJuraszek: I didn't change that. Just extracted from `Enumerable.cs` to `Enumerable.Sum.cs`. But agree, we might want to remove the `if` statement as part of the change.

src/System.Linq/src/System.Linq.csproj
```diff
@@ -23,6 +23,7 @@
       <Link>Common\System\NotImplemented.cs</Link>
     </Compile>
     <Compile Include="System\Linq\Enumerable.cs" />
+    <Compile Include="System\Linq\Enumerable.Sum.cs" />
```
stephentoub: Why the change to separate out just the Sum functions as part of a PR focused on tests?  Seems like if we want to split this file up, it should be a separate discussion/refactoring, at which point we'd do it for all of the members and not just for Sum.

stephentoub:
Other than my one comment about the source being moved, LGTM.

cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.Sum.cs
VSadov: Just use GetValueOrDefault, since we know we are dealing with an int. Add a comment though that we imply an null check here, but not doing it since the result is the same anyways.

src/System.Linq/src/System.Linq.csproj
VSadov: Yes, if we want to split the file, we should come up with a clear guidance/rule on how to do that and do that considering all operators, not just Sum.
It should be done as a separate change.

src/System.Linq/src/System/Linq/Enumerable.Sum.cs
VSadov: This applies to all "primitive?" cases. Perhaps it is better to do as a separate change and apply to all the cases. There could be other operators using this pattern, and such... 
It would be easier to concentrate on one kind of change.

VSadov:
Same as @stephentoub - nice change, but refactoring into separate files should be done in a separate change.

src/System.Linq/src/System/Linq/Enumerable.Sum.cs
MarcinJuraszek: I agree that it should be fixed as separate change.

src/System.Linq/src/System.Linq.csproj
MarcinJuraszek: OK, reverting that change.

MarcinJuraszek:
Updated the PR to address comments.

stephentoub:
Thanks, @MarcinJuraszek!

PR #1782, opened 2015-05-16
Clean up remaining conflicting version warnings

ericstj:
We have a number of MSB3277 warnings from ResolveAssemblyReferences.  These are due to a ProjectReference pulling in a higher version of a nuget package than the referencing project.  Nuget handles this case and flows the references to the referencing project during restore from Visual Studio, but not from the command line yet.  These warnings have been causing many folks grief so I am checking in a workaround.  The workaround is to reference the package representing the project, then rely on the build targets preferring the ProjectReference over the package reference.


stephentoub:
LGTM

Issue #394, opened 2015-01-11
Add GetOrAdd and AddOrUpdate overloads to ConcurrentDictionary<TKey, TValue> that take a TArg factoryArgument


justinvp:
Consider adding `GetOrAdd` and `AddOrUpdate` overloads to `ConcurrentDictionary<TKey, TValue>` that take a `TArg factoryArgument` parameter, to avoid closure allocations.

This would be consistent with what already ships as part of the immutable collections, where there is an extension method for `ImmutableDictionary<TKey, TValue>` that takes such an argument (see [`ImmutableInterlocked.GetOrAdd<TKey, TValue, TArg>(...)`](https://github.com/dotnet/corefx/blob/1a22d7130c8ff1dffaf0d26c6d5d1433da345ba7/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableInterlocked.cs#L71)).

## Motivating Example

I've run into this a number of times. I'll have a `ConcurrentDictionary<TKey, TValue>` and I'll want to call `GetOrAdd` or `AddOrUpdate` with a `valueFactory`, but the factory needs more than just the key itself to create the value, which results in a closure allocation.

For example:

```c#
int id = 1;
PersonRepository repository = ...;
ConcurrentDictionary<int, Person> dictionary = ...;

// This allocates a closure object due to the use of repository in the lambda
Person person = dictionary.GetOrAdd(id, key => repository.Get(key));
```

Possible workarounds for the above example:
 - Move the `ConcurrentDictionary<int, Person>` usage inside the `PersonRepository`. This isn't always possible to do.
 - Create a `struct` key that contains both the `int` and the `PersonRepository`, just to be able to pass the repository to the factory without the closure allocation. This is gross.

It'd be nice if there was a built-in way to pass an extra argument to the factory, to avoid the closure.

Example after the new overloads are added:

```c#
int id = 1;
PersonRepository repository = ...;
ConcurrentDictionary<int, Person> dictionary = ...;

// No closure allocation (and the delegate instance is created once and cached by the compiler)
Person person = dictionary.GetOrAdd(id, (key, repo) => repo.Get(key), repository);
```

## Proposed API

```c#
public class ConcurrentDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IDictionary
{
    // Proposed members
    public TValue GetOrAdd<TArg>(TKey key, Func<TKey, TArg, TValue> valueFactory, TArg factoryArgument);
[...]

PR #1783, opened 2015-05-17
Add ConcurrentDictionary GetOrAdd/AddOrUpdate overloads with generic arg

stephentoub:
Adds one overload to each of GetOrAdd and AddOrUpdate.  These overloads accept a generic argument that is passed through to any invocations of the supplied delegates, enabling developers to avoid delegate/closure allocations when more input is needed than just the key or existing value.  The implementations are exact copies of the existing overloads then augmented with the additional parameter and threading that through to the delegate invocations.

For AddOrUpdate, there are two existing overloads with delegates, and #394 proposes that a new overload be added for each. I went back and forth on that, and for now this PR only provide a new overload for the one that accepts two delegates, as the other just didn't seem worth it (to my recollection we also didn't discuss both of these overloads in the API review... I hadn't actually realized that a new overload was proposed for each of them).  If folks feel like it's really important to have the third one, though, I can add it.

Fixes https://github.com/dotnet/corefx/issues/394.

### DIFF
--- a/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
+++ b/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
@@ -978,6 +978,37 @@ namespace System.Collections.Concurrent
         /// if the key does not already exist.
         /// </summary>
         /// <param name="key">The key of the element to add.</param>
+        /// <param name="valueFactory">The function used to generate a value for the key</param>
+        /// <param name="factoryArgument">An argument value to pass into <paramref name="valueFactory"/>.</param>
+        /// <exception cref="T:System.ArgumentNullException"><paramref name="key"/> is a null reference
+        /// (Nothing in Visual Basic).</exception>
+        /// <exception cref="T:System.ArgumentNullException"><paramref name="valueFactory"/> is a null reference
+        /// (Nothing in Visual Basic).</exception>
+        /// <exception cref="T:System.OverflowException">The dictionary contains too many
+        /// elements.</exception>
+        /// <returns>The value for the key.  This will be either the existing value for the key if the
+        /// key is already in the dictionary, or the new value for the key as returned by valueFactory
+        /// if the key was not in the dictionary.</returns>
+        public TValue GetOrAdd<TArg>(TKey key, Func<TKey, TArg, TValue> valueFactory, TArg factoryArgument)
+        {
+            if (key == null) throw new ArgumentNullException("key");
+            if (valueFactory == null) throw new ArgumentNullException("valueFactory");
+
+            int hashcode = _comparer.GetHashCode(key);
+
+            TValue resultingValue;
+            if (!TryGetValueInternal(key, hashcode, out resultingValue))
+            {
+                TryAddInternal(key, hashcode, valueFactory(key, factoryArgument), false, true, out resultingValue);
+            }
+            return resultingValue;
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @justinvp, @AlfredoMS, @terrajobst

stephentoub:
@dotnet-bot test this please

justinvp:
@stephentoub, thanks for putting this together!

> for now this PR only provide a new overload for the one that accepts two delegates, as the other just didn't seem worth it

I agree; I'm good with just the single new overload for `AddOrUpdate`.

Question: Is there a need for separate `TAddArg` and `TUpdateArg` args for `AddOrUpdate`? Separate args weren't in the original proposal, but I now see that the extension method example you provided had separate args. The single `TArg` is sufficient for the cases I've run into and I like the simplicity of the single `TArg`, so I'm happy with the PR as-is. Just wanted to double-check.

Minor naming question: [`ImmutableInterlocked.GetOrAdd`](https://github.com/dotnet/corefx/blob/b8514c4fb4c4d052669890ab2f9a09d87a022684/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableInterlocked.cs#L157) has a similar overload that takes a `TArg`, but instead of `arg` it is named `factoryArgument`. Should the parameter name be `factoryArgument` here as well to be consistent? I don't have a strong preference either way.

stephentoub:
> The single TArg is sufficient for the cases I've run into and I like the simplicity of the single TArg, so I'm happy with the PR as-is. 

Yeah, that was my reasoning as well (that, and if you really need different args you can package them up into a struct, and then just pay the copying cost of the one that isn't needed).

> Minor naming question

No strong preference, either.  @KrzysztofCwalina, suggestions?

stephentoub:
@dotnet-bot test this please

KrzysztofCwalina:
I agree that factoryArgument would be better. I think it clarifies what the argument is for.
Otherwise looks good! Thanks.

stephentoub:
Thanks, guys; fixed the new parameter's name.

stephentoub:
@dotnet-bot test this please

justinvp:
nit: `factoryArgument` or `factoryArg`?

stephentoub:
Oh, ok.

stephentoub:
Let's try that again ;)

src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
```diff
@@ -1008,6 +1039,59 @@ public TValue GetOrAdd(TKey key, TValue value)
...
+            int hashcode = _comparer.GetHashCode(key);
+
+            while (true)
+            {
+                TValue oldValue;
+                if (TryGetValueInternal(key, hashcode, out oldValue))
+                {
+                    // key exists, try to update
+                    TValue newValue = updateValueFactory(key, oldValue, factoryArgument);
+                    if (TryUpdate(key, newValue, oldValue))
```
justinvp: Sorry I didn't catch this earlier, but should this be `TryUpdateInternal` to reuse the once-computed `hashcode`?

stephentoub: Yes, it should be.  This was just copy-and-paste from the current overload, which highlights that it, too, is calling TryUpdate instead of TryUpdateInternal.  I'll fix this one here, but submit a separate PR to master to fix the current overload.

stephentoub: Fixed. And opened #1833 to address the existing TryUpdate calls.

PR #1785, opened 2015-05-17
Dataflow: Use forward slashes in path of included XML doc file 

akoeplinger:
The previous path resulted in a bunch of warnings when building corefx on Linux, since the XML doc file couldn't be found due to backslashes.

The fix is to change the path to forward slashes as those work on Windows as well.


stephentoub:
LGTM

Issue #1775, opened 2015-05-15
Two warnings when building System.Text.Encodings.Web


stephentoub:
This is with VS2015 RC:
```
System\Text\Encodings\Web\TextEncoder.cs(9,37): warning CS3011: 'TextEncoder.TryEncodeUnicodeScalar(int, char*, int, out int)': only CLS-compliant members can
be abstract [c:\Users\stoub\Source\Repos\corefx\src\System.Text.Encodings.Web\src\System.Text.Encodings.Web.csproj]
System\Text\Encodings\Web\TextEncoder.cs(14,36): warning CS3011: 'TextEncoder.FindFirstCharacterToEncode(char*, int)': only CLS-compliant members can be abstra
ct [c:\Users\stoub\Source\Repos\corefx\src\System.Text.Encodings.Web\src\System.Text.Encodings.Web.csproj]
```

KrzysztofCwalina:
Fixed

PR #1789, opened 2015-05-18
Fixes a build warning.

KrzysztofCwalina:
I added a pragma disables a warning complaining about non-CLS compliant members being abstract, and wants me to mark the type as non-CLS compliant. And it is true that this type cannot be extended by all CLS compliant languages. Having said that, if I marked the type as non-CLS all methods that take it as parameter will now have to be marked CLSCompliant(false), yet consumption of concrete encoders is totally CLS compliant, as it’s mainly to be done by calling helper methods in TextEncoderExtensions class, and so I think the warning is a bit too aggressive.

### DIFF
--- a/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/TextEncoder.cs
+++ b/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/TextEncoder.cs
@@ -2,9 +2,17 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 namespace System.Text.Encodings.Web
-{
+{    
     public abstract class TextEncoder
     {
+        // The following pragma disables a warning complaining about non-CLS compliant members being abstract, 
+        // and wants me to mark the type as non-CLS compliant. 
+        // It is true that this type cannot be extended by all CLS compliant languages. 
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
@stephentoub, this fixes #1775

stephentoub:
> yet consumption of *concrete* encoders is totally CLS compliant

But consumption of the abstract base TextEncoder class is not?

KrzysztofCwalina:
I was not super clear. Let me clarify: the consumption of both the base type and the subtypes using the extensions methods is CLS compliant, i.e. almost nobody should/would consume the non-CLS compliant instance methods on these types. The methods are just low level APIs for the extension methods to call and for people who will want to implement new encoders, which will be extremely rare and will have to be done in a language that supports pointers.

stephentoub:
Thanks for the explanation.  If it's possible to do everything you'd want to do with these from non-CLS compliant languages, my concerns are lessened.  Still seems unfortunate that we'd introduce new abstract classes you can't implement in other languages, but if it's really going to be super rare, I defer to your judgement.  Otherwise, the actual change itself looks fine; if it were me, I'd still put a pragma warning disable at the end of the relevant scope, but it's up to you.

KrzysztofCwalina:
I added pragma warning restore.
Thanks.

KrzysztofCwalina:
@dotnet-bot test this please

PR #1796, opened 2015-05-19
Consume MSBuild Mono package

JeremyKuhne:
Add the Mono MSBuild package and update the build script to consume it and restore tools via xbuild.

@weshaggard, @akoeplinger, @valmenn, @ellismg, @joshfree

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
-__msbuildpath=$__scriptpath/../msbuild/bin/Unix/Debug-MONO/MSBuild.exe
+__packageroot=$__scriptpath/packages
+__msbuildpath=$__packageroot/Microsoft.Build.Mono.Debug.14.1.0.0-prerelease/lib/MSBuild.exe
 __referenceassemblyroot=/usr/lib/mono/xbuild-frameworks
 __monoversion=$(mono --version | grep "version 4.[1-9]")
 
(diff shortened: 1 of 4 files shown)


weshaggard:
LGTM

ellismg:
Is there a way we could not have the xbuild dependency?

ellismg:
(e.g, why can we not just download the .nupkg ourselves with curl and unzip it?  That seems like less work than requiring all of xbuild?)

JeremyKuhne:
@ellismg We could, but we already require mono/xbuild and that isn't likely to go away any time soon (MSBuild requires it). I considered scripting out the tool restore here, but went with expedient for right now.

ellismg:
Okay.  Can you help me understand why we require xbuild now aside from this (I understand if you want to build msbuild from scratch you need it, but with this change you shouldn't have to build it from scratch anymore, right?)

akoeplinger:
@ellismg I had the same thoughts. Instead of unzipping we should rather just fetch nuget.exe and use it to install msbuild via `nuget install`. We could do it on Windows as well and get rid of the DownloadFile inline task.

@JeremyKuhne I don't see why we would need xbuild with this change, the pcl reference assemblies can be installed without pulling in xbuild afaik.

JeremyKuhne:
@akoeplinger @ellismg I would rather just use the script to pull, this was a fast first step. I have about 10 minutes worth of experience in writing bash scripts. :) If we didn't already have xbuild in place from our existing dependencies I wouldn't have used it.

akoeplinger:
Yeah, makes sense. LGTM.

ellismg:
LGTM.  Thanks!

JeremyKuhne:
@dotnet-bot test this please

PR #1799, opened 2015-05-20
Add System.Security.Cryptography.Hashing Source and Tests

bartonjs:
Resolves https://github.com/dotnet/corefx/issues/835

### DIFF
--- a/src/System.Security.Cryptography.Hashing/src/System/Security/Cryptography/KeyedHashAlgorithm.cs
+++ b/src/System.Security.Cryptography.Hashing/src/System/Security/Cryptography/KeyedHashAlgorithm.cs
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+
+namespace System.Security.Cryptography
+{
+    public abstract class KeyedHashAlgorithm : HashAlgorithm
+    {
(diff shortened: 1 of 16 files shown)


src/System.Security.Cryptography.Hashing/src/System/Security/Cryptography/KeyedHashAlgorithm.cs
```diff
@@ -0,0 +1,45 @@
...
+        public virtual byte[] Key
+        {
+            get
+            {
+                return _key.CloneByteArray();
+            }
+
+            set
+            {
+                _key = value.CloneByteArray();
```
ellismg: Desktop had some code to throw if you tried to change the key after hashing began.  Is this something the subclasses need to do now?  Do we not have the code because the checking was only done in the Transform methods which we don't have here?

bartonjs: Yeah, the only time that the key could meaningfully change while a hash is in progress with this contract is a threaded assignment during `CalculateHash(Stream)`.  That looks like it will result in calculating the hash as a skip of however many bytes were already read (from the current implementations).  But, the usual 'this isn't multi-threading-hardened' caveats apply.

ellismg: Gotcha

ellismg:
LGTM

PR #1804, opened 2015-05-20
Remove xbuild dependency.

JeremyKuhne:
Pull down nuget and msbuild in the build.sh script. Check for cURL and fail out if missing.

@weshaggard, @akoeplinger, @chcosta, @ellismg

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,12 @@
 
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
 __packageroot=$__scriptpath/packages
-__msbuildpath=$__packageroot/Microsoft.Build.Mono.Debug.14.1.0.0-prerelease/lib/MSBuild.exe
+__sourceroot=$__scriptpath/src
+__nugetpath=$__packageroot/NuGet.exe
+__nugetconfig=$__sourceroot/NuGet.Config
+__msbuildpackageid="Microsoft.Build.Mono.Debug"
+__msbuildpackageversion="14.1.0.0-prerelease"
+__msbuildpath=$__packageroot/$__msbuildpackageid.$__msbuildpackageversion/lib/MSBuild.exe
 __referenceassemblyroot=/usr/lib/mono/xbuild-frameworks
 __monoversion=$(mono --version | grep "version 4.[1-9]")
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -19,8 +24,30 @@ fi
 __buildproj=$__scriptpath/build.proj
 __buildlog=$__scriptpath/msbuild.log
 
-# Run the restore build tools under xbuild to get NuGet and MSBuild pulled down
-xbuild "$__buildproj" /t:_RestoreBuildTools /nologo /verbosity:minimal "/fileloggerparameters:Verbosity=diag;LogFile=$__buildlog"
+# Pull NuGet.exe down if we don't have it already
+if [ ! -e "$__nugetpath" ]; then
+    which curl > /dev/null 2> /dev/null
+    if [ $? -ne 0 ]; then
```
akoeplinger: We should check for wget as well, it's very likely that at least one of the two is installed on almost all systems.

build.sh
```diff
@@ -19,8 +24,30 @@ fi
...
+if [ ! -e "$__msbuildpath" ]; then
+    echo "Restoring MSBuild..."
+    mono "$__nugetpath" install $__msbuildpackageid -Version $__msbuildpackageversion -ConfigFile "$__nugetconfig" -OutputDirectory "$__packageroot"
+    if [ $? -ne 0 ]; then
+        echo "Failed to restore MSBuild."
+        exit 1
+    fi
+fi
 
 if [ ! -e "$__msbuildpath" ]; then
```
akoeplinger: Do we still need this check? If something goes wrong during nuget install of MSBuild in the previous step we should get an error anyway.

build.sh
JeremyKuhne: Any reason to use cURL over wget? wget seems to be installed by default on Ubuntu?

build.sh
JeremyKuhne: It is pretty unlikely to get hit, true. I can remove it if you think it is confusing or generally a net negative.

build.sh
akoeplinger: wget doesn't ship with OSX, but curl does. We need to check for both :)

build.sh
akoeplinger: I think it complicates the script for little value, so I'd prefer to just remove it.

build.sh
JeremyKuhne: Good to know, thanks. Need to get a Mac. :)

build.sh
JeremyKuhne: Fair enough, removing.

JeremyKuhne:
Updated to look for cURL or wget.

akoeplinger:
LGTM, thanks :+1:

ellismg:
Nice!  LGTM.

JeremyKuhne:
@dotnet-bot test this please

PR #1814, opened 2015-05-21
Fixed some issues based on a Code Analysis run

KrzysztofCwalina:


JeremyKuhne:
LGTM

bpschoch:
var ex = Assert.Throws<ArgumentOutOfRangeException>(() => new UnicodeRange(firstCodePoint, rangeSize));
+            Assert.Equal("length", ex.ParamName);

These kind of asserts can be simplified as:

Assert.Throws<ArgumentOutOfRangeException>("length", () => new UnicodeRange(firstCodePoint, rangeSize));

KrzysztofCwalina:
@bpschoch, it's cute, but is it more readable?

bpschoch:
I started to use this notation in tests I've written after I saw it used in other parts of the library.  It keeps the tests terse especially if you have a sequence of asserts.

stephentoub:
And in such cases it avoids the introduction of unnecessary locals, which I personally find keeps things more readable.

stephentoub:
LGTM

KrzysztofCwalina:
When I work on tests with multiple lines like that, I will consider this overload. Thanks for the tip.

PR #1816, opened 2015-05-21
Add System.Security.Cryptography.Encoding Source and Tests

bartonjs:
Resolves https://github.com/dotnet/corefx/issues/833.

Will create an issue about OidInfo not matching Interop files style.

### DIFF
--- a/src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/AsnFormatter.Unix.cs
+++ b/src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/AsnFormatter.Unix.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Internal.Cryptography
+{
+    internal abstract partial class AsnFormatter
+    {
+        private static readonly AsnFormatter s_instance = new OpenSslAsnFormatter();
+
+        // -----------------------------
+        // ---- PAL layer ends here ----
+        // -----------------------------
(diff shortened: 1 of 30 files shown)


src/System.Security.Cryptography.Encoding/tests/OidCollection.cs
```diff
@@ -0,0 +1,92 @@
...
+            i = c.Add(o0);
+            Assert.Equal(0, i);
+
+            Oid o1 = new Oid(SHA256_Oid, SHA256_Name);
+            i = c.Add(o1);
+            Assert.Equal(1, i);
+
+            Assert.Equal(2, c.Count);
+
+            Assert.True(Object.ReferenceEquals(o0, c[0]));
```
Maxwe11: `Assert.Same`

bartonjs: Good eye, fixed up the lot of them.

src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/AsnFormatter.Unix.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    internal abstract partial class AsnFormatter
+    {
+        private static readonly AsnFormatter s_instance = new OpenSslAsnFormatter();
+
+        // -----------------------------
+        // ---- PAL layer ends here ----
+        // -----------------------------
+    }
+}
```
stephentoub: Nit: some files are missing newlines at the end

stephentoub:
LGTM

PR #1818, opened 2015-05-21
Initial commit of System.IO.Packaging

EricWhiteDev:
New version of System.IO.Packaging, rewritten to use System.IO.Compression.  This assembly is used by the Open XML SDK.  The code was written to work with both .NET 4.5.* and COREFX/CORECLR.  I will be adding additional tests in the near future.  In addition, I have a massive amount of tests that will go into "outer-loop" testing after some cleanup.


stephentoub:
@dotnet-bot test this please

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,744 @@
...
+        #region Private Methods
+
+
+        /// <summary>
+        /// This method validates if the content type string has
+        /// valid CR-LF characters. Specifically we test if '\r' is 
+        /// accompanied by a '\n' in the string, else its an error.
+        /// </summary>
+        /// <param name="contentType"></param>
+        private static void ValidateCarriageReturns(string contentType)
```
stephentoub: Is this going to be correct on platforms where newlines are just "\n"?

src/System.IO.Packaging/src/System/IO/Packaging/Invariant.cs
```diff
@@ -0,0 +1,246 @@
...
+        ///     TreatAsSafe: This code is safe to call.
+        ///         Note that if this code were made to be callable publicly,
+        ///         we would have a potential denial-of-service vulnerability.
+        /// </SecurityNote>
+        // [SecurityCritical, SecurityTreatAsSafe] todo ew
+        // [SecurityCritical]
+        private // DO NOT MAKE PUBLIC OR INTERNAL -- See security note
+            static void FailFast(string message, string detailMessage)
+        {
+            Debug.Assert(false, string.Format("Invariant failure: {0}\n{1}", message, detailMessage));
```
stephentoub: Ideally we'd use Environment.NewLine for newlines, rather than hardcoding "\r\n" or "\n".

src/System.IO.Packaging/src/System.IO.Packaging.csproj
```diff
@@ -0,0 +1,52 @@
...
+    <Compile Include="System\IO\Packaging\TargetMode.cs" />
+    <Compile Include="System\IO\Packaging\XmlCompatibilityReader.cs" />
+    <Compile Include="System\IO\Packaging\XmlWrappingReader.cs" />
+    <Compile Include="System\IO\Packaging\ZipPackage.cs" />
+    <Compile Include="System\IO\Packaging\ZipPackagePart.cs" />
+    <Compile Include="System\IO\Packaging\ZipStreamManager.cs" />
+    <Compile Include="System\IO\Packaging\ZipWrappingStream.cs" />
+  </ItemGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+</Project>
```
stephentoub: Nit: some of the files are missing newlines at the end

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,744 @@
...
+
+        private string _contentType = null;
+        private string _type = String.Empty;
+        private string _subType = String.Empty;
+        private string _originalString;
+        private Dictionary<string, string> _parameterDictionary = null;
+        private bool _isInitialized = false;
+
+        private const string _quote = "\"";
+        private const char _semicolonSeparator = ';';
```
stephentoub: Our style guidelines say that constants should begin with a capital rather than an underscore, e.g. SemicolonSeparator.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,744 @@
...
+        {
+            if (String.CompareOrdinal(token, String.Empty) == 0)
+                throw new ArgumentException(SR.Get(SRID.InvalidToken));
+
+            for (int i = 0; i < token.Length; i++)
+            {
+                if (IsAsciiLetterOrDigit(token[i]))
+                    continue;
+                else
+                    if (IsAllowedCharacter(token[i]))
```
stephentoub: Nit: odd formatting of the else if

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,744 @@
...
+        /// <summary>
+        /// Returns true if the input character is an allowed character
+        /// Returns false if the input cha----ter is not an allowed character
+        /// </summary>
+        /// <param name="character">input character</param>
+        /// <returns></returns>
+        private static bool IsAllowedCharacter(char character)
+        {
+            //We did not use any of the .Contains methods as
+            //it will result in boxing costs.
```
stephentoub: Couldn't it use Array.IndexOf<char>?

src/System.IO.Packaging/src/System/IO/Packaging/Invariant.cs
```diff
@@ -0,0 +1,246 @@
...
+{
+    using System;
+    using System.Security;
+    using System.Diagnostics;
+
+    /// <summary>
+    /// Provides methods that assert an application is in a valid state. 
+    /// </summary>
+    internal // DO NOT MAKE PUBLIC - See security notes on Assert
+        static class Invariant
```
stephentoub: Why not just have the library use Debug.Assert?

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1312 @@
...
+        private static string EscapeSpecialCharacters(string path)
+        {
+            string characterString;
+
+            // Escaping for the following - '%'; '@'; ',' and '?'
+            // !!Important!! - The order is important - The '%' sign should be escaped first.
+            // This is currently enforced by the order of characters in the _specialCharacters array
+            foreach (char c in s_specialCharacters)
+            {
+                characterString = c.ToString();
```
stephentoub: I don't know if allocations matter here, but you could trivially avoid this allocation by using IndexOf instead of Contains.

src/System.IO.Packaging/src/System/IO/Packaging/PackagePartCollection.cs
```diff
@@ -0,0 +1,141 @@
...
+
+//-----------------------------------------------------------------------------
+//
+// Description:
+//  This is a base abstract class for PackagePartCollection. This is a part of the 
+//  MMCF Packaging Layer
+//
+//-----------------------------------------------------------------------------
+
+// Allow use of presharp warning numbers [6506] unknown to the compiler
```
stephentoub: Is "presharp" still being used?

src/System.IO.Packaging/src/System/IO/Packaging/SRID.cs
```diff
@@ -0,0 +1,305 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace System.IO.Packaging
+{
+    internal class SRID
```
stephentoub: Should this library use the same resources design used by the rest of the corefx libraries?

src/System.IO.Packaging/src/System/IO/Packaging/SR.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+
+namespace System.IO.Packaging
+{
+    internal class SR
+    {
+        public static string Get(string arg, params object[] otherArgs)
+        {
+            return "";
```
stephentoub: ?

stephentoub:
@EricWhiteDev, just following up... what's the status of this?

stephentoub:
@dotnet-bot test this please

stephentoub:
Two tests are also failing on Unix:
```
System.IO.Packaging.Tests.Tests.T035_ModifyAllPackageProperties [FAIL]
08:06:16       Assert.Equal() Failure
08:06:16                                        ��� (pos 21)
08:06:16       Expected: ������p.Category: >(null)<\npp.ContentStatus: >(null)<\npp.ContentTyp������
08:06:16       Actual:   ������p.Category: >(null)<\r\npp.ContentStatus: >(null)<\r\npp.ContentT������
08:06:16                                        ��� (pos 21)
08:06:16       Stack Trace:
08:06:16             at System.IO.Packaging.Tests.Tests.T035_ModifyAllPackageProperties()
08:06:16    System.IO.Packaging.Tests.Tests.T029_DeletePackageRelationship [FAIL]
08:06:16       Assert.Equal() Failure
08:06:16                   ��� (pos 2)
08:06:16       Expected: #0\nUri: /docProps/app.xml\nContentType: appl������
08:06:16       Actual:   #0\r\nUri: /docProps/app.xml\r\nContentType: ap������
08:06:16                   ��� (pos 2)
08:06:16       Stack Trace:
08:06:16             at System.IO.Packaging.Tests.Tests.T029_DeletePackageRelationship()
08:06:16 Finished:    System.IO.Packaging.Tests
08:06:16 
08:06:16 === TEST EXECUTION SUMMARY ===
08:06:16    System.IO.Packaging.Tests  Total: 89, Errors: 0, Failed: 2, Skipped: 0, Time: 2.187s
```
The relevant files aren't showing up in the diff, but I believe the issue is that the source is using verbatim string literals that span lines, such that the newline on the platform on which compilation is happening is being embedded in the string, rather than the newline on which the test is running.  The code should be changed to avoid verbatim strings spanning newlines.

stephentoub:
@EricWhiteDev, any updates on this?

EricWhiteDev:
@stephentoub I've been on vacation for the past few days.  I have made progress on making appropriate changes, but have not finished.  I have some questions on things, and will contact you or Matt regarding those.

stephentoub:
Ok, thanks for the update.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
EricWhiteDev: This is a very interesting issue.  RFC-2616 (definition of content types, as referred to by OPC) does not allow for CRLF (or CR or LF) in a content type.  Further, I did a check in my sample documents, and no Open XML documents contain a CR, LF, or CRLF in a content type.  I suspect that this code has never been executed in a real scenario.

However, I am loath to change this code, as changing it could potentially impact existing applications.  There may be some scenario where programs are setting content types with CRLF in them, although I can't think of what that would be.

One of my goals with this module is to implement as closely as possible the existing implementation.  There are certainly aspects of System.IO.Packaging that could be improved upon, such as adding new functionality to make it more LINQ friendly, and so on.  However, I propose that this is a separate project from the project of getting an implementation that is as compatible as possible as the existing one.

I propose that we leave the code as-is for now, and if there are issues that crop up in the Linux or Mac world, we address those as a separate work item.

src/System.IO.Packaging/src/System/IO/Packaging/Invariant.cs
EricWhiteDev: Fixed.  I also searched through out, looking for other places where \n is used instead of Environment.NewLine, and there were no other occurrences.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
EricWhiteDev: Fixed throughout.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
EricWhiteDev: Fixed.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
EricWhiteDev: Yes, this certainly could be re-written.  However, I am reluctant to change this - this code is quite old - I took from the original System.IO.Packaging, as is (with a few super-simple changes such as fixing the formatting of the if statement previously referenced).  As much as is possible, I want to retain, at least for now, the majority of the code as is, particularly code that could potentially impact functionality.  I propose we leave the code as-is for now.

src/System.IO.Packaging/src/System/IO/Packaging/Invariant.cs
EricWhiteDev: Agreed.  Changed throughout.  Removed Invariant.cs.

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
EricWhiteDev: I don't think allocations matter here.  This is one of those cases where I think that we should leave the code as-is.

src/System.IO.Packaging/src/System/IO/Packaging/PackagePartCollection.cs
EricWhiteDev: Removed all comments referencing presharp, also removed pragmas related to presharp.

src/System.IO.Packaging/src/System.IO.Packaging.csproj
EricWhiteDev: Went through every file, made sure that all files have one and exactly one newline at the end.

EricWhiteDev:
Fixed the issue with tests that have strings that span newlines, as those tests failed on Unix.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
stephentoub: Ok.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
stephentoub: Ok.

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
stephentoub: I'd be ok with not fixing it as part of the initial check-in, but this is a case where it should be trivial to avoid multiple allocations and improve code readability at the same time, so I think we should fix things like this. If you'd prefer it be done separately, can you open an issue to track fixing it?

src/System.IO.Packaging/src/project.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": {
+    "System.Collections": "4.0.10-beta-23008",
```
stephentoub: I believe we're trying to avoid hardcoding revision numbers, e.g. these should end with "-*" instead of "-23008".  @ericstj can probably help if you run into issues.

stephentoub:
@dotnet-bot test this please

src/System.IO.Packaging/src/project.json
mellinoe: Right now the latest packages aren't compatible with the version of DNX we are using, so if you use the * syntax you will get an incompatible package.

ericstj: That's right but we're about to fix this.  What I'd like you to do is the following.
1. Use static versions in the project.json.
2. Do a restore.
3. Modify the project.json to contain *, then modify the lock file to set locked = true, and use *s in the versions in the projectFileDependencyGroups at the end of the file, matching the change in project.json.
4. Confirm that future restores do not modify this file.
This will help ensure that the project is set up correctly so that future updates will happen automatically.

stephentoub:
@dotnet-bot test this please

src/System.IO.Packaging/src/project.json
ericstj: You no longer need to do this since I've updated the DNX version.  You should be able to switch back to using *s.

EricWhiteDev: ok, thanks

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        //------------------------------------------------------
+
+        #region Nested Classes
+
+        /// <summary>
+        /// Comparer class makes it easier to put ContentType objects in collections.
+        /// Only compares type and subtype components of the ContentType.  Could be
+        /// expanded to optionally compare parameters as well.
+        /// </summary>
+        internal class StrongComparer : IEqualityComparer<ContentType>
```
stephentoub: Is this type used anywhere?  It appears to be dead code.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+            /// casing passed into the constructor.  ContentTypes that are equal (which
+            /// ignores casing) must have the same hash code.
+            /// </summary>
+            public int GetHashCode(ContentType obj)
+            {
+                return obj.ToString().ToUpperInvariant().GetHashCode();
+            }
+        }
+
+        internal class WeakComparer : IEqualityComparer<ContentType>
```
stephentoub: Same for this one.

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+            //are the same as per the current implementation of System.Uri
+            //Need to use OriginalString property or ToString() here as we are want to 
+            //validate that the input uri given to us was valid in the first place.
+            //We do not want to use GetComponents in this case as it might lead to
+            //performing escaping or unescaping as per the UriFormat enum value and that 
+            //may alter the string that the user created the Uri with and we may not be able 
+            //to verify the uri correctly.
+            //We perform the comparison in a case-insensitive manner, as at this point,
+            //only escaped hex digits (A-F) might vary in casing.
+            if (String.CompareOrdinal(partUri.OriginalString.ToUpperInvariant(), wellFormedPartName.ToUpperInvariant()) != 0)
```
stephentoub: Could this just do a ordinal-ignorecase comparison rather than allocating new strings with ToUpperInvariant?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+                Debug.Assert(String.CompareOrdinal(_type, String.Empty) != 0
+                   || String.CompareOrdinal(_subType, String.Empty) != 0);
+
+                StringBuilder stringBuilder = new StringBuilder(_type);
+                stringBuilder.Append(s_forwardSlashSeparator[0]);
+                stringBuilder.Append(_subType);
+
+                if (_parameterDictionary != null && _parameterDictionary.Count > 0)
+                {
+                    foreach (string paramterKey in _parameterDictionary.Keys)
```
stephentoub: nit: "paramterKey" => "parameterKey"

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+            if (contentType == null)
+                throw new ArgumentNullException("contentType");
+
+            if (String.CompareOrdinal(contentType, String.Empty) == 0)
+            {
+                _contentType = String.Empty;
+            }
+            else
+            {
+                if (IsLinearWhiteSpaceChar(contentType[0]) || IsLinearWhiteSpaceChar(contentType[contentType.Length - 1]))
```
stephentoub: It doesn't appear that we know at this point that contentType isn't empty.  Couldn't this result in indexing out of bounds into contentType?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+                    if (_parameterDictionary != null)
+                    {
+                        if (_parameterDictionary.Count > 0)
+                            return false;
+                    }
+
+                    //Return false if the content type object passed in has parameters
+                    Dictionary<string, string>.Enumerator contentTypeEnumerator;
+                    contentTypeEnumerator = contentType.ParameterValuePairs;
+                    contentTypeEnumerator.MoveNext();
```
stephentoub: Do we know that MoveNext is always going to return true here such that Current will be valid?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+                {
+                    index = contentType.IndexOf(s_linearWhiteSpaceChars[2], ++index);
+                }
+                else
+                    throw new ArgumentException(SR.InvalidLinearWhiteSpaceCharacter);
+            }
+        }
+
+        /// <summary>
+        /// Parses the type ans subType tokens from the string.
```
stephentoub: nit: "ans" => "and"

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        /// Also verifies if the Tokens are valid as per the grammar.
+        /// </summary>
+        /// <param name="typeAndSubType">substring that has the type and subType of the content type</param>
+        /// <exception cref="ArgumentException">If the typeAndSubType parameter does not have the "/" character</exception>
+        private void ParseTypeAndSubType(string typeAndSubType)
+        {
+            //okay to trim at this point the end of the string as Linear White Spaces(LWS) chars are allowed here.
+            typeAndSubType = typeAndSubType.TrimEnd(s_linearWhiteSpaceChars);
+
+            string[] splitBasedOnForwardSlash = typeAndSubType.Split(s_forwardSlashSeparator);
```
stephentoub: There's a lot of string manipulation / allocation / etc. happening in these various helper functions.  Doesn't need to be fixed now, as it's not clear that it's a problem.  But it'd be interesting to do some perf testing / profiling and see what pops as hot spots that we'll potentially want to fix.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+                {
+                    if (IsAllowedCharacter(token[i]))
+                    {
+                        continue;
+                    }
+                    else
+                    {
+                        throw new ArgumentException(SR.InvalidToken);
+                    }
+                }
```
stephentoub: The body of this loop could be simplified to just:
```C#
if (!IsAsciiLetterOrDigit(token[i]) && !IsAllowedCharacter(token[i]))
    throw new ArgumentException(SR.InvalidToken);
```

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        /// The following checks are being made - 
+        /// 1. If all the characters in the token are either ASCII letter or digit.
+        /// 2. If all the characters in the token are either from the remaining allowed cha----ter set.
+        /// </summary>
+        /// <param name="token">string token</param>
+        /// <returns>validated string token</returns>
+        /// <exception cref="ArgumentException">If the token is Empty</exception>
+        private static string ValidateToken(string token)
+        {
+            if (String.CompareOrdinal(token, String.Empty) == 0)
```
stephentoub: Should this be:
```C#
if (string.IsNullOrEmpty(token))
```
?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        /// <summary>
+        /// Validating if the value of a parameter is either a valid token or a 
+        /// valid quoted string
+        /// </summary>
+        /// <param name="parameterValue">paramter value string</param>
+        /// <returns>validate parameter value string</returns>
+        /// <exception cref="ArgumentException">If the paramter value is empty</exception>
+        private static string ValidateQuotedStringOrToken(string parameterValue)
+        {
+            if (String.CompareOrdinal(parameterValue, String.Empty) == 0)
```
stephentoub: Same question as above.  It seems a bit odd to be explicit about the kind of comparison we want for the characters and to be doing a Compare rather than Equal when what we're checking for is just an empty string.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+            if (IsAsciiLetter(character))
+            {
+                return true;
+            }
+            if (character >= '0')
+            {
+                return (character <= '9');
+            }
+            return false;
+        }
```
stephentoub: Probably just a style thing, but I'd personally have written this whole body as just:
```C#
return IsAsciiLetter(character) || (character >= '0' && character <= '9');
```

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        /// <param name="ch">input character</param>
+        /// <returns></returns>
+        private static bool IsLinearWhiteSpaceChar(char ch)
+        {
+            if (ch > ' ')
+            {
+                return false;
+            }
+
+            foreach (char c in s_linearWhiteSpaceChars)
```
stephentoub: This loop could be replaced by a call to ```Array.IndexOf<char>```.

src/System.IO.Packaging/src/System/IO/Packaging/FileFormatException.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </SecurityNote>
+        public Uri SourceUri
+        {
+            get
+            {
+                return _sourceUri;
+            }
+        }
+
+        private Uri _sourceUri;
```
stephentoub: readonly?

src/System.IO.Packaging/src/System/IO/Packaging/IgnoreFlushAndCloseStream.cs
```diff
@@ -0,0 +1,225 @@
...
+        /// </summary>
+        /// <value>Bool, true if the stream can be read from, else false</value>
+        public override bool CanRead
+        {
+            get
+            {
+                if (_disposed)
+                    return false;
+                else
+                    return _stream.CanRead;
```
stephentoub: Again, maybe a style thing, but I'd have written the whole body as:
```C#
get { return !_disposed && _stream.CanRead; }
```
I'll stop commenting on these.

src/System.IO.Packaging/src/System/IO/Packaging/IgnoreFlushAndCloseStream.cs
```diff
@@ -0,0 +1,225 @@
...
+//------------------------------------------------------------------------------
+
+using System;
+using System.IO;
+
+namespace System.IO.Packaging
+{
+    /// <summary>
+    /// This class ignores all calls to Flush() and Close() methods
+    /// depending on whether the IgnoreFlushAndClose property is set to true
```
stephentoub: This comment appears out-of-date: there is no IgnoreFlushAndClose property, and Flush/Close never delegate to the wrapped stream's corresponding members.

src/System.IO.Packaging/src/System/IO/Packaging/InternalRelationshipCollection.cs
```diff
@@ -0,0 +1,703 @@
...
+                                        ProcessRelationshipAttributes(reader);
+
+                                        //Skip the EndElement for Relationship
+                                        if (!reader.IsEmptyElement)
+                                            ProcessEndElementForRelationshipTag(reader);
+                                    }
+                                    else throw new XmlException(SR.RelationshipTagDoesntMatchSchema, null, reader.LineNumber, reader.LinePosition);
+                                }
+                                else
+                                    if (!(String.CompareOrdinal(s_relationshipsTagName, reader.LocalName) == 0 && (reader.NodeType == XmlNodeType.EndElement)))
```
stephentoub: nit: formatting off?

src/System.IO.Packaging/src/System/IO/Packaging/InternalRelationshipCollection.cs
```diff
@@ -0,0 +1,703 @@
...
+            }
+            else
+            {
+                if (target.IsAbsoluteUri)
+                {
+                    if (String.CompareOrdinal(target.Scheme, PackUriHelper.UriSchemePack) == 0)
+                        return PackUriHelper.GetPartUri(target);
+                }
+                else
+                    Debug.Assert(false, "Uri should not be relative at this stage");
```
stephentoub: Debug.Fail?

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+            string file = Path.GetFileName(partName);
+
+            Debug.Assert((partName.Length - file.Length) > 0,
+                "The partname may not be wellformed");
+
+            // Get the parname without the last segment
+            partName = partName.Substring(0, partName.Length - file.Length);
+
+            partName = Path.Combine(partName, s_relationshipPartSegmentName);         // Adding the "_rels" segment
+            partName = Path.Combine(partName, file);                                 // Adding the last segment back
```
stephentoub: Any reason not to combine these Path.Combine calls?

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+        
+        private static readonly char[] HexUpperChars = {
+                                   '0', '1', '2', '3', '4', '5', '6', '7',
+                                   '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
+        
+        private static string HexEscape(char character) {
+            if (character > '\xff') {
+                throw new ArgumentOutOfRangeException("character");
+            }
+            char[] chars = new char[3];
```
stephentoub: I have no idea if this function is perf critical or not, but the chars array could be stackalloc'd instead of heap allocated.

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+        {
+            Debug.Assert(partName != null, "Null reference check for this partName parameter should have been made earlier");
+
+            // Uri.GetComponents may return a single forward slash when there is no absolute path.  
+            // This is Whidbey PS399695.  Until that is changed, we check for both cases - either an entirely empty string,
+            // or a single forward slash character.  Either case means there is no part name.
+            if (partName.Length == 0 || ((partName.Length == 1) && (partName[0] == '/')))
+                return true;
+            else
+                return false;
```
stephentoub: Why not just:
```C#
return the_condition;
```
rather than:
```C#
if (the_condition)
    return true;
else
    return false;
```
?

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+        // !!Important!! - The order is important - The '%' sign should be escaped first.
+        // If any more characters need to be added to the array below they should be added at the end.
+        private static readonly char[] s_specialCharacters = { '%', '@', ',', '?' };
+
+        //Rels segment and extension
+        private static readonly string s_relationshipPartSegmentName = "_rels";
+        private static readonly string s_relationshipPartExtensionName = ".rels";
+
+        // Forward Slash
+        internal static readonly char ForwardSlashChar = '/';
```
stephentoub: There are various places thus far in the implementation that use '/' hardcoded... should they instead be using ForwardSlashChar?

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+            //------------------------------------------------------
+            //
+            //  Internal Constructors
+            //
+            //------------------------------------------------------
+
+            #region Internal Constructors
+
+            internal ValidatedPartUri(string partUriString)
+                : this(partUriString, false /*isNormalized*/, true /*computeIsRelationship*/, false /*dummy value as we will compute it later*/)
```
stephentoub: Rather than using comments, these could use named arguments.

src/System.IO.Packaging/src/System/IO/Packaging/PackUriHelper.cs
```diff
@@ -0,0 +1,1306 @@
...
+            private static readonly string s_relsrelsUpperCaseExtension = String.Concat(s_relationshipPartUpperCaseExtension, s_relationshipPartUpperCaseExtension);
+
+            //need to use the private constructor to initialize this particular partUri as we need this in the 
+            //IsRelationshipPartUri, that is called from the constructor.
+            private static readonly Uri s_containerRelationshipNormalizedPartUri = new ValidatedPartUri("/_RELS/.RELS",
+                                                                                                         true /*isnormalized*/,
+                                                                                                         false /*computeIsRelationship*/,
+                                                                                                         true /*IsRelationship*/);
+
+            private static readonly char[] s_forwardSlashSeparator = { '/' };
```
stephentoub: I think I saw at least one other place where the same thing was declared.  Should they be consolidated?

src/System.IO.Packaging/src/System/IO/Packaging/Package.cs
```diff
@@ -0,0 +1,1255 @@
...
+                {
+                    if (package != null)
+                    {
+                        package.Close();
+                    }
+
+                    throw;
+                }
+            }
+            finally
```
stephentoub: What's the purpose of the try block with an empty finally?

src/System.IO.Packaging/src/System/IO/Packaging/PackagePart.cs
```diff
@@ -0,0 +1,877 @@
...
+        //return false. These properties do not throw ObjectDisposedException.
+        //So we rely on the values of these properties to determine if a stream 
+        //has been closed.
+        private bool IsStreamClosed(Stream s)
+        {
+            if (s.CanRead == false && s.CanSeek == false && s.CanWrite == false)
+                return true;
+            else
+                return false;
+        }
```
stephentoub: This could just be:
```C#
return !s.CanRead && !s.CanSeek && !s.CanWrite;
```

stephentoub:
Thanks, Eric.  I left some additional comments.  As they're mostly minor issues, I'd be ok with us merging this and then addressing such issues subsequently; up to you.  I do think some/most of the commits should be squashed down before it's merged, e.g. the initial commit from dotnet-bot and then one or a few commits from you addressing feedback.  We try to avoid commits in the history that don't build or pass tests, that are just undoing changes from previous commits in the same PR, etc.

src/System.IO.Packaging/src/System/IO/Packaging/Package.cs
```diff
@@ -0,0 +1,1255 @@
...
+            // These will get flushed to disk as a part of the DoFlush operation
+            FlushRelationships(); // Flush into .rels part.
+
+            //Write out the Relationship XML for the parts
+            //These streams will get flushed in the DoFlush operation.
+            DoOperationOnEachPart(DoWriteRelationshipsXml);
+
+            // Flush all the parts that are currently open.
+            // This will flush part relationships.
+            DoOperationOnEachPart(DoFlush);
```
Maxwe11: Delegate allocation. What about cached lambda `() => DoFlush()` ?

stephentoub: DoFlush is an instance method; the delegate allocation can't be avoided with the current signatures, so addressing that would be a larger change.  Using ```() => DoFlush()``` would actually be worse, as it'd force a closure allocation as well in order to capture the implicit ```this``` reference.

stephentoub:
One more thing: I just noticed the test files in the PR (GitHub was hiding them from view).  It appears you've Base64-encoded various test assets (e.g. .docx files) and included them in a .cs file for use by the tests.  That might be the right approach, but it's not how we've handled test assets elsewhere, e.g. in System.IO.Compression, where test data has been provided in a separate NuGet package:
https://github.com/dotnet/corefx/blob/master/src/System.IO.Compression/tests/project.json#L18 . Please sync with @FiveTimesTheFun to decide which approach we want to follow here.  I'm ok with either as long as you guys agree.  Thanks!

src/System.IO.Packaging/src/System/IO/Packaging/Package.cs
Maxwe11: yeap, I was wrong initially but we can
`DoOperationOnEachPart((thisRef, packagePart) => thisRef.DoFlush(packagePart));`
and pass `this` as parameter inside `DoOperationOnEachPart`.
All these changes related to private/internal API.

EricWhiteDev:
@stephentoub Thanks for doing this detailed review.  Good points all.  To a large extent, these code issues are in code that I inherited from the original System.IO.Packaging.  In general, I opted to make as few changes as possible to existing code, given that this code is used in mission critical applications where stability of the application that uses this library is of the highest importance.

Regarding performance issues, the majority of processing time occurs in System.IO.Compression, and no matter how much we optimize the code in System.IO.Packaging, we can't impact the performance of System.IO.Compression.  In the process of developing the new System.IO.Packaging, I did some measurements on performance, and our biggest issues are in the Open XML SDK proper, not System.IO.Packaging, so it seems good to address those first.  In any case, I have metrics that show that the new System.IO.Packaging has better performance than the old one (mainly due to the better perf of System.IO.Compression when compared to MS.Internal.IO.Compression).  Performance in this area is not a concern of the users of the library.

I would like to go with your suggested plan of merging now, and then addressing code issues such as these in a subsequent phase of work.  The code in its current state is the code that I "hardened" by testing on 370,000 Open XML files, and making even simple changes requires that I again run the code through the my massive test.  This test is run using OxRunner (https://www.youtube.com/watch?v=AcZFexDxL0U), which requires setting up 8 quad-core machines, all properly updated, with correct software installed, git repos properly installed on all machines, and so on.  It is not possible to test efficiently on such a massive document library without taking an approach that uses parallelism.

Further, I have a super-high priority project of making the Open XML SDK ready for Office 2016 (time is ticking on this one), so I propose that I address these issues after I make headway on the Open XML SDK.

stephentoub:
> I would like to go with your suggested plan of merging now, and then addressing code issues such as these in a subsequent phase of work.

@EricWhiteDev: that's fine, thanks.  Please do just verify about the test assets with @FiveTimesTheFun, and squash as mentioned.

stephentoub:
@EricWhiteDev, just checking in on this.  I think this is good to go except for the test assets and then squashing.  Any ETA for when you plan to update this accordingly?

stephentoub:
@EricWhiteDev, @FiveTimesTheFun, how are we doing on this?

EricWhiteDev:
@stephentoub We are meeting in a couple of hours so that I can understand the details of how to put together the test assets.  I'll put together the changes tomorrow.  Should be good to go in short order.

stephentoub:
Great!  Thanks.

EricWhiteDev:
We rescheduled our meeting until Thursday, so I expect to have this in good shape early next week.

stephentoub:
@EricWhiteDev, checking in again :smile: Any updates?

EricWhiteDev:
@stephentoub Yes!  Matt and I got the test assets moved to the corefx-testdata repo, and I have most of the tests working using those test assets.  I have a small amount of cleanup left, and then will push latest commit, and in theory, should be good to go.  I'll ping you hopefully before the end of week, you can take a final look.  I have been buried last couple of days on another project, but should have time to finalize this today or tomorrow.

stephentoub:
Thanks, Eric :smile:

EricWhiteDev:
@stephentoub Ok, it is ready once again for you to look at.  All tests now pull from the corefx-testdata repo.  TestFileLib.cs (which had the base64 encoded test assets) has been deleted.  Further, some tests were leaving behind temporary files.  Now all tests clean up after themselves.  The temporary files were named with Guids, so no chance of conflicting with other tests, but in any case, was messy and wrong to leave them around.

So take a look, and let me know if there is anything else I need to do.

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+        /// <exception cref="ArgumentNullException">If the contentType parameter is null</exception>
+        /// <exception cref="ArgumentException">If the contentType string has leading or 
+        /// trailing Linear White Spaces(LWS) characters</exception>
+        /// <exception cref="ArgumentException">If the contentType string invalid CR-LF characters</exception>
+        internal ContentType(string contentType)
+        {
+            if (contentType == null)
+                throw new ArgumentNullException("contentType");
+
+            if (String.CompareOrdinal(contentType, String.Empty) == 0)
```
stephentoub: ```if (contentType.Length == 0)``` ?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
stephentoub: What was the verdict on this?  Does there need to be a check on the return value of MoveNext?

src/System.IO.Packaging/src/System/IO/Packaging/ContentType.cs
```diff
@@ -0,0 +1,727 @@
...
+                    if (contentTypeEnumerator.Current.Key != null)
+                        return false;
+                }
+
+                // Perform a case-insensitive comparison on the type/subtype strings.  This is a
+                // safe comparison because the _type and _subType strings have been restricted to
+                // ASCII characters, digits, and a small set of symbols.  This is not a safe comparison
+                // for the broader set of strings that have not been restricted in the same way.
+                result = (String.Compare(_type, contentType.TypeComponent, StringComparison.OrdinalIgnoreCase) == 0 &&
+                          String.Compare(_subType, contentType.SubTypeComponent, StringComparison.OrdinalIgnoreCase) == 0);
```
stephentoub: These could use String.Equals instead of String.Compare, right?  I know there are a variety of such cleanup things you're hoping to defer until after the initial PR; this can be on that list (which would ideally be opened as an issue so that we can track it).

stephentoub:
Thanks, Eric.  I took a quick skim through and left a few more minor comments.  At this point, here's what I think you should do:
- There are a bunch of issues not addressed in the code.  That's ok for this PR, but I want to make sure it doesn't get lost.  Please open an issue to track fixing up those and other issues.
- Please squash this down to a smaller number of commits.  In generally we try to keep history clean, which means no merge commits as part of a PR's commit history, commits that compile and pass all tests at the time, etc.  I think it'd make sense to have two (maybe three) commits here: one for the initial dotnet-bot commit of the source, and then one for the subsequent changes you made.

Once that's in, I can do a final pass through, and we can get this merged.

Thanks!

EricWhiteDev:
@stephentoub I have consolidated the code cleanup into a single issue:

https://github.com/dotnet/corefx/issues/2699

EricWhiteDev:
@stephentoub Sorry, I am out of my depth here with regards to squashing commits in this scenario.  (I am able to do this with my own super-simple repos, but worry about doing this one.)  Don't want to mess it up, so I will get with @FiveTimesTheFun to do the squashing.  I will schedule a short meeting with him.

stephentoub:
Thanks, @EricWhiteDev.  I can help;  I'll squash things down now, make sure it all builds/tests appropriately, and put up a replacement PR that contains your commits.

EricWhiteDev:
@stephentoub thank you, appreciate it.  As far as I am concerned, don't need any of the commit history.

stephentoub:
Replaced by #2701

PR #1823, opened 2015-05-22
Fix build script for Mac.

JeremyKuhne:
Installing latest Mono (4.0) then building latest github you can build successfully on Mac.

@weshaggard, @akoeplinger, @ellismg

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,7 +8,16 @@ __nugetconfig=$__sourceroot/NuGet.Config
 __msbuildpackageid="Microsoft.Build.Mono.Debug"
 __msbuildpackageversion="14.1.0.0-prerelease"
 __msbuildpath=$__packageroot/$__msbuildpackageid.$__msbuildpackageversion/lib/MSBuild.exe
-__referenceassemblyroot=/usr/lib/mono/xbuild-frameworks
+
+if [ $(uname) == "Linux" ]; then
+    __monoroot=/usr
+else
+    __monoroot=/Library/Frameworks/Mono.framework/Versions/Current
+fi
+
+__referenceassemblyroot=$__monoroot/lib/mono/xbuild-frameworks
+
+
 __monoversion=$(mono --version | grep "version 4.[1-9]")
 
 if [ $? -ne 0 ]; then


build.sh
```diff
@@ -8,7 +8,16 @@ __nugetconfig=$__sourceroot/NuGet.Config
 __msbuildpackageid="Microsoft.Build.Mono.Debug"
 __msbuildpackageversion="14.1.0.0-prerelease"
 __msbuildpath=$__packageroot/$__msbuildpackageid.$__msbuildpackageversion/lib/MSBuild.exe
-__referenceassemblyroot=/usr/lib/mono/xbuild-frameworks
+
+if [ $(uname) == "Linux" ]; then
+    __monoroot=/usr
+else
+    __monoroot=/Library/Frameworks/Mono.framework/Versions/4.0.0
```
akoeplinger: Use `Current` instead of the explicit 4.0.0 to make it future proof.

akoeplinger:
LGTM apart from the one comment. With which version did you test on Mac? I thought it'd run into the same Roslyn issue we're running into with Mono 4.0 on Linux (that's why we're checking for Mono >= 4.1)?

build.sh
JeremyKuhne: It wasn't in the Current directory after building and installing per http://www.mono-project.com/docs/compiling-mono/mac/. (e.g. 'make' 'make install')

JeremyKuhne:
@akoeplinger 4.1, built from latest source. I didn't see any other way to get 4.1.

build.sh
akoeplinger: If you followed that guide then you installed Mono into `/usr/local`. The PCL assemblies are still in the folder of the regular package, which should have a Current folder.

akoeplinger:
@JeremyKuhne makes sense. Btw. I've seen a commit in the mono-4.0 branch that fixes the issue, so as soon as this hits a 4.0 service release it'll be a moot point anyway :)

build.sh
JeremyKuhne: Sorry, when I was searching for it I didn't see the portable folder - love .Net being hidden. :)

ellismg:
LGTM

PR #1826, opened 2015-05-22
System.Text.Encoding: Add missing dnxcore50 frameworks element to project.json

akoeplinger:
@ericstj can I just commit the project.lock.json as well? I noticed there are a couple other lockfiles that aren't checked in.


mellinoe:
Yes, I think we should check in all of the lock files that are missing now,

akoeplinger:
Ok, I think this should be done as a separate PR then (partly because I'm not sure I got the correct versions right now).

@dotnet-bot test this please

Maxwe11:
@dotnet-bot test this please

ericstj:
LGTM.  For lock files please check them in.  I'm ok with it being a separate commit.  If the versions showing up don't match the rest of the repo simply find/replace.

stephentoub:
LGTM

PR #1832, opened 2015-05-23
Use immutable collections 1.1.36 in metadata

tmat:


stephentoub:
This is failing to compile with the error:
```
d:\j\workspace\dotnet_corefx_windows_debug_prtest\bin\Windows_NT.AnyCPU.Debug\System.Reflection.Metadata\System.Reflection.Metadata.dll : error CS1705: Assembly 'System.Reflection.Metadata, Version=1.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' uses 'System.Collections.Immutable, Version=1.1.36.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Collections.Immutable, Version=1.1.34.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

jasonwilliams200OK:
@tmat if you rebase, you will find out [project.lock.json](https://github.com/dotnet/corefx/blob/e9b144e6882b972fbca043f5f5b156ff45813c5f/src/System.Reflection.Metadata/src/project.lock.json) in metadata src root, which probably needs to be updated as well.

tmat:
@jasonwilliams200OK  What is project.lock.json and why is it in the repo?

jasonwilliams200OK:
@tmat, it was added as part of changes merged from TFS through https://github.com/dotnet/corefx/pull/1809. The commit message of https://github.com/dotnet-bot/corefx/commit/d2fbc005707933db1cebca035d6cd6073af6cae4 reads:

> Lock dependencies to ensure we build against a predictable package version
>
This change adds the lock files to repo to ensure we build against a stable set of dependencies.  To explicitly update dependencies use build.cmd /T:RestorePackages /P:LockDependencies=true and this will pick up the latest depenedencies.
>
Existing projects adding new dependencies will automatically update the lock file incrementally.
>
New projects should make sure to check their lock file in with locked set to true.
>
[tfs-changeset: 1474296]

tmat:
@ericstj Could you point me to the reasoning behind .lock.json files? I don't understand why specifying the version in packages.json file is not "predicable" enough. Now the process of updating a dependency version is much more complicated than it used to be.

ericstj:
It's actually much simpler.   We switched to '*' versioning for most dependencies.  This allows us to "float" versions to the latest build.  The lock file ensures that the floating version is still predictable at a given sync point.  To update the lock (and thus, all the floating versions across the repo), just run /t:RestorePackages /p:LockDependencies=true.
For further discussion of the design of project.json and project.lock.json please take that to https://github.com/aspnet/dnx.

PR #1834, opened 2015-05-23
Improve performance of HtmlEncoder.TryWriteEncodedScalarAsNumericEntity

bbowyersmyth:
Change TryWriteEncodedScalarAsNumericEntity to do the shifting twice and use the provided buffer instead of the char[8] creation and indexed access.

~8% improvement within an entire Encoder.Encode() call depending on string provided. Encode has many branches so to make sure this function was singled out the following 2 strings were used for performance comparison.

    "+++++++"
    "\u2601\u2601\u2601\u2601\u2601\u2601\u2601"

In addition I fixed some typos in the tests and corrected one missed refactoring that resulted in compilation fail in a conditional RELEASE section.

### DIFF
--- a/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/HtmlEncoder.cs
+++ b/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/HtmlEncoder.cs
@@ -102,16 +102,18 @@ namespace System.Text.Encodings.Web
         {
             Debug.Assert(buffer != null && bufferLength >= 0);
 
-            // We're building the characters up in reverse
-            char* chars = stackalloc char[8 /* "FFFFFFFF" */];
+            // We're writing the characters in reverse, first determine
+            // how many there are
+            const int nibbleSize = 4;
             int numberOfHexCharacters = 0;
+            int compareUnicodeScalar = unicodeScalar;
+
             do
             {
                 Debug.Assert(numberOfHexCharacters < 8, "Couldn't have written 8 characters out by this point.");
-                // Pop off the last nibble
-                chars[numberOfHexCharacters++] = HexUtil.Int32LsbToHexDigit(unicodeScalar & 0xF);
-                unicodeScalar >>= 4;
-            } while (unicodeScalar != 0);
+                numberOfHexCharacters++;
+                compareUnicodeScalar >>= nibbleSize;
+            } while (compareUnicodeScalar != 0);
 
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
Looking at the code, I like the change. Just a question: how did you measure the 8% improvement? Did you use the checked in perf tests? If yes, could you copy-paste the output from the tests? If you don't know it's in .\corefx\bin\tests\Windows_NT.AnyCPU.Release\System.Text.Encodings.Web.Tests\dnxcore50\testResults.xml; provide just the CDATA sections with the perf mesaurements.
Thanks!

bbowyersmyth:
The existing perf tests do not hit this function as the angle brackets have a special case in `TryEncodeUnicodeScalar`.

To single `TryWriteEncodedScalarAsNumericEntity` out the following test code was used.

        {
            var encoder = System.Text.Encodings.Web.HtmlEncoder.Default;

            Profile("Run 1", 10000000, () =>
            {
                WriteEncodedScalarAsNumericEntity1(encoder, (uint)15);
            });

            Profile("Run 2", 10000000, () =>
            {
                WriteEncodedScalarAsNumericEntity2(encoder, (uint)15);
            });

            Console.Read();
        }

        private static void WriteEncodedScalarAsNumericEntity1(System.Text.Encodings.Web.TextEncoder encoder, uint value)
        {
            string result = encoder.Encode("+++++++");
        }

        private static void WriteEncodedScalarAsNumericEntity2(System.Text.Encodings.Web.TextEncoder encoder, uint value)
        {
            string result = encoder.Encode("\u2601\u2601\u2601\u2601\u2601\u2601\u2601");
        }

        static void Profile(string description, int iterations, Action func)
        {
            // warm up 
            func();
            // clean up
            GC.Collect();

            var watch = new Stopwatch();
            watch.Start();
            for (int i = 0; i < iterations; i++)
            {
                func();
            }

            watch.Stop();
            Console.Write(description);
            Console.WriteLine(" Time Elapsed {0} ms", watch.ElapsedMilliseconds);
        }

That is in a traditional console app, platform target = x64, project reference to System.Text.Encodings.Web, Release build.

On an old PC, these are the timings when run from the command line:

Test | Before | After
------ | -------- | -----
Run 1 | 4199 | 3836
Run 2 | 4710 | 4358

KrzysztofCwalina:
Ah, I see. Thanks! 
The change looks good to me. Thanks for the contribution. It's a nice small improvement.

jasonwilliams200OK:
:+1:

stephentoub:
LGTM

Issue #1805, opened 2015-05-20
Section name in PEReader contains null padding


krwq:
String in section name (i.e. peReader.PEHeaders.SectionHeaders[0].Name) contains null characters.

I believe we should remove it, unless it is by design.

ECMA description of field:
```
An 8-byte, null-padded ASCII string. There is no terminating null if the string is exactly eight characters long.
```

Here is how we get the value:
```csharp
// reader is PEBinaryReader
// _name is string
// PEFileConstants.SizeofSectionName = 8
_name = reader.ReadUTF8(PEFileConstants.SizeofSectionName);

// impl of ReadUTF8:
public string ReadUTF8(int byteCount)
{
    byte[] bytes = ReadBytes(byteCount);
    return Encoding.UTF8.GetString(bytes, 0, byteCount);
}
```

In theory it can also be encoding bug.

ellismg:
Given the spec, it also seems a little odd that we are using a UTF8 decoder here.

mikedn:
> Given the spec, it also seems a little odd that we are using a UTF8 decoder here.

Actually it's the ECMA spec that it is odd :smile:. The latest PECOFF spec says

> An 8-byte, null-padded UTF-8 encoded string...

A quick test shows that the VC++ compiler can indeed emit UTF-8 encoded section names.

jbevain:
Wouldn't be the first time that the C++ compiler emitted something not conforming with ECMA-335 :)

nguerrera:
Fixed by #1837

PR #1837, opened 2015-05-25
Remove null-padding from parsed section-header names.

benjamin-bader:
Fixes issue #1805.

As I understand the issue, the ECMA spec is outdated, and section-header names are 8-byte null-padded UTF-8 strings.  While UTF-8 strings can legally hold null bytes (the NUL codepoint), the nulls here are padding and not properly part of the header name.

This diff truncates the null padding and adds some test coverage for PEBinaryReader and SectionHeader.

As this is my first contribution, I'm new to the codebase and have probably missed something important (besides CLA); thanks in advance for letting me know what those important things are!

### DIFF
--- a/src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
+++ b/src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
--- a/src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
+++ b/src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
@@ -19,6 +19,10 @@ namespace System.Reflection.Metadata.Tests.PortableExecutable
         [Theory]
         [InlineData(".debug", 0, 0, 0x5C, 0x152, 0, 0, 0, 0, SectionCharacteristics.LinkerInfo)]
         [InlineData(".drectve", 0, 0, 26, 0x12C, 0, 0, 0, 0, SectionCharacteristics.Align1Bytes)]
+        [InlineData("", 1, 1, 2, 3, 5, 8, 13, 21, SectionCharacteristics.Align16Bytes)]
+        [InlineData("x", 1, 1, 2, 3, 5, 8, 13, 21, SectionCharacteristics.MemSysheap)]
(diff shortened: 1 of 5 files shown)


benjamin-bader:
Hm, Jenkins doesn't like the test data - will fix later today!

nguerrera:
cc @tmat. Thoughts? It looks good to me. It's technically a behavioral breaking change (and we just shipped a stable revision). I looked at all internal usage of SectionHeader.Name and they all involve eight-character sections (.cormeta, .rsrc$01, .rsrc$02) :smiley:

tmat:
:+1: I think it's ok. If someone finds NUL at the end of a section name with the current reader they'll trim it as a workaround. If they update to the new reader and they will not find NUL anymore the workaround won't kick in. Of course, it is possible they do it poorly and later get broken. Meh.

benjamin-bader:
@tmat Changes are in, tests are passing.  Thanks very much for the review - please let me know what else might be needed.

src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace System.Reflection.Metadata.Tests.PortableExecutable
+{
+    public class SectionHeaderTests
+    {
+        [Theory]
+        [InlineData(".debug", 0, 0, 0x5C, 0x152, 0, 0, 0, 0, SectionCharacteristics.LinkerInfo)]
+        [InlineData(".drectve", 0, 0, 26, 0x12C, 0, 0, 0, 0, SectionCharacteristics.Align1Bytes)]
+        [InlineData("", 1, 1, 2, 3, 5, 8, 13, 21, SectionCharacteristics.Align16Bytes)]
+        [InlineData("x", 1, 1, 2, 3, 5, 8, 13, 21, SectionCharacteristics.MemSysheap)]
+        [InlineData(".बग", int.MaxValue, int.MinValue, int.MaxValue, int.MinValue, int.MaxValue, int.MaxValue, ushort.MaxValue, ushort.MaxValue, SectionCharacteristics.GPRel)]
```
nguerrera: Please use the \uNNNN equivalent here. We have been trying to standardize on that.

nguerrera:
Looks good to me apart from the small comment about using \uNNNN instead of actual non-ASCII characters in the source code.

Thanks for the contribution! I'll wait for that fix and @tmat's sign-off before merging.

src/System.Reflection.Metadata/tests/PortableExecutable/SectionHeaderTests.cs
benjamin-bader: Will do!

benjamin-bader:
Syntax is fixed and the build is green.  Thanks again :)

tmat:
:+1:

nguerrera:
Thanks, @benjamin-bader!

PR #1087, opened 2015-03-04
update to latest rc of OpenCover to address closedown issue encountered ...

sawilde:
...with System.Diagnostics.Process tests #1069

PR #1842, opened 2015-05-26
Fix AltDirectorySeparatorChar on Unix

BrennanConroy:
https://github.com/dotnet/coreclr/issues/1026

### DIFF
--- a/src/System.Runtime.Extensions/src/System/IO/Path.cs
+++ b/src/System.Runtime.Extensions/src/System/IO/Path.cs
@@ -38,7 +38,7 @@ namespace System.IO
     public static class Path
     {
         // Platform specific directory separator character.  This is backslash
-        // ('\') on Windows, slash ('/') on Unix, and colon (':') on Mac.
+        // ('\') on Windows and slash ('/') on Unix.
         // 
 #if !PLATFORM_UNIX        
         public static readonly char DirectorySeparatorChar = '\\';
@@ -49,14 +49,10 @@ namespace System.IO
 #endif // !PLATFORM_UNIX
 
         // Platform specific alternate directory separator character.  
-        // This is backslash ('\') on Unix, and slash ('/') on Windows 
-        // and MacOS.
+        // This is slash ('/') on Windows
+        // There is no alternate separator on Unix, so the default separator is used.
         // 
-#if !PLATFORM_UNIX        
         public static readonly char AltDirectorySeparatorChar = '/';
-#else
-        public static readonly char AltDirectorySeparatorChar = '\\';
-#endif // !PLATFORM_UNIX
 
         // Platform specific volume separator character.  This is colon (':')
         // on Windows and MacOS, and slash ('/') on Unix.  This is mostly


src/System.Runtime.Extensions/src/System/IO/Path.cs
```diff
@@ -49,13 +49,13 @@ public static class Path
 #endif // !PLATFORM_UNIX
 
         // Platform specific alternate directory separator character.  
-        // This is backslash ('\') on Unix, and slash ('/') on Windows 
-        // and MacOS.
+        // This is slash ('/') on Windows and MacOS.
```
kangaroo: This comment is out of date, and incorrect.  Probably related to the ':' on Mac above.  It should be cleaned up at the same time.

src/System.Runtime.Extensions/src/System/IO/Path.cs
```diff
@@ -49,13 +49,13 @@ public static class Path
...
+        // There is no alternate separator on Unix, so the normal separator is used.
         // 
 #if !PLATFORM_UNIX        
         public static readonly char AltDirectorySeparatorChar = '/';
 #else
-        public static readonly char AltDirectorySeparatorChar = '\\';
+        public static readonly char AltDirectorySeparatorChar = '/';
 #endif // !PLATFORM_UNIX
```
kangaroo: This now looks odd.  If we're !UNIX its '/' else its '/'?  Can't we just remove this and clean up the file?

jasonwilliams200OK: Given the comment above, is it necessary to have preprocessor directive at all? 
IMO, this can be trimmed down to:

```c#
// This is backslash ('/') on Windows and MacOS.
// There is no alternate separator on Unix, so the default separator ('/') is used.
//
public static readonly char AltDirectorySeparatorChar = '/';
```

BrennanConroy: Good point

ellismg:
Can you please squash? /cc @stephentoub

BrennanConroy:
Done

kangaroo:
My comment was removed when you force pushed.  Could you please fix the comment above about ':' being the primary sep on Mac.  Thats anciently out of date.

VSadov:
Interestingly this change it matches assumptions made in Roslyn compilers, who treat AltDirectorySeparatorChar to be '/' unconditionally.

https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/FileSystem/PathUtilities.cs

kangaroo:
@dotnet-bot test this please

stephentoub:
Note that this Path.cs file isn't currently being used when running on CoreCLR, so this PR doesn't actually fix anything at present... this file isn't being compiled into anything used on CoreCLR, and rather is only used presently by the .NET Native build of System.Runtime.Extensions.  System.Runtime.Extensions is a "partial facade," which means that it contains some types, but type forwards to mscorlib for any of the types in the System.Runtime.Extensions contract that aren't included in the project, and Path is one of those (note that Path.cs isn't listed in the .csproj).  As such, the Path from mscorlib in the coreclr repo is being used, both on Windows and on Unix.  Eventually we would like to switch over to using this Path.cs in corefx, but that requires more work, as when we do so we'd want to reimplement all of the P/Invokes it relies on to target Unix functionality directly rather than going through libcoreclr's Win32-like PAL (there are also some changes that have been made to the Path.cs in mscorlib that need to be ported here as well).

If/when we make this change to the Path.cs in mscorlib, we'll also need to be cognizant of the corefx tests.  I think it's possible some tests will start breaking (due to test issues) once Path is updated with this change, and since the CI system in corefx currently picks up the latest successful build of mscorlib from the coreclr build, as soon as a change like this gets checked in to coreclr, any tests in corefx that depended on this functionality will immediately start failing in all corefx builds / PRs / etc.  As such, before checking in such a change to coreclr, any such tests (if there are any) would need to be fixed up appropriately in corefx.

stephentoub:
I submitted a similar fix to coreclr: https://github.com/dotnet/coreclr/pull/1087

Thankfully only one corefx test starts to fail with this, and I've disabled it with: https://github.com/dotnet/corefx/pull/1905

stephentoub:
@BrennanConroy, now that #1087 has been merged, I'm going to go ahead and close this PR.  When we're ready to switch to using the Path.cs in corefx rather than in coreclr, we'll be sure to bring this change over along with a bunch of others.  Thanks!

PR #1843, opened 2015-05-26
Workaround for perf issue with the new jit pointer dereference in serialization

khdang:
With the new jit, the *chars deference is done twice which is more costly compared to legacy jit


mconnew:
:shipit:

stephentoub:
LGTM

PR #1845, opened 2015-05-27
Add missing project.lock.json files

akoeplinger:


akoeplinger:
/cc @ericstj I can't seem to get the versions in the lockfile to change, even with running `build /t:RestorePackages /p:LockDependencies=true` like you advised in https://github.com/dotnet/corefx/pull/1832#issuecomment-105604672.

The versions in the repo right now are also not consistent, e.g. there's version [22816](https://github.com/dotnet/corefx/blob/e9b144e6882b972fbca043f5f5b156ff45813c5f/src/System.Security.Cryptography.Encryption/src/project.lock.json#L6) and [22914](https://github.com/dotnet/corefx/blob/e9b144e6882b972fbca043f5f5b156ff45813c5f/src/System.Security.Cryptography.Hashing.Algorithms/src/project.lock.json#L6) of System.Diagnostics.Contracts in lockfiles.

ericstj:
Something here is fishy.  I wonder why DNX isn't choosing the highest version in the feed that matches the *.  I can confirm that I see the same behavior, even when deleting all the lock.json's.

If I look at the feeds:
https://www.myget.org/F/dotnet-core/ -> has 22921
https://www.nuget.org/api/v2/ -> has 22816

/cc @anurse @davidfowl @lodejard

ericstj:
Looks like there are a few new project.jsons with explicit versions that are muddying the waters.  Let me clean those up and systematically fix this.

ericstj:
Ok,the old dependencies are coming from test projects which are bringing in the dependencies via xunit.netcore.extensions.  That can be fixed as a separate issue.

I think I'd like to fix this more systematically to ensure our infrastructure is working OK.  Do you mind if I close this PR?

akoeplinger:
Thanks for looking into it :)

PR #752, opened 2015-02-12
Call Dispose in Finalizer; remove unnecessary Null check

jbe2277:
Please compare with the original implementation which behaves identical to my changes: http://referencesource.microsoft.com/#System/compmod/system/componentmodel/BackgroundWorker.cs,85d60b0d93a826fa

PR #1858, opened 2015-05-28
Improve System.CM.EventBasedAsync tests

### LINKED ISSUE #922: Improve code coverage for System.ComponentModel.EventBasedAsync
According to the current code coverage report, this library is only at 74.5%.  Our general goal is closer to 90%, though this will depend on the library.  We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

Maxwe11:
Closes #922

### DIFF
--- a/src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
+++ b/src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
@@ -139,7 +139,7 @@ namespace System.ComponentModel.EventBasedAsync.Tests
         [Fact]
         public void TestThrowExceptionInDoWork()
         {
-            var orignal = SynchronizationContext.Current;
+            var original = SynchronizationContext.Current;
             try
             {
                 SynchronizationContext.SetSynchronizationContext(null);
@@ -170,11 +170,11 @@ namespace System.ComponentModel.EventBasedAsync.Tests
                 };
 
                 bw.RunWorkerAsync(expectedArgument);
-                Assert.True(workerCompletedEvent.Wait(TimeoutShort), "Background work timeout");
+                Assert.True(workerCompletedEvent.Wait(TimeoutLong), "Background work timeout");
             }
             finally
             {
-                SynchronizationContext.SetSynchronizationContext(orignal);
+                SynchronizationContext.SetSynchronizationContext(original);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -214,6 +214,42 @@ public void RunWorkerAsyncTwice()
         }
 
         [Fact]
+        public void TestCancelInsideDoWork()
+        {
+            var orignal = SynchronizationContext.Current;
```
stephentoub: Did you intend to then do SynchronizationContext.SetSynchronizationContext(null)?  Otherwise, why are we capturing and restoring the context if we're not changing it?

Nit: orignal => original

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -214,6 +214,42 @@ public void RunWorkerAsyncTwice()
...
+                bw.RunWorkerCompleted += (sender, e) =>
+                {
+                    Assert.True(e.Cancelled);
+                };
+
+                bw.RunWorkerAsync();
+                barrier.SignalAndWait();
+                bw.CancelAsync();
+                barrier.SignalAndWait();
+            }
```
stephentoub: Shouldn't we wait for the background operation to complete here?  Otherwise we may not actually execute the Assert.True(e.Cancelled), if the tests were to end prior to that executing.

stephentoub:
Thanks for doing this.  Two comments, but otherwise LGTM.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
Maxwe11: Yeah, accidentally missed it)

Maxwe11:
@stephentoub updated.
Could you also take a look at code coverage report for `BackgroundWorker`. With this test addition only one line still uncovered inside `WorkerThreadStart` method. I think it's a potential dead code and we should remove that line or replace it with `Debug.Assert`.
`BackgroundWorker` [at referencesource.microsoft.com](http://referencesource.microsoft.com/#System/compmod/system/componentmodel/BackgroundWorker.cs,236) also doesn't have that line.

src/System.ComponentModel.EventBasedAsync/tests/BackgroundWorkerTests.cs
```diff
@@ -214,6 +214,46 @@ public void RunWorkerAsyncTwice()
...
+                {
+                    Assert.True(e.Cancelled);
+                    barrier.SignalAndWait();
+                };
+
+                bw.RunWorkerAsync();
+                barrier.SignalAndWait();
+                bw.CancelAsync();
+                barrier.SignalAndWait();
+                Assert.True(barrier.SignalAndWait(TimeoutShort), "Background work timeout");
```
stephentoub: Should this be TimeoutLong rather than TimeoutShort?  I believe TimeoutLong is typically used where we expect the work to always complete and we require waiting until it's done, but we want to bail if it takes so long that there must be an issue... by using TimeoutShort here, I think it's likely now and then we'd hit a spurious failure.

stephentoub:
> With this test addition only one line still uncovered inside WorkerThreadStart method (line number 209). I think it's a potential dead code and we should remove that line

Yes, PR #752 actually does that, but it hasn't been merged yet (for a long time) due to debate about the other part of that PR.  Seems like we should just go ahead and make that part of the change here, and then #752 can be updated to just be about the dispose/finalization part.

Maxwe11:
updated

stephentoub:
LGTM

PR #1868, opened 2015-05-29
Add link to example report

richlander:
Providing an example portability report would help people understand how this tool is going to help them.

/cc @kuhlenh


stephentoub:
LGTM

akoeplinger:
LGTM, though you may want to link to https://github.com/Moq/moq4 in case people aren't familiar with Moq.

Orthogonal to this PR: I saw the report lists `Mono 4.5`. I guess this is meant to show the ".NET 4.5 profile on Mono", but this is bound to get really confusing soon as Mono's version itself (which is independent from .NET) is already at 4.0 and 4.2 is around the corner.
Can we modify the analyzer to show the real Mono version used for the dataset? This will be much more useful to users since "4.5" doesn't really tell anything (Mono 2.10 shipped with that profile as well iirc, just a very limited implementation).

stephentoub:
@richlander, is this good to merge?

richlander:
@akoeplinger Thanks for that feedback. We're going to update ApiPort. We'll apply that feedback for that.

/cc @kuhlenh

PR #1869, opened 2015-05-29
System.Linq.Parallel remove use of 'Shared' in enumerators

Clockwork-Muse:
Several PLINQ classes use a wrapper type called `Shared` for maintaining the enumerator counts, even though this is never exported from the object (much less a thread, as the type purports to be for).  This removes that use in favor of a simple int.

All tests (both the original PLINQ ones, and my expanded ones) passing.

I'm not sure why it's being used in this instance.  `Shared` purports to be for 'sharing a value between threads', but seems to really be used as a substitute for passing a `ref` field.  However, this wasn't necessary for where it was used in the enumerators, and seems misleading besides:

 1. `Shared` itself isn't implemented in a threadsafe manner.
 2. None of the enumerators using it would have been safe if it was.
 3. There was no synchronization between `MoveNext` and `Current`, and doing so is ill-advised/difficult (easier for synchronization the next level up).

This change leaves the use as passing a `ref` field alone, as that's a trickier problem.

<hr /> 
Personally, I'd like to remove use of the type, especially, for the various early-exit `bool` flags; `Shared` doesn't use `volatile` with the wrapped field, so early-exit really becomes "if I eventually see it" (annoying and eats resources, but not fatal per-se).
Where important, PLINQ is using locking/`Interlocked` to deal with the update problem, but this begs the question of why a `ref` field wasn't used in the first place...

### DIFF
--- a/src/System.Linq.Parallel/src/System/Linq/Parallel/Enumerables/RangeEnumerable.cs
+++ b/src/System.Linq.Parallel/src/System/Linq/Parallel/Enumerables/RangeEnumerable.cs
@@ -76,7 +76,7 @@ namespace System.Linq.Parallel
             private readonly int _from; // The initial value.
             private readonly int _count; // How many values to yield.
             private readonly int _initialIndex; // The ordinal index of the first value in the range.
-            private Shared<int> _currentCount; // The 0-based index of the current value. [allocate in moveNext to avoid false-sharing]
+            private int _currentIndex = -1; // The 0-based index of the current value.
 
             //-----------------------------------------------------------------------------------
             // Creates a new enumerator.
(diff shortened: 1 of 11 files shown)


stephentoub:
I'm very hesitant to take a change like this without extensive performance testing on a wide range of scenarios and hardware.  IIRC, one of the reasons ```Shared<T>``` was being used was to keep certain data separate in memory from other data, so as to avoid false sharing issues that were leading to slowdowns.  Some of the comments that were removed as part of this change allude to this.

Clockwork-Muse:
...ah.  I thought the "false sharing" was alluding to sharing with respect to _threading_ which confused me (as in, sharing the values across threads, which as near as I could tell wasn't happening).

Test runs were in the same ballpark as the original version, but those tests aren't really written as perf tests, so I wouldn't trust that much.  Were perf tests for this library open sourced, or just part of some other solution/project?

So the idea was that, what, by moving the instantiation to `MoveNext`, that the value would be located near the data/instructions for the running loop, as opposed to with the readonly values?  Or what's meant by 'sharing'?
This is a surprising and somewhat non-intuitive optimization for me!  Do you have a reference where I could learn more?

stephentoub:
>  the value would be located near the data/instructions for the running loop

It's less about what it would be near and more about what it wouldn't be near, i.e. keeping a memory location that'll be mutated frequently away from values that'll be read frequently by many threads.

> Do you have a reference where I could learn more?

https://msdn.microsoft.com/en-us/magazine/cc872851.aspx

Clockwork-Muse:
You've written some very enlightening stuff, thanks!
I don't know how well I'd be able to do the necessary performance testing, especially because I only have access to this one machine (and the type of testing it requires).  Ah, the pros and cons of not having had to worry about stuff that close to the metal...

(of course, now the purist in me wants to remove it due to the nature of the work-around, but that's not a good idea if there's no better solution available)

stephentoub:
Thanks, anyway :)

PR #1870, opened 2015-05-29
Fix AV in MemoryMappedViewAccessor reading/writing

stephentoub:
On Windows, when a MemoryMappedViewAccessor is created with DelayAllocatePages and a capacity of 0 (meaning to use the whole capacity of the map), VirtualAlloc is incorrectly being called with a size of 0 rather than the actual viewSize.  This leads to an AV when trying to read/write via the view accessor.


stephentoub:
cc: @mellinoe, @ericstj

ericstj:
This looks familiar.  I made the same change to desktop earlier this release and missed flowing to corefx :frowning:.  LGTM

Issue #1841, opened 2015-05-26
TimerQueue.DeleteAppDomainTimer sometimes failing during finalization


stephentoub:
This has been happening periodically while running our tests in CI.
```
  Unhandled Exception: System.Exception: The handle is invalid. (Exception from HRESULT: 0x80070006 (E_HANDLE))
     at System.Threading.TimerQueue.DeleteAppDomainTimer(IntPtr handle)
     at System.Threading.TimerQueue.AppDomainTimerSafeHandle.ReleaseHandle()
     at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
     at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
     at System.Runtime.InteropServices.SafeHandle.Finalize()
```

cc: @ericeil... Eric, have you seen this before?

bpschoch:
I've gotten these while testing pipes.  It maybe related to pipes but don't know how to really debug it when I occasionally get this error.  Some of the pipes testing does use read timeouts and cancellation tokens.

stephentoub:
It's possible this is actually being caused by a handle recycling problem, fixed in #1871.  I'll close this once #1871 is merged in, and we can reopen it subsequently if we see it again.

stephentoub:
I believe #1871 fixes this.  Closing for now; we can reopen if we see this again.

PR #1871, opened 2015-05-29
Fix handle recycling / double-close bug in System.IO.Pipes

stephentoub:
The design of AnonymousServerPipeStream calls for a developer to instantiate the server stream and then grab from it the client handle to use to instantiate an AnonymousClientPipeStream.  The handle can be grabbed in two ways, either via ClientSafePipeHandle (which returns the internal SafePipeHandle) or via GetClientHandleAsString (which returns the handle's value serialized as a string).  In either case, the server marks the client handle as being exposed, which means it won't call Dispose on the handle when the server stream is Disposed or finalized.

However, just because the stream doesn't call Dispose on it doesn't change the fact that the client SafePipeHandle has its own finalizer.  If the consuming code calls GetClientHandleAsString and passes that to an AnonymousClientPipeStream, that AnonymousClientPipeStream will construct a SafePipeHandle from it.  Since the AnonymousClientPipeStream could be in another process (it doesn't know where the string comes from), it marks the SafePipeHandle as ownsHandle=true, which means that disposing of the client stream or letting it or the handle get finalized will close the client handle.  And since the AnonymousServerPipeStream also has a SafePipeHandle that thinks it owns the same handle value in the same process, we now run the risk of double-disposal of the same handle value, and the resulting recycling concerns.

Given the existing design, there's no perfect solution here, but I think the best solution that maps to the expected usage pattern is to simply suppress the finalization of the client handle in the server stream's GetClientHandleAsString.  In the case where the string is passed to another process, the developer is expected to call DisposeLocalCopyOfClientHandle; that's always been the case, this change just makes it that much more important.

The commit also fixes two handle recylcling issues in the System.IO.Pipes tests.  A SafePipeHandle is being constructed around an arbitrary file handle, and since it's being constructed as ownsHandle=true, that SafePipeHandle will get finalized and will close a handle that's already been closed and potentially recycled.  The fix is simply to change ownsHandle to false.

(As an aside, I think this may address issue #1841, but I'm not positive.)

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/AnonymousPipeServerStream.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/AnonymousPipeServerStream.cs
@@ -98,6 +98,7 @@ namespace System.IO.Pipes
         public String GetClientHandleAsString()
         {
             _clientHandleExposed = true;
+            GC.SuppressFinalize(_clientHandle);
             return _clientHandle.DangerousGetHandle().ToString();
         }
 
--- a/src/System.IO.Pipes/tests/NamedPipesThrowsTests.cs
+++ b/src/System.IO.Pipes/tests/NamedPipesThrowsTests.cs
@@ -225,7 +225,7 @@ namespace System.IO.Pipes.Tests
                     safeHandle.DangerousAddRef(ref gotRef);
                     IntPtr handle = safeHandle.DangerousGetHandle();
 
-                    SafePipeHandle fakePipeHandle = new SafePipeHandle(handle, true);
+                    SafePipeHandle fakePipeHandle = new SafePipeHandle(handle, ownsHandle: false);
                     Assert.Throws<IOException>(() => new NamedPipeServerStream(PipeDirection.InOut, false, true, fakePipeHandle));
                 }
                 finally
@@ -326,7 +326,7 @@ namespace System.IO.Pipes.Tests
                     safeHandle.DangerousAddRef(ref gotRef);
                     IntPtr handle = safeHandle.DangerousGetHandle();
 
-                    SafePipeHandle fakePipeHandle = new SafePipeHandle(handle, true);
+                    SafePipeHandle fakePipeHandle = new SafePipeHandle(handle, ownsHandle: false);
                     Assert.Throws<IOException>(() => new NamedPipeClientStream(PipeDirection.InOut, false, true, fakePipeHandle));
                 }
                 finally


stephentoub:
cc: @ericstj, @mellinoe

stephentoub:
@dotnet-bot test this please

ericstj:
What if someone wrote the following code instead of calling DisposeLocalCopyOfClientHandle:
```
using (var server =new AnonymousPipeServerStream(PipeDirection.Out))
{
    using (server.ClientSafePipeHandle)
    {
        StartNewProcess(“exename.exe”, “arg1 arg2 “ + server.GetClientHandleAsString()); // (pseudo method)
        // wait for client to connect.
    }
    // communication
}
```
Would this still work?

stephentoub:
Yes, that'd be fine; that's basically identical to:
```C#
using (var server =new AnonymousPipeServerStream(PipeDirection.Out))
{
    StartNewProcess(“exename.exe”, “arg1 arg2 “ + server.GetClientHandleAsString()); // (pseudo method)
        // wait for client to connect.
    }
    server.DisposeLocalCopyOfClientHandle();
    // communication
}
```

ericstj:
I see.  Are you sure that DisposeLocalCopyOfClientHandle is going to close the handle when ownsHandle = false?  Looking at https://github.com/dotnet/coreclr/blob/master/src/vm/safehandle.cpp#L180 it looks like it won't.

stephentoub:
> Are you sure that DisposeLocalCopyOfClientHandle is going to close the handle when ownsHandle = false?

No, it won't, that's by design (the purpose of ownsHandle is to mean whether disposing of the SafeHandle closes the underlying handle).  But... can you elaborate on your concerns?  If the SafeHandle is ownsHandle=false, then it's already marked as suppressed finalization (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Runtime/InteropServices/SafeHandle.cs#L168), so my change should be a nop for that case.

ericstj:
Ok, I misuderstood.  I thought you were changing the code to actually set ownsHandle to false on the ClientSafePipeHandle.  If I actually looked at the code instead of just reading the text I would have seen that is not the case.  Now I see that you are just suprressing finalization.  

This mostly safe.  The only thing that'll happen is that if someone wasn't calling ClientSafePipeHandle.Dispose or DisposeLocalCopyOfClientHandle before they'll now leak the handle.  If this were a named pipe I'd be concerned since the leak could prevent future opens, but since it is anonymous it shouldn't get in the way of future pipes.  It will be a leak, however, and if someone had some long running process with this pattern it could cause them problems.  I think if we took this code to desktop we might want to quirk it, for corefx I'm ok taking the change.

One last concern: is it possible that a client was detecting the closure of the handle from the server side and making a decision based on this (eg: detecting that the server object was finalized)?

stephentoub:
> The only thing that'll happen is that if someone wasn't calling ClientSafePipeHandle.Dispose or DisposeLocalCopyOfClientHandle before they'll now leak the handle.

Correct.

> is it possible that a client was detecting the closure of the handle from the server side and making a decision based on this (eg: detecting that the server object was finalized)?

I guess it's theoretically possible, but it'd be incredibly strange.  The server in this case is the code calling GetClientHandleAsString(), and closing the inherited handle in the child process doesn't cause the pipe to break (you can pass the same client handle to multiple client processes and have each of them close their handle without affecting the pipe's ability to be used).  Plus, the AnonymousServerPipeStream holds a reference to the ClientSafePipeHandle, so as long as the server process is holding a reference to the AnonymousServerPipeStream (which it would use to detect any theoretical break), the ClientSafePipeHandle wouldn't be finalized anyway.

ericstj:
Ok.  In that case I sign off.  Thanks for entertaining my Qs.  :+1:

stephentoub:
:smile: Thanks for the review!

Issue #1772, opened 2015-05-15
Unix: System.IO.Pipes; Operations not throwing exceptions after Dispose


bpschoch:
```C#
[Fact]
        public static void ServerWhenDisplosedThrows()
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream("unique2", PipeDirection.InOut))
            {
                server.Dispose();

                Assert.Throws<ObjectDisposedException>(() => server.Disconnect());  // disconnect when disposed
                Assert.Throws<ObjectDisposedException>(() => server.WaitForConnection());  // disconnect when disposed
                Assert.Throws<ObjectDisposedException>(() => server.WriteByte(5));
                Assert.Throws<ObjectDisposedException>(() => server.ReadByte());
                Assert.Throws<ObjectDisposedException>(() => server.IsMessageComplete);
            }
        }
```

13:28:39 [31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerWhenDisplosedThrows [FAIL]
13:28:39 [37m      Assert.Throws() Failure
13:28:39       Expected: typeof(System.ObjectDisposedException)
13:28:39       Actual:   typeof(System.InvalidOperationException): Pipe handle has not been set.  Did your PipeStream implementation call InitializeHandle?
13:28:39 [30m      Stack Trace:
13:28:39 [37m            at System.IO.Pipes.NamedPipeServerStream.CheckDisconnectOperations()
13:28:39             at System.IO.Pipes.NamedPipeServerStream.Disconnect()
13:28:39             at System.IO.Pipes.Tests.NamedPipesThrowsTests.<>c__DisplayClass64.<ServerWhenDisplosedThrows>b__5a()

stephentoub:
Fixed by #1884

Issue #1766, opened 2015-05-15
System.IO.Pipes; NamePipeServerStream hangs when trying to connect when using SafePipeHandle parameter on constructor


bpschoch:
```C#
[Fact]
public static void ServerCloneTests()
    {
        const string pipeName = "fooclone";
        byte[] msg1 = new byte[] { 5, 7, 9, 10 };
        byte[] received1 = new byte[] { 0, 0, 0, 0 };

        using (NamedPipeServerStream serverBase = new NamedPipeServerStream(pipeName, PipeDirection.InOut, 1, PipeTransmissionMode.Byte))
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream(PipeDirection.InOut, false, false, serverBase.SafePipeHandle))
            {
                using (NamedPipeClientStream client = new NamedPipeClientStream(".", pipeName, PipeDirection.InOut, PipeOptions.None))
                {
                    Task clientTask = Task.Run(() =>
                    {
                        client.Connect();

                        client.Write(msg1, 0, msg1.Length);

                        int serverCount = client.NumberOfServerInstances;
                        Assert.Equal(1, serverCount);
                    });
                }

                server.WaitForConnection();
                int len1 = server.Read(received1, 0, msg1.Length);
                Assert.Equal(msg1.Length, len1);
                Assert.Equal(msg1, received1);
            }
        }
    }
```
This test hangs

stephentoub:
The test is buggy: it's closing the client stream as soon as it's scheduled the task to connect from the client.  Since it's then ignoring the clientTask and never awaiting it, the exceptions that are getting generated are getting ignored, too.  I have a fix locally and will submit it as part of a set of pipes-related fixes.

stephentoub:
Fixed by #1884

Issue #1765, opened 2015-05-15
Unix: System.IO.Pipes; Disconnected pipe not throwing exceptions on other side of pipe


bpschoch:
```C#
[Fact]
        public static void ClientDisconnectedPipeThrows()
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream("testServer2", PipeDirection.InOut))
            {
                using (NamedPipeClientStream client = new NamedPipeClientStream("testServer2"))
                {
                    byte[] buffer = new byte[] { 0, 0, 0, 0 };

                    Task clientConnect1 = client.ConnectAsync();
                    server.WaitForConnection();
                    clientConnect1.Wait();

                    Assert.True(client.IsConnected);
                    Assert.True(server.IsConnected);

                    client.Dispose();
                    
                    Assert.Throws<IOException>(() => server.Write(buffer, 0, buffer.Length));
                    Assert.Throws<IOException>(() => server.WriteByte(123));
                    Assert.Throws<IOException>(() => server.Flush());
                    int length = server.Read(buffer, 0, buffer.Length);
                    Assert.Equal(0, length);
                    int byt = server.ReadByte();
                }
            }
        }

        [Fact]
        public static void ServerDisconnectedPipeThrows()
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream("testServer3", PipeDirection.InOut))
            {
                using (NamedPipeClientStream client = new NamedPipeClientStream("testServer3"))
                {
                    byte[] buffer = new byte[] { 0, 0, 0, 0 };

                    Task clientConnect1 = client.ConnectAsync();
                    server.WaitForConnection();
                    clientConnect1.Wait();

                    Assert.True(client.IsConnected);
                    Assert.True(server.IsConnected);

                    server.Dispose();

                    Assert.Throws<IOException>(() => client.Write(buffer, 0, buffer.Length));
                    Assert.Throws<IOException>(() => client.WriteByte(123));
                    Assert.Throws<IOException>(() => client.Flush());
                    int length = client.Read(buffer, 0, buffer.Length);
                    Assert.Equal(0, length);
                    int byt = client.ReadByte();
                }
            }
        }
```

13:30:03 �[31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ClientDisconnectedPipeThrows [FAIL]
13:30:03 �[37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.IO.IOException)
[...]

Issue #1764, opened 2015-05-15
Unix: System.IO.Pipes; NamePipeServerStream throwing exception when reading TransmissionMode Property when not connected


bpschoch:
```C#
 [Fact]
        public static void ServerNotConnectedThrows()
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream("tempnotconnected", PipeDirection.InOut, 1))
            {
                // doesn't throw execeptions
                PipeTransmissionMode transmitMode = server.TransmissionMode;
                Assert.Throws<ArgumentOutOfRangeException>(() => server.ReadMode = (PipeTransmissionMode)999);

                if (Interop.IsWindows)
                {
                    int inbuffersize = server.InBufferSize;
                    int outbuffersize = server.OutBufferSize;
                }
                else
                {
                    Assert.Throws<PlatformNotSupportedException>(() => server.InBufferSize);
                    Assert.Throws<PlatformNotSupportedException>(() => server.OutBufferSize);
                }
                PipeTransmissionMode readMode = server.ReadMode;
                server.ReadMode = PipeTransmissionMode.Byte;

                Assert.Throws<InvalidOperationException>(() => server.WriteByte(5));
                Assert.Throws<InvalidOperationException>(() => server.ReadByte());
                Assert.Throws<InvalidOperationException>(() => server.Disconnect());    // disconnect when not connected 
                Assert.Throws<InvalidOperationException>(() => server.IsMessageComplete);
            }
        }
```

13:30:03 �[31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerNotConnectedThrows [FAIL]
13:30:03 �[37m      System.InvalidOperationException : Pipe handle has not been set.  Did your PipeStream implementation call InitializeHandle?
13:30:03 �[30m      Stack Trace:
13:30:03 �[37m            at System.IO.Pipes.PipeStream.CheckPipePropertyOperations()
13:30:03             at System.IO.Pipes.PipeStream.get_TransmissionMode()
13:30:03             at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerNotConnectedThrows()

stephentoub:
Fixed by #1884

PR #1884, opened 2015-06-01
Fix System.IO.Pipes bugs (src and test)

### LINKED ISSUE #1762: Unix: System.IO.Pipes; Operations not throwing exceptions after disconnect
```C#
 [Fact]
        public static void ServerAfterDisconnectThrows()
        {
            using (NamedPipeServerStream server = new NamedPipeServerStream("unique3", PipeDirection.InOut, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous))
            {
                using (NamedPipeClientStream client = new NamedPipeClientStream("unique3"))
                {
                    Task clientConnect = client.ConnectAsync(); 
                    server.WaitForConnection();
                    Assert.Throws<InvalidOperationException>(() => server.IsMessageComplete);
                    Assert.Throws<InvalidOperationException>(() => server.WaitForConnection());
                    Assert.ThrowsAsync<InvalidOperationException>(() => server.WaitForConnectionAsync());

                    server.Disconnect();
                    Assert.Throws<InvalidOperationException>(() => server.Disconnect());    // double disconnect
                    Assert.Throws<InvalidOperationException>(() => server.WriteByte(5));
                    Assert.Throws<InvalidOperationException>(() => server.ReadByte());
                    Assert.Throws<InvalidOperationException>(() => server.IsMessageComplete);
                }
            }
        }
```

13:30:02 �[39;49m�[31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerAfterDisconnectThrows [FAIL]
13:30:02 �[37m      Assert.Throws() Failure
13:30:02       Expected: typeof(System.InvalidOperationException)
13:30:02       Actual:   (No exception was thrown)
13:30:02 �[30m      Stack Trace:
13:30:02 �[37m            at System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerAfterDisconnectThrows()
13:30:03 �[31m   System.IO.Pipes.Tests.NamedPipesThrowsTests.ServerWhenDisplosedThrows [FAIL]
13:30:03 �[37m      Assert.Throws() Failure
13:30:03       Expected: typeof(System.ObjectDisposedException)
13:30:03       Actual:   typeof(System.InvalidOperationException): Pipe handle has not been set.  Did your PipeStream implementation call InitializeHandle?
13:30:03 �[30m      Stack Trace:
13:30:03 �[37m            at System.IO.Pipes.NamedPipeServerStream.CheckDisconnectOperations()
13:30:03             at System.IO.Pipes.NamedPipeServerStream.Disconnect()
13:30:03             at System.IO.Pipes.Tests.NamedPipesThrowsTests.<>c__DisplayClass64.<ServerWhenDisplosedThrows>b__5a()

stephentoub:
Fix multiple test bugs in System.IO.Pipes.Tests and multiple product bugs in the Unix implementation of System.IO.Pipes, primarily around error cases surfacing an exception or the right exception and at the right time.

Product changes:
- In the Windows implementation, avoids a few unnecessary allocations when registring for cancellation.
- Common code shared between Windows and Unix is used to check the state of the streams prior to various operations, including validating the current state of the pipe's handle.  On Windows, the handle is always initialized at creation, whereas in the current Unix implementation, the handle's creation is deferred until the connection is established.  As such, code that checked the validity of the handle needed to be updated to not require that the handle be set on Unix.
- The Unix NamedPipeServerStream.WaitForConnection implementation was missing a check to fail if the connection was already established.  On Windows, this is handled by the underlying OS function; on Unix, we just open a file, which any number of consumers can do, so there's no valdation provided by the OS for the constraint we need.
- The Unix implementation of PipeStream was not overriding Read/WriteAsync, such that validation of the inputs was happening asynchronously.  I've refactored out the Read/WriteAsync validation logic to be shared between Windows and Unix, with the shared code overriding these methods.
- In the Unix implementation, if any operation detects that the pipe has been broken, it needs to transition the pipe to a state of Broken.  The Windows implementation was already doing this.

Test changes:
- A bunch of calls to Assert.ThrowsAsync were ignoring the returned Task, which meant that a) any assertion failures were being ignored, b) operations may have continued to run well past the end of the test, and c) race conditions were occurring due to unexpected concurrency between operations.  These have been addressed by either awaiting on the resulting task (when the operation is expected to run asynchronously) or by changing to using Throw instead of ThrowAsync (when we expect the exception to propagate out synchronously).
- Some tasks were being waited on synchronously, which could have potentially led to deadlock situations in the tests.  I changed everything relevant to use async/await.
- The ServerCloneTests was failing (on both Windows and Unix) because it was closing the client stream prior to being done with the test.  I fixed up the disposal ordering and re-enabled the test. I also had to change the order in which operations occurred around waiting for the connection, due to the Unix implementation not establishing the handle until the connection is established, rather than during construction.
- I added some Interop.IsWindows checks for functionality known to be unavailable on Unix, e.g. NumberOfServerInstances.
- I changed a few tests to use In vs Out rather than InOut, to focus the test on what specifically was being tested, and because the Unix support for InOut isn't quite the same as on Windows.

Fixes #1762, #1763, #1764, #1765, #1766, #1772.

(This commit also currently includes the changes in #1871; I will fix those appropriately once #1871 is merged.)

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
@@ -6,11 +6,20 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using System.Security;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace System.IO.Pipes
 {
     public abstract partial class PipeStream : Stream
     {
+        // The Windows implementation of PipeStream sets the stream's handle during 
+        // creation, and as such should always have a handle, but the Unix implementation 
+        // sometimes sets the handle not during creation but later during connection.  
+        // As such, validation during member access needs to verify a valid handle on 
+        // Windows, but can't assume a valid handle on Unix.
+        internal const bool CheckOperationsRequiresSetHandle = false;
+
         private const string PipeDirectoryPath = "/tmp/corefxnamedpipes/";
 
         internal static string GetPipePath(string serverName, string pipeName)
@@ -62,7 +71,7 @@ namespace System.IO.Pipes
 
         /// <summary>Throws an exception if the supplied handle does not represent a valid pipe.</summary>
         /// <param name="safePipeHandle">The handle to validate.</param>
-        internal static void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
+        internal void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
         {
             SysCall(safePipeHandle, (fd, _, __) =>
             {
@@ -113,6 +122,19 @@ namespace System.IO.Pipes
             }
         }
 
+        [SecuritySafeCritical]
+        private Task<int> ReadAsyncCore(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            Debug.Assert(_handle != null, "_handle is null");
+            Debug.Assert(!_handle.IsClosed, "_handle is closed");
+            Debug.Assert(CanRead, "can't read");
+            Debug.Assert(buffer != null, "buffer is null");
+            Debug.Assert(offset >= 0, "offset is negative");
+            Debug.Assert(count >= 0, "count is negative");
+
+            return base.ReadAsync(buffer, offset, count, cancellationToken);
+        }
+
         [SecurityCritical]
         private unsafe void WriteCore(byte[] buffer, int offset, int count)
         {
@@ -137,6 +159,19 @@ namespace System.IO.Pipes
                     offset += bytesWritten;
                 }
             }
+        }
+
+        [SecuritySafeCritical]
+        private Task WriteAsyncCore(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            Debug.Assert(_handle != null, "_handle is null");
+            Debug.Assert(!_handle.IsClosed, "_handle is closed");
+            Debug.Assert(CanWrite, "can't write");
+            Debug.Assert(buffer != null, "buffer is null");
+            Debug.Assert(offset >= 0, "offset is negative");
+            Debug.Assert(count >= 0, "count is negative");
+
+            return base.WriteAsync(buffer, offset, count, cancellationToken);
         }
 
         // Blocks until the other end of the pipe has read in all written buffer.
@@ -277,7 +312,7 @@ namespace System.IO.Pipes
         /// Arguments are expected to be passed via <paramref name="arg1"/> and <paramref name="arg2"/>
         /// so as to avoid delegate and closure allocations at the call sites.
         /// </remarks>
-        private static long SysCall(
+        private long SysCall(
             SafePipeHandle handle,
             Func<int, IntPtr, int, long> sysCall,
             IntPtr arg1 = default(IntPtr), int arg2 = default(int))
@@ -292,9 +327,23 @@ namespace System.IO.Pipes
                 int fd = (int)handle.DangerousGetHandle();
                 Debug.Assert(fd >= 0);
 
-                long result;
-                while (Interop.CheckIo(result = sysCall(fd, arg1, arg2))) ;
-                return result;
+                while (true)
+                {
+                    long result = sysCall(fd, arg1, arg2);
+                    if (result < 0)
+                    {
+                        int errno = Marshal.GetLastWin32Error();
+
+                        if (errno == Interop.Errors.EINTR)
+                            continue;
+
+                        if (errno == Interop.Errors.EPIPE)
+                            State = PipeState.Broken;
+
+                        throw Interop.GetExceptionForIoErrno(errno);
+                    }
+                    return result;
+                }
             }
             finally
             {
--- a/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Windows.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Windows.cs
@@ -14,6 +14,7 @@ namespace System.IO.Pipes
 {
     public abstract partial class PipeStream : Stream
     {
+        internal const bool CheckOperationsRequiresSetHandle = true;
         private ThreadPoolBoundHandle _threadPoolBinding;
 
         internal static string GetPipePath(string serverName, string pipeName)
@@ -28,7 +29,7 @@ namespace System.IO.Pipes
 
         /// <summary>Throws an exception if the supplied handle does not represent a valid pipe.</summary>
         /// <param name="safePipeHandle">The handle to validate.</param>
-        internal static void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
+        internal void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
         {
             // Check that this handle is infact a handle to a pipe.
             if (Interop.mincore.GetFileType(safePipeHandle) != Interop.mincore.FileTypes.FILE_TYPE_PIPE)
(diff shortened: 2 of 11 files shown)


bpschoch:
wow great work.  I have a version almost ready to go that refactors many of the throws that should be done on each pipe variant into a common base class.  I'll wait until this commits and merge your changes in.

Also when I did the tests, I wondered why on the xxxDisconnectedPipeThrows test, after a disconnect, the write type operations throw exceptions while the read type don't.  It seemed strange that they both wouldn't do the same.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

src/System.IO.Pipes/tests/AnonymousPipesThrowsTests.cs
```diff
@@ -1,17 +1,15 @@
...
-using Xunit;
-
 using Microsoft.Win32.SafeHandles;
+using Xunit;
 
 namespace System.IO.Pipes.Tests
 {
     public class AnonymousPipesThrowsTests
     {
+        private static void NotReachable(object obj) { Assert.True(false, "This should not be reached."); }
```
KrzysztofCwalina: Any reason you are not using Assert.Fail here?

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
```diff
@@ -62,7 +71,7 @@ internal static string GetPipePath(string serverName, string pipeName)
 
         /// <summary>Throws an exception if the supplied handle does not represent a valid pipe.</summary>
         /// <param name="safePipeHandle">The handle to validate.</param>
-        internal static void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
+        internal void ValidateHandleIsPipe(SafePipeHandle safePipeHandle)
```
KrzysztofCwalina: Why did you change it to an instance method?

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Windows.cs
```diff
@@ -120,6 +128,13 @@ private unsafe void WriteCore(byte[] buffer, int offset, int count)
...
+        [SecuritySafeCritical]
+        private Task WriteAsyncCore(Byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            ReadWriteAsyncParams state = new ReadWriteAsyncParams(buffer, offset, count, cancellationToken);
+            return Task.Factory.FromAsync(BeginWrite, EndWrite, state);
+        }
+
         // Blocks until the other end of the pipe has read in all written buffer.
         [SecurityCritical]
         public void WaitForPipeDrain()
```
KrzysztofCwalina: First time I read it, I misread it as "WaitForPipeDream" :-)

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
```diff
@@ -292,9 +327,23 @@ internal static Interop.libc.OpenFlags TranslateFlags(PipeDirection direction, P
                 int fd = (int)handle.DangerousGetHandle();
                 Debug.Assert(fd >= 0);
 
-                long result;
-                while (Interop.CheckIo(result = sysCall(fd, arg1, arg2))) ;
-                return result;
+                while (true)
```
KrzysztofCwalina: Should we worry about this silently never terminating?

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
stephentoub: Because SysCall that it calls became an instance method, too.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
stephentoub: I don't think so. The only way that'd happen is if every system call got interrupted each and every time.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Windows.cs
stephentoub: Heh :smile:

src/System.IO.Pipes/tests/AnonymousPipesThrowsTests.cs
stephentoub: Because there's no such method? :smile:

bpschoch: Perhaps XUnit needs a call like Assert.ThrowsAsyncImmediate()  To ensure the throw happens before the task is created (I assume that is the purpose of NotReachable())

stephentoub: > I assume that is the purpose of NotReachable()

Yes.  Assert.Throws would actually be the right thing here, except that as a "help" it detects that the object returned is a Task and throws an exception to say that Task-returning methods should instead use Async.ThrowsAsync.

ericstj:
I reviewed the product code and skimmed the test code and it looks ok to me.  :+1: 
/cc @krwq who did the initial test port.

stephentoub:
Thanks Eric and Krzysztof C. for reviewing.  Other Krzysztof, any additional thoughts?

src/System.IO.Pipes/tests/AnonymousPipesThrowsTests.cs
KrzysztofCwalina: Oops. I confused XUnit with VS test.

stephentoub: :smile:

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
KrzysztofCwalina: And why did SysCall become an instance method?

ericstj: Because he's updating the (instance) pipe state when he gets an error.

stephentoub: Right. Is there some reason I've missed that this shouldn't be an instance method?

KrzysztofCwalina: Ah, I see it now. Thanks.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.cs
```diff
@@ -166,6 +196,35 @@ public override void Write(byte[] buffer, int offset, int count)
             return;
         }
 
+        [SecuritySafeCritical]
+        public override Task WriteAsync(Byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```
krwq: nit: byte[] (low case)

krwq:
LGTM :+1:

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.cs
stephentoub: Thanks, fixed.

stephentoub:
> wow great work. I have a version almost ready to go that refactors many of the throws that should be done on each pipe variant into a common base class. I'll wait until this commits and merge your changes in.

Thanks :smile:

PR #1886, opened 2015-06-01
Output Console ANSI sequences on Unix only for CHR devices

### LINKED ISSUE #1806: Don't write ANSI escape sequences when output is not a tty?
When you redirect standard out to a file and then change the console color, the ANSI escape sequences are thrown into the file as well, which is not great.

Would it be possible to detect and not write the ASNI codes codes if we are redirected?

Steve, what do you think about this?

stephentoub:
Modifies Console.*Color to output ANSI escape sequences only when wrapping a UnixConsoleStream that itself has an underlying file descriptor for a character device.  This is done to try to avoid outputting escape sequence gibberish when Console.SetOut is used to target an arbitrary stream or when Unix redirection is used to redirect stdout.  This commit includes adding a Unix-only test to verify we're outputting ANSI escape sequences and that we don't do so when SetOut is used to target an arbitrary stream.

Fixes #1806.

### DIFF
--- a/src/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/System.Console/src/System/ConsolePal.Unix.cs
@@ -59,10 +59,43 @@ namespace System
 
         public static void ResetColor()
         {
+            if (!ConsoleOutIsTerminal)
+                return;
+
             string resetFormat = TerminalColorInfo.Instance.ResetFormat;
             if (resetFormat != null)
             {
                 Console.Write(resetFormat);
+            }
+        }
+
+        /// <summary>Gets whether Console.Out is targeting a terminal display.</summary>
+        private static bool ConsoleOutIsTerminal
+        {
+            get
+            {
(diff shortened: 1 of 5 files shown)


src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -66,6 +69,36 @@ public static void ResetColor()
             }
         }
 
+        /// <summary>Gets whether Console.Out is targeting a terminal display.</summary>
+        private static bool ConsoleOutIsTty
```
pgavlin: `ConsoleOutIsTerminal` might be a bit more readable than `ConsoleOutIsTty`.

stephentoub: Sure, I'll change it.

pgavlin:
LGTM modulo the one comment re: naming.

stephentoub:
Thanks, Pat!

stephentoub:
@dotnet-bot test this please

Issue #1887, opened 2015-06-01
Microsoft.Framework.WebEncoders.UnicodeRangesTests.Range_All failed in Windows CI build


stephentoub:
Failed in Linux CI build:
```
Starting:    System.Text.Encodings.Web.Tests
     Microsoft.Framework.WebEncoders.UnicodeRangesTests.Range_All [FAIL]
        Assert.Same() Failure
        Expected: UnicodeRange { FirstCodePoint = 0, Length = 65536 }
        Actual:   UnicodeRange { FirstCodePoint = 0, Length = 65536 }
        Stack Trace:
d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Text.Encodings.Web\tests\UnicodeRangesTests.cs(209,0): at Microsoft.Framework.WebEncoders.UnicodeRangesTests.Range_Unicode(UInt16 first, UInt16 last, String blockName)
d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Text.Encodings.Web\tests\UnicodeRangesTests.cs(32,0): at Microsoft.Framework.WebEncoders.UnicodeRangesTests.Range_All()
```

stephentoub:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_debug_prtest/1508/console

KrzysztofCwalina:
@stephentoub, I created a PR to fix that: https://github.com/dotnet/corefx/pull/1889

stephentoub:
Fixed by #1889

PR #1889, opened 2015-06-01
Fixes bug #1887

KrzysztofCwalina:
Some time ago, we agreed to not guarantee that instances of encoders and ranges can be different.
This change just removes the remaining tests testing the old behavior.

### DIFF
--- a/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/UnicodeRanges.cs
+++ b/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/UnicodeRanges.cs
@@ -32,9 +32,8 @@ namespace System.Text.Unicode
         {
             // If the range hasn't been created, create it now.
             // It's ok if two threads race and one overwrites the other's 'range' value.
-            var newRange = new UnicodeRange(0, 0);
-            Volatile.Write(ref range, newRange);
-            return newRange;
+            range = new UnicodeRange(0, 0);
+            return range;
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)] // the caller should be inlined, not this method
@@ -42,10 +41,8 @@ namespace System.Text.Unicode
         {
             // If the range hasn't been created, create it now.
             // It's ok if two threads race and one overwrites the other's 'range' value.
-            Debug.Assert(last > first, "Code points were specified out of order.");
-            var newRange = UnicodeRange.Create(first, last);
-            Volatile.Write(ref range, newRange);
-            return newRange;
+            range = UnicodeRange.Create(first, last);
+            return range;
         }
     }
 }
--- a/src/System.Text.Encodings.Web/tests/UnicodeRangesTests.cs
+++ b/src/System.Text.Encodings.Web/tests/UnicodeRangesTests.cs
@@ -20,10 +20,6 @@ namespace Microsoft.Framework.WebEncoders
             // Test 1: the range should be empty
             Assert.Equal(0, range.FirstCodePoint);
             Assert.Equal(0, range.Length);
-
-            // Test 2: calling the property multiple times should cache and return the same range instance
-            UnicodeRange range2 = UnicodeRanges.None;
-            Assert.Same(range, range2);
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/UnicodeRanges.cs
```diff
@@ -32,9 +32,8 @@ private static UnicodeRange CreateEmptyRange(ref UnicodeRange range)
         {
             // If the range hasn't been created, create it now.
             // It's ok if two threads race and one overwrites the other's 'range' value.
-            var newRange = new UnicodeRange(0, 0);
-            Volatile.Write(ref range, newRange);
-            return newRange;
+            var range = new UnicodeRange(0, 0);
+            return range;
```
stephentoub: Do we need the barrier implied by the Volatile.Write to guarantee that another racing thread sees a fully instantiated object?

KrzysztofCwalina: I don't think we need it. CLR must ensure partially instantiated objects are not visible outside of their ctors, and the assignment is atomic as it's a reference type.

stephentoub: > CLR must ensure partially instantiated objects are not visible outside of their ctors

Yes, on the same thread.

> the assignment is atomic as it's a reference type

That just means the reference itself won't get torn.

This might be fine, as the CLR does add extra fences to provide a stronger memory model than the CLI spec, but it'd be good to verify. There's a reason Volatile.Write exists ;)

KrzysztofCwalina: Yes, the reference won't get torn and ctor finishes before the object is visible. What else could cause racing threads to see partially instantiated object? I am not following; we are talking about very straightforward code that is all over the place in our code base. 

Also, not that the CLUI spec talks about this explicitly:

I.12.6.8 Other memory model i s sues
All memory allocated for static variables (other than those assigned RVAs within a PE file, see
Partition II) and objects shall be zeroed before they are made visible to any user code.
A conforming implementation of the CLI shall ensure that, even in a multi-threaded environment
and without proper user synchronization, objects are allocated in a manner that prevents
unauthorized memory access and prevents invalid operations from occurring. In particular, on
multiprocessor memory systems where explicit synchronization is required to ensure that all
relevant data structures are visible (for example, vtable pointers) the Execution Engine shall be
responsible for either enforcing this synchronization automatically or for converting errors due to
lack of synchronization into non-fatal, non-corrupting, user-visible exceptions.
It is explicitly not a requirement that a conforming implementation of the CLI guarantee that all
state updates performed within a constructor be uniformly visible before the constructor
completes. CIL generators can ensure this requirement themselves by inserting appropriate calls
to the memory barrier or volatile write instructions.

stephentoub: > What else could cause racing threads to see partially instantiated object? 

In theory the compilers/hardware are free to reorder the memory accesses such that the writes to initialize the instance are done after the instance is published; this wouldn't be visible to the initializing thread but could be to others. This is why some forms of double-checked locking are theoretically problematic and why the CLR sometimes incurs extra fences/costs to prevent dev bugs from manifesting. Igor wrote some nice articles on this stuff: https://msdn.microsoft.com/en-us/magazine/jj863136.aspx
Now, this may very well be one of the cases where the CLR will insert extra fences, I'm simple saying you should confirm before you remove the fences. @ericeil will know...

stephentoub: > Also, not that the CLUI spec talks about this explicitly:

Yes, and it ends with this:
"It is explicitly *not* a requirement that a conforming implementation of the CLI guarantee that all
 state updates performed within a constructor be uniformly visible before the constructor
 completes. CIL generators can ensure this requirement themselves by inserting appropriate calls
 to the memory barrier or volatile write instructions."

KrzysztofCwalina:
@stephentoub, I submitted an additional fix and it seemed to deleted (hid) our conversation about memory barriers.

stephentoub:
@KrzysztofCwalina, if you want to proceed with this, I'm ok with that; I'd suggest just either only merging the test changes and not the source changes, or opening an issue to track whether the removal of the Volatile.Writes was acceptable.

KrzysztofCwalina:
Thanks. I will merge it, file an issue, and talk to Eric about it.

KrzysztofCwalina:
I filed issue #1900

PR #1919, opened 2015-06-03
Reconciling recent changes in System.Linq.Expressions and System.Dyna…

VSadov:
…mic.Runtime

The branch were these fixes were made should be switched over to use
dotnet\corefx, but before that happens the sources must match.


VSadov:
@stephentoub @AlekseyTs - please review

VSadov:
@weshaggard

src/Common/src/System/Dynamic/Utils/CacheDict.cs
```diff
@@ -11,42 +12,71 @@ namespace System.Dynamic.Utils
...
     {
-        private readonly Dictionary<TKey, KeyInfo> _dict = new Dictionary<TKey, KeyInfo>();
-        private readonly LinkedList<TKey> _list = new LinkedList<TKey>();
-        private readonly int _maxSize;
+
+        // cache size is always ^2. 
+        // items are placed at [hash ^ mask]
+        // new item will displace previous one at the same location.
+        private readonly int mask;
+        private readonly Entry[] entries;
```
stephentoub: I understand from your PR comments that you need this code to match what's in an internal branch.  Will we separately re-update this to fix up the names to match the style guidelines, e.g. starting these fields with underscores?

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
```diff
@@ -10,6 +10,10 @@
 using System.Runtime.CompilerServices;
 using System.Reflection;
 
+#if FEATURE_REFEMIT
```
stephentoub: What project sets this?  I don't see FEATURE_REFEMIT in any existing .cs or .csproj file in the repo.

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
```diff
@@ -22,5 +26,138 @@ internal static partial class DelegateHelpers
...
+                paramTypes[i + 1] = parameters[i].ParameterType;
+            }
+
+            DynamicMethod thunkMethod = new DynamicMethod("Thunk", returnType, paramTypes);
+            ILGenerator ilgen = thunkMethod.GetILGenerator();
+
+            LocalBuilder argArray = ilgen.DeclareLocal(typeof(object[]));
+            LocalBuilder retValue = ilgen.DeclareLocal(typeof(object));
+
+            // create the argument array
```
stephentoub: Separate from this PR, should we consider special casing parameters.Length == 0 and instead generating code that uses ```Array.Empty<object>```?

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
```diff
@@ -22,5 +26,138 @@ internal static partial class DelegateHelpers
...
+
+            if (hasReturnValue)
+            {
+                ilgen.Emit(OpCodes.Ldloc, retValue);
+                ilgen.Emit(OpCodes.Unbox_Any, ConvertToBoxableType(returnType));
+            }
+
+            ilgen.Emit(OpCodes.Ret);
+
+            // TODO: we need to cache these.
```
stephentoub: Is this TODO still relevant?  Is it tracked by some issue or bug?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
```diff
@@ -264,6 +264,8 @@ internal sealed class LightCompiler
 
         private readonly LightCompiler _parent;
 
+        private static LocalDefinition[] s_emptyLocals = new LocalDefinition[0];
```
stephentoub: Should we use ```Array.Empty<LocalDefinition>``` at call sites instead of adding this field?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
```diff
@@ -1416,28 +1433,139 @@ private void CompileSwitchExpression(Expression expr)
...
-                throw new PlatformNotSupportedException(SR.NonConstantCaseValuesInSwitchNotSupported);
+                var switchCase = node.Cases[i];
+
+                int caseOffset = _instructions.Count - switchIndex;
+                foreach (ConstantExpression testValue in switchCase.TestValues)
+                {
+                    var key = (T)testValue.Value;
+                    if (!caseDict.ContainsKey(key))
+                    {
+                        caseDict.Add(key, caseOffset);
```
stephentoub: @KrzysztofCwalina, I see this pattern fairly frequently.  Have we considered adding a ```bool TryAdd(TKey key, TValue value)``` to ```Dictionary<TKey, TValue>```, so as to avoid the need to do multiple lookups when the key isn't yet in the dictionary?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
```diff
@@ -2390,7 +2529,7 @@ private void CompileInvocationExpression(Expression expr)
 
             if (typeof(LambdaExpression).IsAssignableFrom(node.Expression.Type))
             {
-                var compMethod = node.Expression.Type.GetMethod("Compile", Array.Empty<Type>());
+                var compMethod = node.Expression.Type.GetMethod("Compile", new Type[0]);
```
stephentoub: Why did we change this?

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
VSadov: The flag is set by special project that produces debug versions of System.Linq.Expressions and System.Dynamic.Runtime for NetNative. This part is likely to change once project files are redirected to use these sources.

VSadov: I may need another flag to switch between compiling and interpreting flavors. I just want to get the sources to be the same as a first step.

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
VSadov: Yes. In a number of places the change has been made already in one or another branch. Having  just one place for fixes like this is my priority here. I will make a note about checking if Array.Empty can be used in more places.

stephentoub: Sounds good.

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
stephentoub: Ok.

src/Common/src/System/Dynamic/Utils/DelegateHelpers.cs
VSadov: entered https://github.com/dotnet/corefx/issues/1920
It is not clear whether caching these thunks is possible/profitable. It would need some researching. 
There was no issue to track this, primarily since the code was not in the official repo.

src/Common/src/System/Dynamic/Utils/CacheDict.cs
VSadov: I just copied it as-is. There could be many style mismatches here.
I think this falls under the rule of preexisting code, since this is from a bug fix on the desktop runtime. I expect that that at some point the style will be addressed via some style-fixing tool.

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
VSadov: There were a bunch of cases where Array.Empty change was applied to one branch or another and I tried to keep the change.
I somehow missed this one.

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
VSadov: I think this was proposed many times, in particular since concurrent dictionary has this. Not sure why that was not done. Perhaps because IDictionary does not have this and it would be breaking to add?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
VSadov: Will fix. We should not lose such changes.

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
VSadov: Fixed

src/Common/src/System/Dynamic/Utils/CacheDict.cs
AlekseyTs: We are actually reformatted all preexisting code before putting it on GitHub.

AlekseyTs: I mean reformatted in order to follow the style guidelines.

PR #1923, opened 2015-06-04
Fix FileVersionInfo test for CompanyName

stephentoub:
Code coverage highlighted that we had a bug in VerifyVersionInfo where the test for CompanyName was actually testing Comments.  Fixing that then highlighted a bug in the test data when compared to the actual native DLL.


stephentoub:
cc: @pallavit

KrzysztofCwalina:
looks good

stephentoub:
Thanks, Krzysztof.

PR #1924, opened 2015-06-04
Test both managed/zlib System.IO.Compression implementations

stephentoub:
System.IO.Compression has two implementations for compression/decompression: one based on zlib, and one implemented in managed code.  As we can use the managed implementation as a fallback in case zlib is unavailable, we need to ensure it's functioning properly.

This commit adds a debug-only static field to DeflateStream that supports overriding the normal logic for whether zlib or the managed implementation is used.  The DeflateStream and GZipStream tests are then augmented to force both implementations to be used.

### DIFF
--- a/src/System.IO.Compression/tests/DeflateStreamTests.cs
+++ b/src/System.IO.Compression/tests/DeflateStreamTests.cs
@@ -1,23 +1,40 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections.Generic;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
-using System.IO;
-using System.Reflection;
-using System.IO.Compression;
 using Xunit;
 
 namespace System.IO.Compression.Tests
 {
-    public partial class DeflateStreamTests
+    public class ZLibDeflateStreamTests : DeflateStreamTests, IDisposable
+    {
+        public ZLibDeflateStreamTests() { SetWorkerMode("zlib"); }
+        public void Dispose() { SetWorkerMode("unknown"); }
+    }
+
+    public class ManagedDeflateStreamTests : DeflateStreamTests, IDisposable
+    {
+        public ManagedDeflateStreamTests() { SetWorkerMode("managed"); }
+        public void Dispose() { SetWorkerMode("unknown"); }
+    }
+
+    public abstract class DeflateStreamTests
     {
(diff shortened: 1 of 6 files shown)


stephentoub:
cc: @FiveTimesTheFun

src/System.IO.Compression/tests/DeflateStreamTests.cs
```diff
@@ -1,23 +1,40 @@
...
     {
         static string gzTestFile(String fileName) { return Path.Combine("GZTestData", fileName); }
 
+        internal static void SetWorkerMode(string mode)
+        {
+            FieldInfo forceType = typeof(DeflateStream).GetTypeInfo().GetDeclaredField("s_forcedTestingDeflaterType");
+            if (forceType != null)
+            {
+                forceType.SetValue(null, mode == "zlib" ? (byte)2 : mode == "managed" ? (byte)1 : (byte)0);
+            }
```
FiveTimesTheFun: Else (eg. running these tests on ret) put a warning on the console, or fail the test?

src/System.IO.Compression/tests/DeflateStreamTests.cs
```diff
@@ -1,23 +1,40 @@
...
+    {
+        public ManagedDeflateStreamTests() { SetWorkerMode("managed"); }
+        public void Dispose() { SetWorkerMode("unknown"); }
+    }
+
+    public abstract class DeflateStreamTests
     {
         static string gzTestFile(String fileName) { return Path.Combine("GZTestData", fileName); }
 
+        internal static void SetWorkerMode(string mode)
```
FiveTimesTheFun: Since this is shared between deflatestream and gzipstream tests, is corefx\src\Common\tests\Compression\Utilities\ a better place?

FiveTimesTheFun:
Thanks, pair of comments, but otherwise LGTM :)

src/System.IO.Compression/tests/DeflateStreamTests.cs
stephentoub: I changed it to output a warning message.

src/System.IO.Compression/tests/DeflateStreamTests.cs
stephentoub: Moved it to a Common class.

stephentoub:
Thanks for the review, Matt.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #1949, opened 2015-06-07
Implement System.IO.Pipes buffer sizes on Linux

### LINKED ISSUE #1730: Implement buffer-size support in Pipes on Linux
In System.IO.Pipes on Unix, we currently ignore custom buffer sizes and just use the system default, and we throw PlatformNotSupportedException when accessing the current buffer sizes from the relevant In/Out properties.  We should be able to do better than this, at least on Linux, e.g. via fcntl and F_SETPIPE_SZ.

stephentoub:
On Linux, fcntl supports F_GETPIPE_SZ and F_SETPIPE_SZ to get and set the capacity of a pipe.  This commit uses that functionality to implement the buffer size support in System.IO.Pipes.  (On OSX, we still ignore attempts to change the buffer size and throw a PlatformNotSupportedException when attempting to get it.)

Fixes #1730.

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
@@ -99,9 +99,7 @@ namespace System.IO.Pipes
                 TranslateFlags(_direction, _options, _inheritability), 
                 (int)Interop.libc.Permissions.S_IRWXU);
 
-            // Ignore _inBufferSize and _outBufferSize.  They're optional, and the fcntl F_SETPIPE_SZ for changing 
-            // a pipe's buffer size is Linux specific.
-
+            InitializeBufferSize(serverHandle, _outBufferSize); // there's only one capacity on Linux; just use the out buffer size
             InitializeHandle(serverHandle, isExposed: false, isAsync: (_options & PipeOptions.Asynchronous) != 0);
             State = PipeState.Connected;
         }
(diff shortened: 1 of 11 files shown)


stephentoub:
cc: @ericstj, @ellismg, @bpschoch

jasonwilliams200OK:
Does this also apply to FreeBSD? 
copy @josteink, @saper

josteink:
@jasonwilliams200OK 

I can't see those arguments mentioned or defined anywhere in `/usr/include` so I'm going to assume "no".

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
```diff
@@ -99,9 +99,7 @@ public void WaitForConnection()
                 TranslateFlags(_direction, _options, _inheritability), 
                 (int)Interop.libc.Permissions.S_IRWXU);
 
-            // Ignore _inBufferSize and _outBufferSize.  They're optional, and the fcntl F_SETPIPE_SZ for changing 
-            // a pipe's buffer size is Linux specific.
-
+            InitializeBufferSize(serverHandle, _outBufferSize); // there's only one capacity on Linux; just use the out buffer size
```
ellismg: What about using Math.Max(_outBufferSize, _inBufferSize) instead?

stephentoub: Interestingly, that's actually what I wrote initially :smile:  But then I hesitated because I wasn't actually sure the max would be desirable... maybe, for example, the user was actually trying to lower the size?  Since the out size is already what's being passed to the base (on Windows as well), I decided to just go with that.  Let me know if you still think I should change it to a max... I don't feel strongly about it either way, so if you think it's the right thing to do, just let me know.

PR #1950, opened 2015-06-07
Optimize ImmutableSortedSet<T>.LeafToRootRefill

### LINKED ISSUE #1943: Performance issue in ImmutableSortedSet.LeafToRootRefill
The implementation is using a SortedSet, and then build the balance tree based on it.  It should not use the SortedSet here.  Internally, the SortedSet will use a list to sort items (unless the input collection is a sortedSet), and convert it to an array, then build into its own balance tree.  Because the SortedSet doesn’t support indexers, the immutable collection code will convert the SortedSet to an array, then use it to build its own AVL tree.  Those conversions are totally wasted, including allocating the entire tree in SortedSet.  The code can directly sort items with a list.

stephentoub:
Some basic microbenchmarks show that this can double throughput and halve garbage production, depending on the inputs.

Fixes #1943.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
@@ -1107,6 +1107,16 @@ namespace System.Collections.Immutable
             List<T> list;
             if (this.IsEmpty)
             {
+                // If the additional items enumerable list is known to be empty, too,
+                // then just return this empty instance.
+                int count;
+                if (addedItems.TryGetCount(out count) && count == 0)
+                {
+                    return this;
+                }
+
+                // Otherwise, construct a list from the items.  The Count could still
+                // be zero, in which case, again, just return this empty instance.
                 list = new List<T>(addedItems);
                 if (list.Count == 0)
                 {
@@ -1115,6 +1125,9 @@ namespace System.Collections.Immutable
             }
             else
             {
+                // Build the list from this set and then add the additional items.
+                // Even if the additional items is empty, this set isn't, so we know
+                // the resulting list will not be empty.
                 list = new List<T>(this);
                 list.AddRange(addedItems);
             }
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
@@ -1102,10 +1102,39 @@ namespace System.Collections.Immutable
             // and can index into that sequence like a list, so the limited
             // garbage produced is a temporary mutable data structure we use
             // as a reference when creating the immutable one.
-            // The (mutable) SortedSet<T> is much faster at constructing its collection 
-            // when passed a sequence into its constructor than into its Union method.
-            var sortedSet = new SortedSet<T>(this.Concat(addedItems), this.KeyComparer);
-            Node root = Node.NodeTreeFromSortedSet(sortedSet);
+
+            // Produce the initial list containing all elements, including any duplicates.
+            List<T> list;
+            if (this.IsEmpty)
+            {
+                list = new List<T>(addedItems);
+                if (list.Count == 0)
+                {
+                    return this;
+                }
+            }
+            else
+            {
+                list = new List<T>(this);
+                list.AddRange(addedItems);
+            }
+            Debug.Assert(list.Count > 0);
+
+            // Sort the list and remove duplicate entries.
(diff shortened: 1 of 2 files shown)


stephentoub:
cc @AArnott, @lifengl

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
```diff
@@ -1102,10 +1102,39 @@ private ImmutableSortedSet<T> LeafToRootRefill(IEnumerable<T> addedItems)
...
+            else
+            {
+                list = new List<T>(this);
+                list.AddRange(addedItems);
+            }
+            Debug.Assert(list.Count > 0);
+
+            // Sort the list and remove duplicate entries.
+            IComparer<T> comparer = this.KeyComparer;
+            list.Sort(comparer);
```
sharwell: :exclamation: The combination of an unstable sort here with the use of a list of all elements when the current set is not empty has what I believe is a particularly bad side effect. Unlike methods using `ISet.UnionWith`, the behavior of `ImmutableSortedSet` with respect to duplicate entries is *unspecified*. The `ISet.UnionWith` method explicitly states that entries already in the collection take precedence in the face of duplicates to a `Union` operation.

stephentoub: This is not a change from the existing implementation.  If you believe ```SortedSet<T>``` should use a stable sort, please open a separate issue for that; I would not want to make such a change here.

sharwell: Added issue #1967 to address this concern separately.

stephentoub: Thanks.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
```diff
@@ -1102,10 +1102,39 @@ private ImmutableSortedSet<T> LeafToRootRefill(IEnumerable<T> addedItems)
...
-            // The (mutable) SortedSet<T> is much faster at constructing its collection 
-            // when passed a sequence into its constructor than into its Union method.
-            var sortedSet = new SortedSet<T>(this.Concat(addedItems), this.KeyComparer);
-            Node root = Node.NodeTreeFromSortedSet(sortedSet);
+
+            // Produce the initial list containing all elements, including any duplicates.
+            List<T> list;
+            if (this.IsEmpty)
+            {
+                list = new List<T>(addedItems);
```
AArnott: Tip: I think we have a TryGetCount extension method available we use to sniff the length of an `IEnumerable<T>` if possible in order to avoid work such as your `new List<T>` here where we can. You might consider using that here.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
```diff
@@ -1102,10 +1102,39 @@ private ImmutableSortedSet<T> LeafToRootRefill(IEnumerable<T> addedItems)
...
+            list.Sort(comparer);
+            int index = 1;
+            for (int i = 1; i < list.Count; i++)
+            {
+                if (comparer.Compare(list[i], list[i - 1]) != 0)
+                {
+                    list[index++] = list[i];
+                }
+            }
+            list.RemoveRange(index, list.Count - index);
```
AArnott: All of this math screams danger of 'off by 1' errors. While I suspect you got it all just right, is our testing of this method adequate to catch any errors that this method may have now, or in the future if anyone messes with this code?

AArnott:
LGTM if you feel the testing is already adequate. Otherwise I'd like to see some tests that try to exploit any off by 1 bugs your code may be vulnerable to.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
stephentoub: Done.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableSortedSet`1.cs
stephentoub: Done.

stephentoub:
Thanks, Andrew.  I added the use of TryGetCount and added some more tests, just in case.

AArnott:
:+1: Thanks

Issue #1739, opened 2015-05-13
Interop\Windows\BCrypt\Cng.cs does not follow the Interop guidelines


bartonjs:
Cng.cs was ported as-was from closed source into open. It should be broken up into multiple Interop.Method.cs and/or Interop.MethodCollection.cs files, and the SafeHandle types extracted into files of their own right.

robert-matusewicz:
@karelz @bartonjs - is there still a need to implement that tickete? If yes, I would be interested to do it.

Just to confirm that I understand that correctly - that the file that needs to adapt to guidelines is corefx/src/Common/src/Interop/Windows/BCrypt/Cng.cs

The work that needs to be done is:
- extract BCrypt* methods to separate files, so BCryptImportKey should be extracted to Interop.BCryptImportKey.cd, BCryptDecrypt to Interop.BCryptDecrypt, etc?
- handles should be extracted to their own files, so SafeAlgorithmHandle into SafeAlgorithmHandle .cs, SafeKeyHandle into SafeKeyHandle.cs

I'm not sure what do you mean by Interop.MethodCollection.cs - any example of how that should be done?

Cheers

karelz:
@krwq @GrabYourPitchforks can you provide some guidance while Jeremy is on vacation?

bartonjs:
While "one thing per file" can work, we've been accepting "one file per logical group".

* `BCRYPT_KEY_DATA_BLOB_HEADER` is only used with `BCryptImportKey`, so they could both be in one file (probably Interop.BCryptImportKey.cs).
* `BCryptEncrypt` and `BCryptDecrypt` are almost always needed together, so leave them in one file (Interop.BCryptEncryptDecrypt.cs?)
* Put the helper methods for the P/Invokes in the same file as the P/Invokes
* I'd probably put the extension methods `SetCipherMode` and `SetEffectiveKeyLength` with all of the SetProperty helpers and P/Invokes as Interop.BCryptSetProperty.cs.  If that requires an assembly to pull in SafeAlgorithmHandle when it otherwise didn't need to, then I'd come up with something else.

robert-matusewicz:
awesome, thank you for explaining that to me:) @karelz can you assign me to that ticket - I will be slowly working on it.

PR #1952, opened 2015-06-07
Separate BCryptGenRandom out of Cng.cs

stephentoub:
As part of #1739, this commit begins to refactor Cng.cs, starting by pulling out the functionality needed for System.Security.Cryptography.RandomNumberGenerator.  The assembly was being filled with all of the Cng code, when in reality it only needed a very small amount of it.  And since the assembly doesn't have any string resources, I also modified the .csproj to suppress the buildtools inclusion of the common resources support.

As a result of the removals, this change boosts the code coverage number of System.Security.Cryptography.RandomNumberGenerator.dll from 13% to 96%.

(Note that there are still copies of the code in Interop.NTSTATUS.cs in Cng.cs, due to other dependencies in Cng.cs that I didn't want to change.  As we refactor Cng.cs further, we can remove those copies.)

### DIFF
--- a/src/Common/src/Interop/Windows/BCrypt/Cng.cs
+++ b/src/Common/src/Interop/Windows/BCrypt/Cng.cs
@@ -129,14 +129,6 @@ namespace Internal.NativeCrypto
             }
         }
 
-        public static void BCryptGenRandom(byte[] buffer)
-        {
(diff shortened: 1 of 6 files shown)


stephentoub:
cc: @bartonjs

bartonjs:
LGTM. Thanks, Stephen.

PR #1955, opened 2015-06-08
Improve SortedSet performance

### LINKED ISSUE #1953: Constructing a SortedSet<T> from an enumerable with lots of dups is O(N^2)
The set's implementation iterates from start to end, comparing each element against the previous one.  Each time it finds a match, it calls list.RemoveAt to remove that element.  This results in the list needing to shift down all elements after the current one. There are some other inefficiencies, too, such as copying to temporary arrays.

stephentoub:
Primary changes:
- SortedSet's ctor has a loop that's O(D * N), where N is the number of elements being added to the set and D is the number of duplicates in those elements.  As the number of duplicates D grows, the time it takes to construct a sorted set grows polynomially, approaching O(N^2), which is particularly bad given that the purpose of a set is to help remove duplicates.  The core cause is a loop over the elements to determine whether element n is equal to n-1; if it is, element n is removed, but it's removed via a call to List<T>.RemoveAt, which will shift down all of the elements above it in the list.  A better implementation can make this O(N) instead; the ctor overall is still then O(N log N), due to the sort, but that's better than O(N^2).
- To sort the elements, the ctor builds a List<T> of the elements but then uses ToArray to get an array used to actually build the set; this incurs an unnecessary (and potentially large) array allocation and copy.  We can avoid that entirely by just building up the array manually using the same logic that List<T> does.  That same logic is actually duplicated in multiple places, including in Stack<T>, so I've extracted it as a helper to use in both places (there are a few other places in corefx that use the same logic, hence I put the helper into a Common file, but I did not condense those other uses as part of this commit).  This then further helps with Stack<T> perf in a few corner cases, such as when a Stack<T> is constructed with an empty source. (It's not clear to me why Queue<T> doesn't have the same IColection<T>-related logic; probably just an oversight, but I didn't want to change its behavior with the additional ICollection<T> check.)
- SortedSet's BreadthFirstTreeWalk (which is used by several members, such as RemoveWhere) is currently using a List<T> as a queue, doing a RemoveAt(0) to dequeue, which will incur the cost of shifting all elements down.  Using a Queue<T> is both cleaner and better performing.
- As I was touching Stack<T> and Queue<T>, I noticed that they had an empty array static field, which was just being initialized to Array.Empty<T>().  The field is unnecessary, and the call site can just use Array.Empty<T> directly.

Fixes #1953.

### DIFF
--- a/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
+++ b/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Collections.Generic
+{
+    /// <summary>Internal helper functions for working with enumerables.</summary>
+    internal static class EnumerableHelpers
+    {
+        /// <summary>Converts an enumerable to an array using the same logic as does List{T}.</summary>
+        /// <param name="length">The number of items stored in the resulting array, 0-indexed.</param>
+        /// <returns>
+        /// The resulting array.  The length of the array may be greater than <paramref name="length"/>,
+        /// which is the actual number of elements in the array.
+        /// </returns>
+        internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
+        {
+            T[] arr;
+            int count = 0;
+
+            ICollection<T> ic = source as ICollection<T>;
+            if (ic != null)
+            {
+                count = ic.Count;
+                if (count == 0)
+                {
+                    arr = Array.Empty<T>();
+                }
+                else
+                {
+                    // Allocate an array of the desired size, then copy the elements into it. Note that this has the same 
+                    // issue regarding concurrency as other existing collections like List<T>. If the collection size 
+                    // concurrently changes between the array allocation and the CopyTo, we could end up either getting an 
+                    // exception from overrunning the array (if the size went up) or we could end up not filling as many 
+                    // items as 'count' suggests (if the size went down).  This is only an issue for concurrent collections 
+                    // that implement ICollection<T>, which as of .NET 4.6 is just ConcurrentDictionary<TKey, TValue>.
+                    arr = new T[count];
+                    ic.CopyTo(arr, 0);
+                }
+            }
+            else
+            {
+                arr = Array.Empty<T>();
+                foreach (var item in source)
+                {
+                    if (count == arr.Length)
+                    {
+                        // MaxArrayLength is defined in Array.MaxArrayLength and in gchelpers in CoreCLR.
+                        // It represents the maximum number of elements that can be in an array where
+                        // the size of the element is greater than one byte; a separate, slightly larger constant,
+                        // is used when the size of the element is one.
+                        const int MaxArrayLength = 0x7FEFFFFF;
+
+                        // This is the same growth logic as in List<T>:
+                        // If the array is currently empty, we make it a default size.  Otherwise, we attempt to 
+                        // double the size of the array.  Doubling will overflow once the size of the array reaches
+                        // 2^30, since doubling to 2^31 is 1 larger than Int32.MaxValue.  In that case, we instead 
+                        // constrain the length to be MaxArrayLength (this overflow check works because of of the 
+                        // cast to uint).  Because a slightly larger constant is used when T is one byte in size, we 
+                        // could then end up in a situation where arr.Length is MaxArrayLength or slightly larger, such 
+                        // that we constrain newLength to be MaxArrayLength but the needed number of elements is actually 
+                        // larger than that.  For that case, we then ensure that the newLength is large enough to hold 
+                        // the desired capacity.  This does mean that in the very rare case where we've grown to such a 
+                        // large size, each new element added after MaxArrayLength will end up doing a resize.
+                        const int DefaultCapacity = 4;
+                        int newLength = count == 0 ? DefaultCapacity : count * 2;
+                        if ((uint)newLength > MaxArrayLength)
+                        {
+                            newLength = MaxArrayLength;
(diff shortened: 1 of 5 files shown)


stephentoub:

ellismg:
Aside from the naming of the local, the change looks good to me.

src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -32,14 +32,13 @@ public class Queue<T> : IEnumerable<T>,
...
         private const int DefaultCapacity = 4;
-        private static T[] s_emptyArray = Array.Empty<T>();
 
         // Creates a queue with room for capacity objects. The default initial
         // capacity and grow factor are used.
         /// <include file='doc\Queue.uex' path='docs/doc[@for="Queue.Queue"]/*' />
         public Queue()
         {
-            _array = s_emptyArray;
+            _array = Array.Empty<T>();
```
sharwell: :bulb: this should probably have been part of a different commit.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -0,0 +1,64 @@
...
+                foreach (var item in source)
+                {
+                    if (found == arr.Length)
+                    {
+                        const int DefaultCapacity = 4;
+                        const int MaxArrayLength = 0x7FEFFFFF; // from Array.MaxArrayLength
+
+                        int newLength = found == 0 ? DefaultCapacity : found * 2;
+                        if ((uint)newLength > MaxArrayLength) newLength = MaxArrayLength;
+                        if (newLength < found + 1) newLength = found + 1;
```
sharwell: :bulb: I believe the statement here should be on a separate line.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                        arr = ArrayT<T>.Resize(arr, newLength, found);
```
sharwell: Can you not use `Array.Resize`?

sharwell:
No major problems stood out to me. I didn't comment on some minor items and had a question.

I do wish this was separated into one commit (or perhaps even one PR) per topic, but I say this as a reminder for the future and not as a reason to reject *this* PR.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
sharwell: :exclamation: The only way this condition can happen is if `found == MaxArrayLength`. There are two possibilities for the behavior here:

1. The value of `MaxArrayLength` is correct, and the call to `Resize` throws an exception`
2. The value of `MaxArrayLength` is *not* correct, and the loop degrades immediately from amortized O(n) to O(n²).

Neither of these behaviors is obvious by reading the code. Rather than leave things up in the air, I believe it would be better to simply use the following single `if` statement to handle bounds:

```cs
if ((uint)newLength > MaxArrayLength)
{
  if (found == MaxArrayLength)
    throw new OutOfMemoryException(...);

  newLength = MaxArrayLength;
}
```

sharwell:
I take that back... I'm concerned about the current `MaxArrayLength` behavior. See my comment in the diff for details.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
stephentoub: The value of MaxArrayLength may not actually be correct, and not just because its value could change over time, but because there's a higher value used when T is one byte in size (search coreclr for MaxByteArrayLength).  However, I do not want this code to be in the business of explicitly setting some hard-and-fast limit and explicitly throwing an OutOfMemoryException, which is why I chose to exactly copy the logic that's been in use in ``List<T>``` for a long time (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L402-L406).  If we decide that's the wrong logic, we should update ```List<T>``` and keep this in sync with it, but for now I plan to leave this as-is.  I do realize that for the last few elements between MaxArrayLength and MaxByteArrayLength, it'll end up degrading in performance.  I can add a comment explaining what's going on, to address your concerns around obviousness.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
stephentoub: The corefx codebase is currently inconsistent about whether it uses ```Array.Resize``` or ```ArrayT<T>.Resize```, the latter of which was added for performance reasons with .NET Native.  I chose to use the latter for now, as it's what's used elsewhere in System.Collections.  We do separately need to rationalize its use across the codebase, and ideally get whatever additional methods or improvements are necessary into the real Array rather than relying on a separate method like this.

src/System.Collections/src/System/Collections/Generic/SortedSet.cs
```diff
@@ -138,23 +134,27 @@ public SortedSet(IEnumerable<T> collection, IComparer<T> comparer)
...
+                        }
                     }
+                    count = index;
+
+                    _root = ConstructRootFromSortedArray(els, 0, count - 1, null);
+                    _count = count;
                 }
-                _root = ConstructRootFromSortedArray(els.ToArray(), 0, els.Count - 1, null);
-                _count = els.Count;
-                _version = 0;
```
AlexGhiondea: You no longer need to set the version?

sharwell: This is the constructor. Setting `_version` to 0 here is redundant because the CLR initializes all fields to their default values.

stephentoub:
Thanks. Updated.

stephentoub:
@dotnet-bot test this please

PR #1957, opened 2015-06-08
Remove empty .resx files

stephentoub:
These were causing the build system to generate SR gook even though the assembly never uses it.  I searched and these appear to be the only such empty .resx files currently in the corefx repo.


stephentoub:
cc: @tarekgh, @bartonjs

bartonjs:
LGTM

tarekgh:
LGTM too.

stephentoub:
Thanks, guys.

PR #1960, opened 2015-06-08
Fixes bug 1958

KrzysztofCwalina:
Another remaining test for sameness that we agreed to remove (i.e. cost to synchronize too high
in comparison to the benefit)

There are a few other tests like that (they all test sameness of codepointfilter creation).
I will review them as a separate workitem as they are probably also not very reliable. But for now,
I am removing this test as it is actually failing during runs.

### DIFF
--- a/src/System.Text.Encodings.Web/tests/UnicodeHelpersTests.cs
+++ b/src/System.Text.Encodings.Web/tests/UnicodeHelpersTests.cs
@@ -21,17 +21,6 @@ namespace Microsoft.Framework.WebEncoders
         private static readonly UTF8Encoding _utf8EncodingThrowOnInvalidBytes = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
 
         [Fact]
-        public void GetDefinedCharacterBitmap_ReturnsSingletonInstance()
-        {
-            // Act
-            uint[] retVal1 = UnicodeHelpers.GetDefinedCharacterBitmap();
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
BTW, I filed #1961 to track holistic review of sameness tests in this tests suite.

src/System.Text.Encodings.Web/tests/UnicodeHelpersTests.cs
```diff
@@ -21,17 +21,6 @@ public unsafe class UnicodeHelpersTests
         private static readonly UTF8Encoding _utf8EncodingThrowOnInvalidBytes = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
 
         [Fact]
-        public void GetDefinedCharacterBitmap_ReturnsSingletonInstance()
```
stephentoub: I don't see any other calls to GetDefinedCharacterBitmap in this file.  Are there tests for it somewhere else?

stephentoub:
LGTM

src/System.Text.Encodings.Web/tests/UnicodeHelpersTests.cs
KrzysztofCwalina: It's an internal method. It's tested indirectly in many tests.

stephentoub: Ah, ok.

PR #1970, opened 2015-06-09
Add missing null-argument check to System.Linq.Parallel.Cast().

Clockwork-Muse:
Add missing null-argument check.
Update tests as necessary.
This makes Cast act the same as the other extension methods, and follow its documented behavior.

### DIFF
--- a/src/System.Linq.Parallel/src/System/Linq/ParallelEnumerable.cs
+++ b/src/System.Linq.Parallel/src/System/Linq/ParallelEnumerable.cs
@@ -5363,6 +5363,8 @@ namespace System.Linq
         /// </exception>
         public static ParallelQuery<TResult> Cast<TResult>(this ParallelQuery source)
         {
+            if (source == null) throw new ArgumentNullException("source");
+
             return source.Cast<TResult>();
         }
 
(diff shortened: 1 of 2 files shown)


Clockwork-Muse:
@stephentoub , @AlfredoMS - separate PR for updating how PLINQ deals with null arguments for `Cast()`

stephentoub:
@weshaggard, this is turning a NullReferenceException into an ArgumentNullException. I'm ok with this, as it makes Cast more consistent with the rest of the PLINQ APIs and with LINQ's Cast, but it is technically a breaking change.  Any concerns?

Clockwork-Muse:
It looks like test failures are coming from System.IO.Pipes.  I don't think it's related (nothing in the path seems to be using PLINQ), especially as it seems to be claiming that a coreclr type can't be loaded (I'm reading that as "class loaded")?  Also, other PRs are showing it, so...

stephentoub:
No, it's not related:
https://github.com/dotnet/corefx/issues/1972

weshaggard:
@stephentoub It is technically a breaking change but people should never really be handling NullRefrerenceException anyway and in general I don't like our code NullRef'ing. In either case this is a bug in the callers code which they will need to fix so I'm OK with this change.

stephentoub:
Cool.  Thanks, Wes.  LGTM.

stephentoub:
@dotnet-bot test this please

PR #1978, opened 2015-06-09
Avoid InvalidCastException from the IList implementation of ImmutableArray<T>

dattathreya:
1. Crib the IsCompatibleObject predicate from List
2. Make IList.Contains and IList.Index call IsCompatibleObject to decide
whether the value to search for is compatible with T
3. Fix typos in comments

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -793,24 +793,18 @@ namespace System.Collections.Immutable
             Requires.NotNull(items, "items");
             Requires.NotNull(equalityComparer, "equalityComparer");
 
-            List<int> indexesToRemove = null;
-            for (int i = 0; i < self.Length; i++)
+            var indexesToRemove = new SortedSet<int>();
+            foreach (var item in items)
             {
-                foreach (var item in items)
+                int index = self.IndexOf(item, equalityComparer);
+                while (index >= 0 && !indexesToRemove.Add(index) && index + 1 < self.Length)
                 {
-                    T elem = self.array[i];
-                    if (equalityComparer.Equals(item, elem))
-                    {
-                        indexesToRemove = indexesToRemove ?? new List<int>();
-                        indexesToRemove.Add(i);
-                        break;
-                    }
+                    // This is a duplicate of one we've found. Try hard to find another instance in the list to remove.
+                    index = self.IndexOf(item, index + 1, equalityComparer);
                 }
             }
 
-            return indexesToRemove != null ?
-                self.RemoveAtRange(indexesToRemove) :
-                self;
+            return self.RemoveAtRange(indexesToRemove);
         }
 
         /// <summary>
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -1286,6 +1286,13 @@ namespace System.Collections.Immutable
             return self.Replace(oldValue, newValue, equalityComparer);
         }
 
+        private static bool IsCompatibleObject(object value)
+        {
+            // Non-null values are fine.  Only accept nulls if T is a class or Nullable<U>.
+            // Note that default(T) is not equal to null for value types except when T is Nullable<U>. 
+            return ((value is T) || (value == null && default(T) == null));
+        }
+
         /// <summary>
         /// Adds an item to the <see cref="IList"/>.
         /// </summary>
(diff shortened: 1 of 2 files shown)


mellinoe:
I believe we made some decisions regarding this a while back and leaned on the side of performance for these cases (rather than checking things to avoid exceptions). I may be mis-remembering, though.

@nguerrera

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -1052,7 +1052,7 @@ public static ImmutableArray<T> CastUp<TDerived>(ImmutableArray<TDerived> items)
         [Pure]
         public ImmutableArray<TOther> CastArray<TOther>() where TOther : class
         {
-            return new ImmutableArray<TOther>((TOther[])(object)array);
+            return new ImmutableArray<TOther>((TOther[])(object)this.array);
```
sharwell: :bulb: I don't see why this needed to change. Seems better to leave it the way it was.

sharwell:
@mellinoe Performance is not a concern in these methods, which explicitly implement the legacy non-generic `IList` interface.

My big concern is the following:

:exclamation: There are four commits. The first (80ec16bb2320364f3a7b5181916aa7511045c52f) introduces certain errors, and they are reverted by the third (2f0cabb34fa2d17262c83d1dc2db10a432eb838d). I definitely want to see all four squashed before they are merged.

nguerrera:
I agree with @sharwell re: performance. The perf hit is fine through the non-generic IList. We should be consistent with other collections on this.

Regarding squashing, please do eliminate the false starts + reverts. However, I would move the unrealted doc comment and `this.` consistency changes out to separate commits.

LGTM apart from that. @AArnott, @FiveTimesTheFun: thoughts?

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
nguerrera: This seems like a good maintenance change in keeping with the comment here: https://github.com/dotnet/corefx/pull/1978/files#diff-c4e56d2c84bd6553b00763a60e46ebd0R21

It is unrelated, though, so I'd split it out to its own commit. Same for the doc comment fixes above it.

dattathreya:
> There are four commits. The first (80ec16b) introduces certain errors, and they are reverted by the  third (2f0cabb).

Apologies for the commits and reverts. I am unfamiliar with github, and did not know that experimental commits to my private branch are visible publicly.
>  I definitely want to see all four squashed before they are merged.

I will separate this into two changesets and resubmit.

sharwell:
> Apologies for the commits and reverts. I am unfamiliar with github, and did not know that experimental commits to my private branch are visible publicly.

@dattathreya Actually I'm sorry. Looking back over my replies, I see they were poorly worded (too harsh). I couldn't be happier that you decided to make your first GitHub contribution a pull request to this project. The issue you identified (the invalid casts) obviously needs to be fixed, and I expect your pull request will be readily incorporated when the concerns are addressed.

If you have any trouble updating this pull request, please add a comment or ask in the Gitter room. We're all in this together and many people would be happy to help you out. :+1:

jasonwilliams200OK:
@dattathreya, 
```powershell
cd into\corefx\directory
git rebase -i HEAD~4

# notepad will open. On first line, replace the word 'pick' with 'reword'
# On second, and third line, replace the word 'pick' with 'fixup'.
# save and close notepad
# another notepad will open to let you reword the commit
# you can replace the commit log with the following (note
# that line breaks are important):
#
# /Commit log start (not part of log itself)
#
# IList: Avoid InvalidCastException
#
# 1. ImmutableArray.cs: fence post error in condition for
#   `Requires.Range`.
#
# 2. ImmutableArray\`1.cs:
#  i. Fixes typos in comments.
#  ii. RemoveRange() speedups.
#  iii. IndexOf does not throw if argument cannot be cast to `T`.
#
# /Commit log ends here (not part of commit log itself)
#
# save and close notepad

git push -f
```

nguerrera:
@dotnet-bot test this please

nguerrera:
One more thing I forgot to mention above: please add test cases for this change. 

Also, thanks for this contribution, it's a good catch. :)

AArnott:
The problem looks like a good one to solve. The solution is not too complicated but I'm just a touch surprised it was necessary -- but I can't think of any other solution so I guess it's good. 

But yes as @nguerrera  said, we definitely need tests for this. Please be creative and add several test cases including various combinations of structs, nullables, and ref types. Basically everything you had in your head when you coded up all those if conditions and compound expressions, we want tests to hit every path. Then think of a couple more tests you haven't thought of yet and try to break your changes. Bonus points if you break it and fix them. :) I'd love to see the commits survive through to merge that show the evolution of the code (e.g. commit 1: code, commit 2: add test, commit 3: fix code to pass after new tests found flaws.). Squashing away history when it's just reverts makes some sense, but I like to see good history like that preserved. :)

dattathreya:
Thank you for your help and feedback. Please reject this patch. I will create and resubmit one patch for the documentation changes and another for the IList implementation changes and unit tests.

stephentoub:
> Please reject this patch. I will create and resubmit one patch for the documentation changes and another for the IList implementation changes and unit tests.

Ok, thanks.

PR #1980, opened 2015-06-09
Disabling System.Diagnostics.Process.MainModule test against Issue#1896

pallavit:
Process class uses Win32 API EnumProcessModules to get the process's MainModule. This API is a best effort API which may not work in certain scenarios. I am debating over whether we should remove this dependency completely so for now, I am simply disabling the test against Issue#1896 while I gather more data.


stephentoub:
LGTM

pallavit:
@stephentoub Thanks for the quick review :).

PR #1988, opened 2015-06-10
Mark project.lock.json files as locked:true

stephentoub:
Whenever I do a full build, a handful of project.lock.json files are getting modified due to their locked value being false.  I've changed them to be true.

### DIFF
--- a/src/Common/tests/SystemXml/BaseLibManaged/project.lock.json
+++ b/src/Common/tests/SystemXml/BaseLibManaged/project.lock.json
@@ -3,14 +3,14 @@
   "version": -9997,
   "targets": {
     "DNXCore,Version=v5.0": {
-      "System.Private.Uri/4.0.0-beta-22927": {
+      "System.Private.Uri/4.0.0-beta-23008": {
         "runtime": [
           "lib/DNXCore50/System.Private.Uri.dll"
         ]
       },
-      "System.Runtime/4.0.20-beta-22927": {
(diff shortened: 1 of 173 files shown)


stephentoub:
cc: @ericstj

ericstj:
This is happening because folks are checking in project.lock.json files without setting locked=true.  I'm tempted to make that a build error.  /cc @venkat-raman251 @Priya91 @bartonjs 

Instead of this can you just run build.cmd /t:RestorePackage /p:LockDependencies=true from the root?  This will update all dependencies to a consistent package version and lock the lock files.  That way we don't have a hodge-podge of versions in the repo.

stephentoub:
> Instead of this can you just run build.cmd /t:RestorePackages /p:LockDependencies=true from the root?

Done.

> That way we don't have a hodge-podge of versions in the repo.

Won't the inconsistencies grow as new libraries/test projects are brought into the repo (and concurrently versions are updated in the myget feeds)?  Do we have a system in place for continuing to ensure these remain consistent?

bartonjs:
@ericstj Could you point me to where we have the guidance to change it to locked, in case there's anything else that I've missed with porting in these new libraries?

ericstj:
@bartonjs where would you expect to find the docs?   Point me and I will make sure to doc it.  

Initially I outlined the process in an issue, but that's probably not completely up to date. This has mostly only been discussed in PRs and then folks have been watching for it when they review incoming PRs.  I agree that this is not the best place to have such guidance.

PR #1989, opened 2015-06-10
Remove [ActiveIssue] for ILGenerator.EmitWriteLine test

### LINKED ISSUE #1915: Reflection.EmitWriteLine fails on Linux
```
System.Reflection.Emit.ILGeneration.Tests.ILGeneratorEmitWriteLine.PosTest1 [FAIL]
17:59:28       System.ArgumentNullException : Value cannot be null.
17:59:28       Parameter name: meth
17:59:28       Stack Trace:
17:59:28             at System.Reflection.Emit.ILGenerator.Emit(OpCode opcode, MethodInfo meth)
17:59:28             at System.Reflection.Emit.ILGenerator.EmitWriteLine(FieldInfo fld)
17:59:28             at System.Reflection.Emit.ILGeneration.Tests.ILGeneratorEmitWriteLine.PosTest1()
17:59:28 Finished:    System.Reflection.Emit.ILGeneration.Tests
```

Quoting @stephentoub 

> The EmitWriteLine method in mscorlib is using reflection to access Console.Out in mscorlib (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/ILGenerator.cs#L1233), but Console.Out is getting removed by the BCL rewriter, so the MethodInfo is null, causing the ArgumentNullException. But even if it wasn't getting removed, this is not the right Console implementation for code to be targeting: the Console in System.Console.dll is the one that's cross-platform. @jkotas, FYI. (I wonder if the right solution is just to remove these EmitWriteLine methods from the contract and not expose them.)

stephentoub:
Fixed by https://github.com/dotnet/coreclr/pull/1126

Resolves #1915

### DIFF
--- a/src/System.Reflection.Emit.ILGeneration/tests/ILGenerator/ILGeneratorEmitWriteLine.cs
+++ b/src/System.Reflection.Emit.ILGeneration/tests/ILGenerator/ILGeneratorEmitWriteLine.cs
@@ -11,7 +11,7 @@ namespace System.Reflection.Emit.ILGeneration.Tests
 {
     public class ILGeneratorEmitWriteLine
     {
-        [Fact, ActiveIssue(1915, PlatformID.AnyUnix)]
+        [Fact]
         public void PosTest1()
         {
             string name = "Assembly1";


stephentoub:
cc: @Priya91

Issue #1926, opened 2015-06-04
Open Source System.Security.Cryptography.X509Certificates


bartonjs:

bartonjs:
This was resolved with https://github.com/dotnet/corefx/pull/1990.

Issue #1986, opened 2015-06-10
RSA/X509Certificates tests are segfaulting on Unix


bartonjs:
Unfortunately it's hard to find what test/tests is/are failing due to the segfault.

If the tests would have the decency to fail on my machine the same way as they do in the CI build I'd be able to catch it in a debugger; but as it is the RSA tests are passing locally :smile:.

Since this failure is blocking opening up these libraries, I'm disabling the tests on Linux/OSX temporarily.

ellismg:
@mmitche Can get you access to one of the CI machines if you need it to investigate.

bartonjs:
It seems that my local environment was missing Xunit.NetCore,Extensions.dll, and (our build of) xUnit stops test discovery when it hits an attribute it doesn't recognize.

I have regained a local repro of the segfault, so investigations are now continuing.  (And I almost have the X509 tests appropriately tagged)

Issue #1977, opened 2015-06-09
Add test assets package for X509Certificates tests, and enable the tests


bartonjs:
The file-load-based tests for X509Certificates are disabled pending having assets to load.

bartonjs:
This was resolved by https://github.com/dotnet/corefx/pull/2017.

PR #1990, opened 2015-06-10
Add System.Security.Cryptography.X509Certificates Source and Test

bartonjs:
Resolves https://www.github.com/dotnet/corefx/issues/1926.


bartonjs:
Built with LockDependencies, so it shouldn't re-introduce the ever-changing project.lock.json problem.

Linux/OSX tests are disabled per #1986, and a few Windows tests are disabled per #1977.

stephentoub:
Two failures on Windows:
```
System.Security.Cryptography.X509Certificates.Tests.PropsTests.TestNotAfter [FAIL]
        Assert.Equal() Failure
        Expected: 2014-04-24T15:33:39.0000000
        Actual:   2014-04-24T22:33:39.0000000+00:00
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Security.Cryptography.X509Certificates\tests\PropsTests.cs(136,0): at System.Security.Cryptography.X509Certificates.Tests.PropsTests.TestNotAfter()
     System.Security.Cryptography.X509Certificates.Tests.PropsTests.TestNotBefore [FAIL]
        Assert.Equal() Failure
        Expected: 2013-01-24T14:33:39.0000000
        Actual:   2013-01-24T22:33:39.0000000+00:00
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Security.Cryptography.X509Certificates\tests\PropsTests.cs(125,0): at System.Security.Cryptography.X509Certificates.Tests.PropsTests.TestNotBefore()
```

bartonjs:
Were those tests only capable of passing when run in a local timezone of UTC-08:00?  Yes, I believe they were.

stephentoub:
:smile:

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
```diff
@@ -0,0 +1,119 @@
...
+                sb.AppendFormat("{0:X2}", b);
+            }
+            return sb.ToString();
+        }
+
+        // Decode a hex string-encoded byte array passed to various X509 crypto api. The parsing rules are overly forgiving but for compat reasons,
+        // they cannot be tightened.
+        public static byte[] DecodeHexString(this String s)
+        {
+            String hexString = DiscardWhiteSpaces(s);
```
stephentoub: DiscardWhiteSpace does a bunch of allocations.  Could DecodeHexString in its for loop instead just skip over whitespace characters?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
```diff
@@ -0,0 +1,119 @@
...
+            int i, iCount = 0;
+            for (i = 0; i < inputCount; i++)
+            {
+                if (Char.IsWhiteSpace(inputBuffer[inputOffset + i]))
+                {
+                    iCount++;
+                }
+            }
+
+            char[] rgbOut = new char[inputCount - iCount];
```
stephentoub: If iCount is 0, inputOffset is 0, and inputCount is inputBuffer.Length, could we just return inputBuffer?

stephentoub: Although it looks like the only place DiscardWhiteSpaces is used is in DecodeHexString, so if we could get rid of that dependency as noted earlier, this won't matter :smile:

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public static ICertificatePal FromBlob(byte[] rawData, string password, X509KeyStorageFlags keyStorageFlags)
+        {
+            return new OpenSslX509CertificateReader(rawData);
+        }
+
+        public static ICertificatePal FromFile(string fileName, string password, X509KeyStorageFlags keyStorageFlags)
+        {
+            throw new NotImplementedException();
```
stephentoub: I assume these NotImplementedExceptions are all essentially TODOs? (Otherwise they'd be PlatformNotSupportedException?)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -0,0 +1,416 @@
...
+        public string Subject
+        {
+            get { return SubjectName.Name; }
+        }
+
+        public byte[] Thumbprint
+        {
+            get
+            {
+                int negativeSize = Interop.NativeCrypto.GetX509Thumbprint(_cert, null, 0);
```
stephentoub: This pattern looks very much like the GetAsn1StringBytes where the actual P/Invoke is private and a wrapper is provided in the interop type.  Should the same be done for this in order to be consistent?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -0,0 +1,416 @@
...
+                IntPtr oidPtr = Interop.NativeCrypto.GetX509PublicKeyAlgorithm(_cert);
+                return Interop.libcrypto.OBJ_obj2txt_helper(oidPtr);
+            }
+        }
+
+        public byte[] KeyAlgorithmParameters
+        {
+            get
+            {
+                int negativeLen = Interop.NativeCrypto.GetX509PublicKeyParameterBytes(_cert, null, 0);
```
stephentoub: Ditto.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -0,0 +1,416 @@
...
+        }
+
+        private static X500DistinguishedName LoadX500Name(IntPtr namePtr)
+        {
+            if (namePtr == IntPtr.Zero)
+            {
+                throw new CryptographicException();
+            }
+
+            int negativeSize = Interop.NativeCrypto.GetX509NameRawBytes(namePtr, null, 0);
```
stephentoub: ditto.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
```diff
@@ -0,0 +1,343 @@
...
+        public byte[] EncodeX509SubjectKeyIdentifierExtension(byte[] subjectKeyIdentifier)
+        {
+            throw new NotImplementedException();
+        }
+
+        public unsafe void DecodeX509SubjectKeyIdentifierExtension(byte[] encoded, out byte[] subjectKeyIdentifier)
+        {
+            using (SafeAsn1OctetStringHandle octetString = Interop.libcrypto.OpenSslD2I(Interop.libcrypto.d2i_ASN1_OCTET_STRING, encoded))
+            {
+                subjectKeyIdentifier = Interop.NativeCrypto.GetAsn1StringBytes(octetString.DangerousGetHandle());
```
stephentoub: Are we guaranteed here that octedString is a valid handle?  DangerousGetHandle is, well, "dangerous" and doesn't do any check.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
```diff
@@ -0,0 +1,343 @@
...
+            return outFlags;
+        }
+
+        private static unsafe RSA BuildRsaPublicKey(byte[] encodedData)
+        {
+            using (SafeRsaHandle rsaHandle = Interop.libcrypto.OpenSslD2I(Interop.libcrypto.d2i_RSAPublicKey, encodedData))
+            {
+                if (rsaHandle.IsInvalid)
+                {
+                    throw new CryptographicException(Interop.libcrypto.GetOpenSslErrorString());
```
stephentoub: This pattern:
```
if (someSafeHandle.IsInvalid)
{
    throw new CryptographicException(Interop.libcrypto.GetOpenSslErrorString());
}
```
shows up a lot.  Should we move it into its own function?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/CertificatePal.cs
```diff
@@ -0,0 +1,538 @@
...
+                    cspKeyContainerInfo = new CspKeyContainerInfo(parameters);
+                }
+            }
+            // We could not access the key container. Just return.
+            catch (CryptographicException) { }
+
+            if (cspKeyContainerInfo == null)
+                return;
+
+            sb.Append(Environment.NewLine + Environment.NewLine + "[Private Key]");
```
stephentoub: So many unnecessary interim string allocations :(  No idea how much this function matters perf-wise, but at some point it'd be good to break these apart to avoid the extra allocations.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
bartonjs: Yep. Moved the work item to https://github.com/dotnet/corefx/issues/1993.

Some of them may end up being PlatformNotSupportedException, the NIE being a quantum state of uncertainty.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Interop.Libraries.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Internal.Cryptography.Pal.Native
+{
+    internal static partial class Interop
+    {
+        private static class Libraries
```
stephentoub: Why do we have this Libraries rather than using the main one in Common\src\Interop\Windows?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Interop.localization.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Internal.Cryptography.Pal.Native
+{
+    internal static partial class Interop
+    {
+        public static class localization
+        {
+            [DllImport(Libraries.Localization, CharSet = CharSet.Unicode, SetLastError = true, EntryPoint = "FormatMessageW")]
```
stephentoub: Don't we already have a FormatMessage (and associated wrapper) in Common?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Interop.crypt32.cs
```diff
@@ -0,0 +1,286 @@
...
+using CryptographicException = System.Security.Cryptography.CryptographicException;
+
+using SafeX509ChainHandle = Microsoft.Win32.SafeHandles.SafeX509ChainHandle;
+using X509KeyUsageFlags = System.Security.Cryptography.X509Certificates.X509KeyUsageFlags;
+
+namespace Internal.Cryptography.Pal.Native
+{
+    internal static partial class Interop
+    {
+        public static class crypt32
```
stephentoub: I assume, as with some of the other crypto libs, that we plan to but are deferring rationalizing this with how we're doing interop files elsewhere in corefx?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Primitives.cs
```diff
@@ -0,0 +1,773 @@
...
+
+    // CRYPTOAPI_BLOB has many typedef aliases in the C++ world (CERT_BLOB, DATA_BLOB, etc.) We'll just stick to one name here.
+    [StructLayout(LayoutKind.Sequential)]
+    internal unsafe struct CRYPTOAPI_BLOB
+    {
+        public CRYPTOAPI_BLOB(int cbData, byte* pbData)
+        {
+            this.cbData = cbData;
+            this.pbData = pbData;
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            get
+            {
+                return handle == IntPtr.Zero;
+            }
+        }
+
+        public static T InvalidHandle
+        {
+            get { return new T(); }
```
stephentoub: We don't want to cache an InvalidHandle?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/StorePal.Find.cs
```diff
@@ -0,0 +1,553 @@
...
+                throw new CryptographicException(Marshal.GetLastWin32Error());
+
+            return sb.ToString();
+        }
+
+        private static BigInteger PositiveBigIntegerFromByteArray(byte[] bytes)
+        {
+            // To prevent the big integer from misinterpreted as a negative number, add a "leading 0" to the byte array.
+            // Since BigInteger(bytes[]) requires a little-endian byte array, the "leading 0" actually goes at the end of the array.
+            byte[] newBytes = new byte[bytes.Length + 1];
```
stephentoub: Are there any quick checks that could be done on ```bytes``` to determine whether it would be interpreted as negative?  That would let us avoid the unnecessary array allocation and copy.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/X509Pal.CustomExtensions.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            int numUsages;
+            using (SafeHandle usagesSafeHandle = usages.ToLpstrArray(out numUsages))
+            {
+                unsafe
+                {
+                    CERT_ENHKEY_USAGE enhKeyUsage = new CERT_ENHKEY_USAGE()
+                    {
+                        cUsageIdentifier = numUsages,
+                        rgpszUsageIdentifier = (IntPtr*)(usagesSafeHandle.DangerousGetHandle()),
```
stephentoub: Same question as earlier around whether the safe handle could be invalid here.

src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            return ChainPal.ReleaseSafeX509ChainHandle(handle);
+        }
+
+        internal static SafeX509ChainHandle InvalidHandle
+        {
+            get { return new SafeX509ChainHandle(); }
```
stephentoub: Caching of InvalidHandle?  (Will stop asking about these.)

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X500DistinguishedName.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        private static void ThrowIfInvalid(X500DistinguishedNameFlags flags)
+        {
+            // All values or'ed together. Change this if you add values to the enumeration.
+            uint allFlags = 0x71F1;
+            uint dwFlags = (uint)flags;
+            if ((dwFlags & ~allFlags) != 0)
+                throw new ArgumentException(String.Format(CultureInfo.CurrentCulture, SR.Arg_EnumIllegalVal, "flag"));
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509BasicConstraintsExtension.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                return _pathLenConstraint;
+            }
+        }
+
+        public override void CopyFrom(AsnEncodedData asnEncodedData)
+        {
+            base.CopyFrom(asnEncodedData);
+            _decoded = false;
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
```diff
@@ -0,0 +1,365 @@
...
+using Internal.Cryptography;
+using Internal.Cryptography.Pal;
+
+namespace System.Security.Cryptography.X509Certificates
+{
+    public class X509Certificate : IDisposable
+    {
+        public X509Certificate()
+        {
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
```diff
@@ -0,0 +1,365 @@
...
+                if (subject == null)
+                    subject = _lazySubject = Pal.Subject;
+                return subject;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
```diff
@@ -0,0 +1,365 @@
...
+
+        public virtual String ToString(bool fVerbose)
+        {
+            if (fVerbose == false || Pal == null)
+                return GetType().ToString();
+
+            StringBuilder sb = new StringBuilder();
+
+            // Subject
+            sb.Append("[Subject]" + Environment.NewLine + "  ");
```
stephentoub: Same comment about interim string allocations.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
```diff
@@ -0,0 +1,365 @@
...
+            if (notBefore == DateTime.MinValue)
+                notBefore = _lazyNotBefore = Pal.NotBefore;
+            return notBefore;
+        }
+
+        internal void ThrowIfInvalid()
+        {
+            if (Pal == null)
+                throw new CryptographicException(SR.Format(SR.Cryptography_InvalidHandle, "m_safeCertContext")); // Keeping "m_safeCertContext" string for backward compat sake.
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2.cs
```diff
@@ -0,0 +1,521 @@
...
+
+            StringBuilder sb = new StringBuilder();
+            String newLine = Environment.NewLine;
+            String newLine2 = newLine + newLine;
+            String newLinesp2 = newLine + "  ";
+
+            // Version
+            sb.Append("[Version]");
+            sb.Append(newLinesp2);
+            sb.Append("V" + this.Version);
```
stephentoub: Ditto on interim strings.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509CertificateCollection.cs
```diff
@@ -0,0 +1,235 @@
...
+        bool ICollection.IsSynchronized
+        {
+            get { return false; }
+        }
+
+        Object ICollection.SyncRoot
+        {
+            get
+            {
+                return Interlocked.CompareExchange(ref _syncRoot, new Object(), null);
```
stephentoub: This will allocate a new object on every access, even if _syncRoot is already non-null.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Chain.cs
```diff
@@ -0,0 +1,137 @@
...
+using Internal.Cryptography.Pal;
+
+namespace System.Security.Cryptography.X509Certificates
+{
+    public class X509Chain : IDisposable
+    {
+        public X509Chain()
+        {
+            Reset();
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Extension.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (extension == null)
+                throw new ArgumentException(SR.Cryptography_X509_ExtensionMismatch);
+            base.CopyFrom(asnEncodedData);
+            Critical = extension.Critical;
+        }
+
+        internal X509Extension(String oidValue)
+        {
+            base.Oid = Oid.FromOidValue(oidValue, OidGroup.ExtensionOrAttribute);
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509ExtensionEnumerator.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return _extensions[_current];
+            }
+        }
+
+        Object IEnumerator.Current
+        {
+            get
+            {
+                return _extensions[_current];
```
stephentoub: This could just delegate to Current.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509KeyUsageExtension.cs
```diff
@@ -0,0 +1,58 @@
...
+                }
+                return _keyUsages;
+            }
+        }
+
+        public override void CopyFrom(AsnEncodedData asnEncodedData)
+        {
+            base.CopyFrom(asnEncodedData);
+            _decoded = false;
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Store.cs
```diff
@@ -0,0 +1,142 @@
...
+        public StoreLocation Location { get; private set; }
+
+        public String Name { get; private set; }
+
+
+        public void Open(OpenFlags flags)
+        {
+            Close();
+            _storePal = StorePal.FromSystemStore(this.Name, this.Location, flags);
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Store.cs
```diff
@@ -0,0 +1,142 @@
...
+        public void Remove(X509Certificate2 certificate)
+        {
+            if (certificate == null)
+                throw new ArgumentNullException("certificate");
+
+            if (_storePal == null)
+                throw new CryptographicException(SR.Cryptography_X509_StoreNotOpen);
+
+            _storePal.Remove(certificate.Pal);
+            return;
```
stephentoub: nit: unnecessary return

src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        [Fact]
+        public static void X509Cert2ToStringVerbose()
+        {
+            X509Store store = new X509Store("My", StoreLocation.CurrentUser);
+            store.Open(OpenFlags.ReadOnly);
+
+            foreach (X509Certificate2 c in store.Certificates)
+            {
+                Assert.NotNull(c.ToString(true));
+            }
```
stephentoub: Should we dispose of the individual ```c```s?

src/System.Security.Cryptography.X509Certificates/tests/CollectionTests.cs
```diff
@@ -0,0 +1,379 @@
...
+
+                using (X509Certificate2 second = cs[1])
+                {
+                    Assert.NotSame(pfxCer, second);
+                    Assert.Equal(pfxCer, second);
+                }
+            }
+        }
+
+        private static X509Certificate2[] ToArray(this X509Certificate2Collection col)
```
stephentoub: We can't use LINQ for helpers like this?

src/System.Security.Cryptography.X509Certificates/tests/InteropTests.cs
```diff
@@ -0,0 +1,329 @@
...
+                byte[] publicKey = c.GetPublicKey();
+                Assert.Equal(expectedPublicKey, publicKey);
+
+                byte[] expectedThumbPrint = "108e2ba23632620c427c570b6d9db51ac31387fe".HexToByteArray();
+                byte[] thumbPrint = c.GetCertHash();
+                Assert.Equal(expectedThumbPrint, thumbPrint);
+            }
+        }
+
+        [DllImport("crypt32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
```
stephentoub: This interop code is duplicative of what's used by the product, right?  We're doing that to avoid the tests taking a dependency on it in case it's the source of the problem?

stephentoub:
Thanks, Jeremy.  There's a lot of code being ported here, so I just sort of spot checked it.  Left a bunch of comments, but they're all relatively minor and can be addressed subsequently.  Otherwise, once the tests pass, LGTM.

justinvp:
I noticed a bunch of uses of `String` and `Object` (uppercase) where `string` and `object` (lowercase) should be used instead per the [Coding Style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-style.md):

> We use language keywords instead of BCL types (i.e. int, string, float instead of Int32, String, Single, etc) for both type references as well as method calls (i.e. int.Parse instead of Int32.Parse). See issue 391 for examples.

src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
bartonjs: Honestly, I've never been sure what the right thing to do with looping over the store is.  "Yes" is probably the correct answer, because X509Store::Close can't (because it doesn't know if you kept a reference).  But I imagine very little customer code Disposes them.

src/System.Security.Cryptography.X509Certificates/tests/CollectionTests.cs
bartonjs: This one, I think, is because System.Linq doesn't like .NET 1.0/1.1 typed collections.  One would have to run it through coll.Cast<X509Certificate2> (or OfType) in order to call ToArray() and have the strong reference.

Unless I'm entirely mistaken, of course.

src/System.Security.Cryptography.X509Certificates/tests/InteropTests.cs
bartonjs: I think I had exactly the same thought while prepping the tests for publication :smile:.  I decided that it was testing that a customer could write their own interop if they want to use handles out of our types into CAPI themselves.

"Legitimately wasteful", perhaps?

bartonjs:
Created one uber-issue to encompass the feedback given.  I'm going to merge as-is, since this PR is largely just for "get the code moved from private source to github", and the cleanup can be done later.

PR #1992, opened 2015-06-10
Fix overflowing key size in AesImplementation

### LINKED ISSUE #1981: Potential overflow in key size validation in AesImplementation::CreateTransform
This is another instance of https://github.com/dotnet/corefx/issues/1755 which was revealed by publishing the AES library.

stephentoub:
Check that the key's size in bits doesn't overflow Int32.MaxValue, or else we could potentially be allowing in a gigantic key that happens to overflow to a valid key size.

Fixes #1981.

### DIFF
--- a/src/System.Security.Cryptography.Encryption.Aes/src/Internal/Cryptography/AesImplementation.cs
+++ b/src/System.Security.Cryptography.Encryption.Aes/src/Internal/Cryptography/AesImplementation.cs
@@ -46,8 +46,12 @@ namespace Internal.Cryptography
             if (keySize > int.MaxValue || !((int)keySize).IsLegalSize(this.LegalKeySizes))
                 throw new ArgumentException(SR.Cryptography_InvalidKeySize, "key");
 
-            if (rgbIV != null && rgbIV.Length * BitsPerByte != BlockSize)
-                throw new ArgumentException(SR.Cryptography_InvalidIVSize, "iv");
+            if (rgbIV != null)
+            {
+                long ivSize = rgbIV.Length * (long)BitsPerByte;
+                if (ivSize != BlockSize)
+                    throw new ArgumentException(SR.Cryptography_InvalidIVSize, "iv");
+            }
 
             if (encrypting)
                 return CreateEncryptor(Mode, Padding, rgbKey, rgbIV, BlockSize / BitsPerByte);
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @bartonjs, @rielz

src/System.Security.Cryptography.Encryption.Aes/src/Internal/Cryptography/AesImplementation.cs
```diff
@@ -41,11 +41,13 @@ private ICryptoTransform CreateTransform(byte[] rgbKey, byte[] rgbIV, bool encry
...
-            if (!keySize.IsLegalSize(this.LegalKeySizes))
-                throw new ArgumentException(SR.Format(SR.Cryptography_InvalidKeySize, "key"));
+
+            long keySize = rgbKey.Length * (long)BitsPerByte;
+            if (keySize > int.MaxValue || !((int)keySize).IsLegalSize(this.LegalKeySizes))
+                throw new ArgumentException(SR.Cryptography_InvalidKeySize, "key");
+
             if (rgbIV != null && rgbIV.Length * BitsPerByte != BlockSize)
-                throw new ArgumentException(SR.Format(SR.Cryptography_InvalidIVSize, "iv"));
+                throw new ArgumentException(SR.Cryptography_InvalidIVSize, "iv");
```
stephentoub: I just noticed this multiplication could overflow as well.  Will address in subsequent commit.

stephentoub: Ok, fixed.

bartonjs:
LGTM, matches the other fix.

I'm wondering, though, if it wouldn't read better and/or perform better if instead the array length was compared to be strictly less than `0x2000000` (the smallest value which can't be multiplied by 8).

Another alternative that I think wins on legibility (but perhaps fails on "don't simplify this") is `long keySize = rgbKey.LongLength * BitsPerByte;`, since it removes the "cast" (which likely isn't really a cast because `const`s are magical to the compiler).

So the long and short of it is: "Looks good, consider if it could look 'better'" :)

stephentoub:
Thanks, Jeremy.  LongLength doesn't exist in the .NET Core surface area (probably because you can't have an array length larger than an Int32's MaxValue, and the max is actually slightly smaller than that).  While it's clever, I actually prefer what's currently here over the 0x2000000 thing because I think it's more understandable, and the few extra instructions here are going to be completely dwarfed by all of the cryptographic calculations we're about to do, so I can't imagine they would actually show up in any measurable way.  I'd like to leave it as is, but let me know if you disagree.

bartonjs:
Nah, it was definitely just "consider". I think the clarity/understandability aspect is potato/potato (ah, things that don't work in a written medium); the important thing is that there's a test for it, so it shouldn't regress.

LGTM, thanks for taking this one.

stephentoub:
Thanks :smile:

PR #1998, opened 2015-06-11
Remove usage of ArrayT<T>

stephentoub:
```ArrayT<T>``` was a temporary workaround for some performance issues during the implementation of .NET Native.  Those performance issues have been fixed and the workaround is no longer needed.

This commit removes all usage of ```ArrayT<T>``` from the code currently on GitHub. The ArrayT.cs file itself is left to avoid breaking any libraries that haven't yet been ported to GitHub when the mirror moves these changes back internally.

### DIFF
--- a/src/Common/src/System/Collections/Generic/LowLevelList.cs
+++ b/src/Common/src/System/Collections/Generic/LowLevelList.cs
@@ -137,14 +137,9 @@ namespace System.Collections.Generic
                 {
                     if (value > 0)
                     {
-#if TYPE_LOADER_IMPLEMENTATION
                         T[] newItems = new T[value];
-                        for (int i = 0; i < _size; i++)
-                            newItems[i] = _items[i];
+                        Array.Copy(_items, 0, newItems, 0, _size);
                         _items = newItems;
-#else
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

**Comment by stephentoub** [comment]
cc: @jkotas, @ellismg

File: src/Common/src/System/Collections/Generic/LowLevelList.cs [hunk 1]
```diff
@@ -137,14 +137,14 @@ public int Capacity
                 {
                     if (value > 0)
                     {
-#if TYPE_LOADER_IMPLEMENTATION
                         T[] newItems = new T[value];
+#if TYPE_LOADER_IMPLEMENTATION
```
jkotas: This ifdef is not needed anymore - Array.Copy will work fine even for TYPE_LOADER_IMPLEMENTATION .

jkotas:
LGTM otherwise.

src/Common/src/System/Collections/Generic/LowLevelList.cs
stephentoub: Thanks, Jan.  Fixed.

PR #2004, opened 2015-06-11
Fix potential AVs from poorly written HashAlgorithm usage

### LINKED ISSUE #1996: Potential native out of bounds read caused by HashAlgorithm.ComputeHash(Stream)
`ComputeHash(Stream)` reads bytes to a buffer and remembers the bytes read indicated by the return value of `Stream.Read`. It then passes both to `HashAlgorithm.HashCore` of the respective implementations.
The built-in hash algorithms (such as SHA1) simply pass that on to `HashProvider.AppendHashData` which (in the case of Windows, same problem may exist in the Unix part) calls `BCryptHashData` without verifying that the byte length is valid for the buffer.

Thus the returned value of `Stream.Read` is passed unfiltered and unchecked to `BCryptHashData`, so an evil/buggy/wrong implementation of Stream can cause an out of bounds read in native `BCryptHashData` with values larger than the actual buffer.

As the `Stream` is a public parameter of `ComputeHash` and `Stream` is freely extensible, every return value of its members should be verified!
In this case the return value of `Stream.Read` should be compared to the buffer size or capped with `Math.Min`.

stephentoub:
The implementation of HashCore in all of our HashAlgorithms on both Windows and Unix uses the buffer/offset/count in unsafe code without argument validation.  While most usage of HashCore comes from internal data that's constructed to be correct, there are two paths where externally supplied data can arrive to HashCore unverified:

1. A Stream passed to ComputeHash can return a number of bytes read that's larger than the buffer size.  This value will be passed as the count to HashCore.
2. A type can derive from one of the hashing algorithms and access HashCore directly, which is protected.

This PR adds argument validation to our Windows and Unix implementations prior to using unsafe code to process those arguments.

Fixes #1996

### DIFF
--- a/src/System.Security.Cryptography.Hashing.Algorithms/src/Internal/Cryptography/HashProvider.cs
+++ b/src/System.Security.Cryptography.Hashing.Algorithms/src/Internal/Cryptography/HashProvider.cs
@@ -12,7 +12,28 @@ namespace Internal.Cryptography
     internal abstract class HashProvider
     {
         // Adds new data to be hashed. This can be called repeatedly in order to hash data from incontiguous sources.
-        public abstract void AppendHashData(byte[] data, int offset, int count);
+        public void AppendHashData(byte[] data, int offset, int count)
+        {
+            // AppendHashData can be called via exposed APIs (e.g. a type that derives from
+            // HMACSHA1 and calls HashCore) and could be passed bad data from there.  It could
+            // also receive a bad count from HashAlgorithm reading from a Stream that returns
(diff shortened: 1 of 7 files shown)


stephentoub:
cc: @bartonjs

stephentoub:
@bartonjs, addressed your feedback. Thanks!

bartonjs:
LGTM

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #1602, opened 2015-05-01
added unit test for ScaffoldColumnAttribute

ghuntley:
This progresses issue #921 - "Improve code coverage for System.ComponentModel.Annotations" and is my first contribution to corefx. :tada: 

unit-test

Question: I've noticed that throughout the codebase unit tests generally do not test the attribute contract `AttributeUsage` - why would we not test `AllowMultiple = false` or `AttributeTargets.Field`?

PR #2005, opened 2015-06-11
Add tests for EditableAttribute and ScaffoldColumnAttribute

stephentoub:
@ghuntley, I've cherry-picked your commits from #1602 and #1619, and I've fixed them up according to the remainder of the feedback from those PRs.  This PR replaces those two, which I will close out.


ghuntley:
:bow:

PR #2006, opened 2015-06-11
Fix and enable RSA and X509Certificates tests on Unix

### LINKED ISSUE #1986: RSA/X509Certificates tests are segfaulting on Unix

bartonjs:
Resolves #1986.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
@@ -58,7 +58,7 @@ namespace Internal.Cryptography.Pal
 
         public IntPtr Handle
         {
-            get { return IntPtr.Zero; }
+            get { return _cert == null ? IntPtr.Zero : _cert.DangerousGetHandle(); }
         }
 
         public string Issuer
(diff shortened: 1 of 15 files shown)


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -53,7 +58,7 @@ public AsymmetricAlgorithm PrivateKey
 
         public IntPtr Handle
         {
-            get { return IntPtr.Zero; }
+            get { return _cert == null ? IntPtr.Zero : _cert.DangerousGetHandle(); }
```
stephentoub: Heh, this one made me smile :smile:

src/System.Security.Cryptography.X509Certificates/tests/TestData.cs
```diff
@@ -370,5 +370,20 @@ internal static class TestData
             "29008DE8B851E2C30B6BF73F219BCE651E5972E62D651BA171D1DA9831A449D99" +
             "AF4E2F4B9EE3FD0991EF305ADDA633C44EB5E4979751280B3F54F9CCD561AC27D" +
             "3426BC6FF32E8E1AAF9F7C0150A726B").HexToByteArray();
+
+        internal static string NormalizeX500String(string expected)
+        {
+            if (StringComparer.Ordinal.Equals("Windows_NT", Environment.GetEnvironmentVariable("OS")))
```
stephentoub: Can you use Interop.PlatformDetection.cs instead? i.e. Interop.IsWindows?

stephentoub:
One minor nit, otherwise LGTM.

src/System.Security.Cryptography.X509Certificates/tests/TestData.cs
bartonjs: Done.

Issue #1995, opened 2015-06-10
Code cleanup for X509Certificates


bartonjs:
This issue is capturing feedback from the PR which copied the source into github.

- [x] Change String/Object to string/object (and any other Type=>keyword) changes encountered. (https://github.com/dotnet/corefx/pull/1990#issuecomment-110880727)
- [x] While cleaning things up, remove unnecessary "this." expressions.
- [x] Interop types are redefined, do not match the guidance (https://github.com/dotnet/corefx/pull/1990#discussion_r32144939, and others)
- [x] Internal.Cryptography.Helpers::DecodeHexString should skip whitespace, not use a string-allocating interior trim method. (https://github.com/dotnet/corefx/pull/1990#discussion_r32141572)
- [x] The "negativeSize" calling methods should have helper methods in the interop type and have the P/Invoke be private (https://github.com/dotnet/corefx/pull/1990#discussion_r32142618, and others).
- [x] OpenSslX509Encoder::DecodeX509SubjectKeyIdentifierExtension should validate the handle before calling DangerousGetHandle (https://github.com/dotnet/corefx/pull/1990#discussion_r32143320)
- [x] \(Windows) Internal.Cryptography.Pal.X509Pal.Encodex509EnhancedKeyUsageExtension() should validate the handle before calling DangerousGetHandle (https://github.com/dotnet/corefx/pull/1990#discussion_r32146020)
- [x] `if (handle.IsInvalid) throw new CryptographicException(GetOpenSslErrorString();` happens a lot, consider a helper method (https://github.com/dotnet/corefx/pull/1990#discussion_r32143432)
- [x] X509Certificate.ToString() and X509Certificate2.ToString() do a lot of allocations, try to reduce them (https://github.com/dotnet/corefx/pull/1990#discussion_r32144615).
- [x] A lot of unnecessary `return;`s as the last/only statement in a method. (https://github.com/dotnet/corefx/pull/1990#discussion_r32145182, and others)
- [x] SafeHandle types have an InvalidHandle property that allocates each call, make them return a single instance (https://github.com/dotnet/corefx/pull/1990#discussion_r32145333, and others)
- [x] \(Windows) Internal.Cryptography.Pal.StorePal.PositiveBigIntegerFromByteArray() should test if the number "is" negative before doing the copy. (Return the existing array if it's OK) (https://github.com/dotnet/corefx/pull/1990#discussion_r32145804)
[...]

PR #2007, opened 2015-06-11
Remove some allocations from X509 library

stephentoub:
This addresses three of the items listed in https://github.com/dotnet/corefx/issues/1995, mostly related to reducing allocations:

1. "Internal.Cryptography.Helpers::DecodeHexString should skip whitespace, not use a string-allocating interior trim method"
2. "X509Certificate.ToString() and X509Certificate2.ToString() do a lot of allocations, try to reduce them"
3. "SyncRoot allocates a new object every call (which is thrown away on all calls aside from the first)"

The third one also addresses a bug: SyncRoot was returning null the first time it was accessed.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
@@ -2,10 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Text;
 using System.Diagnostics;
 using System.Globalization;
-using System.Security.Cryptography;
 
 namespace Internal.Cryptography
 {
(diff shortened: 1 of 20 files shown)


stephentoub:
cc: @bartonjs, @KrzysztofCwalina

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
```diff
@@ -83,6 +72,14 @@ private static byte HexToByte(char val)
                 return 0xFF;
         }
 
+        private static char ByteToHex(byte b)
```
KrzysztofCwalina: Should it be called NibbleToHex? And why does it return "?" in what seems like an error case?

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
```diff
@@ -262,32 +262,42 @@ public virtual String ToString(bool fVerbose)
...
 
             // Serial Number
-            sb.Append(Environment.NewLine + Environment.NewLine + "[Serial Number]" + Environment.NewLine + "  ");
+            sb.AppendLine();
+            sb.AppendLine("[Serial Number]");
+            sb.Append("  ");
             byte[] serialNumber = this.GetSerialNumber();
             Array.Reverse(serialNumber);
-            sb.Append(serialNumber.ToHexStringUpper());
+            sb.AppendLine(serialNumber.ToHexStringUpper());
```
KrzysztofCwalina: It seems to me like ToHexStringUpper should return char[] instead of string and then the char[] should be appended here.

bartonjs:
The pedant in me cares about Byte vs Nybble.  The pragmatist in me doesn't care.
Pragmatist: "Hey, pedant, look over there!" ::points::
(Pedant turns to see the the intended target)
Pragmatist: LGTM

KrzysztofCwalina:
@bartonjs, yeah, normally I would not care, but there seems to be an error condition there. If you pass more than a nibble, you will get "?" back. I though the naming might make it clearer.

bartonjs:
@KrzysztofCwalina Oh, I agree :smile:, which is why I had also pointed it out.  I just won't withhold my blessing based on the naming of those two private methods.

stephentoub:
Thanks, guys.  While I was just trying to mirror the function that already existed, both in name and in function, since you both had the same objection, I'll go ahead and make the change.  I'm hesitant to change the implementation of the existing function in case some code somewhere depended on it, but I'll take a look to see if I can easily prove it doesn't matter.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs
stephentoub: Good idea.

KrzysztofCwalina:
BTW, I don't feel super strongly about the name; treat it as a suggestion. I do care a bit more about converting char[] to string so that it can immediately be converted back to chars inserted into the string builder.

stephentoub:
Yeah, the char[] to string was just an oversight on my part: I was focused on re-implementing the body while maintaining the signature (which already returned string), and didn't go look at the call sites to see how it was used.

stephentoub:
Thanks for the review, guys.  I fixed up the name and the implementation of NibbleToHex, avoided roundtripping through a string for the two call sites where a char[] was acceptable (there were six others that needed a string), and per your request Jeremy, added a commit to remove the extraneous "this." usage.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
```diff
@@ -16,59 +14,54 @@ public static byte[] CloneByteArray(this byte[] src)
...
-            byte[] hex = new byte[cbHex];
-            int i = 0;
-            for (int index = 0; index < cbHex; index++)
+            int whitespaceCount = 0;
+
+            for (int i = 0; i < s.Length; i++)
             {
-                hex[index] = (byte)((HexToByte(hexString[i]) << 4) | HexToByte(hexString[i + 1]));
-                i += 2;
+                if (Char.IsWhiteSpace(s[i]))
```
ellismg: NIT: char not Char, correct?

KrzysztofCwalina:
Looks good!

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
stephentoub: Ugh, yeah, I was just copying and pasting what was there.  I'll fix it.

PR #2014, opened 2015-06-11
Disable more FileSystemWatcher tests on OSX

stephentoub:
Turns out my previous run of the OSX tests wasn't actually running all tests, due to my having an outdated xunit.netcore.extensions.dll and xunit silently not running some tests as a result.


stephentoub:
cc: @sokket, @ellismg

sokket:
LGTM

PR #2017, opened 2015-06-12
Enable file-load tests for X509Certificates

bartonjs:
For the Unix implementation the removal of the 1977 ActiveIssue required adding the 1993 ActiveIssue, since file-load is not yet implemented for the Unix builds.

Borrowed the package unpack/copy target from the System.IO.Compression tests.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
+++ b/src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
@@ -10,13 +10,13 @@ namespace System.Security.Cryptography.X509Certificates.Tests
     public static class CertTests
     {
         [Fact]
-        [ActiveIssue(1977, PlatformID.Any)]
+        [ActiveIssue(1993, PlatformID.AnyUnix)]
         public static void X509CertTest()
         {
             const string CertSubject =
                 @"CN=Microsoft Corporate Root Authority, OU=ITG, O=Microsoft, L=Redmond, S=WA, C=US, E=pkit@microsoft.com";
 
-            using (X509Certificate cert = new X509Certificate("microsoft.cer"))
+            using (X509Certificate cert = new X509Certificate("TestData\\microsoft.cer"))
             {
                 Assert.Equal(CertSubject, cert.Subject);
                 Assert.Equal(CertSubject, cert.Issuer);
@@ -50,13 +50,13 @@ namespace System.Security.Cryptography.X509Certificates.Tests
         }
 
         [Fact]
-        [ActiveIssue(1977, PlatformID.Any)]
+        [ActiveIssue(1993, PlatformID.AnyUnix)]
         public static void X509Cert2Test()
         {
             const string CertName =
                 @"E=admin@digsigtrust.com, CN=ABA.ECOM Root CA, O=""ABA.ECOM, INC."", L=Washington, S=DC, C=US";
 
-            using (X509Certificate2 cert2 = new X509Certificate2("test.cer"))
+            using (X509Certificate2 cert2 = new X509Certificate2("TestData\\test.cer"))
             {
                 Assert.Equal(CertName, cert2.IssuerName.Name);
                 Assert.Equal(CertName, cert2.SubjectName.Name);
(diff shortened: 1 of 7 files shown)


bartonjs:
Resolves #1977.

ellismg:
LGTM.

bartonjs:
It occurs to me that these should all use Path.Combine, but that'll really come up when load-certificate-from-file support is added.  The tests that already believed their files to live under TestData don't show as having anything weird going on.

src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
```diff
@@ -10,13 +10,13 @@ namespace System.Security.Cryptography.X509Certificates.Tests
...
         [Fact]
-        [ActiveIssue(1977, PlatformID.Any)]
+        [ActiveIssue(1993, PlatformID.AnyUnix)]
         public static void X509CertTest()
         {
             const string CertSubject =
                 @"CN=Microsoft Corporate Root Authority, OU=ITG, O=Microsoft, L=Redmond, S=WA, C=US, E=pkit@microsoft.com";
 
-            using (X509Certificate cert = new X509Certificate("microsoft.cer"))
+            using (X509Certificate cert = new X509Certificate("TestData\\microsoft.cer"))
```
stephentoub: I assume we'll eventually enable these for Unix as well?  Can we use Path.Combine for these paths instead, e.g. Path.Combine("TestData", "microsoft.cert")?

src/System.Security.Cryptography.X509Certificates/tests/System.Security.Cryptography.X509Certificates.Tests.csproj
```diff
@@ -39,4 +39,19 @@
     <None Include="project.json" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+  <!-- Temporary until we have new work in build tools to
+       deploy content from nuget packages -->
+  <Target Name="CopyX509TestDataToTestDirectory" AfterTargets="CopyTestToTestDirectory" Inputs="$(PackagesDir)$(X509TestPackage)\content" Outputs="$(TestPath)%(TestTargetFramework.Folder)\*.*">
```
stephentoub: Similarly, I seem to remember there being some discussion about what slash to use in project files... @JeremyKuhne, where did that land?

stephentoub:
I just commented on the Path.Combine thing.  Let's change those now and do it right from the start.  Otherwise, LGTM.

src/System.Security.Cryptography.X509Certificates/tests/System.Security.Cryptography.X509Certificates.Tests.csproj
JeremyKuhne: It is safer to use forward slash for xplat compatibility. MSBuild will attempt to fix up paths if it can identify them as such, but that won't always work (it depends on the root directory in the given string existing).

src/System.Security.Cryptography.X509Certificates/tests/CertTests.cs
```diff
@@ -50,13 +50,13 @@ public static void X509CertTest()
         }
 
         [Fact]
-        [ActiveIssue(1977, PlatformID.Any)]
+        [ActiveIssue(1993, PlatformID.AnyUnix)]
```
stephentoub: Looks like the Windows runs failed due to another timezone issue:
```
System.Security.Cryptography.X509Certificates.Tests.CertTests.X509Cert2Test [FAIL]
        Assert.Equal() Failure
        Expected: 2009-07-09T10:33:53.0000000
        Actual:   2009-07-09T17:33:53.0000000+00:00
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Security.Cryptography.X509Certificates\tests\CertTests.cs(70,0): at System.Security.Cryptography.X509Certificates.Tests.CertTests.X509Cert2Test()
```

stephentoub:
LGTM

PR #1589, opened 2015-04-29
Reimplement TaskExtensions.Unwrap

stephentoub:
The currently checked in implementation of TaskExtensions.Unwrap isn't what's in .NET 4.5/6 desktop.  Rather, it's a slight improvement over what was shipped in .NET 4, before a bunch of optimizations were made to it for .NET 4.5.  In .NET 4.5, we effectively built the implementation into Task itself, taking advantage of Task internals to minimize allocations and other overheads; the Unwrap extension methods then had internals access and could delegate to that implementation in mscorlib.  For the separate implementation now in System.Threading.Tasks.dll in .NET Core, we don't have that internals access, and need to implement Unwrap on top of Task's public surface area.

Even so, we can do much better than the checked in implementation, and this PR overhauls it.  In the common case, the current implementation incurs 17 allocations (!) per Unwrap, whereas mscorlib's implementation incurs only 1 (along with a commensurate reduction in the number of bytes allocated).  This implementation gets the number down from 17 to 4, again with a commensurate size reduction.  It also gets the single-threaded speed of the implementation (i.e. not taking into account the GC impact of the extra allocations) to effectively the same as the implementation in mscorlib, whereas the old implementation is measurably worse.

It also addresses an inconsistency between the implementations.  In .NET 4.5/6, when the inner task has a CancellationToken that causes the task to be canceled, that CancellationToken is stored into the proxy task, a fact visible due to that token propagating out of any OperationCanceledException thrown when await'ing the proxy task.  The current implementation doesn't have that, but rather the pre-.NET 4.5 behavior where the token wasn't marshaled.  This PR brings it back.

I will hold off on merging this until the System.Threading.Tasks tests are up in the repo and I can validate this against them.

Issue #702, opened 2015-02-09
Open-source System.Threading.Tasks unit tests


nslottow:
Note: The implementation of this contract lives in mscorlib in the [coreclr](https://github.com/dotnet/coreclr) repository, but we are planning to put tests in this repo and test against the NuGet package produced by the coreclr repo.

jasonwilliams200OK:
@mellinoe any updates on this?

(Bob created a duplicate issue https://github.com/dotnet/corefx/issues/850. Would be useful to have dedup functionality implemented in Bob! :traffic_light:)

stephentoub:
He can correct me if I'm wrong, but I believe Eric is waiting for a new xunit to be used (#1876), as it should contain a change that makes it easier to test Tasks (https://github.com/xunit/xunit/commit/2f2efc5be1be468a3efe72ff5faa858b9f99926a).

mellinoe:
To give an update: Yeah, I'm waiting on the latest Xunit because it should let us bring our existing tests forward without having to make a ton of changes to avoid deadlocks. These are Tasks tests, so naturally there's a LOT of calls that would have to be modified...

For what it's worth, I built my own simple console test runner yesterday and tried running the tests I have ported from our internal sources, and they ran just fine without any deadlocks. I also ran them all in parallel with Task.Run and they still worked fine.

Granted, they took an insanely long time (6+ minutes) to finish, so I will still need to choose a useful subset and mark the rest as OuterLoop before they can get up into here.

AlfredoMS:
Any updates?

mellinoe:
We still need the latest version of Xunit. There was a PR but it went stale:

https://github.com/dotnet/corefx/pull/1876

That is still what is blocking these tests in particular.

mellinoe:
Fixed with #2775

PR #2019, opened 2015-06-12
Add TaskExtensions.Unwrap tests

stephentoub:
This commit provides a System.Threading.Tasks.Tests project that doesn't include all of the tests that @mellinoe is porting (#702).  However, it does include a new set of tests written from scratch for TaskExtensions.Unwrap, which is the portion of System.Threading.Tasks whose source lives in corefx rather than in coreclr.  

These new tests are meant to be the primary unit testing for Unwrap, as the existing tests only superficially exercise the functionality. Once these are committed, #1589 can be validated against them (and the TODOs in the tests can be uncommented, as they're addressed by the new implementation).

### DIFF
--- a/src/System.Threading.Tasks/System.Threading.Tasks.sln
+++ b/src/System.Threading.Tasks/System.Threading.Tasks.sln
@@ -1,9 +1,11 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @mellinoe, @AlfredoMS

mellinoe:
Looks good to me. I can add on top of this when we upgrade to the latest Xunit. The tests look much better/cleaner than the old ones, thanks!

stephentoub:
Thanks, Eric.

PR #2021, opened 2015-06-12
Create FreeBSD build-target for corefx

josteink:
This partially addresses
https://github.com/dotnet/corefx/issues/1626

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -137,6 +137,7 @@
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('Windows'))">Windows_NT</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('Linux'))">Linux</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('OSX'))">OSX</OSGroup>
+    <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('FreeBSD'))">FreeBSD</OSGroup>
     <OSGroup Condition="'$(OSGroup)'==''">Windows_NT</OSGroup>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


stephentoub:
The changes to dir.props look good.  The changes to System.Runtime.Environment.csproj are going to conflict with #1999 and would be better left for a separate PR that fully augments that library after the revised version is merged.

josteink:
This PR is getting slimmer and slimmer by the minute. :)

I'll add the remaining bits once #1999 is merged. Seems like a small job and then we can start getting things done.

stephentoub:
:smile: Could you just revert the System.Runtime.Environment.csproj file entirely?  The commit still show that whitespace-only changes were made to it.

Even with it being a small change, it's a good first step, in that it'll allow you to build with ```build /p:OSGroup=FreeBSD /p:SkipTests=true``` on Windows and that should produce outputs in a FreeBSD specific output directory, building the FreeBSD-variant of each assembly.  Now, that said, not all assemblies will build successfully with this, only those that either aren't specialized based on OS (e.g. System.Threading.Tasks.Dataflow.dll) or those that only differentiate between Windows and   Unix (e.g. System.Security.Cryptography.*).  Some of the core libraries (e.g. System.IO.FileSystem.dll) will fail to compile because they have some files included under TargetsUnix and then others included under either TargetLinux or TargetsOSX... the types/members defined in the latter group will be missing on FreeBSD and will cause build errors.  I suggest the next step would be to stub out all such files, with members throwing PlatformNotSupportedException, so that you can at least get a clean build.  Some of the files simply contain enums and constants with values from header files, so those should be quick and easy.

josteink:
Done and done.

I had to actually work the command-line for that one, but I *do* agree that a good commit should only contain actually functional changes, so good thing somebody is keeping a watchful eye on me ;)

stephentoub:
LGTM :smile:

PR #2023, opened 2015-06-12
Makes a static field used by multiple tests ThreadStatic

### LINKED ISSUE #1544: System.Dynamic.Runtime failure in CI
```
MESSAGE:
Assert.Equal() Failure Expected: 0 Actual: 1
+++++++++++++++++++
STACK TRACE:
at ManagedTests.DynamicCSharp.Conformance.dynamic.context.indexer.genclass.genclass003.genclass003.Test.DynamicCSharpRunTest() in d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Dynamic.Runtime\tests\Dynamic.Context\Conformance.dynamic.context.indexer.genclass.cs:line 517
```
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/1026/

VSadov:
If tests run concurrently we could have random failures.

Fixes #1544

### DIFF
--- a/src/System.Dynamic.Runtime/tests/Dynamic.Context/Conformance.dynamic.context.indexer.genclass.cs
+++ b/src/System.Dynamic.Runtime/tests/Dynamic.Context/Conformance.dynamic.context.indexer.genclass.cs
@@ -29,6 +29,7 @@ namespace ManagedTests.DynamicCSharp.Conformance.dynamic.context.indexer.genclas
 
     public class MemberClass<T>
     {
+        [ThreadStatic]
         public static int Status;
         public bool? this[string p1, float p2, short[] p3]
         {
@@ -238,6 +239,7 @@ namespace ManagedTests.DynamicCSharp.Conformance.dynamic.context.indexer.genclas
     public class MemberClassWithClassConstraint<T>
         where T : class
     {
+        [ThreadStatic]
         public static int Status;
         public int this[int x]
         {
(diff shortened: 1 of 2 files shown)


VSadov:
@stephentoub @AlekseyTs - please take a look. A trivial fix.

stephentoub:
I see several other static Status fields in other classes in the same file (I didn't look at other files).   Do they need a similar fix?

VSadov:
@stephentoub - it seems the common pattern is that a helper class is used only by a single test. The cases where multiple tests use same helper class are rare. I could not see any right away, but I cannot say that with confidence.
This one is a definite problem. I will look for more of these, but it could be that we have just one that needs fixing. (I saw another similarly fixed already).

stephentoub:
Ok, thanks.  In that case, LGTM.

VSadov:
Found more cases. Perhaps we should fix more comprehensively.

stephentoub:
:smile:

stephentoub:
I haven't looked at the tests in detail, but would it make sense to make these members all instance instead of static (or thread static), and have the tests construct their own instance so that they don't conflict with anyone else?  Or do they need to be static by the nature of the tests?

VSadov:
Yes, I am considering making them instance members.

VSadov:
It is just there are 400+ of them :-)

VSadov:
Ok, many cases are hard to turn from static into instance because it is often used to track sideeffects. I.E. consuming code, typically dynamic, does some stuff and wants to know how many conversions happened.
On the other hand, after reviewing all of them, it seems like they are all contained logically by individual tests. I have found only 2 cases the are truly shared by multiple tests and are not yet ThreadStatic. I am just fixing those.

stephentoub:
Thanks for investigating.

LGTM.

AlekseyTs:
LGTM

PR #2027, opened 2015-06-12
Eliminate closure allocation in ZipArchiveEntry

Maxwe11:


src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
```diff
@@ -643,17 +643,21 @@ private CheckSumAndSizeWriteStream GetDataCompressor(Stream backingStream, Boole
...
-                        _crc32 = checkSum;
-                        _uncompressedSize = currentPosition;
-                        _compressedSize = backingStream.Position - initialPosition;
+            var checkSumStream = new CheckSumAndSizeWriteStream(
+                compressorStream, 
+                backingStream, 
+                leaveCompressorStreamOpenOnClose, 
+                this, 
+                onClose,
+                (Int64 initialPosition, Int64 currentPosition, UInt32 checkSum, Stream backing, ZipArchiveEntry entry, EventHandler onclose) =>
```
stephentoub: Same comment as on the other PR: maybe ```thisRef``` or ```innerThis``` or something like that instead of ```entry```.  I think it just better highlights that we want to be using ```this```, but can't due to the resulting closure.

And also same comment about not shadowing the ```onClose``` name from the outer scope.

src/System.IO.Compression/src/System/IO/Compression/ZipCustomStreams.cs
```diff
@@ -20,11 +20,11 @@ internal class WrappedStream : Stream
...
         #region constructors
 
-        internal WrappedStream(Stream baseStream, Boolean canRead, Boolean canWrite, Boolean canSeek, EventHandler onClosed)
-            : this(baseStream, canRead, canWrite, canSeek, false, onClosed)
+        internal WrappedStream(Stream baseStream, Boolean closeBaseStream)
+            : this(baseStream, true, true, true, closeBaseStream, null)
         { }
 
-        internal WrappedStream(Stream baseStream, Boolean canRead, Boolean canWrite, Boolean canSeek, Boolean closeBaseStream, EventHandler onClosed)
+        private WrappedStream(Stream baseStream, Boolean canRead, Boolean canWrite, Boolean canSeek, Boolean closeBaseStream, EventHandler onClosed)
```
stephentoub: Is WrappedStream only ever constructed with ```true``` for ```canRead``, ```canWrite```, and ```canSeek```?  If so, can we just get rid of those parameters and associated fields on the stream instance?  If the only reason they still exist is to be able to start returning false from ```Can*``` after disposal, ```_isDisposed``` could be checked in the ```Can*``` properties instead of those fields, right?

src/System.IO.Compression/src/System/IO/Compression/ZipCustomStreams.cs
```diff
@@ -317,10 +317,11 @@ internal class CheckSumAndSizeWriteStream : Stream
 
         //this is the position in BaseBaseStream
         private Int64 _initialPosition;
-
+        private readonly ZipArchiveEntry _zipArchiveEntry;
+        private readonly EventHandler _onClose;
         // Called when the stream is closed.
         // parameters are initialPosition, currentPosition, checkSum
-        private readonly Action<Int64, Int64, UInt32> _saveCrcAndSizes;
+        private readonly Action<Int64, Int64, UInt32, Stream, ZipArchiveEntry, EventHandler> _saveCrcAndSizes;
```
stephentoub: The comment on this is now out-of-date with regards to the parameters.

src/System.IO.Compression/src/System/IO/Compression/ZipCustomStreams.cs
```diff
@@ -331,8 +332,9 @@ internal class CheckSumAndSizeWriteStream : Stream
          *  currentPosition (in this CheckSumAndSizeWriteStream),
          *  checkSum (of data passed into this CheckSumAndSizeWriteStream)
         */
-        public CheckSumAndSizeWriteStream(Stream baseStream, Stream baseBaseStream,
-                                            Boolean leaveOpenOnClose, Action<Int64, Int64, UInt32> saveCrcAndSizes)
+        public CheckSumAndSizeWriteStream(Stream baseStream, Stream baseBaseStream, Boolean leaveOpenOnClose,
```
stephentoub: Comment is out-of-date.

src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
```diff
@@ -701,12 +705,13 @@ private Stream OpenInWriteMode()
                                                             (object o, EventArgs e) =>
                                                             {
                                                                 //release the archive stream
-                                                                _archive.ReleaseArchiveStream(this);
-                                                                _outstandingWriteStream = null;
+                                                                var entry = (ZipArchiveEntry)o;
```
stephentoub: GitHub won't let me leave a comment on L735, but could you do a similar cast-to-avoid-closure in the lambda in OpenInUpdateMode?

stephentoub:
Some comments to be addressed, but otherwise LGTM.

Maxwe11:
updated

Maxwe11:
`WrappedStream` used only twice and the [second usage still require closure](https://github.com/dotnet/corefx/pull/2027/files#diff-ba8d5a9ef0ce8bd24cf0576f0ab0972aR728), maybe we can replace `EventHandler` with `Action<ZipArchiveEntry>`? Since it's internal API

stephentoub:
> maybe we can replace EventHandler with Action<ZipArchiveEntry>?

Makes sense to me.

Maxwe11:
done

stephentoub:
LGTM.  Thanks!

PR #2030, opened 2015-06-13
FreeBSD enable: Make build.sh FreeBSD compatible

janhenke:
This patch adds FreeBSD to build.sh. Also use /usr/bin/env to get the real path of bash, since that can be different across platforms.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 __scriptpath=$(cd "$(dirname "$0")"; pwd -P)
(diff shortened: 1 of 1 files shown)


saper:
It would be really cool to make the script `/bin/sh` compatibile. I think the only real obstacle is whether we need the project directory guessed from the script location.

josteink:
Making the script `sh`-compatible would be nice, but I honestly don't consider it's crucial.

LGTM

stephentoub:
LGTM. @JeremyKuhne?

JeremyKuhne:
LGTM- thanks!

PR #2031, opened 2015-06-13
Add FreeBSD support to System.Console.

janhenke:
This commit adds the necessary Interop pieces necessary for System.Console. Also add FreeBSD to the System.Console project.

### DIFF
--- a/src/Common/src/Interop/FreeBSD/Interop.Errors.cs
+++ b/src/Common/src/Interop/FreeBSD/Interop.Errors.cs
@@ -1,0 +1,80 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+internal static partial class Interop
+{
+    /// <summary>Common Unix errno error codes.</summary>
+    internal static partial class Errors
+    {
+        // These values were defined in:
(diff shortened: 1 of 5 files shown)


Maxwe11:
>Add FreeBSD support to System.Console

Sounds like implementation but this just interops :smile: 
Why this should get in repo before implementation?

janhenke:
FreeBSD is a Unix system. There is already a Unix implementation. If you take a look into the files, the Interops are actually the only thing being different between Linux/OS X (and FreeBSD).

Maxwe11:
@janhenke :+1: yeah but I'm confused that `src/Common/src/Interop/FreeBSD/Interop.Libraries.cs` not included in csproj

janhenke:
That one file is not used by System.Console. It is actually identical to the Linux Interop file with the same name. Since the content is trivial, I added it as well. I assume some other assembly will need it.

stephentoub:
I didn't check the values used, but the mechanics of the change LGTM. Nice job plugging this in.

josteink:
I agree with @janhenke on this one.

While `Libraries` isn't used right now, it logically belongs with the rest of those files and it makes sense to introduce them all in the same commit. It will probably come to use sooner rather than later.

LGTM

PR #2032, opened 2015-06-13
Add FreeBSD to System.Private.Uri

janhenke:
Add FreeBSD to the System.Private.Uri.CoreCLR.csproj file. This commit depends on the Interop bits from PR#2031.


josteink:
LGTM

stephentoub:
LGTM

PR #2033, opened 2015-06-13
Avoid unnecessary allocations related to X509ChainElementCollection

justinvp:
- Avoid intermediate `LowLevelList<T>`, `Array`, and `Enumerator` allocations.
- Minor cleanup.

/cc @bartonjs

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/ChainPal.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/ChainPal.cs
@@ -2,21 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Text;
-using System.Diagnostics;
-using System.Globalization;
-using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
 
-using Internal.NativeCrypto;
-using Internal.Cryptography;
 using Internal.Cryptography.Pal.Native;
 
-using System.Security.Cryptography;
-
-using FILETIME = Internal.Cryptography.Pal.Native.FILETIME;
-using SafeX509ChainHandle = Microsoft.Win32.SafeHandles.SafeX509ChainHandle;
-using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
 
 namespace Internal.Cryptography.Pal
 {
@@ -49,7 +41,7 @@ namespace Internal.Cryptography.Pal
             return status.dwError == 0;
         }
 
-        public IEnumerable<X509ChainElement> ChainElements
+        public X509ChainElement[] ChainElements
         {
             get
             {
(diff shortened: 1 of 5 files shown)


jasonwilliams200OK:
@justinvp, isn't using generics and LINQ etc. better as there already are dedicated optimizers in compiler?

justinvp:
@jasonwilliams200OK, there are allocation costs with the use of LINQ.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/ChainPal.cs
```diff
@@ -49,7 +41,7 @@ internal sealed partial class ChainPal : IDisposable, IChainPal
             return status.dwError == 0;
         }
 
-        public IEnumerable<X509ChainElement> ChainElements
+        public X509ChainElement[] ChainElements
```
stephentoub: This isn't introduced by your change @justinvp, but @bartonjs, I'd feel better if property getters didn't introduce array allocations like this on every call... makes it too easy to accidentally introduce significant performance problems, e.g. by doing an iteration like ```for (int i = 0; i < cp.ChainElements.Length; i++) { var element = cp.ChainElements[i]; ... }```.  Should we change this to be a method?  I've not looked at the rest of the type, but there might be some other similar properties.

That's just a note for later and shouldn't prevent this PR from going ahead.

stephentoub:
LGTM

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/ChainPal.cs
bartonjs: You mean @dotnet-bot, of course :smile:.  The properties on the PAL types for the certificates are memoized by the calling type, and tend to have the same signature as their entrypoint.  So the allocating property is never called twice.

This part of X509 was, shall we say, before my time.  I'd probably have done it as methods, but don't know that changing it is justified.  (I'm merely providing insight here, not trying to add to inertia)

stephentoub: > You mean @dotnet-bot, of course :smile:

Sure :smile: Wasn't trying to assign blame, just address the person most likely to follow-up on it :wink:

> I'd probably have done it as methods, but don't know that changing it is justified

I personally think it'd be worthwhile, but up to you. It's just too easy to accidentally use it incorrectly.

bartonjs: It's amazing how different that sentence reads when you see comma/comma instead of comma/period.

For me the properties/methods debate for this is in the middle grey area.  I like the symmetry that the author introduced, but I don't like allocating properties.  So if someone wanted to change the properties to methods as a change in and of itself, I'd likely support it; but as part of another change I'd say it was excessively noisy.  Probably.

bartonjs:
X509ChainElementEnumerator's changes were form without function, and (Windows) ChainPal was largely function without form (at least, missed a String => string, whereas Object => object was changed in the enumerator).  It would have pleased me more if "cleanup only" files were in a "cleanup only" change; but there's nothing here that upset me enough to start quoting Pink Floyd (except that now I have to to explain the comment: ::clears throat:: "WROOOONG, do it again!" (voiceover in The Wall)).

Short form: LGTM.

PR #2038, opened 2015-06-14
Remove unnecessary SafeHandle allocations in X509Certificates library

stephentoub:
Several related issues:
- The custom SafeHandle types expose an InvalidHandle property.  As a property, one typically expects that each call to the property returns the same instance, but each call to these properties is manufacturing a new SafeHandle instance.
- Lots of the accesses to InvalidHandle are to get an initially "null" handle to pass to a native function that expects a null handle on the first invocation.  This means that we're allocating a finalizable object only to then throw it away.
- In some cases, we were completely ignoring the allocated handle, e.g. initializing a handle to InvalidHandle and then immediately using that variable as an out argument.
- etc.

This commit addresses these issues, such that the number of allocated handles noticeably decreases.  As an example, during the execution of the System.Security.Cryptography.X509Certificates.Tests as they currently exist, prior to this change 1661 SafePointerHandles were being allocated; after this change, that number drops to 1410, for a 15% reduction.

Contributes to #1995.

### DIFF
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Microsoft.Win32.SafeHandles
+{
+    /// <summary>Provides a cache for special instances of SafeHandles.</summary>
+    /// <typeparam name="T">Specifies the type of SafeHandle.</typeparam>
+    internal static class SafeHandleCache<T> where T : SafeHandle
+    {
+        private static T s_invalidHandle;
+
+        /// <summary>
+        /// Gets a cached, invalid handle.  As the instance is cached, it should either never be Disposed
+        /// or it should override <see cref="SafeHandle.Dispose(bool)"/> to prevent disposal when the
+        /// instance is the <see cref="InvalidHandle"/>.
+        /// </summary>
+        internal static T GetInvalidHandle(Func<T> invalidHandleFactory)
+        {
+            T currentHandle = Volatile.Read(ref s_invalidHandle);
+            if (currentHandle == null)
+            {
+                T newHandle = invalidHandleFactory();
+                currentHandle = Interlocked.CompareExchange(ref s_invalidHandle, newHandle, null);
+                if (currentHandle == null)
+                {
+                    GC.SuppressFinalize(newHandle);
+                    currentHandle = newHandle;
+                }
+                else
+                {
+                    newHandle.Dispose();
+                }
+            }
+            Debug.Assert(currentHandle.IsInvalid);
+            return currentHandle;
+        }
+
+        /// <summary>Gets whether the specified handle is <see cref="InvalidHandle"/>.</summary>
+        /// <param name="handle">The handle to compare.</param>
+        /// <returns>true if <paramref name="handle"/> is <see cref="InvalidHandle"/>; otherwise, false.</returns>
+        internal static bool IsCachedInvalidHandle(SafeHandle handle)
+        {
+            Debug.Assert(handle != null);
+            bool isCachedInvalidHandle = ReferenceEquals(handle, Volatile.Read(ref s_invalidHandle));
+            Debug.Assert(!isCachedInvalidHandle || handle.IsInvalid, "The cached invalid handle must still be invalid.");
+            return isCachedInvalidHandle;
+        }
+    }
+}
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Win32.SafeHandles;
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
@@ -12,22 +13,27 @@ namespace Internal.Cryptography.Pal.Native
     /// </summary>
     internal abstract class SafePointerHandle<T> : SafeHandle where T : SafeHandle, new()
     {
-        public SafePointerHandle()
+        protected SafePointerHandle()
             : base(IntPtr.Zero, true)
         {
         }
 
         public sealed override bool IsInvalid
         {
-            get
-            {
-                return handle == IntPtr.Zero;
-            }
+            get { return handle == IntPtr.Zero; }
         }
 
         public static T InvalidHandle
         {
-            get { return new T(); }
+            get { return SafeHandleCache<T>.GetInvalidHandle(() => new T()); }
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (!SafeHandleCache<T>.IsCachedInvalidHandle(this))
+            {
+                base.Dispose(disposing);
+            }
         }
     }
 
@@ -146,14 +152,11 @@ namespace Internal.Cryptography.Pal.Native
     /// </summary>
     internal sealed class SafeLocalAllocHandle : SafePointerHandle<SafeLocalAllocHandle>
     {
-        public SafeLocalAllocHandle()
-            : base()
-        {
-        }
-
         public static SafeLocalAllocHandle Create(int cb)
         {
-            return new SafeLocalAllocHandle(Marshal.AllocHGlobal(cb));
+            var h = new SafeLocalAllocHandle();
+            h.SetHandle(Marshal.AllocHGlobal(cb));
+            return h;
         }
 
         protected sealed override bool ReleaseHandle()
@@ -161,13 +164,5 @@ namespace Internal.Cryptography.Pal.Native
             Marshal.FreeHGlobal(handle);
             return true;
         }
-
-        private SafeLocalAllocHandle(IntPtr p)
-            : base()
-        {
-            handle = p;
-        }
     }
 }
-
-
--- a/src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
@@ -1,15 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.IO;
-using System.Text;
-using System.Diagnostics;
-using System.Globalization;
-using System.Runtime.InteropServices;
-
-using Internal.Cryptography;
 using Internal.Cryptography.Pal;
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.Win32.SafeHandles
 {
@@ -25,15 +20,22 @@ namespace Microsoft.Win32.SafeHandles
             get { return handle == IntPtr.Zero; }
         }
 
+        public static SafeX509ChainHandle InvalidHandle
+        {
+            get { return SafeHandleCache<SafeX509ChainHandle>.GetInvalidHandle(() => new SafeX509ChainHandle()); }
+        }
+
         protected override bool ReleaseHandle()
         {
             return ChainPal.ReleaseSafeX509ChainHandle(handle);
         }
 
-        internal static SafeX509ChainHandle InvalidHandle
+        protected override void Dispose(bool disposing)
         {
-            get { return new SafeX509ChainHandle(); }
+            if (!SafeHandleCache<SafeX509ChainHandle>.IsCachedInvalidHandle(this))
+            {
+                base.Dispose(disposing);
+            }
         }
     }
 }
-
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/CertificatePal.Import.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/CertificatePal.Import.cs
@@ -137,7 +137,7 @@ namespace Internal.Cryptography.Pal
                     certInfo.SerialNumber.pbData = pCmsgSignerInfo->SerialNumber.pbData;
                 }
 
-                SafeCertContextHandle pCertContext = SafeCertContextHandle.InvalidHandle;
+                SafeCertContextHandle pCertContext = null;
                 if (!Interop.crypt32.CertFindCertificateInStore(hCertStore, CertFindType.CERT_FIND_SUBJECT_CERT, &certInfo, ref pCertContext))
                     throw new CryptographicException(Marshal.GetHRForLastWin32Error());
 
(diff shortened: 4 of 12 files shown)


stephentoub:
cc: @bartonjs, @KrzysztofCwalina

stephentoub:
@dotnet-bot test this please

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
```diff
@@ -12,22 +13,19 @@ namespace Internal.Cryptography.Pal.Native
...
         }
 
-        public sealed override bool IsInvalid
+        internal static T CreateInvalid()
         {
-            get
-            {
-                return handle == IntPtr.Zero;
-            }
+            return new T();
```
sharwell: :bulb: You could call `GC.SuppressFinalize()` on the returned handle immediately.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
```diff
@@ -12,22 +13,19 @@ namespace Internal.Cryptography.Pal.Native
...
     internal abstract class SafePointerHandle<T> : SafeHandle where T : SafeHandle, new()
     {
-        public SafePointerHandle()
+        protected SafePointerHandle()
             : base(IntPtr.Zero, true)
         {
         }
 
-        public sealed override bool IsInvalid
+        internal static T CreateInvalid()
```
sharwell: :bulb: The introduction of `CreateInvalid()` at this location makes the diff unnecessarily complex. It would be better to move it below the `IsInvalid` property.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
```diff
@@ -12,22 +13,19 @@ namespace Internal.Cryptography.Pal.Native
...
-                return handle == IntPtr.Zero;
-            }
+            return new T();
         }
 
-        public static T InvalidHandle
+        public sealed override bool IsInvalid
         {
-            get { return new T(); }
+            get { return handle == IntPtr.Zero; }
```
sharwell: :bulb: This property didn't change. But the change to the brace layout means we all get to review it as though it did.

src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
```diff
@@ -20,6 +20,11 @@ private SafeX509ChainHandle()
         {
         }
 
+        internal static SafeX509ChainHandle CreateInvalid()
```
sharwell: :bulb: This method is better placed after the `IsInvalid` property.

src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
```diff
@@ -20,6 +20,11 @@ private SafeX509ChainHandle()
...
+        {
+            return new SafeX509ChainHandle();
```
sharwell: :bulb: You can call `GC.SuppressFinalize()` on the new instance immediately.

sharwell:
If you override `SafeHandle.Dispose(bool)`, you might actually be able to get away with a singleton cached invalid instance, and bring back the `InvalidHandle` property. I'm thinking something like the following (untested, never tried before browser code):

```
{
  if (this != SafePointerHandleCache<T>.Instance)
    base.Dispose(disposing);
}

private static class SafePointerHandleCache<T>
  where T : SafePointerHandle<T>, new
{
  public static T Instance;

  static SafePointerHandleCache()
  {
    Instance = new T();
    GC.SuppressFinalize(Instance);
  }
}
```

sharwell:
BTW: @stephentoub I'm not following you around to bash your changes. You just happen to be on a streak submitting pull requests on subject matters that particularly interest me.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
stephentoub: I disagree. I put it here because it's a factory and I believe should be grouped with the ctors. If the resulting diff is not ideal, that's lead about the change and more about the diff tool.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
stephentoub: I was already editing the code around this, and the change makes it more consistent with others elsewhere in the file. I'd rather change it as I notice it then not change it at all, which would be the effect of my skipping it "until later", and it's not worth subsequently trying to split the changes apart into multiple commits. I'm sorry if such whitespace changes makes it slightly harder to review.

src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
stephentoub: As noted on the same comment previously, I respectfully disagree.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
sharwell: :thought_balloon: There are a few problems with treating the method as a factory and thus placing it here. The primary problem is there is no way to define "factory" method such that it can be identified by an analyzer. Special cases like this make it difficult to impossible to transition to using analyzers for improving the overall consistency within source files. For example, it is very easy to check the order Constructors &rarr; Properties &rarr; Methods, but as soon as you allow any non-constructor "factory" method above a property, the best we can do is Constructors &rarr; `static` Methods &rarr; Properties &rarr; More `static` Methods &rarr; Methods. This is a more complicated structure for developers to recognize and adapt to, and requires more complicated analyzers and documentation. I believe special treatment for "factory" methods within the layout of a file raises more questions than it answers.

stephentoub: Thanks for the additional rationale. I still think a factory like this that simply provides a different name for the ctor makes more logical sense grouped with the ctor.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/SafeHandles.cs
stephentoub: In theory I like the suggestion.  In practice it makes me a bit nervous.  Since CreateInvalid is just another name for the constructor, a derived type could easily do something like:
```C#
public SafeSomethingHandle Create(IntPtr ptr)
{
    var h = CreateInvalid();
    h.SetHandle(ptr);
    return h;
}
```
and now we run the risk of never free'ing the handle.  Since the API does make it possible to reassign the handle values, I'm hesitant to suppress finalization anywhere else than in the API  generally understood to close the handle (at least in a base type like this).

stephentoub:
> If you override SafeHandle.Dispose(bool)

I like that suggestion.  Let me play around with it a bit...

stephentoub:
I updated the PR with a change that adds a SafeHandleCache and uses it from InvalidHandle. I'll squash before merging.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// Gets a cached, invalid handle.  As the instance is cached, it should either never be Disposed
+        /// or it should override <see cref="SafeHandle.Dispose(bool)"/> to prevent disposal when the
+        /// instance is the <see cref="InvalidHandle"/>.
+        /// </summary>
+        internal static T InvalidHandle
+        {
+            get
+            {
+                T currentHandle = Volatile.Read(ref s_invalidHandle);
```
sharwell: :bulb: This read does not need to be volatile.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        /// <summary>Gets whether the specified handle is <see cref="InvalidHandle"/>.</summary>
+        /// <param name="handle">The handle to compare.</param>
+        /// <returns>true if <paramref name="handle"/> is <see cref="InvalidHandle"/>; otherwise, false.</returns>
+        internal static bool IsCachedInvalidHandle(SafeHandle handle)
+        {
+            Debug.Assert(handle != null);
+            bool isCachedInvalidHandle = ReferenceEquals(handle, Volatile.Read(ref s_invalidHandle));
```
sharwell: :bulb: On X86, this read definitely does not need to be volatile because the `CompareExchange` instruction which sets the field uses the **lock** prefix, and thus has a full memory barrier. I don't know of any system which implements `CompareExchange` in any other way.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: According to ECMA and CLI, I believe it does.  While the CLR provides a stronger model, I prefer to use volatile to better match CLI and to better document the requirements on the memory model.  This shouldn't have any negative effect on performance.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: Why limit this to x86?

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// instance is the <see cref="InvalidHandle"/>.
+        /// </summary>
+        internal static T InvalidHandle
+        {
+            get
+            {
+                T currentHandle = Volatile.Read(ref s_invalidHandle);
+                if (currentHandle == null)
+                {
+                    T newHandle = Activator.CreateInstance<T>();
```
sharwell: :question: Does this work for `SafeX509ChainHandle`, which has a private constructor? I would have thought it needed the following:

```csharp
T newHandle = Activator.CreateInstance(typeof(T), true);
```

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
sharwell: If you read a stale value here, the worst that can happen is you create an additional handle. You then dispose the newly created handle because `CompareExchange` will do a volatile read when it tries to update the field.

A volatile read at this specific location provides no benefit, but is more expensive than the equivalent non-volatile read.

sharwell:
@stephentoub Have you had the chance to re-run the allocation test with the cache in place?

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: Actually, I still prefer to leave this as a volatile read.  There's no measurable impact from this on the system.  But if there is a system that allowed for this to be a stale read, then we could end up reading a null value when it's non-null, in which case IsCachedInvalidHandle will return false when it should have returned true, and we'll end up disposing of the cached handle.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: No, it doesn't, and we didn't have a test for it for those paths through SafeX509ChainHandle, which is why everything still passed.  But that overload of CreateInstance also doesn't exist in .NET Core.  I've changed it to find the right ctor via reflection and Invoke it, and I added a test.

stephentoub:
Yes, just measuring the number of those handles that got allocated throughout the tests, this resulted in another 46 fewer than before.

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: > provides no benefit

I think there is a benefit: t highlights the intent, and it prevents the unnecessary instantiation in the case you mention.  Now, those benefits may be very small, but so too is any performance impact.  Across a billion invocations, run repeatedly, I couldn't find any measurably difference between the two.  Are you seeing otherwise?

src/Common/src/Microsoft/Win32/SafeHandles/SafeHandleCache.cs
stephentoub: Actually, I've just changed it to pass in a delegate to create the instance; that way in the future it can be used even when there's no parameterless ctor at all, regardless of visibility.

src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafeX509ChainHandle.cs
```diff
@@ -1,15 +1,11 @@
...
 using System;
-using System.IO;
-using System.Text;
 using System.Diagnostics;
-using System.Globalization;
 using System.Runtime.InteropServices;
-
-using Internal.Cryptography;
-using Internal.Cryptography.Pal;
+using System.Threading;
```
bartonjs: I don't see anything new in this file using Threading... I assume it was left from a previous iteration?

stephentoub: Yup, good catch, it was left over from when I was using LazyInitializer.  I'll remove this.

bartonjs:
@stephentoub Since I don't know what test(s) you are referring to, but I probably should: How would I measure such a thing if I wanted to? :smile:

bartonjs:
LGTM, modulo one potentially errant new using namespace directive.

stephentoub:
> I don't know what test(s) you are referring to

All of the tests for X509 certs, i.e. everything in System.Security.Cryptography.X509Certificates.Tests.  I ran all of the tests and kept track of how many handles were allocated.

> How would I measure such a thing if I wanted to?

I (locally and temporarily) added this to the test project (which doesn't reference System.Console.dll, hence the use of reflection):
```C#
internal static class HandleTracker
{
    private static int s_handleCount;

    public static void Add()
    {
        int count = System.Threading.Interlocked.Increment(ref s_handleCount);
        var writeLines = System.Reflection.IntrospectionExtensions.GetTypeInfo(Type.GetType("System.Console")).GetDeclaredMethods("WriteLine");
        foreach (var writeLine in writeLines)
        {
            var p = writeLine.GetParameters();
            if (p.Length == 1 && p[0].ParameterType == typeof(string))
            {
                writeLine.Invoke(null, new[] { count.ToString() });
                break;
            }
        }
    }
}
```
and then added a call ```HandleTracker.Add();``` to the ctor of the base SafePointerHandle type.

Though in just doing it again to respond to you, I realized I made a small mistake in the numbers earlier... basically the same reduction percentage, but higher numbers overall: before the change, 1661 handles allocated, and after, 1410 handles allocated, for a 15% reduction.

bartonjs:
Ah, gotcha.  I had read it as there was a meta-test that tracked allocations by type and gave a report at the end.

Straightforwardly simple approach :smile:.

Issue #1086, opened 2015-03-04
Feature Request - Add ability to get actual file casing in path


ellismg:
Per @jasonwilliams200OK in dotnet/coreclr#390

Based on this answer: http://stackoverflow.com/a/81493/1712065 (further redirected from: http://stackoverflow.com/a/326153). 

Please add the ability to retrieve path with actual case via `FileInfo` and `DirectoryInfo` classes. The candidate member being `FullPath` and `Name`. Perhaps there is some sophisticated way of getting it from win32 file system API, but [that](http://stackoverflow.com/a/81493/1712065) seems to be a working solution.

Expected:

```c#
using System.IO.FileInfo;
// ..
// actual path was:
// C:\\SharePoint\\scripts\\MasterDeployment.ps1
FileInfo info = new FileInfo("c:\\sharepoint\\scriPts\\maSTerdeploymnet.PS1");
Console.WriteLine(info.FullPath);
// C:\\SharePoint\\scripts\\MasterDeployment.ps1
Console.WriteLine(info.Name);
// MasterDeployment.ps1
```

Actual result:

```c#
using System.IO.FileInfo;
// ..
// actual path was:
// C:\\SharePoint\\scripts\\MasterDeployment.ps1
FileInfo info = new FileInfo("c:\\sharepoint\\scriPts\\maSTerdeploymnet.PS1");
Console.WriteLine(info.FullPath);
// c:\\sharepoint\\scriPts\\maSTerdeploymnet.PS1
Console.WriteLine(info.Name);
// maSTerdeploymnet.PS1
```

jasonwilliams200OK:
NTFS is case-sensitive, its Windows which makes it look bad. :)

Thank you @ellismg. But honestly I opened this issue in coreclr repo because I think the solution should come from lowest possible API, instead of C# manipulation of canonical paths, or making repeated expensive IO operations from a distance. I still standby and vote for the coreclr option. 

No solution is better than pure managed solution (like present), so people don't start relying on bad-performing algorithm out of ignorance.

CoreCLR solution = C/C++ = fast performance
CoreFX solution = .NET runtime = not so fast performance

ellismg:
@jasonwilliams200OK For .NET Core, the source for the lowest level path APIs are in the CoreFX repository.  The implementation in mscorlib.dll in CoreCLR is not what we expose as part of .NET core and will be removed if possible.

[...]

PR #2041, opened 2015-06-14
Add FreeBSD support to System.IO.Compression.ZipFile

janhenke:
Adding support for FreeBSD to System.IO.Compression.ZipFile. PathInternal is identical to Linux, as they share the same file system requirements (paths are case sensitive and can contain all Unicode characters except NULL).

/cc @josteink @stephentoub

### DIFF
--- a/src/System.IO.Compression.ZipFile/System.IO.Compression.ZipFile.sln
+++ b/src/System.IO.Compression.ZipFile/System.IO.Compression.ZipFile.sln
@@ -9,6 +9,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.IO.Compression.ZipFi
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		FreeBSD_Debug|Any CPU = FreeBSD_Debug|Any CPU
+		FreeBSD_Release|Any CPU = FreeBSD_Release|Any CPU
 		Linux_Debug|Any CPU = Linux_Debug|Any CPU
 		Linux_Release|Any CPU = Linux_Release|Any CPU
 		OSX_Debug|Any CPU = OSX_Debug|Any CPU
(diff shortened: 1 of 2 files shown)


Maxwe11:
@dotnet-bot test this please

josteink:
I'm a bit weary about the code-duplication going on and I think we should have a future plan for getting some of the code which is common across Unixes merged into one common file.

However I see the way things are done here adheres to the conventions and practices already in place in corefx for cross-platform support.

As such this LGTM.

janhenke:
I agree it does not seem a good path to add more platforms in the future. For now it seems the way to go, as it is the same structure across the different platforms.

The problem is that OS X is different from the other platforms. I still think this should be merged as it is. We need to discuss that issue in a wider range and implement a consistent solution later.

stephentoub:
I think the PathInternal files should be refactored slightly...

Today we have:
- PathInternal.Windows.cs: contains a HasIllegalCharacters specific to Windows behavior and a GetComparison that's case-insensitive
- PathInternal.Linux.cs contains a HasIllegalCharacters generally applicable to Unix and a GetComparison that's case-sensitive
- PathInternal.OSX.cs contains a HasIllegalCharacters generally applicable to Unix and a GetComparison that's case-insensitive
and now with this PR:
- PathInternal.FreeBSD.cs: contains a HasIllegalCharacters specific to Windows behavior and a GetComparison that's case-sensitive

I suggest we change it to this:
- PathInternal.Windows.cs: contains a HasIllegalCharacters specific to Windows behavior
- PathInternal.Unix.cs: contains a HasIllegalCharacters specific Unix behavior
- PathInternal.CaseSensitive.cs: contains a GetComparison that's case-sensitive
- PathInternal.CaseInsensitive.cs: contains a GetComparison that's case-insensitive
Then we just include the right files in the right parts of the projects:
- TargetsWindows gets PathInternal.Windows.cs and PathInternal.CaseInsensitive.cs
- TargetsUnix gets PathInternal.Unix.cs
- TargetsLinux and TargetsFreeBSD get PathInternal.CaseSensitive.cs
- TargetsOSX gets PathInternal.CaseInsensitive.cs

I think that makes things cleaner and avoids unnecessary code duplication.

(This whole PathInternal.GetComparison thing will potentially change in the future if/when we're able to detect the right casing: #1086).

stephentoub:
If folks agree with that suggestion, I've submitted a PR (#2045) to refactor it accordingly.  This PR could then be updated to use it once it's merged and avoid duplicating the files.

janhenke:
Originally I did not want to touch other code, but the path looks good. I am absolutely in favour of going ahead with it like you outlined.

josteink:
I'm perfectly fine with it. Go ahead :)

jasonwilliams200OK:
@janhenke https://github.com/dotnet/corefx/pull/2045 is merged. Time for a rebase! :sunglasses:

janhenke:
With the re-factored code base the patch became much simpler.

stephentoub:
LGTM

josteink:
This is much better. LGTM.

PR #2043, opened 2015-06-14
Update build instructions to include FreeBSD.

josteink:
These instructions presume that that https://github.com/dotnet/corefx/pull/2030 is committed and merged.

They also correct the error where build on all Unixes is referred to as building on Linux.

CC: @janhenke

PR #2044, opened 2015-06-14
Fix detection of supported Mono-versions.

josteink:
According to akoeplinger, the required mono version is now 4.0.1.44 or higher. Build-script is now updated to reflect this.

This is needed to provide build-support for FreeBSD, but also for Linux-distros with less up to date mono-versions.

CC: @akoeplinger @janhenke

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,8 +23,14 @@ __referenceassemblyroot=$__monoroot/lib/mono/xbuild-frameworks
 __monoversion=$(mono --version | grep "version 4.[1-9]")
 
 if [ $? -ne 0 ]; then
-    echo "Mono 4.1 or later is required to build corefx. Please see https://github.com/dotnet/corefx/blob/master/Documentation/unix-instructions.md for more details."
-    exit 1
+    # if built from tarball, mono only identifies itself as 4.0.1
+    __monoversion=$(mono --version | egrep "version 4.0.1(.[0-9]+)?")
+    if [ $? -ne 0 ]; then
+        echo "Mono 4.0.1.44 or later is required to build corefx. Please see https://github.com/dotnet/corefx/blob/master/Documentation/unix-instructions.md for more details."
+        exit 1
+    else
+        echo "WARNING: Mono 4.0.1.44 or later is required to build corefx. Unable to asses if current version is supported."
+    fi
 fi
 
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @JeremyKuhne

build.sh
```diff
@@ -21,8 +21,14 @@ __referenceassemblyroot=$__monoroot/lib/mono/xbuild-frameworks
...
 if [ $? -ne 0 ]; then
-    echo "Mono 4.1 or later is required to build corefx. Please see https://github.com/dotnet/corefx/blob/master/Documentation/linux-instructions.md for more details."
-    exit 1
+    # if built from tarball, mono only identifies itself as 4.0.1
+    __monoversion=$(mono --version | egrep "version 4.0.1(.[0-9]+)?")
+    if [ $? -ne 0 ]; then
+        echo "Mono 4.0.1.44 or later is required to build corefx. Please see https://github.com/dotnet/corefx/blob/master/Documentation/linux-instructions.md for more details."
+        exit 1
+    else
+        echo "WARNING: Mono 4.0.1.44 or later is required to build corefx. Unable to asses if current version is supported."
```
akoeplinger: Hmm, I'm wondering a bit if we really need to complicate the logic as much or if simply changing the regex to check for Mono>=4.0 is sufficient as well. Those that run Mono from the xamarin repos run 4.0.1.44 anyway now (assuming their system is reasonably updated), which is the vast majority of people I hope. What do you think?

josteink: Given how the complexity of this logic is entirely self-contained I don't consider it a problem. I would as much as possible like to avoid causing uncertainty or confusion among contributors by failing silently or creating needless warnings.

Ofcourse this is a judgement call and if you insist I can simplify it and just check for Mono 4.x instead. Let me know.

josteink:
This PR should probably be postponed until https://github.com/dotnet/corefx/pull/2043 has been committed as there's certain incompatible overlap in `build.sh`.

build.sh
akoeplinger: Yeah, I think you're right, avoiding confusion for contributors should be the driving force behind this. LGTM :+1:

josteink:
Fixed up code to match changes for PR 2043. Should now merge cleanly.

JeremyKuhne:
LGTM - Thanks!

PR #2048, opened 2015-06-15
Improve test coverage of System.IO.Compression

stephentoub:
This PR adds more tests for System.IO.Compression, bringing the code coverage numbers for DeflateStream and GZipStream up to close to 100%.

In some cases, dead code was removed (there's more related to the zlib inflation code paths that are currently unused, but I've left that code in because we may actually want to use them in the future).  

In two places, I added the throwing of an exception that occurs later and deeper in the call tree, but only in release builds, and is something that we catch sooner, especially if we decide to switch any more of the implementation to using unsafe code.

### DIFF
--- a/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
@@ -42,39 +42,32 @@ namespace System.IO.Compression
         {
             if (stream == null)
                 throw new ArgumentNullException("stream");
+            
+            switch (mode)
+            {
+                case CompressionMode.Decompress:
+                    if (!stream.CanRead)
+                    {
+                        throw new ArgumentException(SR.NotReadableStream, "stream");
+                    }
+                    _inflater = new Inflater();
+                    break;
 
-            if (CompressionMode.Compress != mode && CompressionMode.Decompress != mode)
-                throw new ArgumentException(SR.ArgumentOutOfRange_Enum, "mode");
+                case CompressionMode.Compress:
+                    if (!stream.CanWrite)
+                    {
+                        throw new ArgumentException(SR.NotWriteableStream, "stream");
+                    }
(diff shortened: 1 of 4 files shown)


stephentoub:
cc: @FiveTimesTheFun

stephentoub:
@FiveTimesTheFun, mind taking a look?

src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
```diff
@@ -273,9 +259,13 @@ public override int Read(byte[] array, int offset, int count)
...
                 int bytes = _stream.Read(_buffer, 0, _buffer.Length);
-                if (bytes == 0)
+                if (bytes <= 0)
                 {
-                    break;      //Do we want to throw an exception here?
+                    break;
+                }
+                else if (bytes > _buffer.Length)
+                {
+                    throw new InvalidDataException(SR.GenericInvalidData);
```
ericstj: Protecting against a poorly implemented Stream? I'd call that out with a comment.

stephentoub: Yes.  I'll add a comment.  With the implementation today, that'll end up throwing later on but almost by accident, and if we end up doing any perf work that would switch to using unsafe code, those checks could end up going away, hence I wanted to move it up here close to the site of where the bad data enters the system.

ericstj: same below at 381.  What exception were we throwing later on?  This could be a compat concern, though unlikely.

stephentoub: > What exception were we throwing later on?

InvalidDataException, as I've done here.

ericstj:
LGTM

stephentoub:
Thanks, Eric!

PR #2049, opened 2015-06-15
Implement System.IO.FileSystem assembly on FreeBSD

janhenke:
This adds the bits necessary to build the System.IO.FileSystem assembly for FreeBSD. I moved the Interop.posix_fdadvise.cs from Linux to Unix, since that system call is identical between Linux and FreeBSD. That file is only used for Linux and FreeBSD.

/cc @josteink @stephentoub


stephentoub:
LGTM

cc: @ellismg

PR #2051, opened 2015-06-15
[PortablePdb] Local signature in method body, entry point, #Pdb stream

tmat:
Adds local signature token to method body, so that PDB readers have access to them without loading IL stream.
Adds a new #Pdb stream to standalone debug metadata that includes sizes of type-system tables referenced by the debug metadata tables. 
Replaces entry point CDI with an entry in #Pdb stream.


tmat:
@nguerrera FYI.

tmat:
Portable PDB spec updated accordingly: https://github.com/dotnet/roslyn/blob/portable-pdb/docs/specs/PortablePdb-Metadata.md

src/System.Reflection.Metadata/src/System/Reflection/Metadata/PortablePdb/MethodBody.cs
```diff
@@ -37,5 +36,16 @@ public BlobHandle SequencePoints
...
+        /// <summary>
+        /// Returns local signature handle.
+        /// </summary>
+        public StandaloneSignatureHandle LocalSignature
+        {
+            get
+            {
+                return StandaloneSignatureHandle.FromRowId(_reader.GetBlobReader(SequencePoints).ReadCompressedInteger());
+            }
+        }
```
nguerrera: Not directly related to the change or line, but this struct should be renamed. It conflicts with System.Reflection.MethodBody and also ends up looking very strange when there is also a S.R.M.MethodBodyBlock with the IL. We should convey that this has the debug info somehow in the name.

Also, it seems a bit odd that SequencePoints starts with the LocalSignature. Should we give the Blob a more opaque name?

tmat: Agreed. Just not sure what the better names would be. Suggestions?

nguerrera: For the type: MethodBodyDebugInfo? MethodDebugInfo?
For the SequencePoints property, maybe just Blob?

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
```diff
@@ -431,68 +476,91 @@ private void ReadMetadataTableHeader(ref BlobReader memReader, out int[] metadat
...
+            // Size of reference tags in each table.
             this.TableRowCounts = rowCounts;
 
-            // Compute ref sizes for tables that can have pointer tables for it
-            int fieldRefSize = referenceSizes[(int)TableIndex.FieldPtr] > SmallIndexSize ? LargeIndexSize : referenceSizes[(int)TableIndex.Field];
-            int methodRefSize = referenceSizes[(int)TableIndex.MethodPtr] > SmallIndexSize ? LargeIndexSize : referenceSizes[(int)TableIndex.MethodDef];
-            int paramRefSize = referenceSizes[(int)TableIndex.ParamPtr] > SmallIndexSize ? LargeIndexSize : referenceSizes[(int)TableIndex.Param];
-            int eventRefSize = referenceSizes[(int)TableIndex.EventPtr] > SmallIndexSize ? LargeIndexSize : referenceSizes[(int)TableIndex.Event];
-            int propertyRefSize = referenceSizes[(int)TableIndex.PropertyPtr] > SmallIndexSize ? LargeIndexSize : referenceSizes[(int)TableIndex.Property];
+            // TODO: Shouldn't XxxPtr table be always the same size or smaller than the corresponding Xxx table?
```
nguerrera: Is this TODO just for extra validation? Should this be tracked by a bug?

tmat: I'll file a bug. Do we do any testing o Ptr tables at all? The ECMA spec doesn't define them at all. Perhaps we should write a spec for them. I'm not even sure in what scenarios they may show up.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
```diff
@@ -92,20 +92,33 @@ public unsafe MetadataReader(byte* metadata, int length, MetadataReaderOptions o
             _options = options;
             this.utf8Decoder = utf8Decoder;
 
-            BlobReader memReader = new BlobReader(this.Block);
-
-            this.ReadMetadataHeader(ref memReader);
+            BlobReader headerReader = new BlobReader(this.Block);
```
nguerrera: Nit: I'd use var here.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
```diff
@@ -431,68 +476,91 @@ private void ReadMetadataTableHeader(ref BlobReader memReader, out int[] metadat
...
-        private const int SmallIndexSize = 2;
-        private const int LargeIndexSize = 4;
-
-        private void InitializeTableReaders(MemoryBlock metadataTablesMemoryBlock, int[] compressedRowCounts)
+        private void ReadStandalonePortablePdbStream(MemoryBlock block, out DebugMetadataHeader debugMetadataHeader, out int[] externalTableRowCounts)
         {
-            // Only sizes of tables present in metadata are recorded in rowCountCompressedArray.
-            // This array contains a slot for each possible table, not just those that are present in the metadata.
-            int[] rowCounts = new int[TableIndexExtensions.Count];
+            BlobReader reader = new BlobReader(block);
```
nguerrera: Same nit.

nguerrera:
:+1:

nguerrera:
Feel free to leave the naming issue to a separate discussion / API review before going to master.

tmat:
@dotnet-bot Retest this, please.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/PortablePdb/MethodBody.cs
tmat: OK, I think we can make it Blob (or Data) and add GetSequencePointReader() method so that it is obvious how to get sequence points out of it.

MethodDebugInfo sounds good. We should also rename CustomDebugInformation to CustomDebugInfo if we are going to use the short "Info" form.

Will leave renaming for another time as you suggest :)

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
tmat: Done.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
tmat: Fixed.

tmat:
@dotnet-bot Retest this, please.

PR #2053, opened 2015-06-15
Fix exception message differences when deserializing empty string

khdang:
There is differences in the exception message when deserializing empty string using DataContractSerializer in NetCore/NetNative compared to Desktop.
This is due to XmlDictionaryReader uses different kind of XmlReader in one of the CreateTextReader overloads. This PR updates the overload to initialize to the same XmlUtf8TextReader.

### DIFF
--- a/src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
+++ b/src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
@@ -1271,6 +1271,16 @@ public static class DataContractJsonSerializerTests
         Assert.True(output.OnDeserializedMethodInvoked, "output.OnDeserializedMethodInvoked is false");
     }
 
+    [Fact]
+    public static void DCJS_DeserializeEmptyString()
+    {
+        var serializer = new DataContractJsonSerializer(typeof (object));
+        Assert.Throws<SerializationException>(() =>
+        {
+            serializer.ReadObject(new MemoryStream());
+        });
+    }
+
     private static T SerializeAndDeserialize<T>(T value, string baseline, DataContractJsonSerializerSettings settings = null, Func<DataContractJsonSerializer> serializerFactory = null)
     {
         DataContractJsonSerializer dcjs;
--- a/src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
+++ b/src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
@@ -1703,6 +1703,16 @@ public static class DataContractSerializerTests
         Assert.StrictEqual(value.TestProperty, actual.TestProperty);
     }
 
+    [Fact]
+    public static void DCS_DeserializeEmptyString()
+    {
+        var serializer = new DataContractSerializer(typeof (object));
+        Assert.Throws<XmlException>(() =>
+        {
+            serializer.ReadObject(new MemoryStream());
+        });
+    }
+
     private static T SerializeAndDeserialize<T>(T value, string baseline, DataContractSerializerSettings settings = null, Func<DataContractSerializer> serializerFactory = null)
     {
         DataContractSerializer dcs;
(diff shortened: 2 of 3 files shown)


khdang:
@SGuyGe @zhenlan

src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
```diff
@@ -1271,6 +1271,20 @@ public static void DCJS_SerializationEvents()
         Assert.True(output.OnDeserializedMethodInvoked, "output.OnDeserializedMethodInvoked is false");
     }
 
+    [Fact]
+    public static void DCJS_DeserializeEmptyString()
+    {
+        var exception = Assert.Throws<XmlException>(() =>
+        {
+            var serializer = new DataContractSerializer(typeof(object));
```
zhenlan: Should this be DCJS instead?

khdang: Yes, this is a typo. After fixing this I've realized this issue is only for DCS. DCJS throws SerializationException which aligns with the behavior on Desktop.

src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
```diff
@@ -1271,6 +1271,19 @@ public static void DCJS_SerializationEvents()
         Assert.True(output.OnDeserializedMethodInvoked, "output.OnDeserializedMethodInvoked is false");
     }
 
+    [Fact]
+    public static void DCJS_DeserializeEmptyString()
+    {
+        var exception = Assert.Throws<SerializationException>(() =>
```
stephentoub: Where does the exception actually come from... ReadObject?  If so, should we wrap just that call in the Assert?

src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
```diff
@@ -1703,6 +1703,20 @@ public static void DCS_DerivedException()
         Assert.StrictEqual(value.TestProperty, actual.TestProperty);
     }
 
+    [Fact]
+    public static void DCS_DeserializeEmptyString()
+    {
+        var exception = Assert.Throws<XmlException>(() =>
+        {
+            var serializer = new DataContractSerializer(typeof(object));
```
stephentoub: Same question.

stephentoub:
Two nits, otherwise LGTM.

src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
khdang: Yes, exception comes from ReadObject. I've updated the tests.

src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
```diff
@@ -1703,6 +1703,17 @@ public static void DCS_DerivedException()
...
 
+    [Fact]
+    public static void DCS_DeserializeEmptyString()
+    {
+        var serializer = new DataContractSerializer(typeof (object));
+        var exception = Assert.Throws<XmlException>(() =>
+        {
+            serializer.ReadObject(new MemoryStream());
+        });
+        Assert.StrictEqual("Unexpected end of file.", exception.Message);
```
stephentoub: Does this message get localized?

khdang: Yes it is. I think I will do this check with stack trace instead.

stephentoub: That could be affected by inlining as well as easily by refactorings. What are you trying to verify here?

khdang: The issue is there are differences in the message and location of the XmlException in NetCore compared to Desktop. So I want to ensure that this scenario throws an XmlException that is not too different from Desktop.

stephentoub:
Thanks.  Revised version LGTM.

zhenlan:
LGTM. Thanks.

PR #2054, opened 2015-06-15
Update string values in OSPlatform.

Priya91:
cc @akoeplinger


stephentoub:
LGTM.  It'll be nice when we can start using C# 6 in corefx and are able to use ```nameof``` to avoid problems like this.

PR #2060, opened 2015-06-15
Small optimization in BigInteger

KrzysztofCwalina:
While debugging something, I noticed that we run same BigInteger ctor repeatedly in a loop.
This change runs the ctor once and caches the constant value to be used inside the loop.


KrzysztofCwalina:
@axelheer

stephentoub:
Nice. LGTM.

axelheer:
:+1:

PR #2063, opened 2015-06-16
Add TextInfo casing tests

ellismg:
An initial set of TextInfo casing tests. I looked at the scinerios we
have coverage for in ToF, but wrote the actual tests from scratch in
idomatic XUnit style.

In addition, I have added some specific tests for the Turkish I casing,
since the characters have special casing behavior in Turkish and some
other locales as well as test cases to ensure we have the simple casing
in places where the Unicode standard normally defines special cases
which are either context sensitive or would change the number of
codepoints in a cased string.

### DIFF
--- a/src/System.Globalization/tests/System.Globalization.Tests.csproj
+++ b/src/System.Globalization/tests/System.Globalization.Tests.csproj
@@ -1,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{7F5F5134-00FE-4DE8-B20C-3DA8BA2EBA68}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <RootNamespace>System.Globalization.Tests</RootNamespace>
+    <AssemblyName>System.Globalization.Tests</AssemblyName>
+    <RestorePackages>true</RestorePackages>
+  </PropertyGroup>
+  <!-- Default configurations to help VS understand the configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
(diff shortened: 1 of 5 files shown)


src/System.Globalization/tests/TextInfo.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        Assert.Equal("THIS IS A LONGER TEST CASE", ti.ToUpper("this is a longer test case"));
+        Assert.Equal("THIS IS A LONGER MIXED CASE TEST CASE", ti.ToUpper("this Is A LONGER mIXEd casE test case"));
+
+        Assert.Equal("THIS \t HAS \t SOME \t TABS", ti.ToUpper("this \t HaS \t somE \t TABS"));
+
+        Assert.Equal('1', ti.ToUpper('1'));
+        Assert.Equal("123", ti.ToUpper("123"));
+
+        Assert.Equal("EMBEDED\0NULL\0BYTE\0", ti.ToUpper("embeded\0NuLL\0Byte\0"));
```
stephentoub: Extreme nit: embedded

src/System.Globalization/tests/TextInfo.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        Assert.Equal('\u00D3', ti.ToUpper('\u00F3'));
+        Assert.Equal("\u00D3", ti.ToUpper("\u00F3"));
+
+        Assert.Equal('\u2603', ti.ToUpper('\u2603'));
```
stephentoub: For these specific unicode character tests, is it worth adding a comment indicating what it is / why it's chosen?

ellismg: Can do (the first is a character that actually cases (an o with an accent mark), the second is a character that does not (a snowman)).

src/System.Globalization/tests/TextInfo.cs
ellismg: Fixed.

src/System.Globalization/tests/System.Globalization.Tests.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
stephentoub: Could you add a .sln with this as well?

stephentoub:
A few minor comments, but otherwise LGTM. Thanks!

src/System.Globalization/tests/TextInfo.cs
ellismg: I also added a test for a character outside the BMP that does have a casepair.

src/System.Globalization/tests/System.Globalization.Tests.csproj
ellismg: Can do.

ellismg:
Thanks Steve, I addressed all your feedback, hopefully fixed the issue that was causing the change to fail in CI (which I didn't hit locally because I am on Windows 10) and squashed.

stephentoub:
Thanks, Matt.  Assuming it passes, LGTM.

PR #2066, opened 2015-06-16
Remove <AllowUnsafeBlocks/> from various test projects

stephentoub:
Most test projects don't need unsafe code, but many were forced to allow unsafe blocks due to Interop.PlatformDetection.cs using pointers, and others were allowing it for no apparent reason.  I've modified Interop.PlatformDetection.cs to not use unsafe code, and updated the relevant test projects accordingly. (Eventually we'll switch it over to using the new RuntimeInformation assembly, anyway.)


ellismg:
LGTM.

stephentoub:
Thanks, Matt.

PR #2067, opened 2015-06-16
Add more tests for System.Console

### LINKED ISSUE #923: Improve code coverage for System.Console
According to the current code coverage report, this library is only at 58%.  Our general goal is closer to 90%, though this will depend on the library.  We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

stephentoub:
This improves code coverage as much as we can reasonably, given the difficulty in automating certain console-related functionality in inner loop testing, e.g. allowing ctrl-c to take down the testing process.

As part of this, I removed a bit of dead code, but also added some more overrides to ConsoleEncoding, which was missing some that should have been delegating to the wrapped encoding (some of the new tests failed without these).

Fixes #923.

### DIFF
--- a/src/System.Console/tests/SyncTextReader.cs
+++ b/src/System.Console/tests/SyncTextReader.cs
@@ -66,8 +66,13 @@ public class SyncTextReader
         }
         finally
         {
+            TextWriter oldWriter = Console.Out;
             Console.SetOut(savedStandardOutput);
+            oldWriter.Dispose();
+
+            TextReader oldReader = Console.In;
             Console.SetIn(savedStandardInput);
+            oldReader.Dispose();
         }
     }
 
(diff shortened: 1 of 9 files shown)


stephentoub:
cc: @pallavit, @ellismg

src/System.Console/tests/SyncTextReader.cs
```diff
@@ -66,8 +66,15 @@ private static void Test(string content, Action action)
         }
         finally
         {
-            Console.SetOut(savedStandardOutput);
-            Console.SetIn(savedStandardInput);
+            using (TextWriter oldWriter = Console.Out)
```
ellismg: Can you add some comments around this using pattern here?  It's a little confusing to me since you don't use oldWriter in the body of the block (I think you are doing this for an exception safe dispose on the oldWriter?)

stephentoub: > I think you are doing this for an exception safe dispose on the oldWriter

Yes.  I can just remove the using if it'll make it clearer, and just call Dispose directly.

ellismg:
Aside from the using thing (I'm fine for you to continue to use using, just have a comment about what's going on), LGTM.

stephentoub:
Thanks, Matt.

stephentoub:
@dotnet-bot test this please

PR #2081, opened 2015-06-17
Update NetCore DCJS to use the same code with NetNative DCJS

khdang:
DCJS implementation in NetCore at the moment is based on the JavascriptSerializer/Deserializer which is rather different from the ones used in NetNative and on Desktop. This introduces unneccessary effort in making improvement and providing consistent behaviors.
This change switches DCJS in NetCore to the same code that is used in NetNative with MERGE_DCJS directive. If there is no issues I will follow to clean up this directive and unused code.

@SGuyGe @zhenlan @roncain @shmao


src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/CodeGenerator.cs
```diff
@@ -79,6 +79,33 @@ private static MethodInfo ArraySetValue
...
+#if MERGE_DCJS
+        [SecurityCritical]
+        private static MethodInfo s_objectToString;
+        private static MethodInfo ObjectToString
+        {
+            [SecuritySafeCritical]
+            get
+            {
+                if (s_objectToString == null)
+                    s_objectToString = typeof(object).GetMethod("ToString", new Type[0]);
```
stephentoub: ```Array.Empty<Type>()```?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Globals.cs
```diff
@@ -591,6 +591,34 @@ internal static Type TypeOfSafeSerializationManager
         }
 #endif
 
+#if MERGE_DCJS
+        [SecurityCritical]
+        private static object[] s_emptyObjectArray;
```
stephentoub: Why have a static field for storing the result of ```Array.Empty<object>()```?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/JsonFormatGeneratorStatics.cs
```diff
@@ -0,0 +1,475 @@
...
+            }
+        }
+        public static MethodInfo WriteEndElementMethod
+        {
+            [SecuritySafeCritical]
+            get
+            {
+                if (s_writeEndElementMethod == null)
+                {
+                    s_writeEndElementMethod = typeof(XmlWriterDelegator).GetMethod("WriteEndElement", Globals.ScanAllMembers, new Type[] { });
```
stephentoub: ```Array.Empty<Type>()``` ?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/BitFlagsGenerator.cs
```diff
@@ -0,0 +1,114 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if MERGE_DCJS
```
stephentoub: Rather than surrounding whole files with a big ```#if```, should the files just be conditionally included in the .csproj?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/DataContractSerializer.cs
```diff
@@ -66,6 +66,8 @@ public DataContractSerializer(Type type, XmlDictionaryString rootName, XmlDictio
 
 #if NET_NATIVE
         public DataContractSerializer(Type type, IEnumerable<Type> knownTypes, int maxItemsInObjectGraph, bool ignoreExtensionDataObject, bool preserveObjectReferences)
+#elif MERGE_DCJS
+        internal DataContractSerializer(Type type, IEnumerable<Type> knownTypes, int maxItemsInObjectGraph, bool ignoreExtensionDataObject, bool preserveObjectReferences)
```
shmao: This "elif" doesn't seem right to me. Should the "endif" be right after line 70?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/DataContractJsonSerializer.cs
```diff
@@ -328,10 +328,12 @@ public void WriteObject(Stream stream, object graph)
 #endif
         }
 
+#if !MERGE_DCJS
```
shmao: This method was used by both NetNative and NetCore. Does this mean NetCore does not need this method?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/XmlObjectSerializerReadContextComplexJson.cs
```diff
@@ -10,20 +10,25 @@
...
 using DataContractDictionary = System.Collections.Generic.Dictionary<System.Xml.XmlQualifiedName, System.Runtime.Serialization.DataContract>;
+#if !NET_NATIVE
+using ExtensionDataObject = System.Object;
+#endif
 
 namespace System.Runtime.Serialization.Json
 {
 #if NET_NATIVE
     public class XmlObjectSerializerReadContextComplexJson : XmlObjectSerializerReadContextComplex
+#elif MERGE_DCJS
```
shmao: Why this class needs to be internal in NetCore? Can it be public? Or can it be internal in NetNative?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/XmlObjectSerializerWriteContextComplex.cs
```diff
@@ -51,18 +51,31 @@ internal override bool WriteClrTypeInfo(XmlWriterDelegator xmlWriter, Type dataC
             return false;
         }
 
+#if NET_NATIVE
```
shmao: My understanding of this change is that, regardless of the consolidation, we'd like to make this method internal in NetCore. Am I right? I just wondered why we need to have the method have different visibility in NetNative and NetCore while our goal is to consolidate them.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/DataContractSerializer.cs
khdang: Yes, this is not right. My intention is that this constructor only exists in NetNative or with the new DCJS with different visibility.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/CodeGenerator.cs
khdang: Thanks.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Globals.cs
khdang: Yes, this is not needed.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/BitFlagsGenerator.cs
khdang: That sounds like a better way to include some of these files. Thanks.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/DataContractJsonSerializer.cs
khdang: Actually NetNative doesn't use this method since it is calling to the _jsonSerializer which is of type JavascriptSerializer. So this method is not needed in NetNative or when we're using the NetNative DCJS.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/XmlObjectSerializerReadContextComplexJson.cs
khdang: It needs to be public in NetNative since the serialization component has reference to it. We're figuring out a way to eliminate that. When it is done we would want this and a couple of other types to be internal.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/XmlObjectSerializerWriteContextComplex.cs
khdang: As previous comment, some of these types/members are referenced in NetNative serialization component. We will have iterations to get rid of the directives and minimize these visibility differences.

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/JsonFormatGeneratorStatics.cs
khdang: Updated.

khdang:
@dotnet-bot test this please

khdang:
@dotnet-bot test this please

PR #2082, opened 2015-06-17
Improve reliability of System.IO.MemoryMappedFiles on Unix

stephentoub:
When a MemoryMappedFile is created backed by memory (rather than by a file), in most cases we create a temporary backing store for it to be used by mmap when creating the views.  That backing store may be either a POSIX shared memory object or a temporary file, depending on the platform, and it enables us to better match the exposed API, e.g. enabling multiple views over the same data where some have copy-on-write access, etc.  However, the current implementation only unlinks that backing store when it's Disposed; this means that if the process crashes, the backing store could be left around indefinitely (in the case of a memory object, until restart, and in the case of a temporary file, until tmps are cleared out).

This change moves the unlink up to be done immediately after the backing store is created and opened, and then cleans up a bunch of cruft left over that's no longer necessary.  This has multiple advantages: in particular, we'll clean up the backing store much more reliably in the face of crashes, and the name will only be visible to others on the system for the briefest of moments.

As I was changing these files, I took the opportunity to rename them away from the initial platform they were targeting and instead based on their characteristics.  This should make it easier for other systems, e.g. FreeBSD, to use whichever implementation best suits its need.


stephentoub:
cc: @ellismg, @nguerrera

stephentoub:
@ellismg, mind taking a look?

ellismg:
Not sure how hard it is to write a test for the new precondition you added, but if it's something you can do it would be nice.  LGTM otherwise.

stephentoub:
Thanks, Matt.  I'm not sure what I was thinking when I added that check, but it's not actually possible for that condition to be met at that point (it's effectively checked earlier in the function), so I've removed it.

Issue #1432, opened 2015-04-18
standardNumericalFunctions_PowTest failed in CI


stephentoub:
```
ErRoR!!! Actual Complex (-Infinity, -6.19524038874532E+307) imaginary part to be equal to Double:-Infinity
  Error pOw-Err2534! Pow ((-1.28872245606257E+308, 2.39662253395911E+307), (0.999164775479196, -0.335819169569676))
     System.Numerics.Tests.standardNumericalFunctions_PowTest.RunTests_RandomValidValues [FAIL]
        Verification Failed
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime.Numerics\tests\Complex\standardNumericalFunctions_Pow.cs(34,0): at System.Numerics.Tests.standardNumericalFunctions_PowTest.VerifyPow(Double a, Double b, Double c, Double d)
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime.Numerics\tests\Complex\standardNumericalFunctions_Pow.cs(110,0): at System.Numerics.Tests.standardNumericalFunctions_PowTest.RunTests_PowComplex(Double a, Double b)
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime.Numerics\tests\Complex\standardNumericalFunctions_Pow.cs(149,0): at System.Numerics.Tests.standardNumericalFunctions_PowTest.RunTests_RandomValidValues()
```

stephentoub:
This failed again:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release/744/console 
```
System.Numerics.Tests.standardNumericalFunctions_PowTest.RunTests_RandomValidValues [FAIL]
        Pow ((-1.47312966505531E+308, 1.08130045515798E+307), (0.999750766903046, -0.309201333815791))
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime.Numerics\tests\Complex\ComplexTestSupport.cs(170,0): at ComplexTestSupport.Support.VerifyRealImaginaryProperties(Complex complex, Double real, Double imaginary, String message)
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime.Numerics\tests\Complex\standardNumericalFunctions_Pow.cs(29,0): at System.Numerics.Tests.standardNumericalFunctions_PowTest.VerifyPow(Double a, Double b, Double c, Double d)
[...]

PR #2083, opened 2015-06-17
Fix Exception type thrown in Vector<T> ctor and CopyTo methods

mellinoe:
When the methods below are handled specially by the JIT, a NullReferenceException is produced, rather than the ArgumentNullException we are throwing in IL. Conceptually, it makes more sense to throw an ArgumentNullException here, but that would degrade the JIT's performance as, currently, the constructor is doing very little work (should be as simple as a load/store instruction for some cases).

* Vector\<T\>.ctor
* Vector\<T\>.CopyTo
* Vector2/3/4.CopyTo

To help diagnosability a bit, you will still get a reasonable exception message when using the IL version (so in Debug builds): "The method was called with a null array argument."

Thoughts on this? It is possible we could go the other way, and have the JIT throw the ArgumentNullException.

@CarolEidt , @nguerrera , @terrajobst

### DIFF
--- a/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs
+++ b/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs
@@ -363,7 +363,8 @@ namespace System.Numerics
         {
             if (values == null)
             {
-                throw new ArgumentNullException("values");
+                // Match the JIT's exception type here. For perf, a NullReference is thrown instead of an ArgumentNull.
+                throw new NullReferenceException(SR.Arg_NullArgumentNullRef);
             }
             if (index < 0 || (values.Length - index) < Count)
             {
@@ -772,7 +773,8 @@ namespace System.Numerics
         {
             if (destination == null)
             {
-                throw new ArgumentNullException("values");
+                // Match the JIT's exception type here. For perf, a NullReference is thrown instead of an ArgumentNull.
+                throw new NullReferenceException(SR.Arg_NullArgumentNullRef);
             }
             if (startIndex < 0 || startIndex >= destination.Length)
             {
--- a/src/System.Numerics.Vectors/src/System/Numerics/Vector4_Intrinsics.cs
+++ b/src/System.Numerics.Vectors/src/System/Numerics/Vector4_Intrinsics.cs
@@ -106,11 +106,18 @@ namespace System.Numerics
         public void CopyTo(Single[] array, int index)
         {
             if (array == null)
-                throw new ArgumentNullException("values");
+            {
+                // Match the JIT's exception type here. For perf, a NullReference is thrown instead of an ArgumentNull.
+                throw new NullReferenceException(SR.Arg_NullArgumentNullRef);
+            }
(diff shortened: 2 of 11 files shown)


mellinoe:
Potential fix for #2084

nguerrera:
I'm OK with this direction since SIMD intrinsics are all about perf. So LGTM.

src/System.Numerics.Vectors/src/System/Numerics/Vector4_Intrinsics.cs
```diff
@@ -106,7 +106,8 @@ public void CopyTo(Single[] array)
         public void CopyTo(Single[] array, int index)
         {
             if (array == null)
-                throw new ArgumentNullException("values");
+                // Match the JIT's exception type here. For perf, a NullReference is thrown instead of an ArgumentNull.
```
nguerrera: Personal style opinion, but I'd put braces around single-statement bodies that also have comment lines associated with them.

mellinoe: I just moved them all to have braces because I prefer that anyways, and those were the only arg checks in the file so it wasn't very much code to change.

CarolEidt:
I like this change - the JIT could potentially optimize way some of the explicit checks, but won't always be able to, so I prefer the NullReferenceException.

mellinoe:
@dotnet-bot  Test this, please.

stephentoub:
(CI failed due to unrelated #1432)

stephentoub:
LGTM

src/System.Numerics.Vectors/src/System/Numerics/Vector.cs
```diff
@@ -363,7 +363,8 @@ public unsafe Vector(T[] values, int index)
         {
             if (values == null)
             {
-                throw new ArgumentNullException("values");
+                // Match the JIT's exception type here. For perf, a NullReference is thrown instead of an ArgumentNull.
+                throw new NullReferenceException(SR.Arg_NullArgumentNullRef);
             }
             if (index < 0 || (values.Length - index) < Count)
```
VSadov: I wonder whether the explicit check is even needed. 
Considering the perf is important, could you just do "int dummy = values.Length" and let the hardware throw NRE?

mellinoe: That would be an option, but I actually like at least giving the exception message here. It will help (a bit) with figuring out your error, whereas with the default exception, you might not immediately know what happened (should still be obvious based on which method is throwing it).

Also, since this is just the IL version, perf here doesn't really matter. This type will be pretty slow in general unless you are running with JIT intrinsics.

VSadov: just noticed that if I was expecting to use these without SIMD, I'd definitely appreciate the perf more than telling me which of the args was the cause of NRE, if I happen to get one.

It is ok if that is not a targeted scenario.

PR #2085, opened 2015-06-17
Fix serialization of generic queue to align with NetNative

khdang:
Generic queue is serialized differently in NetCore compared to NetNative since it skips the queue's _syncRoot property of type object. 
This change is to fix the behavior when serializing property of type object in general. 
As Serialization of generic queue on desktop is controlled with Serializable, NonSerialized attributes, which are not available in NetCore, this results in a different load compared to desktop. I have verified that cross platform serialization/deserialization works fine.

cc: @SGuyGe @zhenlan @shmao

### DIFF
--- a/src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
+++ b/src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
@@ -562,29 +562,31 @@ public static class DataContractSerializerTests
         Assert.True(y.RO1.Count == 1, getCheckFailureMsg("RO1"));
         Assert.True(y.RO2.Count == 1, getCheckFailureMsg("RO2"));
 
-
-
-
         foreach (var item in x.F1)
         {
             Assert.True(y.F1.Contains(item), getCheckFailureMsg("F1"));
         }
+
         foreach (var item in x.F2)
         {
             Assert.True(y.F2.Contains(item), getCheckFailureMsg("F2"));
         }
+
         foreach (var item in x.P1)
         {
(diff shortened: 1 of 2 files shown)


src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
```diff
@@ -1444,7 +1444,14 @@ public static void DCS_GenericQueue()
     {
         Queue<int> value = new Queue<int>();
         value.Enqueue(1);
-        var deserializedValue = SerializeAndDeserialize<Queue<int>>(value, "<QueueOfint xmlns=\"http://schemas.datacontract.org/2004/07/System.Collections.Generic\" xmlns:i=\"http://www.w3.org/2001/XMLSchema-instance\"><_array xmlns:a=\"http://schemas.microsoft.com/2003/10/Serialization/Arrays\"><a:int>1</a:int><a:int>0</a:int><a:int>0</a:int><a:int>0</a:int></_array><_head>0</_head><_size>1</_size><_tail>1</_tail><_version>2</_version></QueueOfint>");
+        object syncRoot = ((ICollection)value).SyncRoot;
```
stephentoub: What's the purpose of accessing SyncRoot... to lazily-initialize it, and if so, does that affect the serialization?  If so, do you need a test for when it hasn't been initialized?

khdang: Yes, it does affect the serialized string. I will update the test.

PR #2086, opened 2015-06-17
Fix Matrix/Plane/Quaternion field offset tests, Issue #1002

### LINKED ISSUE #1002: Using 64-bit CoreCLR causes four test failures for System.Numerics.Vectors
After updating to CoreCLR 1.0.2 the following System.Numerics.Vectors tests are now failing.

QuaternionFieldOffsetTest
PlaneFieldOffsetTest
Matrix4x4FieldOffsetTest
Matrix3x2FieldOffsetTest

The tests fail with a NullReferenceException when attempting to get field offsets.  Here's an example.

Quaternion* ptr = (Quaternion*)0;
Assert.Equal(new IntPtr(0), new IntPtr(&ptr->X));

This used to work and works as expected on desktop CLR.  This is likely a regression in the JIT.

mellinoe:
This fixes #1002. For some reason these tests work fine on x86, but doing these pointer accesses on x64 fails (understandably). I've re-worked the tests so they do roughly the same thing as before but in a less-strange manner.

### DIFF
--- a/src/System.Numerics.Vectors/tests/Matrix3x2Tests.cs
+++ b/src/System.Numerics.Vectors/tests/Matrix3x2Tests.cs
@@ -1009,19 +1009,22 @@ namespace System.Numerics.Tests
 
         // A test to make sure the fields are laid out how we expect
         [Fact]
-        [ActiveIssue(1002)]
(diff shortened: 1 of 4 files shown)


stephentoub:
LGTM

PR #2090, opened 2015-06-18
Optimize ImmutableArray<T>.RemoveRange for 0/1 items input

### LINKED ISSUE #2029: ImmutableArray<T>.RemoveRange should not allocate if length is 0
Currently `ImmutableArray<T>.RemoveRange` always creates a new underlying array. If `length` is 0, no changes are made to the array so the method should simply return `self`.

stephentoub:
Adds an optimization for the RemoveRange(int,int) overload to not allocate if there are 0 items in the input range, and another for the RemoveRange(ImmutableArray<int>, ...) overloads to minimize allocations for when there are 0 or 1 items in the input array.

Prior to that, I also added some RemoveRange tests to improve the coverage for these different inputs; without the additional tests, which are basically copies of the RemoveRange(IEnumerable<T>) tests but with ImmutableArray<T> inputs, some of the new branches were uncovered.  I also added a few more argument validation tests, which highlight some inconsistencies in the exceptions that get thrown when various inputs are incorrect, but I didn't make any effort to change these.  The tests all pass before and after the subsequent commit of the optimizations.

Fixes #2029.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -756,6 +756,11 @@ namespace System.Collections.Immutable
             Requires.Range(index >= 0 && index < self.Length, "index");
             Requires.Range(length >= 0 && index + length <= self.Length, "length");
 
+            if (length == 0)
+            {
+                return self;
+            }
+
             T[] tmp = new T[self.Length - length];
             Array.Copy(self.array, 0, tmp, 0, index);
             Array.Copy(self.array, index + length, tmp, index, self.Length - index - length);
@@ -817,7 +822,7 @@ namespace System.Collections.Immutable
         [Pure]
         public ImmutableArray<T> RemoveRange(ImmutableArray<T> items)
         {
-            return this.RemoveRange(items.array);
+            return this.RemoveRange(items, EqualityComparer<T>.Default);
         }
 
         /// <summary>
@@ -833,7 +838,22 @@ namespace System.Collections.Immutable
         [Pure]
         public ImmutableArray<T> RemoveRange(ImmutableArray<T> items, IEqualityComparer<T> equalityComparer)
         {
-            return this.RemoveRange(items.array, equalityComparer);
+            var self = this;
+            Requires.NotNull(items.array, "items");
+
+            if (items.IsEmpty)
+            {
+                self.ThrowNullRefIfNotInitialized();
+                return self;
+            }
+            else if (items.Length == 1)
+            {
+                return self.Remove(items[0], equalityComparer);
+            }
+            else
+            {
+                return self.RemoveRange(items.array, equalityComparer);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @AArnott, @sharwell

sharwell:
:+1:

jasonwilliams200OK:
:+1: 
@stephentoub, would you please comment on similar fixes at https://github.com/dotnet/coreclr/pull/637 ? :)
IMO a millisecond save in `String.{whatever}` is a big win, since people use those methods carelessly all the time.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -756,6 +756,11 @@ public ImmutableArray<T> RemoveRange(int index, int length)
             Requires.Range(index >= 0 && index < self.Length, "index");
             Requires.Range(length >= 0 && index + length <= self.Length, "length");
 
+            if (length == 0)
+            {
+                return self;
```
AArnott: What if `this.array == null`? Did this diff change the behavior of RemoveRange to return an uninitialized instance instead of throwing?

AArnott: Probably no change, since `self.Length` would have thrown. I just thought I'd check... is there a test for the case of calling this on an uninitialized instance?

stephentoub: We're already accessing self.Length earlier, so that behavior stays the same.

stephentoub: > is there a test for the case of calling this on an uninitialized instance?

Yes, in this PR :)

AArnott:
:+1:

stephentoub:
Thanks, guys.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -833,7 +838,22 @@ public ImmutableArray<T> RemoveRange(ImmutableArray<T> items)
...
+            Requires.NotNull(items.array, "items");
+
+            if (items.IsEmpty)
+            {
+                self.ThrowNullRefIfNotInitialized();
+                return self;
+            }
+            else if (items.Length == 1)
+            {
+                return self.Remove(items[0], equalityComparer);
```
AArnott: Why is it safe to assume that `items` is a 0-index based array?

AArnott: Should we have tests for passing in arrays that are not 0-based?

sharwell: :question: How would you even do that?

stephentoub: Because (correct me if I'm wrong) the library only accepts T[], not System.Array. The latter can have non-zero lower bounds, but the former is always 0-based.

AArnott: Ah, ok that makes sense.

AArnott: @sharwell I was thinking about [Array.CreateInstance](https://msdn.microsoft.com/en-us/library/x836773a(v=vs.110).aspx) but as @stephentoub points out that doesn't let you then cast the resulting `System.Array` to `T[]` so it's not relevant to this discussion after all.

PR #2092, opened 2015-06-18
Replace MRE with MRES in PipeAsyncResult class

Maxwe11:
Replace ManualResetEvent with ManualResetEventSlim in PipeAsyncResult class.
Internal usage of MRES won't allocate MRE, but it will be allocated if user asked `IAsyncResult.AsyncWaitHandle` property.

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
@@ -258,16 +258,8 @@ namespace System.IO.Pipes
 
             // Create and store async stream class library specific data in the 
             // async result
-            PipeAsyncResult asyncResult = new PipeAsyncResult();
-            asyncResult._threadPoolBinding = _threadPoolBinding;
-            asyncResult._userCallback = callback;
-            asyncResult._userStateObject = state;
-
+            PipeAsyncResult asyncResult = new PipeAsyncResult(_threadPoolBinding, callback, state);
             IOCancellationHelper cancellationHelper = state as IOCancellationHelper;
-
-            // Create wait handle and store in async result
-            ManualResetEvent waitHandle = new ManualResetEvent(false);
-            asyncResult._waitHandle = waitHandle;
 
             NativeOverlapped* intOverlapped = _threadPoolBinding.AllocateNativeOverlapped((errorCode, numBytes, pOverlapped) =>
             {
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -355,20 +347,14 @@ private unsafe void EndWaitForConnection(IAsyncResult asyncResult)
...
             }
 
-            // Obtain the WaitHandle, but don't use public property in case we
-            // delay initialize the manual reset event in the future.
-            WaitHandle wh = afsar._waitHandle;
-            if (wh != null)
+            // We must block to ensure that ConnectionIOCallback has completed,
+            // and we should close the WaitHandle in here.  AsyncFSCallback
+            // and the hand-ported imitation version in COMThreadPool.cpp 
+            // are the only places that set this event.
```
stephentoub: This comment appears to be out-of-date.

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -59,6 +53,11 @@ public bool CompletedSynchronously
             get { return _completedSynchronously; }
         }
 
+        internal ManualResetEventSlim WaitHandle
```
stephentoub: Calling this WaitHandle is confusing to me, as a ManualResetEventSlim isn't actually a WaitHandle.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -258,17 +258,9 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
 
             // Create and store async stream class library specific data in the 
             // async result
-            PipeAsyncResult asyncResult = new PipeAsyncResult();
-            asyncResult._threadPoolBinding = _threadPoolBinding;
-            asyncResult._userCallback = callback;
-            asyncResult._userStateObject = state;
-
+            PipeAsyncResult asyncResult = new PipeAsyncResult(_threadPoolBinding, callback, state);
```
stephentoub: As evidenced by the code coverage report, we have very little test coverage for this type and code paths using it.  I don't want to take such a change without such tests.  Can you write some to go along with this?  It's of course more than just about ensuring lines are executed, but about making sure everything behaves the way we expect.  I'd prefer to see a commit that adds a bunch of such tests for various conditions, verifying that all such tests pass, and then subsequently a commit like this one that attempts to improve the implementation.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -414,16 +400,14 @@ unsafe private static void AsyncWaitForConnectionCallback(uint errorCode, uint n
             asyncResult._isComplete = true;
 
             // The OS does not signal this event.  We must do it ourselves.
-            ManualResetEvent wh = asyncResult._waitHandle;
-            if (wh != null)
+            ManualResetEventSlim wh = asyncResult.WaitHandle;
+            Debug.Assert(!wh.IsSet, "ManualResetEventSlim already signaled!");
+            wh.Set();
+            bool isSet = wh.IsSet;
```
stephentoub: This check and the subsequent assert and throw is unnecessary.  (And even if it could fail, it wouldn't result in a Win32 error.)

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -22,6 +21,14 @@ internal unsafe sealed class PipeAsyncResult : IAsyncResult
         internal bool _isComplete;              // Value for IsCompleted property        
         internal bool _completedSynchronously;  // Which thread called callback
 
+        public PipeAsyncResult(ThreadPoolBoundHandle handle, AsyncCallback callback, Object state)
+        {
+            _threadPoolBinding = handle;
+            _userCallback = callback;
+            _userStateObject = state;
+            _waitHandle = new ManualResetEventSlim(false);
```
stephentoub: false is the default used by the parameterless ctor.

stephentoub:
Thanks for submitting this, @Maxwe11.  I think we should hold off on merging this... two high-level points:
- We should ensure we have better code coverage before messing with these functions so much; our coverage right now here is pretty poor.
- Rather than just changing how this particular part of the implementation is done in an attempt to change the allocation of a ManualResetEvent into an allocation of a ManualResetEventSlim, I think it'd be better to look at trying to remove the Begin/End methods entirely.  They're a hold-over from when they were exposed publically, but now they're an implementation detail under the Task-returning WaitForConnectionAsync method.  I'd rather see that method reimplemented to be purely Task-based rather than going through an IAsyncResult-based implementation.  At that point neither the MRE nor the MRES should be needed.

Maxwe11:
Got it. Thanks for the clues :smile:

stephentoub:
Thanks for your continued contributions :smile:

PR #2096, opened 2015-06-18
Increase test coverage for ImmutableArray

stephentoub:
Scratching an itch... combined with #2090, this brings sequence coverage for ```ImmutableArray``` and ```ImmutableArray<T>``` up to 100%, and branch coverage for the same up to within a few branches of 100%. In doing so, I removed a bit of dead code and noticed some calls to Array.Copy that could be improved slightly.

There are some odd inconsistencies in exceptions thrown (or not thrown) for invalid arguments to various APIs, but regardless I added tests for them to help avoid accidental changes to the existing behavior; if we want to change the behavior, we can change the tests at the same time.  As an example, this throws a NullReferenceException:
```
builder.AddRange(default(ImmutableArray<int>));
```
but this completes successfully:
```
builder.AddRange(default(ImmutableArray<int>), 42);
```
It's clear why from looking at the implementation, but it seems like an unintentional difference.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using Validation;
@@ -107,11 +108,12 @@ namespace System.Collections.Immutable
             {
                 immutableArray.ThrowInvalidOperationIfNotInitialized();
 
-                var existingImmutableArray = immutableArray.Array as T[];
-                if (existingImmutableArray != null || immutableArray.Array == null)
-                {
-                    return new ImmutableArray<T>(existingImmutableArray);
-                }
+                // immutableArray.Array must not be null at this point, and we know it's an
+                // ImmutableArray<T> or ImmutableArray<SomethingDerivedFromT> as they are
+                // the only types that could be both IEnumerable<T> and IImmutableArray.
+                // As such, we know that items is either an ImmutableArray<T> or
+                // ImmutableArray<TypeDerivedFromT>, and we can cast the array to T[].
+                return new ImmutableArray<T>((T[])immutableArray.Array);
             }
 
             // We don't recognize the source as an array that is safe to use.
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @AArnott, @terrajobst

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
```diff
@@ -537,11 +539,7 @@ public static int BinarySearch<T>(this ImmutableArray<T> array, int index, int l
         /// <param name="items">The array to use or copy from. May be null for "default" arrays.</param>
```
sharwell: :exclamation: This documentation is no longer correct.

sharwell:
I reviewed all the non-test code changes. It looks good aside from the indicated comment.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
stephentoub: Yup, will fix.  The only caller of this internal method is already doing the check; no need to do it twice.

stephentoub:
Thanks, Sam.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
```diff
@@ -550,7 +548,7 @@ internal static ImmutableArray<T> CreateDefensiveCopy<T>(T[] items)
 
             // defensive copy
             var tmp = new T[items.Length];
-            Array.Copy(items, tmp, items.Length);
+            Array.Copy(items, 0, tmp, 0, items.Length);
```
AArnott: I'm curious what effect this has. Does it call directly into an overload that the old call would indirectly call into?

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
```diff
@@ -534,14 +536,10 @@ public static int BinarySearch<T>(this ImmutableArray<T> array, int index, int l
...
         /// </summary>
-        /// <param name="items">The array to use or copy from. May be null for "default" arrays.</param>
+        /// <param name="items">The array from which to copy.</param>
         internal static ImmutableArray<T> CreateDefensiveCopy<T>(T[] items)
         {
-            // Some folks lazily initialize fields containing these structs, so retaining a null vs. empty array status is useful.
-            if (items == null)
-            {
-                return default(ImmutableArray<T>);
-            }
```
AArnott: Did you remove this code because you proved it was never used anyway?

stephentoub: Yes, there's a single call site that itself is doing the check.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
stephentoub: Yes, but after looking up the lower bounds of rank 0 (the overload can't assume the array starts at 0).

AArnott:
:+1:

stephentoub:
Thanks, Andrew.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray.cs
AArnott: Why is it safe for _this_ method to assume that `items` starts at 0 then?

sharwell: Because `items` is a `T[]`.

PR #2099, opened 2015-06-18
Add System.Runtime.InteropServices tests

mellinoe:
ComAwareEventInfo does not yet have tests, not sure if we want/need any. I wasn't able to find any internal tests for the type and it's unclear whether we need any, given CoreCLR's COM support (not to mention cross-plat support).

Note: I needed to explicitly list the dependency version in project.json because the latest packages aren't compatible with our version of DNX yet.

### DIFF
--- a/src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/DefaultParameterValueAttributeTests.cs
+++ b/src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/DefaultParameterValueAttributeTests.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.Runtime.InteropServices
+{
+    public class DefaultParameterValueAttributeTests
+    {
+        [Theory]
+        [InlineData(0)]
+        [InlineData((byte)255)]
+        [InlineData(5.0)]
+        [InlineData(5.0f)]
+        [InlineData("ExpectedValue")]
+        [InlineData(null)]
+        public static void Constructor(object expected)
+        {
+            DefaultParameterValueAttribute attribute = new DefaultParameterValueAttribute(expected);
+            Assert.Equal(expected, attribute.Value);
+        }
+    }
+}
--- a/src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
+++ b/src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
@@ -1,0 +1,125 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.Runtime.InteropServices
+{
+    public static class HandleCollectorTests
+    {
+        private const int LowLimitSize = 20;
+        private const int HighLimitSize = 100000;
+
+        [Fact]
+        public static void NegativeInitialThresholdCtor()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => new HandleCollector("NegativeInitial", -1));
+        }
+
+        [Fact]
+        public static void NegateMaximumThresholdCtor()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => new HandleCollector("NegativeMax", 0, -1));
+        }
+
+        [Fact]
+        public static void InitialGreaterThanMaxThresholdCtor()
+        {
+            Assert.Throws<ArgumentException>(() => new HandleCollector("InitialGreaterThanMax", 100, 1));
+        }
+
+        [Fact]
+        public static void SimplePropertyValidation()
+        {
+            string name = "ExampleName";
+            int initial = 10;
+            int max = 20;
+
+            HandleCollector collector = new HandleCollector(name, initial, max);
+
+            Assert.Equal(0, collector.Count);
+            Assert.Equal(name, collector.Name);
+            Assert.Equal(initial, collector.InitialThreshold);
+            Assert.Equal(max, collector.MaximumThreshold);
+        }
+
+        [Fact]
+        public static void NullNameCtor()
+        {
+            HandleCollector collector = new HandleCollector(null, 0, 0);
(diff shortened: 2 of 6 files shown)


src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/DefaultParameterValueAttributeTests.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.Runtime.InteropServices
+{
+    public class DefaultParameterValueAttributeTests
+    {
+        [Fact]
+        public static void Constructor()
```
stephentoub: Maybe make this a Theory and test a few values other than just 0?

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
```diff
@@ -0,0 +1,129 @@
...
+        public static void SimplePropertyValidation()
+        {
+            int initial = 10;
+            int max = 20;
+            HandleCollector collector = new HandleCollector(null, initial, max);
+
+            Assert.Equal(0, collector.Count);
+            Assert.Equal(initial, collector.InitialThreshold);
+            Assert.Equal(max, collector.MaximumThreshold);
+            Assert.Equal(string.Empty, collector.Name);
```
stephentoub: I may have missed it in skimming quickly, but do we have any validation for the case of a non-null name ctor argument?

mellinoe: I'll add one

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            Tuple<int, int, int> postLowLimitState = new Tuple<int, int, int>(
+                GC.CollectionCount(0),
+                GC.CollectionCount(1),
+                GC.CollectionCount(2));
+
+            Assert.True(intialGcState.Item1 + intialGcState.Item2 + intialGcState.Item3 < postLowLimitState.Item1 + postLowLimitState.Item2 + postLowLimitState.Item3, "Low limit handle did not trigger a GC");
+
+            for (int i = 0; i < HighLimitSize + 10; ++i)
```
stephentoub: What's the significance of the ```+ 10```?

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
```diff
@@ -0,0 +1,129 @@
...
+        [Fact]
+        public static void SimplePropertyValidation()
+        {
+            int initial = 10;
+            int max = 20;
+            HandleCollector collector = new HandleCollector(null, initial, max);
+
+            Assert.Equal(0, collector.Count);
+            Assert.Equal(initial, collector.InitialThreshold);
+            Assert.Equal(max, collector.MaximumThreshold);
```
stephentoub: I must be missing something, but the implementation of HandleCollector doesn't appear to do anything with MaximumThreshold... it just exposes it from the property.  What's its purpose?

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                TestHighLimitHandle t = new TestHighLimitHandle();
+            }
+
+            Tuple<int, int, int> postHighLimitState = new Tuple<int, int, int>(
+                GC.CollectionCount(0),
+                GC.CollectionCount(1),
+                GC.CollectionCount(2));
+
+            Assert.True(postLowLimitState.Item1 + postLowLimitState.Item2 + postLowLimitState.Item3 < postHighLimitState.Item1 + postHighLimitState.Item2 + postHighLimitState.Item3, "High limit handle did not trigger a GC");
```
stephentoub: HighLimitSize is 100,000, which means the test is creating a 100,000 objects... isn't it likely that these allocations are themselves going to at least force a gen0 collection, whether or not GC.Collect is explicitly called?  That said, I don't know of a better way to test this, as to my knowledge there's no exposed API to determine the cause of a collection.

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
```diff
@@ -0,0 +1,129 @@
...
+                GC.KeepAlive(this);
+            }
+
+            ~TestLowLimitHandle()
+            {
+                s_lowLimitCollector.Remove();
+            }
+        }
+
+        private sealed class TestHighLimitHandle
```
stephentoub: You could collapse these two Test*LimitHandle implementations by passing the relevant HandleCollector into the ctor.

stephentoub:
Some nits/questions, but otherwise LGTM.

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
mellinoe: Yep, I think it's completely pointless actually... the only thing it does is compare the initialThreshold parameter against it and throws if initial > max.

That said, it's part of the contract that already shipped so we can't really do much about it, I think.

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
mellinoe: I think it is to "force" the handle count to go over the threshold and cause a collection to occur. I guess it's 10 just to "be sure".

src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollector.cs
mellinoe: Yeah, I think checking GC Collection counts like this is a bit "fuzzy" and imprecise but I'm not sure how to improve it, especially since our tests aren't really isolated from each other in test runs.

stephentoub:
The updates LGTM.

PR #2107, opened 2015-06-19
[WIP] Add samples to corefx

richlander:
- Pulling samples from MSDN.
- Improving them as I copy them.
- Create a new 'samples' directory as a peer to tests and src.

These samples are intended to be very basic demonstrations of a feature. We are adding ref docs, soon. We'd probably want to add a 'docs' directory at the same place for content that auguments /// content.

### DIFF
--- a/src/System.Collections/samples/List/List.cs
+++ b/src/System.Collections/samples/List/List.cs
@@ -9,7 +9,7 @@ public class Program
         // Doc: https://msdn.microsoft.com/library/6sh2ey19.aspx
         
         // Create a list of parts.
-        List<Part> parts = new List<Part>();
+        var parts = new List<Part>();
 
         // Add a part to the list.
         var crank = new Part();
@@ -18,11 +18,11 @@ public class Program
         parts.Add(crank);
 
         // Add parts to the list with more compact syntax via object initializers.
-        parts.Add(new Part() {PartName = "chain ring", PartId = 1334 });
-        parts.Add(new Part() {PartName = "regular seat", PartId = 1434 });
-        parts.Add(new Part() {PartName = "banana seat", PartId = 1444 });
-        parts.Add(new Part() {PartName = "cassette", PartId = 1534 });
-        parts.Add(new Part() {PartName = "shift lever", PartId = 1634 }); ;
+        parts.Add(new Part() { PartName = "chain ring", PartId = 1334 });
+        parts.Add(new Part() { PartName = "regular seat", PartId = 1434 });
+        parts.Add(new Part() { PartName = "banana seat", PartId = 1444 });
+        parts.Add(new Part() { PartName = "cassette", PartId = 1534 });
+        parts.Add(new Part() { PartName = "shift lever", PartId = 1634 }); ;
  
         PrintParts(parts);
         
--- a/src/System.Collections/samples/List/List.cs
+++ b/src/System.Collections/samples/List/List.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Collections.Generic;
+
+public class Program
+{
+    public static void Main()
+    {
+        // Demonstrates common uses of List<T>.
+        // Doc: https://msdn.microsoft.com/library/6sh2ey19.aspx
+        
+        // Create a list of parts.
+        List<Part> parts = new List<Part>();
+
+        // Add a part to the list.
+        var crank = new Part();
+        crank.PartName = "crank arm";
+        crank.PartId = 1234;
+        parts.Add(crank);
+
+        // Add parts to the list with more compact syntax via object initializers.
+        parts.Add(new Part() {PartName = "chain ring", PartId = 1334 });
+        parts.Add(new Part() {PartName = "regular seat", PartId = 1434 });
+        parts.Add(new Part() {PartName = "banana seat", PartId = 1444 });
+        parts.Add(new Part() {PartName = "cassette", PartId = 1534 });
+        parts.Add(new Part() {PartName = "shift lever", PartId = 1634 }); ;
+ 
+        PrintParts(parts);
+        
+        Console.WriteLine();
+        Console.WriteLine("Insert brake lever part at index 2");
+        parts.Insert(2, new Part() { PartName = "brake lever", PartId = 1834 });
+
+        PrintParts(parts);
+        
+        Console.WriteLine();
+        Console.WriteLine("Remove part at index 3");
+        parts.RemoveAt(3);
+        
+        PrintParts(parts);
+         
+        /*
+        Expected console output:
+        
+        6 Parts:
+        ID: 1234, Name: crank arm
+        ID: 1334, Name: chain ring
+        ID: 1434, Name: regular seat
+        ID: 1444, Name: banana seat
+        ID: 1534, Name: cassette
+        ID: 1634, Name: shift lever
+        
+        Insert brake lever part at index 2
+        
+        7 Parts:
+        ID: 1234, Name: crank arm
+        ID: 1334, Name: chain ring
+        ID: 1834, Name: brake lever
+        ID: 1434, Name: regular seat
+        ID: 1444, Name: banana seat
+        ID: 1534, Name: cassette
+        ID: 1634, Name: shift lever
+        
+        Remove part at index 3
+        
+        6 Parts:
+        ID: 1234, Name: crank arm
+        ID: 1334, Name: chain ring
+        ID: 1834, Name: brake lever
+        ID: 1444, Name: banana seat
+        ID: 1534, Name: cassette
+        ID: 1634, Name: shift lever
+
+     */
+	}
+    
+    // Write out the parts in the list.
+    static void PrintParts(List<Part> parts)
+    {
+        Console.WriteLine();
+        Console.WriteLine("{0} Parts:", parts.Count);
+        
+        // List<T> implements IEnumerable, so supports foreach.
+        foreach (var part in parts)
+        {
+            // WriteLine calls the overridden ToString method in the Part class.
+            Console.WriteLine(part);
+        }
+    }
+}
+
+// Part describes a 'part' in an order system
+public class Part
+{
+    public string PartName {get; set;}
+    public int PartId {get; set;}
+
+    public override string ToString()
+    {
+        return String.Format("ID: {0}, Name: {1}", PartId, PartName);
+    }
+}
--- a/src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
+++ b/src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
@@ -15,16 +15,16 @@ public class Program
 		Console.WriteLine();
 		
 		// short / long formatting
-		Console.WriteLine("Print short date: {0}", date.ToString("d"));
-		Console.WriteLine("Print long date: {0}", date.ToString("D"));
-		Console.WriteLine("Print short time: {0}", date.ToString("t"));
-		Console.WriteLine("Print long time: {0}", date.ToString("T"));
+		Console.WriteLine("Print short date: {0:d}", date);
+		Console.WriteLine("Print long date: {0:D}", date);
+		Console.WriteLine("Print short time: {0:t}", date);
+		Console.WriteLine("Print long time: {0:T}", date);
 		Console.WriteLine();
 		
 		// custom formatting with templating
-		Console.WriteLine("Print date with month/year format: {0}", date.ToString("MM/yy"));
-		Console.WriteLine("Print date with month, year in verbose format: {0}", date.ToString("MMMM, yyyy"));
-		Console.WriteLine("Print time: {0}", date.ToString("hh:mm:ss"));
+		Console.WriteLine("Print date with month/year format: {0:MM/yy}", date);
+		Console.WriteLine("Print date with month, year in verbose format: {0:MMMM, yyyy}", date);
+		Console.WriteLine("Print time: {0:hh:mm:ss}", date);
 		Console.WriteLine();
 		
 		// print with locale-specific formatting
--- a/src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
+++ b/src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
@@ -1,0 +1,44 @@
+using System;
+using System.Globalization;
+
+public class Program
+{
+	public static void Main()
+	{
+		// Demonstrates common uses of formatting the DateTime struct.
+		// Doc: https://msdn.microsoft.com/library/az4se3k1.aspx
+		
+		var date = DateTime.Now;
+		
+		// default formatting
+		Console.WriteLine("Print DateTime with default formatting: {0}", date);
+		Console.WriteLine();
+		
+		// short / long formatting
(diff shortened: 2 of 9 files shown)


mellinoe:
Couple questions for clarification:

* Do we intend for CoreFx to house all of our samples, or is this just a placeholder before we move them into a dedicated samples repo? Personally, I'd prefer samples to live elsewhere for a few reasons, the main one being organization/cleanliness. Some other reasons:
  * People wanting to learn the platform don't necessarily want to get the source code or be tied with our weird build system.
  * Assuming we only have a small (relatively speaking) number of samples, it might seem a bit arbitrary which libraries have samples and which don't.
  * Samples will probably touch many libraries, so it will seem arbitrary which "folder" they live in. This would seem less arbitrary if there was a separate repo with more specialized organization.

* How are we intending these to be built/run? We don't really have a standard way of making a "CoreCLR" Console app right now. We could use the ASP.NET 5 templates, but those would probably be better off in an ASP.NET samples repo (or there might already be one). This will be less of an issue once our packages are more mature and we have better VS and build tooling.

richlander:
@mellinoe great questions. The primary conversation I'm wanting to have is what we're going to do for ref docs. These are not 'sample' in the traditional sense. They are samples that will get pulled into the ref docs.

The samples (so far) are DNX apps.

@danroth27, @blackdwarf Any thoughts on samples for ref docs and where you plan to store of them?

Sure looks like I should have opened an issue for this. Opened one.

src/System.Collections/samples/List/List.cs
```diff
@@ -0,0 +1,101 @@
...
+
+public class Program
+{
+    public static void Main()
+    {
+        // Demonstrates common uses of List<T>.
+        // Doc: https://msdn.microsoft.com/library/6sh2ey19.aspx
+        
+        // Create a list of parts.
+        List<Part> parts = new List<Part>();
```
KrzysztofCwalina: should this use "var"

src/System.Collections/samples/List/List.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        // Add a part to the list.
+        var crank = new Part();
+        crank.PartName = "crank arm";
+        crank.PartId = 1234;
+        parts.Add(crank);
+
+        // Add parts to the list with more compact syntax via object initializers.
+        parts.Add(new Part() {PartName = "chain ring", PartId = 1334 });
```
KrzysztofCwalina: Should there be a space after the first "{"?

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
...
+		// default formatting
+		Console.WriteLine("Print DateTime with default formatting: {0}", date);
+		Console.WriteLine();
+		
+		// short / long formatting
+		Console.WriteLine("Print short date: {0}", date.ToString("d"));
+		Console.WriteLine("Print long date: {0}", date.ToString("D"));
+		Console.WriteLine("Print short time: {0}", date.ToString("t"));
+		Console.WriteLine("Print long time: {0}", date.ToString("T"));
+		Console.WriteLine();
```
KrzysztofCwalina: I think it would be better to teach people to do:

            Console.WriteLine("Print short date: {0:d}", date);
            Console.WriteLine("Print long date: {0:D}", date);
            Console.WriteLine("Print short time: {0:t}", date);
            Console.WriteLine("Print long time: {0:T}", date);

Not that it's important for console scenarios, but this form can be optimized to not allocate strings.

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
...
+		
+		// custom formatting with templating
+		Console.WriteLine("Print date with month/year format: {0}", date.ToString("MM/yy"));
+		Console.WriteLine("Print date with month, year in verbose format: {0}", date.ToString("MMMM, yyyy"));
+		Console.WriteLine("Print time: {0}", date.ToString("hh:mm:ss"));
+		Console.WriteLine();
+		
+		// print with locale-specific formatting
+		var ptBRCulture = new CultureInfo("pt-BR");
```
KrzysztofCwalina: Similarly, these should be CultureInfo.GetCultureInfo(...) instead of using the ctor. GetCultureInfo returns cached culture.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,53 @@
...
+
+public class Program
+{
+	public static void Main()
+	{
+		// Demonstrates common uses of DateTime struct.
+		// Doc: https://msdn.microsoft.com/library/system.datetime.aspx
+		
+		// Get and manipulate time
+		Console.WriteLine("Print the DateTime now");
```
KrzysztofCwalina: Should this say: "print local date and time" ?

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,53 @@
...
+		Console.WriteLine(DateTime.Now);
+		Console.WriteLine();
+		
+		Console.WriteLine("Print the UTC DateTime now");
+		Console.WriteLine(DateTime.UtcNow);
+		Console.WriteLine();
+		
+		Console.WriteLine("Print the DateTime in 10 mins");
```
KrzysztofCwalina: This is even more strangely worded. Sounds like it will print in 10 minutes.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,53 @@
...
+		Console.WriteLine(DateTime.Now.AddMinutes(10));
+		Console.WriteLine();
+		
+		Console.WriteLine("Print the DateTime 5 days and 4 hours ago");
+		Console.WriteLine(DateTime.Now.AddDays(-5).AddHours(-4));
+		Console.WriteLine();
+		
+		// Create a DateTime for a specific date/time
+		// Uses Unix epoch as the example date - https://en.wikipedia.org/wiki/Unix_time
+		DateTime epoch = new DateTime(1970,1,1);
```
KrzysztofCwalina: We should teach people to never create DateTimeKind.Unspecified values. Would it make sense to use one of the ctors with the DateTimeKind parameter?

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
richlander: @KrzysztofCwalina That's such an improvement! Will totally fix that.

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
richlander: What package do I need to bring in for that?

KrzysztofCwalina: Ah, I just noticed it's not in PCL

src/System.Collections/samples/List/List.cs
```diff
@@ -0,0 +1,101 @@
...
+        ID: 1834, Name: brake lever
+        ID: 1444, Name: banana seat
+        ID: 1534, Name: cassette
+        ID: 1634, Name: shift lever
+
+     */
+	}
+    
+    // Write out the parts in the list.
+    static void PrintParts(List<Part> parts)
```
stephentoub: If we're going to have samples like this in the repo, we should follow the coding style conventions of the repo, e.g. this should be explicitly marked as "private". (https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md rule # 5.)

src/System.Collections/samples/List/List.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        Console.WriteLine();
+        Console.WriteLine("{0} Parts:", parts.Count);
+        
+        // List<T> implements IEnumerable, so supports foreach.
```
stephentoub: This comment is a bit misleading.  While it's true that ```List<T>``` does implement ```IEnumerable``` and ```IEnumerable<T>```, that's not what ends up being used here, but rather its strongly-typed GetEnumerator method (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L564) that returns an instance of the public, nested, struct Enumerator type (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L1135)

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
+using System;
```
stephentoub: The List.cs class is in a directory src/System.Collections/samples/List/.  Shouldn't this file be in a directory src/System.Runtime/samples/DateTime/ ?

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Globalization;
+
+public class Program
+{
+	public static void Main()
+	{
+		// Demonstrates common uses of formatting the DateTime struct.
+		// Doc: https://msdn.microsoft.com/library/az4se3k1.aspx
+		
+		var date = DateTime.Now;
```
stephentoub: Could we change ```date``` here to be ```localData```, or something that highlights better what ```Now``` returns?

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
...
+		// default formatting
+		Console.WriteLine("Print DateTime with default formatting: {0}", date);
+		Console.WriteLine();
+		
+		// short / long formatting
+		Console.WriteLine("Print short date: {0:d}", date);
+		Console.WriteLine("Print long date: {0:D}", date);
+		Console.WriteLine("Print short time: {0:t}", date);
+		Console.WriteLine("Print long time: {0:T}", date);
+		Console.WriteLine();
```
stephentoub: Just as below the culture name is included in the output, it'd be good here to include what qualifier is being used, e.g.
```C#
Console.WriteLine("Print short date ('d'): {0:d}", date);
Console.WriteLine("Print long date ('D'): {0:D}", date);
Console.WriteLine("Print short time ('t'): {0:t}", date);
Console.WriteLine("Print long time ('T'): {0:T}", date);
```

src/System.Runtime/samples/DateTime-formatting/DateTime-formatting.cs
```diff
@@ -0,0 +1,44 @@
...
+		var deDECulture = new CultureInfo("de-DE");
+		var zhCNCulture = new CultureInfo("zh-CN");
+		Console.WriteLine("Print DateTime with pt-BR local: {0}", date.ToString(ptBRCulture));
+		Console.WriteLine("Print DateTime with en-US local: {0}", date.ToString(enUSCulture));
+		Console.WriteLine("Print DateTime with de-DE local: {0}", date.ToString(deDECulture));
+		Console.WriteLine("Print DateTime with zh-CN local: {0}", date.ToString(zhCNCulture));
+		Console.WriteLine("Print long DateTime with zh-CN local: {0}", date.ToString("D",zhCNCulture));
+		Console.WriteLine("Print long DateTime with de-DE local: {0}", date.ToString("D",deDECulture));
+		
+		Console.WriteLine();
```
stephentoub: This List examples includes a comment that shows the expected output.  Should the same be done here?  Obviously it'll change based on the result of DateTime.Now, but a simple comment at the top of the output could help explain that this is an example output.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,54 @@
...
+		Console.WriteLine("Print UTC DateTime");
+		Console.WriteLine(DateTime.UtcNow);
+		Console.WriteLine();
+		
+		Console.WriteLine("Print local DateTime + 10 mins");
+		Console.WriteLine(DateTime.Now.AddMinutes(10));
+		Console.WriteLine();
+		
+		Console.WriteLine("Print local DateTime - 5 days and 4 hours");
+		Console.WriteLine(DateTime.Now.AddDays(-5).AddHours(-4));
```
stephentoub: How about showing doing these manipulations on a UTC time instead of a local time?  Math on local time is generally discouraged.

stephentoub:
@richlander, if the decision is to include samples in the repo, who's responsible for maintaining and augmenting them?  I worry we're going to find ourselves with a smattering of examples that are incomplete and eventually just languish.

stephentoub:
@richlander, what's the status of this?

stephentoub:
@richlander?

src/System.Runtime/samples/DateTime/DateTime.cs
mj1856: Echoing @stephentoub's comment - I've seen too many people get bit by math on local `DateTime` values.  We shouldn't perpetuate bad practices in our examples.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,54 @@
...
+		Console.WriteLine();
+		
+		// Create a DateTime for a specific date/time
+		// Uses Unix epoch as the example date - https://en.wikipedia.org/wiki/Unix_time
+		// Always specific DateTimeKind when creating a new DateTime
+		DateTime epoch = new DateTime(1970,1,1,0,0,0,DateTimeKind.Local);
```
mj1856: Should be `DateTimeKind.Utc`, not `DateTimeKind.Local`.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,54 @@
...
+		var unixTime = 1581941594; // 2/17/2020 12:13:14 PM
+		Console.WriteLine("Create and print a DateTime with Unix time");
+		Console.WriteLine("The math: {0} + {1} seconds", epoch.Year, unixTime);
+		Console.WriteLine(epoch.AddSeconds(unixTime));
+		Console.WriteLine();
+
+		// Operate on parts of DateTime
+		Console.WriteLine("Print local year: {0}",DateTime.Now.Year);
+		Console.WriteLine("Print local day of month: {0}", DateTime.Now.Day);
+		Console.WriteLine("Print local date without time: {0}", DateTime.Now.Date);
```
mj1856: use `{0:d}` to hide the artificial midnight.

src/System.Runtime/samples/DateTime/DateTime.cs
```diff
@@ -0,0 +1,54 @@
...
+		Console.WriteLine("Print local time: {0}", DateTime.Now.TimeOfDay);
```
mj1856: Consider dropping `.TimeOfDay` and using `{0:t}` instead.

richlander:
May come back to this later when I have a better plan in place.

Issue #1597, opened 2015-04-30
Running CoreFx tests from VS is broken


tmat:
I’m using TestDriven.NET.
1)  Right-click on a test method and selecting “Run Test” works well for System.Reflection.Metadata tests. Good.
2)  Right-click on a test class or a test file reports an error in output:
The target type doesn't contain tests from a known test framework or a 'Main' method.
3)  Right-click on a test method and selecting “Test with … Debugger” freezes VS.

Test Explorer doesn’t show any tests even after full solution build.

Trying to set the test project as a startup project and F5 it.

Error:
“The working directory does not exist: 'C:\FX0\bin\tests\Windows_NT.AnyCPU.Debug\System.Reflection.Metadata.Tests\dnxcore50'.”

Seems like building the entire CoreFx repo fixes this issue (although I’m not quite sure). In any case I don’t want to build the entire repo to run a single test.

Couldn’t get F5 work from Dev12 VS, yet the project is Dev12 VS. With Dev14 I was able to debug the test.

krwq:
I think it might be related: https://github.com/dotnet/corefx/issues/1460

tmat:
@krwq Yes, seems like the F5 scenario issues are related to #1460.

Another problem is that the test references are populated by some magic from packages.config file and TestDriven.NET can't find the xunit reference.

mellinoe:
I think this just isn't implemented yet. We have a sort of F5 hack in place, but building the project (or hitting F5) doesn't actually produce/construct the directory structure for tests. All of that work is done by our test targets, which aren't involved there. And I don't think debugging CoreCLR is going to be supported from Dev12 at all; only Dev14, unfortunately.

> Seems like building the entire CoreFx repo fixes this issue (although I’m not quite sure)

I think it's because the full build produces the test directories and runs the tests in all of them, so when you open in VS and hit F5, the test directory already exists. I would bet that re-building the product/tests doesn't actually re-populate the test directory, only the binaries directory.

I don't know how TestDriven.NET works, but I highly doubt it will just work given the nature of our projects and product.

mellinoe:
In my opinion, we should invest in a VS test explorer plugin. That would probably provide the best experience in VS, and be closer to what people are used to. We have an internal one, but it may need to be modified heavily (and probably stripped down) to be useful for CoreFx.

tmat:
@mellinoe Investing into test infrastructure is critical. I don't see how anybody can be productive writing and debugging unit tests with the current setup.

[...]

PR #2109, opened 2015-06-19
Add System.Reflection.Metadata.Test.Desktop, which targets FX 4.6 and  Xunit 1.9.2

tmat:
This makes it possible to run and debug tests in VS.

Caveats: can't use Theory attribute since it's not available in Xunit 1.9.2


nguerrera:
Why can't we use xunit 2.0?

tmat:
I guess we could. Roslyn is using 1.9.2 not sure why.

nguerrera:
Looks OK to me. Waiting on CI.

akoeplinger:
> Caveats: can't use Theory attribute since it's not available in Xunit 1.9.2

This feature was available through the xunit extensions package in 1.x.

Apart from that, the new project doesn't follow the other corefx project conventions (packages.config instead of project.json, hardcoding the snk, importing Microsoft.CodeSharing.* targets that aren't yet available on other platforms afaik). Is that expected or just a temporary thing?

tmat:
@akoeplinger  This is expected and temporary until it is possible to run and debug Core CLR targeted tests properly from VS.

tmat:
The issues with running tests from VS are tracked by https://github.com/dotnet/corefx/issues/1597.

PR #2110, opened 2015-06-19
TD.NET doesn't work well with shared projects, replacing with manual sharing

tmat:


tmat:
@nguerrera @KevinH-MS

tmat:
@KevinH-MS Could you ask the TD.NET maintainer about support for shared projects? The problem is that when a test method is in a shared project TD.NET doesn't use the current head to find the assembly, but rather tries to get the assembly of the shared project, which there is none.

KevinH-MS:
Per hallway conversation, we should be able to use xunit.runner.dnx for this.  We just need to get it working with the latest K projects.  It  *should* just involve:
```
  "dependencies": {
    "xunit": "2.1.0-beta2-*",
    "xunit.runner.dnx": "2.1.0-beta2-*"
  },

  "commands": {
    "test": "xunit.runner.dnx"
  },
```
(in project.json)

KevinH-MS:
I'm currently seeing:
```
------ Test started: Project: CoreLibraryTest ------
Starting  Microsoft.Framework.TestHost [C:\Users\kevinh\.dnx\runtimes\dnx-coreclr-win-x86.1.0.0-beta5-11911\bin\dnx.exe --appbase "C:\Users\kevinh\Documents\Visual Studio 2015\Projects\CoreLibrary\src\CoreLibraryTest" Microsoft.Framework.ApplicationHost --port 63951 Microsoft.Framework.TestHost --port 64001]
Unable to start Microsoft.Framework.TestHost
========== Discover test finished: 0 found (0:00:03.1301446) ==========
```

PR #2120, opened 2015-06-22
Optimize array/list.Select(...).ToArray()

stephentoub:
LINQ already special-cases the very common case of using Select on input ```T[]```s and ```List<T>```s, but it doesn't take this to the next step of optimizing the case where that's then turned back into an array, a common pattern in its own right, e.g.
```C#
var names = people.Select(p => p.Name).ToArray();
```
This commit does so, modifying the internal Iterator type to support customization of ToArray, and updating the internal Buffer type to use it.  Changing Buffer not only helps ToArray but also Reverse and OrderBy, both of which convert the input sequence into an array internally.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -144,6 +144,11 @@ namespace System.Linq
 
             public abstract IEnumerable<TSource> Where(Func<TSource, bool> predicate);
 
+            public virtual TSource[] ToArray()
+            {
+                return new Buffer<TSource>(this, queryInterfaces: false).ToArray();
+            }
+
             object IEnumerator.Current
             {
                 get { return Current; }
@@ -431,6 +436,21 @@ namespace System.Linq
             {
                 return new WhereEnumerableIterator<TResult>(this, predicate);
             }
+
+            public override TResult[] ToArray()
+            {
+                if (_predicate != null)
+                {
+                    return base.ToArray();
+                }
+
+                var results = new TResult[_source.Length];
+                for (int i = 0; i < results.Length; i++)
+                {
+                    results[i] = _selector(_source[i]);
+                }
+                return results;
+            }
         }
 
         internal class WhereSelectListIterator<TSource, TResult> : Iterator<TResult>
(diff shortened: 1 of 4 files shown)


stephentoub:
cc: @VSadov, @AlekseyTs, @KrzysztofCwalina

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -431,6 +436,21 @@ public override IEnumerable<TResult> Where(Func<TResult, bool> predicate)
...
+            public override TResult[] ToArray()
+            {
+                if (_predicate != null)
+                {
+                    return base.ToArray();
+                }
+
+                var results = new TResult[_source.Length];
+                for (int i = 0; i < _source.Length; i++)
+                {
```
VSadov: result.Length   could result in a better range check here. Since result is a local, it is easier to prove its Length cannot change.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3109,6 +3158,7 @@ internal Buffer(IEnumerable<TElement> source)
                     count++;
                 }
             }
+
             this.items = items;
```
VSadov: There is a "new TElement[0];"   in the ToArray() implementation. Are we not using Array.Empty there for compat reasons?

VSadov:
LGTM 
" < results.Length" comment is probably worth addressing.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: I was hesitant to change it for that reason.  Is it ok to make it Array.Empty<TElement>()? (Regardless, if we want to change it I'd like us to do it as part of a separate PR / commit.)

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: Good suggestion, fixed.

src/System.Linq/src/System/Linq/Enumerable.cs
davidfowl: That's a good question. I'd be interested to know how implementation dependencies affect compatibility in general. /cc @ericstj

stephentoub: That's a fine separate question, but to be clear mine wasn't about dependencies. When I previously did a sweep to replace new array allocations with Array.Empty, I generally refrained from doing so in ToArray methods, as many are documented to allocate or create a new arrays, and I didn't want to inadvertently introduce compat problems for folks relying on the methods returning new instances. If we don't believe that's an issue in this case, I'd be happy to see it changed, too.

Issue #382, opened 2015-01-09
Add ctor to HashSet<T> that allows the initial capacity to be specified


mattwarren:
Unlike most other collections within the `System.Collections` namespace, `HashSet<T>` doesn't have a [constructor overload](http://msdn.microsoft.com/en-us/library/bb354549(v=vs.110).aspx) that allows you to pass in the expected capacity. It will only let you pass in an [`IEnumerable<T> collection`](http://referencesource.microsoft.com/#System.Core/System/Collections/Generic/HashSet.cs,116) of items to pre-populate it with.

Having this would enable the use case where you know how many items will be added to the `HashSet<T>` (or at least an upper bound), but don't have them all up front, i.e. they are to be added over time. 

It would therefore let pre-size the `HashSet<T>` so that it doesn't have to grow over time, with all the associated copying that is needed. This pattern is exposed by other collections, for instance [`List<T>`](http://msdn.microsoft.com/en-us/library/dw8e0z9z(v=vs.110).aspx), to cover this scenario.

There's already a private [`Initialize` method](http://referencesource.microsoft.com/#System.Core/System/Collections/Generic/HashSet.cs,831ca3e6d9c0d78b) that will do the work, so unless I'm missing something, it's just a case of calling this.

Is this something that is worth adding?

mikedn:
I'd very much like to see this added too. It is useful to avoid resizing in cases where a reasonable approximation of the final count is known, especially for hashtables where resizing is more expensive than for lists.

KrzysztofCwalina:
I think it's a good suggestion.

mattwarren:
Cool, glad people agree. BTW I'm more than happy to do the pull-request for the actual work if that's possible?

Petermarcu:
@mattwarren,Please do. We'd love to have you contribute!

mattwarren:
Cool, I guess I have to wait until `System.Collection` is open-sourced though, or have I missed that announcement?

terrajobst:
That's correct. I'll mark this issue as "Accepting PRs". If you want to call dips, say so here :-)

mattwarren:
Yeah I'd love to call dibs on this one ;-) I'll put together a PR when it's open-sourced.

terrajobst:
[...]

PR #2122, opened 2015-06-22
HashSet<T> constructor with capacity

MarkusSintonen:
Issue #382
-Added constructor with capacity parameter to HashSet<T>
-Added two unit tests for HashSet<T> capacity constructor

### DIFF
--- a/src/System.Collections/src/System/Collections/Generic/HashSet.cs
+++ b/src/System.Collections/src/System/Collections/Generic/HashSet.cs
@@ -75,22 +75,16 @@ namespace System.Collections.Generic
         #region Constructors
 
         public HashSet()
-            : this(EqualityComparer<T>.Default)
+            : this(0, EqualityComparer<T>.Default)
         { }
 
         public HashSet(IEqualityComparer<T> comparer)
-        {
-            if (comparer == null)
-            {
-                comparer = EqualityComparer<T>.Default;
-            }
+            : this(0, comparer)
+        { }
 
-            _comparer = comparer;
-            _lastIndex = 0;
-            _count = 0;
-            _freeList = -1;
-            _version = 0;
-        }
+        public HashSet(int capacity)
+            : this(capacity, EqualityComparer<T>.Default)
+        { }
 
         public HashSet(IEnumerable<T> collection)
             : this(collection, EqualityComparer<T>.Default)
@@ -128,6 +122,30 @@ namespace System.Collections.Generic
                 (_count > 0 && _slots.Length / _count > ShrinkThreshold))
             {
                 TrimExcess();
+            }
+        }
+
+        public HashSet(int capacity, IEqualityComparer<T> comparer)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException("capacity");
+            }
+
+            if (comparer == null)
+            {
+                comparer = EqualityComparer<T>.Default;
+            }
+
+            _comparer = comparer;
+            _lastIndex = 0;
+            _count = 0;
+            _freeList = -1;
+            _version = 0;
+
+            if (capacity > 0)
+            {
+                Initialize(capacity);
             }
         }
 
--- a/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
+++ b/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
@@ -1248,6 +1248,9 @@ namespace System.Collections.Concurrent
                 result = tail._value;
             }
 
+            /// <summary>
+            /// Resets the local list
+            /// </summary>
             internal void Reset()
             {
                 _head = null;
--- a/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
+++ b/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
@@ -48,6 +48,9 @@ namespace System.Collections.Concurrent
         // GlobalListsLock lock
         private bool _needSync;
 
+        // Count of the thread lists. This is used to determine initial capacity for the version lists
+        private volatile int _threadListCount;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ConcurrentBag{T}"/>
         /// class.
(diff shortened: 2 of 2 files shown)


svick:
I take it this is the implementation for #382?

MarkusSintonen:
Yes, I updated the comment

mellinoe:
Note: this is the error in the CI build:

Generic\HashSet\HashSet_ConstructorTests.cs(291,36): error CS1502: The best overloaded method match for 'System.Collections.Generic.HashSet<int>.HashSet(System.Collections.Generic.IEnumerable<int>)' has some invalid arguments

This is expected because this new method hasn't actually been added to the contract assembly we build internally, which is what the test compiles against.

src/System.Collections/tests/Generic/HashSet/HashSet_ConstructorTests.cs
```diff
@@ -281,5 +281,27 @@ public static void HashSetConstructor_IEn_IEq_Comparer_NotIEquatable()
         }
 
         #endregion
+
+        #region Constructor_Capacity
+
+        //Test 1: Check that EqualityComparer<T>.Default is used
```
KrzysztofCwalina: How does it verify that the default comparer is used?

src/System.Collections/tests/Generic/HashSet/HashSet_ConstructorTests.cs
```diff
@@ -281,5 +281,27 @@ public static void HashSetConstructor_IEn_IEq_Comparer_NotIEquatable()
...
+        public static void HashSetConstructor_Capacity_Negative()
+        {
+            HashSet<int> hashSet = new HashSet<int>(13);
+            int[] items = new int[0];
+            HashSetTestSupport<int> driver = new HashSetTestSupport<int>(hashSet, s_intGenerator, items);
+            driver.VerifyHashSet_NegativeTests();
+        }
+
+        #endregion
     }
```
KrzysztofCwalina: Should there be tests for the ctor that takes capacity and comparer?

stephentoub:
@mellinoe, the way we currently build our test projects for partial facades makes it impossible to add tests for new APIs, since they're not part of the contract but we build the tests against the contract:
https://github.com/MarkusSintonen/corefx/blob/future/src/System.Collections/tests/System.Collections.Tests.csproj#L18
Do we have a plan for solving this?

mellinoe:
Other than bringing out our contract generation code to GitHub, I don't know that we have a plan for this necessarily. Traditionally we've always built our tests against the contract, so a change like this would force us to update the contract internally as well. Unfortunately it's not really possible for people adding new APIs to do that until the contract generation stuff is out here.

Technically it should be possible to compile these tests against the partial facade itself, but that will require also compiling against mscorlib.dll, as well as potentially other partial facades that are used by the library. For example, System.Diagnostics.Debug's tests are compiled against the library project file, because it uses internal types, and System.Diagnostics.Debug is a partial facade.

The basic process is to add this to project.json to pull in mscorlib:

 "Microsoft.DotNet.CoreCLR": "1.0.0-prerelease",

Additionally you need a few project properties to force-resolve mscorlib.dll from that assembly if you aren't a Partial Facade project:

    <NuGetTargetFrameworkMoniker>DNXCore,Version=v5.0</NuGetTargetFrameworkMoniker>
    <IgnoreArchitectureMismatches>true</IgnoreArchitectureMismatches>
    <PostFilterNugetReferences>true</PostFilterNugetReferences>

stephentoub:
> Technically it should be possible to compile these tests against the partial facade

Separate from (and before) this change, should we update our projects as you suggest? Or is the stance "no new APIs in partial facades" for now until the rolling issues are sorted?

MarkusSintonen:
I wondered why the test compilation failed on my machine. Well at least it wasn't a problem with my own setup. I think it would be valuable to get the full tooling available since this would allow test coverage for future brach API additions.

src/System.Collections/tests/Generic/HashSet/HashSet_ConstructorTests.cs
MarkusSintonen: More tests should be done once the tooling problems are resolved related to API changes and unit testing.

src/System.Collections/src/System/Collections/Generic/HashSet.cs
```diff
@@ -131,6 +125,30 @@ public HashSet(IEnumerable<T> collection, IEqualityComparer<T> comparer)
             }
         }
 
+        public HashSet(int capacity, IEqualityComparer<T> comparer)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException("capacity");
```
jasonwilliams200OK: Using C# 6, we can say:

```c#
throw new ArgumentOutOfRangeException(nameof(capacity)); 
```

src/System.Collections/src/System/Collections/Generic/HashSet.cs
```diff
@@ -131,6 +125,30 @@ public HashSet(IEnumerable<T> collection, IEqualityComparer<T> comparer)
...
+                comparer = EqualityComparer<T>.Default;
+            }
+
+            _comparer = comparer;
+            _lastIndex = 0;
+            _count = 0;
+            _freeList = -1;
+            _version = 0;
+
+            if (capacity > 0)
```
jasonwilliams200OK: This condition is not required?

src/System.Collections/src/System/Collections/Generic/HashSet.cs
akoeplinger: @jasonwilliams200OK corefx doesn't use C#6 yet.

stephentoub:
@MarkusSintonen, it looks like you may have accidentally pushed your concurrent bag changes to this PR. Would you be able to fix that? We're still blocked in taking this change on the infrastructure issues around tests, but hopefully that will be stored out soon and then this can be merged. Thanks!

stephentoub:
@mellinoe, I tried your suggested workaround, but it then fails to compile due to lots of duplicated types between mscorlib and the various contracts.  Could you see if there's an easy way to unblock this?

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
@MarkusSintonen, sorry for the delay on this.  Let's go ahead on this without the test for the new ctor, and when we have the infrastructure in place, we can then add the test.  Can you squash this down to a single commit? Thanks!

stephentoub:
Replaced by #2862.

PR #2130, opened 2015-06-23
Switch S.R.Metadata to standard Strings.resx pattern

nguerrera:
This library was missed during the bulk changes to how resources were
handled. That had a side effect that the build logic to inject the
NeutralResourcesLanguage attribute was not triggered when no Strings.resx
was detected.

cc @tarekgh  @ericstj


stephentoub:
LGTM

ericstj:
LGTM :shipit:

tarekgh:
LGTM too.

PR #2133, opened 2015-06-23
System.Linq.Parallel - Other cleanup

Clockwork-Muse:
Cleanup of some of the other files and tests in PLINQ.


Clockwork-Muse:
From the look of things, the tests are still being run in parallel (just maybe not PLINQ's individually?), which allowed it to deadlock.  
...I'm not sure how the original version was able to avoid the deadlock.

Clockwork-Muse:
PLINQ was also affected by #2239 .  Added tests and the fix for it.

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
+        };
+
+        public static IEnumerable<object[]> PartitioningData(object[] counts)
         {
-            for (int i = 1; i <= 128; i *= 2)
-            {
-                SimplePartitionMergeWhereScanTest1(1024 * 8, i, true);
-                SimplePartitionMergeWhereScanTest1(1024 * 8, i, false);
-            }
+            foreach (object[] results in Sources.Ranges(counts.Cast<int>(), x => new[] { 1, 2, 4 })) yield return results;
```
stephentoub: This is a long single-line of code.  Can you reformat it to be multiple lines? (Same for the other similar ones below.)

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
+                Labeled<ParallelQuery<int>> labeled = Labeled.Label("ThrowOnEnumeration " + count, new ThrowOnEnumeration<int>(count).AsParallel().AsOrdered());
+                foreach (ParallelMergeOptions option in Options)
+                {
+                    yield return new object[] { labeled, count, option };
+                }
             }
+        }
+
+        // The following tests attempt to test internal behavior,
+        // but there doesn't appear to be any way to deterministically observe it.
```
stephentoub: I don't understand this comment.  You're saying we're adding tests for things that are not deterministic such that they could randomly fail?

src/System.Linq.Parallel/tests/ExchangeTests.cs
stephentoub: It'd be helpful to add comments for what this is yielding.  The 1,2,4 are numbers of partitions, right?

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
-            PartitioningTestCore(true, 1, 0, 500);
-            PartitioningTestCore(false, 1, 0, 520);
+            Merge_Ordered(labeled, count, options);
+        }
 
-            PartitioningTestCore(true, 4, 0, 500);
-            PartitioningTestCore(false, 4, 0, 520);
+        [Theory]
+        [MemberData("MergeData", (object)(new int[] { 16, 1024 }))]
+        [MemberData("FailingMergeData", (object)(new int[] { 16, 1024 }))]
```
stephentoub: I'm confused by the naming.  What is "FailingMergeData"?  I'd have expected to see some Asserts.Throws or try/catches or something in the test, as I assumed from the name that meant something was going to fail and throw.

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
-            OrderedPipeliningTest1(10007, false);
-
-            OrderedPipeliningTest2(true);
-            OrderedPipeliningTest2(false);
+            Assert.Throws<ArgumentNullException>(() => ((ParallelQuery<int>)null).WithMergeOptions(ParallelMergeOptions.AutoBuffered));
         }
 
         /// <summary>
-        /// Checks whether an ordered pipelining merge produces the correct output.
+        /// The plinq chunk partitioner calls takes an IEnumerator over the source, and disposes the enumerator when it is
```
stephentoub: Nit: typo "calls takes"

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
+        /// finished.
+        /// If an exception occurs, the calling enumerator disposes the enumerator... but then other callers generate ODEs.
+        /// These ODEs either should not occur (preferred), or should not get into the aggregate exception.
```
stephentoub: I don't understand this exception.  Are you commenting about what you believe to be a bug in the implementation?  If so, why is this in the XML comment for a test?

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,177 @@ namespace Test
...
+        ///
+        /// Also applies to the plinq stand-alone chunk partitioner.
+        /// Does not apply to other partitioners unless an exception in one consumer would cause flow-on exception in others.
         /// </summary>
-        private static void OrderedPipeliningTest2(bool buffered)
+        /// <returns></returns>
+        [Theory]
+        [MemberData("BinaryRanges", new int[] { 16 }, new int[] { 16 }, MemberType = typeof(UnorderedSources))]
```
stephentoub: I'm having trouble following along with all of these theories, in particular because I have to jump somewhere else in the code to understand what function is feeding in the data, with arguments to that other function being provided in this attribute without names.  Can we do something to make this all more understandable, e.g. use comments consistently to describe what these things all are?

src/System.Linq.Parallel/tests/Helpers/Comparers.cs
```diff
@@ -67,6 +72,17 @@ public int Compare(int x, int y)
...
     }
 
+    internal class ExtremeComparer<T> : IComparer<T>
+    {
+        private IComparer<T> _def = Comparer<T>.Default;
+
+        public int Compare(T x, T y)
+        {
+            int direction = _def.Compare(x, y);
+            return direction == 0 ? 0 : direction > 0 ? int.MaxValue : int.MinValue;
```
stephentoub: Can you split this across multiple lines, e.g.
```C#
return
    direction == 0 ? 0 :
    direction > 0 ? int.MaxValue :
    int.MinValue;
```
?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
 {
     public static class PlinqModesTests
     {
-        [Fact]
-        public static void RunPlinqModesTests()
+        private static IEnumerable<Labeled<Action<ParVerifier, ParallelQuery<int>>>> EasyUnorderedQueries()
```
stephentoub: What does "Easy" mean here?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
...
-                (mode,verifier) => ParallelEnumerable.Range(0, 1000).WithExecutionMode(mode)
-                    .Select(x => verifier.Verify(x)).TakeWhile(x => true).Iterate(),
-
-                (mode,verifier) => ParallelEnumerable.Range(0, 1000)
-                    .OrderBy(x=>x).Select(x => verifier.Verify(x)).WithExecutionMode(mode).ElementAt(5),
-
-                (mode,verifier) => ParallelEnumerable.Range(0, 1000).WithExecutionMode(mode)
-                    .OrderBy(x=>x).Select((x,i) => verifier.Verify(x)).Iterate(),
+            // Some queries may be brittle, depending on source type.
+            // In particular, ParallelEnumerable.Range with Take+Select+foreach failed until the count in Take was increased.
```
stephentoub: Failed how? And why?  Increased from what?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
...
+            yield return Labeled.Label<Action<ParVerifier, ParallelQuery<int>>>("Select+TakeWhile+ToArray", (verifier, query) => query.Select(x => verifier.Verify(x)).TakeWhile(x => true).ToArray());
+            yield return Labeled.Label<Action<ParVerifier, ParallelQuery<int>>>("Select+TakeWhile+foreach", (verifier, query) => query.Select(x => verifier.Verify(x)).TakeWhile(x => true).ForEach());
+            yield return Labeled.Label<Action<ParVerifier, ParallelQuery<int>>>("OrderBy+Select+ElementAt", (verifier, query) => query.OrderBy(x => x).Select(x => verifier.Verify(x)).ElementAt(8));
+            yield return Labeled.Label<Action<ParVerifier, ParallelQuery<int>>>("OrderBy+Select+foreach", (verifier, query) => query.OrderBy(x => x).Select(x => verifier.Verify(x)).ForEach());
+            yield return Labeled.Label<Action<ParVerifier, ParallelQuery<int>>>("Where+Select+OrderBy+Take+foreach", (verifier, query) => query.Where(x => true).Select(x => verifier.Verify(x)).OrderBy(x => x).Take(128).ForEach());
+        }
```
stephentoub: How did you choose this combination of queries?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
...
 
-            // Verify that all queries in 'easyQueries' always run in forced mode
-            for (int i = 0; i < hardQueries.Length; i++)
+        public static IEnumerable<object[]> HardQueryData(object[] counts, object[] modes)
+        {
+            // Test doesn't apply to DOP == 1.  It verifies that work is actually
+            // happening in parallel, which won't be the case with DOP == 1.
+            if (Environment.ProcessorCount == 1) yield break;
+
+            foreach (object[] results in UnorderedSources.Ranges(counts))
```
stephentoub: EasyQueryData returns both unordered and ordered, but HardQueryData is just unordered?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
...
-        private abstract class Verifier
+        [Fact]
+        public static void WithExecutionMode_ArgumentNullException()
         {
-            internal abstract int Verify(int x);
+            Assert.Throws<ArgumentNullException>(() => ((ParallelQuery<int>)null).WithExecutionMode(ParallelExecutionMode.Default));
+        }
 
-            internal abstract bool Passed { get; }
+        private static void ForEach<T>(this IEnumerable<T> e)
```
stephentoub: Kind of a strange name to me, in particular because it doesn't take an Action<T> to run "for each" element.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,135 @@ namespace Test
...
 
-            internal override int Verify(int x)
+            internal int Verify(int x)
             {
                 lock (this)
                 {
                     _counter++;
                     if (_counter == 1)
                     {
-                        if (Monitor.Wait(this, TIMEOUT_LIMIT))
```
stephentoub: Why did you remove the timeout?

stephentoub:
Thanks for the further PLINQ work, @Clockwork-Muse.  I went through the PR and left some comments.  At a high-level, could you separate out the the src changes from the test changes?  They're mostly unrelated.  I mentioned this in one or two of the comments, but I'd also find some more commenting on the tests helpful, in particular of what the theory inputs are.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
Clockwork-Muse: I believe "Easy" was chosen by the original author to denote queries that PLINQ would have an easy time parallelizing - so they'd be run that way by default (in contrast to hard, which might not).  I can add some explanatory comments; unless you have a better name?  I just stuck with the original, because I didn't really have anything better.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
Clockwork-Muse: I didn't - those are the original ones, I just translated them.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
Clockwork-Muse: "Because it offended my sensibilities".  Probably much better to fail the test than hang, yes, changing back.

Clockwork-Muse:
Separated out the source changes (with the related tests).  I'll post that one later, as I found at least one other minor fix (`Average<long?>()` doesn't use `checked`)

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: I was trying to say "We have tests for internal behavior, but there isn't a reliable way to externally observe it" (The tests already existed, I just translated/widened them.  I otherwise left them in their "current" state). 

 Take the "partitioning" case.  From the DegreesOfParallelism tests, we know PLINQ will use multiple threads (.... eventually, when large).  What we don't have a direct path to observe is the number of internal partitions it creates and uses, at least when using the current types directly (array, enumerable, etc).

I guess I could write that custom partitioner, for these tests?  The feeling I got was that degrees of parallelism wasn't necessarily directly related to the partitions created. 
(Although I'm realizing you're right, and I should create a bunch of custom implementations for various error conditions, like null enumerables out of `GetDynamicPartitions`).

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: Side note: do you want me to hit everything, everywhere (this or a separate commit)?  Some of the stuff already in is 200+ characters.  That particular line is ~120, what's the limit you want?

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: Not me - that's the original.  Revising.

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: I'm too close to the test implementation now, everything just makes sense to me.  Did you mean adding something like:
```
// Generate combinations of two ParallelQuery s from different sources with the given counts.
```
Of course, that's still clear to me...
(I was hoping to keep the external stuff simple enough I wouldn't need comments for this...)

src/System.Linq.Parallel/tests/PlinqModesTests.cs
Clockwork-Muse: I thought it had been failing for ordered queries (not being run in parallel even when forced), but now I'm not sure why.  It certainly isn't failing now.  Adding it.

src/System.Linq.Parallel/tests/ExchangeTests.cs
stephentoub: I don't think we have a guideline set (though maybe we should).  In general for me it's less about length and more about parsability... there's just a lot going on here (e.g. a loop, a variable being declared, multiple methods being called, LINQ being used, data being yielded, etc.), and I found it more difficult to comprehend what was going on with it on a single line.

src/System.Linq.Parallel/tests/ExchangeTests.cs
stephentoub: > The feeling I got was that degrees of parallelism wasn't necessarily directly related to the partitions created

It controls the number of partitions created / the number of tasks created, but that doesn't provide any guarantees about how much work each task will end up doing.  If, for example, the partitioning scheme employed has each task going to the source to get the next item as it's needed, then the first task to be scheduled could end up running so quickly that it ends up completing all of the processing before any of the other tasks get a chance to help.

> We have tests for internal behavior, but there isn't a reliable way to externally observe it

Ok, that makes more sense to me.  Thanks.

src/System.Linq.Parallel/tests/ExchangeTests.cs
stephentoub: It might be sufficient just to add comments to the various MemberData functions (that describe the inputs and outputs).  I'm reading a test, I see it has a MemberData attribute, so I go in search of that, then when I find it I have to read its implementation to understand what data it's providing based on the inputs, and at that point I've lost the context of the test I was originally looking at.  If the MemberData function at least had a comment describing what it was doing, I might have an easier time.  (It'd be even nicer if VS could provide that in IntelliSense at the usage site, but as these are just strings, it has little chance of that... maybe once we move to C# 6 these can be switched to use nameof.)

src/System.Linq.Parallel/tests/PlinqModesTests.cs
stephentoub: I see, thanks.  EasilyParallelizedQueries?

src/System.Linq.Parallel/tests/PlinqModesTests.cs
stephentoub: Ah, ok.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
stephentoub: :smile:

stephentoub:
@Clockwork-Muse, thanks for continuing to work on this.  Seems like the majority of remaining things are about reformatting, revising comments, etc.  Rather than hold up this PR for thiat (since much of it was preexisting), shall we just go ahead and merge this, and then you can do such follow-up separately?

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: Okay, that makes more sense.  Do you want me to cover everything I've already done, or only things already in this PR?  I am working on more stuff for PLINQ anyways (expanding cancellation, for one).

stephentoub: I'm ok with it being done separately.

stephentoub:
@Clockwork-Muse, I just noticed that not everything you'd said you'd address (e.g. reverting the removal of the timeout on the Monitor.Wait) has actually been addressed, so I'll hold off on merging this.

Clockwork-Muse:
@stephentoub - that's because I hadn't uploaded any of them, so from your perspective _none_ of them had been addressed!
I've now pushed, and once that passes review I'll squash.

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,180 @@ namespace Test
...
         }
 
-        [Fact]
-        [OuterLoop]
-        public static void SimplePartitionMergeWhereScanTest_LongRunning()
+        // For each source, run with each buffering option.
+        public static IEnumerable<object[]> MergeData(object[] counts)
         {
-            for (int i = 1; i <= 128; i *= 2)
+            foreach (object[] results in Sources.Ranges(counts.Cast<int>(), x => Options))
```
stephentoub: For the previous two descriptions, how does ``counts``` factor in?

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,180 @@ namespace Test
...
             {
-                SimplePartitionMergeWhereScanTest1(1024 * 1024 * 2, i, true);
-                SimplePartitionMergeWhereScanTest1(1024 * 1024 * 2, i, false);
+                yield return results;
             }
         }
 
-        private static void SimplePartitionMergeWhereScanTest1(int dataSize, int partitions, bool pipeline)
+        // For each count, return an Enumerable source that fails (throws an exception) on that count,
+        // with each buffering option.
```
stephentoub: That description is helpful, thanks.  Should the method itself be renamed then to something more descriptive, i.e. instead of ```FailingMergeData```, something like ```ThrowOnCount_AllMergeOptions_MemberData```?  I'm trying to come up with ways that make the tests self-explanatory (to folks besides you :smile:) without having to do lots of exploring and code reading.

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,180 @@ namespace Test
...
-        [Fact]
-        public static void SimplePartitionMergeWhereScanTest()
+        private static readonly ParallelMergeOptions[] Options = new[] {
+            ParallelMergeOptions.AutoBuffered,
+            ParallelMergeOptions.Default,
+            ParallelMergeOptions.FullyBuffered,
+            ParallelMergeOptions.NotBuffered
+        };
+
+        // For each source, test with different degrees of parallelism to force PLINQ to partition.
```
stephentoub: If you made these XML comments instead, then when we eventually switch to using ```nameof(PartitioningData)``` instead of ```"PartitioningData```, hovering over ```PartitioningData``` at the usage site will show the description in IntelliSense.

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -10,183 +11,180 @@ namespace Test
...
         }
 
-        /// <summary>
-        /// Checks whether an ordered pipelining merge produces the correct output.
-        /// </summary>
-        private static void OrderedPipeliningTest1(int dataSize, bool buffered)
+        // The plinq chunk partitioner takes an IEnumerator over the source, and disposes the
+        // enumerator when it is finished. If an exception occurs, the calling enumerator disposes
+        // the source enumerator... but then other worker threads may generate ODEs.
+        // This test verifies any such ODEs are not reflected in the output exception.
```
stephentoub: Much better, thanks.

src/System.Linq.Parallel/tests/Helpers/Enumerables.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Test
+{
+    internal class ThrowOnEnumeration<T> : IEnumerable<T>
```
stephentoub: Did you consider replacing such a custom enumerable with something like:
```C#
public static IEnumerable<T> ThrowOnEnumeration() 
{ 
    return ThrowOnEnumeration(1); 
}

public static IEnumerable<T> ThrowOnEnumeration(int count)
{
    return Enumerable.Range(0, count).Select(i => {
        if (i == count - 1) throw new DeliberateTestException();
        return default(T);
    });
}
```
?

stephentoub:
Thanks, @Clockwork-Muse.  A few little nits, but otherwise LGTM.  Please take a look at the comments, address what you feel is appropriate, and squash.  I'll take another quick pass through then before merging.

src/System.Linq.Parallel/tests/Helpers/Enumerables.cs
Clockwork-Muse: ... I'm liking LINQ, but tend to still think too much in creating my own enumerables.  Thanks, though: stealing.

Clockwork-Muse:
@stephentoub - ... and done.
I've added doc comments for the UnorderedSources (which are pretty heavily used, and were touched in this branch anyways), but will hold off on doing that for anything else until `nameOf` becomes available.

src/System.Linq.Parallel/tests/PlinqModesTests.cs
```diff
@@ -11,157 +11,195 @@ namespace Test
...
         // A class that checks whether the Verify method got called from at least two threads.
         // The first call to Verify() blocks. If another call to Verify() occurs prior to the timeout
         // then we know that Verify() is getting called from multiple threads.
-        private class ParVerifier : Verifier
+        public class ParVerifier
         {
             private int _counter = 0;
             private bool _passed = false;
-            private const int TIMEOUT_LIMIT = 30000;
+            private const int TimeoutLimit = 100;
```
stephentoub: Hmm, 100 milliseconds seems dangerously low... a brief glitch during testing could easily exceed that.

stephentoub:
@Clockwork-Muse, I left one more comment that should be addressed before merging. (Sorry, not sure if I missed that before or if the issue wasn't there in the previous push, but I'm worried it could actually cause problems during CI runs.)

Clockwork-Muse:
Missed, but an easy thing.  Reverted.

stephentoub:
Thanks.

PR #2136, opened 2015-06-23
Refactoring DriveInfo on OS X, Linux, and FreeBSD

sokket:
Refactoring DriveInfo to work on Linux, OS X, and FreeBSD 

@stephentoub @ellismg @josteink @ghuntley

This moves DriveInfo into 3 distinct areas -
1) DriveInfo.cs which is OS agnostic
2) DriveInfo.Unix.cs which is common to all OSes where we can use statfs to get info about a specific mount point
3) DriveInfo.\<OS\>.cs which is specific to each OS, which is where we need to enumerate all mount points, which does not have a common PInvoke for platforms

Also fixing readdir to use the same pattern as DriveInfo


sokket:
Also fixed issue #815

josteink:
That looks way more thorough and proper than anything i would have prepared for FreeBSD alone. I really like the direction this is heading with regard to code reuse.Commenting on the FreeBSD specific stuff that looks good to me, but I haven't had the chance to verify the correctness of it on proper iron yet. @janhenke ?

src/Common/src/Interop/BSD/libc/Interop.readdir.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
```
janhenke: There is no "BSD" interop until now. We discussed it in the FreeBSD team and feel that is not a good idea.

This file should either in the Unix folder, if it works for most Unix like platforms or in the FreeBSD one is if just works on (Free)BSD.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
```diff
@@ -19,6 +19,12 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Release|AnyCPU' " />
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Debug|AnyCPU' " />
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Debug|AnyCPU' "/>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Release|AnyCPU' "/>
```
janhenke: We kept alphabetical ordering until now. Just a nit, but would be nice.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
```diff
@@ -19,6 +19,12 @@
...
+  <PropertyGroup>
+    <TargetsBSD Condition="'$(TargetsOSX)' == 'true' or '$(TargetsFreeBSD)' == 'true'">true</TargetsBSD>
+  </PropertyGroup>
```
janhenke: We should not introduce a "is BSD" here. We already have the "is Unix". Also OS X is not a BSD system, even if inherited some parts from it.

Please remove the "is BSD" part.

src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.BSD.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
janhenke: Again we are not confident about having a generic "BSD" code path. This should either go to the Unix folder or the specific FreeBSD one.

janhenke:
I do not like at all to start a "BSD" flavour in our interop and so forth. The FreeBSD porting team agreed that it does not make sense. We already have the generic Unix category, which should contain all code that is common to most of the Unix-like platforms (currently for 2 out of the 3 FreeBSD, Linux and OS X).

Please move this "BSD" code to either the generic Unix part of the platform specific folders.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
stephentoub: @janhenke, I'm not understanding the objection to "BSD" in cases like this.  We need some way to identify the functions that come from the BSD specification, e.g. from the man page at https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/statfs.2.html comments like "The statfs() function first appeared in 4.4BSD" and "BSD System Calls Manual".  What would you prefer we call it?

josteink:
I've now verified the implementation of `DriveInfo.GetDrives()` in the most basic sense on FreeBSD and we do get the correct results back.

Good job, @sokket . Not bad for working blind-folded :)

sokket:
@josteink No problem, thanks a bunch for verifying! I'm glad I could help you guys out :)

josteink:
Extending it to dump DriveType causes some errors to be printf'ed, but this is clearly better than what we used to have :)

sokket:
@josteink What sort of errors? It might be a struct signature issue...

josteink:
Using this test-program:

````c#
public static class Program
{
        {
                var drives = System.IO2.DriveInfo.GetDrives();
                foreach (var drive in drives)
                {
                        Console.WriteLine(drive.ToString());
                        Console.WriteLine(drive.DriveType);
                }
        }
}
````

I get this output:

````
/
472
Should be here
Unknown
/dev
472
Should be here
Ram
/tmp
472
Should be here
Unknown
/usr/home
472
Should be here
Unknown
/usr/ports
472
Should be here
Unknown
/usr/src
472
Should be here
Unknown
/var/crash
472
Should be here
Unknown
/var/log
472
Should be here
Unknown
/var/mail
472
Should be here
Unknown
/var/tmp
472
Should be here
Unknown
/dev/fd
472
Should be here
Unknown
/usr/compat/linux/proc
472
Should be here
Unknown
/proc
472
Should be here
Unknown
````

It seems like this code here gets triggered, so at least it's a known error-case :)

https://github.com/dotnet/corefx/pull/2136/files#diff-780b231eab7cf8e6a47933108ea64cf6R57

File: src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+using System.Runtime.InteropServices;
+
+using uid_t = System.UInt32;
+
+internal static partial class Interop
+{
+    internal static partial class libc
+    {
+        [DllImport("libc")]
+        static extern int printf(string data);
```
stephentoub: This printf DllImport should be removed.

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                int errno = Marshal.GetLastWin32Error();
+                if (errno == Interop.Errors.ENOENT)
+                    throw new System.IO.DriveNotFoundException(SR.Format(SR.IO_DriveNotFound_Drive, name)); // match Win32 exception
+                else
+                    throw Interop.GetExceptionForIoErrno(errno, isDirectory: true);
+            }
+            else
+            {
+                printf(System.Runtime.InteropServices.Marshal.SizeOf<statfs>().ToString() + "\r\n");
```
stephentoub: These printfs should be removed.

src/Common/src/Interop/FreeBSD/libc/Interop.statfs.cs
```diff
@@ -0,0 +1,54 @@
...
+            internal fixed byte f_mntfromname[MNAMELEN];    /* mounted filesystem */
+            internal fixed byte f_mntonname[MNAMELEN];      /* directory on which mounted */
+        }
+
+        internal static unsafe String GetMountPointFsType(statfs data)
+        {
+            return Marshal.PtrToStringAnsi((IntPtr)data.f_fstypename);
+        }
+    }
+}
```
stephentoub: nit: missing newline

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
sokket: Interesting...I don't have them on my local machine and my repo is clean; I thought I removed them before the PR but maybe I didn't push.

src/Common/src/Interop/Linux/libc/Interop.statfs.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    internal static partial class libc
+    {
+        /// <summary>
+        /// Internal FileSystem names and magic numbers taken from man(2) statfs
+        /// </summary>
+        /// <remarks>
+        /// These value names MUST be kept in sync with those in DriveInfo.Unix.GetDriveType
+        /// </remarks>
+        internal enum LinuxFileSystemTypes : long
```
stephentoub: Can this enum be private?

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
sokket: Same as below...I might not have pushed to origin before the PR, which would have removed these. Let me check

src/Common/src/Interop/Linux/libc/Interop.statfs.cs
```diff
@@ -0,0 +1,94 @@
...
+            internal long f_flags;
+            internal fixed long f_space[4];
+        }
+
+        internal static unsafe String GetMountPointFsType(statfs data)
+        {
+            return ((LinuxFileSystemTypes)data.f_type).ToString();
+        }
+    }
+}
```
stephentoub: nit: missing newline... will stop commenting on these

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
sokket: Oops, my findstr was pointing to the wrong directory so I missed these....removing now (same with all printfs)

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
sokket: Fixed (see above)

src/Common/src/Interop/OSX/libc/Interop.statfs.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Runtime.InteropServices;
+
+using uid_t = System.UInt32;
+
+internal static partial class Interop
+{
+    internal static partial class libc
+    {
+        internal const int MAXPATHLEN = 1024;
+        internal const int MFSTYPENAMELEN = 16;
```
stephentoub: Can we make these consts private?

src/Common/src/Interop/FreeBSD/libc/Interop.statfs.cs
sokket: Fixed

sokket:
@josteink Those are debug printf's that slipped through the crack...pushing a fix to remove them :)

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
```diff
@@ -0,0 +1,62 @@
...
+        internal static unsafe extern int getmntinfo(statfs** ppBuffer, int flags);
+
+        /// <summary>
+        /// Gets a statfs struct for the given path that describes that mount point
+        /// </summary>
+        /// <param name="path">The path to retrieve the statfs for</param>
+        /// <param name="buffer">The output statfs struct describing the mount point</param>
+        /// <returns>Returns 0 on success, -1 on failure</returns>
+        [DllImport(Interop.Libraries.libc, EntryPoint = "statfs" + Interop.Libraries.INODE64SUFFIX, SetLastError = true)]
+        private static unsafe extern int get_statfs(string path, statfs* buffer);
```
stephentoub: Could this use out instead of a pointer?

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
```diff
@@ -5,6 +5,12 @@ internal static partial class Interop
...
     private static partial class Libraries
     {
-        internal const string LibRt = "librt"; // POSIX Realtime Extensions library
+        /// <summary>
+        /// We aren't OS X so don't have an INODE64 suffix to entry points
+        /// </summary>
+        internal const string INODE64SUFFIX = "";
+
+        internal const string LibRt = "librt";  // POSIX Realtime Extensions library
+        internal const string libc = "libc";    // C runtime
```
stephentoub: Why is libc being defined in each of these Interop.Libraries.cs files... isn't it already in the Unix/Libraries.Interop.cs file?  Are we not including that for some reason?

src/Common/src/Interop/Linux/libc/Interop.statfs.cs
sokket: Fixed

src/Common/src/Interop/Linux/libc/Interop.statfs.cs
sokket: Fixed

src/Common/src/Interop/OSX/libc/Interop.statfs.cs
sokket: Fixed

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
sokket: I can't include the Unix Libraries.cs file since that's included in OS X and Linux; I need to declare the INODE64SUFFIX constant for each platform (ie the Libraries.cs file) since there isn't a shared file between BSD and Linux that doesn't also include OS X. Due to this constant and the niceness of code re-use by this refactor, it made it necessary to include libc in each platform's Libraries.cs. It seemed like a small price to pay for all the goodness we get from code re-use; however, I'm definitely open to cleaner solutions :)

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
sokket: Fixed

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
stephentoub: I might be misunderstanding, but I wasn't suggesting just including the Unix Interop.Libraries.cs, I suggesting included the Unix Interop.Libraries.cs, and then also including each of the platform-specific Interop.Libraries.cs as is relevant, and libc would remain in the Unix one.

janhenke: @sokket Thanks for your implementation, but still there is no "BSD". There are at least 4 major platforms with BSD in their name and we should not make the mistake to assume they all behave the same. Please drop that idea of a common BSD before this one get merged. Thanks!

stephentoub: @janhenke, he hasn't pushed new commits, so this is still the same set of commits you previously looked at and commented on.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
janhenke: Starting another "is BSD" condition into the code makes it very very complicated, since it means we had two layers of generic Unix-flavours. Written in a set notation: 

BSD = {FreeBSD, OS X} (I still very much deny OS X being a BSD), 
Unix = {Linux, BSD} = {Linux, {FreeBSD, OS X}}.

As I said we discussed this topic in the FreeBSD porting team and feel it is a path that will cause more problems than it solves in the future.

We should stick to the solution we used before. If it is common among most Unix platforms, it should be implemented in a common "Unix" file. If one platform, e.g. Linux, does not have it, add a file just for LInux specializing the behaviour away from the "normal" Unix implementation.

Also OS X is not really a BSD. It is a strange hybrid, with some parts coming from FreeBSD while other parts are completely different.

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
janhenke: Mea culpa, sorry.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
stephentoub: > Starting another "is BSD" condition into the code makes it very very complicated

I don't see how that's the case.  Yes, it adds another level; I personally think that helps to simplify things, not add complication.  If there's a logical way of saying that a bunch of things are relevant together, then in my opinion that organization helps to lend clarity.  It becomes confusing if we have a bunch of functionality under Unix that doesn't apply to Linux; the case we saw before wasn't about a function not applying to OSX, but rather about that function needing to be specialized for OS X slightly.  In contrast, getmntinfo doesn't exist on Linux at all.

> that will cause more problems than it solves in the future.

If it causes problems in the future, we can revise it in the future. :smile:

> We should stick to the solution we used before.

That made sense when there was no logical grouping that allowed for it to be separated out.  That's not the case here.

> Also OS X is not really a BSD

No one's claiming it's BSD; the claim is that these functions are from BSD.  Both the FreeBSD man page (http://www.freebsd.org/cgi/man.cgi?query=getmntinfo&apropos=0&sektion=0&manpath=FreeBSD+10.1-RELEASE&arch=default&format=html) and the OS X man page (https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/getmntinfo.3.html) agree on that, stating that these functions "function first appeared in 4.4BSD".  And at least for our needs (if not entirely), the function behaves the same way on both platforms, as specified for BSD.

I'm not trying to be difficult, but I'm honestly not understanding the problem.

janhenke: I think we have a naming problem here. BSD has a well defined meaning, referring to the open source OSes that derive from the original BSD release. That means FreeBSD, NetBSD, OpenBSD, DargonFly BSD etc. OS X is not a BSD, as we seem to agree.

Still that already defined term is used here in a different way, to refer to an API originating from the BSD family, not the OSes themselves. If at all, at least it should never be "TargetsBSD" but rather "TargetsBSDAPI". The latter makes clear we are speaking about API classification, not platform classification. Still better would be to just remove the "TargetsBSD" all together and below just include the individual files in the "TargetsFreeBSD" and "TargestOSX" sections.

Still I would strongly suggest to not add another folder in the Interop folder. Rather put those files in the Unix folder and name them as *.BSDAPI.cs.

We really must be clear and precise what exactly we are talking about. Using a term like BSD in this way is not really helping.

With the above points taken into account, I am fine with having implementations named by API families. But we really have to be clear and do not mix platform names and API family names in a confusing manner like it happened here.

stephentoub: > That means FreeBSD, NetBSD, OpenBSD, DargonFly BSD etc. OS X is not a BSD, as we seem to agree.

I wasn't making any claims about whether OS X is or is not BSD. I personally don't see that as particularly important, at least not for this aspect of the project (I agree it's more interesting if you start exposing functionality that tells you whether something is or is not something, but even then it's almost always being used as a proxy to understand whether something does or does not support something). What matters to me is whether a particular API can be used on a particular OS and being able to describe the heritage of that API so as to more easily determine its potential applicability elsewhere and to help differentiate it when needed.

> to refer to an API originating from the BSD family

Yes, that's how the entire Interop folder and Targets* scheme is set up. If it makes you more comfortable, imagine that "API" is an implicit suffix on the name of all the Interop subfolders (including Windows) and on all of the Targets conditions, as that is what's meant. But as it's relevant everywhere, it doesn't need to be included anywhere.

> Rather put those files in the Unix folder and name them as *.BSDAPI.cs.

Taking that logic to the extreme, you could imagine a single Interop flat folder containing just files where every file ends in .WindowsAPI.cs, .UnixAPI.cs, .BSDAPI.cs, .OSXAPI.cs, .FreeBSDAPI.cs, etc. The whole point of the subfolders is to provide that same meaning but with more structure and less clutter.

janhenke: I still think the problem here is mixing names of APIs and platforms. So far it seemed to me that the Interop folders are for platforms, not for APIs. Because if we want to name the implementations by the origin of the APIs, we need a whole new naming scheme (starting with calling all POSIX APIs POSIX).

Also if we go for APIs, we must not have TargetOSX, TargetLinux etc. Then you can only define TargetsPOSIX2008, TargetsBSD4.4, TargetsWin32, etc. For any platform you then have to define the set of APIs that are implemented there.

Right now it is a big mess, we mix APIs and platforms, both in naming and implementations. Like this it just makes absolutely no sense. We have to decide for one system (either platforms or APIs) and stick with that. Not mix everything together.

josteink: Just throwing in a vote for the FreeBSD port-team and saying I agree with @janhenke. Like he says this has already been debated extensively inside the team and in the end the current direction received unanimous support.
I would really appreciate if we could get past bikeshedding this single issue in a otherwise most excellent and productive patch.
--
Jostein Kjønigsen
http://jostein.kjonigsen.net
On 24 Jun 2015 5:17 pm, Jan Henke <notifications@github.com> wrote:In src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj:
> @@ -19,6 +19,12 @@
>    <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Release|AnyCPU' " />
>    <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Debug|AnyCPU' " />
>    <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Release|AnyCPU' " />
> +  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Debug|AnyCPU' "/>
> +  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Release|AnyCPU' "/>
> +  <PropertyGroup>
> +    <TargetsBSD Condition="'$(TargetsOSX)' == 'true' or '$(TargetsFreeBSD)' == 'true'">true</TargetsBSD>
> +  </PropertyGroup>

I still think the problem here is mixing names of APIs and platforms. So far it seemed to me that the Interop folders are for platforms, not for APIs. Because if we want to name the implementations by the origin of the APIs, we need a whole new naming scheme (starting with calling all POSIX APIs POSIX).

Also if we go for APIs, we must not have TargetOSX, TargetLinux etc. Then you can only define TargetsPOSIX2008, TargetsBSD4.4, TargetsWin32, etc. For any platform you then have to define the set of APIs that are implemented there.

Right now it is a big mess, we mix APIs and platforms, both in naming and implementations. Like this it just makes absolutely no sense. We have to decide for one system (either platforms or APIs) and stick with that. Not mix everything together.

stephentoub: > Right now it is a big mess [...] Like this it just makes absolutely no sense.

Really? I do appreciate your point of view (and your contributions and insights), and I'm happy to have these kinds of discussions.  I'd like it if we could avoid exaggeration and hyperbole, though; at least that's how I interpret these statements.  Apologies if I've misconstrued them.

> Also if we go for APIs, we must not have TargetOSX, TargetLinux etc. Then you can only define TargetsPOSIX2008, TargetsBSD4.4, TargetsWin32, etc.

We could very well find that we need such specificity and level of categorization in the future; if and when we do, we can introduce it... the code base is malleable in that regard, and we're not locked into any one approach (I expect we'll see some changes here related to #2137, mostly around consolidating things which we've had to split apart).  But thus far we haven't seen such a need, and I don't believe we're currently in a place that would benefit from it.  While I agree that if we were to apply pure academic rigor such an approach could be warranted, perfection can be the enemy of the good, and I believe where we are strikes a nice balance.  If that changes as the code base grows, we'll all adapt with it.

> I would really appreciate if we could get past bikeshedding this single issue in a otherwise most excellent and productive patch.

Completely agree.  Jon should go ahead with fixing up the other issues in the PR and getting it checked in.

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
sokket: Fixed to remove Libc from the individual Libraries.cs and included the Unix Libraries.cs file to take that.

sokket:
Updated with a new version

stephentoub:
Thanks, Jon.

PR #2140, opened 2015-06-23
Bug fixes to OSX FSW

sokket:
Fixing bugs in the OSX FileSystemWatcher to make almost all tests pass (minus 3)

@stephentoub @ellismg

### DIFF
--- a/src/Common/src/Interop/OSX/Interop.EventStream.cs
+++ b/src/Common/src/Interop/OSX/Interop.EventStream.cs
@@ -21,7 +21,7 @@ internal static partial class Interop
         /// <summary>
         /// This constant specifies that we don't want historical file system events, only new ones
         /// </summary>
-        internal const ulong kFSEventStreamEventIdSinceNow = ulong.MaxValue;
+        internal const ulong kFSEventStreamEventIdSinceNow = 0xFFFFFFFFFFFFFFFF;
 
         /// <summary>
         /// Flags that describe what happened in the event that was received. These come from the FSEvents.h header file in the CoreServices framework.
--- a/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
+++ b/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
@@ -22,7 +22,6 @@ namespace System.IO
         {
             // Make sure we cleanup
             StopRaisingEvents();
-            CleanupStreamAndWatcher();
         }
 
         private void StartRaisingEvents()
@@ -35,7 +34,10 @@ namespace System.IO
                 // Always re-create the filter flags when start is called since they could have changed
                 if ((_notifyFilters & (NotifyFilters.Attributes | NotifyFilters.CreationTime | NotifyFilters.LastAccess | NotifyFilters.LastWrite | NotifyFilters.Size)) != 0)
                 {
-                    _filterFlags = Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemInodeMetaMod | Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemFinderInfoMod;
+                    _filterFlags = Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemInodeMetaMod  | 
+                                   Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemFinderInfoMod |
+                                   Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemModified      |
+                                   Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemChangeOwner;
                 }
                 if ((_notifyFilters & NotifyFilters.Security) != 0)
                 {
@@ -46,7 +48,6 @@ namespace System.IO
                     _filterFlags |= Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemIsDir |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemIsSymlink |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated |
-                                    Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemModified |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRemoved |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRenamed;
                 }
@@ -55,23 +56,11 @@ namespace System.IO
                     _filterFlags |= Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemIsFile |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemIsSymlink |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated |
-                                    Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemModified |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRemoved |
                                     Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRenamed;
                 }
 
-                // The directory has not changed then just restart receiving events. If the directory has changed 
-                // then we need to cleanup the old stream and kick off a new one
-                if (_lastWatchedDirectory.Equals(_directory, StringComparison.CurrentCultureIgnoreCase))
-                {
-                    RestartStream();
-                }
-                else
-                {
-                    CleanupStreamAndWatcher();
-                    CreateStreamAndStartWatcher();
-                    _lastWatchedDirectory = _directory;
-                }
+                CreateStreamAndStartWatcher();
             }
         }
 
@@ -82,8 +71,26 @@ namespace System.IO
             // RunLoop or StreamRef because IntPtrs are value types and can't be locked
             lock (_syncLock)
             {
-                // Be optimistic and only stop the stream, don't clean up unless we need to.
-                StopStream();
+                // Make sure there's a loop to clear
+                if (_watcherRunLoop != IntPtr.Zero)
+                {
+                    // Stop the RunLoop and wait for the thread to exit gracefully
+                    Interop.RunLoop.CFRunLoopStop(_watcherRunLoop);
+                    _watcherThread.Join();
+                    _watcherRunLoop = IntPtr.Zero;
+                }
+
+                // Clean up the EventStream, if it exists
+                if (!_eventStream.IsInvalid)
+                {
+                    _eventStream = new SafeEventStreamHandle(IntPtr.Zero);
+                }             
+
+                // Cleanup the callback
+                if (_callback != null)
+                {
+                    _callback = null;
+                }
             }
         }
 
@@ -92,12 +99,9 @@ namespace System.IO
         // -----------------------------
 
         // Flags used to create the event stream
-        private const Interop.EventStream.FSEventStreamCreateFlags EventStreamFlags = (Interop.EventStream.FSEventStreamCreateFlags.kFSEventStreamCreateFlagFileEvents |
+        private const Interop.EventStream.FSEventStreamCreateFlags EventStreamFlags = (Interop.EventStream.FSEventStreamCreateFlags.kFSEventStreamCreateFlagFileEvents | 
+                                                                                        Interop.EventStream.FSEventStreamCreateFlags.kFSEventStreamCreateFlagNoDefer   | 
                                                                                         Interop.EventStream.FSEventStreamCreateFlags.kFSEventStreamCreateFlagWatchRoot);
-
-        // The last directory we were told to watch. This allows us to be smart about how we create streams
-        // and to only teardown and recreate streams when we need to.
-        private string _lastWatchedDirectory = String.Empty;
 
         // The user can input relative paths, which can muck with our path comparisons. Save off the 
         // actual full path so we can use it for comparing
@@ -155,12 +159,15 @@ namespace System.IO
                 // Create the callback for the EventStream
                 _callback = new Interop.EventStream.FSEventStreamCallback(FileSystemEventCallback);
 
+                // Make sure the OS file buffer(s) are fully flushed so we don't get events from cached I/O
+                Interop.libc.sync();
+
                 // Create the event stream for the path and tell the stream to watch for file system events.
                 _eventStream = Interop.EventStream.FSEventStreamCreate(
                     _callback,
                     arrPaths,
                     Interop.EventStream.kFSEventStreamEventIdSinceNow,
-                    0.2f,
+                    0.0f,
                     EventStreamFlags);
                 if (_eventStream.IsInvalid)
                 {
@@ -176,58 +183,6 @@ namespace System.IO
                 _watcherThread = new Thread(new ThreadStart(WatchForFileSystemEventsThreadStart));
                 _watcherThread.Start();
                 _runLoopStartedEvent.WaitOne();
-            }
-        }
-
-        private void CleanupStreamAndWatcher()
-        {
-            // If we're currently streaming, stop
-            StopStream();
-
-            // Clean up the EventStream, if it exists
-            if (!_eventStream.IsInvalid)
-            {
-                _eventStream = new SafeEventStreamHandle(IntPtr.Zero);
-            }
-
-            // Make sure there's a loop to clear
-            if (_watcherRunLoop != IntPtr.Zero)
-            {
-                // Stop the RunLoop and wait for the thread to exit gracefully
-                Interop.RunLoop.CFRunLoopStop(_watcherRunLoop);
-                _watcherThread.Join();
-                _watcherRunLoop = IntPtr.Zero;
-            }          
-
-            // Cleanup the callback
-            if (_callback != null)
-            {
-                _callback = null;
-            }
-        }
-
-        private void RestartStream()
-        {
-            Debug.Assert(!_eventStream.IsInvalid);
-
-            // We don't need to rebuild the stream since the path is the same so just restart the stream.
-            if (Interop.EventStream.FSEventStreamStart(_eventStream) == false)
-            {
-                // Pass the error back to the user and set the last watched directory to null so that we'll
-                // cleanup and recreate our streams next time Start is called.
-                OnError(new ErrorEventArgs(new IOException("", Marshal.GetLastWin32Error())));
-                _lastWatchedDirectory = String.Empty;
-                _fullDirectory = String.Empty;
-            }
-        }
-
-        private void StopStream()
-        {
-            // Just stop the EventStream to be optimistic that we'll be restarted with the same path
-            // and not have to teardown everything and rebuild.
-            if (!_eventStream.IsInvalid)
-            {
-                Interop.EventStream.FSEventStreamStop(_eventStream);
             }
         }
 
@@ -281,6 +236,14 @@ namespace System.IO
 
             for (long i = 0; i < numEvents.ToInt32(); i++)
             {
+                // We need to special case the root path for pattern matching, so cache the path this way in case it is the root
+                // and use it for pattern matching. We'll go back to the eventPaths[i] to get the actual path for notifications
+                string path = eventPaths[i];
+                if (path.Equals(_fullDirectory, StringComparison.OrdinalIgnoreCase))
+                {
+                    path += "/.";
+                }
+
                 // First, we should check if this event should kick off a re-scan since we can't really rely on anything after this point if that is true
                 if (ShouldRescanOccur(eventFlags[i]))
                 {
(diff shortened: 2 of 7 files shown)


src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -387,7 +344,7 @@ private bool DoesPathPassNameFilter(string eventPath)
             Interop.EventStream.FSEventStreamEventFlags[] eventFlags,
             FSEventStreamEventId[] eventIds)
         {
-            Debug.Assert((currentIndex + 1 < eventPaths.Length) && (currentIndex + 1 < eventFlags.Length) & (currentIndex + 1 < eventIds.Length));
+            //Debug.Assert((currentIndex + 1 < eventPaths.Length) && (currentIndex + 1 < eventFlags.Length) && (currentIndex + 1 < eventIds.Length));
```
ellismg: Is there nothing we can assert here?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -295,7 +247,7 @@ private void WatchForFileSystemEventsThreadStart()
                 else if ((DoesPathPassNameFilter(eventPaths[i])) && ((_filterFlags & eventFlags[i]) != 0))
                 {
                     // Remove the base directory prefix (including trailing / that OS X adds)
-                    string relativePath = eventPaths[i].Remove(0, _fullDirectory.Length + 1);
+                    string relativePath = (eventPaths[i].Equals(_fullDirectory, StringComparison.CurrentCultureIgnoreCase) ? _fullDirectory : eventPaths[i].Remove(0, _fullDirectory.Length).TrimStart('/'));
```
ellismg: I think you want to be doing an ordinal comparison here (and I guess ignoring case, if that's what we do other places in OSX land).

src/System.IO.FileSystem.Watcher/tests/Utility/Utility.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading;
 using System.Runtime.CompilerServices;
 using Xunit;
+using System.Runtime.InteropServices;
```
ellismg: Seems like you could remove this and leave the file unchanged.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: The previous assumptions about this assert were incorrect in that we can have a rename without a pair when an item is moved out / in from / to the watcher directory. The current index may be the last item in the array of events, meaning the only thing we could assert is that the index is within the bounds of the array, but we would have other (more fundamental issues) if that ever occurred. I should remove the line though...forgot to do that

ellismg:
I would rebase and squash 9af76dd into the previous commit for a nicer history.

src/System.IO.FileSystem.Watcher/tests/Utility/Utility.cs
sokket: Ah forgot to remove that...thanks, will do

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -160,7 +164,7 @@ private void CreateStreamAndStartWatcher()
                     _callback,
                     arrPaths,
                     Interop.EventStream.kFSEventStreamEventIdSinceNow,
-                    0.2f,
+                    0.0f,
```
ellismg: Can you help me understand what was going on here before?  It seems like now we just disable coalescing from the OS, was the coalescing causing problems?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: Fixed

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
```diff
@@ -178,7 +178,14 @@ public DateTimeOffset CreationTime
...
                         default(DateTimeOffset);
                 }
-                set { } // Not supported
+                set 
+                {
+                    // The ctime in Unix can be interpreted differently by different formats so there isn't
+                    // a reliable way to set this; however, we can't just do nothing since the FileSystemWatcher
+                    // specifically looks for this call to make a Metatdata Change, so we should rev the 
+                    // LastAccessTime of the file to cause the metadata change we need.
+                    LastAccessTime = DateTimeOffset.Now;
```
stephentoub: Is a call to utime sufficient to trigger the event, or do we actually need to change the value?  I'm not loving actually changing the last access time if we can avoid it.  In other words, would this be sufficient to trigger the event?
```C#
LastAccessTime = LastAccessTime;
```
?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
stephentoub: Rather than doing a Remove and then a Trim, can we just manually count how many '/'s there are after _fullDirectory.Length and pass that value as the length to Remove?  As it is we're potentially creating two strings when we could instead just do one.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -326,23 +280,26 @@ private void WatchForFileSystemEventsThreadStart()
...
                         //       So look for deletes first since those will mean the listening app can't do anything with the item since it doesn't exist (such as stat 
                         //       the file to determine what exactly changed, in this example).
-                        if ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRemoved) == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRemoved)
-                        {
-                            NotifyFileSystemEventArgs(WatcherChangeTypes.Deleted, relativePath);
-                        }
-                        else if ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated) == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated)
+                        if ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated) == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated)
```
stephentoub: The length of these fully-qualified consts makes it a bit difficult to read, and the duplication of the name can lead to accidental typos; did you consider adding a helper somewhere like:
```C#
private static bool IsSet(int value, int flags)
{
    return (value & flags) != flags;
}
```
and then using it like:
```C#
if (IsSet(eventFlags[i], Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemCreated) { ... }
```

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -326,23 +280,26 @@ private void WatchForFileSystemEventsThreadStart()
...
+
+                        if (((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemInodeMetaMod)  == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemInodeMetaMod)  ||
+                            ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemModified)      == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemModified)      ||
+                            ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemFinderInfoMod) == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemFinderInfoMod) ||
+                            ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemChangeOwner)   == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemChangeOwner)   ||
+                            ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemXattrMod)      == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemXattrMod))
```
stephentoub: Such an IsSet could be used here as well.  If all of these flags are a single bit, you could change the IsSet to do ```!= 0``` instead of ```== flags``` and then make all of them in a single call rather than multiple checks and conditionals, e.g.
```C#
if (IsSet(eventFlags[i], 
    kFSEventStreamEventFlagItemInodeMetaMod |
    kFSEventStreamEventFlagItemModified |
    kFSEventStreamEventFlagItemFinderInfoMod |
    ...))
{
    ...
}
```

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -304,7 +256,9 @@ private void WatchForFileSystemEventsThreadStart()
                         long pairedId = FindRenameChangePairedChange(i, eventPaths, eventFlags, eventIds);
                         if (pairedId == long.MinValue)
                         {
-                            OnError(new ErrorEventArgs(new ArgumentOutOfRangeException("pairedId")));
+                            // Getting here means we have a rename without a pair, meaning it should be a create for the 
+                            // move from unwatched folder to watcher folder scenario
+                            NotifyFileSystemEventArgs(WatcherChangeTypes.Created, relativePath);
```
stephentoub: Could it also be the opposite, that we get a rename for the "from" but not for the "to" because the file moved out of the direction being watched rather than into the directory being watched?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: This value didn't seem to cause a difference in coalescing...the OS still coalesced events as expected. I'm not 100% sure what this value is supposed to affect since changing it to 0 made no noticeable difference; due to that, making it 0 is much more clear that it isn't some magic value

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
sokket: Looks like that works...changed to not actually change the time but still trigger the metadata mod

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: Fixed the comment

sokket:
@ellismg I'll squash before I put out my final iteration

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -295,16 +247,18 @@ private void WatchForFileSystemEventsThreadStart()
...
-                    if ((eventFlags[i] & Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRenamed) == Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRenamed)
+                    if (IsFlagSet(eventFlags[i], Interop.EventStream.FSEventStreamEventFlags.kFSEventStreamEventFlagItemRenamed))
                     {
                         // Find the rename that is paired to this rename, which should be the next rename in the list
                         long pairedId = FindRenameChangePairedChange(i, eventPaths, eventFlags, eventIds);
                         if (pairedId == long.MinValue)
                         {
-                            OnError(new ErrorEventArgs(new ArgumentOutOfRangeException("pairedId")));
+                            // Getting here means we have a rename without a pair, meaning it should be a create for the 
+                            // move from unwatched folder to watcher folder scenario or a move from the watcher folder out
```
stephentoub: Your most recent commit updated the comment to include moving out of the watched folder, but it didn't update the implementation accordingly... wouldn't moving out be treated as deletion, not creation?

sokket:
@stephentoub @ellismg updated with squashed version and all tests passing

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -292,19 +256,43 @@ private void WatchForFileSystemEventsThreadStart()
...
-                    // Remove the base directory prefix (including trailing / that OS X adds)
-                    string relativePath = eventPaths[i].Remove(0, _fullDirectory.Length + 1);
+                    // The base FileSystemWatcher does a match check against the relative path before combining with 
+                    // the root dir; however, null is special cased to signify the root dir, so check if we should use that.
+                    string relativePath = null;
+                    if (eventPaths[i].Equals(_fullDirectory, StringComparison.OrdinalIgnoreCase) == false)
+                    {
+                        // Check if the event path, with the root directory removed, begins with a / and
+                        // if so, remove it; otherwise, just remove the root path (which contains a trailing /)
+                        if (eventPaths[i][_fullDirectory.Length] == '/')
```
sokket: @stephentoub Now that I think about it, should this be Path.DirectorySeparatorChar?

sokket:
@dotnet-bot test this please

src/Common/src/Interop/OSX/Interop.EventStream.cs
```diff
@@ -21,7 +21,7 @@ internal static partial class EventStream
         /// <summary>
         /// This constant specifies that we don't want historical file system events, only new ones
         /// </summary>
-        internal const ulong kFSEventStreamEventIdSinceNow = ulong.MaxValue;
+        internal const ulong kFSEventStreamEventIdSinceNow = 0xFFFFFFFFFFFFFFFF;
```
stephentoub: Why this change?  Just curious.  I had to count the number of Fs to verify it was correct.

src/Common/src/Interop/OSX/libc/Interop.sync.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: I believe ```sync``` is POSIX compliant.  Consider putting it in the Unix rather than OSX directory.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -155,12 +159,16 @@ private void CreateStreamAndStartWatcher()
                 // Create the callback for the EventStream
                 _callback = new Interop.EventStream.FSEventStreamCallback(FileSystemEventCallback);
 
+                // Make sure the OS file buffer(s) are fully flushed so we don't get events from cached I/O
+                // like we saw in the Unit Tests
```
stephentoub: Comments like "like we saw in the Unit Tests" get stale pretty quickly... someone reading this in the not-too-distant-future won't know what that's referring to.  I'd suggest just cutting that phrase; the previous part of the comment does a good job explaining the reasoning.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
```diff
@@ -281,6 +237,14 @@ private void WatchForFileSystemEventsThreadStart()
 
             for (long i = 0; i < numEvents.ToInt32(); i++)
             {
+                // We need to special case the root path for pattern matching, so cache the path this way in case it is the root
+                // and use it for pattern matching. We'll go back to the eventPaths[i] to get the actual path for notifications
+                string path = eventPaths[i];
+                if (path.Equals(_fullDirectory, StringComparison.OrdinalIgnoreCase))
+                {
+                    path += System.IO.Path.DirectorySeparatorChar.ToString() + ".";
```
stephentoub: Consider caching this into a static readonly string, e.g.
```C#
private static readonly string CurrentDirectorySuffix = Path.DirectorySeparatorChar + ".";
...
path += CurrentDirectorySuffix;
```
As it's currently written, not only will string.Concat be called by the compiler to produce a new string, but each call will end up allocating a temporary string containing "/" for the separator.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
stephentoub: Ideally, but it's not a big deal in this case, since the file is already specific to OS X.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.cs
```diff
@@ -455,8 +455,9 @@ private void NotifyFileSystemEventArgs(WatcherChangeTypes changeType, string nam
                     Debug.Fail("Unknown FileSystemEvent change type!  Value: " + changeType);
                     break;
             }
-
-            if (handler != null && MatchPattern(name))
+            
+            bool passedPatternMatch = (string.IsNullOrEmpty(name) ? MatchPattern(_directory) : MatchPattern(name));
+            if (handler != null && passedPatternMatch)
```
stephentoub: This is going to invoke MatchPattern even if handler is null.  Will it be common for handler to be null?  You could instead write it like:
```C#
if (handler != null && MatchPattern(string.IsNullOrEmpty(name) ? _directory : name))
{
    handler(...);
}
```

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -75,4 +75,22 @@ public static void FileSystemWatcher_Changed_Negative()
             Utility.ExpectNoEvent(eventOccured, "changed");
         }
     }
+
+        [Fact]
```
stephentoub: Nit: odd indenting

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -75,4 +75,22 @@ public static void FileSystemWatcher_Changed_Negative()
...
+            watcher.Filter = "*";
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+
+            watcher.EnableRaisingEvents = true;
+
+            Directory.SetLastAccessTime(watcher.Path, DateTime.Now);
+
+            Utility.ExpectEvent(eventOccured, "changed");
+        }
+    }
```
stephentoub: Good, thanks for adding this test.

src/Common/src/Interop/OSX/Interop.EventStream.cs
sokket: This is just making it exactly what it is in the header file and not having to look up the specific value for MaxValue....it just made it more explicit

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: Good idea

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs
sokket: The second option looks like the best option...fixed

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.cs
sokket: Good idea, plus it keeps the if statement mostly like previous versions. Fixed

src/Common/src/Interop/OSX/Interop.EventStream.cs
stephentoub: Ok.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Oops...

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: :+1:

src/Common/src/Interop/OSX/libc/Interop.sync.cs
sokket: So it is :) added to the Unix directory

sokket:
@stephentoub @ellismg latest version passing with all comments fixed

ellismg:
LGTM

sokket:
Thanks all! Merging now

PR #2142, opened 2015-06-23
Check for state before fetching threadID

VSadov:
It is unlikely that fetching CurrentManagedThreadId is cheaper than fetching an int field, so lets check the state first.


VSadov:
CC @stephentoub 
simple change I meant to do a while ago.

stephentoub:
LGTM

Issue #2152, opened 2015-06-24
[Perf] Fix XmlDocument/XPath to properly compare strings by ref


krwq:
XmlDocument/XPath when comparing namespaces it uses fixed string object (from private class defined in XmlConst.cs). During the split of System.Xml.dll to smaller libraries (XmlDocument, ReaderWriter, XPath, XPath.XmlDocument) we introduced shared XmlConst.cs file for those constants and thus different references for those strings. To optimize implementation is comparing built-in namespaces by reference and thus in some cases it doesn't work properly and causes some undefined behavior. The cases which were breaking tests were already fixed although there are some other broken scenarios which got introduced when running product on Desktop (runtime).
Steps:
- [x] Workaround to not compare by ref (one line change in Ref.cs) to unblock Desktop ASAP (we are gonna lose some perf but correctness is more important than perf) - https://github.com/dotnet/corefx/pull/2154
- [ ] Make sure there is only one copy of string constants in one of the libraries (ReaderWriter or XmlDocument) and other libraries should get the valid and needed references to string some other way (i.e. through XmlDoc's NameTable or exposing something new in the contract if NameTable is too slow or not possible)
- [ ] Re-enable comparison by ref and validate it is passing when running on Desktop runtime (on Win 10) and coreclr

**Except for the first checkbox it is up for grabs**

Please revert Ref.cs to its original form with the last checkbox:
```csharp
        public static bool Equal(string strA, string strB)
        {
#if DEBUG
            if (((object)strA != (object)strB) && string.Equals(strA, strB))
                System.Diagnostics.Debug.Fail("Ref.Equal: Object comparison used for non-atomized string '" + strA + "'");
#endif
            return (object)strA == (object)strB;
        }
```

joshfree:
@ianhays do we have good perf test coverage here?

ianhays:
> do we have good perf test coverage here?

No, we don't have any. System.Xml.XPath.XDocument didn't come up in any of my usage data (though System.Xml.XDocument did). I can add it to my list of places to perf test next though.

krwq:
Closing this as System.Private.Xml should not doing that anymore

PR #2154, opened 2015-06-24
Fix XPath tests to run on Desktop

krwq:
Fixes 2 issues:
- New line endings issue - XmlDocument on Desktop is preserving \r\n while on CoreCLR it replaces it with \n (standard defines to replace it: http://www.w3.org/TR/2008/REC-xml-20081126/#sec-line-ends ; Desktop needs to remain as is for compat reasons)
- References issues - workaround for a bug introduced during the split of System.Xml.dll into smaller libraries: More details here: https://github.com/dotnet/corefx/issues/2152

### DIFF
--- a/src/System.Xml.XPath.XDocument/tests/XDocument/NavigatorComparer.cs
+++ b/src/System.Xml.XPath.XDocument/tests/XDocument/NavigatorComparer.cs
@@ -29,8 +29,7 @@ namespace System.Xml.XPath.XDocument.Tests.XDocument
 
             if (AreComparableNodes(a.NodeType, b.NodeType))
             {
-                Assert.Equal(a.Value, b.Value);
-                Assert.Equal(a.Value, b.Value);
+                CompareValues(a, b);
                 Assert.Equal(a.Name, b.Name);
             }
 #endif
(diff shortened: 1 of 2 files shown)


mellinoe:
Why aren't we just having the portable implementation match Desktop's behavior in this case (referring to the \n vs \r\n difference)?

src/System.Xml.XPath.XDocument/tests/XDocument/NavigatorComparer.cs
```diff
@@ -69,6 +68,14 @@ private static void CompareNodeTypes(XPathNodeType a, XPathNodeType b)
                 Assert.Equal(a, b);
             }
         }
+        
+        private static void CompareValues(XPathNavigator a, XPathNavigator b)
+        {
+            string sa = a.Value.Replace("\r\n", "\n").Replace("\r", "\n");
```
mellinoe: A comment might help explain why this is necessary

krwq:
The implementation which was preserving \r\n is not available directly in the surface area (XmlTextReader). To get it we would have to either expose it or add some new settings to XmlReader. Right now we are using XmlReader.Create which doesn't provide an option to not touch line endings - there is an option to preserve whitespaces but it doesn't affect line endings since they're defined by a standard. We do provide customization of line endings in the XmlWriter settings though.

Until there is no real need to add support for preserving line endings we decided to leave it as is especially that the behavior is exactly what XML specification defines.

src/System.Xml.XPath.XDocument/tests/XDocument/NavigatorComparer.cs
krwq: Will add before merging

mellinoe:
Sounds fine to me. Other than the comment in the test above, LGTM

src/System.Xml.XPath.XDocument/tests/XDocument/NavigatorComparer.cs
mellinoe: Sounds okay to me. Maybe "account for" instead of "eliminate"

krwq: :+1:

PR #2156, opened 2015-06-24
Update packages to latest

nguerrera:
cc @ericstj @weshaggard 

0c6b87e is just an msbuild /t:RestorePackages /p:LockDependencies of the whole tree to get the latest packages consumed by the build. (You'll likely want to review the other changes more carefully.)

f0f4cb8 makes sure that the upcoming revision of metadata references the upcoming revision of immutable.

97cd962 bumps the FileVersionInfo dependencies on these libraries to version with changes here rather than against the previous build on nuget.org.

463a8aa gets rid of the System.Collections.nupkg build from a build.cmd. We don't ship from that nupkg and it's just something left over that would get out of sync

a4c2db0 adjust the nuspec that we do use in the dev/metadata branch to have version numbers that are coherent with the master build. This  is just to avoid confusion since we don't do anything with the nupkgs out of master at this time.


ericstj:
LGTM, nice commit stats.  Maybe we start need to start using dotnet-bot for these package updates :wink:  (guilty)

nguerrera:
@ericstj Can you look at dc79c1c. I'm troubled that I had to do that as System.Reflection.Metadata using the old-style build should have worked, but maybe this is unique to dotnet/corefx's build system?

ericstj:
I suspect the failure you were seeing was because you were resolving the package as dnxcore50 (defined in your project.json) but the project itself was portable.  This would cause conflicts for all the assemblies brought in by the closure of Immutable.
 
We could instead try to resolve the project as .NETPortable,Version=v4.6,Profile=Profile7 but this would fail on the build machines because DNX requires the PCL targeting pack when installing to portable projects. https://github.com/aspnet/dnx/issues/1814

Pretty soon I am going to switch projects to using 'DotNET' moniker + guardrails so it will appear less bad that we're lying about targeting here.  To do that for this project we'll need the change for phone support in the packages I just checked in to TFS :wink:

nguerrera:
@ericstj So is this OK in the meantime?

ericstj:
Yeah.

PR #2160, opened 2015-06-25
Exclude injected SR functions from code coverage

stephentoub:
Build tools inject resource helper functions into any library that has resource strings.  These should be excluded from code coverage.

We already have an ExcludeFromCodeCoverage attribute in our Common folder, but I copied it here to be in the same file as SR.cs/vb so that build tools didn't need to inject yet another file dependency; I put the copy in a different namespace to avoid any conflicts.

### DIFF
--- a/src/Common/src/System/SR.cs
+++ b/src/Common/src/System/SR.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Resources;
+using System.Resources.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace System
@@ -12,6 +13,7 @@ namespace System
 
         private static ResourceManager ResourceManager
         {
+            [ExcludeFromCodeCoverage]
             get
             {
                 if (SR.s_resourceManager == null)
@@ -23,13 +25,15 @@ namespace System
         }
 
         // This method is used to decide if we need to append the exception message parameters to the message when calling SR.Format. 
-        // by default it returns false.
+        // It is hardcoded to return false, but on some platforms an IL transform may rewrite it to instead return true.
+        [ExcludeFromCodeCoverage]
         [MethodImpl(MethodImplOptions.NoInlining)]
         private static bool UsingResourceKeys()
         {
             return false;
         }
 
+        [ExcludeFromCodeCoverage]
         internal static string GetResourceString(string resourceKey, string defaultString)
         {
             string resourceString = null;
@@ -44,6 +48,7 @@ namespace System
             return resourceString;
         }
 
+        [ExcludeFromCodeCoverage]
         internal static string Format(string resourceFormat, params object[] args)
         {
             if (args != null)
@@ -59,6 +64,7 @@ namespace System
             return resourceFormat;
         }
 
+        [ExcludeFromCodeCoverage]
         internal static string Format(string resourceFormat, object p1)
         {
             if (UsingResourceKeys())
@@ -69,6 +75,7 @@ namespace System
             return String.Format(resourceFormat, p1);
         }
 
+        [ExcludeFromCodeCoverage]
         internal static string Format(string resourceFormat, object p1, object p2)
         {
             if (UsingResourceKeys())
@@ -79,6 +86,7 @@ namespace System
             return String.Format(resourceFormat, p1, p2);
         }
 
+        [ExcludeFromCodeCoverage]
         internal static string Format(string resourceFormat, object p1, object p2, object p3)
         {
             if (UsingResourceKeys())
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @tarekgh, @nguerrera, @ellismg

src/Common/src/System/SR.cs
```diff
@@ -89,3 +97,20 @@ internal static string Format(string resourceFormat, object p1, object p2, objec
         }
     }
 }
+
+namespace System.Resources.Diagnostics
+{
+    [AttributeUsage(AttributeTargets.All)]
+    [ExcludeFromCodeCoverage]
+    internal sealed class ExcludeFromCodeCoverageAttribute : Attribute
```
weshaggard: Do you worry that others will start depending on this ExcludeFromCodeCoverageAttribute that is being brought in from SR? If so does it make since to put it in its own file and just include it by default where SR.cs/vb is included?

weshaggard:
It would be nice if we didn't need to use a custom attribute for this is there any support for providing this information in a separate file?

src/Common/src/System/SR.cs
weshaggard: Sorry I missed your other comment about this already existing. In which case how do we prevent them from conflicting?

stephentoub: > In which case how do we prevent them from conflicting?

They're in different namespaces, and I made up this System.Resources.Diagnostics namespace, so no one else should be importing it.

stephentoub:
> It would be nice if we didn't need to use a custom attribute for this is there any support for providing this information in a separate file?

There are only two mechanisms at our disposal:
- Filter by type
- Filter by attribute name

If you prefer, we could just modify buildtools to filter out SR.  I wanted to avoid that though because there's value in having the resource string members show up in coverage, just not these helpers injected everywhere.

Instead of duplicating the attribute, we could also modify buildtools to just include the existing .cs file with the attribute.  I was avoiding that because I didn't want to add another .cs file dependency that magically gets included in a project, but we could go that route if you prefer.

Or we could just skip this entirely and live with the noise.

src/Common/src/System/SR.cs
weshaggard: So the code coverage tool actually only looks at the simple type name and not the fully qualified type name?

nguerrera: I don't feel very strongly about this, but maybe we should use a namespace that would be more obvious code smell if anyone but this file imported it. e.g. Internal.ExcludeSRFromCodeCoverge or something like that...

weshaggard:
Could we use an existing attribute like GeneratedCodeAttribute or EditorBrowsableAttribute or something of that nature. If not I don't mind having this it would just be nice if we didn't need to litter our code with yet more custom attributes.

src/Common/src/System/SR.vb
```diff
@@ -92,3 +103,23 @@ Namespace System
...
+        ' The code coverage tools are configured to ignore any code attributed with an attribute
+        ' named ExcludeFromCodeCoverage, regardless of its namespace.  We have it in a specialized namespace
+        ' so as to avoid conflicts with other ExcludeFromCodeCoverageAttribute types used elsewhere in corefx.
+        ' It's applied to the individual members in SR rather than to SR as a whole because we still
+        ' want code coverage to include the individual resource keys; doing so helps to highlight whether
+        ' we're exercising all error paths, whether any resource strings are stale and can be removed, etc.
+
+    End Class
+
+End Namespace
```
tarekgh: do we need to exclude only these methods or the whole SR class. I am just wondering if we need to set the attribute on the whole class instead. 

but in general the change looks good to me.

src/Common/src/System/SR.cs
stephentoub: > So the code coverage tool actually only looks at the simple type name and not the fully qualified type name?

We get to specify the filter, which we've done with a * for namespace.

nguerrera:
LGTM

src/Common/src/System/SR.vb
stephentoub: > do we need to exclude only these methods or the whole SR class. I am just wondering if we need to set the attribute on the whole class instead.

If we wanted to exclude the whole SR class, we could do so without code changes, by just modifying the command-line arguments passed to the code coverage tool to exclude the SR class.  But I don't think we should do that.  There's value in knowing which of the resource strings are being accessed, via seeing whether the associated generated properties are hit during testing: it helps to highlight stale strings that can be removed, and it helps to highlight error paths that are never being used.  Unfortunately all of these helpers and those resource strings are on the same class.

> Could we use an existing attribute like GeneratedCodeAttribute or EditorBrowsableAttribute or something of that nature.

GeneratedCodeAttribute is a good idea, but the compiler puts it more places than you might expect, e.g. the state machine type that's generated for async methods is annotated as GeneratedCode, which means that effectively all async methods would end up being excluded from code coverage.  EditorBrowsableAttribute could work, but just because something is advanced or meant to be internal and should be hidden from users in VS doesn't mean we don't want that code to be tested, e.g. https://github.com/dotnet/corefx/blob/41e203011152581a6c65bb81ac44ec037140c1bb/src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFileExtensions.cs#L9-L10 .  I'm not aware unfortunately of any other attributes that could be used without inadvertently suppressing code that should actually be tested.

> it would just be nice if we didn't need to litter our code with yet more custom attributes.

I may have found another solution.  I won't merge this until I've investigated that.  We might be able to just filter out this specific file from coverage with a change to build tools.

tarekgh: > There's value in knowing which of the resource strings are being accessed
I agree.

stephentoub:
Found a simpler fix that can be made in buildtools.

ViktorHofer:
@stephentoub you said you made a change in buildtools but it that change still enabled? When I produce a coverage report for one assembly it contains the SR file.

PR #2166, opened 2015-06-25
Add Asserts wrapper to assist Unix test debugging

stephentoub:
CoreCLR on Unix doesn't currently support PDBs.  This means that any failed xunit Asserts during testing include a stack trace but without any file or line information, which means that it can be time consuming to diagnose the location of a failure, in particular when there are multiple asserts of the same kind in the same test method, which is fairly common.

This commit adds an internal Assert class that can be included in a test project while debugging.  It provides wrappers over xunit's Assert class, and those wrappers use [Caller*] attributes to grab the call site's file/line info and include the information in any assertion failures.


stephentoub:
cc: @nguerrera, @ellismg, @sokket

ellismg:
LGTM

sokket:
LGTM

PR #2169, opened 2015-06-25
Fix serialization for type which has DataMember with special Name values

khdang:
This issue happens only for DCJS and affecting both NetCore and NetNative, due to missing method from recent merging of DCJS.

cc: @SGuyGe @shmao @zhenlan

### DIFF
--- a/src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/XmlObjectSerializerWriteContextComplexJson.cs
+++ b/src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/XmlObjectSerializerWriteContextComplexJson.cs
@@ -326,6 +326,12 @@ namespace System.Runtime.Serialization.Json
                 scopedKnownTypes.Pop();
             }
         }
+
+        internal static void WriteJsonNameWithMapping(XmlWriterDelegator xmlWriter, XmlDictionaryString[] memberNames, int index)
+        {
+            xmlWriter.WriteStartElement("a", JsonGlobals.itemString, JsonGlobals.itemString);
+            xmlWriter.WriteAttributeString(null, JsonGlobals.itemString, null, memberNames[index].Value);
+        }
 #endif
 
         internal static void VerifyObjectCompatibilityWithInterface(DataContract contract, object graph, Type declaredType)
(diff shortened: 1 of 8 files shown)


shmao:
LGTM.

stephentoub:
Should we add Debug.Asserts in all of the places that access such methods via reflection, e.g.
https://github.com/dotnet/corefx/blob/dfe8f72f7f4cbf61f08cf71b4e9f7c455c1e3dbf/src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/JsonFormatGeneratorStatics.cs#L404-L405
to make sure that this doesn't happen again (or isn't already happening elsewhere)?

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/Json/XmlObjectSerializerWriteContextComplexJson.cs
```diff
@@ -326,6 +326,12 @@ private void VerifyType(DataContract dataContract, Type declaredType)
                 scopedKnownTypes.Pop();
             }
         }
+
+        internal static void WriteJsonNameWithMapping(XmlWriterDelegator xmlWriter, XmlDictionaryString[] memberNames, int index)
```
weshaggard: Who calls this method?

weshaggard: Nevermind @stephentoub already pointed me to who calls it. I do agree with trying to add some validation where we can.

stephentoub: It's accessed via reflection (I had to search for it, as I had a similar question in mind); see my comment for a link to the access site.

khdang: I've added Debug.Assert to all applicable places that access to type members through reflection. I think there is validation check in NetNative toolchain but it shows as warning and not failure.

stephentoub: Thanks for adding those, looks good.

PR #2170, opened 2015-06-25
Fix DNU restore on Unix

### LINKED ISSUE #2157: Incorrect urls for dnu restore after PR2147, breaks building on Linux
https://github.com/dotnet/corefx/pull/2147 broke building on Linux:

````bash
$ ./build.sh 
  Restoring build tools...
  Restoring all packages...
/home/alexander/dev/corefx/build.proj(24,5): error MSB3073: The command ""/home/alexander/dev/corefx/packages/dnx-mono.1.0.0-beta5-12101/bin/dnu" restore --parallel --packages "/home/alexander/dev/corefx/packages" --source https:/www.myget.org/F/dotnet-core/ --source https:/www.myget.org/F/dotnet-coreclr/ --source https:/www.myget.org/F/dotnet-corefx/ --source https:/www.myget.org/F/dotnet-corefxtestdata/ --source https:/www.myget.org/F/dotnet-buildtools/ --source https:/www.nuget.org/api/v2/ "/home/alexander/dev/corefx/src" "/home/alexander/dev/corefx/packages/Microsoft.DotNet.BuildTools.1.0.25-prerelease-00056/lib/"" exited with code 1.
```

If you look closely, the URLs like `--source https:/www.myget.org/F/dotnet-core/` are wrong, they only have one slash after https. This might be an issue with xplat MSBuild incorrectly fixing up the slashes (just a guess, I haven't investigated yet).

JeremyKuhne:
MSBuild on Unix normalizes double forward slashes to a single slash.
Escape to avoid the issue with URLs.

Fixes #2157
@akoeplinger, @jhendrixMSFT, @weshaggard, @stephentoub

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -60,12 +60,13 @@
 
   <!-- list of nuget package sources passed to dnu -->
   <ItemGroup>
-    <DnuSourceList Include="https://www.myget.org/F/dotnet-core/" />
-    <DnuSourceList Include="https://www.myget.org/F/dotnet-coreclr/" />
-    <DnuSourceList Include="https://www.myget.org/F/dotnet-corefx/" />
-    <DnuSourceList Include="https://www.myget.org/F/dotnet-corefxtestdata/" />
-    <DnuSourceList Include="https://www.myget.org/F/dotnet-buildtools/" />
-    <DnuSourceList Include="https://www.nuget.org/api/v2/" />
+    <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
+    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-core/" />
+    <DnuSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-coreclr/" />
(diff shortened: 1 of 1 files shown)


stephentoub:
Is this a bug in MSBuild, and if so, is there an issue/bug # we can include?

Otherwise, LGTM.

JeremyKuhne:
@stephentoub I'm not sure if it is a "bug" or not. I was going to play around with it a bit more before opening an issue on them.

weshaggard:
Looks like a reasonable fix but I think it will break immediately once it comes internally because we have some UNC network paths in that list which will not like being prepended with https.

JeremyKuhne:
@weshaggard Oh, bummer- I'll have to try something more complicated :(

JeremyKuhne:
@weshaggard Thankfully escaping on the include worked as well. Updated.

akoeplinger:
@JeremyKuhne Thanks for fixing it :+1:

I'm curious too if this is an MSBuild bug or not. If not, we should probably start a list of "xplat MSBuild gotchas" where things like this are documentated.

PR #2175, opened 2015-06-26
Fix broken coding style link in contributing.md

kevin-david:
Relative URL seemed preferable to absolute... although if it's moved again it'll break anyway :)


stephentoub:
LGTM.  Thanks for the fix.

PR #2176, opened 2015-06-26
Fix System.Linq.Expressions.Tests build warnings and Console.WriteLine output

stephentoub:
With recently ported tests, there are now a handful of warnings being generated during the build along with some unnecessary console output.

### DIFF
--- a/src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
+++ b/src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


stephentoub:
cc: @VSadov

src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
```diff
@@ -734,7 +735,7 @@ public static void ElementInit()
             {
                 ElementInit ei1 = Expression.ElementInit(mi1, new Expression[] { ce1 });
             }
-            catch (ArgumentException e)
+            catch (ArgumentException)
```
Maxwe11: Assert.Throws ?

VSadov:
LGTM

src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
stephentoub: I'm not familiar with what exactly this is testing, e.g. without examining it more I don't know if this is expecting an ArgumentException or just wants to allow it.  An Assert.Throws would likely make sense, but I'll leave that for @VSadov to look into later... my goal now was just to clean up the build and test output.

PR #2178, opened 2015-06-26
Add CompareInfo tests

ellismg:
An initial set of tests, based on the coverage we have internally in our
conformance tests, but rewritten in idomatic xunit style.

In addition, I have added some more tests to bring up the code coverage
numbers of the Windows implementation of globalization.

### DIFF
--- a/src/System.Globalization/tests/CompareInfo.cs
+++ b/src/System.Globalization/tests/CompareInfo.cs
@@ -1,0 +1,424 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+using Xunit;
+
+public partial class CompareInfoTests
+{
+    [Theory]
+    [InlineData("")]
+    [InlineData("en")]
+    [InlineData("en-US")]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public static void GetCompareInfo(string localeName)
+    {
+        CompareInfo ci = CompareInfo.GetCompareInfo(localeName);
+
+        Assert.Equal(localeName, ci.Name);
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


src/System.Globalization/tests/CompareInfo.cs
```diff
@@ -0,0 +1,424 @@
...
+        }
+    }
+
+    [Fact]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public static void IndexOfMinusOneEverettCompatability()
+    {
+        CompareInfo ci  = CultureInfo.InvariantCulture.CompareInfo;
+
+        // This behavior was for Everett compatability.  We early outed for empty source strings
```
stephentoub: Could we change the comment to list a known version number rather than a codename?

src/System.Globalization/tests/System.Globalization.Tests.csproj
```diff
@@ -16,6 +16,8 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="CompareInfo.cs" />
+    <Compile Include="TestData.cs" />
```
stephentoub: Should this be CompareInfo.TestData.cs, or something like that?

stephentoub:
LGTM

src/System.Globalization/tests/System.Globalization.Tests.csproj
ellismg: That's a good idea.  I'll change it.

src/System.Globalization/tests/CompareInfo.cs
ellismg: Sure, I will change this to ".NET Framework 1.1"

PR #2179, opened 2015-06-26
Switch System.Runtime.Extensions to use BitConverter.cs in corefx

stephentoub:
The System.Runtime.Extensions.dll partial facade is currently forwarding to the System.BitConverter class in mscorlib.  This PR changes it to use the BitConverter.cs in corefx.  To help verify nothing broke in doing so, I increased the test coverage for BitConverter.  I also did a bit of code cleanup in the BitConverter.cs file, e.g. using Debug.Assert instead of Contract.Assert, using "string" instead of "String", eliminating some unnecessary branches, ensuring we're including a parameter name in ArgumentExceptions, consolidating how we throw exceptions (matching what's done in mscorlib), etc.

### DIFF
--- a/src/System.Runtime.Extensions/src/Resources/Strings.resx
+++ b/src/System.Runtime.Extensions/src/Resources/Strings.resx
@@ -136,7 +136,7 @@
     <value>The parameter '{0}' cannot be an empty string.</value>
   </data>
   <data name="Arg_ArrayPlusOffTooSmall" xml:space="preserve">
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @mellinoe

ellismg:
db test this please

mellinoe:
LGTM, there shouldn't be any need to use the one in mscorlib for this.

stephentoub:
Thanks, guys.

PR #2181, opened 2015-06-26
Finished incorporating API review feedback for System.Text.Encodings.Web

KrzysztofCwalina:
With the exception of feedback/workitems that are are being debated or researched:
1. Research ability to encode UTF8 text buffers
2. Come up with a better namespace name

After this is checked in, we will start moving ASP.NET to these APIs

### DIFF
--- a/src/System.Text.Encodings.Web/src/System.Text.Encodings.Web.csproj
+++ b/src/System.Text.Encodings.Web/src/System.Text.Encodings.Web.csproj
@@ -18,11 +18,11 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Text\Encodings\Web\AllowedCharactersBitmap.cs" />
-    <Compile Include="System\Text\Encodings\Web\CodePointFilter.cs" />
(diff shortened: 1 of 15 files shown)


KrzysztofCwalina:
cc: @Eilon, @GrabYourPitchforks, @terrajobst

jasonwilliams200OK:
ASPE.NET => American Society for Precision Engineering
For a moment, I thought that has something to do with ASP.NET :smile:

Eilon:
:shipit:

ellismg:
What do you think about putting TextEncoder and TextEncoderSettings in the System.Text.Encodings namespace (and keeping the concrete encoders in System.Text.Encodings.Web?)

KrzysztofCwalina:
@jasonwilliams200OK, good catch :-)

KrzysztofCwalina:
@ellismg, I would prefer to come up with a namespace name that can contain all these types and not be web specific. The probability of these abstractions actually being used for non-web encoders is not super high, the number of types is low, and so probably not worth having them distributed across namespaces.

Issue #1307, opened 2015-04-01
Improve performance of BigInteger


axelheer:
Looking at other implementations like...

* [IntXLib][0] by @devoyster (faster for really huge numbers)
* [NeinMath][1] by @axelheer (faster for numbers with a few thousand bits)
* [GNU MP][2] (not .NET based, but super fast, if not the fastest at all)

...there should be room for one or another performance tweak.

Using managed code is by it's nature not as fast as working bare to the metal, but if the algorithms can be improved significantly, that would be great. Or other optimizations (pointer arithmetic, optimised assembly code) are viable as well for this big number stuff, maybe down at *CoreCLR* level (don't know if *unsafe* is allowed here).

What do you think?

[0]: https://github.com/devoyster/IntXLib
[1]: https://github.com/axelheer/nein-math
[2]: https://gmplib.org/

terrajobst:
Do you have a concrete scenario for which the performance is substantially worse?

axelheer:
Any scenario involving big numbers with thousand bits or more.

The current multiplication implementation relies on "grammer-school" methods (as called by literature), which is good for "small" big numbers, but becomes very inefficient for "not so small" big numbers since it operates at *Θ(n^2)*. Usually a switch to [Karatsuba's algorithm][0] (*Θ(n^1.585)*) at some threshold helps, after that a switch to [Toom-3][1] (*Θ(n^1.465)*), then a switch to another faster algorithm with more overhead albeit less complexity... and I'm only talking about multiplication.

The *IntXLib* seems to use some kind of FFT based algorithm operating at *Θ(n log n log log n)*. Maybe it's the [Schönhage–Strassen algorithm][2], which is used for numbers with at least 33,000 decimal digits by the GNU MP library. The [GNU MP][3] library seems to be the fastest implementation available, it's also used by commercial software like *Mathematica* or *Maple*. Thus, it may be possible to just integrate that in a open source .NET framework. :)

On the other hand, my implementation contains many minor tweaks to get this calculations done faster, since I was interested in only a few thousand bits. It also contains the first step to optimize multiplication, and an optimized algorithm for squaring. You can find a simple performance comparison [there][4]. But note, like *IntXLib* it's just a small fun implementation -- if you want to see something serious, you'd have to look at GNU MP... (if you prefer "pure managed code" I could contribute mine, of course)

[0]: https://en.wikipedia.org/wiki/Karatsuba_algorithm
[...]

PR #1618, opened 2015-05-03
Improve performance of BigInteger.Add/Sub/Div/Rem

axelheer:
To introduce further performance tweaks, some minor improvements are added to `BigIntegerCalculator`, which cover all the basic operations. Having these algorithms based on raw uint[] objects will lead to more interesting tweaks of the more fancy stuff (`ModPow`).

A basic performance comparison based on [this code](https://gist.github.com/axelheer/3b701c91271b7c762586) unveils the following results:

**Add**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |        40 |       37 |
|        64 | 1,000,000 |        97 |       94 |
|       256 | 1,000,000 |       110 |      107 |
|     1,024 | 1,000,000 |       163 |      152 |
|     4,096 |   100,000 |        40 |       37 |
|    16,384 |   100,000 |       147 |      122 |
|    65,536 |   100,000 |       515 |      440 |

**Subtract**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |        44 |       38 |
|        64 | 1,000,000 |       106 |       95 |
|       256 | 1,000,000 |       128 |      111 |
|     1,024 | 1,000,000 |       182 |      164 |
|     4,096 |   100,000 |        42 |       38 |
|    16,384 |   100,000 |       143 |      131 |
|    65,536 |   100,000 |       511 |      444 |

**Divide**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |        49 |       29 |
|        64 | 1,000,000 |       125 |      110 |
|       256 | 1,000,000 |       349 |      287 |
|     1,024 |   100,000 |       215 |      119 |
|     4,096 |    10,000 |       254 |      124 |
|    16,384 |     1,000 |       378 |      172 |
|    65,536 |       100 |       595 |      263 |

*Note:* during making this pull request I noted that all the `Debug.Assert` aren't doing anything... (?)

*Attention:* because of using `stackalloc` I had to reduce an unit test value, which is currently very high and leads to a stack overflow. Would it be better to switch to `fixed` or should the actual stack size be increased, if someone really needs that huge numbers?

#1307

PR #2182, opened 2015-06-27
Improve performance of BigInteger.Pow/ModPow

axelheer:
To introduce further performance tweaks, the exponentiation algorithms are ported to `BigIntegerCalculator`. Furthermore the newly introduced `FastReducer` triggers a bad corner case within the division algorithm, which gets fixed too.

A basic performance comparison based on [this code](https://gist.github.com/axelheer/3b701c91271b7c762586) unveils the following results:

**ModPow**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 |   100,000 |        64 |       23 |
|        64 |    10,000 |       145 |      202 |
|       256 |     1,000 |       298 |      181 |
|     1,024 |       100 |     1,344 |      637 |
|     4,096 |        10 |     7,778 |    2,548 |
|    16,384 |         1 |    48,292 |    9,595 |

#1307

### DIFF
--- a/src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
+++ b/src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
@@ -19,7 +19,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Numerics\BigIntegerCalculator.AddSub.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.BitsBuffer.cs" />
     <Compile Include="System\Numerics\BigIntegerCalculator.DivRem.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.FastReducer.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.PowMod.cs" />
     <Compile Include="System\Numerics\BigIntegerCalculator.SquMul.cs" />
     <Compile Include="System\Numerics\BigInteger.cs" />
     <Compile Include="System\Numerics\BigIntegerBuilder.cs" />
--- a/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
@@ -1049,45 +1049,6 @@ namespace System.Numerics
             return right;
         }
 
-
-        private static void ModPowUpdateResult(ref BigIntegerBuilder regRes, ref BigIntegerBuilder regVal, ref BigIntegerBuilder regMod, ref BigIntegerBuilder regTmp)
-        {
-            NumericsHelpers.Swap(ref regRes, ref regTmp);
-            regRes.Mul(ref regTmp, ref regVal);   // result = result * value;
-            regRes.Mod(ref regMod);               // result = result % modulus;                       
-        }
-
-        private static void ModPowSquareModValue(ref BigIntegerBuilder regVal, ref BigIntegerBuilder regMod, ref BigIntegerBuilder regTmp)
-        {
-            NumericsHelpers.Swap(ref regVal, ref regTmp);
-            regVal.Mul(ref regTmp, ref regTmp);   // value = value * value;
-            regVal.Mod(ref regMod);               // value = value % modulus;
-        }
-
-        private static void ModPowInner(uint exp, ref BigIntegerBuilder regRes, ref BigIntegerBuilder regVal, ref BigIntegerBuilder regMod, ref BigIntegerBuilder regTmp)
-        {
-            while (exp != 0)          // !(Exponent.IsZero)
-            {
-                if ((exp & 1) == 1)   // !(Exponent.IsEven)
-                    ModPowUpdateResult(ref regRes, ref regVal, ref regMod, ref regTmp);
-                if (exp == 1)         // Exponent.IsOne - we can exit early
-                    break;
-                ModPowSquareModValue(ref regVal, ref regMod, ref regTmp);
-                exp >>= 1;
-            }
-        }
-
-        private static void ModPowInner32(uint exp, ref BigIntegerBuilder regRes, ref BigIntegerBuilder regVal, ref BigIntegerBuilder regMod, ref BigIntegerBuilder regTmp)
-        {
-            for (int i = 0; i < 32; i++)
-            {
-                if ((exp & 1) == 1)   // !(Exponent.IsEven)
-                    ModPowUpdateResult(ref regRes, ref regVal, ref regMod, ref regTmp);
-                ModPowSquareModValue(ref regVal, ref regMod, ref regTmp);
-                exp >>= 1;
-            }
-        }
-
         public static BigInteger ModPow(BigInteger value, BigInteger exponent, BigInteger modulus)
         {
             if (exponent.Sign < 0)
@@ -1098,36 +1059,31 @@ namespace System.Numerics
             exponent.AssertValid();
             modulus.AssertValid();
 
-            int signRes = +1;
-            int signVal = +1;
-            int signMod = +1;
-            bool expIsEven = exponent.IsEven;
-            BigIntegerBuilder regRes = new BigIntegerBuilder(BigInteger.One, ref signRes);
-            BigIntegerBuilder regVal = new BigIntegerBuilder(value, ref signVal);
-            BigIntegerBuilder regMod = new BigIntegerBuilder(modulus, ref signMod);
-            BigIntegerBuilder regTmp = new BigIntegerBuilder(regVal.Size);
+            bool trivialValue = value._bits == null;
+            bool trivialExponent = exponent._bits == null;
+            bool trivialModulus = modulus._bits == null;
 
-            regRes.Mod(ref regMod);   // Handle special case of exponent=0, modulus=1
+            if (trivialModulus)
+            {
+                long bits = trivialValue && trivialExponent ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent._sign), NumericsHelpers.Abs(modulus._sign)) :
+                            trivialValue ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, NumericsHelpers.Abs(modulus._sign)) :
+                            trivialExponent ? BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent._sign), NumericsHelpers.Abs(modulus._sign)) :
+                            BigIntegerCalculator.Pow(value._bits, exponent._bits, NumericsHelpers.Abs(modulus._sign));
 
-            if (exponent._bits == null)
-            {   // exponent fits into an Int32
-                ModPowInner((uint)exponent._sign, ref regRes, ref regVal, ref regMod, ref regTmp);
+                return value._sign < 0 && !exponent.IsEven ? -1 * bits : bits;
             }
             else
-            {   // very large exponent
-                int len = Length(exponent._bits);
-                for (int i = 0; i < len - 1; i++)
-                {
-                    uint exp = exponent._bits[i];
-                    ModPowInner32(exp, ref regRes, ref regVal, ref regMod, ref regTmp);
-                }
-                ModPowInner(exponent._bits[len - 1], ref regRes, ref regVal, ref regMod, ref regTmp);
-            }
+            {
+                uint[] bits = trivialValue && trivialExponent ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent._sign), modulus._bits) :
+                              trivialValue ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, modulus._bits) :
+                              trivialExponent ? BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent._sign), modulus._bits) :
+                              BigIntegerCalculator.Pow(value._bits, exponent._bits, modulus._bits);
 
-            return regRes.GetInteger(value._sign > 0 ? +1 : expIsEven ? +1 : -1);
+                return new BigInteger(bits, value._sign < 0 && !exponent.IsEven);
+            }
         }
 
-        public static BigInteger Pow(BigInteger value, Int32 exponent)
+        public static BigInteger Pow(BigInteger value, int exponent)
         {
             if (exponent < 0)
                 throw new ArgumentOutOfRangeException("exponent", SR.ArgumentOutOfRange_MustBeNonNeg);
@@ -1136,75 +1092,27 @@ namespace System.Numerics
             value.AssertValid();
 
             if (exponent == 0)
-                return BigInteger.One;
+                return s_bnOneInt;
             if (exponent == 1)
                 return value;
-            if (value._bits == null)
+
+            bool trivialValue = value._bits == null;
+
+            if (trivialValue)
             {
                 if (value._sign == 1)
                     return value;
                 if (value._sign == -1)
-                    return (exponent & 1) != 0 ? value : 1;
+                    return (exponent & 1) != 0 ? value : s_bnOneInt;
                 if (value._sign == 0)
                     return value;
             }
 
-            int sign = +1;
-            BigIntegerBuilder regSquare = new BigIntegerBuilder(value, ref sign);
+            uint[] bits = trivialValue
+                        ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent))
+                        : BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent));
 
-            // Get an estimate of the size needed for regSquare and regRes, so we can minimize allocations.
-            int cuSquareMin = regSquare.Size;
-            int cuSquareMax = cuSquareMin;
-            uint uSquareMin = regSquare.High;
-            uint uSquareMax = uSquareMin + 1;
-            if (uSquareMax == 0)
-            {
-                cuSquareMax++;
-                uSquareMax = 1;
-            }
-            int cuResMin = 1;
-            int cuResMax = 1;
-            uint uResMin = 1;
-            uint uResMax = 1;
-
-            for (int expTmp = exponent; ;)
-            {
-                if ((expTmp & 1) != 0)
-                {
-                    MulUpper(ref uResMax, ref cuResMax, uSquareMax, cuSquareMax);
-                    MulLower(ref uResMin, ref cuResMin, uSquareMin, cuSquareMin);
-                }
-
-                if ((expTmp >>= 1) == 0)
-                    break;
-
-                MulUpper(ref uSquareMax, ref cuSquareMax, uSquareMax, cuSquareMax);
-                MulLower(ref uSquareMin, ref cuSquareMin, uSquareMin, cuSquareMin);
-            }
-
-            if (cuResMax > 1)
-                regSquare.EnsureWritable(cuResMax, 0);
-            BigIntegerBuilder regTmp = new BigIntegerBuilder(cuResMax);
-            BigIntegerBuilder regRes = new BigIntegerBuilder(cuResMax);
-            regRes.Set(1);
-
-            if ((exponent & 1) == 0)
-                sign = +1;
-
-            for (int expTmp = exponent; ;)
-            {
-                if ((expTmp & 1) != 0)
-                {
-                    NumericsHelpers.Swap(ref regRes, ref regTmp);
-                    regRes.Mul(ref regSquare, ref regTmp);
-                }
-                if ((expTmp >>= 1) == 0)
-                    break;
-                NumericsHelpers.Swap(ref regSquare, ref regTmp);
-                regSquare.Mul(ref regTmp, ref regTmp);
-            }
-
-            return regRes.GetInteger(sign);
+            return new BigInteger(bits, value._sign < 0 && (exponent & 1) != 0);
         }
 
         #endregion public static methods
@@ -1810,8 +1718,8 @@ namespace System.Numerics
 
             if (trivialDivisor)
             {
-                uint[] bits = BigIntegerCalculator.Remainder(dividend._bits, NumericsHelpers.Abs(divisor._sign));
-                return new BigInteger(bits, dividend._sign < 0);
+                long bits = BigIntegerCalculator.Remainder(dividend._bits, NumericsHelpers.Abs(divisor._sign));
+                return dividend._sign < 0 ? -1 * bits : bits;
             }
 
             if (dividend._bits.Length < divisor._bits.Length)
@@ -2032,11 +1940,10 @@ namespace System.Numerics
         [Pure]
         internal static int Length(uint[] rgu)
         {
-            int cu = rgu.Length;
-            if (rgu[cu - 1] != 0)
-                return cu;
-            Debug.Assert(cu >= 2 && rgu[cu - 2] != 0);
-            return cu - 1;
+            Debug.Assert(rgu[rgu.Length - 1] != 0);
+
+            // no leading zeros
+            return rgu.Length;
         }
 
         internal int _Sign { get { return _sign; } }
@@ -2086,47 +1993,6 @@ namespace System.Numerics
             }
             xl = (x._bits == null ? 1 : x._bits.Length);
             return x._sign < 0;
-        }
-
-        // Gets an upper bound on the product of uHiRes * (2^32)^(cuRes-1) and uHiMul * (2^32)^(cuMul-1).
-        // The result is put int uHiRes and cuRes.
-        private static void MulUpper(ref uint uHiRes, ref int cuRes, uint uHiMul, int cuMul)
-        {
-            ulong uu = (ulong)uHiRes * uHiMul;
-            uint uHi = NumericsHelpers.GetHi(uu);
-            if (uHi != 0)
-            {
-                if (NumericsHelpers.GetLo(uu) != 0 && ++uHi == 0)
-                {
-                    uHi = 1;
-                    cuRes++;
-                }
-                uHiRes = uHi;
-                cuRes += cuMul;
-            }
-            else
-            {
-                uHiRes = NumericsHelpers.GetLo(uu);
-                cuRes += cuMul - 1;
-            }
-        }
-
-        // Gets a lower bound on the product of uHiRes * (2^32)^(cuRes-1) and uHiMul * (2^32)^(cuMul-1).
-        // The result is put int uHiRes and cuRes.
-        private static void MulLower(ref uint uHiRes, ref int cuRes, uint uHiMul, int cuMul)
-        {
-            ulong uu = (ulong)uHiRes * uHiMul;
-            uint uHi = NumericsHelpers.GetHi(uu);
-            if (uHi != 0)
-            {
-                uHiRes = uHi;
-                cuRes += cuMul;
-            }
-            else
-            {
-                uHiRes = NumericsHelpers.GetLo(uu);
-                cuRes += cuMul - 1;
-            }
         }
 
         internal static int GetDiffLength(uint[] rgu1, uint[] rgu2, int cu)
--- a/src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
@@ -49,7 +49,7 @@ namespace System.Numerics
             return quotient;
         }
 
-        public static uint[] Remainder(uint[] left, uint right)
+        public static uint Remainder(uint[] left, uint right)
         {
             Debug.Assert(left != null);
             Debug.Assert(left.Length >= 1);
@@ -63,7 +63,7 @@ namespace System.Numerics
                 carry = value % right;
             }
 
-            return new uint[] { (uint)carry };
+            return (uint)carry;
         }
 
         [SecuritySafeCritical]
@@ -79,18 +79,21 @@ namespace System.Numerics
             // Switching to unsafe pointers helps sparing
             // some nasty index calculations...
 
-            uint[] quotient = new uint[left.Length - right.Length + 1];
-            remainder = new uint[right.Length];
+            // NOTE: left will get overwritten, we need a local copy
 
-            fixed(uint* l = left, r = right, q = quotient, e = remainder)
+            uint[] localLeft = CreateCopy(left);
+            uint[] bits = new uint[left.Length - right.Length + 1];
+
+            fixed (uint* l = localLeft, r = right, b = bits)
             {
-                Divide(l, left.Length,
+                Divide(l, localLeft.Length,
                        r, right.Length,
-                       q, quotient.Length,
-                       e, remainder.Length);
+                       b, bits.Length);
             }
 
-            return quotient;
+            remainder = localLeft;
+
+            return bits;
         }
 
         [SecuritySafeCritical]
@@ -104,17 +107,19 @@ namespace System.Numerics
 
             // Same as above, but only returning the quotient.
 
-            uint[] quotient = new uint[left.Length - right.Length + 1];
+            // NOTE: left will get overwritten, we need a local copy
 
-            fixed (uint* l = left, r = right, q = quotient)
+            uint[] localLeft = CreateCopy(left);
+            uint[] bits = new uint[left.Length - right.Length + 1];
+
+            fixed (uint* l = localLeft, r = right, b = bits)
             {
-                Divide(l, left.Length,
+                Divide(l, localLeft.Length,
                        r, right.Length,
-                       q, quotient.Length,
-                       null, 0);
+                       b, bits.Length);
             }
 
-            return quotient;
+            return bits;
         }
 
         [SecuritySafeCritical]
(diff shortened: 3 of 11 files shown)


stephentoub:
Thanks, @axelheer. Will take a look, but in the meantime, do you know why the 64/10K case suffers a 33% regression?

axelheer:
@stephentoub nope, that's a bit odd and needs further investigation. Maybe the Barrett reduction backfires for smaller numbers.

src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
```diff
@@ -36,6 +36,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.PowMod.cs" />
```
stephentoub: Shouldn't this be in the ItemGroup above with the rest of the BigIntegerCalculator.* files?

src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
```diff
@@ -1098,36 +1059,30 @@ public static BigInteger ModPow(BigInteger value, BigInteger exponent, BigIntege
...
+                        ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, NumericsHelpers.Abs(modulus._sign))
+                        : trivialValue && trivialExponent
+                        ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent._sign), modulus._bits)
+                        : trivialValue
+                        ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, modulus._bits)
+                        : trivialExponent
+                        ? BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent._sign), modulus._bits)
+                        : trivialModulus
+                        ? BigIntegerCalculator.Pow(value._bits, exponent._bits, NumericsHelpers.Abs(modulus._sign))
+                        : BigIntegerCalculator.Pow(value._bits, exponent._bits, modulus._bits);
```
stephentoub: I personally find this formatting of ```? :``` hard to read, and prefer to see it written:
```C#
uint[] bits =
    condition ? result :
    condition ? result :
    condition ? result :
    ...
    condition ? result :
    result;
```

stephentoub: That said, I wonder if in this particular case it would be both slightly better performing and easier to read if it were instead broken up into a few pieces like this (you'll want to double-check that I haven't messed up the logic here in copying-and-pasting):
```C#
uint[] bits;
if (trivialValue)
{
    bits = 
        trivialExponent & trivialModulus ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent._sign), NumericsHelpers.Abs(modulus._sign)) :
        trivialModulus ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, NumericsHelpers.Abs(modulus._sign)) :
        trivialExponent ? BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), NumericsHelpers.Abs(exponent._sign), modulus._bits) :
        BigIntegerCalculator.Pow(NumericsHelpers.Abs(value._sign), exponent._bits, modulus._bits);
}
else if (trivialExponent)
{
    bits = trivialModulus ?
        BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent._sign), NumericsHelpers.Abs(modulus._sign)) :
        BigIntegerCalculator.Pow(value._bits, NumericsHelpers.Abs(exponent._sign), modulus._bits);
}
else
{
    bits = trivialModulus ?
        BigIntegerCalculator.Pow(value._bits, exponent._bits, NumericsHelpers.Abs(modulus._sign)) :
        BigIntegerCalculator.Pow(value._bits, exponent._bits, modulus._bits);
}
```

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
```diff
@@ -0,0 +1,536 @@
...
+        public static uint[] Pow(uint value, uint power, uint[] modulus)
+        {
+            Debug.Assert(modulus != null);
+
+            // The big modulus pow method for a 32-bit integer
+            // raised by a 32-bit integer...
+
+            BitsBuffer temp = new BitsBuffer(modulus.Length * 2, value);
+            BitsBuffer result = new BitsBuffer(modulus.Length * 2, 1);
+            FastReducer reducer = new FastReducer(modulus);
```
stephentoub: I know you've been making all of these changes in the name of performance and been measuring along the way.  That's great.  I'm still concerned with what appears to be a bunch of new allocations, though; BitBuffer and FastReducer are both classes being allocated here, and each of those instances ends up allocating multiple arrays.  As I've mentioned before, the impact of such allocations often doesn't show up when just doing single-threaded performance benchmarking.  What happens if, for example, this functionality is used under stress in a server environment?  Are we going to see performance regressions as a result?

src/System.Runtime.Numerics/tests/BigInteger/modpow.cs
```diff
@@ -161,6 +161,23 @@ public static void ModPow2Large1SmallInt()
         }
 
         [Fact]
+        public static void ModPow3LargeInt()
+        {
+            byte[] tempByteArray1 = new byte[0];
+            byte[] tempByteArray2 = new byte[0];
+            byte[] tempByteArray3 = new byte[0];
```
stephentoub: Why are these being initialized here?  Even in a test, seems like the allocations are unnecessary.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
mellinoe: If they are just helper objects for encapsulating a few things internally, can we just make them structs? The internal array allocation still wouldn't be helped, though.

src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
axelheer: Sure. Will fix that.

src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
axelheer: Sure. I'll update this.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
axelheer: Actually, I invested a bit more in reducing allocations, since I've learned from our previous work that these are damn crucial (my original implementation just used the big-integer structure, thus relocated with every operation...). But let's have a look at the numbers:

- The current implementation needs 4 "builders" (`regRes`, `regVal`, `regMod`, `regTmp`), the algorithm needs its values. I don't know any way how to use less than those 2 buffers.
- One "buffer" allocates 2 arrays, one for the current value and one for the next, and switches between them. So the 2 is a constant value, no matter how many calculations are needed (the bigger the numbers, the more calculations). For 32-bit integers we are even buffer-free.
- The "reducer" creates some overhead though, but the Barrett reduction needs it's μ, q and ˜q. As with the buffer the 3 is at least constant.

So, how can we squeeze this further?

- As suggested, turning buffer and reducer into structures shouldn't be any problem. I've no idea why I've chosen classes anyway.
- As already suspected at the beginning, the Barrett reduction may be not that efficient for smaller numbers (I'll investigate this further!). If that's the case, we can define a threshold for using an ordinary remainder, so we have this overhead only for the really big ones.

Sounds ok?

src/System.Runtime.Numerics/tests/BigInteger/modpow.cs
axelheer: Copy and paste from above, nevertheless you're totally right. Will fix this.

axelheer:
@stephentoub is it of equal value to just "amend" further changes? Or has the push further commits / squash in the end process any pros?

stephentoub:
In this case I'd suggesting adding additional commits and then squashing later.  Personally, I prefer to just ammend the existing commit when the changes are small / trivial and I don't expect anyone to need to re-review, e.g. fixing up typos or making simple one-line fixes.  For larger things, I prefer to have the additional commits so that it's clear to a reviewer what's changed.  That's just me, though.

src/System.Runtime.Numerics/tests/BigInteger/modpow.cs
stephentoub: Thanks; I didn't look at the rest of the file and thus didn't realize this was copied from elsewhere... please feel free to fix those up as well ;)

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
stephentoub: Sounds good, thanks.  It'd be good to also try running some perf tests in parallel, both with and without the server GC enabled, to see how things are affected.

axelheer: How do I enable the server GC with the new .NET Core? :flushed:

jasonwilliams200OK: > How do I enable the server GC with the new .NET Core?

For instance, your compiled exe name is `benchmark-bigint.exe`, create a `benchmark-bingint.exe.config` with following contents:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <runtime>
    <gcServer enabled="true" />
  </runtime>
</configuration>
```

In your code, you can verify it as:

```c#
using System.Diagnostics;

{
    Debug.Assert(System.Runtime.GCSettings.IsServerGC);
    // ...
}
```

axelheer:
@stephentoub @mellinoe the two internal helpers are structures now, the reducer seems to pay off already at 256 bits -- below that an ordinary remainder is used. But I still measure that regression at 64 bits, a bit smaller but still there. I have no idea what's wrong here...

@jasonwilliams200OK a DNX console app doesn't produce an .exe. And a classic .NET 4.6 console application does not work since there are type forwards from `System.Runtime.Numberics` to `System.Numerics`. Thus, just putting the modified DLL to the build / run directory as described [here](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/oss-signing.md) doesn't work.

jasonwilliams200OK:
Ah, I thought you are using coreclr: `CoreRun.exe yours.exe`.

axelheer:
@dotnet-bot test this please

axelheer:
*Update:* it seems that the current `mod` implementation is faster at 64 bits, and since this is the most used operation here (if we don't use the "reducer") we have a regression. (And the "reducer" isn't that fast at 64 bits either...) But the `mod` operation has already been replaced with #1618, because I only compared performance data of `div` there. (I assumed no difference, because the algorithms are the same, but `BigIntegerBuilder` is better within this special case...) Yeah, that sucks.

Thus, I don't think right here is much more to optimize. But, I should review integer division for 64 bit numbers (for 64 bit divisors?) and prepare a separate PR when I figured out a solution.

@stephentoub  What do you think?

stephentoub:
> But the mod operation has already been replaced with #1618, because I only compared performance data of div there

Just so I understand, you're saying we already took a regression in the previous change, and that's what you'll be looking to repair in a subsequent one? Can you quantify that regression?

axelheer:
> Just so I understand, you're saying we already took a regression in the previous change, and that's what you'll be looking to repair in a subsequent one?

Yeah. If that's a problem we can rollback this stuff and wait until I figured out why it gets stuck at 64 bits.

> Can you quantify that regression?

It's the same situation for `mod` as here for `modpow`, only a regression at 64 bits, but about 80%.

axelheer:
I finally rewrote the integer division, I just combined the algorithms of `BigIntegerBuilder` and `BigIntegerCalculator`. *Bonus:* no more allocations within integer division.

The performance regression seems to be fixed now:

**ModPow**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 |   100,000 |        64 |       23 |
|        64 |    10,000 |       145 |      131 |
|       256 |     1,000 |       298 |      241 |
|     1,024 |       100 |     1,344 |      634 |
|     4,096 |        10 |     7,778 |    2,542 |
|    16,384 |         1 |    48,292 |    9,573 |

I'm still not happy with the division code, maybe I tweak it a bit tomorrow.

axelheer:
@stephentoub since the last commit the temporary memory of buffers can be shared between them, which means n / 2 - 1 less allocations (okay, just 3 instead of 4 in our case). I'm quite sure doing it with less allocations isn't possible.

Agreed?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,123 @@
...
+namespace System.Numerics
+{
+    internal static partial class BigIntegerCalculator
+    {
+        // To spare memory allocations a buffer helps reusing memory!
+        // We just create the target array twice and switch between every
+        // operation. In order to not compute unnecessarily with all those
+        // leading zeros we take care of the current actual length.
+
+        // ATTENTION: always pass this by reference,
```
stephentoub: And in particular it's a mutable struct.  Could you note that and move this warning to above the class?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            public BitsBuffer(int length, uint[] value)
+            {
+                Debug.Assert(value != null);
+                Debug.Assert(length >= ActualLength(value));
+
+                _bits = new uint[length];
+                _length = ActualLength(value);
+
+                Array.Copy(value, _bits, _length);
```
stephentoub: ```Array.Copy(value, 0, _bits, 0, _length);``` helps to save having to query for the lower bounds of the arrays, which can be non-zero with Array.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,123 @@
...
+                Apply(ref temp, _length + _length);
+            }
+
+            public void Reduce(ref FastReducer reducer)
+            {
+                _length = reducer.Reduce(_bits, _length);
+            }
+
+            [SecuritySafeCritical]
+            public unsafe void Reduce(uint[] modulus)
```
stephentoub: Could you add some comments to explain what some of these functions are doing?  Some like MultipleSelf and SquareSelf are relatively self-explanatory, but this one is less clear, to me at least.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,123 @@
...
+                        Divide(b, _length,
+                               m, modulus.Length,
+                               null, 0);
+                    }
+
+                    _length = ActualLength(_bits, modulus.Length);
+                }
+            }
+
+            public uint[] ToArray()
```
stephentoub: This is all internal so it's not as big a deal, but could you rename this?  Typically ToArray has connotations of creating a new copy, whereas this is just returning the underlying storage that could still be in use.  Maybe GetBuffer like MemoryStream uses?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,123 @@
...
+                return _bits;
+            }
+
+            private void Apply(ref BitsBuffer temp, int maxLength)
+            {
+                Array.Clear(_bits, 0, _length);
+
+                uint[] t = temp._bits;
+                temp._bits = _bits;
+                _bits = t;
```
stephentoub: temp's _length doesn't need to be updated / reset to 0?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -79,18 +79,19 @@ public static uint[] Remainder(uint[] left, uint right)
             // Switching to unsafe pointers helps sparing
             // some nasty index calculations...
 
-            uint[] quotient = new uint[left.Length - right.Length + 1];
-            remainder = new uint[right.Length];
+            uint[] localLeft = CreateCopy(left);
```
stephentoub: This copy is necessary because Divide writes over the input?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
-                if (targetLength > valueLength)
-                {
-                    target[valueLength] =
-                        value[valueLength - 1] >> backShift;
-                }
-            }
-            else
+            // Repairs the dividend, if the last subtract was too much
+
+            int i = 0;
```
stephentoub: Nit: why is the declaration of ```i``` separated out from the for loop?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
+                                                   uint* right, int rightLength,
+                                                   ulong q)
         {
-            Debug.Assert(valueLength >= 1);
-            Debug.Assert(targetLength == valueLength);
-            Debug.Assert(shift >= 0 && shift < 32);
+            Debug.Assert(leftLength >= 0);
+            Debug.Assert(rightLength >= 0);
+            Debug.Assert(leftLength >= rightLength);
+            Debug.Assert(q <= 0xFFFFFFFF);
```
stephentoub: ```Debug.Assert(q <= uint.MaxValue)``` ?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
-                    (value[valueLength - 1] >> shift);
-            }
-            else
+            // Combines a subtract and a multiply operation, which is naturally
+            // more efficient than multiplying and then subtracting...
+
+            int i = 0;
+            ulong carry = 0L;
+
+            for (; i < rightLength; i++)
```
stephentoub: Same question about ```i```.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
+        private unsafe static uint SubtractDivisor(uint* left, int leftLength,
+                                                   uint* right, int rightLength,
+                                                   ulong q)
         {
-            Debug.Assert(valueLength >= 1);
-            Debug.Assert(targetLength == valueLength);
-            Debug.Assert(shift >= 0 && shift < 32);
+            Debug.Assert(leftLength >= 0);
+            Debug.Assert(rightLength >= 0);
+            Debug.Assert(leftLength >= rightLength);
```
stephentoub: Doesn't matter much, but if we're asserting that ```leftLength >= rightLength``` and that ```rightLength >= 0```, is it necessary to also assert that ```leftLength >= 0``` as is done above?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
+                left[i] -= digit;
             }
+
+            return (uint)carry;
         }
 
-        [SecuritySafeCritical]
-        private unsafe static void MultiplyDivisor(uint* left, int leftLength,
-                                                   ulong right, uint* bits)
+        private static bool CompareDivisor(uint divHi, uint divLo,
```
stephentoub: Another example where I think comments on the function would be helpful, i.e. what does the bool return mean?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
+
+            return false;
+        }
+
+        private static uint[] CreateCopy(uint[] value)
+        {
+            Debug.Assert(value != null);
+
+            var bits = new uint[value.Length];
+            Array.Copy(value, bits, bits.Length);
```
stephentoub: ```Array.Copy(value, 0, bits, 0, bits.Length);```

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -128,203 +129,194 @@ public unsafe static uint[] Remainder(uint[] left, uint[] right)
...
+                return true;
+
+            return false;
+        }
+
+        private static uint[] CreateCopy(uint[] value)
+        {
+            Debug.Assert(value != null);
+
+            var bits = new uint[value.Length];
```
stephentoub: Will it be common for ```value.Length == 0``` ?  If no, nevermind.  If yes, consider special-casing it to return ```Array.Empty<uint>()```.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
```diff
@@ -0,0 +1,432 @@
...
+
+            BitsBuffer tempBuffer = new BitsBuffer(length, 0);
+            BitsBuffer valueBuffer = new BitsBuffer(length, value);
+            BitsBuffer resultBuffer = new BitsBuffer(length, 1);
+
+            PowCore(power, ref valueBuffer, ref resultBuffer,
+                ref tempBuffer);
+
+            return resultBuffer.ToArray();
+        }
```
stephentoub: The two Pow methods are almost identical.  I wonder if they should be refactored to share most of the code, e.g.
```C#
public static uint[] Pow(uint value, uint power)
{
    int length = PowBound(power, 1, 1);
    BitsBuffer valueBuffer = new BitsBuffer(length, value);
    return PowCore(length, power, ref valueBuffer);
}

public static uint[] Pow(uint[] value, uint power)
{
    Debug.Assert(value != null);
    int length = PowBound(power, value.Length, 1);
    BitsBuffer valueBuffer = new BitsBuffer(length, value);
    return PowCore(length, power, ref valueBuffer);
}

private static uint[] PowCore(int length, uint power, ref BitsBuffer valueBuffer)
{
    BitsBuffer tempBuffer = new BitsBuffer(length, 0);
    BitsBuffer resultBuffer = new BitsBuffer(length, 1);
    PowCore(power, ref valueBuffer, ref resultBuffer, ref tempBuffer);
    return resultBuffer.ToArray();
}
```
or something like that?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
```diff
@@ -0,0 +1,432 @@
...
+            }
+
+            return (uint)(result % modulus);
+        }
+
+#if DEBUG
+        private const int ReducerThreshold = 8;
+#else
+        private const int ReducerThreshold = 32;
+#endif
```
stephentoub: Why is it different for debug?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
```diff
@@ -0,0 +1,432 @@
...
+            return ActualLength(value, value.Length);
+        }
+
+        private static int ActualLength(uint[] value, int length)
+        {
+            Debug.Assert(value != null);
+            Debug.Assert(length <= value.Length);
+
+            while (length > 0 && value[length - 1] == 0)
+                --length;
```
stephentoub: ```while (length > 0 && value[length--] == 0);``` ?

src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
```diff
@@ -19,7 +19,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Numerics\BigIntegerCalculator.AddSub.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.Buffer.cs" />
     <Compile Include="System\Numerics\BigIntegerCalculator.DivRem.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.PowMod.cs" />
+    <Compile Include="System\Numerics\BigIntegerCalculator.Reducer.cs" />
     <Compile Include="System\Numerics\BigIntegerCalculator.SquMul.cs" />
     <Compile Include="System\Numerics\BigInteger.cs" />
     <Compile Include="System\Numerics\BigIntegerBuilder.cs" />
```
stephentoub: With all of the changes that have been made, it seems like a large chunk of BigIntegerBuilder.cs is now dead code / unreachable methods.  If that's true, we should clean that up (can be done separately).

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
axelheer: Sure.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
axelheer: Ah, didn't know that.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
axelheer: Sure, I'll write some comments.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
axelheer: Okay.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
axelheer: The temps should be empty between operations, I'll add asserts to clarify.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Yeah, I'll add comments to clarify.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Good question, I'll change that.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: I prefer explicit values, is that ok?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: I'll change that.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Nope, not necessary, I just try to write asserts as verbose as meaningful.

stephentoub:
Also, I know we don't have any formal performance harnesses set up, but are your perf tests checked in anywhere?  If not, could you please add them?  You could check them in as unit tests with an [ActiveIssue("PerfTest")] or something like that for now, until we have a better performance-measurement system in place.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: I tried to describe that in the comment below: "[..] return true, which means the current guess is still too big"

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Sure.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Should not be zero, I'll add an assert to clarify.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
stephentoub: It's ok, and you can leave it if you like.  I just worry about typos, accidentally having one too few or one too many Fs, etc.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
stephentoub: Yes, but that requires someone to read the implementation to understand the signature of a method.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
axelheer: Sure.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
axelheer: To test / debug the reducer code with smaller values.

src/System.Runtime.Numerics/src/System.Runtime.Numerics.csproj
axelheer: I'd like to port the remaining code to BigIntegerCalculator, then we could drop it. If that's okay, I'd prefer a separate PR, since it will take some time.

stephentoub: Yup, no problem.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
stephentoub: My concern is that a lot of our testing is done against debug builds, and I get a bit concerned any time I see debug vs release settings changing behavior, which this essentially does.  What would the downside be to just always making it 32?  Are we unable to write tests that appropriately exercise the reducer and still complete in a reasonable amount of time?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: I see, then I've chosen the method's name badly? Maybe `GuessTooBig` or something like that?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.PowMod.cs
axelheer: I totally understand that concern, but it would certainly compromise the speed of the unit tests, since the test implementation is optimized for safety and not for speed (I'd assume).

I added the "missing" `ModPow3LargeInt` test, which increased the test time from 0:10 to 1:48 on my machine... and the test values have only a few hundred bits. :(

What should I do?

axelheer:
@stephentoub thanks, I pushed a few changes addressing your input so far. For your debug / release concern I don't came up with a solution, maybe you've an idea what I should do? The performance tests I've made are based on a very primitive console app (I linked to its gist in the PR message), it's far from a bunch of unit tests. If you want me to check in that code (although I don't think it's that exciting), where? If you want me to write some tests based on it, how?

stephentoub:
> For your debug / release concern I don't came up with a solution, maybe you've an idea what I should do?

What is the time distribution like amongst the various tests?  Are there any we could move to be [OuterLoop] such that we'd be able to always use the 32 value, still have reasonably good inner-loop test coverage, have that inner-loop coverage run relatively quickly, but still have the full suite available to run as part of outer loop?

> If you want me to write some tests based on it, how?

I was thinking you could do something like this:
1. Create a PerformanceTests.cs file in the tests project.
2. Add to it one [Fact] per perf test, something like:
```C#
[Fact]
[ActiveIssue("PerformanceTest")]
public static void ModPow()
{
    ... // most of the contents of your existing console test for ModPow
}
```
Then when someone wants to run the test, they can just remove the [ActiveIssue] attribute, or something like that.  And it should hopefully make it easier for us to promote these to real and harnessed performance tests when we have the infrastructure available publically.

Is that reasonable?  I was aiming for something easy for you to do and with minimal ceremony.

axelheer:
> What is the time distribution like amongst the various tests?

Just `ModPow3LargeInt` is already a bit slow for values < 32.

> Are there any we could move to be [OuterLoop] such that we'd be able to always use the 32 value

What does `OuterLoop` mean? I didn't find any documentation on it. Note that we don't have tests for >= 32 values -- if I remove the debug / release stuff, we won't cover the reducer code. And skipping the whole reducer code for an "ordinary" test run... ?

I might add, that the changed behavior should be very harmless, it's just a threshold when to switch to the "more sophisticated" code, which doesn't depend on that value. On the contrary, if it works even for smaller values, it's a good thing.

> I was thinking you could do something like this [...]

I see, I can add that for all those operations, but I'd prefer to do that separately on a subsequent date.

By the way, I did more cleanup on the "modpow" code based on your inputs and will push that shortly.

stephentoub:
> What does OuterLoop mean?

We currently have two main classes of tests.  The first, called "inner loop", are meant to be fast, do basic validation, and run whenever you do a build locally or as part of CI on the server.  The second, called "outer loop", are meant to be more robust and are ok to take longer; these don't run by default when you build locally, nor as part of normal CI runs... we have a special job on the server that runs them a few times a day.  You can also run them locally with xunit by not passing "-notrait category=outerloop" on the command-line (this argument is specified by default when you do a test run locally).

> Note that we don't have tests for >= 32 values -- if I remove the debug / release stuff, we won't cover the reducer code

That feels like a gap... doesn't that mean that the reducer stuff won't be used at all when tests are run on release builds?

>  it's just a threshold when to switch to the "more sophisticated" code

That also means that we're not testing the normal algorithm in debug for values between 8 and 32 bits, yeah that's the algorithm we'll be using on that range in production.  That feels wrong.

How important is it to performance that ReducerThreshold be a const?  If it were, for example, a static  int that was read each time it was needed, would that be measurable?  If that wouldn't affect anything, we could potentially test things by mucking with the threshold value from the tests via reflection at runtime.  Not ideal, for sure, but could be the least evil thing.

> I can add that for all those operations, but I'd prefer to do that separately on a subsequent date.

That's fine, thanks.

axelheer:
Hm, we've that debug / release thing already three times: multiply, square, modpow.

I can remove this "trick" and change the thresholds to static values as you've suggested, which should work since xunit runs the tests at class level single threaded, and add further tests / extend existing tests to change the thresholds accordingly. That should do it, yeah. Not ideal, but better than the current one.

axelheer:
@stephentoub the last commit should test as discussed. I actually like that solution now. The (new) tests for three big values became `OuterLoop` tests, since they only add coverage for a very small function but take a long time.

stephentoub:
Left a few minor comments, but otherwise LGTM.  Feel free to squash when ready.

stephentoub:
LGTM.  Thanks for doing this, @axelheer.

@KrzysztofCwalina, look good to you?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System.Diagnostics;
```
KrzysztofCwalina: Feel free to commit the change as is, but it would be better if this file was called ...BitsBuffer.cs to match the type name.

KrzysztofCwalina:
Looks good. Thanks!

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.Buffer.cs
jasonwilliams200OK: @KrzysztofCwalina, which name is more suitable in this case: `BigIntegerCalculator.BitsBuffer.cs` or `BitsBuffer.cs`?

KrzysztofCwalina: This project already uses the <outer_type>.<inner_type>.cs conventions, so I would stick to it. And in general, I like the convention, though I am not sure if it's standard over the whole CoreFx repo. @terrajobst, do we have a standard convention for naming files with partial types containing nested type?

axelheer: I'd prefer `BigIntegerCalculator.BitsBuffer.cs` instead of `BitsBuffer.cs` (lexicographical reasons). Since the other files have just abbreviations as suffix, I've chosen the shorter `BigIntegerCalculator.Buffer.cs` (long file names are so java...).

But I haven't that much emotion for these file names, just say and I amend. ;)

stephentoub: Let's go with BigIntegerCalculator.BitsBuffer.cs.

axelheer:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
Thanks!

PR #2186, opened 2015-06-28
Fixed paths to unix build instructions.

MI3Guy:
I noticed that the paths to the build instructions in the shell script were incorrect so I updated them.


stephentoub:
LGTM. Thanks.

PR #2187, opened 2015-06-28
Fix config file headline comment

peterblazejewicz:
This is a small PR corrects unwanted line break added to original
EditorConfig example content headline: http://editorconfig.org/#example-file

Thanks!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,4 @@
-# EditorConfig is awesome:
-http://EditorConfig.org
+# editorconfig.org
 
 # top-most EditorConfig file
 root = true


.editorconfig
```diff
@@ -1,5 +1,4 @@
-# EditorConfig is awesome:
-http://EditorConfig.org
+# EditorConfig is awesome: http://EditorConfig.org
```
stephentoub: Could we just change this to be:
```
# editorconfig.org
```
?

stephentoub:
Thanks, @peterblazejewicz. LGTM.

.editorconfig
peterblazejewicz: @stephentoub :+1:  rebased, squashed and commit message changed. Thanks!

Issue #1727, opened 2015-05-13
Investigate implementing Process.Modules on Unix


stephentoub:
On Unix, we currently return an empty array from Process.Modules, and we throw a PlatformNotSupportedException from Process.MainModule.  Is there something better we can reasonably do on Linux and OSX?  If so, we should do it.

ianhays:
Mono on Linux implemented this functionality and will print the names of the loaded dll’s or exe’s, including that of the running program. In a Hello World example, getting the Modules of the running process will return:
* Msorlib
* Hello
* System

The source for Mono’s implementation is archaic and resides chiefly in the [ves_icall_System_Diagnostics_Process_GetModules_internal()] (https://github.com/mono/mono/blob/55d8296440ee2c62b7f22dd7ad6e3836f09e5db4/mono/metadata/process.c) function. In Mono, the ability to list out the loaded Modules is handwritten in C, whereas in .NET Windows it relies upon the C++ function [EnumProcessModules] (https://msdn.microsoft.com/en-us/library/windows/desktop/ms682631(v=vs.85).aspx). It would require a great deal of effort to replicate the Mono functionality and although it is possible, it is frankly beyond me. Perhaps another set of eyes could discern more than mine, but as far as I could tell Modules in Mono rely heavily on the underlying runtime implementation and wouldn't translate to .NET.


I'm still searching through OSX and Linux docs for APIs that do similar things, but have thus far come upon nothing of use.


As far as alternatives go, I personally like the suggestions in [github issue 1081] (https://github.com/dotnet/corefx/issues/1081). 
* @terrajobst suggests that we add something like Process.CanQueryModuleInformation if we are going to have Process.MainModule throw an exception on some platforms. I agree.
* @Giorgi suggests we add a Process.Path property to get the path of the executable. This would take some of the use cases away from Process.MainModule and Process.Modules (which would throw/return the empty array on Unix) while also providing value in the Windows implementations.

stephentoub:
Closed by #2188

stephentoub:
Still needs an OS X implementation...

PR #2188, opened 2015-06-29
Implement basic System.Diagnostics.Process.Modules support on Linux

stephentoub:
I'd previously left a comment in the implementation that we could potentially implement Process.Modules support by enumerating the contents of the procfs /procs/[pid]/maps file.  This commit does that.  It uses the heuristic that if the pathname isn't empty and the mapped region has both read and execute permissions, then the entry should be considered a module.  We can tweak this logic over time as needed, and potentially augment the results with a different data source if we find an appropriate one.

#1727

### DIFF
--- a/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
+++ b/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 
@@ -12,6 +16,7 @@ internal static partial class Interop
         internal const string SelfExeFilePath = RootPath + "self/exe";
         internal const string ProcUptimeFilePath = RootPath + "uptime";
         private const string StatFileName = "/stat";
+        private const string MapsFileName = "/maps";
         private const string TaskDirectoryName = "/task/";
 
         internal struct ParsedStat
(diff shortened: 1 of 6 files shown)


stephentoub:
cc: @ellismg, @sokket, @ianhays

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
...
         internal static string GetTaskDirectoryPathForProcess(int pid)
         {
             return RootPath + pid.ToString(CultureInfo.InvariantCulture) + TaskDirectoryName;
         }
 
+        internal static IEnumerable<ParsedMapsModule> ParseMapsModules(int pid)
+        {
+            try
+            {
+                return ParseMapsModulesCore(File.ReadAllLines(GetMapsFilePathForProcess(pid)));
```
justinvp: I'm curious, why `File.ReadAllLines` and not `File.ReadLines`? `ReadAllLines` involves some extra allocations in its implementation, not to mention the array allocation. If there is a good reason to stick with `File.ReadAllLines`, perhaps `ParseMapsModulesCore`'s parameter should be changed from `IEnumerable<string>` to `string[]` to avoid the enumerator allocation when enumerating the lines.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
...
+
+            // Parse each line from the maps file into a ParsedMapsModule result
+            foreach (string line in lines)
+            {
+                // Use a StringParser to avoid string.Split costs
+                var parser = new StringParser(line, separator: ' ', skipEmpty: true);
+
+                // Parse the address range 
+                KeyValuePair<long, long> addressRange =  
+                    parser.ParseRaw(delegate (string s, ref int start, ref int end)
```
justinvp: Not that it matters much, but from a coding style perspective, why an anonymous method and not a lambda expression?

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
...
+                            if (long.TryParse(startingString, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out startingAddress))
+                            {
+                                string endingString = s.Substring(pos + 1, end - (pos + 1));
+                                long.TryParse(endingString, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out endingAddress);
+                            }
+                        }
+                        return new KeyValuePair<long, long>(startingAddress, endingAddress);
+                    });
+
+                // Pars the permissions (we only care about entries with 'r' and 'x' set)
```
justinvp: Typo: "Pars" -> "Parse"

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
...
+                        else if (s[i] == 'x')
+                            sawExec = true;
+                    }
+                    return sawRead & sawExec;
+                }))
+                {
+                    continue;
+                }
+
+                // Skip pst the offset, dev, and inode fields
```
justinvp: Typo: "pst" -> "past"?

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
...
+                            {
+                                string endingString = s.Substring(pos + 1, end - (pos + 1));
+                                long.TryParse(endingString, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out endingAddress);
+                            }
+                        }
+                        return new KeyValuePair<long, long>(startingAddress, endingAddress);
+                    });
+
+                // Pars the permissions (we only care about entries with 'r' and 'x' set)
+                if (!parser.ParseRaw(delegate (string s, ref int start, ref int end)
```
justinvp: Same question about anonymous method vs. lambda expression.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: Because of the ref parameters.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: Thanks, will fix.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: Because of the ref parameters.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: Thanks, will fix.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: I'd meant to use ReadLines, hence the ```IEnumerable<string>``` rather than ```string[]``` argument.  I'll change it.

jasonwilliams200OK:
@stephentoub, does `System.Diagnostics.Process.Module` has any relation with `System.Diagnostics.ProcessModule` class and is `ProcessModule` class xplat compatible?

stephentoub:
Process.Modules returns a ProcessModuleCollection, which contains ProcessModule instances, which wrap internal ModuleInfo instances.  Those are just data structures and are platform-agnostic, but the ModuleInfo instances are populated in a platform-specific manner (e.g. https://github.com/dotnet/corefx/pull/2188/files#diff-e7145ead94d1f1c72f6efbc2174060c1R35), as it's gathering data about the system using platform-provided support.

jasonwilliams200OK:
Thanks for the info @stephentoub!
For FreeBSD, we can probably enumerate over [`procstat`](https://www.freebsd.org/cgi/man.cgi?query=procstat&sektion=1) dump to implement this property.
(copy: @ghuntley, @janhenke, @josteink, @saper)

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
justinvp: Should this be codified in the Coding Style, or is it not worth the extra rule for such a trivial thing?

The reason I asked is because the MSDN [documentation](https://msdn.microsoft.com/en-us/library/0yw3tz5k.aspx) prefers the use of the lambda syntax over anonymous method syntax:

> C# 2.0 introduced anonymous methods and in C# 3.0 and later, lambda expressions supersede anonymous methods as the preferred way to write inline code.

Unless omitting the parameter list:

> There is one case in which an anonymous method provides functionality not found in lambda expressions. Anonymous methods enable you to omit the parameter list. This means that an anonymous method can be converted to delegates with a variety of signatures. This is not possible with lambda expressions.

I realize that when you have to specify the parameter types (because of ref), there isn't much difference between `delegate (string s, ref int start, ref int end) { ... }` and `(string s, ref int start, ref int end) => { ... }`.

I was just wondering if the rule should be "Prefer the lambda syntax, unless there is a need to omit the parameter list, in which case the anonymous method syntax can be used".

stephentoub: I don't have a strong opinion on it; personally I think it's a trivial difference and don't think we need to be strict about it, but I'm fine if others disagree and want to codify it one way or the other.  At that point though it should be built into codeformatter and the rule applied consistently across the repo.  Separate from that I'm going to leave this as-is.

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
```diff
@@ -67,16 +72,104 @@ internal struct ParsedStat
             //internal long cguest_time;
         }
```
sokket: Is this struct used anywhere? It doesn't look to be here and is completely commented out

sokket:
LGTM

src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
stephentoub: Which struct are you referring to?

sokket: Ah nevermind...this looks to be some tickery in the GitHub code review tool...pay no attention :)

stephentoub: Ok :smile:

PR #2194, opened 2015-06-29
Remove SignAssembly=false from test assemblies

ChadNedzlek:
Setting SignAssembly to false breaks loading our test assemblies
in XUnit using the current folder structure we use.  It doesn't show
up in the open, because we always sign the assemblies, ignoring
this setting. But in our official drop environment, we respect the
property, and are unable to run these 12 tests.


stephentoub:
LGTM

mellinoe:
LGTM

PR #2196, opened 2015-06-29
Initial implementation of X509Chain for Unix.

### LINKED ISSUE #1812: Implement X509Chain on Unix

bartonjs:
* Only supports X509RevocationMode.NoCheck
* Does not support AIA record chaining
* Adds support for the LM\Root store, based off of the system CA directory.
  * OSX has a directory for this, but it is empty.
* Implements X509ChainPolicy.CertificatePolicy and X509ChainPolicy.ApplicationPolicy matching in managed code, because the OpenSSL code doesn't provide equivalent matching.
* Differs from Windows behaviors in cases of multiple errors. Windows reports every error for every cert in the chain, OpenSSL reports only the first error it finds.
* Adds support for reading PEM-encoded certificate files
* Adds support for creating X509Certificate2 objects from an existing IntPtr on Unix.
* Adds support for the SHA-2 family of algorithms when using OpenSSL.

Resolves #1812, though new issues will need to be created to track the still-missing aspects.


bartonjs:
Individual commits can be viewed from the working branch, for anyone who wants to see the smaller versions: https://github.com/dotnet/corefx/compare/master...bartonjs:x509chain?expand=1.

src/Common/src/Interop/Unix/libcrypto/Interop.Initialization.cs
```diff
@@ -21,8 +18,15 @@ static libcrypto()
...
             }
 
+            // Load the SHA-2 hash algorithms, and anything else not in the default
+            // support set.
+            OPENSSL_add_all_algorithms_conf();
+
             // Ensure that the error message table is loaded.
-            Interop.libcrypto.ERR_load_crypto_strings();
+            ERR_load_crypto_strings();
         }
```
jasonwilliams200OK: In the PR description you mentioned:

> * Differs from Windows behaviors in cases of multiple errors. Windows reports every error for every cert in the chain, OpenSSL reports only the first error it finds.

But looking at manual: http://www.openssl.org/docs/crypto/ERR_load_crypto_strings.html, http://wiki.openssl.org/index.php/Manual:ERR_print_errors%283%29 and PHP implementation: http://php.net/manual/en/function.openssl-error-string.php, it appears that all the errors are recorded as a stack and can be read by iterating over the function, something like:

```
// pseudo-code
for(auto i : ERR_load_crypto_strings()) { .. }
```

?

@khouzam, do you know of better way of reading "all" crypt errors in OpenSSL?

bartonjs: OpenSSL normal API/library errors are a stack.  The chain validation errors are [X509_STORE_CTX_get_error](https://www.openssl.org/docs/crypto/X509_STORE_CTX_get_error.html), which talks a lot about *the* error.  But I'm certainly open to finding out that more than one error is possible, and from more than one certificate in the chain.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.NativeCrypto.cs
```diff
@@ -51,6 +51,33 @@ internal static partial class NativeCrypto
         [DllImport(Libraries.CryptoInterop)]
         private static extern int GetAsn1StringBytes(IntPtr asn1, byte[] buf, int cBuf);
 
+        [DllImport(Libraries.CryptoInterop)]
+        internal static extern int GetX509StackFieldCount(SafeX509StackHandle stack);
+
+        /// <summary>
+        /// Gets a pointer to a certificate within a STACK_OF(X509). This pointer will later
+        /// be freed, so it should be cloned via new X509Certificate2(IntPtr)
+        /// </summary>
```
stephentoub: I'm supportive of XML comments... but just wondering what sets this method apart that it gets an XML comment and nothing else does?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePolicy.cs
```diff
@@ -0,0 +1,378 @@
...
+                // sequence we don't particularly need to drain out here.
+            }
+
+            return policies;
+        }
+
+        private static LowLevelList<CertificatePolicyMapping> ReadCertPolicyMappingsExtension(X509Extension extension)
+        {
+            DerSequenceReader reader = new DerSequenceReader(extension.RawData);
+            LowLevelList<CertificatePolicyMapping> mappings = new LowLevelList<CertificatePolicyMapping>();
```
stephentoub: Why ```LowLevelList<>``` instead of ```List<>```?  I'd thought we just used ```LowLevelSomeType``` when we couldn't reference the appropriate library containing ```SomeType``` and so instead included a copy of the source for it.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/ChainPal.cs
```diff
@@ -25,7 +25,26 @@ public static bool ReleaseSafeX509ChainHandle(IntPtr handle)
...
+                applicationPolicy,
+                certificatePolicy,
+                verificationTime);
+        }
+
+        private static void CheckRevocationMode(X509RevocationMode revocationMode)
+        {
+            if (revocationMode != X509RevocationMode.NoCheck)
+            {
+                throw new NotImplementedException(SR.WorkInProgress);
```
stephentoub: Is this a TODO, or do we expect this to remain this way?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        internal int ReadInteger()
+        {
+            byte[] integerBytes = ReadIntegerBytes();
+
+            // integerBytes is currently Big-Endian, need to reverse it for
+            // Little-Endian to pass into BigInteger.
+            Array.Reverse(integerBytes);
+            BigInteger bigInt = new BigInteger(integerBytes);
+            return (int)bigInt;
```
stephentoub: Are we expecting that integerBytes will just be 4-bytes long, or are we expecting this to wrap around?  If the former, could we just use BitConverter instead of going through a BigInteger?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
```diff
@@ -0,0 +1,262 @@
...
+            EatTag(DerTag.ObjectIdentifier);
+            int contentLength = EatLength();
+
+            // Each byte could cause 3 decimal characters to be written, plus a period. Over-allocate
+            // and avoid re-alloc.
+            StringBuilder builder = new StringBuilder(contentLength * 4);
+
+            byte firstByte = _data[_position];
+            byte first = (byte)(firstByte / 40);
+            byte second = (byte)(firstByte % 40);
```
stephentoub: Where does "40" come from?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -21,29 +21,54 @@ internal sealed class OpenSslX509CertificateReader : ICertificatePal
...
+            {
+                cert = Interop.libcrypto.OpenSslD2I(Interop.libcrypto.d2i_X509, data);
+            }
 
-                // X509_check_purpose has the effect of populating the sha1_hash value,
-                // and other "initialize" type things.
-                bool init = Interop.libcrypto.X509_check_purpose(_cert, -1, 0);
+            if (cert.IsInvalid)
+            {
+                throw new CryptographicException();
```
stephentoub: We don't get an error code we can use to get a message for either of the functions used to create the cert?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -16,23 +18,366 @@ public void Dispose()
...
+            OidCollection applicationPolicy,
+            OidCollection certificatePolicy,
+            DateTime verificationTime)
         {
-            get { throw new NotImplementedException(); }
+            X509ChainElement[] elements;
+            LowLevelList<X509ChainStatus> overallStatus = new LowLevelList<X509ChainStatus>();
+
+            // An X509_STORE is more comparable to Cryptography.X509Certificate2Collection than to
+            // Cryptography.X509Store. So read this with OpenSSL eyes, not CAPI/CNG eyes.
```
stephentoub: :smile:

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -16,23 +18,366 @@ public void Dispose()
...
+
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION:
+                    return X509ChainStatusFlags.HasNotSupportedCriticalExtension;
+
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_CERT_CHAIN_TOO_LONG:
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_OUT_OF_MEM:
+                    throw new CryptographicException();
+
+                default:
+                    Debug.Assert(false, "Unrecognized X509VerifyStatusCode:" + code);
```
stephentoub: Debug.Fail?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -16,23 +18,366 @@ public void Dispose()
...
+
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_CERT_REJECTED:
+                    return X509ChainStatusFlags.ExplicitDistrust;
+
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION:
+                    return X509ChainStatusFlags.HasNotSupportedCriticalExtension;
+
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_CERT_CHAIN_TOO_LONG:
+                case Interop.libcrypto.X509VerifyStatusCode.X509_V_ERR_OUT_OF_MEM:
+                    throw new CryptographicException();
```
stephentoub: Is CryptographicException the right exception type for *_ERR_OUT_OF_MEM, or should that be an OutOfMemoryException?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -16,23 +18,366 @@ public void Dispose()
...
+                        current,
+                        extraStore,
+                        intermediateCerts,
+                        rootCerts);
+
+                    if (results != null)
+                    {
+                        foreach (X509Certificate2 result in results)
+                        {
+                            if (result != null)
```
stephentoub: The collection can legitimately contain null?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
```diff
@@ -25,12 +30,123 @@ public static IStorePal FromCertificate(ICertificatePal cert)
...
         {
-            throw new NotImplementedException();
+            return new OpenSslX509StoreProvider(certificates);
         }
 
         public static IStorePal FromSystemStore(string storeName, StoreLocation storeLocation, OpenFlags openFlags)
         {
+            if (storeLocation != StoreLocation.LocalMachine)
+            {
+                throw new NotImplementedException();
```
stephentoub: PlatformNotSupportedException?  Or is this something we plan to implement soon?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
```diff
@@ -25,12 +30,123 @@ public static IStorePal FromCertificate(ICertificatePal cert)
...
+            {
+                byte[] bytes;
+
+                try
+                {
+                    bytes = File.ReadAllBytes(file.FullName);
+                }
+                catch (IOException)
+                {
+                    continue;
```
stephentoub: I assume we're eating these exceptions to handle race conditions where the file gets deleted concurrently with this loop?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
```diff
@@ -25,12 +30,123 @@ public static IStorePal FromCertificate(ICertificatePal cert)
...
+
+                X509Certificate2 cert;
+
+                try
+                {
+                    cert = new X509Certificate2(bytes);
+                }
+                catch (CryptographicException)
+                {
+                    continue;
```
stephentoub: Does this imply that the store is corrupted?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
```diff
@@ -25,12 +30,123 @@ public static IStorePal FromCertificate(ICertificatePal cert)
...
+                }
+
+                if (StringComparer.Ordinal.Equals(cert.Subject, cert.Issuer))
+                {
+                    if (!rootStore.Contains(cert))
```
stephentoub: How many certs do we expect we'll find?  This loop is O(N) in the number of certs, which is fine if the number is always small but concerning if it could grow to be large.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
```diff
@@ -25,12 +30,123 @@ public static IStorePal FromCertificate(ICertificatePal cert)
...
+                    {
+                        rootStore.Add(cert);
+                    }
+                }
+                else
+                {
+                    if (!s_machineIntermediateStore.Contains(cert))
```
stephentoub: Either here or at the beginning of this method, I'd suggest asserting that Monitor.IsHeld(s_machineIntermediateStore).

src/System.Security.Cryptography.X509Certificates/src/Resources/Strings.resx
```diff
@@ -204,4 +204,7 @@
   <data name="WorkInProgress" xml:space="preserve">
     <value>WorkInProgress.</value>
   </data>
-</root>
+  <data name="Chain_NoPolicyMatch" xml:space="preserve">
+    <value>The certificate has invalid policy.</value>
+  </data>
+</root>
```
stephentoub: Nit: newline got deleted

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -16,23 +18,366 @@ public void Dispose()
...
+
+                    if (!candidates.Contains(current))
+                    {
+                        candidates.Add(current);
+                    }
+
+                    X509Certificate2Collection results = FindIssuer(
+                        current,
+                        extraStore,
+                        intermediateCerts,
```
stephentoub: FindIssuer is defined to take a params array, so each call in this loop is going to allocate a new three-element array.  I'd suggest either changing the FindIssuer signature, or at least caching the array into a local so that it's allocated once rather than once per iteration.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.NativeCrypto.cs
bartonjs: The fact that the pointer has a short lifespan, and needs to be duplicated.  Most everything else here returns things whose lifetime is caller-determined (and thus marshalls through as a SafeHandle).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePolicy.cs
bartonjs: Since LowLevelList was already being used in the project I was assuming I was following the conventions that I'd never heard before.  I'll happily change it to `List<T>` (which the library also already had a reference for, since I was able to use `HashSet<T>` without needing to update the dependencies).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/ChainPal.cs
bartonjs: TODO. It's dependent on a stable networking stack (since I can't make use of the OpenSSL functions on OSX's 0.9.8). So not beta6, but I'd feel really bad if this wasn't done by X-Plat RTM.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
bartonjs: While BitConverter would throw for me if the array length is greater than 4, it wouldn't handle it being less than 4, such as `-1` (`{ 0xFF }`).

Integer in DER-land always means BigInteger in C#, but they're usually small enough to fit into an Int32, and this function is for those contextual cases.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
bartonjs: ISO/IEC 8825-1:2003 (BER/CER/DER encoding)

> ##### 8.19.4
  The numerical value of the first subidentifier is derived from the values of the first two object identifier  components in the object identifier value being encoded, using the formula:
    `(X*40)  +  Y`
    where X is the value of the first object identifier component and Y is the value of the second object identifier 
component. 
        <sub>NOTE – This packing of the first two object identifier components recognizes that only 
three values are allocated from the root node, and at most 39 subsequent values from nodes reached by X = 0 and X = 1</sub>

Bit partitioning would have been more computer-sciency :smile:.  I've taken the hint and amended the code:
```C#
            // The first byte is ((X * 40) + Y), where X is the first segment and Y the second.
            // ISO/IEC 8825-1:2003 section 8.19.4

            byte firstByte = _data[_position];
            byte first = (byte)(firstByte / 40);
            byte second = (byte)(firstByte % 40);
```

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/ChainPal.cs
stephentoub: Could we add comments for such NotImplementedExceptions calling out why they're not implemented, e.g. ```// TODO: Implement once a stable networking stack is available"?  Presumably the only NotImplementedExceptions we have are ones with TODOs associated with them, as otherwise they'd be promoted to NotSupportedException or PlatformNotSupportedException.  If there was a GitHub issue number to go along with the TODO, even better.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: The PEM page doesn't talk specifically about ERR_get_error, but some digging revealed that it does use the same error stack; so I've added the call to `Interop.libcrypto.GetOpenSslErrorString()`.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: I had the impression that only the CLR threw `OutOfMemoryException`s, but I see a handful of counter-examples to that theory; so I've made the change here.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: Fixed.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: FindIssuer definitely benefits from the ability to loop over the data.  I changed it to a non-params method and used an array that is allocated before the while.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: Nope, probably left from debugging a rogue crash somewhere; removed this check.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
bartonjs: Coming up with a plan for CurrentUser\My is on my radar ("RSA private key management"), and that should generalize to the rest of CU.  Other LocalMachine things might be... trickier :smile:.

CU\My is where support for ReadWrite comes in.  So all of these NotImplementedExceptions are expected to be functional or a different exception by RTM (I'll open issues for them).

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.NativeCrypto.cs
stephentoub: Ok, thanks.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
stephentoub: Ok, thanks.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DerSequenceReader.cs
stephentoub: :smile:

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
stephentoub: Cool.  What happens if one calls ERR_get_error and there wasn't an error?  I'm wondering if we should centralize everything into a helper like (pseudo-code):
```C#
internal static void ThrowOpenSslCryptographicException()
{
    int error = ERR_get_error();
    throw error == 0 ?
        new CryptographicException() :
        new CryptographicException(GetOpenSslErrorString(error));
}
```
and then just always use this helper, rather than relying on the incomplete docs to determine whether to get the error string or not.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
stephentoub: Thanks.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: I was thinking about that this morning.  It'd probably want to be a bit more complicated:

```C#
internal static void ThrowOpenSslCryptographicException()
{
    int error = ERR_get_error();
    int last = error;

    while (last != 0)
    {
        last = ERR_get_error();
    }

    throw error == 0 ? new CryptographicException() : new CryptographicException(GetOpenSslErrorString(error));
}
```

in order to deal with any error codes set while we hadn't bothered checking.  But, I think we're going to need to do something complicated in OpenSsl for error management; I'm pretty sure that their stack is cross-threaded, and we'd presumably want it to be unique per execution context.  I'll open an issue about the threading problem, and a separate one to make the helper function and clean up existing usages.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
bartonjs: Or that there's a format we don't know how to read (yet).  Or someone/some-distro dropped a README in the directory to remind people that on -this- system you should change this by running some-tool-or-another... (I'll add a comment)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
bartonjs: My main concern would be a broken symlink, actually.  I think we'd rather have an incomplete root store than blow up the application, so sort of just a general relaxed attitude at reading contents off of the disk. (I'll add a comment, and am certainly open to criticism)

src/System.Security.Cryptography.X509Certificates/src/Resources/Strings.resx
bartonjs: \*sigh\*. Fixed. (Is there a way to make VS always write one? Opening a file in vim *just* to hit save so it fixes the last line feels like a sad use of vim.)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
stephentoub: Thanks.  Comments in all of these cases would be helpful.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: Opened #2210 (Multi-threading implications) and #2211 (helper method).  I feel like there's guidance somewhere that helper methods should return the exceptions instead of throw them... anyone who wants to express a preference can do so on #2211 :smile:.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/StorePal.cs
bartonjs: Ubuntu 15.04 x64 has 520 files in the directory which produce 127 unique root certs, and 1 intermediate (lots of redundant symlinks).

I've gone ahead and switched to using HashSets for uniqueness tests (adding to the X509Certificate2Collections when Add returns true).  If we have a more preferred O(faster) uniqueness operator I'm more than happy to learn.

stephentoub:
Thanks for the fixes.  LGTM!

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePolicy.cs
```diff
@@ -0,0 +1,378 @@
...
+
+        private static void ReadCertPolicyConstraingsExtension(X509Extension extension, CertificatePolicy policy)
+        {
+            DerSequenceReader reader = new DerSequenceReader(extension.RawData);
+
+            while (reader.HasData)
+            {
+                switch (reader.PeekTag())
+                {
+                    case 0x80:
```
rielz: This number should be explained. Something like:
`Identifier of context-specific, primitive tag for field 0 in PolicyConstraints (see X.509 section 4.2.1.11)`

bartonjs: Good call.  I added the attribution comment, but went a step further and made the Context-Specific bit a const and rewrote the line here as
```C#
case DerSequenceReader.ContextSpecificTagFlag | 0:
```

PR #2197, opened 2015-06-30
Adding System.Data.Common Source

saurabh500:
Build for System.Data.Common sources

#885


src/System.Data.Common/src/System.Data.Common.csproj
```diff
@@ -0,0 +1,50 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- A reference to the entire .NET Framework is automatically included -->
+  </ItemGroup>
+  <ItemGroup>
+	<Compile Include="System\DBNull.cs" />
```
stephentoub: Nit: indentation

src/System.Data.Common/src/System.Data.Common.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <AssemblyName>System.Data.Common</AssemblyName>
+    <AssemblyVersion>4.0.0.0</AssemblyVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
```
stephentoub: Nit: this item group can be removed (the comment seems out-of-date)

src/System.Data.Common/src/System/DBNull.cs
```diff
@@ -0,0 +1,36 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+using System;
+
+
+namespace System
+{
+    public sealed class DBNull
+    {
+        //Package private constructor
```
stephentoub: I'm not sure what "package" means in this context.

src/System.Data.Common/src/System/DBNull.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+////////////////////////////////////////////////////////////////////////////////
+// Void
```
stephentoub: Shoudl this be "DBNull" instead of "Void", or should this header be removed entirely (leaving the copyright in place)?

src/System.Data.Common/src/System/Data/CommandBehavior.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace System.Data
+{
+    [Flags()]
+    public enum CommandBehavior
+    {
+        Default = 0,  // with data, multiple results, may affect database, MDAC 68240
+        SingleResult = 1,  // with data, force single result, may affect database
+        SchemaOnly = 2,  // column info, no data, no effect on database
+        KeyInfo = 4,  // column info + primary key information (if available)
+        //
```
stephentoub: What does this "//" mean... is it a placeholder for something?

src/System.Data.Common/src/System/Data/CommandBehavior.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+
+//------------------------------------------------------------------------------
+
+namespace System.Data
+{
+    [Flags()]
```
stephentoub: Nit: unneeded parens

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using Res = System.SR;
+
+namespace System
+{
+    internal static partial class SR
```
stephentoub: We should use the same SR mechanism employed by the rest of corefx.

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
```diff
@@ -0,0 +1,206 @@
...
+        // location so that the catcher of the exception will have the appropriate call stack.
+        // This class is used so that there will be compile time checking of error messages.
+        // The resource Framework.txt will ensure proper string text based on the appropriate
+        // locale.
+
+        static internal Task<T> CreatedTaskWithException<T>(Exception ex)
+        {
+            TaskCompletionSource<T> completion = new TaskCompletionSource<T>();
+            completion.SetException(ex);
+            return completion.Task;
```
stephentoub: This can all just be:
```C#
return Task.FromException<T>(ex);
```
or, better yet, the call sites for this could just use FromException directly.

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        static internal Task<T> CreatedTaskWithCancellation<T>()
+        {
+            TaskCompletionSource<T> completion = new TaskCompletionSource<T>();
+            completion.SetCanceled();
+            return completion.Task;
+        }
```
stephentoub: Likewise, this can just be:
```C#
return Task.FromCanceled(new CancellationToken(true));
```
If there's a CancellationToken that could be passed in, that could be passed to FromCanceled instead of manufacturing a dummy canceled token.

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
```diff
@@ -0,0 +1,206 @@
...
+            {
+                if (s_falseTask == null)
+                {
+                    s_falseTask = Task.FromResult<bool>(false);
+                }
+                return s_falseTask;
+            }
+        }
+
+        static internal void TraceExceptionAsReturnValue(Exception e)
```
stephentoub: What is this method meant to be doing?

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            UnknownTransactionFailure = 60,
+        }
+        static internal Exception InternalError(InternalErrorCode internalError)
+        {
+            return InvalidOperation(Res.GetString(Res.ADP_InternalProviderError, (int)internalError));
+        }
+        internal const int DefaultConnectionTimeout = DbConnectionStringDefaults.ConnectTimeout;
+
+        static internal bool IsEmpty(string str)
```
stephentoub: Can code just use string.IsNullOrEmpty?

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
```diff
@@ -0,0 +1,401 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+
+namespace System.Data.Common
+{
+    internal sealed class DataRecordInternal : DbDataRecord, ICustomTypeDescriptor
+    {
+        private SchemaInfo[] _schemaInfo;
+        private object[] _values;
```
stephentoub: readonly?

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
```diff
@@ -0,0 +1,401 @@
...
+            return _schemaInfo[i].typeName;
+        }
+
+        public override Type GetFieldType(int i)
+        {
+            return _schemaInfo[i].type;
+        }
+
+        public override int GetOrdinal(string name)
+        { // MDAC 69015
```
stephentoub: What do these comments mean?

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
```diff
@@ -0,0 +1,401 @@
...
+            return cbytes;
+        }
+
+        public override char GetChar(int i)
+        {
+            string s;
+
+            s = (string)_values[i];
+            char[] c = s.ToCharArray();
+            return c[0];
```
stephentoub: Converting a char array just to extract the first char seems unnecessary... why not just:
```C#
public override char GetChar(int i)
{
    return ((string)_values[i])[0];
}
```
?

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+using System;
+using System.Data;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace System.Data.Common
+{
+    public abstract class DbCommand :
+        IDisposable
+    { // V1.2.3300
```
stephentoub: What does this comment mean?

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+        protected DbCommand() : base()
```
stephentoub: ```: base()``` is unnecessary.

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+        public Task<int> ExecuteNonQueryAsync()
+        {
+            return ExecuteNonQueryAsync(CancellationToken.None);
+        }
+
+        public virtual Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return ADP.CreatedTaskWithCancellation<int>();
```
stephentoub: Here's an example where cancellationToken could be passed in to CreatedTaskWithCancellation, which could then pass it to FromCanceled.  That stores the token into the returned Task, such that the token will be propagated as part of any OperationCanceledExceptions that get thrown, enabling consumers to correlate what token got a cancellation request with what exception emerged.

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+            }
+            else
+            {
+                CancellationTokenRegistration registration = new CancellationTokenRegistration();
+                if (cancellationToken.CanBeCanceled)
+                {
+                    registration = cancellationToken.Register(CancelIgnoreFailure);
```
stephentoub: This allocates a delegate on every call.  To avoid that and allow the compiler to cache the delegate so that it's reused for all calls, this could instead be:
```C#
registration = cancellationToken.Register(s => ((DbCommand)s).CancelIgnoreFailure(), this);
```

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+                }
+
+                try
```
stephentoub: What happens if cancellation is requested prior to invoking ExecuteNonQuery?  I'm wondering if the implementation appropriately handles that race condition (which could happen because at this point a cancellation request has been wired up to CancelIgnoreFailure).

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
```
stephentoub: Same comments as in the previous function.  I'll refrain from commenting on these further if they arise again.

src/System.Data.Common/src/System/Data/Common/DbConnection.cs
```diff
@@ -0,0 +1,154 @@
...
+        abstract public void Open();
+
+        public Task OpenAsync()
+        {
+            return OpenAsync(CancellationToken.None);
+        }
+
+        public virtual Task OpenAsync(CancellationToken cancellationToken)
+        {
+            TaskCompletionSource<object> taskCompletionSource = new TaskCompletionSource<object>();
```
stephentoub: Why use a TaskCompletionSource here?  Since everything is happening synchronously, instead of the calls to Set* below, the code can just call Task.FromResult/FromException/FromCanceled as was done in other methods previously.

src/System.Data.Common/src/System/Data/Common/DbConnectionOptions.cs
```diff
@@ -0,0 +1,621 @@
...
+        private static class KEY
+        {
+            internal const string Integrated_Security = "integrated security";
+            internal const string Password = "password";
+            internal const string Persist_Security_Info = "persist security info";
+            internal const string User_ID = "user id";
+        };
+
+        // known connection string common synonyms
+        private static class SYNONYM
```
stephentoub: Do these types need to be cased this way, or could they just be Key and Synonym?

src/System.Data.Common/src/System/Data/Common/DbConnectionOptions.cs
```diff
@@ -0,0 +1,621 @@
...
+        private readonly Hashtable _parsetable;
+        internal readonly NameValuePair KeyChain;
+        internal readonly bool HasPasswordKeyword;
+
+
+        // synonyms hashtable is meant to be read-only translation of parsed string
+        // keywords/synonyms to a known keyword string
+        public DbConnectionOptions(string connectionString, Hashtable synonyms)
+        {
+            _parsetable = new Hashtable();
```
stephentoub: _parsetable only appears to be used in the constructor.  Does it need to be a field?  And if so, could it be a strongly-typed ```Dictionary<TKey, TValue>``` instead of the legacy, weakly-typed ```Hashtable```?

src/System.Data.Common/src/System/Data/Common/DbConnectionOptions.cs
```diff
@@ -0,0 +1,621 @@
...
+            _usersConnectionString = ((null != connectionString) ? connectionString : "");
```
stephentoub: Could just be:
```C#
_usersConnectionString = connectionString ?? string.Empty;
```

src/System.Data.Common/src/System/Data/Common/DbConnectionOptions.cs
```diff
@@ -0,0 +1,621 @@
...
+                            {
+                                isEquivalent = true;
+                            }
+                        }
+                    }
+                    Debug.Assert(isEquivalent, "ParseInternal code vs regex message mismatch: <" + msg1 + "> <" + msg2 + ">");
+                }
+                else
+                {
+                    Debug.Assert(false, "ParseInternal code vs regex throw mismatch " + f.Message);
```
stephentoub: Debug.Fail?

src/System.Data.Common/src/System/Data/Common/DbConnectionStringBuilder.cs
```diff
@@ -0,0 +1,318 @@
...
+            set
+            {
+                ADP.CheckArgumentNull(keyword, "keyword");
+                bool flag = false;
+                if (null != value)
+                {
+                    string keyvalue = DbConnectionStringBuilderUtil.ConvertToString(value);
+                    DbConnectionOptions.ValidateKeyValuePair(keyword, keyvalue);
+
+                    flag = CurrentValues.ContainsKey(keyword);
```
stephentoub: Doesn't look like flag is being consumed anywhere, just set.

src/System.Data.Common/src/System/Data/Common/DbConnectionStringBuilder.cs
```diff
@@ -0,0 +1,318 @@
...
+            }
+        }
+
+        void System.Collections.IDictionary.Add(object keyword, object value)
+        {
+            Add(ObjectToString(keyword), value);
+        }
+        public void Add(string keyword, object value)
+        {
+            this[keyword] = value;
```
stephentoub: ```Add``` will end up overwriting an existing value that may already be there; is that the desired behavior?

src/System.Data.Common/src/System/Data/Common/DbConnectionStringCommon.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+
+
+namespace System.Data.Common
+{
+    internal sealed class ReadOnlyCollection<T> : System.Collections.ICollection, ICollection<T>
+    {
+        private T[] _items;
```
stephentoub: readonly?

src/System.Data.Common/src/System/Data/Common/DbDataReader.cs
```diff
@@ -0,0 +1,297 @@
...
+                long bytesReadTotal = 0;
+                byte[] buffer = new byte[4096];
+                do
+                {
+                    bytesRead = GetBytes(ordinal, bytesReadTotal, buffer, 0, buffer.Length);
+                    bufferStream.Write(buffer, 0, (int)bytesRead);
+                    bytesReadTotal += bytesRead;
+                } while (bytesRead > 0);
+
+                return new MemoryStream(bufferStream.ToArray(), false);
```
stephentoub: Why copy to another memory stream rather than just returning bufferStream, i.e.
```C#
public virtual Stream GetStream(int ordinal)
{
    var bufferStream = new MemoryStream();
    ... // copy loop
    bufferStream.Position = 0;
    return bufferStream;
}
```

src/System.Data.Common/src/System/Data/Common/DbDataReader.cs
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+
+        virtual public TextReader GetTextReader(int ordinal)
+        {
+            if (IsDBNull(ordinal))
+            {
+                return new StringReader(String.Empty);
```
stephentoub: ```return TextReader.Null;``` ?

src/System.Data.Common/src/System/Data/Common/DbException.cs
```diff
@@ -0,0 +1,25 @@
...
+
+
+//------------------------------------------------------------------------------
+
+namespace System.Data.Common
+{
+    public abstract class DbException :
+        Exception
+    {
+        protected DbException() : base()
```
stephentoub: ```base()``` isn't necessary.

src/System.Data.Common/src/System/Data/Common/FieldNameLookup.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Diagnostics;
+using System.Globalization;
+using System.Text;
+
+namespace System.Data.ProviderBase
+{
+    internal sealed class FieldNameLookup
+    { // V1.2.3300, MDAC 69015, 71470
+        // hashtable stores the index into the _fieldNames, match via case-sensitive
+        private Hashtable _fieldNameLookup;
```
stephentoub: ```Dictionary<TKey, TValue>``` ?

src/System.Data.Common/src/project.lock.json
```diff
@@ -0,0 +1,615 @@
...
+      "System.Text.RegularExpressions >= 4.0.10-beta-*",
+      "System.Text.Encoding >= 4.0.10-beta-*",
+      "System.Runtime >= 4.0.20-beta-*",
+      "System.Runtime.Extensions >= 4.0.10-beta-*",
+      "System.Threading.Tasks >= 4.0.10-beta-*",
+      "System.Diagnostics.Tools >= 4.0.0-beta-*"
+    ],
+    "DNXCore,Version=v5.0": []
+  }
+}
```
stephentoub: Are there tests to go along with the library?

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
```diff
@@ -0,0 +1,401 @@
...
+            s = (string)_values[i];
+
+            char[] data = s.ToCharArray();
+
+            cchars = data.Length;
+
+            // since arrays can't handle 64 bit values and this interface doesn't
+            // allow chunked access to data, a dataIndex outside the rang of Int32
+            // is invalid
+            if (dataIndex > Int32.MaxValue)
```
Maxwe11: When this could be true?

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        public virtual Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return ADP.CreatedTaskWithCancellation<int>();
+            }
+            else
+            {
+                CancellationTokenRegistration registration = new CancellationTokenRegistration();
```
Maxwe11: Why do we need this allocation? 
Could we initialize it with null and then check it on null before calling Dispose() ?

Maxwe11:
Solution file missing?

stephentoub:
@saurabh500, just checking in on this.  Do you need any assistance making forward progress with this?

saurabh500:
@Maxwe11 The solution file will follow. Thanks for pointing it out. I was meaning to get the project code with .csproj file following which the solution file will be pushed.
@stephentoub I am working on some of code comments and sanitizing a few items. Thanks for the offer :)

src/System.Data.Common/src/project.lock.json
saurabh500: We won't be including the tests in this pull request. There is work in progress on the tests to get the tests out.

src/System.Data.Common/src/System/DBNull.cs
saurabh500: Needs to be removed.

src/System.Data.Common/src/System/DBNull.cs
saurabh500: Thanks for pointing this out. This will be removed.

stephentoub:
@saurabh500, checking in on this again.  Still making forward progress?

saurabh500:
@stephentoub I am going to update the pull request today.

stephentoub:
Great :smile:

src/System.Data.Common/src/System/Data/Common/DbConnectionStringBuilder.cs
saurabh500: This is the desired behavior. The connection string builder needs to have a single value for any connection string key.

src/System.Data.Common/src/System/Data/Common/DbDataReader.cs
saurabh500: Based on the code, it seems this has been done to return a read-only memory stream.

saurabh500:
@stephentoub  Can you take a look at this pull request and help with the merge approval.

stephentoub:
> Can you take a look at this pull request and help with the merge approval.

Two things:
1. Could you please squash this?  There are merges and reverts in the commit history we try to avoid.  Ideally this would be two commits: one with the initial code committed by dotnet-bot, and one from you with fixes based on PR feedback.
2. Can you please open an issue to track addressing the PR feedback that you didn't address?

src/System.Data.Common/src/project.lock.json
```diff
@@ -0,0 +1,615 @@
+{
+  "locked": false,
```
stephentoub: This needs to be set to true before checking in.

mellinoe: To lock it, run msbuild with /p:LockDependencies=true

src/System.Data.Common/src/System.Data.Common.csproj
saurabh500: This is taken care of

src/System.Data.Common/src/System/Data/CommandBehavior.cs
saurabh500: These comments have been cleaned up. 
They came out as a by product of the sanitizer.

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
saurabh500: Method removed during cleanup

src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
saurabh500: I changed the implementation to use the string.IsNullOrEmpty. 
The code is quite old and this kind of code has been around. Thanks for pointing this out.

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
saurabh500: Removed

src/System.Data.Common/src/System/Data/Common/DataRecordInternal.cs
saurabh500: We need to evaluate this.  I will open an issue.

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
saurabh500: Removed

src/System.Data.Common/src/System/Data/Common/DbCommand.cs
saurabh500: I will open an issue to investigate the cancellationToken related comments.

saurabh500:
Tracking with the pull request https://github.com/dotnet/corefx/pull/2443

PR #2200, opened 2015-06-30
Cleanup on FileSystem tests and resolve issue with leftover temporary files

### LINKED ISSUE #1412: FileSystem tests leave temporary files/directories behind
The System.IO.FileSystem tests do not fully clean up after themselves.  Here's the garbage left behind after I just did a successful run of those tests:
```
0eo3v51i.ogq
32nolsz2.r4s
dkriwxyi.tmp
FSEnumeratorTest
GetFileSystemEntries_str_str_TestDir
GetFiles_str_str_test_TestDir
gpkyxkxw.ife
ozi4vtk0.zqb
pbovqtu0.kpw
qpzuye3n.r1j
sa1uvib4.dtw
tn0gt1if.a5d   tn0gt1if.a5d
uc0dzyqf.u5h
udon0itw.mqx
ugmatydr.3eo
xj4wtgwp.ajj
ze4qacvk.fr3
```

ianhays:
-Running the FileSystem tests no longer leaves test directories or files in the running directory. fixes #1412
-Modernized the tests for GetFIleSystemEntries and GetFiles, splitting each test case into a standalone method runnable through xunit. Similar versions of these tests now inherit from each other rather than just being copy-pasted like they were previously. Moved most test cases from using hardcoded file names to instead use GetTestFileName() in FileSystemTest.cs. Tested on Linux and Windows.
-Changed the names of some files and classes to better follow the naming conventions of the other FileSystem tests.


src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str.cs
```diff
@@ -3,271 +3,147 @@
...
-            if (strArr.Length != 0)
-            {
-                iCountErrors++;
-                printerr("Error_207v7! Incorrect number of directories returned");
-            }
+    [Fact]
+    public void FileNameWithSpaces()
+    {
+        String strTempDir = "             ";
+        Assert.Throws<ArgumentException>(() => GetEntries(strTempDir));
```
stephentoub: Hmm.  If this passes on Unix (and it looks like it will, as the same case existed in the old code you're transforming), I think that means we have a bug in our Unix implementation, as an all-spaces name is actually a valid directory name on Unix, e.g.
```
stoub@SToubLinux1:~$ mkdir "    "
stoub@SToubLinux1:~$ cd "    "
stoub@SToubLinux1:~/    $ mkdir something
stoub@SToubLinux1:~/    $ cd something
stoub@SToubLinux1:~/    /something$ 
```

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str.cs
```diff
@@ -3,271 +3,147 @@
...
-        }
-        catch (Exception exc_general)
+    [Fact]
+    public void SubDirectories()
+    {
+        DirectoryInfo testDir = new DirectoryInfo(TestDirectory);
+        testDir.CreateSubdirectory("TestDir1");
+        testDir.CreateSubdirectory("TestDir2");
+        testDir.CreateSubdirectory("TestDir3");
+        using (new FileInfo(Path.Combine(TestDirectory, "TestFile1")).Create())
```
stephentoub: Rather than:
```C#
new FileInfo(...).Create()
```
seems like you could simplify slightly to:
```C#
File.Create(...);
```

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str_str.cs
```diff
@@ -3,503 +3,171 @@
...
-            if (Array.IndexOf(strArr, "TestDir2") < 0)
-            {
-                iCountErrors++;
-                printerr("Error_24gy7! Incorrect name==" + strArr[1]);
-            }
+    [Fact]
+    public void ValidSearchPattern()
+    {
+        String strTempDir = "a..b abc..d";
+        GetEntries(TestDirectory, strTempDir); //Should not throw
```
stephentoub: Should this also Assert.Empty?

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str_str.cs
```diff
@@ -3,503 +3,171 @@
...
-                iCountErrors++;
-                printerr("Error_28gyb! Incorrect number of files");
-            }
-            fs6.Dispose();
-
-            //if(Directory.Exists(dirName))
-            //	Directory.Delete(dirName, true);
+                Assert.Contains(Path.Combine(TestDirectory, "aaabbcc"), strArr);
+            else
+                Assert.DoesNotContain(Path.Combine(TestDirectory, "aaabbcc"), strArr);
```
stephentoub: I think this else case is going to fail on OS X, which like Windows does case-insensitive comparisons by default.  Unfortunately we don't run these tests yet in CI on OS X (we do on Windows and Linux).  Can you test locally?  The condition may need to be updated to have OS X go down the if block.

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str_str_so.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    public override String[] GetEntries(String dirName, String searchPattern)
+    {
+        return Directory.GetFileSystemEntries(dirName, searchPattern, SearchOption.TopDirectoryOnly);
+    }
+
+    public virtual String[] GetEntries(String dirName, String searchPattern, SearchOption option)
+    {
+        return Directory.GetFileSystemEntries(dirName, searchPattern, option);
+    }
```
stephentoub: Your structuring of these tests is clever. :smile:

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str.cs
ianhays: That's interesting! Since that is the case, I'll need to go through the tests again and fix this, as whitespace file/directory name checks are in essentially every test.

stephentoub: If the test is passing now, I'd suggest leaving the tests as you have them, opening an issue to track the potential product bug, and adding a new test that's PlatformSpecific.AnyUnix that validates spaces can be used in this manner... you'll need to mark it as ActiveIssue with the issue you opened, as it would currently be failing.  Then after you get these tests merged, you can go back and investigate/fix the problem, and update the then failing tests accordingly.  Just a suggestion.

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str_str.cs
ianhays: I'll need to wrangle myself up a mac; I don't currently have one available to use. 

The case sensitivity check was frequently brought up in the tests, so again I'll need to go through them and pull out the OS-specific parts. I'll make a new issue for that and mark it Xplat and work on it as I move forward with redoing the worst of the FileSystem tests.

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str.cs
ianhays: Will do!

src/System.IO.FileSystem/tests/Directory/GetFiles_str_Str.cs
```diff
@@ -11,454 +12,274 @@
...
+    }
 
-            strLoc = "Loc_4yg7b";
+    [Fact]
+    public void InvalidSearchPattern()
+    {
+        // '<' and '>' are valid Unix filename chars
+        if (Interop.IsWindows)
+        {
+            Assert.Throws<ArgumentException>(() => GetFiles(".", "<>"));
```
stephentoub: I think elsewhere where we're checking invalid characters, we have a Unix check as well, using '\0' instead of '>'.  Should we do the same here?

ianhays:
Fixing the temp file/directories issue greatly furthered my desire to redo these FileSystem tests. It seemed better to keep this PR small (since it's my first to the dotnet repo) and fix the issue linked for now before anything else though. I'm going to continue restructuring the test classes to be more like GetFiles and GetFileSystemEntries after this PR goes through, as they desperately need it.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_Str.cs
```diff
@@ -11,454 +12,274 @@
...
-            catch (Exception exc)
-            {
-                iCountErrors++;
-                printerr("Err_24y7g! Incorrect exception thrown, exc==" + exc.ToString());
-            }
+    [Fact]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public void CharacterTests()
+    {
+        // TODO: [ActiveIssue(846)] Globalization on Unix
```
stephentoub: Now that we actually have the ActiveIssue on this, the comment for the same thing can be removed.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_Str.cs
```diff
@@ -11,454 +12,274 @@
...
-                    iCountErrors++;
-                    printerr("Err_960tli! Incorrect exception thrown, exc==" + exc.ToString());
-                }
-            }
+        for (int i = 160; i < 256; i++)
+        {
+            GetFiles(".", ((Char)i).ToString());
+        }
 
 #if DESKTOP
```
stephentoub: Do we ever set this?

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
```diff
@@ -8,642 +8,321 @@
...
-                        if (Eval(list.Contains(files[i]), "Err_582bmw! unexpected file found: {0}", files[i]))
-                            list.Remove(files[i]);
-                    }
-                    if (!Eval(list.Count == 0, "Err_891vut! {0} expected files not found.", list.Count))
-                    {
-                        foreach (String fileName in list)
-                            Console.WriteLine(fileName);
-                    }
-                }
+                if (Eval(list.Contains(files[i]), "Err_287kkm! unexpected file found: {0}", files[i]))
```
stephentoub: With these tests having been overhauled, do we still need these identifiers (e.g. "Err_415mbz!")?

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
```diff
@@ -8,642 +8,321 @@
...
             }
+        }
+    }
 
-            // Path is not in the current directory (same drive)
-            //Scenario 4: Ensure that the path contains subdirectories and call this API and ensure that only the top directory files are returned
-            /* Scenario disabled when porting because it modifies the filesystem outside of the test's working directory
-            try
+    [Fact]
+    [ActiveIssue(0, PlatformID.Any)]
```
stephentoub: Do we know why this was previously commented out?  If it's just a bad test, now would be a good time to remove it.  If it has value and needs fixing, it'd be good to fix it.  If that's too much work for now, we could tag it with a real issue number rather than 0.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
```diff
@@ -8,642 +8,321 @@
...
-                    }
-                }
+    [Fact]
+    public void Scenario7()
+    {
+        //Scenario 7: dir is readonly
+        String dirName;
+        String[] expectedFiles;
+        String[] files;
+        List<String> list;
```
stephentoub: These variables don't all need to be declared at the top of the function (same goes for other tests where this is being done).

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
```diff
@@ -8,642 +8,321 @@
...
-        }
-
-        Assert.True(s_pass);
     }
 
+    #endregion
+
     //Checks for error
-    private static bool Eval(bool expression, String msg, params Object[] values)
+    protected static bool Eval(bool expression, String msg, params Object[] values)
```
stephentoub: Are these Eval methods just legacy from before we had Assert.Equal?  If so, could we remove them and just use Assert directly?

src/System.IO.FileSystem/tests/File/Copy_str_str_b.cs
```diff
@@ -290,31 +289,31 @@ public static void runTest()
...
+                        {
+                            Console.WriteLine(fil1.FullName);
+                            iCountErrors++;
+                            printerr("Error_298gc! Incorrect fullname set during copy");
+                        }
+                        new FileInfo(filName).Delete();
+                        fil1.Delete();
+                        File.Delete("\\TestFile.tmp");
+            #endif
+            */
```
stephentoub: Should this be deleted, or fixed?

src/System.IO.FileSystem/tests/Directory/GetFiles_str_Str.cs
ianhays: Yes, it is happening below in the switch statement - it goes to default and throws an ArgumentException.

Side question, do you know what the justification is for "GetFiles(TestDirectory,'\n'.ToString())" to return the empty array but "GetFiles(Directory.GetCurrentDirectory(), String.Format("te{0}st", '\n'ToString()))" to throw an exception? The distinction of allowing newlines by themselves but not in a name seems a bit inconsistent, unless there's a use case I'm missing.

stephentoub:
Overall, nice cleanup!  The tests are much clearer now, and it's always nice to see so much code gook removed :smile:

In addition to verifying that everything passes on Windows, Linux, and OSX, can you also do a code coverage run before and after to help verify we've not regressed in that regard?  I spot checked in a few places, and it seemed like you did a good job keeping the existing validation, but it'd be good to double check.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_Str.cs
stephentoub: > The distinction of allowing newlines by themselves but not in a name seems a bit inconsistent, unless there's a use case I'm missing.

I don't think it's that newlines alone are allowed and disallowed otherwise, but that various whitespace charactes (including '\n') are trimmed from the end of the search string (I don't know what the justification is for such trimming, but it's been the behavior for a long time).  So  "\n" and "something\n" are both valid, but "some\nthing" is not: the string is trimmed and then checked for invalid characters.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
ianhays: I think it's a holdover; the comment says the purpose is to "Ensure that the path contains subdirectories and call this API and ensure that only the top directory files are returned" but when it gets files it uses SearchOption.AllDirectories, which is directly contradicting the point of the test. I don't think it has value since I check subdirectory exclusion/inclusion elsewhere.

src/System.IO.FileSystem/tests/Directory/GetFiles_str_str_so.cs
ianhays: If they do have some value, I couldn't find it :) They should be assertions. They're holdovers from before xunit was being used for testing.

src/System.IO.FileSystem/tests/File/Copy_str_str_b.cs
ianhays: I haven't looked at the File classes yet beyond fixing the temporary files issue, but at a glance it looks like it should be removed. I think the code formatter made my changes in this section.

ianhays:
Thanks @stephentoub! There's still so much to be done in these FileSystem classes! It was hard to leave some of that stuff as is - it hurt my inner designer - but I wanted to get a PR out to fix the persistent temp files issue ASAP. I ran a coverage report after my major changes and didn't have any regression in coverage, but I didn't do one before PR submission so I'll check again.

My next steps are:
* Get a mac and test on it
* Finish restructuring and splitting the tests into separate [Fact] decorated methods
* Remove calls to the unnecessary "Eval" methods
* Get rid of all the debugging strings
* Replace calls to "CheckErrors" with Asserts
* Refactor the "Scenario#" tests and determine which should stay/be removed/moved elsewhere
* Remove calls to Console.WriteLine
* Either modify the CommonUtils.ManageFileSystem class or remove the calls to it. As of now it uses random file names and directory structures which really should be changed to something repeatable and structured.

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str.cs
ianhays: #2205

stephentoub:
db, test this please

(failed due to git repo cloning issues earlier in the day)

stephentoub:
@dotnet-bot test this please

stephentoub:
Looks like there's a file issue that's causing compilation to fail:
```
error CS2001: Source file 'DirectoryInfo\DirectoryInfo_EnumerableAPIs.cs' could not be found [d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.IO.FileSystem\tests\System.IO.FileSystem.Tests.csproj]
```

ianhays:
I see. It's failing because I renamed two of the files to match the format of all of the other FileSystem test files. 

I renamed File_EnumerableAPIs.cs to EnumerableAPIs.cs in the "File" test folder as well as DirectoryInfo_EnumerableAPIs.cs to EnumerableAPIs.cs in the "DirectoryInfo" test folder.

ianhays:
PR updated! All tests for the modified files are passing, and it should be ready to merge now. Any additional feedback before I go through with the changes?

stephentoub:
Thanks, @ianhays.  I left a few more comments.  Most are nits, though the .csproj one should probably be addressed.  Once you've done that, LGTM.  Could you squash before merging?

ianhays:
Squashed it down to two commits: one from before PR feedback and one from after PR feedback. I also removed the unnecessary additions to the csproj. I'll go ahead and merge it after the Jenkins build just to be safe.

Issue #2190, opened 2015-06-29
Move System.IO.Path to corefx


stephentoub:
The Path.cs file currently in the corefx repo is only used for .NET Native; System.IO.Path currently comes from the version built into mscorlib.  The Path exposed from System.Runtime.Extensions.dll should have its code in System.Runtime.Extensions rather than being type forwarded.

stephentoub:
Fixed by #2234 and #2248.

PR #2201, opened 2015-06-30
Add more tests for System.IO.Path

stephentoub:
Path has a bunch of corner cases it tries to cover, especially in normalization.  This brings code coverage up to > 90% to hit many more of those cases.  Part of #2190.

### DIFF
--- a/src/System.Runtime.Extensions/tests/System/IO/Path.cs
+++ b/src/System.Runtime.Extensions/tests/System/IO/Path.cs
@@ -4,42 +4,58 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
 using Xunit;
 
 public static class PathTests
 {
-    [Fact]
-    public static void ChangeExtension()
+    [Theory]
+    [InlineData(null, null, null)]
+    [InlineData(null, null, "exe")]
+    [InlineData("", "", "")]
+    [InlineData("file", "file.exe", null)]
+    [InlineData("file.", "file.exe", "")]
+    [InlineData("file.exe", "file", "exe")]
+    [InlineData("file.exe", "file", ".exe")]
+    [InlineData("file.exe", "file.txt", "exe")]
+    [InlineData("file.exe", "file.txt", ".exe")]
+    [InlineData("file.txt.exe", "file.txt.bin", "exe")]
+    [InlineData("dir/file.exe", "dir/file.t", "exe")]
+    [InlineData("dir/file.t", "dir/file.exe", "t")]
+    [InlineData("dir/file.exe", "dir/file", "exe")]
+    public static void ChangeExtension(string expected, string path, string newExtension)
     {
-        Assert.Null(Path.ChangeExtension(null, null));
-        Assert.Equal("file", Path.ChangeExtension("file.exe", null));
-        Assert.Equal("file.exe", Path.ChangeExtension("file.txt", "exe"));
+        if (expected != null)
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @ericstj, @JeremyKuhne, @ellismg

stephentoub:
db, test this please

(timed out trying to clone the repo)

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -127,12 +198,104 @@ public static void GetTempFileName()
...
+            Assert.Equal(@"\\server\share", Path.GetFullPath(@"\\server\share\.."));
+            Assert.Equal(@"\\server\share\", Path.GetFullPath(@"\\server\share\    "));
+            Assert.Equal(@"\\server\  share\", Path.GetFullPath(@"\\server\  share\"));
+            Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\"));
+            Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\server"));
+            Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\server\"));
+            Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\server\.."));
+            Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\?\GLOBALROOT\"));
+
+            // Windows short paths
```
JeremyKuhne: Short path generation may or may not be turned on for the drive running tests. Is there anything we want to do to allow for that here? (I *think* it may be on for "C:" and off for others in recent OS versions- but that is from limited observation and possibly a side effect of my usage of Storage Spaces.)

stephentoub: Thanks, Jeremy.  Since I'm already doing P/Invokes, is there a Win32 function I can P/Invoke to that will tell me whether short path generation has been disabled?  Or will GetShortPathName just fail in that case?  If the latter, rather than asserting that it returns > 0, I could just change this to conditionally execute these tests if it succeeds and skip them on failure.

stephentoub:
@dotnet-bot, test this please

src/System.Runtime.Extensions/tests/System/IO/Path.cs
JeremyKuhne: Sorry, took a few to find this. It is unclear whether drive specific settings are trumped by overall disable (registry). Here is what I know of:

_Short path policy reg key:_
https://technet.microsoft.com/en-us/library/cc778996.aspx

_Querying for short path availability on a drive:_
https://msdn.microsoft.com/en-us/library/windows/hardware/ff545496.aspx

fsutil has the ability to set/query this stuff. It is something we want to figure out as we can probably save quite a bit of time on the normalization paths by skipping expansion in GetFullPath().

ellismg:
Could you consider some test cases with invalid characters in them?  Embedded nulls would be particularly interesting, I think.

stephentoub:
> Could you consider some test cases with invalid characters in them? Embedded nulls would be particularly interesting, I think.

Good suggestion. I've now added a test to the one for GetPathInvalidChars, checking all of the Path methods against each of the invalid chars to ensure each results in an exception.  GetPathInvalidChars includes '\0' on all platforms.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
stephentoub: Thanks.  I updated the test to check the result of GetShortPathName.

JeremyKuhne:
Good progress- lgtm

PR #2202, opened 2015-06-30
ConcurrentBag performance improvements and new functions

MarkusSintonen:
-Improved ToList() performance by initializing the list with initial capacity
-Use initial capacity for bag version lists
-Added Clear function
-Added Contains/Find/FindAll/Exists/TrueForAll functions that avoid List
allocations. Functions represent moment-in-time state of the bag
-Added GetEnumerableSlim to avoid list allocation. Returned enumerable
does not represent moment-in-time state of the bag (which can be used when
moment-in-time representation is not enforced)

New functions can be used to avoid the list allocations which is
problematic with very large bags.

### DIFF
--- a/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
+++ b/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentBag.cs
@@ -48,10 +48,6 @@ namespace System.Collections.Concurrent
         // GlobalListsLock lock
         private bool _needSync;
 
-        // Approximate item count inside the bag. This is used to determine initial capacity for list copying operations.
-        // Note that this might not represent totally correct item count inside the bag.
-        private volatile int _approxCount;
-
         // Count of the thread lists. This is used to determine initial capacity for the version lists
         private volatile int _threadListCount;
 
@@ -91,18 +87,14 @@ namespace System.Collections.Concurrent
             _locals = new ThreadLocal<ThreadLocalList>();
 
             // Copy the collection to the bag
-            int count = 0;
             if (collection != null)
             {
                 ThreadLocalList list = GetThreadList(true);
                 foreach (T item in collection)
                 {
                     list.Add(item, false);
-                    count++;
                 }
             }
-
-            _approxCount = count;
         }
 
         /// <summary>
@@ -141,9 +133,6 @@ namespace System.Collections.Concurrent
                     Monitor.Enter(list, ref lockTaken);
                 }
                 list.Add(item, lockTaken);
-#pragma warning disable 0420
-                Interlocked.Increment(ref _approxCount);
-#pragma warning restore 0420
             }
             finally
             {
@@ -244,10 +233,6 @@ namespace System.Collections.Concurrent
                         }
                     }
                     list.Remove(out result);
-#pragma warning disable 0420
-                    Interlocked.Decrement(ref _approxCount);
-#pragma warning restore 0420
-                    return true;
                 }
                 else
                 {
@@ -406,12 +391,6 @@ namespace System.Collections.Concurrent
                 if (CanSteal(list))
                 {
                     list.Steal(out result, take);
-                    if (take)
-                    {
-#pragma warning disable 0420
-                        Interlocked.Decrement(ref _approxCount);
-#pragma warning restore 0420
-                    }
                     return true;
                 }
                 result = default(T);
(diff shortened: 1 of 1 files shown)


stephentoub:
@MarkusSintonen, thanks for the PR.  On the whole I'm happy to see PRs that try to improve performance (cc: @KrzysztofCwalina).  But I do have several concerns about this change.

1. This is adding new surface area.  Purely from a process perspective, we should be following the API addition / review process documented at https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md . cc: @terrajobst 

2. This is adding additional interlocked operations on the most common code paths, in order to track an approximate count that can be used to potentially benefit other operations.  I'm not convinced this is a worthwhile tradeoff, given typical usage patterns which in my experience involve lots of threads adding/removing and then only later consulting the contents of the bag, often in a non-parallel region. (It's also worth noting that the "optimization" could in certain circumstances actually hurt the cases attempting to be optimized, by having an artificially higher count that results in a much-larger-than-necessary array allocation.)  If the goal is simply to get a count with which to size the list, you might explore/measure just enumerating twice, at which point potentially the list could be skipped entirely and an array used directly (there would be overhead to that as well, including various memory effects, so measuring lots of scenarios would be important either way).

3. The GetEnumeratorSlim method you added documents itself as being safe to enumerate concurrently with other operations.  I'm not convinced of that, either (and it's the basis for the optimization). Can you walk me through the logic with which you arrived at that conclusion, taking into account memory model issues and the like?  I can see the value in such an enumerator, so it's a worthwhile discussion to have about whether it should be added, but if it can't be done safely, we'd probably want to consider an alternative name highlighting its unsafeness, e.g. DangerousGetEnumerable, UnsafeGetEnumerable, etc.

Thanks.

KrzysztofCwalina:
I definitely think changes like this *must* be accompanied by good set of perf tests comparing before/after results.

stephentoub:
@MarkusSintonen, thanks again for your efforts to contribute.  Given my above comments, here's my suggestion:

1. We close this PR out, as I don't believe it's workable in its current form.
2. You open an issue suggesting the addition of ConcurrentBag.GetEnumeratorSlim (or whatever name it'd end up with), providing the suggested API signature and scenarios for why it'd be valuable, and highlighting any performance numbers you have to motivate it.  (Optionally you could include the other APIs you mention, though my personal feeling is that a non-freezing GetEnumerator represents the 90% case of what you're trying to address, with the other APIs buildable on top of it.)

The second one can then be evaluated both from an API standpoint and from a feasibility standpoint, e.g. someone will need to do the investigation to see whether it's thread-safe to implement.

MarkusSintonen:
Yes indeed interlocked increment/decrement caused slightly lower performance when adding/removing items (benchmarked). But ToList() performance was obviously improved, especially with large bag sizes. I removed the counter variable from the code. Other possibility would be to determine the initial capacity inside ToList() by checking the Count from each of ThreadLocalLists but this would require looping through all the ThreadLocalLists every time ToList() is called. This would be beneficial with large bags but would deteriorate bag enumeration performance on smaller bags.

GetEnumerableSlim(/Unsafe/Whatever) is safe to use concurrently since _headList pointing to the first thread owner linked list is marked as volatile. Also linked list _head is marked as volatile as well as the linked list _next reference and _value. Therefore all the threads are always seeing the newest values which we are enumerating. It is also certain that linked list node _value is always initialized and wont be modified during GetEnumerableSlim or any other operation.

It can be noted through that GetEnumerableSlim might not represent the true moment in time state of the bag as the function summary states. This is obvious since while some thread is enumerating the bag, some other thread might call the new Clear() function. After this the enumerating thread would still see some already deleted values from the bag. But this effect is stated in the function summary (this enumerable does not necessarily represeting the true momement in time state).

I also removed some of the API additions and added more versatile ForEach() function for thread safe enumeration of the bag.

Some of these might be valuable additions since ToList() copying is slow operation and situation is even worse with large bag sizes. Previously there was no way to avoid this and it would kill the performance where it was necessary to enumerate the bag contents.  I will provide some bag content enumeration benchmarks later on.

There might be still some possibility to avoid the list allocations safely when returning the enumerable as other concurrent collections have also avoided it (dictionary/queue/stack).

stephentoub:
Thanks for the update, @MarkusSintonen.  A few things:
- As I noted in my previous response, we have a process around adding new public API surface area: https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md.  Including such changes in a PR like this without going through that process means we won't be able to merge this PR, even if there are other changes in it that we would want to take.  If you'd like additional API surface area like GetEnumerableSlim to be considered, please open an Issue so that it can be discussed; this PR can then be trimmed down to just the performance improvements you're hoping to see made to the existing functionality.
- As @KrzysztofCwalina, it's important to understand the performance tradeoffs associated with a change like this.  I think there's a good chance your thread count tracking change here is worthwhile, as it'll likely save some allocations on each steal operation (or at least the size of those allocations).  However, it's not without its costs.  For example, this could increase the size of the ConcurrentBag itself, depending on architecture.  It's possible (though probably not common) for the explicit sizing of the list to actually make allocations worse, if it's sized to a value that's then immediately invalidated by another thread joining.  It's adding additional volatile writes when each thread joins/departs.  Etc.  I could definitely believe that your changes would be a net win, but it's important to have more concrete data to substantiate that before we take a change like this.
Thanks.

hackcraft:
> This is adding additional interlocked operations on the most common code paths, in order to track an approximate count that can be used to potentially benefit other operations.

I have code at https://bitbucket.org/JonHanna/ariadne/src/e4e6580d40f58c48feea8ff90436afce745f5975/Ariadne/Counter.cs?at=master for a striped counter that reduces the impact of both the interlocking and also the potential for false-sharing on highly-concurrent uses of a counter. The EUPL license currently on it isn't appropriate for this project, but I'd be happy to re-license that class if useful.

It doesn't avoid races on it's `Value` reader, because it's intended to be used heuristically in concurrent collections, so its use-case tolerates inaccuracies. Besides, a perfectly accurate count would be stale the moment you read it anyway.

If your tolerance for inaccuracy is high, you can even try the non-atomic increments and decrements. Obviously inaccuracies will happen, but if the tolerance for such inaccuracy is high enough it can be a nett win to do so.

MarkusSintonen:
Thank you for your comments @stephentoub I'll post new issue to the issue tracker about these additions.

Counter by @hackcraft could be very beneficial in many scenarios for example this where there is a possibly of list copyings (which might be problematic with GetEnumerator) and approximation of the initial list capacity could be formed.

stephentoub:
> Thank you for your comments @stephentoub I'll post new issue to the issue tracker about these additions.

Should we close out this PR now, or were you going to do additional follow-ups on it in addition to the issues you opened?

stephentoub:
Associated issues for reference:
https://github.com/dotnet/corefx/issues/2335
https://github.com/dotnet/corefx/issues/2338

PR #2212, opened 2015-07-01
Fix TaskExtensions.Unwrap stack dive

### LINKED ISSUE #2208: Stack overflow in TaskExtensions.Unwrap
When running the following test:
```c#
        [Fact]
        public static void RunStackGuardTests()
        {
            const int DIVE_DEPTH = 12000;

            // Test stack guard with Unwrap
            {
                Func<long, Task<long>> func = null;
                func = iterationsRemaining =>
                {
                    --iterationsRemaining;
                    return iterationsRemaining > 0 ?
                        Task.Factory.StartNew(() => func(iterationsRemaining)).Unwrap() :
                        Task.FromResult(iterationsRemaining);
                };
                func(DIVE_DEPTH).Wait();
            }

            // These tests will have stack overflowed if they failed.
        }
```

fails because Stack Overflow occurs.

Related to PR https://github.com/dotnet/corefx/pull/1589

stephentoub:
In strange but possible task configurations, it's possible to set up a series of Unwrap calls such that the continuations involved will form a very long chain of synchronously-executing calls.  In such cases, without extra measure, the implementation can overflow the stack.  The mscorlib Unwrap implementation takes precautions to avoid this, and ContinueWith (which was used in the previous, slower, standalone Unwrap implementation) has checks to avoid this as well, but currently await does not.  Since our new Unwrap implementation uses the await infrastructure, it's susceptible again.

This commit adds a check to the new Unwrap implementation to avoid the stack dive, as well as porting a test that triggers the stack dive prior to the fix.  Some local benchmarks show little-to-no impact from the additional checks, and it's very rare to hit the fallback path that triggers the async queueing: while the actual numbers will depend on the actual code being run, the size of stack frames, etc., in the test we only hit the fallback case ~8 times out of 12000 iterations, and that's in code trying very hard to force this condition.

Fixes #2208.

### DIFF
--- a/src/System.Threading.Tasks/src/System/Threading/Tasks/TaskExtensions.CoreCLR.cs
+++ b/src/System.Threading.Tasks/src/System/Threading/Tasks/TaskExtensions.CoreCLR.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace System.Threading.Tasks
 {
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @AlfredoMS, @mellinoe

stephentoub:
@dotnet-bot test this please

mellinoe:
LGTM :+1:

PR #2215, opened 2015-07-01
Bring build warnings to 0

### LINKED ISSUE #1520: Build warnings due to SafeFileHandle conflict
A bunch of warnings like the following are now generated when doing a Linux/OSX build in corefx:
```
\corefx\src\Common\src\Microsoft\Win32\SafeHandles\SafeFileHandle.Unix.cs(26,25): 
warning CS0436: The type 'SafeFileHandle' in '\corefx\src/Common\src\Microsoft\Win32\SafeHandles\SafeFileHandle.Unix.cs' conflicts with the imported type 'SafeFileHandle' in 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'. 
Using the type defined in '\corefx\src/Common\src\Microsoft\Win32\SafeHandles\SafeFileHandle.Unix.cs'. 
[\corefx\src\System.Diagnostics.Debug\src\System.Diagnostics.Debug.csproj]
```

stephentoub:
Fixes a variety of build warnings, bringing us down to 0 warnings and 0 errors (at this point we should be able to enable warnings-as-errors so that we don't regress further):

- System.Diagnostics.Debug warnings.  On Unix builds, SafeFileHandle is built into the partial facade.  Since the code in the assembly also uses SafeFileHandle and because the partial facade references mscorlib, we get a bunch of warnings about type conflicts between the SafeFileHandle in mscorlib and the one in Debug (the one in Debug correctly wins).  I've suppressed the warning in the .csproj.

- System.Private.Uri warnings.  Same as System.Diagnostics.Debug, as it internally builds some of the Debug code into the assembly.

- System.Security.Principal.Windows.Tests warnings.  There are two issues here.  One is that the test .csproj has a P2P reference to the src project, but also includes the corresponding NuGet package in the project.json; I've removed it from the project.json.  The other has to do with a mismatch in referenced System.Diagnostics.Debug version numbers; I've fixed this in the test project's project.json.

Fixes #1520.

### DIFF
--- a/src/System.Diagnostics.Debug/src/System.Diagnostics.Debug.csproj
+++ b/src/System.Diagnostics.Debug/src/System.Diagnostics.Debug.csproj
@@ -11,6 +11,7 @@
     <DefineConstants>$(DefineConstants);PUBLIC_DEBUG</DefineConstants>
     <IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>
     <NuGetTargetFrameworkMoniker>DNXCore,Version=v5.0</NuGetTargetFrameworkMoniker>
+    <NoWarn>0436</NoWarn> <!-- Suppress warnings for type conflicts between SafeFileHandle in partial facade and mscorlib -->
   </PropertyGroup>
   <PropertyGroup Condition=" '$(TargetsUnix)' == 'true' ">
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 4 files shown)


stephentoub:
cc: @weshaggard, @mmitche, @mellinoe

mellinoe:
LGTM

PR #2216, opened 2015-07-01
Enable warnings-as-errors for the build

### LINKED ISSUE #1160: Turn on warnings-as-errors for CoreFX

stephentoub:
I also explicitly enabled WarningLevel 4 for all projects, and fixed a few warnings that fire with the VS2013 compiler but not with the VS2015 compiler.

Fixes #1160.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -186,6 +186,12 @@
     <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
   </PropertyGroup>
 
+  <!-- Set up handling of build warnings -->
+  <PropertyGroup>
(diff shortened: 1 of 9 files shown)


stephentoub:
cc: @weshaggard, @mmitche

stephentoub:
With #2215 merged, this is now passing as well.

ellismg:
LGTM.  Really excited to see this land.

weshaggard:
LGTM @stephentoub thanks for taking care of this. I actually think this is worth going into one of the common targets in buildtools but I'm OK with it only in corefx for the time being.

PR #2218, opened 2015-07-01
Relax regex in build.sh to also recognize Mono 4.0.2

akoeplinger:
Before, the script complained cause it only looked for 4.0.1.x. This change makes it possible to compile with Mono 4.0.2 as well.

/cc @josteink


stephentoub:
LGTM

cc: @JeremyKuhne

JeremyKuhne:
LGTM- thanks

PR #2219, opened 2015-07-01
IO: Add GetActualCasing to File and Directory

jasonwilliams200OK:
Uses kernel32 interoped method `GetLongPathName` to determine the
actual casing of file and directory paths on disk as well as network.

Issue-URL: #1086.

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Directory.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Directory.cs
@@ -115,7 +115,41 @@ namespace System.IO
         {
             return File.GetCreationTimeUtc(path);
         }
- 
+
+        // Gets actual casing of path. The result is string with exact path
+        // as disk or network.
+        //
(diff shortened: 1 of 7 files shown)


jasonwilliams200OK:
@stephentoub, 
Disclaimer: I did all my experiments with this console application https://gist.github.com/jasonwilliams200OK/a10c5c92c1de96f2cc2a and then ported the code to corefx. This is an experiment. I am open for suggestions. Please review and let me know if there is something terribly wrong with this approach. :smile:

ellismg:
We should follow the [API Review Process](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md) for this change.  #1086 should be updated with a Speclet and then we should take the addition through API Review.

If we do end up adding this, the work should be done against the "future" branch, instead of master.

jasonwilliams200OK:
@ellismg, alright. Lets follow the review process. :)

PR #2220, opened 2015-07-01
[WIP] Refactor HttpClient and MultipartContent with async/await

Maxwe11:


Maxwe11:
@dotnet-bot test this please

stephentoub:
@Maxwe11, thanks for taking a stab at this.  My initial reaction is that this doesn't go nearly far enough, as evidenced by the TaskCompletionSources that are still being passed around and that this hasn't really changed the structure of the code, but rather simply done localized changes to replace ContinueWiths in a method with awaits in the same method.  My expectation is that doing this robustly will result in a significant rewrite, and that very few (if any TaskCompletionSources) will be left at the end across the codebase.

Maxwe11:
@stephentoub thanks for the initial feedback, I've updated PR. Now only a couple TaskCompletionSources left.

src/System.Net.Http/src/System/Net/Http/HttpClient.cs
```diff
@@ -157,44 +157,128 @@ public Task<Stream> GetStreamAsync(string requestUri)
...
                 }
-                HttpResponseMessage response = requestTask.Result;
-                if (response.Content == null)
+                else
                 {
-                    tcs.TrySetResult(defaultValue);
-                    return;
+                    // Otherwise option is HttpCompletionOption.ResponseHeadersRead.
+                    // This option used by GetStreamAsync so we expecting that generic argument is Stream.
+                    Debug.Assert(typeof(T) == typeof(Stream), "We expection Stream as generic argument at this branch");
```
Maxwe11: Will fix typo

Maxwe11:
@dotnet-bot test this please

Issue #1277, opened 2015-03-28
Incremental Build gets longer while doing less work


eatdrinksleepcode:
Ingesting the latest build tools brought in dotnet/buildtools#116, which skips running tests when nothing has changed since the last successful run. This means that a fully incremental build (a build after a successful build with no changes) is doing much less work than it was doing before; however, it is also taking a lot longer (about 50%) than before. This apparently has something to do with the diagnostic log output that the build writes to msbuild.log. Reducing the verbosity of that output from "diagnostic" to "normal" reduces the time for an incremental build to a few seconds, which was the intention of dotnet/buildtools#116 in the first place:

&nbsp;&nbsp;&nbsp; | Clean | Incremental
--- | ----- | -----------
Before #116 | 85 | 86
After #116 | 83 | 143
After #116 (diag -> normal) | 79 | 9
(all times in seconds)

I cannot fully explain why doing less work causes the diagnostic output to be written so much slower; my best guess is that some internal log batching mechanism in MSBuild is getting overloaded by the enormous amount of output in such a short time. My proposed solution is to reduce the default verbosity from "diagnostic" to "normal", and add the ability to set that verbosity.

nguerrera:
Fixed by #2225

PR #2225, opened 2015-07-02
Switch default build log from diag to default

nguerrera:
We were logging about 300 MB per incremental build and
the IO had a very noticeable impact on build time.

To get a diag log wen you actually need one:

   build.cmd /flp:v=diag

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -39,7 +39,7 @@ call :build %*
 goto :AfterBuild
 
 :build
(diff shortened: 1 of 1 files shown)


nguerrera:
cc @stephentoub @mmitche

stephentoub:
Good idea.  LGTM.

eatdrinksleepcode:
Fixes #1277

PR #2226, opened 2015-07-02
RSA: Read the exponent without LowLevelList<T>

justinvp:
Avoids unnecessary allocations and having to reverse the array, and removes the need for `LowLevelList<T>` in this library.

### DIFF
--- a/src/System.Security.Cryptography.RSA/src/System.Security.Cryptography.RSA.csproj
+++ b/src/System.Security.Cryptography.RSA/src/System.Security.Cryptography.RSA.csproj
@@ -38,9 +38,6 @@
     <Compile Include="System\Security\Cryptography\RSAParameters.cs" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetsWindows)' == 'true' ">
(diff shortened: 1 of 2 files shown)


justinvp:
cc: @bartonjs

bartonjs:
This does look valid to me; but I'd feel more comfortable with the endianness if you could find (or make) a certificate somewhere that doesn't use a palindrome for an exponent, and we could verify the Windows and OpenSSL implementations off of each other.

OTOH, 99% (handwavy number) of certificates use `{ 0x01, 0x00, 0x01 }` as their exponent value, so maybe it's moot.

@stephentoub I'm on the fence between "just check it in" and "harden the tests, please".  If you want to LGTM it, you can count me as having done so, too.

stephentoub:
If all of the tests we use are of such form, then that feels like a gap to me regardless of how the conversion is implemented, and it'd be good to augment our tests accordingly.  But I think that's an issue whether with the old or new implementation, and the new one does look correct, so I'm ok with it.  How about we go ahead with this and then address the test gap separately.

bartonjs:
Created issue #2228 to track the test gap.  I'll merge in the change. Thanks, @justinvp :smile:.

Issue #2183, opened 2015-06-28
Rename tests in System.Reflection.Emit to better names


Priya91:
The tests in this assembly are named as PosTest* and NegTest*, and does not provide context on what it's testing. There are ~700 tests in this assembly.

Maxwe11:
Fixed by #2230

PR #2230, opened 2015-07-02
Rename test methods in System.Reflection.Emit.Tests to describe the test meaning.

Priya91:
closes issue #2183 

Also in this PR:

1. Added verification for tests that didn't fully test the APIs.
2. Corrected some tests, which meant to test different test-case, but had the same logic as tested before.

cc @stephentoub @nguerrera @joshfree


src/System.Reflection.Emit/tests/ConstructorBuilder/ConstructorBuilderDefineParameter.cs
```diff
@@ -157,15 +157,16 @@ public void PosTest4()
         }
 
         [Fact]
-        public void PosTest5()
+        public void TestParameterWithZeroSequence()
         {
             CreateConstructorBuilder("PosTest5_Type1", new Type[] { }).DefineParameter(0, ParameterAttributes.None, "p");
```
stephentoub: Do we need to update strings like "PosTest5_Type1", too, or does it matter?

src/System.Reflection.Emit/tests/MethodBuilder/MethodBuilderDeclaringType.cs
```diff
@@ -36,51 +36,43 @@ public class MethodBuilderDeclaringType
...
             string methodName = null;
 
             methodName = TestLibrary.Generator.GetString(false, false, true, MinStringLength, MaxStringLength);
-            MethodBuilder builder = TestTypeBuilder.DefineMethod(methodName,
+            TypeBuilder typeBuilder = GetTestTypeBuilder();
+            MethodBuilder builder = typeBuilder.DefineMethod(methodName,
                 MethodAttributes.Public);
 
-            Assert.True(builder.DeclaringType.Equals(TestTypeBuilder));
+            Assert.True(builder.DeclaringType.Equals(typeBuilder));
```
stephentoub: Assert.Equals?  Or is this a special Equals method?

src/System.Reflection.Emit/tests/TypeBuilder/TypeBuilderDefineDefaultConstructor.cs
```diff
@@ -46,85 +46,85 @@ public TypeBuilder RetriveTestTypeBuilder(TypeAttributes typeAtt)
...
         [Fact]
-        public void PosTest1()
+        public void TestDefaultConstructor1()
         {
             ExePosTest(TypeAttributes.Public | TypeAttributes.Class, MethodAttributes.Public, BindingFlags.Public | BindingFlags.Instance);
         }
 
         [Fact]
-        public void PosTest2()
+        public void TestDefaultConstructor2()
```
stephentoub: Not a big deal, but should these be named based on their differences rather than based on a number?  Or should things like this move to become theories?  Or if we don't want to use a theory, just have a method that calls ExePosTest once for each of the combinations?

stephentoub:
Thanks for going through all of those.  I spot checked, and left a few comments, but overall it looks good.

src/System.Reflection.Emit/tests/ConstructorBuilder/ConstructorBuilderDefineParameter.cs
Priya91: It's just a random string, whether xyz or some * to test locally.. I think it doesn't matter.

src/System.Reflection.Emit/tests/TypeBuilder/TypeBuilderDefineDefaultConstructor.cs
Priya91: I thought by looking at the arguments passed over to ExePosTest it was evident that we are testing the different combo of MethodAttributes and BindingFlags. I couldn't think of better names.

nguerrera:
Nice. Thanks for doing this! LGTM.

src/System.Reflection.Emit/tests/TypeBuilder/TypeBuilderDefineDefaultConstructor.cs
stephentoub: Ok.

PR #2231, opened 2015-07-02
Implement RSA Sign/Verify on Unix

### LINKED ISSUE #1964: Implement RSA Sign/Verify on Unix
The Unix (Linux/OSX) builds of RSA have only provided support for Import/Export and Encrypt/Decrypt.

Signing was skipped due to other needs at the time, but we still need to write it.

bartonjs:
Fixes #1964.

Since the RSA contract refactoring hasn't come along yet, this still implements RSA publicly via RSACryptoServiceProvider.

It internally makes use of a temporary version of HashAlgorithmName, trying to set up for when the API changes come.


stephentoub:
LGTM

PR #2237, opened 2015-07-03
Enumerable improvements [WIP]

### LINKED ISSUE #2238: Linq loops could be tighter; causes bug in one obscure case, waste in others.

hackcraft:
Fixes #2238

Changes to System.Linq.Enumerable, focusing on reducing the size of loops and on short-circuiting when an existing test points to a short-circuit opportunity. Also adds tests.


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -12,8 +12,8 @@ public static class Enumerable
     {
         public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate)
         {
-            if (source == null) throw Error.ArgumentNull("source");
-            if (predicate == null) throw Error.ArgumentNull("predicate");
+            Error.CheckSourceNotNull(source);
+            Error.CheckPredicateNotNull(predicate);
```
stephentoub: I'm not sure this change to hide the check-and-throw in a method is an improvement.  I agree it helps clean up the code, but in the past I believe I've seen this have implications on inlining, so while this could be a small readability improvement, I think it's possible it could also be a small performance hit.  I could be remembering incorrectly, or maybe it's so small in this context that it doesn't matter, but it makes me hesitate.

Regardless, I'd prefer to see such a change separated out into its own PR so that it can be more easily evaluated on its own.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -678,55 +678,201 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
+                IEnumerator<TSource> sourceEnumerator = _source.GetEnumerator();
+                try
+                {
+                    for (int count = _count; count != 0; --count)
+                    {
+                        // It should be harmless to have the recipient call MoveNext again.
+                        // but we guard against strange behaviour from shortcuts in the enumerator
+                        // It should also be harmless to dispose the enumerator here if we've
+                        // exhausted it, but we guard against strange behaviour in the caller.
+                        if (!sourceEnumerator.MoveNext()) return new DeadEnumeratorWrapper(sourceEnumerator);
```
stephentoub: Isn't this changing both order of operations and exception behavior? Previously GetEnumerator and MoveNext wouldn't be called on the underlying enumerable/enumerator until MoveNext was called on the enumerator returned from the SkipIterator.  But with this change, it seems like those methods will be called during the call to GetEnumerator instead.  If the underlying GetEnumerator or MoveNext threw, those exceptions would now propagate out of GetEnumerator instead of out of MoveNext.  And if the caller did something in between the call to GetEnumerator and MoveNext, those would now happen after the calls to the underlying enumerator, not before.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -678,55 +678,201 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
+            private readonly IEnumerable<TSource> _source;
+            private readonly Func<TSource, bool> _predicate;
+            public SkipWhileIterator(IEnumerable<TSource> source, Func<TSource, bool> predicate)
+            {
+                _source = source;
+                _predicate = predicate;
+            }
+            public IEnumerator<TSource> GetEnumerator()
+            {
+                return new Enumerator(_source, _predicate);
```
stephentoub: Previously the compiler-generated iterator would be able to reuse the object generated for the SkipWhile enumerable as the iterator.  This is allocating a separate enumerator object.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1234,25 +1380,30 @@ public static TSource Last<TSource>(this IEnumerable<TSource> source)
...
-                    result = element;
-                    found = true;
+                    TSource result = e.Current;
+                    if (predicate(result))
+                    {
+                        while (e.MoveNext())
+                        {
+                            TSource element = e.Current;
+                            if (predicate(element)) result = element;
+                        }
```
stephentoub: Is this really better?  Yes, it saves writing one Boolean flag for each element that matches.  But it does so by duplicating the bulk of the function.  Is the difference really measurable?  I'd expect such a single Boolean write to a local to be dwarfed by the multiple interface calls and the delegate call each iteration, plus any time there's a write to the Boolean there's still a write to the local result.  Unless there's a real measurable difference, I'd prefer to stick with the simpler, more understandable foreach loop.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1318,29 +1469,29 @@ public static TSource Single<TSource>(this IEnumerable<TSource> source)
...
+                    }
                 }
             }
-            switch (count)
-            {
-                case 0: throw Error.NoMatch();
-                case 1: return result;
-            }
-            throw Error.MoreThanOneMatch();
+            throw Error.NoMatch();
```
stephentoub: With regards to the nested loops, I have the same question/comment about this one as on Last.

Separately, I agree with you that it's unfortunate the current implementation continues iterating the remainder of the list after it's found a second match.  Ideally it wouldn't have been written like that, and ideally we could fix it to not do so.  Changing it, however, is a potentially visible behavioral change.  Such changes need to be discussed to determine whether the compatibility risk is acceptable, and as such they should be done as separate PRs (or potentially an issue discussing it first) that can be evaluated independently.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1755,206 +1906,299 @@ public static decimal Sum<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
                 }
-                else
-                {
-                    value = x;
-                    hasValue = true;
-                }
             }
-            if (hasValue) return value;
-            throw Error.NoElements();
+            return value;
```
stephentoub: I do like this reimplementation of ```Min(IEnumerable<int>)```, and in particular that it avoids an extra branch on every iteration.  As I've noted elsewhere, there are pieces of this PR that I think are suspect and require more discussion and investigation, and others (like this) that I think could be worthwhile to review and take, so I'd suggest these be split up into smaller PRs.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1755,206 +1906,299 @@ public static decimal Sum<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
+                    if (cur.HasValue)
+                    {
+                        int x = cur.Value;
+                        if (x < valueVal)
+                        {
+                            valueVal = x;
+                            value = cur;
+                        }
+                    }
+                }
```
stephentoub: In contrast to the previous Min I commented on, this one is significantly more complex than the original implementation, and for what appears to be very little gain, primarily saving the write on each iteration to an int? local.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1755,206 +1906,299 @@ public static decimal Sum<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
                     // negative infinity.
-                    if (x < value || System.Single.IsNaN(x)) value = x;
-                }
-                else
-                {
-                    value = x;
-                    hasValue = true;
+                    // Not testing for NaN therefore isn't an option, but since we
+                    // can't find a smaller value, we can short-circuit.
+                    else if (float.IsNaN(x)) return x;
```
stephentoub: This will change behavior by potentially not enumerating the source to the end where it previously would have (and enumerating the source could have visible side effects).  Something to be discussed.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: (I'm assuming the rest of the Min/Max operators are the same basic transformation and so am not reviewing them, but my comments should likely apply to them equally.)

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2266,186 +2620,247 @@ public static decimal Max<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
                 {
-                    sum += v;
-                    count++;
+                    while (e.MoveNext())
+                    {
+                        sum += e.Current;
+                        ++count;
+                    }
                 }
+                return (double)sum / count;
```
stephentoub: Is this whole change just to avoid the if at the end of the function and one extra addition?  Is there a measurable cost to those?  This transformation doesn't add a ton of complexity, but it does add some, which makes me hesitate if the reward can't be quantified.

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
+﻿using System.Collections.Generic;
```
stephentoub: Thank you for adding more tests.  Could these be done as a separate PR to go in first?

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class MinMaxTests
+    {
+        [Fact]
+        public void MinInt()
+        {
+            var one = Enumerable.Range(1, 10).ToArray();
```
stephentoub: Nit: how about just writing it out like you've done for the other arrays?  It would help with consistency for someone reading along.

src/System.Linq/src/System/Linq/Enumerable.cs
JohanLarsson: How expensive is branching? Probably hard to answer.

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+            var minusTen = new[] { -1, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000, 100, 200, 1000 };
+            Assert.Equal(1, one.Min());
+            Assert.Equal(-10, minusTen.Min());
+            Assert.Equal(100, hundred.Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<int>().Min());
```
stephentoub: Can you add more boundary cases, e.g. where the array includes Int32.MinValue, where there's a single element in the array?  And could you add some tests on non-array sources?

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        [Fact]
+        public void MaxInt()
+        {
+            var ten = Enumerable.Range(1, 10).ToArray();
+            var minusTen = new[] { -100, -15, -50, -10 };
+            var thousand = new[] { -16, 0, 50, 100, 1000 };
+            Assert.Equal(10, ten.Max());
+            Assert.Equal(-10, minusTen.Max());
+            Assert.Equal(1000, thousand.Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<int>().Max());
```
stephentoub: Similar comments as for MinInt... I'll stop providing the same feedback on the rest of these, and will only note other things if I see any.

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+            Assert.Equal(1000F, thousand.Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<float>().Max());
+        }
+
+        [Fact]
+        public void MinDouble()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (double)i).ToArray();
+            var minusTen = new[] { -1D, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000D, 100, 200, 1000 };
```
stephentoub: The 'D' here on the first element is pretty sublte and easy to miss or even accidentally remove, at which point the test changes.  How about explicitly typing the array, i.e. ```new double[]``` instead of ```new []```?

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        public void MinNullableInt()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (int?)i).ToArray();
+            var minusTen = new[] { default(int?), -1, -10, 10, 200, 1000 };
+            var hundred = new[] { default(int?), 3000, 100, 200, 1000 };
+            Assert.Equal(1, one.Min());
+            Assert.Equal(-10, minusTen.Min());
+            Assert.Equal(100, hundred.Min());
+            Assert.Equal(null, Enumerable.Empty<int?>().Min());
+            Assert.Equal(null, Enumerable.Repeat(default(int?), 100).Min());
```
stephentoub: It'd be good to have cases where there are null and non-null values interspersed throughout, not just where the first element is null or where all elements are null.  Plus the same comments from earlier, e.g. testing a single-element array.

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        [Fact]
+        public void MinSingle()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (float)i).ToArray();
+            var minusTen = new[] { -1F, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000F, 100, 200, 1000 };
+            Assert.Equal(1L, one.Min());
+            Assert.Equal(-10L, minusTen.Min());
+            Assert.Equal(100L, hundred.Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<float>().Min());
```
stephentoub: There's special logic in the implementation to handle NaN.  There should be some corresponding tests.

stephentoub: (Nevermind, I see you do have them separated out below. Thanks.)

src/System.Linq/tests/SingleTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        // Warning: At time of writing these two tests, they will incorrectly run for about five millennia.
+        // They'll then probably throw OverflowException rather than the correct InvalidOperationException,
+        // but I haven't had time to confirm. Recommend to skip if using without fix applied in commit 3f17916
+
+        [Fact]
+        public void FailOnManyMatches()
+        {
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Range(0, int.MaxValue).Single(i => i > 0));
```
stephentoub: I would prefer to see this done differently.  I'd have a test like:
```C#
[Fact]
public void DelegateInvokedForAllMatches()
{
    int count = 0;
    int iters = 10;
    Assert.Equal(1, Enumerable.Range(1, iters).Single(i => { count++; return true; }));
    Assert.Equal(iters, count);
}
```
and I'd get that merged with the rest of the tests.  That helps verify the current behavior, which someone could be depending on (though that'd be unfortunate).  Then separately there can be a discussion about changing that behavior, and if/when a fix for it goes in, the test can be updated to reflect that count is now 2 instead of 10. Etc.

src/System.Linq/tests/SkipTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        [Fact]
+        public void SkipWhileHalf()
+        {
+            Assert.Equal(Enumerable.Range(10, 10), Enumerable.Range(0, 20).SkipWhile(i => i < 10));
+            Assert.Equal(Enumerable.Range(10, 10), Enumerable.Range(0, 20).SkipWhile((i, idx) => idx < 10));
+        }
+    }
+}
```
stephentoub: Seems like there should also be tests for when the skip count is 0, for when the input has 0 elements and when it has 1 element, and ideally for different kinds of sources.

stephentoub:
cc: @VSadov, @terrajobst

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: With small enough methods a reason for doing it can be precisely to encourage inlining (if the result is that the containing method drops below the jitter's threshold). I think you've a good point about it being in a different PR though. I shall revert this particular commit soon.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Yes. If keeping with this approach, the logic would have to move into `MoveNext` in which case the improvement on the code generated from `yield` is negligible if any, or it would have to wrap an exception-throwing version of the DeadEnumeratorWrapper (not all that arduous since there's already a try block).

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Understandable is in the eye of the beholder; I find this approach a better logical match to the task than the previous. It's probably the least beneficial change in this whole PR though, and largely included because it matches the style of a lot of the other changes.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Very hard to answer in general, because some of the branches are more likely to be mis-predicted than others.
Some of the changes to Min allow for a short-circuit, specifically those on floating-point values, which is both the biggest possible gain for some sequences in this whole PR (this whole thing was inspired by a bug I had a while back with how Single doesn't short-circuit) and the thing that could cause issues if people are expecting side-effects (if you are coding side-effects into your Linq use you are going to have a bad time, but that doesn't mean people won't have done so). I haven't added short-circuits where there isn't a test in the original, but it's still at least 5 cases of short-circuiting that are new. Perhaps whether or not to short-circuit needs to be discussed separately.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: I'd actually thought it would make things worse, as I've seen the JIT refuse to inline methods that contain ```throw```, and I thought it would refuse to inline these helpers because of that, thus we'd actually be increasing the number of function calls that couldn't be inlined.  But, I just checked and tried it out, and it appears at least the VS2015 x86 and x64 JITs are able to inline methods with throws in some cases, just not in others (e.g. https://github.com/dotnet/coreclr/blob/master/src/jit/importer.cpp#L12588-L12612), and these helpers are actually able to be inlined, at least in my experiments.  

In any event, thanks for separating them out.  Since we'd be changing such error checking paths, I'd also hope to see tests in place that would exercise all of those checks being changed.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: > if you are coding side-effects into your Linq use you are going to have a bad time, but that doesn't mean people won't have done so

Unfortunately people definitely do this.

> I haven't added short-circuits where there isn't a test in the original

We have very low coverage of LINQ in our tests right now. Please don't infer that a lack of a test for a particular behavior means that behavior can be changed (though it'd be nice if we lived in such a world :smile:).

> Perhaps whether or not to short-circuit needs to be discussed separately.

Yes :smile:   I think it'd be best to separate out anything that could result in a functional change in behavior.  Otherwise, such things could very easily derail other suggested improvements.

src/System.Linq/src/System/Linq/Enumerable.cs
AdamSpeight2008: @stephentoub The `Enumerator` is disposed of anyway because of the `using` block.

stephentoub: @AdamSpeight2008, yes, but that's not what I was referring to.  In the current implementation, the enumerator's MoveNext/Current will be invoked for every element, regardless of whether a NaN value is found.  With this change, if a NaN is found, the new behavior will stop processing immediately, such that MoveNext/Current will not be invoked for elements beyond the NaN.  That can be a visible change in behavior, e.g. if an expression earlier in the query was mutating every element or was updating some shared state.

AdamSpeight2008: @stephentoub 
 `Any`, `All` also don't iterate through the entire source.

stephentoub: @AdamSpeight2008, I understand, I'm not talking about what the ideal behavior would be (I agree the ideal behavior would be to short-circuit whenever possible).  The issue is one of compatibility and introducing breaking changes.  Any/All already short-circuit, whereas this would be introducing new short-circuiting.

src/System.Linq/src/System/Linq/Enumerable.cs
AdamSpeight2008: @JohanLarsson
 The difference is that the PR reduces the size of the IL for this method by 10 instructions. Same for `Max`

src/System.Linq/src/System/Linq/Enumerable.cs
AdamSpeight2008: @stephentoub It breaks early, potentially useful if the source sequence is infinite in length.

stephentoub: @AdamSpeight2008, as I noted in response to your other comments, I understand the value in breaking early like this.  My point is that it's a change in behavior from the current implementation, and thus could break existing code.

hackcraft: I think I shall soon create a PR for a request that solely introduces short-circuits where appropriate, so that this issue can be discussed in isolation. Is that the best way to do that, in terms of procedures here?

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: The change seems observable and will break code relying on delayed execution of GetEnumerator and MoveNext. It is not hard to imagine such code to exist. 
However inconvenient, the former order of sideeffects needs to be preserved.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -679,35 +680,155 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
+                    _started = true;
+                    while (_source.MoveNext())
+                        if (!_predicate(_source.Current))
+                            return true;
+                    return false;
+                }
+                public void Reset()
+                {
+                    _source.Reset();
+                    _started = false;
```
VSadov: I do not think former version supported Reset. It is better to not introduce it because in will or will not work depending on underlying source which would be inconsistent and hard to program against. More so, once it is supported it will be hard to take back if future optimizations cannot support Reset. 
Generally Linq iterators are not supporting reset. So it is better to not expand the feature surface here.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -715,18 +836,62 @@ public static IEnumerable<TSource> SkipWhile<TSource>(this IEnumerable<TSource>
...
+                private int _index = 0;
+                public Enumerator(IEnumerable<TSource> source, Func<TSource, int, bool> predicate)
+                {
+                    _source = source.GetEnumerator();
+                    _predicate = predicate;
+                }
+                public TSource Current
+                {
+                    get { return _source.Current; }
+                }
```
VSadov: You are calling _source.Current twice - in MoveNext and in Current. There are several problems with that. 
Current is not supposed to be sideeffecting, but it can be, so the difference will be observable. Current could be expensive since former implementation called it just once per iteration. The change would have negative effect on such code.
A chain of iterators tat call Current twice will result in exponential complexity which will become expensive very quickly even if leaf "Current" is cheap.

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: Yes. Based on analysis done a while ago, most iterators are used exactly once and saving allocation of an extra object is a statistical win. That resulted in an optimization that reuses the enumerable object as a first iterator instance. We should not give up that optimization without a proof that it is not longer relevant.

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: Yes. I think the best course of action is to separate pure optimizations (not observable) into a separate PR and include time/allocation measurements for various scenarios to see that the performance is indeed better.
As for observable changes - some changes like the early-out in Single seems to be good ideas, but it would be easier if they are separate changes. Changes in behavior may need a broader discussion and implications of one could differ from another. It would be easier if those are separately reviewable changes.

VSadov:
As a general note, I think this would go better in the following order:
1) tests 
2) Unobservable optimizations - fewer branches, loop unrolling, strength reduction and the like.
3) Changes that come with behavior differences like short-circuiting or different strategy of calling to source methods. What makes these different is that in addition to perf impact, there is some degree of breakage risk which is harder to measure or estimate.

stephentoub:
@hackcraft, what do you think of @VSadov's suggested plan of attack here?  I think it'd make sense to close out this PR and submit PRs instead for each of his numbered items in the order he recommends (potentially splitting into further PRs if deemed appropriate).  Hopefully 1 and 2 would go through without much issue, and then we can have the more difficult conversations around 3.

hackcraft:
I was actually planning to submit 3 PRs in parallel for those three cases, but won't look at this until the weekend.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: > > I haven't added short-circuits where there isn't a test in the original

> We have very low coverage of LINQ in our tests right now. Please don't infer that a lack of a test for a particular behavior means that behavior can be changed (though it'd be nice if we lived in such a world :smile:).

Sorry. What I meant wasn't *test* as in *test-case* but *test* as in the running code tests for a condition, and branches accordingly. We could for example add `if (value == int.MinValue) return value;` into `Min<int>()` and have a short-circuit that benefits sequences containing `int.MinValue` at the cost of adding extra work to those that don't and is probably not worth it. `Min<double>()` on the other hand already has to test for `double.IsNaN(value)` so making that a short-circuit condition doesn't add a test to the loop.

src/System.Linq/src/System/Linq/Enumerable.cs
James-Ko: Just a minor note here, but if we know that the nullable `HasValue` then we can get a [performance gain](http://ericlippert.com/2012/12/20/nullable-micro-optimizations-part-one/) by changing `int?.Value` to `int?.GetValueOrDefault()`.

hackcraft:
> I was actually planning to submit 3 PRs in parallel for those three cases, but won't look at this until the weekend.

And of course, that's a silly idea, because doing them in parallel means either committing code without tests, which is backwards, or else branching the branches for the other two PRs of the branch in #2326 which has its own issues.

Let's look at the tests, as per #2326. Then I'll split off the other two aspects of this, and close this PR.

stephentoub:
@hackcraft, can this be closed now?  Has everything you wanted to save from it been separated out?

hackcraft:
@stephentoub There's still a bunch of stuff that I'm separating out from it a batch at a time (expect Min and Max optimisations with tests on performance improvement this weekend) but it can still be closed in the meantime.

stephentoub:
Thanks, @hackcraft.

PR #2240, opened 2015-07-03
Fix OrderByDescending if comparer returns int.MinValue

### LINKED ISSUE #2239: OrderByDescending sorts incorrectly if a comparer returns int.MinValue
`int.MinValue` is of course traditionally considered less than 0. It is therefore a valid "x is less than y" return value for `IComparer<T>.Compare(T x, T y)`.

But if an implementation returns it when it is correct, then `Enumerable.OrderByDescending` will treat it indicating that `y` is greater than `x`.

Demonstrating test case:

    private class ExtremeComparer : IComparer<int>
    {
        public int Compare(int x, int y)
        {
            if (x == y)
                return 0;
            if (x < y)
                return int.MinValue;
            return int.MaxValue;
        }
    }
    [Fact]
    public void OrderByExtremeComparer()
    {
        var outOfOrder = new[] { 7, 1, 0, 9, 3, 5, 4, 2, 8, 6 };
        Assert.Equal(Enumerable.Range(0, 10), outOfOrder.OrderBy(i => i, new ExtremeComparer()));
        Assert.Equal(Enumerable.Range(0, 10).Reverse(), outOfOrder.OrderByDescending(i => i, new ExtremeComparer()));
    }

The second assertion fails, as rather than the expected `9, 8, 7, 6, 5, 4, 3, 2, 1, 0`, the result is `3, 5, 1, 4, 9, 2, 0, 8, 7, 6`.

hackcraft:
int.MinValue is a valid "x is less than y" result for a comparer but
OrderByDescending treats it incorrectly. Fix this.

Fix #2239

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -3105,7 +3105,10 @@ namespace System.Linq
                 if (next == null) return index1 - index2;
                 return next.CompareKeys(index1, index2);
             }
-            return descending ? -c : c;
+            // -c will result in a negative value for int.MinValue (-int.MinValue == int.MinValue).
+            // Flipping keys earlier is more likely to trigger something strange in a comparer,
+            // particularly as it comes to the sort being stable.
+            return (descending != (c > 0)) ? 1 : -1;
         }
     }
 
--- a/src/System.Linq/tests/EnumerableTests.cs
+++ b/src/System.Linq/tests/EnumerableTests.cs
@@ -194,6 +194,26 @@ namespace System.Linq.Tests
             Assert.Equal(-10, minusTen.Max());
             Assert.Equal(1000, thousand.Max());
         }
+
+        private class ExtremeComparer : IComparer<int>
+        {
+            public int Compare(int x, int y)
+            {
+                if (x == y)
+                    return 0;
+                if (x < y)
+                    return int.MinValue;
+                return int.MaxValue;
+            }
+        }
+
+        [Fact]
+        public void OrderByExtremeComparer()
+        {
+            var outOfOrder = new[] { 7, 1, 0, 9, 3, 5, 4, 2, 8, 6 };
+            Assert.Equal(Enumerable.Range(0, 10), outOfOrder.OrderBy(i => i, new ExtremeComparer()));
+            Assert.Equal(Enumerable.Range(0, 10).Reverse(), outOfOrder.OrderByDescending(i => i, new ExtremeComparer()));
+        }
     }
 }
 


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3105,7 +3105,11 @@ internal override int CompareKeys(int index1, int index2)
                 if (next == null) return index1 - index2;
                 return next.CompareKeys(index1, index2);
             }
-            return descending ? -c : c;
+            // -c will result in a negative value for int.MaxValue (-int.MinValue == int.MinValue).
+            // Flipping keys earlier is more likely to trigger something strange in a comparer,
+            // particularly as it comes to the sort being stable.
+            // 0.CompareTo(c) is always negative for a positive c and positive for a negative.
+            return descending ? 0.CompareTo(c) : c;
```
stephentoub: Would it be clearer (not needing the last line of the comment) and potentially slightly faster (though hopefully the backend compiler will do the right thing) to have:
```C#
return descending ? (c < 0 ? 1 : -1) : c;
```
?

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: At the time of coding I preferred the fact that `0.CompareTo(c)` would also do the right thing if `c` was `0`. That shouldn't happen though, and your suggestion has one less branch (I'd expect the jitter to do the right thing in inlining `CompareTo` but not necessarily in removing the branch that considers the possibility of `c == 0`). I shall change that line to your suggestion.

Clockwork-Muse:
PLINQ also had this problem.  I've added the relevant tests and fix for it in my outstanding PR ( #2133 )

hackcraft:
Yes. I've made that change, including reducing the comment to enough to prevent someone coming along and saying "hey, it'd be more efficient to return -c". Another approach is to flip the order, which is what Clockwork-Muse did in PR #2133. My concern there was we might expose a bug in a comparer that had depended on order; that would be such a bad thing for a comparer to do that I wouldn't worry too much about it, but if it can be considered then so much the better.

Clockwork-Muse:
@hackcraft - I debated about that, but didn't really consider it something to worry about in my case:
 1. (Like you point out) Non-reflexive comparators explicitly break the contract in the first place.
 2. It's PLINQ, you've got no control over which is on the left and right anyways.

It was also such a wonderfully minimal change, it just appealed to me.

hackcraft:
@Clockwork-Muse Agreed; in the parallel case such a use of a broken comparator is just too broken to have a hope of working. It's probably too broken to have a hope in the non-parallel case, but I figure there's no harm in being cautious.

nguerrera:
LGTM. Thanks.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3105,7 +3105,10 @@ internal override int CompareKeys(int index1, int index2)
                 if (next == null) return index1 - index2;
                 return next.CompareKeys(index1, index2);
             }
-            return descending ? -c : c;
+            // -c will result in a negative value for int.MaxValue (-int.MinValue == int.MinValue).
+            // Flipping keys earlier is more likely to trigger something strange in a comparer,
+            // particularly as it comes to the sort being stable.
+            return descending ? (c < 0 ? 1 : -1) : c;
         }
```
VSadov: Can this be just    (descending ^ (c >= 0))  ?   1 :  -1;

stephentoub:
@VSadov, should we merge this as-is, or are you waiting for a response from @hackcraft on your suggestion?

hackcraft:
Sorry, I'm busy with other stuff so I won't be looking at corefx much for the next two days. I agree with VSadov though, and will update my PR in a minute.

stephentoub:
Thanks, @hackcraft; completely understand.

nguerrera:
I liked it much better with the parens as @VSadov had it. This clever expression makes me think enough as it is, and I don't want to be distracted by potential precedence implications.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3105,7 +3105,10 @@ internal override int CompareKeys(int index1, int index2)
                 if (next == null) return index1 - index2;
                 return next.CompareKeys(index1, index2);
             }
-            return descending ? -c : c;
+            // -c will result in a negative value for int.MaxValue (-int.MinValue == int.MinValue).
+            // Flipping keys earlier is more likely to trigger something strange in a comparer,
+            // particularly as it comes to the sort being stable.
+            return (descending ^ (c > 0)) ? 1 : -1;
```
sharwell: :exclamation: Use `!=` instead of `^`.

:memo: Some people understand `^` for Boolean values. Everyone understands `!=` for Boolean values. Functionally they are equivalent.

stephentoub:
LGTM. @hackcraft, could you squash this?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3105,7 +3105,10 @@ internal override int CompareKeys(int index1, int index2)
                 if (next == null) return index1 - index2;
                 return next.CompareKeys(index1, index2);
             }
-            return descending ? -c : c;
+            // -c will result in a negative value for int.MaxValue (-int.MinValue == int.MinValue).
```
sharwell: :bulb: This statement is true, but I believe you meant to reference `int.MinValue` where you currently have `int.MaxValue`.

hackcraft:
> LGTM. @hackcraft, could you squash this?

YWIMC

VSadov:
I like " != " more than ^ as well. Easier to read.

Issue #2244, opened 2015-07-04
MetadataReader.GetNamespaceDefinition is crashing on framework assemblies


jkotas:
Repro:
```
var peReader = new PEReader(File.OpenRead(@"Microsoft.CSharp.dll"));
var reader = peReader.GetMetadataReader();
var root = reader.GetNamespaceDefinitionRoot();
```

Result:
```
System.ArgumentException: An item with the same key has already been added. at
System.Collections.Generic.Dictionary<System.Reflection.Metadata.NamespaceDefinitionHandle, System.__Canon>.Add(System.Reflection.Metadata.NamespaceDefinitionHandle key, System.__Canon value)
System.Reflection.Metadata.Ecma335.NamespaceCache.PopulateNamespaceTable()
```

This bug makes the namespace feature of metadata reader useless.

jkotas:
The crash is caused by same suffix shared by multiple namespaces, like:

```
namespace FxResources.Microsoft.CSharp
{
    public class Resources
    { }
}

namespace Microsoft.CSharp
{
    public class Program
    { }
}
```

I have shared the test update to include this case at https://github.com/jkotas/corefx/commit/7ed83663ef0662a29f21b5d6ea4e6c35b14d45a6

jkotas:
@tmat @nguerrera Thoughts on how to fix it?

nguerrera:
I'll take a look.

tmat:
Oops. We should really get Roslyn to use the reader's namespace API instead of its own namespace tree builder. We'd likely catch this earlier.

nguerrera:
Yes. I have a potential fix #2245.

PR #2245, opened 2015-07-04
 Fix issue with multiple namespaces having same suffix 

nguerrera:
For so-called synthetic namespaces (which have no types of their own,
only child namespaces), we were using the string heap offset to its
dot-terminated name as a unique handle value. However, compilers are
smart in the case where namespace overlap such as
FxResources.Microsoft.CSharp and Microsoft.CSharp and re-use the same
bytes in the string heap. As such, the value is not unique.

It turns out that we don't actually retrieve the simple name from the
handle so the fix is to use an auto-incrementing counter instead of the
simple name offset as the handle value.


nguerrera:
Fixes #2244 

cc @jkotas @tmat

src/System.Reflection.Metadata/tests/Metadata/HandleTests.cs
```diff
@@ -541,7 +541,7 @@ public void NamespaceKinds()
             Assert.False(full.IsVirtual);
             Assert.Equal(123, full.GetHeapOffset());
 
-            var synthetic = NamespaceDefinitionHandle.FromSimpleNameOffset(123);
+            var synthetic = NamespaceDefinitionHandle.FromSyntheticId(123);
```
nguerrera: @tmat do you think we should remove the Synthetic designation  bit and make these use the virtual bit. It feels more like what we use virtual for elsewhere now.

stephentoub:
LGTM

src/System.Reflection.Metadata/tests/Metadata/HandleTests.cs
tmat: @nguerrera Making it virtual makes sense.

nguerrera: OK. I'll do it.

nguerrera:
@tmat Moved to virtual bit as discussed. See 3176e3f for delta. I can squash before committing if folks want.

nguerrera:
@tmat Moved to virtual bit as discussed. See 1d838d3 for delta. I can squash before committing if folks want.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Handles.cs
```diff
@@ -2713,61 +2713,66 @@ public override int GetHashCode()
...
         {
-            return new NamespaceDefinitionHandle(NamespaceHandleType.SyntheticNamespace | (uint)stringHeapOffset);
+            // we arbitrarily disallow 0 virtual index to simplify nil check.
+            Debug.Assert(virtualIndex != 0); 
+
+            if (!HeapHandleType.IsValidHeapOffset(virtualIndex))
+            {
+                // only a pathological assembly would hit this, but id must fit in 29 bits.
+                ThrowTooManyVirtualNamespaces();
+            }
```
nguerrera: Prior conversation hiding behind rebase:
@tmat: 
> I wouldn't surface the notion of virtual namespaces to the user thru an error message (although it's almost impossible to see that message in reality).

@nguerrera 
>> Struggled with this. Felt it should be there as acknowledgement of possibility/completeness with other checks around heap offsets. Better message idea? Maybe just throw new BadImageFormatException() due to extreme unlikeliness?

@tmat 
> Can it actually happen?

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Internal/MetadataFlags.cs
```diff
@@ -230,9 +205,10 @@ internal static class HandleType
...
         internal const uint Guid = 0x72;        // #Guid heap
+        internal const uint Namespace = 0x73;   // #String heap but known to be the full name of a namespace
 
-        // #String heap and its modifications (up to 8 string kinds, virtaul and non-virtual)
-        internal const uint String = 0x78;
+        // #String heap and its modifications (up to 8 string kinds, virtual and non-virtual)
+        internal const uint String = 0x7c;
         internal const uint DotTerminatedString = String | ((StringHandleType.DotTerminatedString & ~HeapHandleType.VirtualBit) >> HeapHandleType.OffsetBitCount);
         internal const uint ReservedString1 = String | ((StringHandleType.ReservedString1 & ~HeapHandleType.VirtualBit) >> HeapHandleType.OffsetBitCount); 
         internal const uint ReservedString2 = String | ((StringHandleType.ReservedString1 & ~HeapHandleType.VirtualBit) >> HeapHandleType.OffsetBitCount);
```
nguerrera: From prior conversation hiding behind rebase:

@tmat:
> This should be  StringHandleType.ReservedString2

nguerrera:
@tmat fixed the remaining issues and squashed. Mind taking a final look?

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Handles.cs
```diff
@@ -2713,61 +2713,66 @@ public override int GetHashCode()
...
+            if (!HeapHandleType.IsValidHeapOffset(virtualIndex))
+            {
+                // only a pathological assembly would hit this, but it must fit in 29 bits.
+                ThrowTooManySubnamespaces();
+            }
+
+            return new NamespaceDefinitionHandle(TokenTypeIds.VirtualBit | virtualIndex);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
stephentoub: Just curious, is this because FromVirtualIndex will get inlined if it doesn't inline ThrowTooManySubnamespaces, but if the throw does get inlined, then FromVirtualIndex won't be?

nguerrera: Yes. Either that or FromVirtualIndex gets inlined but uses unnecessary space in its caller. A similar pattern applied to BlobReader made a very measurable difference. With that said, this case is not very highly travelled and it probably doesn't matter here.  I find it easier to keep the whole library consistent than to decide on each throw on a case-by-case basis. 

I've filed #2261 to re-organize so that the approach is centralized and documented.

stephentoub: Thanks. As part of #2261, should we be standardizing such a thing across corefx, or is that unwarranted?

stephentoub:
LGTM.

tmat:
:+1:

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Handles.cs
nguerrera: Possibly, but maybe that might be too aggressive. I'm not sure. Perhaps we could do it on demand as a real need is observed within a library.

We also have something a little similar in mscorlib: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/ThrowHelper.cs

PR #2247, opened 2015-07-04
Fix FileSystem enumeration with broken symlinks

stephentoub:
Today when enumerating directories, we look only for entries with DT_DIR (directory) or DT_REG (regular file); if we encounter something unknown or a symlink, we stat to resolve it and determine whether it's actually an S_IFDIR (directory) or S_IFREG (regular file). There are (at least) two problems with this:

1) The stat to a symlink will fail if the symlink points to a non-existent entry.
2) We end up skipping lots of special files, e.g. FIFOs.

The commit changes our logic to simply special case directories: we treat everything as a file unless we're sure it's a directory.  This directly fixes (2), and indirectly fixes (1): our logic for DirectoryExists already ignores any stat errors, returning false if the symlink couldn't be resolved, such that this naturally maps to our desired behavior of broken symlinks being categorized as a file.

Fixes https://github.com/aspnet/dnx/issues/2184

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
@@ -318,7 +318,7 @@ namespace System.IO
             return DirectoryExists(fullPath, out errno);
         }
 
-        private bool DirectoryExists(string fullPath, out int errno)
+        private static bool DirectoryExists(string fullPath, out int errno)
         {
             return FileExists(fullPath, Interop.libcoreclr.FileTypes.S_IFDIR, out errno);
         }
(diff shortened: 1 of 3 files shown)


stephentoub:
cc: @ellismg, @davidfowl, @ianhays 

Ian, I added a few tests that are going to conflict with your overhauling of the FileSystem tests.  If you plan to land your changes in the next couple days, I'll hold off on merging this and handle the conflicts.  Otherwise, it'll probably make sense for you to handle the conflicts.

stephentoub:
I fixed up the conflicts with #2200.

stephentoub:
@dotnet-bot test this please

ellismg:
LGTM, assuming green tests.

davidfowl:
:shipit:

PR #2248, opened 2015-07-06
Implement System.IO.Path on Unix

### LINKED ISSUE #2205: Creating files with whitespace names on Linux throws an exception
Creating files with whitespace names on Linux throws an exception when it should allow the operation. The following lines each throw an exception:
```
File.Create(Path.Combine(TestDirectory, "\n"));       //throws
File.Create(Path.Combine(TestDirectory, " "));        //throws
File.Create(Path.Combine(TestDirectory, "      "));   //throws
```
The stack trace for a unit test:
```
      System.UnauthorizedAccessException : Access to the path '/home/ianha/kcorefx/Directory_GetFileSystemEntries_str/' is denied.
      Stack Trace:
            at Interop.CheckIo(Int64 result, String path, Boolean isDirectory, Func`2 errorRewriter)
            at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
            at System.IO.UnixFileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
            at System.IO.UnixFileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
            at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
            at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
            at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync)
            at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
            at System.IO.File.Create(String path, Int32 bufferSize)
            at System.IO.File.Create(String path)
            at Directory_GetFileSystemEntries_str.UnixFileNameWithSpaces()
```
And of course, creating a file in the same directory with a spaceless name <I>not</I> throw an exception:
```
File.Create(Path.Combine(TestDirectory, "cat")); //this is fine
```

stephentoub:
Previously, System.IO.Path was implemented in mscorlib (with calls into Win32 on Windows / the Win32 PAL layer in libcoreclr on Unix) and type-forwarded from System.Runtime.Extensions.  With #2234, we switched the Windows build to using System.IO.Path from System.Runtime.Extensions and Path.cs from corefx, but the Unix version was left still using the mscorlib implementation.  One of the reasons for switching to the Path.cs in corefx, however, was to better enable a bunch of fixes to the behavior of the Unix implementation.

This change implements Path.cs in corefx for Unix and switches the Unix builds (Linux, OS X, and FreeBSD) over to using the implementation in corefx rather than type-forwarding to mscorlib.

Primary changes:
- Defined MaxPath / MaxComponent length in the Unix build in terms of pathconf
- Implemented the Unix build's NormalizePath to properly handle normalizing unnecessary "/", ".", and ".."s in the path.
- Implemented the Unix build's GetTempPath in terms of the TMPDIR environment variable (same logic that's in the Win32 PAL layer's implementation).
- Implemented the Unix build's GetTempFileName in terms of mkstemps.
- Implemented the Unix build's GetRandomFileName in terms of OpenSSL's RAND_pseudo_bytes, same as is done in System.Security.Cryptography.RandomNumberGenerator (which we can't reference due to layering).
- Centralized the wrapping of getcwd, which is now used from both System.Runtime.Extensions and System.IO.FileSystem.  The previous usage of it always created a temporary array of size MaxPath, which can now be quite big, so the wrapper first tries to use stack space, and then graduates to retrying with larger and larger buffer sizes.
- Centralized the MaxPath and MaxName logic, which is now used by three libraries: System.Diagnostics.Process, System.IO.FileSystem, and now System.Runtime.Extensions.
- Centralized the throwing of PathTooLongException, as it now requires arguments for how long are the maximum path and component names.
- Cleaned up a few other minor things in the shared Path.cs (e.g. naming Boolean arguments) that were confusing me as I was working through the code.
- Added FreeBSD to the .sln and .csproj files, as otherwise the FreeBSD build for this library would be broken (previously it inherited the mscorlib implementation as did the other Unix OSes).
- Added a bunch of tests, some general across both Windows and Unix, but others specifically focused on Unix, which is more lenient than is Windows in the paths it allows.

Fixes #2205, #2190


stephentoub:
cc: @ellismg, @sokket, @ianhays, @bartonjs, @JeremyKuhne 

Note that this PR will currently fail the build due to some bogus tests in System.IO.FileSystem that are assuming some paths are invalid when they're actually valid on Unix.  Once Ian's overhauling of the FileSystem tests gets merged, I'll fix up the offending tests and include a commit for that in this PR.

@bartonjs, note that I had to change the type of the exception thrown when we fail to initialize OpenSSL, as CryptographicException isn't available at this level in the stack.  I don't think this is a big deal, but let me know if you disagree.  There's a comment explaining this in the source.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -10,7 +10,57 @@ internal static partial class Interop
...
+            byte* stackBuf = stackalloc byte[StackLimit];
+            string result = getcwd(stackBuf, StackLimit);
+            if (result != null)
+                return result;
+
+            // If that was too small, try increasing large buffer sizes
+            // until we get one that works or until we hit MaxPath.
+            int maxPath = Interop.libc.MaxPath;
+            int bufferSize = StackLimit;
+            while ((bufferSize = Math.Min(bufferSize * 2, maxPath)) < maxPath)
```
bartonjs: This looks to me like it never tries with maxPath.

Assuming maxPath is 4096 the loop will do { 512 => try, 1024 => try, 2048 => try, 4096 => abort, not < maxPath }.

If someone had rounded their system off to 5000 characters, you'd try the 4096, then the next iteration should try 5000, but won't.

```C#
do
{
    bufferSize = Math.Min(bufferSize * 2, maxPath);

    ...
} while (bufferSize < maxPath)
```

perhaps?  It does one extra call in the degenerate case where maxPath <= StackLimit, which could be guarded against if it's really a concern.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -10,7 +10,57 @@ internal static partial class Interop
...
+        internal static unsafe string getcwd()
+        {
+            const int StackLimit = 256;
+
+            // First try to get the path into a buffer on the stack
+            byte* stackBuf = stackalloc byte[StackLimit];
+            string result = getcwd(stackBuf, StackLimit);
+            if (result != null)
+                return result;
+
```
bartonjs: Should you check errno == ERANGE here, before trying with a larger buffer?

bartonjs: Nevermind, you do that in the other overload of getcwd :smile:.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
stephentoub: Oops, you're right, will fix.

bartonjs: Also, note that your queried maxPath value can be -1 in situations where a system describes itself as having no inherent limit.  I don't know that it's anything that would really be encountered in the wild, but the manpage describes it that way.

stephentoub: > note that your queried maxPath value can be -1

It actually won't be here, due to https://github.com/dotnet/corefx/pull/2248/files#diff-57471f013b586ed190c3bbce817edd1cR39

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -10,7 +10,57 @@ internal static partial class Interop
...
+            // If that was too small, try increasing large buffer sizes
+            // until we get one that works or until we hit MaxPath.
+            int maxPath = Interop.libc.MaxPath;
+            int bufferSize = StackLimit;
+            while ((bufferSize = Math.Min(bufferSize * 2, maxPath)) < maxPath)
+            {
+                byte[] buf = new byte[bufferSize];
```
bartonjs: I'm wondering if we should try using `getcwd(NULL, 0)` before doing this loop.  Both the Ubuntu (glibc) and OSX manpages say that they provide the Right Answer when that is called (you'd just need to `free()` the result after making a managed string).

The advantage of that plan is no loop on our normal supported systems.  The disadvantages are that you need a SafeMallocHandle to track the return value, and you have a fallback that's unreachable on our officially supported systems.

Just offering it for consideration.

stephentoub: I'd considered that, but was scared off by the note that "POSIX.1-2001 leaves the behavior of getcwd() unspecified if buf is NULL."  The Linux, OS X, and FreeBSD man pages all imply that NULL is supported, but still.  I'm going to leave this as-is for now.  If it becomes an issue, we can easily update to it.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
bartonjs: Ah, I see; if the system says it has no limit we impose one.  While that makes me slightly uneasy, it does make it such that the consumer can reason about their code :smile:.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -10,7 +10,57 @@ internal static partial class Interop
...
+            }
+
+            // If we couldn't get the cwd with a MaxPath-sized buffer, something's wrong.
+            throw Interop.GetExceptionForIoErrno(Interop.Errors.ENAMETOOLONG);
+        }
+
+        private static unsafe string getcwd(byte* ptr, int bufferSize)
+        {
+            // Call the real getcwd
+            byte* result = getcwd(ptr, (IntPtr)bufferSize);
```
bartonjs: Since the method you're calling wants a size_t (which happens to be defined via a using to IntPtr), I'd cast to `size_t` here in the two places in this method that currently cast to `IntPtr`.

stephentoub: Yup, will do.

src/Common/src/Interop/Unix/libc/Interop.pathconf.cs
```diff
@@ -8,6 +8,21 @@ internal static partial class Interop
 {
     internal static partial class libc
     {
+        /// <summary>The maximum path length for the system.  -1 if it hasn't yet been initialized.</summary>
+        private static int _maxPath = -1;
```
bartonjs: s_maxPath / s_maxName

stephentoub: Yup.

src/System.Runtime.Extensions/src/Resources/Strings.resx
```diff
@@ -187,7 +190,7 @@
     <value>'{0}' must be greater than zero.</value>
   </data>
   <data name="IO_PathTooLong" xml:space="preserve">
-    <value>The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.</value>
+    <value>The specified path, file name, or both are too long. The fully-qualified file name must be less than {0} characters, and the directory name must be less than {1} characters.</value>
```
bartonjs: I think there are some hits on this that didn't get edited:
* src/Common/src/Interop/Unix/Interop.IOErrors.cs
* src/Common/src/System/IO/Win32Marshal.cs
* src/System.IO.FileSystem/src/System/IO/WinRTFileSystem.cs
* src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs

Not to mention that there are a half-dozen or so copies of this string across the various libraries.

stephentoub: Ugh, yup, I had my search set up in correctly.  Fixed.

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -69,15 +147,38 @@ private static string RemoveLongPathPrefix(string path)
...
-            // TODO: Implement this
-            throw new NotImplementedException();
-        }s
+            const string Suffix = ".tmp";
+            const int SuffixByteLength = 4;
+
+            // mkstemps takes a char* and overwrites the XXXXXX with six characters
+            // that'll result in a unique file name.
+            string template = GetTempPath() + "tmpXXXXXX" + Suffix + "\0";
+            byte[] name = Encoding.UTF8.GetBytes(template);
```
bartonjs: While UTF-8 and ASCII agree on 7-bit ASCII values, would this not be more technically correct as Encoding.ASCII?

bartonjs: (If Ubuntu and OSX do UTF-8 instead of Extended ASCII and I just missed the memo, that's fine.  I just don't know where to look to see if such a memo would exist :smile:)

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -61,22 +95,50 @@ public static void GetExtension(string expected, string path)
     [Fact]
     public static void GetFileName()
     {
+        Assert.Equal(null, Path.GetFileName(null));
```
bartonjs: This method (and many others) feels more suited to [Theory] than [Fact].

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
stephentoub: They do UTF8.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
stephentoub: Path.Combine can't be used in an attribute, and it didn't seem worth separating out those that use it.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -215,104 +314,187 @@ public static void GetTempFileName()
...
+        // "C:\..\..\..\.." => "C:\"
+        Assert.Equal(Path.GetPathRoot(curDir), Path.GetFullPath(Path.Combine(Path.GetPathRoot(curDir), "..", "..", "..", "..")));
+
+        // "C:\\\" => "C:\"
+        Assert.Equal(Path.GetPathRoot(curDir), Path.GetFullPath(Path.GetPathRoot(curDir) + new string(Path.DirectorySeparatorChar, 3)));
+
+        // Path longer than MaxPath that normalizes down to less than MaxPath
         var longPath = new StringBuilder(curDir);
-        for (int i = 0; i < 1000; i++)
+        for (int i = 0; i < 10000; i++)
```
bartonjs: This isn't necessarily big enough.  But we don't seem to expose MaxPath (probably because we've always considered it a const?), so this seems fine.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -215,104 +314,187 @@ public static void GetTempFileName()
...
 
-        // Try out a long path that normalizes down to less than MaxPath
+        // "C:\..\..\..\.." => "C:\"
+        Assert.Equal(Path.GetPathRoot(curDir), Path.GetFullPath(Path.Combine(Path.GetPathRoot(curDir), "..", "..", "..", "..")));
+
+        // "C:\\\" => "C:\"
+        Assert.Equal(Path.GetPathRoot(curDir), Path.GetFullPath(Path.GetPathRoot(curDir) + new string(Path.DirectorySeparatorChar, 3)));
+
+        // Path longer than MaxPath that normalizes down to less than MaxPath
         var longPath = new StringBuilder(curDir);
```
bartonjs: Pre-alloc the buffer size to handle all your DirectorySeparators?

src/System.Runtime.Extensions/tests/System/IO/Path.cs
stephentoub: Yeah, I just chose a randomly large value.  I contemplated querying pathconf, but decided against it for now.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
stephentoub: Sure.  I haven't spent time optimizing the tests, but in this case it's easy enough.

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -215,104 +314,187 @@ public static void GetTempFileName()
...
+        Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\server\.."));
+        Assert.Throws<ArgumentException>(() => Path.GetFullPath(@"\\?\GLOBALROOT\"));
+    }
+
+    [PlatformSpecific(PlatformID.Windows)]
+    [Fact]
+    public static void GetFullPath_Windows_83Paths()
+    {
+        string tempFilePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N") + ".txt");
+        File.Create(tempFilePath).Dispose();
```
bartonjs: Could you add a comment as to the objective here?

src/System.Runtime.Extensions/tests/System/IO/Path.cs
```diff
@@ -215,104 +314,187 @@ public static void GetTempFileName()
...
+                string invalidShortName = @"S:\DOESNT~1\USERNA~1.RED\LOCALS~1\Temp\bg3ylpzp";
+                Assert.Equal(invalidShortName, Path.GetFullPath(invalidShortName));
 
-                    // Same thing, but with a long path that normalizes down to a short enough one
-                    var shortLongName = new StringBuilder(invalidShortName);
-                    for (int i = 0; i < 1000; i++)
-                        shortLongName.Append(Path.DirectorySeparatorChar).Append('.');
-                    Assert.Equal(invalidShortName, Path.GetFullPath(shortLongName.ToString()));
+                // Same thing, but with a long path that normalizes down to a short enough one
+                var shortLongName = new StringBuilder(invalidShortName);
```
bartonjs: Pre-allocate the StringBuilder buffer?

src/System.Runtime.Extensions/tests/System/IO/Path.cs
bartonjs: If [InlineData] can handle params you could send it all through Path.Combine(params string[] paths); though that's a really inefficient-looking method when paths.Length == 1.

And there's always good old [MemberData].

Certainly not a critical issue, more of the food-for-thought category.

stephentoub: > you could send it all through Path.Combine(params string[] paths)

It also leads to complications, since some of the inputs are not destined to be part of that path but instead the expected output.

> And there's always good old [MemberData].

Sure, and I considered that, but in a case like this it makes things harder to understand (at least for me) rather than cleaning things up, and the benefit of having each test case isolated for a small number of asserts like this is relatively minimal.

I'm going to leave this as-is.  I have converted a few of the other tests into theories, though.

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
bartonjs: Ah, right.  It's based on whatever the user's LANG value is set to; which looks to be en_US.UTF-8 by default on Ubuntu and whatever account creation I did on OSX.

Are we concerned about Unix situations where someone has their LANG set to C or POSIX (both are ASCII)? Perhaps we need an interop helper somewhere for getting the right encoding depending on our launch locale (preferably by using locale capability detection, not reading those environment variables)?  Might depend on the work @ellismg is doing... or maybe we just document that we assume everyone is UTF-8 these days.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -10,7 +10,57 @@ internal static partial class Interop
...
+            byte* stackBuf = stackalloc byte[StackLimit];
+            string result = getcwd(stackBuf, StackLimit);
+            if (result != null)
+                return result;
+
+            // If that was too small, try increasing large buffer sizes
+            // until we get one that works or until we hit MaxPath.
+            long maxPath = Interop.libc.MaxPath;
+            int bufferSize = StackLimit;
+            while ((bufferSize = (int)Math.Min(bufferSize * 2, maxPath)) < maxPath + 1)
```
bartonjs: Because Math.Min is always going to restrict this down to maxPath, this now is an infinite loop instead of an exception.

stephentoub: Yup, I pushed too quickly; already working on fixing it.  Give me a minute before reviewing stuff.

src/System.IO.FileSystem/tests/Directory/GetDirectories_str_str_so.cs
```diff
@@ -610,21 +610,24 @@ public static void runTest()
...
-            }
-            catch (Exception ex)
-            {
-                s_pass = false;
-                Console.WriteLine("Bug927807! Exception caught in scenario: {0}", ex);
+                    string pathStr = root + "test" + Path.DirectorySeparatorChar + "test.cs";
+                    fullPath = Path.GetFullPath(" " + pathStr);
+                    Eval(fullPath, pathStr, "Wrong Full Path");
+                }
+                catch (Exception ex)
```
ellismg: I know this isn't you, but I really hate seeing stuff like this.  Do we have plans to clean this pattern of explicit catches and Assert.True(s_pass) soon?

stephentoub: @ianhays is working on it, at least in System.IO.FileSystem.  He already started fixing some of these with #2200.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
stephentoub: Ok, hopefully 3rd time's the charm. :smile:

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -43,23 +39,105 @@ private static bool IsDirectoryOrVolumeSeparator(char c)
 
         private static void EmulateFileIOPermissionChecks(string fullPath)
```
ellismg: Can we rename this method to make it more clear what's going on?  I would not expect "EmulateFileIOPermissionChecks" to be a no-op on Unix.  Alternately, pull GetFullPath into the PAL since that seems to be the only caller in the platform agnostic code?

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
ellismg: Is there a reason you have to take this approach instead of using StringBuilder?  Does the marshaller not correctly convert things to UTF-8 and back in that case?

bartonjs:
LGTM, though you should let @ellismg finish his pass, since he's started :smile:.

stephentoub:
Thanks for the review, @bartonjs.

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
stephentoub: Sure, I'll either rename it or move some things around.

PR #2256, opened 2015-07-06
Add tests which use an unusual RSA Exponent value (0x01B1).

### LINKED ISSUE #2228: Add an RSA test which uses a non-palindrome exponent value
A significant number of RSA keys use `{ 0x01, 0x00, 0x01 }` as their exponent value, with the second most popular value being `{ 0x03 }`.  These palindrome answers make it hard to verify that the Exponent value being returned has the proper endianness.

A test should be added which uses a non-palindrome exponent, preferably as a direct RSA test; but a certificate test might be reasonable.

bartonjs:
Generated an RSA key which used an Exponent value of 0x01B1, and verified that it roundtrips through Import/Export, and can be used successfully for encryption and decryption.

The key was generated by a private build of the Unix RSA library, which changed s_defaultExponent, and the cipherBytes value was generated on Windows using that key.  The two systems inter-operated beautifully.

Fixes #2228.


stephentoub:
LGTM. Thanks.

Issue #2184, opened 2015-06-28
Process.*ProcessorTime properties failing on OSX


stephentoub:
Attempting to get any of the Process' processor time properties on OSX is throwing an exception.

Priya91:
fixed with #2262

PR #2262, opened 2015-07-07
Fix Process failing tests on OSX.

Priya91:
proc_pid_rusage returns 0 on success, and -1 on failure. Removing throwing exception for result=0.
Removing debug assert on result, since it only stores 0/1 return code and not rusage_info_v3 struct size.

cc @stephentoub @sokket 

closes issue #2184

### DIFF
--- a/src/Common/src/Interop/OSX/Interop.libproc.cs
+++ b/src/Common/src/Interop/OSX/Interop.libproc.cs
@@ -546,12 +546,10 @@ internal static partial class Interop
 
             // Get the PIDs rusage info
             int result = proc_pid_rusage(pid, RUSAGE_SELF, &info);
-            if (result <= 0)
+            if (result < 0)
             {
                 throw new Win32Exception(SR.RUsageFailure);
             }
-
-            Debug.Assert(result == Marshal.SizeOf<rusage_info_v3>());
 
             return info;
         }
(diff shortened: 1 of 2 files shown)


stephentoub:
Thanks, Priya.  If this fixes the issue, there should be one or more [ActiveIssue]s removed as part of this, right?

Priya91:
Oh yup!, on Mac i've been running all tests without excluding failing ones, so missed removing the ActiveIssue tag. Will update the commit.

stephentoub:
Thanks. LGTM.

Unfortunately the existing test is a bit weak, in that it verifies that the various values are all >= 0, which on its own doesn't say much.  Something to be improved separately, but for now, were you able to verify that these values are now coming back as what you'd expect?

Priya91:
I was able to verify the totalprocessortime by comparing it with the CPU time in activity monitor on Mac. For the other 2 values, i am not sure how i can verify that, although the sum of the privileged and user time is the same value as totalprocessor time, in order of 0.3 sec difference, which is expected, as i make the property access one after the other. Is there another good way to measure the times to verify?

stephentoub:
Thanks, Priya.  I'm glad you were able to verify it manually, that approach seems reasonable, and it's great you were able to find and fix this bug so quickly.

PR #2264, opened 2015-07-07
Add CreateOpenSslCryptographicException helper method and use it.

### LINKED ISSUE #2211: Make all Unix CryptographicExceptions be produced through a helper method.
There are a couple of issues with what we currently do:

1. If OpenSSL didn't set an error but we do `new CryptographicException(Interop.libcrypto.GetOpenSslErrorString())` it will be Exception: Everything's OK.
2. If OpenSSL set an error but we didn't check, then the next (1)-like situation will report the wrong error.
3. OpenSSL has an error stack. If one function set more than one error, or there was a function whose error was ignored, we should fully drain the stack.

The helper method would be something akin to:
```C#
internal static void ThrowOpenSslCryptographicException()
{
    int error = ERR_get_error();
    int last = error;

    while (last != 0)
    {
        last = ERR_get_error();
    }

    throw error == 0 ?
        new CryptographicException() :
        new CryptographicException(GetOpenSslErrorString(error));
}
```

Though maybe having it return the Exception object so it is thrown from the right location?

bartonjs:
CreateOpenSslCryptographicException will drain out the error stack, using the first reported error for the exception.  If the error was an ERR_R_MALLOC_FAILURE code, use an OutOfMemoryException, otherwise a CryptographicException.

A pass was made over the code, and any throw statements which were preceded by a call into libcrypto use the new helper method.  Calls into NativeCrypto throw an unadorned CryptographicException.

This change also introduces CheckValidOpenSslHandle to make the handle checks easier on the eyes, and a few locations were noticed which did not check handle validity with the same thoroughness as the rest of the code.

Fixes #2211.

### DIFF
--- a/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
+++ b/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
@@ -32,19 +32,39 @@ internal static partial class Interop
             const uint ReasonMask = 0x0FFF;
             const uint ERR_R_MALLOC_FAILURE = 64 | 1;
 
+            // The Windows cryptography library reports error codes through
+            // Marshal.GetLastWin32Error, which has a single value when the
+            // function exits, last writer wins.
+            //
+            // OpenSSL maintains an error queue. Calls to ERR_get_error read
+            // values out of the queue in the order that ERR_set_error wrote
+            // them. Nothing enforces that a single call into an OpenSSL
+            // function will guarantee at-most one error being set.
+            //
+            // In order to maintain parity in how error flows look between the
+            // Windows code and the OpenSSL-calling code, drain the queue
+            // whenever an Exception is desired, and report the exception
+            // related to the last value in the queue.
             uint error = ERR_get_error();
-            uint last = error;
+            uint lastRead = error;
 
-            while (last != 0)
+            // 0 (there's no named constant) is only returned when the calls
+            // to ERR_get_error exceed the calls to ERR_set_error.
+            while (lastRead != 0)
             {
-                last = ERR_get_error();
+                error = lastRead;
+                lastRead = ERR_get_error();
             }
 
+            // If we're in an error flow which results in an Exception, but
+            // no calls to ERR_set_error were made, throw the unadorned
+            // CryptographicException.
             if (error == 0)
             {
                 return new CryptographicException();
             }
 
+            // Inline version of the ERR_GET_REASON macro.
             uint reason = error & ReasonMask;
 
             if (reason == ERR_R_MALLOC_FAILURE)
@@ -52,6 +72,8 @@ internal static partial class Interop
                 return new OutOfMemoryException();
             }
 
+            // If there was an error code, and it wasn't something handled specially,
+            // use the OpenSSL error string as the message to a CryptographicException.
             return new CryptographicException(ERR_error_string_n(error));
         }
 
--- a/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
+++ b/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
 using System.Text;
 
 internal static partial class Interop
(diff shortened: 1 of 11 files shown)


bartonjs:
Testing strategy: Private build of libcrypto where RSA_new() to sets RSAerr(RSA_F_RSA_NEW_METHOD, ERR_R_MALLOC_FAILURE) and returns NULL.

No change to the managed code:
```
   System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip [FAIL]
      System.Security.Cryptography.CryptographicException : error:0406A041:rsa routines:RSA_new_method:malloc failure
      Stack Trace:
            at Internal.Cryptography.RSAOpenSsl.CheckInvalidNewKey(SafeRsaHandle key)
            at Internal.Cryptography.RSAOpenSsl.GenerateKey()
            at System.Lazy`1.CreateValue()
            at System.Lazy`1.LazyInitValue()
            at Internal.Cryptography.RSAOpenSsl.Encrypt(Byte[] data, OpenSslRsaPadding padding)
            at System.Security.Cryptography.RSACryptoServiceProvider.Encrypt(Byte[] rgb, Boolean fOAEP)
            at System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip()
```

With the change:
```
   System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip [FAIL]
      System.OutOfMemoryException : Insufficient memory to continue the execution of the program.
      Stack Trace:
            at Interop.libcrypto.CheckValidOpenSslHandle(SafeHandle handle)
            at Internal.Cryptography.RSAOpenSsl.GenerateKey()
            at System.Lazy`1.CreateValue()
            at System.Lazy`1.LazyInitValue()
            at Internal.Cryptography.RSAOpenSsl.Encrypt(Byte[] data, OpenSslRsaPadding padding)
            at System.Security.Cryptography.RSACryptoServiceProvider.Encrypt(Byte[] rgb, Boolean fOAEP)
            at System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip()
```

RSAerr(RSA_F_RSA_NEW_METHOD, RSA_R_NON_FIPS_METHOD);
```
   System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip [FAIL]
      System.Security.Cryptography.CryptographicException : error:0406A08D:rsa routines:RSA_new_method:non fips method
      Stack Trace:
            at Interop.libcrypto.CheckValidOpenSslHandle(SafeHandle handle)
            at Internal.Cryptography.RSAOpenSsl.GenerateKey()
            at System.Lazy`1.CreateValue()
            at System.Lazy`1.LazyInitValue()
            at Internal.Cryptography.RSAOpenSsl.Encrypt(Byte[] data, OpenSslRsaPadding padding)
            at System.Security.Cryptography.RSACryptoServiceProvider.Encrypt(Byte[] rgb, Boolean fOAEP)
            at System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip()
```

No error set, but return NULL
```
   System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip [FAIL]
      System.Security.Cryptography.CryptographicException : Error occurred during a cryptographic operation.
      Stack Trace:
            at Interop.libcrypto.CheckValidOpenSslHandle(SafeHandle handle)
            at Internal.Cryptography.RSAOpenSsl.GenerateKey()
            at System.Lazy`1.CreateValue()
            at System.Lazy`1.LazyInitValue()
            at Internal.Cryptography.RSAOpenSsl.Encrypt(Byte[] data, OpenSslRsaPadding padding)
            at System.Security.Cryptography.RSACryptoServiceProvider.Encrypt(Byte[] rgb, Boolean fOAEP)
            at System.Security.Cryptography.Rsa.Tests.EncryptDecrypt.RsaCryptRoundtrip()
```

stephentoub:
LGTM.

> Calls into NativeCrypto throw an unadorned CryptographicException

What happens in cases where we call OpenSSL functions from NativeCrypto... we just end up ignoring the error codes in those cases?

bartonjs:
Yeah, my original thought was that usually NativeCrypto returning a failure status is about bad inputs.  But since it does wrap OpenSSL there's a chance it was just trying to say that OpenSSL itself reported failure.

Since NativeCrypto returning an error means an unadorned CryptographicException would be thrown, but NativeCrypto transporting an error would mean we want the underlying ERR_get_error value, it makes sense to do another pass and update all of the exceptions which come right after a NativeCrypto call gone wrong.

stephentoub:
Sounds good.

src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
```diff
@@ -17,13 +19,56 @@ internal static partial class libcrypto
...
+        {
+            const uint ReasonMask = 0x0FFF;
+            const uint ERR_R_MALLOC_FAILURE = 64 | 1;
+
+            uint error = ERR_get_error();
+            uint last = error;
+
+            while (last != 0)
+            {
+                last = ERR_get_error();
```
nguerrera: Should we log these? Make them inner exceptions?

stephentoub: Hmm. I'm not sure where it'd make sense to log these. Inner exceptions is an interesting idea; the concern is that they may be (likely are?) unrelated to the current error.

nguerrera: Some comments here would help.

When do we expect the stack to have more than one error and what would it mean? Is it a bug that somebody else left one behind? If so, could this assert?

bartonjs: Sure, I'll add some comments.  A couple scenarios off-hand are
1. We do some sort of fallback and don't drain the error state out after the first call fails
2. OpenSSL internally ends up calling ERR_set_error twice during one call that we make

Though, it also turns out that by doing more testing here this fix also isn't right... `ERR_get_error`/`ERR_set_error` isn't a stack, but a queue.  So I'll be fixing that, too.  Good thing it's consolidated now :smile:.

bartonjs:
Added some comments, and changed the drain algorithm to a queue from the original stack.  I'll do the NativeCrypto call pass in an hour or so.

stephentoub:
Other than the delegate issue, LGTM.

src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
```diff
@@ -17,13 +19,78 @@ internal static partial class libcrypto
...
+            uint error = ERR_get_error();
+            uint lastRead = error;
+
+            // 0 (there's no named constant) is only returned when the calls
+            // to ERR_get_error exceed the calls to ERR_set_error.
+            while (lastRead != 0)
+            {
+                error = lastRead;
+                lastRead = ERR_get_error();
+            }
```
stephentoub: If you wanted to have just one call to ERR_get_error, this could be rewritten like:
```C#
uint error = 0, lastRead;
while ((lastRead = ERR_get_error()) != 0)
{
    error = lastRead;
}
```

PR #2267, opened 2015-07-08
Fix System.IO.FileSystem's PathHelpers for Unix

### LINKED ISSUE #2253: System.IO.FileSystem PathHelpers needs to be revamped
System.IO.FileSystem has its own internal copy of much of the code from System.IO.Path.  This is problematic for a variety of reasons:
- Duplicated code in corefx that has diverged over time
- Lack of proper Unix support

Once System.IO.Path is in corefx (#2248), we need to rationalize all of this.

stephentoub:
The FileSystem's PathHelpers internal class contains some utility functions for processing paths.  Some of these functions are copies of non-exposed Path members, while others are unique.  A bunch of those members have Windows-specifics coded into them, which is causing problems when dealing with various special paths on Unix, for example paths that end with double dots, relative paths that begin with a space, etc.

This commit splits PathHelpers into a PathHelpers.cs, PathHelpers.Windows.cs, and PathHelpers.Unix.cs, and implements the Unix functions appropriately.

It also fixes up a variety of tests that are now failing because they were expecting the Windows behavior.

Fixes #2253, #2205

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/PathHelpers.Unix.cs
+++ b/src/System.IO.FileSystem/src/System/IO/PathHelpers.Unix.cs
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.IO
+{
+    internal static partial class PathHelpers
+    {
+        internal static int GetRootLength(string path)
+        {
+            CheckInvalidPathChars(path);
+            return path.Length > 0 && IsDirectorySeparator(path[0]) ? 1 : 0;
+        }
+
+        internal static void CheckSearchPattern(string searchPattern)
+        {
+            // ".." should not be used to move up directories. On Windows, this is more strict, and ".."
+            // can only be used in particular places in a name, whereas on Unix it can be used anywhere.
+            // So, throw if we find a ".." that's its own component in the path.
+            for (int index = 0; (index = searchPattern.IndexOf("..", index, StringComparison.Ordinal)) >= 0; index += 2)
+            {
+                if ((index == 0 || IsDirectorySeparator(searchPattern[index - 1])) && // previous character is directory separator
+                    (index + 2 == searchPattern.Length || IsDirectorySeparator(searchPattern[index + 2]))) // next character is directory separator
+                {
+                    throw new ArgumentException(SR.Arg_InvalidSearchPattern, "searchPattern");
+                }
+            }
+        }
+
+        internal static bool IsDirectorySeparator(char c)
+        {
+            return c == Path.DirectorySeparatorChar;
+        }
+
+        internal static string GetFullPathInternal(string path)
+        {
+            // The Windows implementation trims off whitespace from the start and end of the path,
+            // and then based on whether that trimmed path is rooted decides to use the trimmed
+            // or original version.  On Unix, a filename can be composed of entirely whitespace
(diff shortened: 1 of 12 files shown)


stephentoub:
cc: @ianhays, @ellismg

src/System.IO.FileSystem/tests/Directory/GetFileSystemEntries_str_str.cs
```diff
@@ -190,6 +190,31 @@ public void UnixFilePathWithSpaces()
         }
     }
 
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void UnixSearchPatternWithSpaces()
```
ianhays: Good addition!

src/System.IO.FileSystem/src/System/IO/PathHelpers.Unix.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        internal static string GetFullPathInternal(string path)
+        {
+            // The Windows implementation trims off whitespace from the start and end of the path,
+            // and then based on whether that trimmed path is rooted decides to use the trimmed
+            // or original version.  On Unix, a filename can be composed of entirely whitespace
+            // characters, so we can't legimately do such trimming.  As such, we just delegate
+            // to the real GetFullPath.
+            return Path.GetFullPath(path);
```
ianhays: This is great - I ran into so many issues with Path curtailing whitespace when I was writing my Unix tests.

src/System.IO.FileSystem/src/System/IO/PathHelpers.Windows.cs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.IO
+{
+    internal static partial class PathHelpers
+    {
+        // Trim trailing white spaces, tabs etc but don't be aggressive in removing everything that has UnicodeCategory of trailing space.
+        // String.WhitespaceChars will trim more aggressively than what the underlying FS does (for ex, NTFS, FAT).    
+        internal static readonly char[] TrimEndChars = { (char)0x9, (char)0xA, (char)0xB, (char)0xC, (char)0xD, (char)0x20, (char)0x85, (char)0xA0 };
```
ianhays: Any reason for the chars to be declared by their hex value instead of by character e.g. TrimEndChars = { '\t', '\n', ' ', etc}?

I ask because where I used similar code I declared them by character.

stephentoub: I just moved this from PathHelpers.cs, but I'm assuming because of the last two chars in the list (horizontal ellipses and non-breaking space) that don't have an ASCII representation.  I'd be fine if we wanted to change this to use the C# character instead for the others where that works, though we can do that separately if desired, and if we do, we should look to do something similar in Path.Windows.cs: https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs#L19-L48

src/System.IO.FileSystem/src/System/IO/PathHelpers.Unix.cs
stephentoub: Yeah, it's been a pain :smile:

ianhays:
LGTM!

stephentoub:
Thanks, Ian.

PR #2268, opened 2015-07-08
Remove string allocation/copy from Path.Combine

stephentoub:
Path.Combine(string, string, string) is essentially implemented as Path.Combine(Path.Combine(string, string), string), resulting in the temporary string from combining the first two strings and then another string from combining that with the third string.  This commit adds an implementation for combining three paths without the intermediate allocation/copy.

### DIFF
--- a/src/System.Runtime.Extensions/src/System/IO/Path.cs
+++ b/src/System.Runtime.Extensions/src/System/IO/Path.cs
@@ -311,7 +311,7 @@ namespace System.IO
             CheckInvalidPathChars(path2);
             CheckInvalidPathChars(path3);
 
-            return CombineNoChecks(CombineNoChecks(path1, path2), path3);
+            return CombineNoChecks(path1, path2, path3);
         }
 
         public static string Combine(params string[] paths)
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @KrzysztofCwalina, @ellismg

justinvp:
:+1:

nguerrera:
LGTM

jasonwilliams200OK:
@stephentoub, I understand that the new public signatures require API approval but can you please comment on the impact on *performance*, if we drop all public overloaded signatures of `Combine` method in favor of one `Path.Combile(params string[])`? In particular, will then the generated machine-code/jitdisasm against `Path.Combile('a', 'b')` (or `Path.Combine('a', 'b', 'c')`) be different than the current (fixed num of args) implementation?

justinvp:
@jasonwilliams200OK there's already a public `Combine(params string[] paths)` method. The benefit of the other overloads is to avoid the `string[]` params array allocation for a common number of paths.

KrzysztofCwalina:
Looking good. Thanks!

stephentoub:
Thanks, guys.

PR #2269, opened 2015-07-08
Move S.R.Metadata throw helpers together

### LINKED ISSUE #2261: Move S.R.Metadata throw helpers together
We have a pattern in the reader to move all throw helpers into simple non-inlined helpers taking no arguments. This makes them have the smallest impact possible on calling methods, which both improves the likelihood that callers will be inlined and reduces the impact on code size from their inlining. The exceptional cases are rare (by definition) so there is no value in inlining their implementation.

Today we have various private static void ThrowXyz() spread throughout. These would be better organized within a single internal Throw static class with documentation around why we have this pattern.

Maxwe11:
Closes #2261


stephentoub:
Looks reasonable to me.  @nguerrera?

nguerrera:
LGTM. Thanks!

PR #2272, opened 2015-07-08
Separate platform-specifics out of shared FileSystem code

stephentoub:
In a few places, we still have platform-specific code in shared System.IO.FileSystem code.  In particular, DirectoryInfo is doing various manipulations related to Windows path roots that can result in doing the wrong thing on Unix, e.g. creating a DirectoryInfo for "C:" on Unix should be for the directory named "C:", but instead today it'll follow the Windows-specific logic of rerouting to target the current directory.  This commit fixes that.  It also moves some code used only in the Windows implementation into Windows-specific files (e.g. Directory.SearchData).

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs
+++ b/src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs
@@ -1,18 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
-using System.Globalization;
-using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using System.Security;
-using System.Text;
-
-using Microsoft.Win32;
 
 namespace System.IO
 {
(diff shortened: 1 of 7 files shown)


src/System.IO.FileSystem/src/System/IO/DirectoryInfo.Unix.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.IO
+{
+    partial class DirectoryInfo
+    {
+        private static bool SpecialCaseToCurrentDirectory(string path)
```
stephentoub: I'm going to wait to merge these changes until #2267 is merged.  Then I'll probably move this helper to PathHelpers.Windows/Unix.cs rather than making DirectoryInfo a partial class.

stephentoub:
cc: @ianhays, @ericstj, @ellismg

stephentoub:
(I had a bug in the new test causing it to fail on Windows when run from a drive other than C:, such that it passed for me locally but failed in CI.  Fixed.)

@ianhays, @ellismg, mind taking a look?

src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs
```diff
@@ -74,17 +49,20 @@ public DirectoryInfo Parent
...
-                if (s.Length > 3 && s[s.Length - 1] == Path.DirectorySeparatorChar)
-                    s = FullPath.Substring(0, FullPath.Length - 1);
-                parentName = Path.GetDirectoryName(s);
-                if (parentName == null)
-                    return null;
-                DirectoryInfo dir = new DirectoryInfo(parentName, null);
-                return dir;
+                string s = FullPath;
+
+                // FullPath might be end in either "parent\child" or "parent\child", and in either case we want
```
ianhays: Comment here is confusing. It should be:

```FullPath might end in either "parent\child" or "parent\child\"```

File: src/System.IO.FileSystem/tests/DirectoryInfo/ctor_str.cs [hunk 3]
```diff
@@ -54,6 +54,28 @@ public void CDriveCase()
         }
 
         [Fact]
```
ianhays: Tests look good.

ianhays:
LGTM, only issue I caught was the typo in the comment.

src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs
stephentoub: Oops, will fix.

PR #2277, opened 2015-07-08
Release the StringBuilder back to the StringBuilderCache.

### LINKED ISSUE #2275: Unix Path.NormalizePath acquires StringBuilder but never releases it.
In this file:
[src\System.Runtime.Extensions\src\System\IO\Path.Unix.cs](https://github.com/dotnet/corefx/blob/7938a0e9264f4fbbe3fa5638494387b08444f383/src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs)
the NormalizePath method uses StringBuilderCache to acquire a StringBuilder, however the StringBuilder is never released back to the cache.

AustinWise:
Fixes #2275

### DIFF
--- a/src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
+++ b/src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
@@ -134,10 +134,12 @@ namespace System.IO
                 throw new PathTooLongException(SR.IO_PathTooLong);
             }
 
-            return
+            string result =
                 sb.Length == 0 ? (fullCheck ? DirectorySeparatorCharAsString : string.Empty) :
                 sb.Length == path.Length ? path :
                 sb.ToString();
+            StringBuilderCache.Release(sb);
+            return result;
         }
 
         private static string RemoveLongPathPrefix(string path)


src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -134,10 +134,19 @@ public static string GetFullPath(string path)
...
+            }
+            else if (sb.Length == path.Length)
+            {
+                StringBuilderCache.Release(sb);
+                return path;
+            }
+            else
+            {
+                return StringBuilderCache.GetStringAndRelease(sb);
+            }
```
stephentoub: Why isn't this releasing the StringBuilder in the first case, too?  Rather than rewrite the whole expression with multiple calls to GetStringAndRelease, rather than the current code that does:
```C#
return stuff;
```
I'd do:
```C#
string result = stuff;
StringBuilderCache.Release(sb);
return result;
```

AustinWise: Opps, missed another release. I'll revise it.

stephentoub:
Thanks for the fix and the revision.  LGTM.

stephentoub:
@AustinWise, could you squash these?  I didn't notice previously that this was done as two commits, and the second really obviates the first rather than builds on it.

AustinWise:
Yeah, I'll do that. I was just not sure if it would mess up the pull request somehow if I edited the commits.

stephentoub:
Cool, thanks :smile:

PR #2283, opened 2015-07-09
Fix bugs in OSX on process type implementations.

Priya91:
In this commit, 

1. Added test coverage for properties on Process type, subsequent PRs will be opened to provide test coverage on Process methods, and other types and scenarios in this contract, as work is done incrementally.
2. Added implementation to get Session ID of a process on OSX.
3. Fixed bug in retrieving the priority of a task on OSX, we were reading the wrong struct value for this.

cc @stephentoub @joshfree @sokket

### DIFF
--- a/src/System.Diagnostics.Process/src/System.Diagnostics.Process.csproj
+++ b/src/System.Diagnostics.Process/src/System.Diagnostics.Process.csproj
@@ -261,6 +261,9 @@
     <Compile Include="$(CommonPath)\Interop\Unix\libcoreclr\Interop.ForkAndExecProcess.cs">
       <Link>Common\Interop\Unix\Interop.ForkAndExecProcess.cs</Link>
     </Compile>
+    <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.getsid.cs">
+      <Link>Common\Interop\Unix\Interop.getsid.cs</Link>
+    </Compile>
     <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.ResourceLimits.cs">
       <Link>Common\Interop\Unix\Interop.ResourceLimits.cs</Link>
     </Compile>
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Win32.SafeHandles;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Pipes;
@@ -13,7 +14,7 @@ namespace System.Diagnostics.ProcessTests
 {
     public partial class ProcessTest : IDisposable
     {
-        private const int WaitInMS = 100 * 1000; 
+        private const int WaitInMS = 100 * 1000;
         private const string CoreRunName = "corerun";
         private const string TestExeName = "System.Diagnostics.Process.TestConsoleApp.exe";
         private const int SuccessExitCode = 100;
@@ -51,8 +52,8 @@ namespace System.Diagnostics.ProcessTests
             _processes.Add(p);
 
             p.StartInfo.FileName = CoreRunName;
-            p.StartInfo.Arguments = string.IsNullOrWhiteSpace(optionalArgument) ? 
-                TestExeName : 
+            p.StartInfo.Arguments = string.IsNullOrWhiteSpace(optionalArgument) ?
+                TestExeName :
                 TestExeName + " " + optionalArgument;
 
             // Profilers / code coverage tools doing coverage of the test process set environment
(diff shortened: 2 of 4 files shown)


src/System.Diagnostics.Process/src/System.Diagnostics.Process.csproj
```diff
@@ -261,6 +261,9 @@
     <Compile Include="$(CommonPath)\Interop\Unix\libcoreclr\Interop.ForkAndExecProcess.cs">
       <Link>Common\Interop\Unix\Interop.ForkAndExecProcess.cs</Link>
     </Compile>
+    <Compile Include="..\..\Common\src\Interop\Unix\libc\Interop.getsid.cs">
```
stephentoub: This should be using $(CommonPath).

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -79,25 +80,40 @@ public void SetAndCheckBasePriority(ProcessPriorityClass exPriorityClass, int pr
...
+                SetAndCheckBasePriority(ProcessPriorityClass.Idle, 4);
+                SetAndCheckBasePriority(ProcessPriorityClass.Normal, 8);
+            }
+            finally
+            {
+                _process.PriorityClass = originalPriority;
+            }
+        }
+
+        [Fact, PlatformSpecific(PlatformID.AnyUnix), OuterLoop] // This test requires admin elevation on Unix
```
stephentoub: Do we run our OuterLoop tests in CI on Unix, and do we run them in a privileged process?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -79,25 +80,40 @@ public void SetAndCheckBasePriority(ProcessPriorityClass exPriorityClass, int pr
             Assert.Equal(priority, _process.BasePriority);
         }
 
-        [Fact]
+        [Fact, PlatformSpecific(PlatformID.Windows)]
```
stephentoub: the Unix test is named Process_BasePriorityUnix... this one should be named to match for Windows, right?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -79,25 +80,40 @@ public void SetAndCheckBasePriority(ProcessPriorityClass exPriorityClass, int pr
...
-                    //SetAndCheckBasePriority(ProcessPriorityClass.RealTime, 24);
-                    SetAndCheckBasePriority(ProcessPriorityClass.High, 13);
-                    SetAndCheckBasePriority(ProcessPriorityClass.Idle, 4);
-                    SetAndCheckBasePriority(ProcessPriorityClass.Normal, 8);
-                }
-                finally
-                {
-                    _process.PriorityClass = originalPriority;
-                }
+                //SetAndCheckBasePriority(ProcessPriorityClass.RealTime, 24);
```
stephentoub: I know it was this way before, but do we know why this one is commented out?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
stephentoub: Also, what about it requires privileges?  If it's just the setting, should we still verify we can get the PriorityClass without privileges?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -184,10 +200,15 @@ public void Process_ExitTime()
 
 
         [Fact]
-        [PlatformSpecific(PlatformID.Windows)]
-        public void Process_GetHandle()
+        public void Process_Id()
         {
-            Assert.Equal(_process.Id, Interop.GetProcessId(_process.SafeHandle));
+            if (global::Interop.IsWindows)
+                Assert.Equal(_process.Id, Interop.GetProcessId(_process.SafeHandle));
```
stephentoub: Nit: ideally this would be wrapped in braces like the else branch

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -443,6 +463,25 @@ public void Process_PriorityBoostEnabled()
             }
         }
 
+        [Fact, PlatformSpecific(PlatformID.AnyUnix), OuterLoop] // This test requires admin elevation on Unix
```
stephentoub: Same questions about privileges.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
```diff
@@ -473,6 +512,23 @@ public void ProcessProcessName()
...
+        public void Process_SessionId()
+        {
+            if (global::Interop.IsWindows)
+            {
+                uint sessionId;
+                ProcessIdToSessionId((uint)_process.Id, out sessionId);
+                Assert.Equal(sessionId, (uint)_process.SessionId);
+            }
+            else Assert.Equal(getsid(_process.Id), _process.SessionId);
+        }
```
stephentoub: Nit: consider restructuring this like:
```C#
uint sessionId;
if (global::Interop.IsWindows)
{
    ProcessIdToSessionId((uint)_process.Id, out sessionId);
}
else
{
    sessionId = getsid(_process.Id);
}
Assert.Equal(sessionId, _process.SessionId);
```
That way we have the one assert, and we're simply switching by platform to get the expected value.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/System.Diagnostics.Process.Tests.csproj
```diff
@@ -18,10 +18,10 @@
   <ItemGroup>
     <Compile Include="Interop.cs" />
     <Compile Include="ProcessTest.cs" />
+    <Compile Include="$(CommonPath)\Interop\Interop.PlatformDetection.cs" />
```
stephentoub: Not a big deal, but why did this move?

stephentoub:
Thanks, Priya!

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
Priya91: We are attempting to change the priority of the process, and on unix, lowering the priority requires privileged access. When setting PriorityClass, we are also setting the PriorityClassCore, which calls setpriority on Unix.

Priya91: Right now, we dont have a good story to run Outerloop tests, are how we are going to run tests in admin access, there's an open issue #1477  to track this. Once that system is in place, then these tests should automatically light up.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
Priya91: From msdn, 

> RealTime priority preempt the threads of all other processes, including operating system processes performing important tasks. Thus, a RealTime priority process that executes for more than a very brief interval can cause disk caches not to flush or cause the mouse to be unresponsive.

I am not sure if we should be doing this in test, as it could potentially hang the machine.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/System.Diagnostics.Process.Tests.csproj
Priya91: I was including and excluding files during testing from project, and VS has automatically changed the lines. Will undo this.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
stephentoub: Ok.  Consider adding a comment explaining why it's commented out.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
stephentoub: I'm wondering whether OuterLoop is really the right category for things that require elevated privileges.  Is that how we've defined it?  Or do we need a separate category for things that require such privileges?  I often run OuterLoop tests locally, because I want to make sure I've run all relevant tests before submitting a PR, and thus far I haven't been doing it elevated... with this I'll need to start doing so.  If that's the path we're on, ok, but we should probably be clear somewhere about what it means.

Priya91: No, OuterLoop doesn't mean all outerloop tests should be run with admin privileges. Outerloop are {longrunning, admin acess, changes system state}.. When @Chrisboh proposal is finalized, we will know what are the subcategories for outerloop, and how we intend to differentiate and run them. So, please continue running the tests as you are doing now. Last i checked with the issue, our plan was to mark the tests in the above set as OuterLoop, if this plan changes, then we will go back and update these. @Chrisboh @joshfree can you guys confirm that this is correct?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTest.cs
Priya91: Done.

PR #2290, opened 2015-07-09
Use Array.Empty<T> for LINQ's Enumerable.Empty<TResult>

justinvp:


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1452,7 +1452,10 @@ private static IEnumerable<TResult> RepeatIterator<TResult>(TResult element, int
 
         public static IEnumerable<TResult> Empty<TResult>()
         {
-            return EmptyEnumerable<TResult>.Instance;
+            // This is a more memory-intensive empty enumerable that allocates a new enumerator each time.
```
stephentoub: I don't believe this comment is accurate any more.  Enumerating an empty array should not allocate a new enumerator: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Array.cs#L2717

justinvp: Excellent! I'll remove the comment.

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
justinvp: As an aside, correct me if I am wrong, but it looks like `SZGenericArrayEnumerator<T>.Empty` is being greedily allocated as soon as `SZGenericArrayEnumerator<T>` is used for the first time (first time the generic `GetEnumerator` is called). Which means lots of possible unnecessary empty enumerator allocations for each T. Wouldn't it be better if this was lazily allocated only when an empty enumerator was actually needed? Something like `EmptySZGenericArrayEnumerator<T>.Instance`. (I can open an issue/PR on CoreCLR to continue the discussion there, if warranted).

VSadov:
LGTM

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: Yeah, that's probably a good idea.

justinvp: Never mind -- I tested it out and the runtime is doing the right thing.

PR #1436, opened 2015-04-19
Improve performance of BigInteger.Multiply

axelheer:
To introduce performance tweaks for BigInteger's operations a new static class `BigCalc` implements those algorithms based on raw uint[] objects, partially switching even to pointer arithmetic to spare some index calculations. First of all a multiplication implementation is included, furthermore an optimized version for squaring.

A basic performance comparison based on [this code](https://gist.github.com/axelheer/3b701c91271b7c762586) unveils the following results:

**Multiply**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |        41 |       35 |
|        64 | 1,000,000 |       127 |       89 |
|       256 | 1,000,000 |       284 |      182 |
|     1,024 |   100,000 |       236 |      147 |
|     4,096 |    10,000 |       323 |      142 |
|    16,384 |     1,000 |       495 |      132 |
|    65,536 |       100 |       785 |      121 |

**Square**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |        41 |       42 |
|        64 | 1,000,000 |       127 |       78 |
|       256 | 1,000,000 |       284 |      147 |
|     1,024 |   100,000 |       236 |      102 |
|     4,096 |    10,000 |       323 |       98 |
|    16,384 |     1,000 |       495 |       92 |
|    65,536 |       100 |       785 |       84 |

*Note:* the performance gain is not as big as expected based on my previous work (as mentioned within the issue), but I'd still call it significant. I actually had some difficulties executing a custom build of *System.Runtime.Numerics* though, since just dropping in the *dll* as described [here](https://github.com/dotnet/corefx/wiki/Open-Source-Signing) does not work 'cause of some "type forwards". I finally managed to execute it using *kre-coreclr-win-x64.1.0.0-beta3* and some hacks... :flushed:

#1307

PR #2291, opened 2015-07-09
Reduce array allocations within BigIntegerCalculator

axelheer:
- An allocation for BigInteger.DivRem with small divisor is unnecessary, which gets fixed. BigIngeger handles these cases better now.
- The square and multiply code submitted with #1436 used stackalloc, which led to stack overflows for huge numbers. With #1618 these has been changed to ordinary array allocations, which put some pressure on the managed heap. Thus, a hybrid solution should be better.

### DIFF
--- a/src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
@@ -26,6 +26,7 @@ namespace System.Numerics
 
         // Mutable for unit testing...
         private static int SquareThreshold = 32;
+        private static int AllocationThreshold = 4096;
 
         [SecuritySafeCritical]
         private unsafe static void Square(uint* value, int valueLength,
@@ -111,9 +112,12 @@ namespace System.Numerics
 
                 int foldLength = valueHighLength + 1;
                 int coreLength = foldLength + foldLength;
-                fixed (uint* fold = new uint[foldLength],
-                             core = new uint[coreLength])
+
+                if (coreLength < AllocationThreshold)
                 {
+                    uint* fold = stackalloc uint[foldLength];
+                    uint* core = stackalloc uint[coreLength];
+
                     // ... compute z_a = a_1 + a_0 (call it fold...)
                     Add(valueHigh, valueHighLength,
                         valueLow, valueLowLength,
@@ -128,6 +132,27 @@ namespace System.Numerics
 
                     // ... and finally merge the result! :-)
                     AddSelf(bits + n, bitsLength - n, core, coreLength);
+                }
+                else
+                {
+                    fixed (uint* fold = new uint[foldLength],
+                                 core = new uint[coreLength])
+                    {
+                        // ... compute z_a = a_1 + a_0 (call it fold...)
+                        Add(valueHigh, valueHighLength,
+                            valueLow, valueLowLength,
+                            fold, foldLength);
+
+                        // ... compute z_1 = z_a * z_a - z_0 - z_2
+                        Square(fold, foldLength,
+                               core, coreLength);
+                        SubtractCore(bitsHigh, bitsHighLength,
+                                     bitsLow, bitsLowLength,
+                                     core, coreLength);
+
+                        // ... and finally merge the result! :-)
+                        AddSelf(bits + n, bitsLength - n, core, coreLength);
+                    }
                 }
             }
         }
@@ -270,10 +295,13 @@ namespace System.Numerics
                 int leftFoldLength = leftHighLength + 1;
                 int rightFoldLength = rightHighLength + 1;
                 int coreLength = leftFoldLength + rightFoldLength;
-                fixed (uint* leftFold = new uint[leftFoldLength],
-                             rightFold = new uint[rightFoldLength],
-                             core = new uint[coreLength])
+
+                if (coreLength < AllocationThreshold)
                 {
+                    uint* leftFold = stackalloc uint[leftFoldLength];
+                    uint* rightFold = stackalloc uint[rightFoldLength];
+                    uint* core = stackalloc uint[coreLength];
+
                     // ... compute z_a = a_1 + a_0 (call it fold...)
                     Add(leftHigh, leftHighLength,
                         leftLow, leftLowLength,
@@ -294,6 +322,34 @@ namespace System.Numerics
 
                     // ... and finally merge the result! :-)
                     AddSelf(bits + n, bitsLength - n, core, coreLength);
+                }
+                else
+                {
+                    fixed (uint* leftFold = new uint[leftFoldLength],
+                                 rightFold = new uint[rightFoldLength],
+                                 core = new uint[coreLength])
+                    {
+                        // ... compute z_a = a_1 + a_0 (call it fold...)
+                        Add(leftHigh, leftHighLength,
+                            leftLow, leftLowLength,
+                            leftFold, leftFoldLength);
+
+                        // ... compute z_b = b_1 + b_0 (call it fold...)
+                        Add(rightHigh, rightHighLength,
+                            rightLow, rightLowLength,
+                            rightFold, rightFoldLength);
+
+                        // ... compute z_1 = z_a * z_b - z_0 - z_2
+                        Multiply(leftFold, leftFoldLength,
+                                 rightFold, rightFoldLength,
+                                 core, coreLength);
+                        SubtractCore(bitsHigh, bitsHighLength,
+                                     bitsLow, bitsLowLength,
+                                     core, coreLength);
+
+                        // ... and finally merge the result! :-)
+                        AddSelf(bits + n, bitsLength - n, core, coreLength);
+                    }
                 }
             }
         }
--- a/src/System.Runtime.Numerics/tests/BigInteger/multiply.cs
+++ b/src/System.Runtime.Numerics/tests/BigInteger/multiply.cs
@@ -38,8 +38,16 @@ namespace System.Numerics.Tests
         public static void RunMultiply_TwoLargeBigIntegers_Threshold()
         {
             // Again, with lower threshold
-            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, RunMultiply_TwoLargeBigIntegers);
-            BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, RunMultiply_TwoLargeBigIntegers);
+            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, () =>
+                BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, RunMultiply_TwoLargeBigIntegers)
+            );
+
+            // Again, with lower threshold
+            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, () =>
+                BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, () =>
+                    BigIntTools.Utils.RunWithFakeThreshold("AllocationThreshold", 8, RunMultiply_TwoLargeBigIntegers)
+                )
+            );
         }
 
         [Fact]
--- a/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
@@ -945,10 +945,10 @@ namespace System.Numerics
 
             if (trivialDivisor)
             {
-                uint[] rest;
+                uint rest;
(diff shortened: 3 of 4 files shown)


axelheer:
cc: @stephentoub

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
```diff
@@ -26,6 +26,7 @@ public unsafe static uint[] Square(uint[] value)
 
         // Mutable for unit testing...
         private static int SquareThreshold = 32;
+        private static int AllocationThreshold = 4096;
```
stephentoub: This is bit length, right?  Just curious how you arrived at this value for the threshold?

src/System.Runtime.Numerics/tests/BigInteger/multiply.cs
```diff
@@ -38,8 +38,16 @@ public static void RunMultiply_TwoLargeBigIntegers()
...
-            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, RunMultiply_TwoLargeBigIntegers);
-            BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, RunMultiply_TwoLargeBigIntegers);
+            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, () =>
+                BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, RunMultiply_TwoLargeBigIntegers)
+            );
+
+            // Again, with lower threshold
+            BigIntTools.Utils.RunWithFakeThreshold("SquareThreshold", 8, () =>
+                BigIntTools.Utils.RunWithFakeThreshold("MultiplyThreshold", 8, () =>
+                    BigIntTools.Utils.RunWithFakeThreshold("AllocationThreshold", 8, RunMultiply_TwoLargeBigIntegers)
```
stephentoub: Is 8 too low here for AllocationThreshold?

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
axelheer: Nope, it's array length. It's just a wild guess and I hoped someone has a more profound suggestion. The idea is that numbers around 8 KB can live on the stack, 4096 means 16 KB values and bigger use array allocations instead.

src/System.Runtime.Numerics/tests/BigInteger/multiply.cs
axelheer: There are two runs to cover all the code: the first run modifies only the square and multiply thresholds to execute the stackalloc based code, the second run modifies all three to execute the array based code.

But it should be valid too, if we modify the allocation threshold to say 16 and do only one run.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
stephentoub: Do you have a sense for the typical size of numbers used?  Obviously there will always be outliers, but if there's a sweet spot where we could get 80% while not having the stack allocation be too large, that'd be great.  4K (or 8K because it's used in a pair) on the stack is a fairly large chunk.  @KrzysztofCwalina, thoughts?

src/System.Runtime.Numerics/tests/BigInteger/multiply.cs
stephentoub: What you have is fine, I was just asking.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.SquMul.cs
axelheer: I cannot really speak for other "users" of big numbers here.

KrzysztofCwalina: I have to say 4096 seems large too me. Once the number is large enough, the cost of allocating it on the heap is probably dwarfed by the cost to process the number. Unfortunately, I am not sure where the cut off point should be. Maybe @Maoni0, or @pdussud could suggest something.

Maoni0:
I haven’t had a chance to look at the code – can you elaborate a bit what sort of processing you are doing with the numbers? Ie, what kind of tradeoff is there for this?

axelheer:
It's the multiplication and squaring algorithms for big numbers, which use very temporary memory to speed up things beyond a certain threshold (there could be thousand(s) of operations within about hundred milliseconds depending on number size, actual usage (and of course available hardware)). And the additional memory, which is necessary to do some recursive calls, shouldn't put too much additional pressure on the managed heap -- if the gain in speed leads to much more frequent garbage collections, it wouldn't be that ideal (especially in server environments as @stephentoub pointed out).

But let me switch to some pseudo code (note that we're talking about lengths of uint arrays):

    multiply (left, right) =>
        # left.Length >= right.Length
        when right.Length < 32
            # classic Θ(n^2) algorithm
        else
            # recursive Θ(n^1.585) algorithm
            # core.Length == ceiling(left.Length / 2)
            #             + ceiling(right.Length / 2) + 2
            when core.Length < 4096
                # stackalloc based code
            else
                # fixed array based code

The first threshold is solely speed based chosen, this pull request is about the second one, where I made just a wild guess lacking better knowledge: the idea was that numbers around 8 KB can live on the 1 MB (?) stack, which are already huge numbers of course, but I cannot really speak for others using this stuff. Using much bigger numbers leads to other more optimized environments anyway I think, but we won't have a stack overflow here.

I might add that numbers up to 10.000 bits should be fairly common (just guessing again), which means 512 as second threshold should do it as well. But I've really no idea what's okay to put on the stack...

@Maoni0 does this explanation help? And thanks for helping by the way.

Maoni0:
I don't actually think that making 32k bytes allocations on the stack for this purpose is outrageous - on 64-bit our stack is default to 1mb and nowadays it's not uncommon to see stacks with number of frames in the 200's range. So if we assume 256 frames, that gives you 4k. Sure you are consuming 8x that but most frames would consume nowhere near 4k. And this part of the allocation is not recursive.

However, I don't think that is the problem here. The problem I would ask is if you need 32k. Does 32k give you the optimal perf? Based on the discussion I've seen so far it really boils down to "what do we use to measure our performance?". For something like BigNum I would go with typical benchmarks like http://www.bytereef.org/mpdecimal/benchmarks.html (you can also compare with the implementations they tried). I would pick something that would use both code paths so you can see the effect of moving from your stack alloc code path to your heap alloc code path (calculating pi or factorial). And adjust your threshold see where you start seeing diminishing returns. And I would set the threshold where that point is.

axelheer:
Thanks, I just thought the less the GC needs to collect the better. Howsoever, just doing some performance testing sounds good. I've chosen an exponentiation benchmark: `ModPow`. It makes very heavy usage of the square and multiply code. Just calling it with up to 65,536 bit values to hit both code paths leads to the following results (`GCSettings.IsServerGC == true`, `GCSettings.LatencyMode == GCLatencyMode.Batch`):

| threshold |        time | gain   |
|----------:|------------:|-------:|
|        32 |  366,863 ms | 0.00 % |
|        64 |  353,903 ms | 3.53 % |
|       128 |  353,278 ms | 3.70 % |
|       256 |  351,351 ms | 4.23 % |
|       512 |  350,649 ms | 4.42 % |
|      1024 |  349,685 ms | 4.68 % |
|      2048 |  348,352 ms | 5.05 % |

The speed gain isn't that exciting though, we could just set the threshold to something between 64 and 128. On the other hand, based on your calculations, 512 sounds like a viable choice too (if the actual application does other things than just exponentiation it maybe helps to nurse the GC a bit?). Or we just say values up to 10.000 bits should be common enough to set it to 320.

Maoni0:
Are you doing many GCs? If you are using Server GC, you may not see the GC effect much at all.

axelheer:
No explicit / manual GC at all. I now tried a slightly different approach: run the benchmark code on all logical CPUs in parallel (so the GC threads should have an effect), do a manual GC before starting the stopwatch (to exclude other garbage), and do a manual GC before stopping the stopwatch (to include remaining garbage of the run). Does that make more sense?

| threshold |        time | gain   |
|----------:|------------:|-------:|
|        32 |  751,753 ms | 0.00 % |
|        64 |  718,186 ms | 4.47 % |
|       128 |  692,652 ms | 7.86 % |
|       256 |  681,461 ms | 9.35 % |
|       512 |  677,634 ms | 9.86 % |

Maoni0:
Thanks for doing all the perf measurements! It's much appreciated.

Actually I didn't mean to manually trigger GCs. I meant if you used workstation GC, you will likely see more GCs naturally. I am not sure what prompted you to use server GC by default (is it because you are thinking most of the users of this would be using svr GC?) but I am suggesting that I'd expect some users will be running in the default mode which is wks GC so we'd like to see how it behaves there. You can use perfview.exe (you can download it from Microsoft.com) to collect some GC ETW events:

perfview /GCCollectOnly /nogui /NoV2Rundown /NoNGENRundown collect

type s to stop when you are done running the test.

and open the resulting file in perfview. The GCStats view shows you exactly how many GCs you are doing and how long they are. This way you know if GCs are even affecting your scenario much at all.

axelheer:
Server GC seems to be the default for DNX, I included the information just in case without any profound reflections. And I actually have no idea how to change that since there is no .exe/.config when running the code as described [here](https://github.com/dotnet/coreclr/blob/master/Documentation/install/get-dotnetcore-dnx-windows.md). Using `dnx . run` was the only way I found working since the "classic" runtime contains type forwards from `System.Runtime.Numerics` to `System.Numerics`, which make it impossible for me to just replace `System.Runtime.Numerics` as decribed [here](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/oss-signing.md).

Besides I'm curious, doesn't my previous measurement help at all? I know server GC isn't the same as workstation GC, but including all the garbage within the benchmark and giving the GC no separate CPU to do it's work, shouldn't that already show the performance impacts / gains?

Nevertheless, I played with perfview as you suggested and did a new (smaller) benchmark. Still using server GC, but hopefully better than nothing.

| threshold | GC count |
|----------:|---------:|
|        32 |       51 |
|        64 |       29 |
|       128 |       15 |
|       256 |        7 |
|       512 |        3 |

axelheer:
I finally was able to get an good old .exe as described [here](https://github.com/dotnet/coreclr/blob/master/Documentation/install/get-dotnetcore-windows.md) (the listed package versions are not available by the way), which uses workstation GC => time for a provisional summary!

**GC counts - server - up to 16384 bits**

| threshold | GC count | GC pause |
|----------:|---------:|---------:|
|        32 |       51 |    1.4 % |
|        64 |       29 |    0.9 % |
|       128 |       15 |    0.5 % |
|       256 |        7 |    0.3 % |
|       512 |        3 |    0.2 % |

**GC counts - workstation - up to 16384 bits**

| threshold | GC count | GC pause |
|----------:|---------:|---------:|
|        32 |     1092 |    1.9 % |
|        64 |      615 |    0.9 % |
|       128 |      342 |    0.4 % |
|       256 |      175 |    0.2 % |
|       512 |       68 |    0.1 % |

**perf multi threaded - server - up to 65536 bits**

| threshold |        time | gain   |
|----------:|------------:|-------:|
|        32 |  751,753 ms | 0.00 % |
|        64 |  718,186 ms | 4.47 % |
|       128 |  692,652 ms | 7.86 % |
|       256 |  681,461 ms | 9.35 % |
|       512 |  677,634 ms | 9.86 % |

**perf single threaded - workstation - up to 65536 bits**

| threshold |        time | gain   |
|----------:|------------:|-------:|
|        32 |  355,055 ms | 0.00 % |
|        64 |  347,863 ms | 2.03 % |
|       128 |  343,574 ms | 3.23 % |
|       256 |  342,955 ms | 3.41 % |
|       512 |  342,415 ms | 3.56 % |

Alright. We can see that doubling the threshold leads to halve the GC collects (which looks somehow naturally to me), putting the machine under "full" pressure and using server GC (which should simulate an ordinary server environment) leads to 256 or 512 or something in between, and doing a simple single threaded benchmark and using workstation GC (which should simulate an ordinary client application) leads to 128 or 256 or something in between. Right?

stephentoub:
Thanks for gathering all of this data, @axelheer!  Seems like a threshold of 256 strikes a nice balance, providing almost all of the performance wins for relatively little stack space.  Unless anyone has a different opinion, I'd suggest going with that.

Maoni0:
Agreed - 256 looks good. And @axelheer, all your previous measurements were helpful too, I just wanted to make sure we don't get surprises when using workstation GC. Thanks again for doing all the detailed measurements!

axelheer:
Okay, I'll prepare a new pull request then. :tada:

And thanks for the insights, my understanding of these allocation / collection effects improved, I think.

PR #2293, opened 2015-07-09
Use "-source" parameter in OSX build script, NuGet.Config was removed

chcosta:
NuGet.Config was removed from source, this breaks the OSX build using build.sh.  Convert OSX build to restore buildtools using the "-source" parameter.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -68,7 +68,7 @@ fi
 # Grab the MSBuild package if we don't have it already
 if [ ! -e "$__msbuildpath" ]; then
     echo "Restoring MSBuild..."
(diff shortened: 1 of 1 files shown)


akoeplinger:
LGTM :+1:

stephentoub:
https://github.com/dotnet/corefx/commit/68e6afc63180a3122f5a85310881aaf3b9d9c27e stated that it needed an additional source for the OS X build... that's not necessary here?

chcosta:
This change is scoped to the msbuild restore, packages had an additional dependency.  My testing showed that this was the only change necessary to unblock msbuild.

stephentoub:
Ok.

PR #2298, opened 2015-07-10
Port additional System.IO tests from ToF

FiveTimesTheFun:
Port additional System.IO tests from ToF for
System.IO.BinaryWriter
System.IO.MemoryStream
System.IO.Stream
System.IO.StreamReader
System.IO.StreamWriter
System.IO.StringReader
System.IO.StringWriter

#862


src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+using System.IO;
+using System.Text;
+
+namespace Tests
+{
+    [ContractsRequired("System.IO, System.Runtime, System.Text.Encoding, System.Threading.Tasks")]
+    public class BinaryWriterTests
+    {
+
+        [TestMethod]
```
stephentoub: ContractsRequired?  TestMethod?  This isn't using xunit, is it?

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            // [] Smoke test to ensure that we can write with the constructed writer
+            mstr = new MemoryStream();
+            dw2 = new BinaryWriter(mstr);
+            dw2.Write(true);
+            dw2.Flush();
+            mstr.Position = 0;
+            dr2 = new BinaryReader(mstr);
+            Assert.IsTrue(dr2.ReadBoolean(), "Error_298yx! Incorrect value on stream");
+            dr2.Dispose();
+            mstr.Dispose();
```
stephentoub: I'd find the tests to be more readable if they used ```using``` to handle disposal.

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            MemoryStream mstr;
+
+            // [] Smoke test to ensure that we can write with the constructed writer
+            mstr = new MemoryStream();
+            dw2 = new BinaryWriter(mstr);
+            dw2.Write(true);
+            dw2.Flush();
+            mstr.Position = 0;
+            dr2 = new BinaryReader(mstr);
+            Assert.IsTrue(dr2.ReadBoolean(), "Error_298yx! Incorrect value on stream");
```
stephentoub: Do we still need strings like"Error_298yx!"?

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            ms2 = new MemoryStream();
+            sw2 = new BinaryWriter(ms2, Encoding.Unicode);
+            sw2.Write(testString);
+            sw2.Flush();
+            ms2.Position = 0;
+            sr2 = new BinaryReader(ms2, Encoding.Unicode);
+            str2 = sr2.ReadString();
+            Assert.AreEqual(testString, str2, "Error_f897h! Incorrect string on stream==" + str2);
+            sr2.Dispose();
+            ms2.Dispose();
```
stephentoub: This looks like it's the same code three times, just with a different encoding.  Could you make it a theory and use MemberData to provide the different encodings to test?

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            catch (Exception exc)
+            {
+                Assert.Fail("Error_398yg! Incorrect exception thrown, exc==" + exc.ToString());
+            }
+            Assert.IsTrue(mstr.Position == 14, "Err_1234!! Unexpected position value .." + mstr.Position);
+
+            dw2.Dispose();
+            mstr.Dispose();
+
+            // [] Seek pas the end from beginning
```
stephentoub: past

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
```diff
@@ -0,0 +1,391 @@
...
+            memstr2 = new MemoryStream();
+            bw2 = new BinaryWriter(memstr2);
+            bw2.Dispose();
+            try
+            {
+                bw2.Flush();
+            }
+            catch (Exception exc)
+            {
+                Assert.Fail("Error_399c;! Unexpected exception exception thrown, exc==" + exc.ToString());
```
stephentoub: Rather than catching an asserting, for cases like this we usually just let the exception propagate out of the test.

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteByteCharTests.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+            }
+
+            Assert.Equal(ii, chArr.Length);
+
+            dw2.Dispose();
+            dr2.Dispose();
+            mstr.Dispose();
+
+            //VSWhidbey #273805
```
stephentoub: Are these comments still needed?

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteByteCharTests.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            dw2.Write(bytArr);
+            dw2.Flush();
+            mstr.Position = 0;
+
+            BinaryReader dr2 = new BinaryReader(mstr);
+
+            for (ii = 0; ii < bytArr.Length; ii++)
+            {
+                Assert.Equal(dr2.ReadByte(), bytArr[ii]);
```
stephentoub: These should be reversed such that the expected value comes first, then the actual.

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteByteCharTests.cs
```diff
@@ -0,0 +1,573 @@
...
+
+        /// <summary>
+        /// Cases Tested:
+        ///VSWhidbey #273805
+        ///If someone writes out characters using BinaryWriter's Write(char[]) method, they must use something like BinaryReader's ReadChars(int) method to read it back in.  
+        ///They cannot use BinaryReader's ReadChar().  Similarly, data written using Write(char) can't be read back using ReadChars(int).
+        ///
+        ///A high-surrogate is a Unicode code point in the range U+D800 through U+DBFF and a low-surrogate is a Unicode code point in the range U+DC00 through U+DFFF
+        ///
+        ///UPDATE: 9/21/2006: We dont throw on the second read but then throws continuously - note the loop count difference in the 2 loops
```
stephentoub: Is all of this still needed?

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteByteCharTests.cs
```diff
@@ -0,0 +1,573 @@
...
+                try
+                {
+                    reader.ReadChar();
+
+                    Assert.Equal(i, 1);
+                }
+                catch (ArgumentException)
+                {
+                    // I guess sometimes it throws and othertimes, it doesn't?
+                    // Based on the comments above.
```
stephentoub: Not particularly confidence inspiring ;)

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteTests.cs
```diff
@@ -0,0 +1,744 @@
...
+                sgl2 = dr2.ReadSingle();
+                Assert.Fail("Error_2d847! Expected exception not thrown, sgl2==" + sgl2);
+            }
+            catch (IOException)
+            {
+            }
+            catch (Exception exc)
+            {
+                Assert.Fail("Error_238gy! Unexpected exception thrown, exc==" + exc.ToString());
+            }
```
stephentoub: This could all just be ```Assert.Throws<IOException>(() => dr2.ReadSingle());```, right?

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteTests.cs
```diff
@@ -0,0 +1,744 @@
...
+                dw2.Write(dblArr[ii]);
+
+            dw2.Flush();
+            mstr.Position = 0;
+            BinaryReader dr2 = new BinaryReader(mstr);
+
+            for (ii = 0; ii < dblArr.Length; ii++)
+            {
+                if ((dbl2 = dr2.ReadDecimal()) != dblArr[ii])
+                    Assert.Fail("Error_948yg! Expected==" + dblArr[ii] + ", got==" + dbl2);
```
stephentoub: Does Assert.Fail actually exist in xunit?  Regardless, this would be better as Assert.Equal, anyway, e.g.
```Assert.Equal(dblArr[ii], dr2.ReadDecimal());```

File: src/System.IO/tests/BinaryWriter/BinaryWriter_WriteTests.cs [hunk 13]
```diff
@@ -0,0 +1,744 @@
...
+            {
+                dbl2 = dr2.ReadDecimal();
+                Assert.Fail("EndOfStreamException expected");
+            }
+            catch (EndOfStreamException)
+            { }
+            catch (Exception e)
+            {
+                Assert.Fail("EndOfStreamException expected but got: " + e);
+            }
```
stephentoub: Assert.Throws

src/System.IO/tests/BinaryWriter/BinaryWriter_WriteTests.cs
```diff
@@ -0,0 +1,744 @@
...
+            for (ii = 0; ii < dblArr.Length; ii++)
+                dw2.Write(dblArr[ii]);
+
+            dw2.Flush();
+            mstr.Position = 0;
+
+            for (ii = 0; ii < dblArr.Length; ii++)
+            {
+                if ((dbl2 = dr2.ReadDouble()) != dblArr[ii])
+                    Assert.Fail("Error_948yg! Expected==" + dblArr[ii] + ", got==" + dbl2);
```
stephentoub: Assert.Equal

src/System.IO/tests/MemoryStream/MemoryStream_TryGetBufferTests.cs
```diff
@@ -0,0 +1,428 @@
...
+    [Fact]
+    public static void TryGetBuffer_Constructor_Int32_AlwaysReturnsTrue()
+    {
+        var stream = new MemoryStream(512);
+
+        ArraySegment<byte> _;
+        bool result = stream.TryGetBuffer(out _);
+
+        Assert.True(result);
+    }
```
stephentoub: There's nothing to also assert here about _?

src/System.IO/tests/MemoryStream/MemoryStream_TryGetBufferTests.cs
```diff
@@ -0,0 +1,428 @@
...
+            Assert.True(default(ArraySegment<byte>).Equals(result));
+        }
+    }
+
+    [Fact]
+    public static void TryGetBuffer_Constructor_ByteArray_Bool_AlwaysReturnsEmptyArraySegment()
+    {
+        var arrays = Inputs.GetArraysVariedBySize();
+
+        foreach (byte[] array in arrays)
```
stephentoub: this is ok, though it could also be done as a [Theory] with MemberData.

src/System.IO/tests/Stream/NullTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace StreamTests
+{
+    public class NullCtorTests
+    {
+        [Fact]
+        public static void TestNullStream()
+        {
+            var t = TestNullStream(Stream.Null);
+            t.Wait();
```
stephentoub: It'd be better to do this as:
```C#
[Fact]
public static Task TestNullStream()
{
    return TestNullStream(Stream.Null);
}
```
or
```C#
[Fact]
public static async Task TestNullStream()
{
    await TestNullStream(Stream.Null);
}
```
The way it's currently written could lead to deadlock problems.

src/System.IO/tests/Stream/TimeoutTests.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.IO;
+using Xunit;
+
+namespace StreamTests
+{
+    public class MyStream : Stream
```
stephentoub: Is there a better name we could use for this, something that describes its purpose?

src/System.IO/tests/StringReader/StringReaderCtorTests.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                StringReader sr = new StringReader(null);
+                String strTemp = sr.ReadToEnd();
+                sr.ReadLine();
+                sr.Peek();
+                sr.Read();
+                sr.Dispose();
+            });
```
stephentoub: It's not clear here which line is intended to throw the ArgumentNullException.

stephentoub:
Thanks, Matt.  I skimmed this and left a smattering of comments; in most of the cases, I commented on just one or two occurrences of a pattern that's repeated many times throughout the tests.

FiveTimesTheFun:
Thanks Stephen.  I've pushed some fixes to the BinaryWriter tests.  Of course I agree with all the comments, but these tests were ported first from the old QA system, then to the ToF, and finally here, and I don't believe I'll be able to go through them all.  I'll continue to improve them before we merge them, however.

FiveTimesTheFun:
I'll also rebase my changes before they're merged, so please forgive the commit messages.

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
FiveTimesTheFun: Thank you, I converted the BinaryWriter tests to xunit.

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
FiveTimesTheFun: I can't make this a theory because the Encoding.UTF8, etc, properties are not constant expressions, and I cannot use them as arguments to an attribute.  I just did factor them though.

src/System.IO/tests/BinaryWriter/BinaryWriterTests.cs
FiveTimesTheFun: :)

src/System.IO/tests/StringWriter/StringWriterTests.cs
```diff
@@ -0,0 +1,369 @@
...
+            sw.WriteLine((UInt64.MaxValue));
+
+            Assert.Equal(@"False
+B
+987
+875634
+1.23457
+45634563
+18446744073709551615
+", sw.ToString());
```
stephentoub: Any of these tests (like this one) that are compiled on Windows using verbatim strings that include newlines are going to fail when run on Unix.  When compiled, these verbatim newlines will be encoded as \r\n, but the WriteLines they're being compared against are going to use the newline for that platform, e.g. \n on Unix.  I'd suggest changing these verbatim strings to instead use Environment.NewLine.

src/System.IO/tests/StringWriter/StringWriterTests.cs
```diff
@@ -0,0 +1,369 @@
...
+
+        [Fact]
+        public static async Task MiscWriteLineAsync()
+        {
+            var sw = new StringWriter();
+            await sw.WriteLineAsync('H');
+            await sw.WriteLineAsync(new char[] { 'e', 'l', 'l', 'o' });
+            await sw.WriteLineAsync("World!");
+
+            Assert.Equal("H\r\nello\r\nWorld!\r\n", sw.ToString());
```
stephentoub: This will fail on Unix, due to the wrong newline being used.

src/System.IO/tests/StringWriter/StringWriterTests.cs
```diff
@@ -0,0 +1,369 @@
...
+18446744073709551615
+", sw.ToString());
+        }
+
+        [Fact]
+        public static void TestWriteLineObject()
+        {
+            var sw = new StringWriter();
+            sw.WriteLine(new Object());
+            Assert.Equal("System.Object\r\n", sw.ToString());
```
stephentoub: This will fail on Unix, due to the wrong newline being used.

src/System.IO/tests/StringWriter/StringWriterTests.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+    
+        [Fact]
+        public static void TestWriteLineAsyncCharArray()
+        {
+            StringWriter sw = new StringWriter();
+            sw.WriteLineAsync(new char[] { 'H', 'e', 'l', 'l', 'o' });
+
+            Assert.Equal("Hello\r\n", sw.ToString());
```
stephentoub: This will fail on Unix, due to the wrong newline being used.

stephentoub:
@FiveTimesTheFun, just checking in on this... you still working on getting these merged?  Thanks.

stephentoub:
Replaced by #2696

Issue #915, opened 2015-02-19
Remove uses of Console.WriteLine in the Unit Tests?


ellismg:
A bunch of our unit tests do `Console.WriteLine` either on regular paths or on error cases.  By my count there's bout 1600 hits for `Console.Write*` in our tests.  This is a bit of an over-count because some of these are the Console tests themselves.

I don't really like this very much.  It feels like something we inherited when porting our legacy test assets.  Is it worthwhile to go clean this up?  I don't mind doing the work, just don't want to put in a bunch of effort if it doesn't stand a chance of getting merged.

I'm sure @nguerrera will be interested in this and my understanding was @davkean had some guidance on how to author and port tests and I'd be interested in understanding how this aligns with his thinking.

davkean:
Agreed with this. A good example of this was what we did with the File IO tests. We ripped all logging that was non-essential and just noise, and moved logging that added value only to the failure case.

nguerrera:
Strongly agree too. I got them down to zero at [one point], but folks keep adding them back with each new library. :(

[one point]: https://github.com/dotnet/corefx/pull/456

ellismg:
The real issue here really is that all of our tests were not written with xUnit in mind.  As we do this, we should make sure that if the output of the test was useful then it's captured in some other way (e.g an assert with a good message).

gregoryyoung:
I was getting ready to send a PR in this area but want to make sure this is what is being refered to:

https://github.com/dotnet/corefx/blob/master/src/System.IO.MemoryMappedFiles/tests/MemoryMappedFile/CreateViewAccessor.cs#L501

If so I'll get together the PR

Cheers,

Greg

ellismg:
@gregoryyoung, Yes.  I would expect in cases like what you had here we would just let the unexpected exception go unhandled, and xUnit would catch it and give a nice message.

gregoryyoung:
OK just wanted to check..

Cheers,

Greg

gregoryyoung:
In looking through some are simple some are less so.

As an example here https://github.com/gregoryyoung/corefx/blob/master/src/System.IO.MemoryMappedFiles/tests/MemoryMappedViewStream/ReadWrite.cs

It appears the reason why they are printing is that basically all the tests are being done in a single method (I guess its expected that you would go read the output).

If you just mechanically change this you will also get some pretty annoying behaviour in trying to run this if its broken (first exception stops all tests).

[...]

PR #2299, opened 2015-07-10
changes for system.globalization to xunit

venkat-raman251:
Need to test System.Globalization tests (XUnit) on other platforms.
@ellismg : Can you please do a quick review of the changes (w.r.t integrating with the existing folder structure/files)?

ToDo: A couple of test have been commented out. I will fix them tomorrow along with changes for marking tests which fail in other platforms.


ellismg:
A high level comment.  We should move away from a bunch of csprojs in favor of a single one for all the globalization tests.

Can you figure out a good way to add all of this stuff to the existing csproj?

venkat-raman251:
That was a mistake. I seem to have redundant .csproj files. I have changed the existing .csproj to cover all files. Will remove the additional ones.

src/System.Globalization/tests/CharNumericCategory/GetNumericValue.cs
```diff
@@ -0,0 +1,145 @@
...
+
+                var unicodeStr = frontPadding + Char.ConvertFromUtf32(item.CodeValue) + backPadding;
+                var actualNumericValue = CharUnicodeInfo.GetNumericValue(unicodeStr, frontPadding.Length);
+
+                Assert.Equal(hasValue ? item.NumericValue.Value : -1, actualNumericValue);
+
+                count++;
+            }
+
+            Console.WriteLine("Tested {0} characters.", count);
```
Maxwe11: #915

src/System.Globalization/tests/Common/Factory.cs
```diff
@@ -0,0 +1,49 @@
...
+                    return typeof(Factory).GetTypeInfo().Assembly.GetManifestResourceStream(name);
+                }
+            }
+
+            return null;
+        }
+
+        public static IEnumerable<IUnicodeChar> GetUnicodeData()
+        {
+            return GetUnicodeData(GetUnicodeDataTxt());
```
Maxwe11: Why this not cached?

stephentoub:
Thanks, @venkat-raman251.  I've lost track of what our desired approach is around initial ports like this: should the primary commit be done as dotnet-bot?

src/System.Globalization/tests/CharNumericCategory/GetNumericValue.cs
```diff
@@ -0,0 +1,145 @@
...
+        public void InputStringBeginningOfStringWithValue()
+        {
+            TestStringInputWithIndex(String.Empty, "ending text", true);
+        }
+
+        [Fact]
+        public void InputStringBeginningOfStringWithoutValue()
+        {
+            TestStringInputWithIndex(String.Empty, "ending text", false);
+        }
```
stephentoub: FYI for cases like this, all of these different facts can be condensed into a single theory, e.g.
```C#
[Theory]
[InlineData("", "", false)]
[InlineData("", "", true)]
[InlineData("some text", "more text", false)]
[InlineData("some text", "", true)]
[InlineData("some text", "", false)]
[InlineData("", "ending text", true)]
[InlineData("", "ending text", false)]
public void TestStringInputWithIndex(string frontPadding, string backPadding, bool hasValue)
{
    ...
}
```
This keeps the tests more concise, makes it easy to add additional data inputs as needed, keeps all of the related tests together, and helps to highlight the differences between test cases.

src/System.Globalization/tests/CharNumericCategory/GetNumericValue.cs
```diff
@@ -0,0 +1,145 @@
...
+            foreach (var item in Factory.GetUnicodeData())
+            {
+                // Skip items that have a value
+                if (item.CodeValue < char.MaxValue && item.NumericValue.HasValue == hasValue)
+                {
+                    var actualNumericValue = CharUnicodeInfo.GetNumericValue(checked((char)item.CodeValue));
+
+                    Assert.Equal(hasValue ? item.NumericValue.Value : -1, actualNumericValue);
+
+                    count++;
```
stephentoub: Where is this used?

src/System.Globalization/tests/CharNumericCategory/GetNumericValue.cs
stephentoub: Yes, please remove the remaining uses of Console.WriteLine.... looks like there are only a few in the whole PR, so it should be easy.

src/System.Globalization/tests/CharNumericCategory/GetNumericValueTests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), CLRTest.Common.props))\CLRTest.Common.props" />
+  <PropertyGroup>
+    <CLRTestOwner>tasou</CLRTestOwner>
+    <CLRTestSuite>WEEKLY</CLRTestSuite>
```
stephentoub: Per Matt's comment and your response, I'm assuming all of these .csproj files will be removed. (And if they were to stay, there are things that would need to be removed, like all of this "CLR"-specific stuff.

src/System.Globalization/tests/Common/UnicodeData.cslib
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
stephentoub: Why is all of this in Common?  Is it used by more than the System.Globalization tests?

src/System.Globalization/tests/Common/Unicode_5_1/PropList.txt
```diff
@@ -0,0 +1,1207 @@
+# PropList-5.1.0.txt
+# Date: 2008-03-20, 17:55:27 GMT [MD]
+#
+# Unicode Character Database
+# Copyright (c) 1991-2008 Unicode, Inc.
+# For terms of use, see http://www.unicode.org/terms_of_use.html
```
stephentoub: Just verifying it's ok to commit this to the repo as is?

src/System.Globalization/tests/Common/Unicode_5_1/README.txt
```diff
@@ -0,0 +1,2 @@
+UnicodeData.txt retrieved from http://www.unicode.org/Public/5.1.0/ucd/UnicodeData.txt on 6 March 2014 by tasou
+PropList.txt retrieved from http://www.unicode.org/Public/5.1.0/ucd/PropList.txt on 1 May 2014 by tasou
```
stephentoub: Should this be a .md?  Do we need the user alias included?

src/System.Globalization/tests/CompareInfo/CompareInfoCompare2.cs
```diff
@@ -0,0 +1,101 @@
...
+                string str2 = TestLibrary.Generator.GetString(-55, false, 5, 20);
+                Assert.Equal(0, CultureInfo.CurrentCulture.CompareInfo.Compare(str1, str1, CompareOptions.Ordinal));
+                Assert.Equal(0, CultureInfo.CurrentCulture.CompareInfo.Compare(str2, str2, CompareOptions.Ordinal));
+                TestStrings(str1, str2);
+            }
+        }
+
+        // PosTest4: Specific regression cases
+        [Fact]
+        public void RegressionTests()
```
stephentoub: "Regression" from what / when?  It'd be better if the name of the test described what it was testing rather than why it was added.

src/System.Globalization/tests/CultureInfo/CultureInfoConstructor2.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class CultureInfoConstructor2
+    {
+        [Fact]
+        public void TestDiffCulture()
+        {
+            CultureInfo myCulture = new CultureInfo("");
+            CultureInfo myCultureEn = new CultureInfo("en");
+            CultureInfo myCultureFr = new CultureInfo("fr-FR");
+            CultureInfo myCultureDe = new CultureInfo("de-DE");
```
stephentoub: What is this testing, just that the ctor doesn't throw?  Should it be validating any properties on the resulting CultureInfo?

src/System.Globalization/tests/CultureInfo/CultureInfoGetHashCode.cs
```diff
@@ -0,0 +1,45 @@
...
+        public void PosTest3()
+        {
+            CultureInfo myCultureInfo = CultureInfo.InvariantCulture;
+
+            // the only guarantee that can be made about HashCodes is that they will be the same across calls
+            int actualValue = myCultureInfo.GetHashCode();
+            int expectedValue = myCultureInfo.GetHashCode();
+            Assert.Equal(actualValue, expectedValue);
+        }
+    }
```
stephentoub: In a case like this, you could use MemberData, e.g.
```C#
[Theory]
[MemberData("CultureInfos")]
public void IdempotentGetHashCode(CultureInfo ci)
{
    Assert.Equal(ci.GetHashCode(), ci.GetHashCode());
}

public static IEnumerable<object[]> CultureInfos
{
    get
    {
        yield return new object[] { new CultureInfo("en-US") };
        yield return new object[] { new CultureInfo("en") };
        yield return CultureInfo.InvariantCulture;
    }
}
```
(If it weren't for the InvariantCulture, you could use InlineData with the string name instead.)

src/System.Globalization/tests/CultureInfo/CultureInfoTest.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            Assert.True(CultureInfo.CurrentCulture.Equals(jaCulture));
+            Assert.True(CultureInfo.CurrentUICulture.Equals(jaCulture));
+
+            DateTime dt = new DateTime(2014, 3, 14, 3, 14, 0);
+
+            Assert.True(dt.ToString().Equals(dt.ToString(jaCulture)));
+
+            CultureInfo.CurrentCulture = culture;
+            CultureInfo.CurrentUICulture = uiCulture;
```
stephentoub: This restoring of the original culture should be done inside a finally block; otherwise if one of the above asserts fails and throws, the wrong culture could be left set for the execution of other tests.

stephentoub:
cc: @eerhardt, @steveharter

stephentoub:
@venkat-raman251, thanks for bringing over all of these globalization tests.  I skimmed and left a few general comments about how some of the employed patterns in the tests can be cleaned up with xunit.  Don't feel like you need to block on addressing those, but it'd be great to clean this stuff up at some point, so consider filing an issue to address it.

src/System.Globalization/tests/Common/Unicode_5_1/PropList.txt
venkat-raman251: @stephentoub @ellismg  : This file just lists the Unicode characters. This test (CharUnicode and CharNumericCategory) seem to be using Unicode6.3 though. Do I raise an issue to update the embedded resource UnicodeData.txt to represent Unicode 8?
I do not know the answer to this question regarding the copyright info. It seems to be the readme.txt from http://www.unicode.org/Public/UCD/latest/ReadMe.txt

venkat-raman251: And this readme.txt and PropList.txt isn't required for the tests. Should I just remove them?

src/System.Globalization/tests/CharNumericCategory/GetNumericValue.cs
venkat-raman251: 2 tests call them with different bool values.

src/System.Globalization/tests/CultureInfo/CultureInfoGetHashCode.cs
ellismg: InvariantCulture's string name is "", so you could still adopt inline data if you wanted.

ellismg:
I think there is some way that needs to be done around our use of the Factory type and the UnicodeHelpers in general.  The way this is structured, we embed the UnicodeData.txt file for a specific Unicode version in the test assembly and then at runtime pull it out, parse it and then compare its well known data to what the platform returns.

I am a little hesitant to check this part of the test in.  I don't know about include the UnicodeData files in the repository and I'm not sure if a test assembly per unicode version is the right long term approach.  For now could we roll back the parts that depend on Factory and the UnicodeHelpers\UnicodeData.cslib stuff?  What do you think about this @stephentoub ?

venkat-raman251:
@stephentoub , @ellismg  : I have currently added <UnsupportedPlatforms>Linux;OSX</UnsupportedPlatforms> to the .csproj as there are 179 tests failing on Linux. Shall I merge this PR?

or would you like me mark each one of those tests as [ActiveIssue(846, PlatformID.AnyUnix)] instead ?

venkat-raman251:
@stephentoub , @ellismg : I have marked the failing tests. Can I merge this now?

ellismg:
I'm happy, LGTM.  Thanks again for slogging through all of this.

PR #2305, opened 2015-07-10
PLINQ - Minor fixes 

Clockwork-Muse:
There's a few things bundled in this:

The PLINQ fix for #2239 (`int.MinValue` returned in a comparer).
Adding `checked` to the inner (partition) enumerators for inlined nullable Average operators (only missing from some).

It _also_ contains a change for `Sum(int)`, doing the intermediate summation in a long, to allow extreme values to be added, so long as the final result would be within range.  eg. -
```
new[] { int.MaxValue, int.MaxValue, int.MinValue }.AsParallel().Sum()
```
... currently throws an overflow exception because of the first two items, even though the total is within range (`int.MaxValue - 1`).

I'm not sure of all the potential ramifications of this change, although it looks fairly benign.  Note that I haven't made the corresponding change to vanilla LINQ yet (which would throw too), while I wait for more direction on this particular piece of the PR.


stephentoub:
> It also contains a change for Sum(int), doing the intermediate summation in a long

Let's revert this.  It should first be discussed as an issue, and if the general agreement is to move forward with that, a PR can address it for both LINQ and PLINQ.  With it as part of this PR, it's going to make it difficult to move forward on the other pieces.

I'll hold off on reviewing this further until that's taking care of, as it seems to represent the bulk of the changes in the diff.

Thanks!

Clockwork-Muse:
@stephentoub - done.  New issue is #2316

src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/Inlined/NullableDecimalAverageAggregationOperator.cs
```diff
@@ -118,8 +118,11 @@ protected override bool MoveNextCore(ref Pair<decimal, long> currentElement)
 
                     if (current.HasValue)
                     {
-                        sum += current.GetValueOrDefault();
-                        count++;
+                        checked
+                        {
+                            sum += current.GetValueOrDefault();
+                            count++;
+                        }
```
stephentoub: Hmm.  These checked changes could actually result in visibly different behavior from before, right? i.e. exceptions occurring where none previously occurred and instead values wrapped around?

Clockwork-Muse: Heh.  Unlikely, actually; [you can't use that many elements, remember](http://blogs.msdn.com/b/pfxteam/archive/2012/11/16/plinq-and-int32-maxvalue.aspx)?  (In theory it means we could remove the checking altogether, even from the non-nullable versions, but that makes me nervous.) 
I ran into this with `LongCount`, too.  Of course, having it wrap around would be a pretty bad bug (silent incorrect values)....

hackcraft: What's the performance impact on 32-bit systems?

stephentoub: > Unlikely, actually

I was actually thinkinga about ```sum``` rather than ```count```.  Why is that unlikely?

Clockwork-Muse: @stephentoub - ah.

I didn't think `checked` worked on decimal?  The documentation states "only on integral types".  The keyword also doesn't appear to propagate into method calls either, so this wouldn't affect that anyways (unless there's something special in the compiler?)... and the type performs its own check anyways (throws overflow regardless).

If you mean for nullable int, it lofts to long for the intermediate sum, so wont overflow before hitting the iteration limit anyways.
For nullable long it's possible, but the current behavior is non-deterministic; the worker threads used non-checked, while the final result uses checked, which means whether you get an exception or an incorrect result depends on where the partitioner split your values.  Heisenbug.

@hackcraft - this just brings it in line with the non-nullable version, which uses checked throughout.  Decimal isn't effected by checked (or unchecked), this just makes the counting safe.  I also have no way to test against a 32-bit system - my machine is 64.

stephentoub: > I didn't think checked worked on decimal? 

Sorry I was unclear... I was just referring to all of the tests together and happened to comment on that one. (Decimal itself does overflow checking, regardless of whether you use checked.)

> The keyword also doesn't appear to propagate into method calls either

That's correct; it purely affects the IL contained within the checked block in that method.

> If you mean for nullable int, it lofts to long for the intermediate sum, so wont overflow before hitting the iteration limit anyways.

Ah, you're right.  That's a good point.

> For nullable long it's possible, but the current behavior is non-deterministic; the worker threads used non-checked, while the final result uses checked, which means whether you get an exception or an incorrect result depends on where the partitioner split your values

Ok, that's reasonable.  Thanks.

stephentoub:
@dotnet-bot test this please

src/System.Linq.Parallel/tests/ExchangeTests.cs
```diff
@@ -131,22 +132,28 @@ public static void Merge_Ordered_Longrunning(Labeled<ParallelQuery<int>> labeled
...
         [Theory]
-        [MemberData("MergeData", (object)(new int[] { 16, 1024 }))]
-        [MemberData("ThrowOnCount_AllMergeOptions_MemberData", (object)(new int[] { 16, 1024 }))]
+        [MemberData("ThrowOnCount_AllMergeOptions_MemberData", (object)(new int[] { 4, 8 }))]
         // FailingMergeData has enumerables that throw errors when attempting to perform the nth enumeration.
         // This test checks whether the query runs in a pipelined or buffered fashion.
         public static void Merge_Ordered_Pipelining(Labeled<ParallelQuery<int>> labeled, int count, ParallelMergeOptions options)
         {
-            Assert.Equal(0, labeled.Item.WithMergeOptions(options).Select(x => x).First());
+            int partitions = Environment.ProcessorCount == 1 ? 1 : count - 1;
```
stephentoub: Why the check for ProcessorCount?  What's special about ProcessorCount == 1 in this case?

src/System.Linq.Parallel/tests/QueryOperators/AverageTests.cs
```diff
@@ -77,6 +79,16 @@ public static void Average_Long_Longrunning(Labeled<ParallelQuery<int>> labeled,
         }
 
         [Theory]
+        [MemberData("Ranges", (object)(new int[] { 2 }), MemberType = typeof(UnorderedSources))]
+        public static void Average_Long_Overflow(Labeled<ParallelQuery<int>> labeled, int count)
```
stephentoub: Is there a corresonding test somewhere for Int32 / Nullable<Int32> ?

src/System.Linq.Parallel/tests/QueryOperators/OrderByThenByTests.cs
```diff
@@ -326,6 +326,54 @@ public static void OrderByDescending_NotPipelined_CustomComparator_Longrunning(L
...
+            int prev = int.MinValue;
+            foreach (int i in labeled.Item.OrderBy(x => x, new ExtremeComparer<int>()))
+            {
+                Assert.InRange(i, prev, int.MaxValue);
+                prev = i;
+            }
+        }
+
+        [Theory]
+        [Trait("category", "run")]
```
stephentoub: What is this for?

src/System.Linq.Parallel/tests/ExchangeTests.cs
Clockwork-Muse: I think I feel a little silly about it now.  Stemmed from some of the other tests which check against processor count for default parallelization checks, but that's not likely to be an issue here.  Potentially `count` could be adjusted based on processor count, but since we're messing with DOP anyways...

src/System.Linq.Parallel/tests/QueryOperators/AverageTests.cs
Clockwork-Muse: Can't overflow the addition for Int32, because of the loft to long, you'd run into the iteration limit first, so I didn't initially plan on including it.  I can certainly add it, although it'd be outerloop (take a few minutes to run).

Kicking myself for missing nullable, especially for this change....

src/System.Linq.Parallel/tests/QueryOperators/OrderByThenByTests.cs
Clockwork-Muse: Test related to solving the PLINQ version of #2239 .  Adding comment.

src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/Inlined/NullableDecimalAverageAggregationOperator.cs
stephentoub: Should we change the use of ```checked`` then to surround just the operations we're trying to check for overflow?  Even if that ends up being the only operation that's affected, it'd be nice from a readability / understandability perspective to be clear on what we're checking.

src/System.Linq.Parallel/tests/QueryOperators/AverageTests.cs
stephentoub: > nullable long is in there

Yeah, I'd written ```Int32 / Nullable<Int32>``` in my question, but I'd neglected to use the triple ticks and the generic qualifier got cut off.

> Can't overflow the addition for Int32, because of the loft to long, you'd run into the iteration limit first, so I didn't initially plan on including it. I can certainly add it, although it'd be outerloop (take a few minutes to run).

That's if we were trying to force the overflow to come from this particular checked area.  But the long will eventually be brought back to an int and an overflow will occur, right?  Shouldn't we have a test that verifies that happens (or do we already)?  That should be very quick.

src/System.Linq.Parallel/tests/QueryOperators/OrderByThenByTests.cs
stephentoub: I meant, what's the ```[Trait("category", "run")]``` for?

src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/Inlined/NullableDecimalAverageAggregationOperator.cs
Clockwork-Muse: Do you mean only checking the count for float/double/decimal?  I didn't do it that way initially because then it wouldn't match the non-nullable versions, so we'd probably want to change those in that case.
(I'm assuming you don't mean only checking the addition for long, which is the only thing that can overflow at the moment?  I debated that too, but it made me a little nervous, relying on the inner enumeration limit....)

src/System.Linq.Parallel/tests/QueryOperators/OrderByThenByTests.cs
Clockwork-Muse: (... ah.... "I can read my own code, I swear")
Removing.

src/System.Linq.Parallel/tests/QueryOperators/AverageTests.cs
Clockwork-Muse: We don't have a test like that, no.  I tried writing one like that, but it didn't work, because it casts directly to a double for the final calculation; it never goes back to int.

If you're thinking of `Sum`, you had me split that off to a separate issue ( #2316 ).

src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/Inlined/NullableDecimalAverageAggregationOperator.cs
stephentoub: > Do you mean only checking the count for float/double/decimal?

Yup.

> (I'm assuming you don't mean only checking the addition for long, which is the only thing that can overflow at the moment?

I have the same nagging thoughts, though it also feels wrong to penalize the generated code because of something that could theoretically change in the future.  Were you able to verify that the addition of checked isn't measurable on the resulting performance, due to the other overheads involved?

Clockwork-Muse: I'll see what I can do about performance checks....
(That's what I'm assuming, too, yes)

Clockwork-Muse: @stephentoub - looks like the answer is "probably not measurable" - there's some slight differences in timing, but sometimes the checked version is faster, and even completely removing the checking can be slower (for whatever reason).

I used the code [in this branch](https://github.com/Clockwork-Muse/corefx/tree/PLINQ_Perf), creating another test.
[Current PLINQ results](https://onedrive.live.com/redir?resid=9AD89593899AA409!38191&authkey=!AJv9-cuHQBnbsjA&ithint=file%2cxlsx)
[Full Checking results](https://onedrive.live.com/redir?resid=9AD89593899AA409!38192&authkey=!AGn2D1U0yhu9w6w&ithint=file%2cxlsx)
[Long-addition-only-checking results](https://onedrive.live.com/redir?resid=9AD89593899AA409!38193&authkey=!AKmO0e60BBEw6EQ&ithint=file%2cxlsx) (for a lark).

(I'm aware this might not have been the best way to do this, but it was quick to figure out, so...)

stephentoub:
Thanks.  Could you squash this down a bit (mainly the last two commits seem like they could be merged back into the first)?  Then I'll take one more pass through.

Clockwork-Muse:
@stephentoub - ... did you want me to do the full squash/rebase at this time, or just combine those?

stephentoub:
I was thinking you could just combine those two into the first one (you can tell me if that doesn't make sense).  I didn't notice that anything else needed to be fixed, so it's possible I could just merge after that, but I want to take one more pass through afterwards.  I'd suggest keeping the other two commits separate, as they're logically for different things.

Clockwork-Muse:
Squashed the last two commits (they belong two the earlier first and third commits).  Did not rebase on current master.

Clockwork-Muse:
@stephentoub - failed checks due to a Cryptography test on Windows release, all PLINQ tests passing

stephentoub:
@dotnet-bot test this please

stephentoub:
Looks good. Thanks, @Clockwork-Muse.

PR #2306, opened 2015-07-10
20% faster

alessandromontividiu03:
20% faster way to sum.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1599,15 +1599,10 @@ namespace System.Linq
         public static int Sum(this IEnumerable<int> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            var array = source.ToArray();
-            var length = array.Length - 1;
             int sum = 0;
             checked
             {
-                for (var i = length; i >= 0; i--)
-                {
-                    sum += array[i];
-                }
+                foreach (int v in source) sum += v;
             }
             return sum;
         }
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1599,10 +1599,15 @@ namespace System.Linq
         public static int Sum(this IEnumerable<int> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            var array = source.ToArray();
+            var length = array.Length - 1;
             int sum = 0;
             checked
             {
-                foreach (int v in source) sum += v;
+                for (var i = length; i >= 0; i--)
+                {
+                    sum += array[i];
+                }
             }
             return sum;
         }


pdelvo:
This might be a bit faster in some cases, but it increases the memory consumption from O(1) to O(n). That's not worth it.

hackcraft:
20% faster with which sources? I'd imagine if the source is `TSource[]` there wouldn't be much of an improvement. I'd imagine we could likely find implementations of `IEnumerable<TSource>` for which this is slower. I know this will increase memory consumption from O(1) to O(n) either way.

Clockwork-Muse:
It also might throw exceptions for some peoples' use cases: it's valid in vanilla LINQ to have more than `int.MaxValue` elements.  You're not going to be able to allocate an array anywhere near that size.

Clockwork-Muse:
...besides which, why did you only convert the sum for (non-nullable, at that) int?

stephentoub:
@alessandromontividiu03, thanks for your interest in contributing.

The throughput improvements you're seeing are primarily due to eliminating interface calls.  When one iterates through an ```IEnumerable<T>```, each iteration requires two interface calls: one to MoveNext, and one to Current.  In contrast, iterating through an array doesn't require any such calls.  Your approach to getting such an array is the primary thing folks are pushing back on here, due to the significant overheads involved in such a conversion... this definitely makes things slower when the source isn't an ```ICollection<T>```, and has non-trivial negative impact even when the source is.

Now, if the goal is to special-case array inputs, that can instead be done with code like:
```C#
int total = 0;
int[] arr = source as int[];
if (arr != null)
{
    foreach (int item in arr) total += item;
}
else
{
    foreach (int item in source) total += item;
}
return total;
```
This incurs the cost of a cast as well as code duplication, but it successfully allows the compiler to special-case array inputs and avoid the enumerator allocation as well as all of those interface calls when the input is an array.  I've seen a change like that result in very measurable performance improvements in certain scenarios, and in at least one place where it was measured to be important we do it in corefx (https://github.com/dotnet/corefx/blob/master/src/System.Threading.Tasks.Dataflow/src/Internal/SourceCore.cs#L418-L447).  I'm not convinced it's worthwhile complexity in the case of Sum, but that's a conversation that could be had.  

As it stands, while I do appreciate your contributions, I think this PR should be closed as it currently is.  If someone did push for a change like this in the future, we'd also want to see additional tests first for Sum.

Thanks.

JohanLarsson:
Would be nice if foreach took care of that. IIRC foreach compiles to a for for arrays. 
Probably requires source to be known to be an Array at compiletime.

stephentoub:
> IIRC foreach compiles to for loops for array

Essentially, yes.  That's why I was able to write it as a foreach in my example.

> Would be nice if foreach took care of that.

That would mean in many cases (unless it could prove it wasn't necessary) the compiler would need to expand the codegen to the equivalent of what I wrote out, which means many usages would be unnecessarily bloated and include type casts.

JohanLarsson:
I have no idea if the pros outweigh the cons. Just asked out of curiosity. I can bounce the idea in the Roslyn chat and see what they think.

stephentoub:
@JohanLarsson, sounds good.

PR #2307, opened 2015-07-10
PLINQ - immediate cancellation of ToArray

Clockwork-Muse:
All the query operators have a check for whether they've already been canceled - except `ToArray` (well, and `SequenceEqual`, but that's a different problem).  `ToArray` calls a different method than the other query operators, and it appears to be missing the pre-emptive cancellation check - the methods are otherwise mostly identical.  This change adds the check.

Current behavior of `ToArray` in such a situation is to fire off the worker threads, which each perform one enumeration before checking for cancellation.  This change would prevent the worker threads from being started, so _may_ change use cases for some people.  Note that some operations (`SelectMany`, binaries) already perform the pre-emptive check anyways before enumerating a source, so the current behavior is very much dependent on what consumers call.

(besides the required discussion, will need to be rebased to account for other PRs I've submitted).

### DIFF
--- a/src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/QueryOperator.cs
+++ b/src/System.Linq.Parallel/src/System/Linq/Parallel/QueryOperators/QueryOperator.cs
@@ -209,6 +209,18 @@ namespace System.Linq.Parallel
 
                 QueryResults<TOutput> results = GetQueryResults(querySettings);
 
+                // Top-level preemptive cancellation test.
+                // This handles situations where cancellation has occured before execution commences
+                // The handling for in-execution occurs in QueryTaskGroupState.QueryEnd()
+
+                if (querySettings.CancellationState.MergedCancellationToken.IsCancellationRequested)
+                {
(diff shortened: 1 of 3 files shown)


src/System.Linq.Parallel/tests/Helpers/Enumerables.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace Test
+{
+    internal class ThrowOnEnumeration<T> : IEnumerable<T>
```
stephentoub: This gets reverted once you merge with your other PR, right?

stephentoub:
I'll review again once you've merged with the other PR and eliminated the duplication, but for now LGTM.

src/System.Linq.Parallel/tests/Helpers/Enumerables.cs
Clockwork-Muse: Yeah.  At this point it was "yes/no?".

Clockwork-Muse:
@stephentoub - rebased.

stephentoub:
Thanks. LGTM.

PR #2308, opened 2015-07-10
Two minor path-related changes

stephentoub:
Two minor-path related changes:

- System.IO.Compression.ZipFile was unnecessarily including PathInternals.Windows/Unix.cs

- A System.Runtime test for PathTooLongException was passing but for the wrong reason: on Windows it was testing a path longer than max path, whereas on Unix it happened to be passing because of limitations on directory length in a path.  I've tweaked it to avoid assumptions about max path length.

I found these by scrubbing through all of corefx looking for additional uses of max path.  The only other one now outside of System.IO.FileSystem / System.Runtime.Extensions and the associated interop files is in Semaphore, which limits a semaphore name to be MAX_PATH.  We'll need to address that separately when we determine how best to handle named synchronization primitives on Unix (the limit is unlikely to be the MAX_PATH equivalent on Unix.)

### DIFF
--- a/src/System.IO.Compression.ZipFile/src/System.IO.Compression.ZipFile.csproj
+++ b/src/System.IO.Compression.ZipFile/src/System.IO.Compression.ZipFile.csproj
@@ -24,17 +24,8 @@
   </ItemGroup>
   <!-- Windows -->
   <ItemGroup Condition="'$(TargetsWindows)' == 'true'">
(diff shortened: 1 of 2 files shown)


nguerrera:
LGTM

stephentoub:
Thanks, Nick.

JeremyKuhne:
LGTM

stephentoub:
Thanks, Jeremy.

PR #2310, opened 2015-07-10
Remove Tools dependency from System.Numerics.Vectors.WindowsRuntime

mellinoe:
System.Diagnostics.Tools isn't actually needed in this library at all, and the version used here is actually wrong anyways (package version is 4.0.0, not 4.1.0).


stephentoub:
LGTM

mellinoe:
Thanks, Stephen, gonna merge this as it seems to be messing with re-generating lock files.

PR #2311, opened 2015-07-10
Add test coverage to System.Diagnostics.Process.

Priya91:
In this commit, removed HandleCount implementation, as it's not exposed in
contract. Added some more functional tests. Reorganized the test structure
to inherit from a TestBase. Renamed the tests to match the Given-When-Then
naming scheme.

cc @stephentoub @Sridhar-MS @joshfree @sokket


src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            Environment2.Clear();
+            Assert.Equal(0, Environment2.Count);
+
+            //ContainsKey 
+            Environment2.Add("NewKey", "NewValue");
+            Environment2.Add("NewKey2", "NewValue2");
+            Assert.True(Environment2.ContainsKey("NewKey"));
+            if (global::Interop.IsWindows)
+            {
+                Assert.True(Environment2.ContainsKey("newkey"));
```
stephentoub: Should this instead be:
```C#
Assert.Equal(global::Interop.IsWindows, Environment2.ContainsKey("newkey"));
```
so that on Windows we're validiting it's found and on Unix we're validating it's not?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+            Assert.Equal("NewKeyNewKey2", result);
+            Assert.Equal(2, index);
+
+            //Contains
+            Assert.True(Environment2.Contains(new System.Collections.Generic.KeyValuePair<string, string>("NewKey", "NewValue")));
+            if (global::Interop.IsWindows)
+            {
+                Assert.True(Environment2.Contains(new System.Collections.Generic.KeyValuePair<string, string>("nEwKeY", "NewValue")));
+            }
```
stephentoub: Ditto.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            Assert.Equal("NewValue", stringout);
+            if (global::Interop.IsWindows)
+            {
+                retval = Environment2.TryGetValue("NeWkEy", out stringout);
+                Assert.True(retval);
+                Assert.Equal("NewValue", stringout);
+            }
+
+            stringout = null;
+            retval = false;
```
stephentoub: Not necessary.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            Environment2.Add(new System.Collections.Generic.KeyValuePair<string, string>("NewKey98", "NewValue98"));
+
+            //Indexed
+            string newIndexItem = Environment2["NewKey98"];
+            Assert.Equal("NewValue98", newIndexItem);
+
+            //TryGetValue
+            string stringout = null;
+            bool retval = false;
```
stephentoub: Not necessary.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            retval = Environment2.TryGetValue("NewKey", out stringout);
+            Assert.True(retval);
+            Assert.Equal("NewValue", stringout);
+            if (global::Interop.IsWindows)
+            {
+                retval = Environment2.TryGetValue("NeWkEy", out stringout);
+                Assert.True(retval);
+                Assert.Equal("NewValue", stringout);
+            }
```
stephentoub: Same IsWindows comment as earlier.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            retval = false;
+            retval = Environment2.TryGetValue("NewKey99", out stringout);
+            Assert.Equal(null, stringout);
+            Assert.False(retval);
+
+            //Exception not thrown with invalid key
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                string stringout1 = null;
+                bool retval1 = false;
```
stephentoub: Not necessary... (I'll stop commenting on these.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            );
+            //Remove Item
+            Environment2.Remove("NewKey98");
+            Environment2.Remove("NewKey98");   //2nd occurrence should not assert
+
+            //Exception not thrown with null key
+            Assert.Throws<ArgumentNullException>(() => { Environment2.Remove(null); });
+
+            //"Exception not thrown with null key"
+            Assert.Throws<System.Collections.Generic.KeyNotFoundException>(() => { string a1 = Environment2["1bB"]; });
```
stephentoub: It's not necessary to store it into a string.  We should also just import the namespace, leaving this at:
```C#
Assert.Throws<KeyNotFoundException>(() => Environment2["1bB"]);
```

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.Diagnostics.ProcessTests
+{
+    public class Given_ProcessStartInfoType : ProcessTestBase
+    {
+        [Fact]
+        public void Then_EnvironmentShouldBe()
```
stephentoub: Are we using the GivenWhenThen scheme elsewhere in corefx?  Seems a bit odd to use it here if we're not going to push it everywhere.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
```diff
@@ -0,0 +1,223 @@
...
+            ProcessStartInfo psi = new ProcessStartInfo();
+
+            // Calling the setter
+            try
+            {
+                psi.UseShellExecute = false;
+            }
+            catch (Exception)
+            {
+                Assert.True(false, "ProcessStartInfo_UseShellExecute001 failed");
```
Maxwe11: Stack trace is not enough?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessType.cs
```diff
@@ -122,26 +75,8 @@ public void Process_BasePriorityUnix()
...
-        {
-            p.Start();
-            Sleep();
-            p.Kill();
-            Assert.True(p.WaitForExit(WaitInMS));
-        }
-
         [Fact]
-        public void Process_EnableRaiseEvents()
+        public void Then_EnableRaiseEventsShouldBe()
```
stephentoub: The GivenWhenThen naming seems strange in cases like this.  Most of these test methods are setting up the When (in some cases even setting up the Given, though most of that's done by the initialization earlier), yet the method is named Then_*ShouldBe.  Strange, at least to me.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessType.cs
```diff
@@ -7,81 +7,34 @@
 using System.IO.Pipes;
 using System.Linq;
 using System.Runtime.InteropServices;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace System.Diagnostics.ProcessTests
 {
-    public partial class ProcessTest : IDisposable
+    public class Given_ProcessType : ProcessTestBase
```
stephentoub: What is "ProcessType"?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace System.Diagnostics.ProcessTests
+{
+    public class ProcessTestBase : IDisposable
+    {
+        protected const int WaitInMS = 100 * 1000;
+        protected const string CoreRunName = "corerun";
+        protected const string TestExeName = "System.Diagnostics.Process.TestConsoleApp.exe";
+        protected const int SuccessExitCode = 100;
+
+        protected Process _process;
```
stephentoub: Now every derived test class is going to have its own test process.  Is that what we want?

stephentoub:
Thanks, Priya.  The additional unit tests look good.  I'm a little wary of the ramifications of moving all of the process management into a base class, in particular because it's now going to significantly change how process management for these tests; if you've worked through all that, great.  I'm much less clear on the value of the test renaming... FWIW, it makes things more confusing to me.

Sridhar-MS:
@stephentoub Test renaming is to follow a new convention called [Given-When-Then](http://martinfowler.com/bliki/GivenWhenThen.html) model. We had success in using that model in Vipr. It was easy to understand what the test is testing by filling the Given-When-Then template. We are trying to replicate it in corefx.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Given_ProcessStartInfoType.cs
sokket: This is a super long test that seems like it'll be difficult to debug if (when) something breaks. Could this be broken up into more cases to help with debug-ability?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Interop.cs
```diff
@@ -30,11 +26,37 @@ public struct PROCESS_MEMORY_COUNTERS
...
         [DllImport("api-ms-win-core-processthreads-l1-1-0", CharSet = System.Runtime.InteropServices.CharSet.Unicode, SetLastError = true)]
         public static extern int GetPriorityClass(SafeProcessHandle handle);
 
         [DllImport("kernel32.dll", CharSet = System.Runtime.InteropServices.CharSet.Ansi, SetLastError = true)]
         public static extern SafeProcessHandle GetCurrentProcess();
+
+        [DllImport("api-ms-win-core-processthreads-l1-1-0.dll")]
+        internal static extern int GetCurrentProcessId();
+
+        [DllImport("libc")]
```
sokket: Do we have precedence for mixing Windows and Unix Interop in one file?

stephentoub:
To avoid stalling the additional unit tests and fixes, could we separate out the naming changes as a separate thing to be discussed?

Sridhar-MS:
@stephentoub Makes sense. I will followup on GivenWhenThen model on a separate discussion.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/Interop.cs
Priya91: There are not a lot of interop to do for tests, and having a file for each platform is not worthwhile as the structure is not so complex. Following with what already existed here.

Priya91:
@stephentoub : I have made the base class disposable, so all derived classes will be disposable, and xunit will call the Dispose which does the cleanup. I have also provided a virtual Dispose, so if there is any additional resources that have be released in the derived, then the dervied classes will have to override the dispose and call the base dispose. So by this design change, nothing should change in how the processes are cleaned. Am i overlooking any case here? I am reverting back the name change, to be followed up once the model proposed by @Sridhar-MS  is finalized.

mellinoe:
Regarding removing HandleCount: does anyone remember why it was removed from the contract, or why it's still sticking around? It might make sense to separate that out from the rest of the changes.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs
mellinoe: What is the run time like with this?

Priya91: On my box, they run for 6 sec. None of the tests actually Wait on a thread. We create processes that are supposed to run for infinite (100s) to ensure the process is alive, till we finish testing. Then we kill it in Dispose.

stephentoub:
> So by this design change, nothing should change in how the processes are cleaned. Am i overlooking any case here?

Nothing jumps out at me, but it is changing the fact that previously we had one pool of processes to clean up and now we have multiple pools of processes to clean up.  That in and of itself isn't a problem.  We just spent a good deal of time fighting with these process tests and stability in the CI system, and I want to make sure that doesn't happen again.  If you're confident this is all going to work well, with test classes running in parallel and whatnot, then I'm fine with it.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs
```diff
@@ -0,0 +1,100 @@
...
+                TestExeName + " " + optionalArgument;
+
+            // Profilers / code coverage tools doing coverage of the test process set environment
+            // variables to tell the targeted process what profiler to load.  We don't want the child process 
+            // to be profiled / have code coverage, so we remove these environment variables for that process 
+            // before it's started.
+            p.StartInfo.Environment.Remove("Cor_Profiler");
+            p.StartInfo.Environment.Remove("Cor_Enable_Profiling");
+            p.StartInfo.Environment.Remove("CoreClr_Profiler");
+            p.StartInfo.Environment.Remove("CoreClr_Enable_Profiling");
```
mellinoe: Could you clarify this part? Why don't we want these to be set? These values seem awfully specific / hard-coded to whatever particular instrumentation we are using right now, which worries me a bit.

mellinoe: I see this code removed below so I guess you just moved it around, but it still looks a bit weird to me.

stephentoub: These are the environment variables used by CLR / CoreCLR.

Priya91:
Hmm, I'm confident this change should not cause any problems with parallel running, since we've made sure we don't share any state between processes, and all cleanup is done correctly. If this does break something, it should be an interesting learning experience :)

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs
mellinoe: That makes sense, I initially thought they were something to do with the specific code coverage tool we were using. Does this mean that we don't get code coverage numbers for whatever blocks these test processes hit?

stephentoub: > Does this mean that we don't get code coverage numbers for whatever blocks these test processes hit?

Yes.

PR #2313, opened 2015-07-10
Move Windows File I/O helper methods into common.

JeremyKuhne:
This facilitates further IO work around long paths.

@stephentoub @nguerrera


src/Common/src/Interop/Windows/mincore/Interop.CopyFile.cs
```diff
@@ -0,0 +1,23 @@
...
+            int cancel = 0;
+            if (!Interop.mincore.CopyFileEx(src, dst, IntPtr.Zero, IntPtr.Zero, ref cancel, copyFlags))
+            {
+                return Marshal.GetLastWin32Error();
+            }
+
+            return Interop.mincore.Errors.ERROR_SUCCESS;
+        }
+    }
+}
```
nguerrera: Several files missing final newline.

JeremyKuhne: Hard habit to break- updating.

src/Common/src/Interop/Windows/mincore/Interop.CopyFile.cs
```diff
@@ -0,0 +1,23 @@
...
+        internal static int CopyFile(string src, string dst, bool failIfExists)
+        {
+            int copyFlags = failIfExists ? Interop.mincore.FileOperations.COPY_FILE_FAIL_IF_EXISTS : 0;
+            int cancel = 0;
+            if (!Interop.mincore.CopyFileEx(src, dst, IntPtr.Zero, IntPtr.Zero, ref cancel, copyFlags))
+            {
+                return Marshal.GetLastWin32Error();
+            }
+
+            return Interop.mincore.Errors.ERROR_SUCCESS;
```
stephentoub: I'm not understanding the benefit of returning the error like this, rather than just returning the Boolean result and letting the caller do with it as they will (as we do with other interop).  Case in point, it looks like there's a single caller of this method, and it only cares whether it succeeded or not.

src/Common/src/Interop/Windows/mincore/Interop.SafeCreateFile.cs
```diff
@@ -0,0 +1,45 @@
...
+        internal static readonly IntPtr INVALID_HANDLE_VALUE = new IntPtr(-1);  // WinBase.h
+
+        /// <summary>
+        /// Does not allow access to non-file devices. This disallows DOS devices like "con:", "com1:",
+        /// "lpt1:", etc.  Use this to avoid security problems, like allowing a web client asking a server
+        /// for "http://server/com1.aspx" and then causing a worker process to hang.
+        /// </summary>
+        [System.Security.SecurityCritical]  // auto-generated
+        internal static SafeFileHandle SafeCreateFile(
+            String lpFileName,
```
stephentoub: Nit: string instead of String

src/Common/src/Interop/Windows/mincore/Interop.SafeCreateFile.cs
```diff
@@ -0,0 +1,45 @@
...
+internal partial class Interop
+{
+    internal partial class mincore
+    {
+        internal static readonly IntPtr INVALID_HANDLE_VALUE = new IntPtr(-1);  // WinBase.h
+
+        /// <summary>
+        /// Does not allow access to non-file devices. This disallows DOS devices like "con:", "com1:",
+        /// "lpt1:", etc.  Use this to avoid security problems, like allowing a web client asking a server
+        /// for "http://server/com1.aspx" and then causing a worker process to hang.
```
stephentoub: What kind of hang is this referring to?

src/Common/src/Interop/Windows/mincore/Interop.UnsafeCreateFile.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.IO;
+using Microsoft.Win32.SafeHandles;
+
+internal partial class Interop
+{
+    internal partial class mincore
+    {
+        [System.Security.SecurityCritical]  // auto-generated
+        internal static SafeFileHandle UnsafeCreateFile(
```
stephentoub: This exists just to add an "Unsafe" prefix to something that already exists?  Do we really need this, or can the rest of the code just use CreateFile and SafeCreateFile?

src/Common/src/Interop/Windows/mincore/WinRT/Interop.UnsafeCreateFile.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
+
+internal partial class Interop
+{
+    internal partial class mincore
+    {
+        internal static unsafe SafeFileHandle UnsafeCreateFile(
```
stephentoub: I see... related to my previous comment on the other UnsafeCreateFile, it exists because this one exists?  I'm wondering if this one should just be renamed to CreateFile and the other UnsafeCreateFile removed.

stephentoub:
cc: @ericstj

src/Common/src/Interop/Windows/mincore/Interop.SafeCreateFile.cs
JeremyKuhne: No idea- just pulled this comment along.

src/Common/src/Interop/Windows/mincore/Interop.UnsafeCreateFile.cs
JeremyKuhne: Transitional- the logic for long paths will probably have to go here as the concept doesn't appear to exist for WinRT. Want to get some basic scenarios up and working and then move away from the other. It is also possible that I might change the signature here as I try implementing.

src/Common/src/Interop/Windows/mincore/Interop.SafeCreateFile.cs
JeremyKuhne: Thanks, I'll fix this- and I'll pay closer attention to existing code I'm moving for this sort of thing.

src/Common/src/Interop/Windows/mincore/Interop.UnsafeCreateFile.cs
jasonwilliams200OK: @JeremyKuhne, regarding long paths, I found some useful implementations in this library: https://github.com/peteraritchie/LongPath (`Pri.LongPath` package @NuGet).

stephentoub:
@JeremyKuhne, what's the status of this? Thanks.

JeremyKuhne:
@stephentoub Shifted focus a bit- did you have any blocking feedback?

stephentoub:
> did you have any blocking feedback?

Nothing blocking per se, but at the same time it seems like if we're going to move this code around, we should fix issues with it as part of doing so, especially if the goal of moving it to Common is so that additional code can depend on it.

I think most of the issues I called out were preexisting, so if you want to go ahead and merge, please do so, but it'd be really good to revisit this stuff before further code uses it.

JeremyKuhne:
I'll follow up with functional changes in another PR.

PR #2319, opened 2015-07-11
Avoid unnecessary allocations in Win32 file system enumeration

justinvp:
Changes:

- Lazily create and cache `SearchResultHandler<TSource>` instances
- Only allocate `_searchStack` when needed (for `SearchOptions.AllDirectories`)
- Remove the unnecessary `SearchOption` field on `SearchData`
- Replace `SearchData` class with unified `PathPair` struct that was used for Unix. 
- Remove the `SearchResult` class
- Pass `WIN32_FIND_DATA` by ref instead of copying (its not a small struct)
- Various minor cleanup
 - Remove unused usings
 - Remove trailing whitespace
 - Remove uses of `this.`
 - `String` -> `string`

I left the individual commits to make it easier to review, but I intend to squash before merging.

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
@@ -1,83 +1,88 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
-using System.Globalization;
 using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using System.Security;
-using System.Text;
-using System.Threading;
 
-using Microsoft.Win32;
 using Microsoft.Win32.SafeHandles;
 
 namespace System.IO
 {
     // Overview:
     // The key methods instantiate Win32FileSystemEnumerableIterators. These compose the iterator with search result
-    // handlers that instantiate the FileInfo, DirectoryInfo, String, etc. The handlers then perform any
-    // additional required permission demands. 
+    // handlers that instantiate the FileInfo, DirectoryInfo, string, etc. The handlers then perform any
+    // additional required permission demands.
     internal static class Win32FileSystemEnumerableFactory
     {
-        internal static IEnumerable<String> CreateFileNameIterator(String path, String originalUserPath, String searchPattern,
+        internal static IEnumerable<string> CreateFileNameIterator(string path, string originalUserPath, string searchPattern,
                                                                     bool includeFiles, bool includeDirs, SearchOption searchOption)
         {
             Contract.Requires(path != null);
             Contract.Requires(originalUserPath != null);
             Contract.Requires(searchPattern != null);
 
-            SearchResultHandler<String> handler = new StringResultHandler(includeFiles, includeDirs);
-            return new Win32FileSystemEnumerableIterator<String>(path, originalUserPath, searchPattern, searchOption, handler);
+            SearchResultHandler<string> handler;
+
+            if (includeFiles && includeDirs)
+            {
+                handler = SearchResultHandler.FileSystemPath;
+            }
+            else if (includeFiles)
+            {
+                handler = SearchResultHandler.FilePath;
+            }
+            else
+            {
+                Debug.Assert(includeDirs, "Should never be excluding both files and directories.");
+                handler = SearchResultHandler.DirectoryPath;
+            }
+
+            return new Win32FileSystemEnumerableIterator<string>(path, originalUserPath, searchPattern, searchOption, handler);
         }
 
-        internal static IEnumerable<FileInfo> CreateFileInfoIterator(String path, String originalUserPath, String searchPattern, SearchOption searchOption)
+        internal static IEnumerable<FileInfo> CreateFileInfoIterator(string path, string originalUserPath, string searchPattern, SearchOption searchOption)
         {
             Contract.Requires(path != null);
             Contract.Requires(originalUserPath != null);
             Contract.Requires(searchPattern != null);
 
-            SearchResultHandler<FileInfo> handler = new FileInfoResultHandler();
-            return new Win32FileSystemEnumerableIterator<FileInfo>(path, originalUserPath, searchPattern, searchOption, handler);
+            return new Win32FileSystemEnumerableIterator<FileInfo>(path, originalUserPath, searchPattern, searchOption, SearchResultHandler.FileInfo);
         }
 
-        internal static IEnumerable<DirectoryInfo> CreateDirectoryInfoIterator(String path, String originalUserPath, String searchPattern, SearchOption searchOption)
+        internal static IEnumerable<DirectoryInfo> CreateDirectoryInfoIterator(string path, string originalUserPath, string searchPattern, SearchOption searchOption)
         {
             Contract.Requires(path != null);
             Contract.Requires(originalUserPath != null);
             Contract.Requires(searchPattern != null);
 
-            SearchResultHandler<DirectoryInfo> handler = new DirectoryInfoResultHandler();
-            return new Win32FileSystemEnumerableIterator<DirectoryInfo>(path, originalUserPath, searchPattern, searchOption, handler);
+            return new Win32FileSystemEnumerableIterator<DirectoryInfo>(path, originalUserPath, searchPattern, searchOption, SearchResultHandler.DirectoryInfo);
         }
 
-        internal static IEnumerable<FileSystemInfo> CreateFileSystemInfoIterator(String path, String originalUserPath, String searchPattern, SearchOption searchOption)
+        internal static IEnumerable<FileSystemInfo> CreateFileSystemInfoIterator(string path, string originalUserPath, string searchPattern, SearchOption searchOption)
         {
             Contract.Requires(path != null);
             Contract.Requires(originalUserPath != null);
             Contract.Requires(searchPattern != null);
 
-            SearchResultHandler<FileSystemInfo> handler = new FileSystemInfoResultHandler();
-            return new Win32FileSystemEnumerableIterator<FileSystemInfo>(path, originalUserPath, searchPattern, searchOption, handler);
+            return new Win32FileSystemEnumerableIterator<FileSystemInfo>(path, originalUserPath, searchPattern, searchOption, SearchResultHandler.FileSystemInfo);
         }
     }
 
     // Overview:
     // Enumerates file system entries matching the search parameters. For recursive searches this
     // searches through all the sub dirs and executes the search criteria against every dir.
-    // 
+    //
     // Generic implementation:
-    // Win32FileSystemEnumerableIterator is generic. When it gets a WIN32_FIND_DATA, it calls the 
-    // result handler to create an instance of the generic type. 
-    // 
+    // Win32FileSystemEnumerableIterator is generic. When it gets a WIN32_FIND_DATA, it calls the
+    // result handler to create an instance of the generic type.
+    //
     // Usage:
-    // Use Win32FileSystemEnumerableFactory to obtain FSEnumerables that can enumerate file system 
-    // entries as String path names, FileInfos, DirectoryInfos, or FileSystemInfos.
-    // 
+    // Use Win32FileSystemEnumerableFactory to obtain FSEnumerables that can enumerate file system
+    // entries as string path names, FileInfos, DirectoryInfos, or FileSystemInfos.
+    //
     // Security:
     // For all the dirs/files returned, demands path discovery permission for their parent folders
     internal class Win32FileSystemEnumerableIterator<TSource> : Iterator<TSource>
@@ -87,11 +92,11 @@ namespace System.IO
         private const int STATE_FIND_NEXT_FILE = 3;
         private const int STATE_FINISH = 4;
 
-        private SearchResultHandler<TSource> _resultHandler;
-        private List<SearchData> _searchStack;
-        private SearchData _searchData;
-        private String _searchCriteria;
-        [System.Security.SecurityCritical]
+        private readonly SearchResultHandler<TSource> _resultHandler;
+        private List<PathPair> _searchList;
+        private PathPair _searchData;
+        private readonly string _searchCriteria;
+        [SecurityCritical]
         private SafeFindHandle _hnd = null;
 
         // empty means we know in advance that we won’t find any search results, which can happen if:
@@ -102,14 +107,14 @@ namespace System.IO
         // searchPattern at each directory level.
         private bool _empty;
 
-        private String _userPath;
-        private SearchOption _searchOption;
-        private String _fullPath;
-        private String _normalizedSearchPath;
-        private uint _oldMode;
+        private readonly string _userPath;
+        private readonly SearchOption _searchOption;
+        private readonly string _fullPath;
+        private readonly string _normalizedSearchPath;
+        private readonly uint _oldMode;
 
-        [System.Security.SecuritySafeCritical]
-        internal Win32FileSystemEnumerableIterator(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler<TSource> resultHandler)
+        [SecuritySafeCritical]
+        internal Win32FileSystemEnumerableIterator(string path, string originalUserPath, string searchPattern, SearchOption searchOption, SearchResultHandler<TSource> resultHandler)
         {
             Contract.Requires(path != null);
             Contract.Requires(originalUserPath != null);
@@ -119,9 +124,7 @@ namespace System.IO
 
             _oldMode = Interop.mincore.SetErrorMode(Interop.mincore.SEM_FAILCRITICALERRORS);
 
-            _searchStack = new List<SearchData>();
-
-            String normalizedSearchPattern = NormalizeSearchPattern(searchPattern);
+            string normalizedSearchPattern = NormalizeSearchPattern(searchPattern);
 
             if (normalizedSearchPattern.Length == 0)
             {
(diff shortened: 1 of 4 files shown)


src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -1,83 +1,91 @@
...
+
+            if (includeFiles && includeDirs)
+            {
+                handler = SearchResultHandler.FileSystemPath;
+            }
+            else if (includeFiles)
+            {
+                handler = SearchResultHandler.FilePath;
+            }
+            else if (includeDirs)
```
stephentoub: Since this is an internal method and a problem with includeFiles / includeDirs would indicate a bug in our code rather than something we need to be throwing an exception for, instead of:
```C#
else if (includeDirs)
{
    handler = SearchResultHandler.DirectoryPath;
}
else
{
    Debug.Fail(...);
    throw ...;
}
```
how about just:
```C#
else
{
    Debug.Assert(includeDirs, "Should never be excluding both files and directories);
    handler = SearchResultHandler.DirectoryPath;
}
```
?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -381,19 +387,20 @@ public override bool MoveNext()
...
         {
             string findData_fileName = findData.cFileName;
             Contract.Requires(findData_fileName.Length != 0 && !Path.IsPathRooted(findData_fileName),
                 "Expected file system enumeration to not have empty file/directory name and not have rooted name");
 
-            String userPathFinal = Path.Combine(localSearchData.userPath, findData_fileName);
-            String fullPathFinal = Path.Combine(localSearchData.fullPath, findData_fileName);
-            return new SearchResult(fullPathFinal, userPathFinal, findData);
+            string userPathFinal = Path.Combine(_searchData.UserPath, findData_fileName);
+            string fullPathFinal = Path.Combine(_searchData.FullPath, findData_fileName);
```
stephentoub: Can these Path.Combines be pushed into the handler's IsResultIncluded such that these strings are only created if, well, the result is included?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
         {
-            _includeFiles = includeFiles;
-            _includeDirs = includeDirs;
+            get
+            {
+                if (s_filePath == null)
+                {
+                    s_filePath = new StringResultHandler(includeFiles: true, includeDirs: false);
+                }
+                return s_filePath;
```
stephentoub: Just personal preference, but FWIW I often write things like this as:
```C#
return s_filePath ?? (s_filePath = new StringresultHandler(includefiles: true, includeDirs: false));
```

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
+
+            result = null;
+            return false;
         }
     }
 
-    // Holds search data that is passed around 
-    // in the heap based stack recursion
-    internal sealed class SearchData
+    internal class FileSystemInfoResultHandler : SearchResultHandler<FileSystemInfo>
```
stephentoub: Can we seal all of these specialized handlers?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
-        public readonly SearchOption searchOption;
+            result = null;
+            return false;
+        }
     }
 
-    internal sealed class SearchResult
+    // Holds search data that is passed around
+    // in the heap based stack recursion
+    internal struct SearchData
```
stephentoub: With this now a struct and with only the two fields, it's now the equivalent of the ```PathPair``` type in the Unix version of this: https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs#L527-L537 .  It'd be nice to separate this out and unify them at some point.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
+        internal override bool IsResultIncluded(string fullPath, string userPath, ref Interop.mincore.WIN32_FIND_DATA findData, out string result)
         {
-            String name = result.FullPath;
-            IFileSystemObject fileSystemObject = new Win32FileSystem.Win32FileSystemObject(name, result.FindData, asDirectory: true);
-            DirectoryInfo di = new DirectoryInfo(name, fileSystemObject);
-            return di;
+            bool includeFile = _includeFiles && Win32FileSystemEnumerableHelpers.IsFile(ref findData);
+            bool includeDir = _includeDirs && Win32FileSystemEnumerableHelpers.IsDir(ref findData);
+            Debug.Assert(!(includeFile && includeDir), findData.cFileName + ": current item can't be both file and dir!");
+            if (includeFile || includeDir)
```
stephentoub: It's minor, so don't feel obliged to change it, but we're computing includeDir even if includeFile is true.  It could be avoided with either:
```C#
if ((_includeFiles && Win32FileSystemEnumerableHelpers.IsFile(ref findData)) ||
    (_includeDirs && Win32FileSystemEnumerableHelpers.IsDir(ref findData)))
{
    ...
}
```
or something like that.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
+            bool isFile = Win32FileSystemEnumerableHelpers.IsFile(ref findData);
+            bool isDir = Win32FileSystemEnumerableHelpers.IsDir(ref findData);
+            Debug.Assert(!(isFile && isDir), findData.cFileName + ": current item can't be both file and dir!");
 
-            this.fullPath = fullPath;
-            this.userPath = userPath;
-            this.searchOption = searchOption;
-        }
+            if (isFile)
+            {
```
stephentoub: Same comment as above about computing isDir.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
+            _userPath = userPath;
         }
 
-        internal String UserPath
+        public string FullPath
         {
-            get { return _userPath; }
+            get
+            {
+                Debug.Assert(_fullPath != null && _fullPath.Length > 0);
```
stephentoub: ```Debug.Assert(!string.IsNullOrEmpty(_fullPath));``` ?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,171 +517,224 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
         }
 
-        internal Interop.mincore.WIN32_FIND_DATA FindData
+        public string UserPath
         {
-            [System.Security.SecurityCritical]
             get
-            { return _findData; }
+            {
+                Debug.Assert(_userPath != null && _userPath.Length > 0);
```
stephentoub: ```Debug.Assert(!string.IsNullOrEmpty(_userPath));``` ?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -492,13 +496,13 @@ private static String GetNormalizedSearchCriteria(String fullSearchString, Strin
...
 
-        private static String GetFullSearchString(String fullPath, String searchPattern)
+        private static string GetFullSearchString(string fullPath, string searchPattern)
         {
             Contract.Requires(fullPath != null);
             Contract.Requires(searchPattern != null);
 
             PathHelpers.ThrowIfEmptyOrRootedPath(searchPattern);
-            String tempStr = Path.Combine(fullPath, searchPattern);
+            string tempStr = Path.Combine(fullPath, searchPattern);
```
stephentoub: Since searchPattern isn't null or empty, Path.Combine isn't going to effect the end of searchPattern, right?  Meaning whether tempStr ends in a directory / volume separator is the same as whether searchPattern ends in a directory / volume separator?  Assuming yes, we can avoid another string allocation here in the case where we need to append the *, e.g. something like:
```C#
private static string GetFullSearchString(string fullPath, string searchPattern)
{
    ...
    PathHelpers.ThrowIfEmptyOrRootedPath(searchPattern);
    char lastChar = searchPattern[searchPattern.Length - 1];
    return PathHelpers.IsDirectorySeparator(lastChar) || lastChar == Path.VolumeSeparatorChar ?
        Path.Combine(fullPath, searchPattern, "*") :
        Path.Combine(fullPath, searchPattern);
}
```
Are there any cases with that I'm not thinking about?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -192,39 +198,38 @@ private void CommonInit()
...
                     }
                 }
             }
-            // for AllDirectories, we first recurse into dirs, so cleanup and add searchData 
+            // for AllDirectories, we first recurse into dirs, so cleanup and add searchData
             // to the stack
             else
             {
                 _hnd.Dispose();
+                _searchStack = new List<SearchData>();
```
stephentoub: Do you know why we're using a ```List<T>``` instead of a ```Stack<T>```?  With most of the operations happening at the 0th element of the list, we're going to incur a bunch of unnecessary moves in the list to continually shift elements around.

Seems like the only non-stack-ish operation being performed is that when iterating through a directory, it looks like we're maintaining a FIFO order of the elements added, just for that set.  If that's not important, seems we should just switch to using a ```Stack<T>```.  If that is important, we could stick with a ```List<T>``` but at least operate at the other end of the list, to avoid all of those shift operations; in the case of the FIFO ordering, we could still just Add everything but then Reverse just the portion that was added to restore that order.

Note that by changing SearchData from a class to a struct, we've changed the amount of data stored in the list, such that each element is now the size of two references rather than just one, so moves will become more expensive.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -130,25 +136,25 @@ internal Win32FileSystemEnumerableIterator(String path, String originalUserPath,
...
-                String searchPatternDirName = Path.GetDirectoryName(normalizedSearchPattern);
-                String userPathTemp = originalUserPath;
+                string searchPatternDirName = Path.GetDirectoryName(normalizedSearchPattern);
+                string userPathTemp = originalUserPath;
                 if (searchPatternDirName != null && searchPatternDirName.Length != 0)
                 {
                     userPathTemp = Path.Combine(userPathTemp, searchPatternDirName);
                 }
-                this._userPath = userPathTemp;
+                _userPath = userPathTemp;
```
stephentoub: Again just personal preference, but if it were me I'd have written the last six lines instead as:
```C#
_userPath = string.IsNullOrEmpty(searchPatternDirName) ?
    originalUserPath :
    Path.Combine(originalUserPath, searchPatternDirName);
```

stephentoub:
cc: @ericstj

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: I noticed this as well. It's using `List<T>` purely for historical reasons since this code was originally in `mscorlib.dll` and `Stack<T>` was in `System.dll`.

The original behavior (pre .NET 4) ended up producing a very strange ordering that was very different from the default order of `dir /s`. When we added the `Enumerate*` methods in .NET 4, we deliberately changed the previous behavior to match more closely with the default order of `dir /s` (although, the order is not guaranteed). I guess it wasn't implemented as efficiently has it could have been. We were able to make this change in .NET 4 without it being a breaking change because the order isn't guaranteed and its stated so in the docs. We should be able to change the order again, but my preference would be to maintain the current behavior as its a nicer ordering.

I went ahead and updated the code to operate at the end of the list, to avoid all the shifting. For adding the subdirectories, I changed it the way you suggested, adding each directory at the end and then calling Reverse on just that portion that was added. Note: I found that `List<PathPair>.Reverse` was slow (it looks like the underlying implementation in `Array.Reverse` ends up using a code path in this case that boxes each swapped element), so I implemented a custom non-boxing Reverse. The custom Reverse is ~28x faster than `List<PathPair>.Reverse`.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Changed.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Yeah, cool. Done.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: I like this better too. Changed.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Done. Also, I moved these to be private inner classes.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Done.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Done.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Done.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: I'm assuming you're correct, but I didn't have a chance to investigate myself. If I don't have a chance to get to it for this PR, I'll open an up-for-grabs issue to track.

justinvp:
I made changes based on your feedback, @stephentoub. Thanks for the initial review.

justinvp:
Some data: in addition to the reduction in allocations, this is ~10% faster on my machine doing `Directory.EnumerateFileSystemEntries(path, "*", SearchOption.AllDirectories)` (for 100 iterations) where `path` is the corefx\src directory.

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
stephentoub: Interesting.  It looks like Array.Reverse tries to special case some types of T, but will fail for all but a known set of primitive types (https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/classlibnative/bcltype/arrayhelpers.cpp#L422-L424).  It'd be nice if in the future we could expose some more generic methods on Array (e.g. ```Reverse<T>```), but in the interim it seems like we should just rewrite ```List<T>.Reverse``` to not use ```Array.Reverse``` (either by just reimplementing ```List<T>.Reverse``` or by adding an internal ```Array.Reverse<T>``` that ```List<T>.Reverse``` could use... in the age of generics, there doesn't seem to be anything particularly magical about the ```TrySZReverse``` implementation in the runtime, anyway. @jkotas, any reason not to do this?

justinvp: I'm going to go ahead and submit a PR for this by adding an internal `Array.Reverse<T>` and making `List<T>.Reverse` use it.

justinvp: https://github.com/dotnet/coreclr/pull/1231

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -454,12 +456,32 @@ private void AddSearchableDirsToStack(SearchData localSearchData)
             }
         }
 
-        private static String NormalizeSearchPattern(String searchPattern)
+        // This is faster than using List<T>.Reverse because it avoids
+        // boxing in List<T>.Reverse's implementation due to the use of
+        // a custom struct for T.
+        private void ReverseSearchList(int index)
```
stephentoub: We'll remove this once https://github.com/dotnet/coreclr/pull/1231 is merged, right?

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
```diff
@@ -513,182 +535,155 @@ private static String GetFullSearchString(String fullPath, String searchPattern)
...
+
+            [SecurityCritical]
+            internal override bool IsResultIncluded(string fullPath, string userPath, ref Interop.mincore.WIN32_FIND_DATA findData, out string result)
             {
-                Debug.Assert(isFile);
-                IFileSystemObject fileSystemObject = new Win32FileSystem.Win32FileSystemObject(name, result.FindData, asDirectory: false);
-                FileInfo fi = new FileInfo(name, fileSystemObject);
-                return fi;
+                if (_includeFiles && Win32FileSystemEnumerableHelpers.IsFile(ref findData) ||
+                    _includeDirs && Win32FileSystemEnumerableHelpers.IsDir(ref findData))
```
stephentoub: I know the parens aren't strictly needed here due to && having higher precedence than ||, but I still find it much easier to understand if there are parents around each grouping:
```C#
if ((_includeFiles && ...) ||
    (_includeDirs && ...))
{
    ...
}
```

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: I was going to just go ahead and update this PR to remove it now under the assumption that dotnet/coreclr#1231 would be merged. Unless you think it'd be better to wait a little bit for the coreclr change to propagate?

stephentoub: I don't think we need to wait until the binaries make it to the relevant NuGet packages, but I think we should at least wait until the change is merged.

stephentoub:
LGTM.  @ericstj, can you review as well?

whoisj:
:heart_eyes: this is beautiful...

src/System.IO.FileSystem/src/System/IO/Win32FileSystemEnumerable.cs
justinvp: Removed, now that dotnet/coreclr#1231 has been merged.

justinvp:
@ericstj, can you review?

ericstj:
LGTM, sorry for the delay.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #2322, opened 2015-07-12
[WIP] Add Linq methods keyed on expressions: MaxBy, DistinctBy, etc.

### LINKED ISSUE #2146: Add commonly required Enumerable methods (DistinctBy, ExceptBy, AsChunked, ...)
We have Distinct, Except, Intersect, Union, etc. Often it is required, though, to base the set operation on some key. OrderBy can do that for example.

I request that the following methods be added:
- DistinctBy (pick the first element deterministically for each key)
- ExceptBy (return all items of the first sequence that do not have a matching key in the second)
- IntersectBy (return the first item in a.Concat(b) for each unique key where the key appears in both collections)
- UnionBy (return the first item in a.Concat(b) for each unique key)
- MaxBy/MinBy

The set operations need to deal with duplicate keys in a sensible way. I have added to that list my idea for how to do that in sane and well-defined way.

<hr>

Another commonly required helper method is one that splits a sequence into chunks of a fixed size:

```
static IEnumerable<IEnumerable<T>> AsChunked<T>(this IEnumerable<T> sequence, int chunkSize)
```

This is useful for many things, for example:
- Split a list into pages
- Make AsParallel operate on coarser items by feeding it chunks
- If you want to save a huge stream of items to files you can use this to split the sequence into one chunk per file
- Request rows from the database in chunks of 1000 IDs

This requirement comes up on Stack Overflow a lot and there is no built-in way to do it.

The chunks should be drawn lazily. In fact the items of each chunk should be lazy as well. This allows for very big chunks such as 1 billion items per chunk.

Maybe we need two variants:
1. `IEnumerable<IEnumerable<T>> AsChunkedLazy`
2. `IEnumerable<T[]> AsChunkedEager`

The second one would be easier to handle for consumers. The first one requires more disciplined use because each chunk must be enumerated at most once and the chunks must be enumerated in order. This is required for an efficient streaming operation.

hackcraft:
Fixes #2146 if it is indeed considered desirable functionality.

Add MaxBy, MinBy, DistinctBy, ExceptBy, IntersectBy, UnionBy, and Chunk
methods to both Queryable (extension methods applied to IQueryable<T>)
and ‎Enumerable (extension methods applied to IEnumerable<T>).

The methods are fully consistent with each other, giving the same result
as observed from the outside, but by different means. The Queryable
methods are all delivered in terms of existing methods, and so will be
supported by existing providers that. In particular care is taken to
ensure that new methods that don't use IComparer<T> don't call into
Queryable methods that do, as they are relatively less-well supported.

The Queryable methods test for the query provider being EnumerableQuery and
in that case pass to the enumerable version, being better optimised for that
provider. Note that this introduces a project dependency of
System.Linq.Queryable.csproj and System.Linq.Queryable.Tests.csproj on
System.Linq.csproj


hackcraft:
I've marked a few places where I could see arguments for a different approach as FIXME.

hackcraft:
Is there any way to know if the CI failure is due solely to my changing the interface or if there's actually a compiler error or test failure that didn't appear on my machine?

akoeplinger:
@hackcraft you can click on the "Details" link next to the build status and dive into the Jenkins jobs that get triggered. In this case, the following tests failed:

```
System.Linq.Tests.EnumerableTests.MinIntByString
Tests.Queryable_Tests.MatchSequencePattern
```

src/System.Linq.Queryable/src/System.Linq.Queryable.csproj
```diff
@@ -32,5 +32,12 @@
   <ItemGroup>
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\System.Linq\src\System.Linq.csproj">
```
weshaggard: Why did you need a project reference for this?

weshaggard:
@hackcraft thank you for your interest in contributing but I don't believe these are ready for PR yet. I realize issue #2146 is discussing these new public APIs but we need to follow the [API Review Process](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md) on the issue before submitting a PR for adding new public APIs.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2498,6 +2499,337 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
         {
             return Enumerable.Average(Enumerable.Select(source, selector));
         }
+
+        public static TSource MaxBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector)
+        {
+            return source.MaxBy(keySelector, null);
+        }
```
JohanLarsson: I prefer pasing Comparer<TKey>.Default over null here even if the result is the same. Think it reads better.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2498,6 +2499,337 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
...
+        {
+            if (source == null) throw Error.ArgumentNull("source");
+            if (keySelector == null) throw Error.ArgumentNull("keySelector");
+            if (comparer == null) comparer = Comparer<TKey>.Default;
+            TSource value = default(TSource);
+            if (value == null)
+            {
+                using (IEnumerator<TSource> e = source.GetEnumerator())
+                {
+                    do
```
JohanLarsson: No check for empty sequence here?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2498,6 +2499,337 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
...
+                    }
+                }
+            }
+
+            return value;
+        }
+
+        public static IEnumerable<TSource> DistinctBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector)
+        {
+            return source.DistinctBy(keySelector, null);
```
JohanLarsson: I have not had any coffee yet but does null get a value later? I prefer passing EqualityComparer<T>.Default same as above for readability and then comparer can be checked for null in the overload.

src/System.Linq.Queryable/src/System.Linq.Queryable.csproj
hackcraft: I couldn't get a build using `msbuild /t:RebuildAndTest` to reference it without it. Perhaps I am not building in the most appropriate way? Or not referencing in the most appropriate way?

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: I agree. However the rest of the file passes `null` in such cases, and I wanted to be consistent with that.

src/System.Linq.Queryable/src/System.Linq.Queryable.csproj
akoeplinger: You need to add it to project.json.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: On the next line an empty sequence would cause `null` to be immediately returned, which is the correct behaviour for an empty sequence of a nullable type, to be consistent with e.g. `new string[0].Max()` and `new DateTime?[0].Max()`, etc. which return `null`. Similarly the code path for non-nullable types throw an `InvalidOperationException` to be consistent with out `new DateTime[0].Max()` throws, etc.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: It does, an again I ruled consistency with what is done in existing code as more important than my own preference, which would also be to pass `EqualityComparer<T>.Default` explicitly.

hackcraft:
> I don't believe these are ready for PR yet

Forgive someone who is a newbie to the project. I don't believe they're ready for PR either (I'm not convinced either way on whether such methods should be added to Linq or live in an external project). I do though think that it can sometimes be more fruitful to discuss possible additions with a working example (for certain values of "working", since this is failing some tests). Is there a conventional way to do that outside of a PR? Just point to the branch in my own repository?

src/System.Linq/src/System/Linq/Enumerable.cs
JohanLarsson: TIL, ty sir.

src/System.Linq.Queryable/src/System.Linq.Queryable.csproj
hackcraft: I'd thought that, but it was already in there. (I'm sure I'm doing something dim here).

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: There is incidentally a technical reason to favour passing null in cases like this, as it increases the chances of the method falling below the jitter's threshold for inlining if you consistently favour null over property access, because sometimes the few IL instructions can make the difference.

JohanLarsson: For perf it would probably make sense to check if source is T[] or IReadonlyList<T> and use a for loop.

hackcraft: On the one hand you could further and check separately and `for` looped separately because the compiler optimises for loops on `T[]` differently to `IList<T>`. On the other, the increase cost in the extra check (and the increased amount of code to potentially have bugs) may not be worth it. I'm inclined to lean toward the simpler approach.

src/System.Linq.Queryable/src/System.Linq.Queryable.csproj
hackcraft: Ah. I've got it. Don't have how to not have to manually overwrite packages though (is there somewhere that's good for these newb-to-project questions?).

hackcraft:
@stephentoub As per your point-of-order on the ToHashSet, should we close this and leave the branch on my fork for discussion to focus on?

stephentoub:
That would be helpful, yes.  Thanks, @hackcraft.  I do appreciate your starting all of these discussions and providing starting implementations, and hopefully at least some of these APIs will find their way into the tree. (Also just FYI that when we do merge in new APIs, we do it into the future branch rather than into the master branch.)

PR #2325, opened 2015-07-12
Add tests for bulk removal from empty collections

AArnott:
Immutable set and list collections have distinct behaviors in their batch remove methods. These tests describe these differences and ensure that these publicly visible behaviors do not change unknowingly.


AArnott:
Of course we should have a discussion on this around whether the right change is the addition of the test for a lack of an optimization, or adding the missing optimization (which would constitute a publicly observable behavior change).

stephentoub:
LGTM.

I think the right thing is to check this in. Then we can potentially decide to take the risk and add the optimization and update the test accordingly.

PR #2326, opened 2015-07-12
More tests for enumerable.

hackcraft:
Tests originally commited as part of pull-request #2237
moved out into separate pull-request.

### DIFF
--- a/src/System.Linq/tests/MinMaxTests.cs
+++ b/src/System.Linq/tests/MinMaxTests.cs
@@ -1,0 +1,822 @@
+﻿using System.Collections.Generic;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class MinMaxTests
+    {
+        [Fact]
+        public void MinInt32()
+        {
+            var one = Enumerable.Range(1, 10).ToArray();
+            var minusTen = new[] { -1, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000, 100, 200, 1000 };
+            Assert.Equal(42, Enumerable.Repeat(42, 1).Min());
+            Assert.Equal(1, one.Min());
+            Assert.Equal(-10, minusTen.Min());
+            Assert.Equal(100, hundred.Min());
+            Assert.Equal(int.MinValue, one.Concat(Enumerable.Repeat(int.MinValue, 1)).Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<int>().Min());
+        }
+
+        [Fact]
+        public void MaxInt32()
+        {
+            var ten = Enumerable.Range(1, 10).ToArray();
+            var minusTen = new[] { -100, -15, -50, -10 };
+            var thousand = new[] { -16, 0, 50, 100, 1000 };
+            Assert.Equal(42, Enumerable.Repeat(42, 1).Max());
+            Assert.Equal(10, ten.Max());
+            Assert.Equal(-10, minusTen.Max());
+            Assert.Equal(1000, thousand.Max());
+            Assert.Equal(int.MaxValue, thousand.Concat(Enumerable.Repeat(int.MaxValue, 1)).Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<int>().Max());
+        }
+
+        [Fact]
+        public void MinInt64()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (long)i).ToArray();
+            var minusTen = new[] { -1L, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000L, 100, 200, 1000 };
+            Assert.Equal(42, Enumerable.Repeat(42L, 1).Min());
+            Assert.Equal(1, one.Min());
+            Assert.Equal(-10, minusTen.Min());
+            Assert.Equal(100, hundred.Min());
+            Assert.Equal(long.MinValue, one.Concat(Enumerable.Repeat(long.MinValue, 1)).Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<long>().Min());
+        }
+
+        [Fact]
+        public void MaxInt64()
+        {
+            var ten = Enumerable.Range(1, 10).Select(i => (long)i).ToArray();
+            var minusTen = new[] { -100L, -15, -50, -10 };
+            var thousand = new[] { -16L, 0, 50, 100, 1000 };
+            Assert.Equal(42, Enumerable.Repeat(42L, 1).Max());
+            Assert.Equal(10, ten.Max());
+            Assert.Equal(-10, minusTen.Max());
+            Assert.Equal(1000, thousand.Max());
+            Assert.Equal(long.MaxValue, thousand.Concat(Enumerable.Repeat(long.MaxValue, 1)).Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<long>().Max());
+        }
+
+        [Fact]
+        public void MinSingle()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (float)i).ToArray();
+            var minusTen = new[] { -1F, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000F, 100, 200, 1000 };
+            Assert.Equal(42F, Enumerable.Repeat(42F, 1).Min());
+            Assert.Equal(1L, one.Min());
+            Assert.Equal(-10L, minusTen.Min());
+            Assert.Equal(100L, hundred.Min());
+            Assert.Equal(float.MinValue, one.Concat(Enumerable.Repeat(float.MinValue, 1)).Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<float>().Min());
+        }
+
+        [Fact]
+        public void MaxSingle()
+        {
+            var ten = Enumerable.Range(1, 10).Select(i => (float)i).ToArray();
+            var minusTen = new[] { -100F, -15, -50, -10 };
+            var thousand = new[] { -16F, 0, 50, 100, 1000 };
+            Assert.Equal(42F, Enumerable.Repeat(42F, 1).Max());
+            Assert.Equal(10F, ten.Max());
+            Assert.Equal(-10F, minusTen.Max());
+            Assert.Equal(1000F, thousand.Max());
+            Assert.Equal(float.MaxValue, thousand.Concat(Enumerable.Repeat(float.MaxValue, 1)).Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<float>().Max());
+        }
+
+        [Fact]
+        public void MinDouble()
+        {
+            var one = Enumerable.Range(1, 10).Select(i => (double)i).ToArray();
+            var minusTen = new[] { -1D, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000D, 100, 200, 1000 };
+            Assert.Equal(42D, Enumerable.Repeat(42D, 1).Min());
+            Assert.Equal(1D, one.Min());
+            Assert.Equal(-10D, minusTen.Min());
+            Assert.Equal(100D, hundred.Min());
+            Assert.Equal(double.MinValue, one.Concat(Enumerable.Repeat(double.MinValue, 1)).Min());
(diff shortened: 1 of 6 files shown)


src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        [Fact]
+        public void MinInt()
+        {
+            var one = Enumerable.Range(1, 10).ToArray();
+            var minusTen = new[] { -1, -10, 10, 200, 1000 };
+            var hundred = new[] { 3000, 100, 200, 1000 };
+            Assert.Equal(1, one.Min());
+            Assert.Equal(-10, minusTen.Min());
+            Assert.Equal(100, hundred.Min());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<int>().Min());
```
stephentoub: In all of these min/max cases, it'd be good to have some more border cases tests, e.g. a source with one element, a source where the minimum value is MinValue, etc.

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+            var minusTen = new[] { -100L, -15, -50, -10 };
+            var thousand = new[] { -16L, 0, 50, 100, 1000 };
+            Assert.Equal(10, ten.Max());
+            Assert.Equal(-10, minusTen.Max());
+            Assert.Equal(1000, thousand.Max());
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Empty<long>().Max());
+        }
+
+        [Fact]
+        public void MinSingle()
```
stephentoub: The names of these functions are mixing the C# keywords with the type names, e.g. int and long vs Single.  Ideally they'd be consistent with one or the other, e.g. Int32, Int64, Single vs Int, Long, Float

src/System.Linq/tests/MinMaxTests.cs
```diff
@@ -0,0 +1,373 @@
...
+            var nine = Enumerable.Range(1, 10).Select(i => i.ToString()).ToArray();
+            var agents = new[] { "Alice", "Bob", "Charlie", "Eve", "Mallory", "Victor", "Trent" };
+            var confusedAgents = new[] { null, "Charlie", null, "Victor", "Trent", null, "Eve", "Alice", "Mallory", "Bob" };
+            Assert.Equal("9", nine.Max());
+            Assert.Equal("Victor", agents.Max());
+            Assert.Equal("Victor", confusedAgents.Max());
+            Assert.Null(Enumerable.Empty<string>().Max());
+        }
+    }
+}
```
stephentoub: It'd be good to also have coverage for other sources besides arrays.

Also, I assume it was an explicit choice not to include tests for the Min/Max functions that take selector delegates?  That's fine, we'll just need to ensure that happens later and before any changes are submitted to those overloads.

src/System.Linq/tests/SingleTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        [Fact]
+        public void DefaultOnEmpty()
+        {
+            Assert.Equal(0, Enumerable.Empty<int>().SingleOrDefault());
+        }
+
+        [Fact(Skip="Runs for thousands of years. Then throws wrong exception")]
+        public void FailOnManyMatches()
+        {
+            Assert.Throws<InvalidOperationException>(() => Enumerable.Range(0, int.MaxValue).Single(i => i > 0));
```
stephentoub: I think I mentioned this in the previous PR, but I'd prefer to see this done differently, testing the current behavior and not being disabled, e.g.
```
[Fact]
public void FailOnManyMatches()
{
    int invocations = 0;
    Assert.Throws<InvalidOperationException>(() => Enumerable.Range(1, 10).Single(i => i > 0));
    Assert.Equal(10, invocations);
}
```
If we decide to take a visible behavioral change that introduces short circuiting, then the test can be updated at the same time, but at least this will help to flag such behavioral changes when they're attempted.

src/System.Linq/tests/SingleTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            Assert.Equal(42, Enumerable.Repeat(42, 1).Single());
+        }
+
+        [Theory]
+        [InlineData(1, 100)]
+        [InlineData(42, 100)]
+        public void FindSingleMatch(int target, int range)
+        {
+            Assert.Equal(target, Enumerable.Range(0, range).Single(i => i == target));
```
stephentoub: Similarly here, it be good to track the number of times the delegate is invoked and ensure it matches range.

src/System.Linq/tests/SkipTests.cs
```diff
@@ -0,0 +1,90 @@
+﻿using System.Collections.Generic;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Linq.Tests
+{
+    public class SkipTests
+    {
+        private readonly ITestOutputHelper _output;
```
stephentoub: Where is this used?

src/System.Linq/tests/SkipTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            _output = output;
+        }
+        [Fact]
+        public void SkipSome()
+        {
+            Assert.Equal(Enumerable.Range(10, 10), Enumerable.Range(0, 20).Skip(10));
+        }
+
+        [Fact]
+        public void SkipExcessive()
```
stephentoub: What about the case of skipping none with Skip(0)?

src/System.Linq/tests/SkipTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            var source = Enumerable.Range(-2, 5).Select(i => (decimal)i).Select(m => 1 / m).Skip(4);
+            using(var en = source.GetEnumerator())
+            {
+                Assert.True(
+                    Assert.Throws<DivideByZeroException>(() => en.MoveNext())
+                    // Check stack-trace still references the WhereSelectEnumerableIterator`2.MoveNext() call
+                    // where it originated. Check will have to be changed if the details of `Select()`
+                    // implementation change accordingly.
+                    .StackTrace.Contains("at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()"));
```
stephentoub: I'd suggest removing this.  There are a variety of reasons this could be faulty.

src/System.Linq/tests/SkipTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [Fact]
+        public void SkipErrorWhenSourceErrors()
+        {
+            var source = Enumerable.Range(-2, 5).Select(i => (decimal)i).Select(m => 1 / m).Skip(4);
+            using(var en = source.GetEnumerator())
+            {
+                Assert.True(
+                    Assert.Throws<DivideByZeroException>(() => en.MoveNext())
```
stephentoub: It'd be good to have a similar verification for the case where the predicate passed to SkipWhile throws.

stephentoub:
Thanks for adding these tests.

src/System.Linq/tests/MinMaxTests.cs
hackcraft: Agreed. Shall add some more.

src/System.Linq/tests/MinMaxTests.cs
hackcraft: The result of thinking the type names should be used, but failing to do that because I'm used to the C# names. I'll change them to the type names.

src/System.Linq/tests/MinMaxTests.cs
hackcraft: At the time I was writing tests to hit everything I touched in the rest of that PR, and I didn't touch them. I'll add copies of these that hit an identity selector.

src/System.Linq/tests/SingleTests.cs
hackcraft: This probably shouldn't be here at all, at this point. I think I'll cut this, and then when it comes to discussing the short-circuiting add tests showing both the potential for breaking current code if making the change but also that as-is the results can be surprising (that `.Where(selector).Single()` behaves differently to `.Single(selector)` is surprising). When a decision is made on that, the tests that pass on the behaviour it was decided to go with can then be added to enforce that decision either way.

src/System.Linq/tests/SingleTests.cs
hackcraft: Similarly, I think I'll add one with the proposal for short-circuiting complaining if it short-circuits and one complaining if it doesn't, to aid consideration of which behaviour is desirable.

src/System.Linq/tests/SkipTests.cs
hackcraft: Agreed. I wanted to be able to say "hey, and the change doesn't even affect stack traces on error" about one of the change in the original proposal, but such a quality is a nice-to-have about a change rather than vital; I'd say checking on stack traces in code would be in the deserves-to-lose category, and hence a change that broke it in the "only because every change is a breaking change" category.

stephentoub:
cc: @VSadov

src/System.Linq/tests/MinMaxTests.cs
stephentoub: Thanks :smile:

src/System.Linq/tests/SingleTests.cs
stephentoub: The nice thing about adding them now is that it locks in the behavior and alerts us to it ever changing, on purpose or accidentally.  Kind of like the Assert.Same/NotSame tests you suggested adding around ToArray.  Then any PR that did want to change the short-circuiting behavior would also need to update the test; lack of doing so would make the contributor aware of something changing in case they weren't already, and the fact that the changed tests would then need to be in a passing PR would make the reviewer(s) aware of the potential break as well.

stephentoub: (That said, I'm ok if you choose not to do so in this PR; the rest of the tests are goodness even without that.)

hackcraft: Since #2350 has a set of tests that work before the change and fail after, and a set of tests that fail before the change and work after, then whatever decision is made there's a set of tests that will enforce that decision.

The tests here couldn't be used as-is anyway; take off the `Skip` and all they would test is how well your CI server holds up against people trying to use tests to DoS it!

stephentoub: > The tests here couldn't be used as-is anyway

I know, that's why I suggested an alternative in my previous comments :wink:

> Since #2350 has a set of tests that work before the change and fail after...

Yes, but that PR also contains the associated product changes, and we may decide not to take them, in which case the tests would be lost with it.  My point is that it's better to have tests in first so that we know when a PR then comes along that changes the behavior.

hackcraft: Got you. I was more thinking that adding them now is premature because I'm more used to workflows where one doesn't test for behaviour one hasn't decided is or isn't desirable. I'll add them to this PR later (one a phone right now).

stephentoub: Thanks, @hackcraft.

> I'm more used to workflows where one doesn't test for behaviour one hasn't decided is or isn't desirable

Yup, completely understand.  We have to worry about an additional category, though, that of backwards compatibility.  That's lessened at the moment for .NET Core, but it's still an issue, and we'll often have tests that verify the current behavior regardless of whether it's a desirable behavior, because someone could have taken a dependency on it and we at least want to know and be thoughtful about any changes that could have a visible behavioral impact.

hackcraft: I meant more that I'm used to "we've decided this particular current behaviour must be maintained as backward-compatibility in the future" first, add test after. In any case, the tests are now in.

src/System.Linq/tests/ShortCircuitingTests.cs
```diff
@@ -0,0 +1,222 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using Xunit;
+
+namespace System.Linq.Tests
```
stephentoub: Thanks for adding these.

Nit: Other than this one blank line, looks like something may have happened with blank lines elsewhere in the file, i.e. there aren't any :smile:  Would you mind adding some between members, as in the other files?

hackcraft: Certainly.

stephentoub:
Thanks!

PR #2328, opened 2015-07-12
More informative failure from Queryable_Tests.MatchSequencePattern()

hackcraft:
When experimenting with #2322 for a while I hit a snag with
Queryable_Tests.MatchSequencePattern. While this does Console.Write()
the error cases, that's not helpful if you don't hit the problem
locally, but do on the CI build.

This change makes the failure message more directly informative. It
leaves the old assertion in place, so no flaw in this change should
hide a bug that the test would previously have caught.

### DIFF
--- a/src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
+++ b/src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
@@ -1789,7 +1789,8 @@ namespace Tests
                     Console.WriteLine(m);
                 }
             }
-
+            Assert.Equal(Enumerable.Empty<string>(), list.Select(mi => mi.Name));
+            Assert.Equal(Enumerable.Empty<string>(), list2.Select(mi => mi.Name));
             Assert.True(list.Count == 0 && list2.Count == 0);
         }
 


src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
```diff
@@ -1789,7 +1789,8 @@ public static void MatchSequencePattern()
                     Console.WriteLine(m);
                 }
             }
-
+            Assert.Equal(Enumerable.Empty<string>(), list.Select(mi => mi.Name));
+            Assert.Equal(Enumerable.Empty<string>(), list2.Select(mi => mi.Name));
```
stephentoub: Are the Console.WriteLines still useful, or does this render them obsolete?

stephentoub:
> While this does Console.Write() the error cases, that's not helpful if you don't hit the problem locally, but do on the CI build.

FWIW, these writelines are visible in the CI log:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/1106/console
```
07:55:58 Enumerable methods not defined by Queryable
07:55:58 
07:55:58 System.Collections.Generic.IEnumerable`1[TSource] IntersectByIterator[TSource,TKey](System.Collections.Generic.IEnumerable`1[TSource], System.Collections.Generic.IEnumerable`1[TSource], System.Func`2[TSource,TKey], System.Collections.Generic.IEqualityComparer`1[TKey])
07:55:58 System.Collections.Generic.IEnumerable`1[TSource] DistinctIterator[TSource,TKey](System.Collections.Generic.IEnumerable`1[TSource], System.Func`2[TSource,TKey], System.Collections.Generic.IEqualityComparer`1[TKey])
07:55:58 System.Collections.Generic.IEnumerable`1[System.Linq.IGrouping`2[System.Int32,TSource]] ChunkIterator[TSource](System.Collections.Generic.IEnumerable`1[TSource], Int32)
07:55:58 System.Collections.Generic.IEnumerable`1[System.Linq.IGrouping`2[System.Int32,TSource]] ChunkListIterator[TSource](System.Collections.Generic.IList`1[TSource], Int32)
```
That said, I'm happy to see such asserts added, ideally instead of the Assert.True and associated Console.WriteLines if nothing is lost by doing so.

hackcraft:
Well, if only I could have found that earlier!

I suggest adding this anyway. If there's a consensus that the Console.WriteLine and/or the final assertion should be removed, I'll remove them. Otherwise there's no harm leaving them. Personally I'd be inclined to cut out the Console.Write and leave the final assertion as it's always safer to leave a correct assertion, but maybe someone cares about that Console.Write.

stephentoub:
> Well, if only I could have found that earlier!

Yeah :smile: Jenkins has the logs for the individual runs that make up the total build nested a few levels deep, so it can be difficult to find if you don't know where to look.

> I suggest adding this anyway.

Completely agree, my question wasn't about whether the asserts should be added (they should :smile:), but whether they then make the writelines superfluous so that we can delete the writelines.  Many of the tests in the corefx repo were ported from older tests written years ago in a more legacy style and use (or at least used) lots of writelines... we've been trying to get rid of them.  Those left in the repo primarily only execute on failure paths, but even those would be nice to replace with better asserts.

> If there's a consensus that the Console.WriteLine

I'll take a look separately to see whether there's any value in the writelines... if there isn't, I'll remove them.

Thanks!

PR #2330, opened 2015-07-13
PLINQ - Correct type used in constructor.

Clockwork-Muse:
For some reason, PLINQ's version of `Zip` references `IEnumerable` for the second operator, instead of `ParallelQuery`.  Since it was immediately casting (pretty much), it's unlikely this had any negative effects, but it's something to correct.


hackcraft:
Does this change what happens if a parallel source is zipped with a non-parallel source?

Clockwork-Muse:
@hackcraft - You can't call any of the binary operations with a non-parallel second source; [there are specific overloads taking `IEnumerable<T>` as the second operation that throw an exception](https://github.com/dotnet/corefx/blob/master/src/System.Linq.Parallel/src/System/Linq/ParallelEnumerable.cs#L682-L705).

stephentoub:
Wasn't hurting anything, but it's fine to correct.  LGTM.

PR #2333, opened 2015-07-13
Fix System.Security.Cryptography.RSA handle leaks

### LINKED ISSUE #2229: System.Security.Cryptography.RSA: All SafeHandles are not correctly released.
The SafeHandles declared in SafeCryptoHandles.cs are not released correctly in `ReleaseHandle`.

* `SafeProvHandle` should call `CryptReleaseContext` but just sets the handle to `IntPtr.Zero`
* `SafeKeyHandle` should call `CryptDestroyKey` but does nothing.
* `SafeHashHandle` should call `CryptDestroyHash` but does nothing.

Instead of fixing these implementations the classes from mscorlib in coreclr ([`System.Security.Cryptography.SafeProvHandle`](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Security/Cryptography/SafeCryptoHandles.cs#L20) etc.) should be used instead.
Looking at the comments here in corefx the classes seem to be copied from that source and modified.

The CLR actually has [special support](https://github.com/dotnet/coreclr/blob/master/src/classlibnative/cryptography/cryptography.h#L157) for these handles, so we should use that.

AtsushiKan:
Fixes #2229. Internal SafeHandles inside System.Security.Cryptography.RSA were ported
from closed source with their ReleaseHandle() implementations commented out
and never finished. This delta finishes them.

### DIFF
--- a/src/System.Security.Cryptography.RSA/src/System/Security/Cryptography/SafeCryptoHandles.cs
+++ b/src/System.Security.Cryptography.RSA/src/System/Security/Cryptography/SafeCryptoHandles.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Runtime.CompilerServices;
 using System.Runtime.Versioning;
+using Internal.NativeCrypto;
 
 namespace System.Security.Cryptography
 {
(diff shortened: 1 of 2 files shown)


AtsushiKan:
Talking to myself...

src/System.Security.Cryptography.RSA/src/System/Security/Cryptography/SafeCryptoHandles.cs
```diff
@@ -247,7 +266,8 @@ internal static SafeHashHandle InvalidHandle
         [SecurityCritical]
         protected override bool ReleaseHandle()
         {
-            //FreeHash(handle);
+            bool successfullyFreed = CapiHelper.CryptDestroyHash(handle);
+            Debug.Assert(successfullyFreed);
```
stephentoub: In all of these cases, should we be returning successfullyFreed instead of the hardcoded true?

stephentoub:
cc: @bartonjs 

One question on the changes, but otherwise LGTM.

bartonjs:
I updated the comment to include the magical autoresolving keywords for the issue.

The way that the interop was done doesn't follow the guidelines, but does match the (Windows version of the) project, so it's fine; while it's creating a tiny bit more work for the cleanup issue it's not creating a new wrong pattern to also have to clean up.

LGTM.

Issue #1061, opened 2015-03-01
MemoryMappedFile CreateFromFileTestCases failed in CI


stephentoub:
```
Discovering: MemoryMappedFile.Tests
  Discovered:  MemoryMappedFile.Tests
  Starting:    MemoryMappedFile.Tests
  ERR999: Unexpected exception in runTest, System.IO.IOException: The process cannot access the file 'd:\j\workspace\dotnet_corefx_windows_debug_prtest\bin\tests\Windows_NT.AnyCPU.Debug\MemoryMappedFile.Tests\aspnetcore50\CreateFromFile_test2_5f7a364a-e0be-4a98-abcd-67568018dfa7.txt' because it is being used by another process.
     at System.IO.Win32FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs)
     at System.IO.Win32FileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
     at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
     at System.IO.FileStream..ctor(String path, FileMode mode)
     at CreateFromFile.RunTest() in d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.IO.MemoryMappedFiles\tests\MemoryMappedFile\CreateFromFile.cs:line 385
     CreateFromFile.CreateFromFileTestCases [FAIL]
        One or more test cases failed.
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.IO.MemoryMappedFiles\tests\MemoryMappedFile\CreateFromFile.cs(34,0): at CreateFromFile.CreateFromFileTestCases()
  Finished:    MemoryMappedFile.Tests
  
  === TEST EXECUTION SUMMARY ===
     MemoryMappedFile.Tests  Total: 7, Errors: 0, Failed: 1, Skipped: 0, Time: 1.641s
```

joshfree:
@venkat-raman251 could you investigate?

stephentoub:
Fixed by #2334

Issue #560, opened 2015-01-30
2 tests fail when run build.cmd


cnblogs-dudu:
Discovering: MemoryMappedFile.Tests
Discovered:  MemoryMappedFile.Tests
Starting:    MemoryMappedFile.Tests
Err445! Got unexpected exception: System.IO.IOException: There is not enough space on the disk.
  
     at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateCore(SafeFileHandle fileHandle, String mapName, HandleInheritability inheritability, MemoryMappedFileAccess access, MemoryMappedFileOptions options, Int64 capacity) in e:\Dev\GitHub\DotNet\corefx\src\System.IO.MemoryMappedFiles\src\System\IO\MemoryMappedFiles\MemoryMappedFile.Windows.cs:line 45
     at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateFromFile(String path, FileMode mode, String mapName, Int64 capacity, MemoryMappedFileAccess access) in e:\Dev\GitHub\DotNet\corefx\src\System.IO.MemoryMappedFiles\src\System\IO\MemoryMappedFiles\MemoryMappedFile.cs:line 180
     at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateFromFile(String path, FileMode mode, String mapName, Int64 capacity) in e:\Dev\GitHub\DotNet\corefx\src\System.IO.MemoryMappedFiles\src\System\IO\MemoryMappedFiles\MemoryMappedFile.cs:line 107
     at CreateViewAccessor.RunTestLargeCapacity(Int64 capacity) in e:\Dev\GitHub\DotNet\corefx\src\System.IO.MemoryMappedFiles\tests\MemoryMappedFile\CreateViewAccessor.cs:line 378
  Err445! Got unexpected exception: System.IO.IOException: There is not enough space on the disk.
  
     at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateCore(SafeFileHandle fileHandle, String mapName, HandleInheritability inheritability, MemoryMappedFileAccess access, MemoryMappedFileOptions options, Int64 capacity) in e:\Dev\GitHub\DotNet\corefx\src\System.IO.MemoryMappedFiles\src\System\IO\MemoryMappedFiles\MemoryMappedFile.Windows.cs:line 45
[...]

PR #2334, opened 2015-07-13
Rewrite MemoryMappedFiles tests

### LINKED ISSUE #928: Rewrite System.IO.MemoryMappedFiles tests
According to the current code coverage report, this library is only at 77%.  Our general goal is closer to 90%, though this will depend on the library.  We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

stephentoub:
More details are in each commit's description, but at a high-level this PR completely rewrites the memory-mapped files tests.  Along the way I found a few bugs in the Unix implementation, and this PR fixes those.  I also found some dead code, which this PR removes.

Fixes #928, #1061, #560


stephentoub:
cc: @ianhays, @ellismg, @ericstj, @FiveTimesTheFun

ellismg:
Thanks Steve!  I did a pass over the tests and left some feedback, nothing major.  Might be good to wait for another set of eyes before merging however.

stephentoub:
Thanks, Matt!  I addressed your feedback.

src/System.IO.MemoryMappedFiles/tests/MemoryMappedFile.CreateFromFile.Tests.cs
```diff
@@ -0,0 +1,810 @@
...
+        /// </summary>
+        [Theory]
+        [InlineData(FileAccess.ReadWrite, MemoryMappedFileAccess.Read)]
+        [InlineData(FileAccess.ReadWrite, MemoryMappedFileAccess.ReadWrite)]
+        [InlineData(FileAccess.ReadWrite, MemoryMappedFileAccess.CopyOnWrite)]
+        [InlineData(FileAccess.Read, MemoryMappedFileAccess.Read)]
+        [InlineData(FileAccess.Read, MemoryMappedFileAccess.CopyOnWrite)]
+        public void FileAccessAndMapAccessCombinations_Valid(FileAccess fileAccess, MemoryMappedFileAccess mmfAccess)
+        {
+            const int Capacity = 4096;
```
ianhays: You've got Capacity separately declared in multiple locations as 4096. Might be better to just move it up with a name like DefaultCapacity so it's only declared in one place.

src/System.IO.MemoryMappedFiles/tests/MemoryMappedViewAccessor.Tests.cs
```diff
@@ -0,0 +1,469 @@
...
+        [Theory]
+        [InlineData(0, 8192)]
+        [InlineData(8100, 92)]
+        [InlineData(0, 20)]
+        [InlineData(1, 8191)]
+        [InlineData(17, 8175)]
+        [InlineData(17, 20)]
+        public void AllReadWriteMethods(long offset, long size)
+        {
+            foreach (MemoryMappedFile mmf in CreateSampleMaps(8192))
```
ianhays: CreateSampleMaps(8192) fine, or should that number instead be "size"?

stephentoub: Good question, but I actually wanted this to be 8192, a value at least as large as any of the offset+size combinations.  I could have also made it offset+size, but part of what I wanted tested was creating accessors that are small slices of the overall map.

ianhays:
Just gave this a look over and found no egregious errors :) Looks good!

src/System.IO.MemoryMappedFiles/tests/MemoryMappedFile.CreateFromFile.Tests.cs
stephentoub: I went back and forth on that.  DefaultCapacity means something specific in this library, 0 being passed to mean "use the size of the underlying file or map", so I didn't want to use that, and I couldn't come up with something better.  Plus, my choices in many of these cases were largely random, any value would do as long as it was consistent within the method, hence my defining a constant in each of the methods where I needed consistency within that method.  I think I'll leave it this way for now... we can always choose to unify them later if desired.

stephentoub:
Thanks for the reviews, guys!

Issue #2335, opened 2015-07-13
Improve ConcurrentBag<T> GetEnumerator performance


MarkusSintonen:
Currently `ConcurrentBag<T>` `GetEnumerator()` causes list copying everytime enumerator is accessed eg. through `foreach`. This causes strain on the GC and list copying is also heavy operation on large bags.

There should be alternative methods to enumarate the bag contents or implement `GetEnumerator()` so that it does not cause list copying. I have presented here 3 possbile solutions to this problem.

**Solution 1)**
Add `ForEach` method to `ConcurrentBag<T>`:

``` CSharp
public void ForEach(Func<T, bool> action)
{
    // Short path if the bag is empty
    if (_headList == null)
        return;

    bool lockTaken = false;
    try
    {
        FreezeBag(ref lockTaken);

        ThreadLocalList currentList = _headList;
        while (currentList != null) // Go through thread lists
        {
            Node currentNode = currentList._head;
            while (currentNode != null) // Go through thead specific linked list
            {
                if (!action(currentNode._value))
                    return; // User wanted to break the operation
                currentNode = currentNode._next;
            }
            currentList = currentList._nextList;
        }
    }
    finally
    {
        UnfreezeBag(lockTaken);
    }
}
```

`ForEach` function takes user specified `Func<T,bool>` delegate. This delegate takes the bag item as an argument and returns true/false depending on whether the user wants to continue or stop the enumeration operation.

Here `ForEach` operation would lock the bag so that other threads would need to wait for the operation to finish. Therefore results of this operation would represent the bag as a moment-in-time snapshot.

**Solution 2)** 
Add `GetEnumerableSlim` function:

``` CSharp
public IEnumerable<T> GetEnumerableSlim()
{
    ThreadLocalList currentList = _headList;
    while (currentList != null)
    {
        Node currentNode = currentList._head;
        while (currentNode != null)
        {
            yield return currentNode._value;
            currentNode = currentNode._next;
        }
        currentList = currentList._nextList;
    }
}
```

This would require an additional change to the internal class `Node`:

``` CSharp
internal class Node
{
    public Node(T value)
    {
        _value = value;
    }
    public readonly T _value;
    public volatile Node _next; // This needs to be marked as volatile
    public Node _prev;
}
```

[...]

PR #2341, opened 2015-07-13
Avoid empty allocations in ConcurrentBag.GetEnumerator()

stephentoub:
Avoid allocating a ```List<T>``` and an enumerator when the bag is empty.

(Noticed this while looking at #2335.)


ellismg:
LGTM.

PR #2343, opened 2015-07-13
Eliminate extra IFileSystemObject allocation for FileSystemInfo/FileInfo/DirectoryInfo on Win32 and Unix

justinvp:
`FileSystemInfo`/`FileInfo`/`DirectoryInfo` carry an internal `IFileSystemObject`, which `FileSystemInfo` calls for the actual implementation. While this enables `FileSystemInfo`/`FileInfo`/`DirectoryInfo` to be used with WinRT on Windows, it comes with the cost of an extra unnecessary object allocation for plain Win32 and Unix. The extra allocation might not be a huge deal for one-off instances of `FileInfo` or `DirectoryInfo` but when enumerating lots of these from the file system, the extra allocations start adding up.

This change eliminates the extra object allocation on Win32 and Unix, while keeping the internal `FileSystem` abstraction intact. After this change, the extra object allocation only happens when the WinRT APIs need to be used on Windows.

It does this by having `FileSystemInfo` implement the internal `IFileSystemObject` interface itself, instead of implementing the interface on separate `Win32FileSystemObject` and `UnixFileSystemObject` types. `FileSytemInfo` becomes a partial class and its internal `FileSystemObject` property that provides the `IFileSystemObject` instance is moved to be defined by platform-specific partials. Win32FileSystemObject.cs becomes the FileSystemInfo.Windows.cs partial along with the FileSystemInfo.Win32.cs partial and its `FileSystemObject` always returns `this`, similarly, UnixFileSystemObject.cs becomes FileSystemInfo.Unix.cs and returns `this` for `FileSystemObject`. FileSystemInfo.WinRT.cs does the multi-plexing behavior: if the Win32 APIs can be used, `FileSystemObject` returns `this`, otherwise it returns a `WinRTFileSystemObject` that uses the WinRT APIs.

Aside: I'd like to use a similar approach to eliminate the need for the extra `FileStreamBase` object allocation inside `FileStream`, but I'll submit that as a separate PR, assuming the approach in this PR (or something like it) is acceptable.

Note: I could not compile/test the WinRT side (`EnableWinRT=true`) so it's possible I missed something there.


src/System.IO.FileSystem/src/System/IO/FileInfo.Win32.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace System.IO
+{
+    partial class FileInfo
+    {
+        [SecuritySafeCritical]
+        internal FileInfo(string fullPath, Interop.mincore.WIN32_FIND_DATA findData) : base(findData)
+        {
+            Debug.Assert(Path.IsPathRooted(fullPath), "fullPath must be fully qualified!");
+            _name = Path.GetFileName(fullPath);
```
justinvp: Question: any reason why we can't use `_name = findData.cFileName` here?

src/System.IO.FileSystem/src/System/IO/DirectoryInfo.Win32.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    partial class DirectoryInfo
+    {
+        [SecuritySafeCritical]
+        internal DirectoryInfo(string fullPath, Interop.mincore.WIN32_FIND_DATA findData) : base(findData)
+        {
+            Debug.Assert(PathHelpers.GetRootLength(fullPath) > 0, "fullPath must be fully qualified!");
+
+            // Fast path when we know a DirectoryInfo exists.
+            OriginalPath = Path.GetFileName(fullPath);
```
justinvp: Question: any reason why we can't use `OriginalPath = findData.cFileName` here?

justinvp:
cc: @ericstj, @stephentoub

whoisj:
@justinvp man you're just all over it today. Thank you from the coreclr community for all this great Windows love. :heart_eyes:

justinvp:
@whoisj :smile:

justinvp:
@ericstj would you mind taking a look? Is there anyone else who should review this?

justinvp:
Is there anyone else willing to take a look at this?

Adding some more folks who've been modifying/reviewing System.IO.FileSystem recently: @ianhays @JeremyKuhne @nguerrera @ericstj @stephentoub 

Note that this PR renames a couple files:
 - Win32FileSystemObject.cs -> FileSystemInfo.Windows.cs
 - UnixFileSystemObject.cs -> FileSystemInfo.Unix.cs

There have been some more recent changes submitted to these existing files after this PR was submitted, so I'll need to update this PR to make sure those changes move to the new file names. I don't want to spend a lot of time keeping this PR up-to-date if the approach isn't going to be acceptable in the end.

stephentoub:
@ericstj, you'd created the original structure here... would you be able to review the proposed direction?

src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs
```diff
@@ -23,7 +23,7 @@ public DirectoryInfo(String path)
         }
 
         [System.Security.SecuritySafeCritical]
-        internal DirectoryInfo(String fullPath, IFileSystemObject fileSystemObject) : base(fileSystemObject)
+        internal DirectoryInfo(String fullPath, bool ignoreThis)
```
JeremyKuhne: The ignored parameter is unfortunate. Would be nice if we could refactor this to be a little clearer. Perhaps an Initialize method with bool fullyQualified? Then you also call from the public and new findData constructors?

justinvp: > The ignored parameter is unfortunate.

I agree. I just went with the ignored parameter as this is what was done [previously](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/IO/FileInfo.cs#L132). I'll hold off on improving this until I hear back on the overall direction of the proposal.

JeremyKuhne: Other than this I'm fine with what you're doing here.

ericstj:
LGTM, sorry for the delay

stephentoub:
Thanks, Justin.  Looks good.  Can you go ahead and rebase and then I'll take another look?

justinvp:
The PR is up-to-date and ready for a final look before merging.

I made a few minor tweaks:

 - I separated the changes into two commits. The first commit just contains the file renames without changes to the content of the files, to help git track the moves. The second commit has the rest of the changes. I can squash if you don't want this.

 - Per Jeremy's feedback, I attempted to get rid of the constructors with the `ignoreThis` parameter, but every alternative I tried had downsides of its own. So instead, I made these constructors private, created internal static `InternalCreateFromFullPath` factory methods for use by callers outside the types, and added some private `Init` methods. I also added `InternalCreateFromFullPath` overloads that take an additional `WIN32_FIND_DATA` parameter on Windows, instead of using an internal constructor (it could have been a constructor), to align with the other factory methods that just take the `fullPath`. Basically, if you need to create one of these with a fullPath internally, use the factory methods. I'm open to better ideas here.

 - The `InternalCreateFromFullPath` factory methods that accept a `WIN32_FIND_DATA` parameter on Windows use the file name from `WIN32_FIND_DATA.cFileName` instead of `Path.GetFileName(fullPath)`, to avoid the string allocation.

Thanks for reviewing, Jeremy, Eric, and Steve!

justinvp:
I was thinking about it more last night and came up with an alternative approach to the factory methods, which keeps the internal ctors, but uses an optional `originalPath` parameter instead of `ignoreThis`.

Here it is in another branch: https://github.com/justinvp/corefx/commit/c4f5162ad7dd1f25872f6bfc4775d65c519c4917

I could go either way. Let me know if the alternative approach is preferred, or if you can think of a better way to do it, and I'll update the PR.

stephentoub:
Thanks, Justin.  Sorry for the delay.  I prefer your alternate approach in the other branch; I really dislike things that need an ignored parameter :smile:

stephentoub:
LGTM.  Thanks, Justin.

PR #1470, opened 2015-04-22
Fix #1339: Change encoding of .cs files to utf-8 without BOM

krwq:
Conversion script:
For each .cs file:
- Read all bytes
- Analyze BOM to get encoding
- If encoding not any of those then end: Encoding.UTF32;  Encoding.BigEndianUnicode; Encoding.Unicode
- Get string from bytes read from the file (encoding.GetString(bytes))
- Get utf-8 (no BOM) bytes from the string ((new Utf8Encoding(false)).GetBytes(str))
- Save to file

I'm not sure to reliably check if we didn't lose anything although I did run all tests (including outerloop) and made sure no test is testing anything related to encoding (looking at the file name)

Issue #1339, opened 2015-04-07
Make encoding of .cs files consistent


akoeplinger:
Here's a list of the .cs files that aren't ASCII/UTF8 in the repo:

```
$ find . -name *.cs -type f -exec file --mime {} \; | grep -v "charset=us-ascii" | grep -v "charset=utf-8"
./src/Common/tests/SystemXml/XmlCoreTest/TestData.g.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/XmlCoreTest/CustomWriter.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/XmlCoreTest/FilePathUtil.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/XmlCoreTest/AsyncUtil.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/interop.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/ctestcase.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/cmodinfo.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/ctestbase.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/ctestexception.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/cparser.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/ctestmodule.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/cltmconsole.cs: text/plain; charset=utf-16le
./src/Common/tests/SystemXml/ModuleCore/cvariation.cs: text/plain; charset=utf-16le
./src/System.Xml.ReaderWriter/tests/Readers/ReaderSettings/TCMaxSettings.cs: text/plain; charset=utf-16le
./src/System.Xml.ReaderWriter/tests/Readers/ReaderSettings/TCCreateOverloads.cs: text/plain; charset=utf-16le
./src/System.Xml.ReaderWriter/tests/Readers/ReaderSettings/TCLineInfo.cs: text/plain; charset=utf-16le
./src/System.Xml.ReaderWriter/tests/Readers/ReaderSettings/FilterSettings.cs: text/plain; charset=utf-16le
[...]

PR #2346, opened 2015-07-13
Convert UTF-16 encoded files into UTF8

ellismg:
UTF16 encoded files trip up both Git and GitHub since the huristic they
use to determine if a file is binary or not depends on the existence of
null bytes in a file, which is common in UTF-16 encoded source.

This converts all the UTF-16 encoded files into UTF-8 with an explicit
byte order marker.


ellismg:
The GitHub provided diff here is pretty terrible, since it believes the source files are binary.  I had success locally using kdiff3 + git difftool -d.  It shows the files and for each one tells me they are text equal but not binary equal.

stephentoub:
Thanks. The only other differences I saw were adding some newlines at the end of files. LGTM.

akoeplinger:
I assume this supersedes https://github.com/dotnet/corefx/pull/1470?

ellismg:
Maybe?  It was unclear if we landed on a BOM vs no BOM call, but this certainly gets us closer.

akoeplinger:
Yep, thanks, I actually didn't care about BOM vs no BOM when I opened https://github.com/dotnet/corefx/issues/1339, so this is fine by me :+1:

PR #2347, opened 2015-07-13
Adding new Unit Tests to the FileSystemWatcher code base

sokket:
Adding new UTs to the FSW class and fixing a bug in the Linux FSW where we would debug assert when cleaning up or when receiving an IGNORE notification for a directory that did not exist on disk

### DIFF
--- a/src/Common/src/Interop/Linux/libc/Interop.inotify.cs
+++ b/src/Common/src/Interop/Linux/libc/Interop.inotify.cs
@@ -14,8 +14,30 @@ internal static partial class Interop
         [DllImport(Libraries.Libc, SetLastError = true)]
         internal static extern int inotify_add_watch(int fd, string pathname, uint mask);
 
-        [DllImport(Libraries.Libc, SetLastError = true)]
-        internal static extern int inotify_rm_watch(int fd, int wd);
+        [DllImport(Libraries.Libc, SetLastError = true, EntryPoint = "inotify_rm_watch")]
+        private static extern int inotify_rm_watch_extern(int fd, int wd);
+
+        internal static int inotify_rm_watch(int fd, int wd)
+        {
+            int result = Interop.libc.inotify_rm_watch_extern(fd, wd);
+            if (result < 0)
+            {
+                int hr = System.Runtime.InteropServices.Marshal.GetLastWin32Error();
+                if (hr == Interop.Errors.EINVAL)
+                {
+                    // This specific case means that there was a deleted event in the queue that was not processed
+                    // so this call is expected to fail since the WatchDescriptor is no longer valid and was cleaned
+                    // up automatically by the OS.
+                    result = 0;
+                }
+                else
+                {
+                    System.Diagnostics.Debug.Fail("inotify_rm_watch failed with " + hr);
+                }
+            }
+
+            return result;
+        }
 
         [Flags]
         internal enum NotifyEvents
--- a/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
+++ b/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
@@ -10,6 +10,9 @@ using System.Threading.Tasks;
 
 namespace System.IO
 {
+    // Note: This class has an OS Limitation where the inotify API can miss events if a directory is created and immediately has
+    //       changes underneath. This is due to the inotify* APIs not being recursive and needing to call inotify_add_watch on
+    //       each subdirectory, causing a race between adding the watch and file system events happening.
     public partial class FileSystemWatcher
     {
         /// <summary>Starts a new watch operation if one is not currently running.</summary>
@@ -166,7 +169,7 @@ namespace System.IO
         }
 
         /// <summary>
-        /// State and processing associatd with an active watch operation.  This state is kept separate from FileSystemWatcher to avoid 
+        /// State and processing associated with an active watch operation.  This state is kept separate from FileSystemWatcher to avoid 
         /// race conditions when a user starts/stops/starts/stops/etc. in quick succession, resulting in the potential for multiple 
         /// active operations. It also helps with avoiding rooted cycles and enabling proper finalization.
         /// </summary>
@@ -470,7 +473,6 @@ namespace System.IO
                         }
 
                         uint mask = nextEvent.mask;
-                        bool addWatch = false;
                         string expandedName = null;
                         WatchedDirectory associatedDirectoryEntry = null;
 
@@ -531,6 +533,15 @@ namespace System.IO
                             previousEventCookie = 0;
                         }
 
+                        // If the event signaled that there's a new subdirectory and if we're monitoring subdirectories,
+                        // add a watch for it.
+                        const Interop.libc.NotifyEvents AddMaskFilters = Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_MOVED_TO;
+                        bool addWatch = ((mask & (uint)AddMaskFilters) != 0);
+                        if (addWatch && isDir && _includeSubdirectories)
+                        {
+                            AddDirectoryWatch(associatedDirectoryEntry, nextEvent.name);
+                        }
+
                         const Interop.libc.NotifyEvents switchMask =
                             Interop.libc.NotifyEvents.IN_Q_OVERFLOW | Interop.libc.NotifyEvents.IN_IGNORED |
                             Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_DELETE |
@@ -543,7 +554,6 @@ namespace System.IO
                                 break;
                             case Interop.libc.NotifyEvents.IN_CREATE:
                                 watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Created, expandedName);
-                                addWatch = true;
                                 break;
                             case Interop.libc.NotifyEvents.IN_IGNORED:
                                 // We're getting an IN_IGNORED because a directory watch was removed.
@@ -582,15 +592,7 @@ namespace System.IO
                                 previousEventName = null;
                                 previousEventParent = null;
                                 previousEventCookie = 0;
-                                addWatch = true; // for either rename or creation, we need to update our state
                                 break;
-                        }
-
-                        // If the event signaled that there's a new subdirectory and if we're monitoring subdirectories,
-                        // add a watch for it.
-                        if (addWatch && isDir && _includeSubdirectories)
-                        {
-                            AddDirectoryWatch(associatedDirectoryEntry, nextEvent.name);
                         }
 
                         // Drop our strong reference to the watcher now that we're potentially going to block again for another read
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
@@ -77,7 +77,7 @@ public partial class FileSystemWatcher_4000_Tests
     }
 
     [Fact, ActiveIssue(2279)]
-    public static void FileSystemWatcher_ChangeWatchedFolder()
+    public static void FileSystemWatcher_Changed_WatchedFolder()
     {
         using (var dir = Utility.CreateTestDirectory())
         using (var watcher = new FileSystemWatcher())
@@ -92,5 +92,30 @@ public partial class FileSystemWatcher_4000_Tests
 
             Utility.ExpectEvent(eventOccured, "changed");
         }
+    }
+
+    [Fact]
+    public static void FileSystemWatcher_Changed_NestedDirectories()
+    {
+        Utility.TestNestedDirectoriesHelper(WatcherChangeTypes.Changed, (AutoResetEvent are, TemporaryTestDirectory ttd) =>
+        {
+            Directory.SetLastAccessTime(ttd.Path, DateTime.Now);
+            Utility.ExpectEvent(are, "changed");
+        },
+        NotifyFilters.DirectoryName | NotifyFilters.LastAccess);
+    }
+
+    [Fact]
+    public static void FileSystemWatcher_Changed_FileInNestedDirectory()
+    {
+        Utility.TestNestedDirectoriesHelper(WatcherChangeTypes.Changed, (AutoResetEvent are, TemporaryTestDirectory ttd) =>
+        {
+            using (var nestedFile = new TemporaryTestFile(Path.Combine(ttd.Path, "nestedFile")))
+            {
+                Directory.SetLastAccessTime(nestedFile.Path, DateTime.Now);
+                Utility.ExpectEvent(are, "changed");
+            }
+        },
+        NotifyFilters.DirectoryName | NotifyFilters.LastAccess | NotifyFilters.FileName);
     }
 }
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
@@ -111,4 +111,16 @@ public partial class FileSystemWatcher_4000_Tests
             Utility.ExpectNoEvent(eventOccured, "created");
         }
     }
+
+    [Fact]
+    public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
+    {
+        Utility.TestNestedDirectoriesHelper(WatcherChangeTypes.Created, (AutoResetEvent are, TemporaryTestDirectory ttd) =>
+        {
+            using (var nestedFile = new TemporaryTestFile(Path.Combine(ttd.Path, "nestedFile")))
+            {
+                Utility.ExpectEvent(are, "nested file created", 1000 * 30);
+            }
+        });
+    }
 }
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
@@ -80,4 +80,78 @@ public partial class FileSystemWatcher_4000_Tests
             }
         }
     }
+
+    [Fact]
+    public static void FileSystemWatcher_Deleted_NestedDirectories()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        using (AutoResetEvent createOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created))
+        using (AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Deleted))
+        {
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.IncludeSubdirectories = true;
+            watcher.EnableRaisingEvents = true;
+
+            using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
+            {
+                // Wait for the created event
+                Utility.ExpectEvent(createOccured, "create", 1000 * 30);
+
+                using (var secondDir = new TemporaryTestDirectory(Path.Combine(firstDir.Path, "dir2")))
+                {
+                    // Wait for the created event
+                    Utility.ExpectEvent(createOccured, "create", 1000 * 30);
+
+                    using (var nestedDir = new TemporaryTestDirectory(Path.Combine(secondDir.Path, "nested")))
+                    {
+                        // Wait for the created event
+                        Utility.ExpectEvent(createOccured, "create", 1000 * 30);
+                    }
+
+                    Utility.ExpectEvent(eventOccured, "deleted");
+                }
+
+                Utility.ExpectEvent(eventOccured, "deleted");
+            }
+
+            Utility.ExpectEvent(eventOccured, "deleted");
+        }
+    }
+
+    [Fact]
+    public static void FileSystemWatcher_Deleted_FileDeletedInNestedDirectory()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        using (AutoResetEvent createOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created))
+        using (AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Deleted))
+        {
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.IncludeSubdirectories = true;
+            watcher.EnableRaisingEvents = true;
+
+            using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
+            {
+                // Wait for the created event
+                Utility.ExpectEvent(createOccured, "create", 1000 * 30);
+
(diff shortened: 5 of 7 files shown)


ellismg:
Super nit-picky, but do you mind spelling out Unit Tests in the commit message?

sokket:
@ellismg  Will do. This PR is still a work-in-progress...my Windows box was hosed so I was using this to test in Jenkins. I'll be updating it to fix the Unit Test break soon

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
```diff
@@ -14,8 +14,30 @@ internal static partial class libc
...
+                if (hr == Interop.Errors.EINVAL)
+                {
+                    // This specific case means that there was a deleted event in the queue that was not processed
+                    // so this call is expected to fail since the WatchDescriptor is not longer valid and was cleaned
+                    // up automatically by the OS.
+                    result = 0;
+                }
+                else
+                {
+                    System.Diagnostics.Debug.Assert(false, "inotify_rm_watch failed with " + hr);
```
Maxwe11: `Debug.Fail("inotify_rm_watch failed with " + hr.ToString())`

src/System.IO.FileSystem.Watcher/tests/project.json
```diff
@@ -12,6 +12,7 @@
     "System.Threading": "4.0.10-beta-*",
     "System.Threading.Overlapped": "4.0.0-beta-*",
     "System.Threading.Tasks": "4.0.10-beta-*",
+    "System.Threading.Thread": "4.0.0-beta-*",
```
Maxwe11: Thread used just for Sleep method? If so use `Task.Delay(...).Wait();`

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
stephentoub: There are currently two call sites to inotify_rm_watch, both of which have a:
```Debug.Assert(result >= 0);```
after them.  Rather than adding this new wrapper function (which may not be what some callers in the future want), could we instead just augment both of those asserts to be:
```Debug.Assert(result >= 0 || Marshal.GetLastWin32Error() == Interop.Errors.EINVAL);```
?

File: src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs [hunk 3]
```diff
@@ -93,4 +93,69 @@ public static void FileSystemWatcher_ChangeWatchedFolder()
...
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.IncludeSubdirectories = true;
+            watcher.EnableRaisingEvents = true;
+
+            using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
+            {
+                // We need to sleep to allow Linux FSW to get the event and get a WatchDescriptor on the directory
+                System.Threading.Thread.Sleep(1000);
```
stephentoub: I get nervous any time I see explicit sleeps being put in code like this, for two reasons:

1. Under certain conditions the pause time could end up being too low, resulting in sporadic failures.
2. The longer the forced pause time, the longer the test takes.  We have a goal (we're not there yet) of each suite of inner loop tests taking no more than 5 seconds, and just from the sleeps added in this PR, the running time for this suite will increase almost 20 seconds.

Instead, how about just blocking here until the event is fired for the creation of firstDir?  Create the child directory, wait for the event, create the next child, wait for the event, etc.  I'm ok with having long timeouts on those waits because they're not forcing a normal execution to slowdown, but rather just providing some wiggle room in case something's running a little slower here and there.  This is why in some of our tests you see things like ```Assert.True(someCondition.Wait(LongTimeout));``` (where LongTimeout is 30 seconds or something)... usually the condition being waited for will happen immediately, and it should always happen, but we give it 30 seconds wiggle room just in case the lab is running very slowly or something, enough time that if the assert ever does fail, something is definitely wrong.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -93,4 +93,69 @@ public static void FileSystemWatcher_ChangeWatchedFolder()
...
+                using (var nestedDir = new TemporaryTestDirectory(Path.Combine(firstDir.Path, "nested")))
+                {
+                    // We need to sleep to allow Linux FSW to get the event and get a WatchDescriptor on the directory
+                    System.Threading.Thread.Sleep(1000);
+                    using (var nestedFile = new TemporaryTestFile(Path.Combine(nestedDir.Path, "nestedFile")))
```
stephentoub: If we're able to follow my suggestion in the previous unit test, then we probably don't even need a second test here, since we're verifying the events are being received as we run the previous test, in which case this nestedFile could just be added to the innermost directory in that test.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -111,4 +111,73 @@ public static void FileSystemWatcher_Created_Negative()
...
+        using (var watcher = new FileSystemWatcher())
+        {
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created);
+
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.IncludeSubdirectories = true;
+            watcher.EnableRaisingEvents = true;
+
+            using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
```
stephentoub: There's a lot of repetition in the structure of these tests.  If you're able to condense them due to my previous suggestion, great.  If such repetition still remains, consider turning this into a helper function, e.g. pseudo-code:
```C#
private static void TestNestedDirectories(WatcherChangeTypes change, Action<AutoResetEvent, TemporaryTestDirectory> action)
{
    using (var dir = Utility.CreateTestDirectory())
    using (var watcher = new FileSystemWatcher())
    using (AutoResetEvent eventOccurred = Utility.WatchForEvents(watcher, change))
    {
        watcher.Path = ...;
        ... // watcher setup

        using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
        {
            WaitForEventForFirstDir(); // whatever this ends up being
            using (var secondDir = new TemporaryTestDirectory(Path.Combine(firstDir.Path, "dir2")))
            {
                action(eventOccurred, secondDir);
            }
        }
    }
}
```
or something like that.

stephentoub:
Thanks for starting to improve the suite, Jon.  I left some comments.

Additionally, I think you said this is just the beginning of adding more tests.  That's good. For future PRs, off the top of my head a few thoughts on additional things related to nested directories we should vet:

- Moving directories and files into and out of the watched hierarchy, both the most nested ones and ones in the middle of the hierarchy.
- IncludeSubdirectories == false and verifiying that activity in the nested directories doesn't trigger events
- Very deeply nested, e.g. 20+ levels deep.
- Starting watching a directory that already has nested directories/files, and starting/stopping/restarting with nested directories.
- The top-level or middle directory in a hierarchy is renamed and the events for inside its children contain the right updated path information.

I'm sure there are some other interesting cases, too.

src/System.IO.FileSystem.Watcher/tests/project.json
sokket: Good idea

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
sokket: I optimized for what the current code has, which is two calling sites checking the same thing. I did this because it's more likely that the current code will have churn and fix bugs and need the same checks; I figure that it's more unlikely that we will add future calls to inotify_rm_watch, let alone calls that should not take this check into consideration since it could always return this failure code. 

Thoughts?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: The point of these waits is to *not* proceed immediately on create due to a limitation of the Linux inotify API; once notified about a new create, the listening code (the FileSystemWatcher) must execute code to get a Watch Descriptor on that new directory (and traverse it if necessary) as well, otherwise we will not get notifications about changes underneath. The delay was necessary otherwise the tests fail on Linux since the FSW doesn't have time to create a WatchDescriptor and call inotify_add_watch before the new directory is created, thus missing changes.

I completely agree about having Sleep in test code, I'm just not sure what other options we have since that's exactly what we need to do in order to work around the limitations in the OS API...thoughts?

sokket:
Good idea, adding those to the list I have for new Unit Tests to add in future PRs

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Ah sorry, I misunderstood...I think that will work...testing now

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Good catch

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: This will work for some of them...the Deleted ones will need to stay as-is otherwise I would need to add an action to fire after every using block ended, which is basically special casing the Deleted events anyway.

sokket:
@piotrpMSFT FYI

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
```
sokket: Will remove the extra using statements that are no longer necessary

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
```diff
@@ -14,8 +14,30 @@ internal static partial class libc
...
+                if (hr == Interop.Errors.EINVAL)
+                {
+                    // This specific case means that there was a deleted event in the queue that was not processed
+                    // so this call is expected to fail since the WatchDescriptor is not longer valid and was cleaned
+                    // up automatically by the OS.
+                    result = 0;
+                }
+                else
+                {
+                    System.Diagnostics.Debug.Fail("inotify_rm_watch failed with " + hr);
```
Maxwe11: nit: `hr` boxed, use `hr.ToString()`

stephentoub: That'll only happen in a Debug build, though, and only if we're about to fail an assert.  I prefer clarity in such a case.

sokket:
@dotnet-bot please test this

sokket:
@dotnet-bot test this please

sokket:
@stephentoub looks like ETW tests are failing on Linux...???

stephentoub:
Yes, I commented to that affect on several of the other PRs. Not sure yet what's going on.

sokket:
Cool, I'll hold off with this until the branch is in a good state. I'll make a new branch for more UTs

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
```diff
@@ -14,8 +14,30 @@ internal static partial class libc
...
+        internal static int inotify_rm_watch(int fd, int wd)
+        {
+            int result = Interop.libc.inotify_rm_watch_extern(fd, wd);
+            if (result < 0)
+            {
+                int hr = System.Runtime.InteropServices.Marshal.GetLastWin32Error();
+                if (hr == Interop.Errors.EINVAL)
+                {
+                    // This specific case means that there was a deleted event in the queue that was not processed
+                    // so this call is expected to fail since the WatchDescriptor is not longer valid and was cleaned
```
stephentoub: nit: "not longer" => "no longer"

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -230,6 +233,10 @@ private sealed class RunningInstance
             /// Maximum length of a name returned from inotify event.
             /// </summary>
             private const int NAME_MAX = 255; // from limits.h
+            /// <summary>
+            /// Flags that signify that we need to add a new directory to the watch.
+            /// </summary>
+            private const Interop.libc.NotifyEvents AddMaskFilters = Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_MOVED_TO;
```
stephentoub: Does this need to be a const on the type like this, or can it be a local const as is done elsewhere in the method, e.g. just before it's used to set addWatch?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -531,6 +538,13 @@ private void ProcessEvents()
                             previousEventCookie = 0;
                         }
 
+                        // If the event signaled that there's a new subdirectory and if we're monitoring subdirectories,
+                        // add a watch for it.
+                        if (addWatch && isDir && _includeSubdirectories)
+                        {
+                            AddDirectoryWatch(associatedDirectoryEntry, nextEvent.name);
+                        }
+
```
stephentoub: Just curious, what was the bug that caused this to need to move up here?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -470,7 +477,7 @@ private void ProcessEvents()
                         }
 
                         uint mask = nextEvent.mask;
-                        bool addWatch = false;
+                        bool addWatch = ((mask & (uint)AddMaskFilters) != 0);
```
stephentoub: Can we move this down now to where it's used?  I'm not sure why I had it so far away.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -111,4 +111,46 @@ public static void FileSystemWatcher_Created_Negative()
...
+        TestNestedDirectoriesHelper(WatcherChangeTypes.Created, (AutoResetEvent are, TemporaryTestDirectory ttd) =>
+        {
+            using (var nestedFile = new TemporaryTestFile(Path.Combine(ttd.Path, "nestedFile")))
+            {
+                Utility.ExpectEvent(are, "nested file created", 1000 * 30);
+            }
+        });
+    }
+
+    protected static void TestNestedDirectoriesHelper(
```
stephentoub: Since this is used across multiple files, consider putting it into some more central/shared location.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -111,4 +111,46 @@ public static void FileSystemWatcher_Created_Negative()
...
+        WatcherChangeTypes                              change, 
+        Action<AutoResetEvent, TemporaryTestDirectory>  action, 
+        NotifyFilters                                   changeFilers = NotifyFilters.LastWrite | NotifyFilters.FileName | NotifyFilters.DirectoryName)
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        {
+            AutoResetEvent createdOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created);
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, change);
```
stephentoub: Ideally we'd dispose of these AREs when we're done with them.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -111,4 +111,46 @@ public static void FileSystemWatcher_Created_Negative()
...
+            watcher.IncludeSubdirectories = true;
+            watcher.EnableRaisingEvents = true;
+
+            using (var firstDir = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
+            {
+                Utility.ExpectEvent(createdOccured, "dir1 created", 1000 * 30);
+
+                using (var nestedDir = new TemporaryTestDirectory(Path.Combine(firstDir.Path, "nested")))
+                {
+                    Utility.ExpectEvent(createdOccured, "nested created", 1000 * 30);
```
stephentoub: Consider putting these timeout values into a named const.

src/System.IO.FileSystem.Watcher/tests/Utility/Utility.cs
```diff
@@ -91,12 +91,12 @@ public static AutoResetEvent WatchForEvents(FileSystemWatcher watcher, WatcherCh
...
         string message = String.Format("Didn't observe a {0} event within {1}ms", eventName, timeout);
-        Assert.True(eventOccured.WaitOne(Utility.Timeout), message);
+        Assert.True(eventOccured.WaitOne(timeout), message);
     }
 
     public static void ExpectNoEvent(WaitHandle eventOccured, string eventName, int timeout = Utility.Timeout)
     {
         string message = String.Format("Should not observe a {0} event", eventName);
-        Assert.False(eventOccured.WaitOne(Utility.Timeout), message);
+        Assert.False(eventOccured.WaitOne(timeout), message);
```
stephentoub: Wonder how long that's been like that :smile:

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
```diff
@@ -80,4 +80,78 @@ public static void FileSystemWatcher_Deleted_Negative()
...
+                    // Wait for the created event
+                    Utility.ExpectEvent(createOccured, "create", 1000 * 30);
+
+                    using (var nestedDir = new TemporaryTestFile(Path.Combine(secondDir.Path, "nestedFile"))) { }
+
+                    Utility.ExpectEvent(eventOccured, "deleted");
+                }
+            }
+
+            Utility.ExpectEvent(eventOccured, "deleted");
```
stephentoub: Are we explicitly choosing not to check for the deleted event for secondDir?  If yes, a comment explaining why would be helpful.  If no, should we?

stephentoub:
Thanks, Jon.  A few nits, but generally looks good.

stephentoub:
@dotnet-bot test this please

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: I'll make it local outside of the while loop

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: The bug was that the Creation event would be fired to clients before adding the watch for the folder; this meant that your previous comment about waiting for the creation event before continuing (to get rid of the Thread.Sleep) would be incorrect since the race condition would still occur since the event would be fired and the UT would proceed before the FSW had a chance to add the new watch

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Fixed. I'm keeping it as a bool rather than inlining the bit checking for readability purposes

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Moved to the utility class

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
stephentoub: I see.  Makes sense.  Thanks.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Ah forgot that ARE is IDisposable; nice catch

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Good idea

src/System.IO.FileSystem.Watcher/tests/Utility/Utility.cs
sokket: :) found this when my tests were timing out in < 1 second instead of 30

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
sokket: Oops, missed a ctrl+v there

sokket:
@stephentoub @ellismg New iteration with all comments addressed and tests passing

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -455,6 +458,8 @@ private void ProcessEvents()
                     WatchedDirectory previousEventParent = null;
                     uint previousEventCookie = 0;
 
+                    const Interop.libc.NotifyEvents AddMaskFilters = Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_MOVED_TO;
+
```
stephentoub: I'm still not clear why this isn't defined just before it's used below.  But not a big deal.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
```diff
@@ -80,4 +80,80 @@ public static void FileSystemWatcher_Deleted_Negative()
...
     }
+
+    [Fact]
+    public static void FileSystemWatcher_Deleted_NestedDirectories()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        {
+            AutoResetEvent createOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created);
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Deleted);
```
stephentoub: Were you going to dispose these?

stephentoub:
A few nits, otherwise LGTM.  Feel free to either merge as-is or address them; up to you.

sokket:
Merged, thanks all!

PR #2350, opened 2015-07-14
Short-circuiting versions of Enumerable methods.

### LINKED ISSUE #2349: Short-circuit when possible during Enumerable methods.
As mentioned at #2237 and #2238, there are methods of `Enumerable` that can benefit from short-circuiting.

Now, logically a great many could such as e.g. `Max(this IEnumerable<int>)` testing for `int.MaxValue`, but that would require an additional check that would penalise all sequences that didn't have that value.

However:

1. The logical behaviour of `Single` and `SingleOrDefault` is "look for a match, then check there isn't another" (just like `Where(pred).Single()`). Following that rather than keeping count removes work from non-error paths as well as shortening error paths.

2. `Min`/`MinOrDefault` already have to check for `NaN` explicitly, so they don't suffer from any further work being done.

3. Adding a check for `IList<T>` sources to `Last`/`LastOrDefault` adds only a single check to the entire operation. The search can then proceed in reverse order and break on the first match.

These changes all affect existing behaviour, in that if either the enumerable expected to be consumed, or they were used with a predicate that had a side-effect, that behaviour would now change.

On the one hand, any change to observed behaviour bar pure performance increase is bad.

On the other hand:

1. If you have side-effects in a `Func` you are going to have problems elsewhere, particularly elsewhere in Linq which tends to short-circuit where it can.
2. If you depend on an enumerable being consumed you are going to have problems elsewhere, , particularly elsewhere in Linq which tends to short-circuit where it can.
3. The idea of Linq being side-effect free is supported not just in the documentation, but as a justification for the lack of a `ForEach` outside of cases (parallel, asynchronous) where some other feature is added (eh, being parallel or asynchronous).
4. The current behaviour would be confusing to someone depending on side-effects anyway: Why does `Single(pred)` behave differently to `Where(pred).Single()`?
5. The behaviour of `Single` can exasperate bugs. (I had a case where certain values would throw a sequence into an incorrect repetition. Rather than `Single()` throwing at this bug, it hung).

Side-benefit: I noticed when working on this that not only do the current `Min(IEnumerable<double>)` etc. fail to take the short-circuit opportunity if it encounters a `NaN`, but they can actually perform much slower on a sequence with a `NaN` in the middle than one without, presumably due to branch mis-prediction. It's reasonable to believe that there may be other branch mis-prediction cases that such short-circuiting would help remove.

hackcraft:
Fixes #2349

`Last<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)`
`LastOrDefault<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)`
`Single<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)`
`SingleOrDefault<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)`
`Min(this IEnumerable<float>)`
`Min(this IEnumerable<float?>)`
`Min(this IEnumerable<double>)`
`Min(this IEnumerable<double?>)`

`Last` and `LastOrDefault` only short-circuit if the source is an `IList<T>`.

`Last` and `LastOrDefault` add a check for if the source is an `IList<T>` comparable
to that take by the form that don't take a predicate.

Apart from that no tests are added: While we could check for e.g. `int.MaxValue`
being seen by `Max(this IEnumerable<int>)`, and so on, that would add a test for
every element, and so penalise sequences that did not contain it.

`Min`/`MinOrDefault` on `double` and `float` already require an explicit check for `IsNaN` so
short-circuiting on it does not add such a penalty.

Included are tests for both the state before and after the changes.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1242,17 +1242,34 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
-            TSource result = default(TSource);
-            bool found = false;
-            foreach (TSource element in source)
+            IList<TSource> list = source as IList<TSource>;
+            if (list != null)
             {
-                if (predicate(element))
+                for (int i = list.Count - 1; i >= 0; --i)
                 {
-                    result = element;
-                    found = true;
+                    TSource result = list[i];
+                    if (predicate(result)) return result;
                 }
             }
-            if (found) return result;
+            else
+            {
+                using (IEnumerator<TSource> e = source.GetEnumerator())
+                {
+                    while (e.MoveNext())
+                    {
+                        TSource result = e.Current;
+                        if (predicate(result))
+                        {
+                            while (e.MoveNext())
+                            {
+                                TSource element = e.Current;
+                                if (predicate(element)) result = element;
+                            }
+                            return result;
+                        }
+                    }
+                }
+            }
             throw Error.NoMatch();
         }
 
@@ -1287,15 +1304,28 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
-            TSource result = default(TSource);
-            foreach (TSource element in source)
+            IList<TSource> list = source as IList<TSource>;
+            if (list != null)
             {
-                if (predicate(element))
+                for (int i = list.Count - 1; i >= 0; --i)
                 {
-                    result = element;
+                    TSource element = list[i];
+                    if (predicate(element)) return element;
                 }
+                return default(TSource);
             }
-            return result;
+            else
+            {
+                TSource result = default(TSource);
+                foreach (TSource element in source)
+                {
+                    if (predicate(element))
+                    {
+                        result = element;
+                    }
+                }
+                return result;
+            }
         }
 
         public static TSource Single<TSource>(this IEnumerable<TSource> source)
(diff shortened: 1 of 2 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1821,73 +1850,103 @@ public static long Min(this IEnumerable<long> source)
...
         {
             if (source == null) throw Error.ArgumentNull("source");
-            float value = 0;
-            bool hasValue = false;
-            foreach (float x in source)
+            using (IEnumerator<float> e = source.GetEnumerator())
             {
-                if (hasValue)
+                if (!e.MoveNext()) throw Error.NoElements();
+                float value = e.Current;
```
hackcraft: I don't check if the first element of an enumerable `IsNaN`, because that adds a test where the current code does not. Such a check could be easily added as well.

stephentoub:
cc: @VSadov

hackcraft:
Updated to squash in changes in the meantime.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1242,17 +1242,34 @@ public static TSource Last<TSource>(this IEnumerable<TSource> source, Func<TSour
...
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
-            TSource result = default(TSource);
-            bool found = false;
-            foreach (TSource element in source)
+            IList<TSource> list = source as IList<TSource>;
+            if (list != null)
             {
-                if (predicate(element))
+                for(var i = list.Count - 1; i >= 0; --i)
```
stephentoub: Nit: please use int instead of var, and add a space before the paren

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1287,15 +1304,28 @@ public static TSource LastOrDefault<TSource>(this IEnumerable<TSource> source, F
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
-            TSource result = default(TSource);
-            foreach (TSource element in source)
+            IList<TSource> list = source as IList<TSource>;
+            if (list != null)
             {
-                if (predicate(element))
+                for (var i = list.Count - 1; i >= 0; --i)
```
stephentoub: Nit: int instead of var

VSadov:
LGTM

stephentoub:
A couple minor things to be fixed, but otherwise LGTM2.  I believe we decided we're ok with the short-circuiting behavior.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Funny, I managed to subdue my normally var-heavy style, and then it crept in where I'd normally use int. 
On a phone now, but will get this soon.

hackcraft:
Well, I picked those nits. I've also removed the tests `[ActiveIssue("Demonstrates old behaviour. See #2349")]` as those were only really useful while we were discussing the backwards-compatibility impact.

stephentoub:
Thanks, Jon!

PR #2351, opened 2015-07-14
Enumerable.ToHashSet() Method

### LINKED ISSUE #2323: Add Enumerable.ToHashSet
We have `Enumerable.To*` methods for many common collections. `HashSet` is sorely missing. This feature should be cheap to implement and not be problematic in any way that I can see.

Evidence that this feature is sorely missing: https://www.google.com/webhp?complete=1&hl=en&gws_rd=cr,ssl&ei=#complete=1&hl=en&q=.net+tohashset People are writing this method over and over again.

hackcraft:
Fixes #2323

Overload for explicit selection of IEqualityComparer, overload
for defaulting to HashSet's default (which is EqualityComparer.Default).

### DIFF
--- a/src/System.Linq/tests/ToHashSetTests.cs
+++ b/src/System.Linq/tests/ToHashSetTests.cs
@@ -1,0 +1,64 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class ToHashSetTests
+    {
+        private class CustomComparer<T> : IEqualityComparer<T>
+        {
+            public bool Equals(T x, T y) { return EqualityComparer<T>.Default.Equals(x, y); }
+            public int GetHashCode(T obj) { return EqualityComparer<T>.Default.GetHashCode(obj); }
+            public override bool Equals(object obj)
+            {
+                return obj is CustomComparer<T>;
+            }
+            public override int GetHashCode()
+            {
+                return -465576218; // Equals all instances of same class, so randomly-produced constant.
+            }
+        }
+        [Fact]
+        public void NoExplicitComparer()
+        {
+            var hs = Enumerable.Range(0, 50).ToHashSet();
+            Assert.IsType<HashSet<int>>(hs);
+            Assert.Equal(50, hs.Count);
+            Assert.Equal(EqualityComparer<int>.Default, hs.Comparer);
+        }
+        [Fact]
+        public void ExplicitComparer()
+        {
+            var hs = Enumerable.Range(0, 50).ToHashSet(new CustomComparer<int>());
+            Assert.IsType<HashSet<int>>(hs);
+            Assert.Equal(50, hs.Count);
+            Assert.Equal(new CustomComparer<int>(), hs.Comparer);
+        }
+        [Fact]
+        public void TolerateNullElements()
+        {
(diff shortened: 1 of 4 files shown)


src/System.Linq/tests/ToHashSetTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            var hs = Enumerable.Range(0, 50).ToHashSet(new CustomComparer<int>());
+            Assert.IsType<HashSet<int>>(hs);
+            Assert.Equal(50, hs.Count);
+            Assert.Equal(new CustomComparer<int>(), hs.Comparer);
+        }
+        [Fact]
+        public void TolerateNullElements()
+        {
+            // Unlike the keys of a dictionary, HashSet tolerates null items.
+            new string[] { "abc", null, "def" }.ToHashSet();
```
JohanLarsson: No assert?

hackcraft: If it the behaviour it was testing for didn't happen, it would fail with an exception.

JohanLarsson: No Assert.DoesNotThrow in xunit?

hackcraft: Though now I think of it. Asserting that `.Contains(null)` would be a good idea.

hackcraft: > No Assert.DoesNotThrow in xunit?

Not apparently in the version of XUnit referenced. There's no `Timeout` property on `[Fact]` in the version of XUnit used, either, which I miss more.

JohanLarsson: While DoesNotThrow is not super nice and not often right it reads pretty good in tests like the above.

hackcraft: Yes, but not available when I go to build. The change I've made since is nicer still anyway, as it checks on the state of the produced set, along with verifying that it isn't an always-true state.

stephentoub: DoesNotThrow has been removed from xunit: https://github.com/xunit/xunit/issues/188

stephentoub:
Just a reminder that for new API surface area, we really need to follow the API design process outlined at https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md.  If the goal is to have somewhere to put example implementations, they can go in the associated issue, they can go in a gist, they can go in a branch in your own fork, etc.  Having them as PRs makes it more challenging for folks to know which PRs are headed towards merging and which aren't.  Thanks for understanding.

cc: @terrajobst

src/System.Linq/tests/ToHashSetTests.cs
hackcraft: @stephentoub Ah. Thanks for the info. I see that `Timeout` also got removed as per http://xunit.github.io/docs/test-migration.html on the grounds of accurate timing being infeasible, though I preferred it when used with inaccurate timing anyway (a case that should never take more than a second taking more than 10, etc.).

hackcraft:
@stephentoub 

> they can go in a branch in your own fork, etc

Perfect. Well it's obviously there. I shall close this PR.

stephentoub:
Thanks, @hackcraft. :smile:

PR #2360, opened 2015-07-15
Fix several test projects to use per-test-class random generator

### LINKED ISSUE #2355: System.Globalization.Tests.StringInfoGetNextTextElement2.TestRandomIndex failed
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/2253/
```
System.Globalization.Tests.StringInfoGetNextTextElement2.TestRandomIndex [FAIL]
        System.IndexOutOfRangeException : Index was outside the bounds of the array.
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Globalization\tests\StringInfo\StringInfoGetNextTextElement2.cs(19,0): at System.Globalization.Tests.StringInfoGetNextTextElement2.TestRandomIndex()
```

stephentoub:
Several of our test projects have a copy of a Generator.cs file that’s used to produce random data for use by the tests.  In multiple projects, this led to failures due to Generator.cs containing a static Random instance that’s used by tests running in parallel.  In a couple of projects, these issues were spotted and locks were added to Generator.cs, but not in others, resulting sporadic test failures.

This PR consolidates all of those copies into a single Common file, and converts the generator to be used as an instance rather than as a static class; each test class that needs to use it then gets its own instance, avoiding the race conditions and the need for locks and so on.

(As an aside, the design of this class is quite strange, in particular asround its use of seeds and attempts to change them, which are often ignored.  But I’ve not messed with that in this PR, as it would require significant changes to how the consuming tests use it.  Beyond making the type non-static, I’ve only made minimal changes to it, mostly around formatting, style, etc.)

The changes in this PR were done fairly programmatically:
- All usage of "TestLibrary.Generator." was replaced by "_generator"
- Classes that were then using "_generator" got a "private readonly RandomDataGenerator _generator;" field added to it
- The test project was updated to refer to the shared RandomDataGenerator.cs.
- The old Generator.cs was deleted.

Fixes #2355.


stephentoub:
cc: @ellismg, @venkat-raman251, @Priya91

Priya91:
:+1:

ellismg:
LGTM.  Thanks for making this improvement, Steve.

PR #2361, opened 2015-07-15
Restoring the return type of the Expression<TDelegate>.Update

VSadov:
The method is a public API and is used to return Expression< TDelegate >

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/DebugViewWriter.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/DebugViewWriter.cs
@@ -430,7 +430,7 @@ namespace System.Linq.Expressions
             return node;
         }
 
-        protected internal override Expression VisitLambda(LambdaExpression node)
+        protected internal override Expression VisitLambda<T>(Expression<T> node)
         {
             Out(
                 String.Format(CultureInfo.CurrentCulture,
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/ExpressionStringBuilder.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/ExpressionStringBuilder.cs
@@ -326,7 +326,7 @@ namespace System.Linq.Expressions
             return node;
         }
 
-        protected internal override Expression VisitLambda(LambdaExpression node)
+        protected internal override Expression VisitLambda<T>(Expression<T> node)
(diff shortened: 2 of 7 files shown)


src/System.Globalization/tests/project.lock.json
```diff
@@ -3,6 +3,17 @@
   "version": -9996,
   "targets": {
     "DNXCore,Version=v5.0": {
+      "System.AppContext/4.0.0-beta-23024": {
```
akoeplinger: The changes to this file aren't related to your change and looks like they were missed when the globalization tests were added, I think we should separate them out into a distinct commit/PR.

stephentoub: Agreed.

VSadov: The changes happen automatically when I build. I was not sure if they are needed. I can exclude them.

VSadov:
CC @weshaggard @Pilchie

weshaggard:
cc @nguerrera 
Is this going to be compatible? What did we change from the desktop to require this to be a newslot?

nguerrera:
Why is dropping override OK? Is anything calling through the base Update method? The base method is not part of the public surface seen through contracts. If nothing is using it, it should be removed. If something is, then it needs to be renamed (can't overload on return type) and overridden. In the latter case, I'd also mark them internal to draw attention to the fact that this is implementation-only.

Are there tests covering this? I realize it may be source compatible with them (stronger return type than before). Perhaps it's worth making the tests require `Expression<TDelegate>` in some way?

VSadov:
@nguerrera - you are right, the change is more involved than just changing the Update. I also need to remove Update from the LambdaExpressin (thus making newslot unnecessary).

I think the whole thing might have been added as a workaround for generic virtual Visit that takes Expression<T>, but it appears that it is still there, so I am not sure. 
Will need to check if this change is ok with .Net native before committing.

nguerrera:
There should also have been a test that failed when the override became newslot. Can you make sure that stress that code path?

stephentoub:
@dotnet-bot test this please

VSadov:
@dotnet-bot retest this please

VSadov:
retest this please

VSadov:
@AlekseyTs  @gafter @jaredpar @agocke - please review

Issue #2137, opened 2015-06-23
Investigate/implement native shims for Unix functions


stephentoub:
Our interop layer for accessing Unix functions (mostly libc) from managed code is currently fairly fragile, with hard-coded assumptions about the sizes of various types, with error values and various flag argument values needed to be copied into managed code per platform, etc.  Related to https://github.com/dotnet/corefx/issues/371, we should investigate (and subsequently implement as is appropriately) having a very thin native shim layer (one per managed library? one all-up?) that provides a P/Invokable and consistent surface area we can target, e.g. an ```open``` function in a System.IO.FileSystem.Native.so that in turn just calls the underlying libc's ```open``` function, doing whatever conversions are necessary between the standardized values/sizes/etc. expected by the managed code and what's actually used by the underlying platform.

omariom:
Is the approach [SharpDX devs took](http://sharpdx.org/documentation/tools/sharpgen) applicable in this case? They generate a very thin interop layer right from  DirectX header files.

nguerrera:
@omariom I think interop code generation, while definitely something to consider long term, is somewhat orthogonal to this. It could certainly help in theory, but as I dig into this it becomes apparent that there are lots of judgment calls to be made and that would either require very complex/configurable tooling or lots of opting out to manual mode. I think our best bet is to start with plain old hand-written C++ and C# code.

nguerrera:
Making note that this issue is about the code changes needed to remove the fragility and reuse more managed code across unixes. Not covered here are:

1. Building the native shim code in the corefx repo (Issue #2301)
2. How to package and ship the shims (Issue #2302)

nguerrera:
Still TODO after #2445:

- [ ] errno conversion
- [ ] declare fixed constants for stat file modes at shim layer

janhenke:
Note that #2445 does not compile on FreeBSD. Is this still work in progress or are we supposed to submit a PR fixing the issue?

nguerrera:
Please feel free to submit a PR and thanks for the help!

janhenke:
Just a small problem. PR#2536

bartonjs:
@nguerrera I know that there's still conversion to be done; but is there anything left for this particular issue? It seems like we have a plan forward, which sounds to me like the objective of this issue.

Just glancing through the open beta7 issues.

nguerrera:
@bartonjs This is basically the only thing tracking the rest of the conversion so I'd leave it open.

[...]

Issue #715, opened 2015-02-10
Unix interop uses non-posix open64


kangaroo:
The unix interop layer currently pinvokes open64.  open64 is a non-POSIX extension.  We should move to using open() with O_LARGEFILE.

This currently blocks using System.Console and System.IO on osx.

kangaroo:
Trivial fix without O_LARGEFILE:

```
diff --git a/src/Common/src/Interop/Unix/libc/Interop.open64.cs b/src/Common/src/Interop/Unix/libc/Interop.open64.cs
index d5810ce..e6ec079 100644
--- a/src/Common/src/Interop/Unix/libc/Interop.open64.cs
+++ b/src/Common/src/Interop/Unix/libc/Interop.open64.cs
@@ -11,7 +11,7 @@ internal static partial class Interop
     internal static partial class libc
     {
         [DllImport(Libraries.Libc, SetLastError = true)]
-        internal static extern int open64(string filename, OpenFlags flags, mode_t mode);
+        internal static extern int open(string filename, OpenFlags flags, mode_t mode);

         [Flags]
         internal enum OpenFlags
diff --git a/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs b/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
index e7cf81c..49f0bed 100644
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Win32.SafeHandles
             try { } finally
             {
                 int fd;
-                while (Interop.CheckIo(fd = Interop.libc.open64(path, flags, mode))) ;
+                while (Interop.CheckIo(fd = Interop.libc.open(path, flags, mode))) ;
                 Contract.Assert(fd >= 0);
                 handle.SetHandle((IntPtr)fd);
             }
diff --git a/src/System.Console/src/System/ConsolePal.Unix.cs b/src/System.Console/src/System/ConsolePal.Unix.cs
index a51ab83..b3de57c 100644
--- a/src/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/System.Console/src/System/ConsolePal.Unix.cs
@@ -448,7 +448,7 @@ namespace System
[...]

PR #2362, opened 2015-07-15
CoreFX 32 bit support

benpye:
Whilst working on the ARM version of CoreCLR I ran into a few issues because CoreFX assumes that the file functions will be the large file support versions in some cases, where they are not. I have added support to those that I have found and know that the LFS version exists. I wasn't sure on the best way to do this change, so I have currently added an EntryPoint parameter to the DllImport attribute, though I could also rename the functions but that touches a lot of code for no benefit in my eyes. I don't believe this should cause any breaks on other platforms but CI should show if it does.

Related to https://github.com/dotnet/coreclr/pull/1210

### DIFF
--- a/src/Common/src/Interop/Linux/libc/Interop.readdir.cs
+++ b/src/Common/src/Interop/Linux/libc/Interop.readdir.cs
@@ -11,7 +11,7 @@ internal static partial class Interop
 {
     internal static partial class libc
     {
-        [DllImport(Libraries.Libc, SetLastError = true)]
+        [DllImport(Libraries.Libc, SetLastError = true, EntryPoint = "readdir64")]
         internal static extern IntPtr readdir(SafeDirHandle dirp);
 
         internal static unsafe DType GetDirEntType(IntPtr dirEnt)
--- a/src/Common/src/Interop/Unix/libc/Interop.ftruncate.cs
+++ b/src/Common/src/Interop/Unix/libc/Interop.ftruncate.cs
@@ -10,7 +10,7 @@ internal static partial class Interop
 {
     internal static partial class libc
     {
-        [DllImport(Libraries.Libc, SetLastError = true)]
+        [DllImport(Libraries.Libc, SetLastError = true, EntryPoint = "ftruncate64")]
         internal static extern int ftruncate(int fd, off_t length);
     }
 }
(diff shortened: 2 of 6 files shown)


OtherCrashOverride:
Also related to dotnet/coreclr#930

benpye:
Is that relevant? These methods exist on all modern Unix like systems, on 64 bit they are equivalent to the standard methods, on 32 bit systems they provide support for larger files.

stephentoub:
Thank you for trying to fix this! Unfortunately at the moment we can't take this as it'll break OS X; we previously used some of these functions and moved away from them for issues like https://github.com/dotnet/corefx/issues/715 . There are currently many assumptions about 64-bit, and these should be addressed in the next few weeks as part of https://github.com/dotnet/corefx/issues/2137. cc: @nguerrera. Now that while our CI system currently builds for OS X, it doesn't currently run tests on it, only on Ubuntu and Windows.

benpye:
Ah okay, I didn't realise that OS X did things differently in this regard, I was the impression that at least these methods should be supported there. In that case I guess the native shims will be required as you say. Something else to consider, you already for some cases such as https://github.com/dotnet/corefx/blob/master/src/Common/src/Interop/BSD/libc/Interop.readdir.cs#L14 add a suffix for specific platforms, at least with the cases I have submitted, this could also work, with Mac OS X having no suffix and other platforms having "64".

stephentoub:
Thanks, @benpye.  Yes, there are things we could do to make this work, e.g. duplicating the INODE64 hack you mentioned, having duplicate Interop.*.cs files and including the right ones in the right builds, etc.  It's just that at present we're aiming to get rid of those hacks and consolidate everything back together, and this would push it further in the other direction.  Since we're expecting the rearchitecture of these P/Invokes to land in the not-too-distant future, I'd prefer to wait for that when committing changes to master... reasonable?

benpye:
Sure, makes sense to avoid hacks where possible, I appreciate that. Hopefully #2137 goes well then. Thanks for the explanation.

OtherCrashOverride:
> Is that relevant?

Yup. (as illustrated by OSX)

stephentoub:
Thanks, @benpye. I'm closing this for now.  If for some reason we have trouble completing #2137 in a timely fashion, we can revisit another solution here.  Thanks for your help.

PR #2364, opened 2015-07-15
Ensure all Common includes use a <Link/>

### LINKED ISSUE #144: All the linked files are in the root folder of the XML projects.
When you reference files outside the project directory, such as the common directory in the XML code case, these links by default show up in the root of the project in VS which is very annoying and clutters the solution unnecessarily. We should add the appropriate <Link> metadata to these Compile items so that VS will group them correctly in the solution explorer.

stephentoub:
Modifies various .csproj files to ensure that all Includes for files under Common use a <Link/> so that Visual Studio shows the files under a Common folder rather than flattened in the root.

Fixes #144

### DIFF
--- a/src/System.Xml.XmlSerializer/tests/System.Xml.XmlSerializer.Tests.csproj
+++ b/src/System.Xml.XmlSerializer/tests/System.Xml.XmlSerializer.Tests.csproj
@@ -17,7 +17,9 @@
     <Compile Include="XmlSerializerTests.cs" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="$(CommonTestPath)\System\Console.cs" />
+    <Compile Include="$(CommonTestPath)\System\Console.cs">
+      <Link>Common\System\Console.cs</Link>
+    </Compile>
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\System.Private.DataContractSerialization\src\System.Private.DataContractSerialization.csproj">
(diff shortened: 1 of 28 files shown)


stephentoub:
cc: @weshaggard, @akoeplinger

Maxwe11:
https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/tests/System.Net.Http.Tests.csproj#L26

stephentoub:
Thanks, @Maxwe11; I addressed those, too.

Maxwe11:
:+1:

src/System.Xml.XmlSerializer/tests/System.Xml.XmlSerializer.Tests.csproj
```diff
@@ -17,7 +17,9 @@
     <Compile Include="XmlSerializerTests.cs" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="$(CommonTestPath)\System\Console.cs" />
+    <Compile Include="$(CommonTestPath)\System\Console.cs">
+      <Link>Common\System\Console.cs</Link>
```
weshaggard: Orthogonal but any ideas whey we need this common Console.cs?

weshaggard:
LGTM thanks for cleaning this up @stephentoub

src/System.Xml.XmlSerializer/tests/System.Xml.XmlSerializer.Tests.csproj
stephentoub: The comment in that file says:
```C#
    // A dummy class for passing through System.Console calls until the System.Console contract is available.
    // To-do: Remove this workaround when possible.
```
and then implements WriteLine to call Debug.WriteLine.

stephentoub:
Thanks for reviewing, Wes.

PR #782, opened 2015-02-15
Scrub [ComVisible] attribution from CoreFX library sources

stephentoub:
Per #779, this PR scrubs [ComVisible] attribute from the CoreFX library sources.

@jkotas, one thing I'm unsure of: since [ComVisible(true)] is the default for desktop, and since some of these libraries (e.g. immutable collections, dataflow, etc.) can be used with the desktop, should we actually be applying [ComVisible(false)] to every assembly?  Some of the assemblies already had this, and with the current commit I've removed those, but if they're needed I can put them back and make sure the other libraries have such attribution, too.

PR #2371, opened 2015-07-15
Scrub additional [ComVisible] attributes from corefx

stephentoub:
In #782 we scrubbed all of the [ComVisible] attributes from corefx, but more have snuck in since then.  Doing another pass.

### DIFF
--- a/src/System.Xml.XmlSerializer/src/System/Collections/ArrayList.cs
+++ b/src/System.Xml.XmlSerializer/src/System/Collections/ArrayList.cs
@@ -9,7 +9,6 @@
 **
 =============================================================================*/
 
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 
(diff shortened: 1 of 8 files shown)


stephentoub:
cc: @jkotas, @weshaggard

src/System.Xml.XmlSerializer/src/System/Collections/ArrayList.cs
```diff
@@ -17,7 +16,6 @@
 namespace System.Collections
 {
     [DebuggerDisplay("Count = {Count}")]
-    [System.Runtime.InteropServices.ComVisible(true)]
     internal class ArrayList : List<object>
```
weshaggard: @khdang Any ideas why we have a duplicated ArrayList as opposed to using something like LowLeveList or `List<T>`.

weshaggard:
LGTM

PR #2390, opened 2015-07-16
FileSystem test update

### LINKED ISSUE #2359: System.IO.FileSystem.Tests.FileInfo_get_LastWriteTime.WriteTimeRefreshes failed
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/1161/console
```
System.IO.FileSystem.Tests.FileInfo_get_LastWriteTime.WriteTimeRefreshes [FAIL]
15:18:01       Assert.InRange() Failure
15:18:01       Range:  (635725090770029800 - 635725090780030360)
15:18:01       Actual: 635725090770000000
15:18:01       Stack Trace:
15:18:01 Skipped test. Make sure you have a parent directory (not root) with somes files before running this test
15:18:01             at System.IO.FileSystem.Tests.FileInfo_get_LastWriteTime.WriteTimeRefreshes()
```

ianhays:
- Updated Directory tests. Note that there are still some unresolved issues pending a solution to #1221
- Consolidated AccessTime, WriteTime, and CreationTime tests for FileInfo, File, DirectoryInfo, and Directory into one file per folder - GetSetTimes.cs
- Removed failing tests for the above properties
- Added some Unix tests
- Resolves #2359


src/System.IO.FileSystem/tests/Directory/GetDirectoryRoot_str.cs
```diff
@@ -1,191 +1,58 @@
...
-            ++iCountErrors;
-            Console.WriteLine(s_strTFName + " : Error Err_8888yyy!  strLoc==" + strLoc + ", exc_general==" + exc_general.ToString());
+            string root = Directory.GetDirectoryRoot(Path.Combine("a", "a", "a", "b") + Path.DirectorySeparatorChar);
+            Assert.Equal(Path.GetPathRoot(TestDirectory), root);
         }
-        ////  Finish Diagnostics
 
-        if (iCountErrors != 0)
+        [Fact]
+        public void UNCShares()
```
stephentoub: [PlatformSpecific(PlatformID.Windows)]?

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -1,462 +1,407 @@
...
         }
-    }
 
-    [Fact]
-    public static void CreateDirectory_ExistingDirectoryWithTrailingSlashAsPath_DoesNotThrow()
-    {
-        using (TemporaryDirectory directory = new TemporaryDirectory())
+        [Fact]
+        [PlatformSpecific(PlatformID.AnyUnix)]
+        [ActiveIssue(645)]
```
stephentoub: The way this is written, this isn't going to run on any platform, but it's fine on Unix, right?  I think what you meant was:
```C#
[Fact]
[ActiveIssue(645, PlatformID.Windows)]
public void LongPath_Allow() { ... }
```
which says "run this test on every platform except for Windows because there's an active issue preventing this from working on Windows".

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -1,462 +1,407 @@
...
+        public void UnixLongPath_Allowed()
         {
-            string path = IOServices.AddTrailingSlashIfNeeded(directory.Path);
-
-            DirectoryInfo result = Directory.CreateDirectory(path);
-
-            Assert.Equal(path, result.FullName);
-            Assert.True(Directory.Exists(result.FullName));
+            var path = new string('a', 280);
+            Create(Path.Combine(TestDirectory, path));
```
stephentoub: Except looking at this now, this isn't creating a long path, but rather a path that contains a component with a name longer than the max allowed component... this should fail on Unix now and in the future (that's not an active issue, that's just the way things are).

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -1,462 +1,407 @@
...
-#if !TEST_WINRT // Cannot set current directory to root from appcontainer with it's default ACL
-    /*
-    [Fact]
-    [ActiveIssue(1220)] // SetCurrentDirectory
-    public static void CreateDirectory_DotDotAsPath_WhenCurrentDirectoryIsRoot_DoesNotThrow()
-    {
-        string root = Path.GetPathRoot(Directory.GetCurrentDirectory());
+        [Fact]
+        [PlatformSpecific(PlatformID.AnyUnix)]
+        public void UnixLongPath_Disallowed()
```
stephentoub: Long component, not long path.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -1,462 +1,407 @@
...
+            });
         }
-    }
 
-    [Fact]
-    public static void CreateDirectory_ValidPathWithTrailingSlash_CanBeCreated()
-    {
-        using (TemporaryDirectory directory = new TemporaryDirectory())
+        [Fact]
+        [ActiveIssue(1221)]
```
stephentoub: Why is the test for a non-existent drive an active issue?  I understand the problem with mounting drives, but this seems to not need to mount and instead just wants to make sure things work in the opposite case.

src/System.IO.FileSystem/tests/Directory/Exists.cs
```diff
@@ -1,382 +1,307 @@
...
-    }
-
-    [Fact]
-    [PlatformSpecific(PlatformID.Windows)] // drive labels
-    public static void Exists_SubdirectoryOnNotReadyDriveAsPath_ReturnsFalse()
-    {
-        var drive = IOServices.GetNotReadyDrive();
-        if (drive == null)
+        [Fact]
+        [PlatformSpecific(PlatformID.Linux)]
```
stephentoub: | PlatformID.FreeBSD

src/System.IO.FileSystem/tests/Directory/GetSetTimes.cs
```diff
@@ -1,118 +1,131 @@
...
-                        if (setUtc)
-                        {
-                            expectedUtc = dt;
-                        }
-                        else
-                        {
-                            expected = dt;
-                        }
-                    }
+            Task.Delay(2000).Wait();
```
stephentoub: Rather than blocking, can we just subtract and add several seconds?  Even in outer loop it'd be great to avoid such long delays.

src/System.IO.FileSystem/tests/File/GetSetTimes.cs
```diff
@@ -1,118 +1,132 @@
...
+            long utcAfterTime = DateTime.UtcNow.Ticks;
+
+            Assert.All(TimeFunctions(), (tuple) =>
+            {
+                Assert.InRange(tuple.Item2(testFile.FullName).Ticks, tuple.Item3 == DateTimeKind.Local ? beforeTime : utcBeforeTime, tuple.Item3 == DateTimeKind.Local ? afterTime : utcAfterTime);
+            });
         }
 
-        File.Delete(fileName);
+        [OuterLoop]
```
stephentoub: Why outer loop?

stephentoub:
A few more comments, but otherwise LGTM.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
ianhays: Yeah you're right. This needs to be split into two distinct tests - one to test maxcomponent fails on both platforms, one to test longpath passes on unix but fails on Windows.

src/System.IO.FileSystem/tests/Directory/GetSetTimes.cs
ianhays: Yes, potentially. I previously was using DateTime.AddSeconds(-1) but was still running into InRange errors so I switched it up to Task.Delay(2000) and they stopped. I can switch back to DateTime.AddSeconds() but I would only do so with a larger buffer. 

It seemed theoretically sound that DateTime.AddSeconds(-1) would resolve the InRange second/ms resolution issue we discussed, but that didn't end up being true.

stephentoub: My point is just that there shouldn't be any difference from a time perspective between artificially changing the time via addition and doing so by blocking, but in the latter case we're using up precious resources.  I expect you'd have similar problems if you used blocking with a shorter time.

Also, with regards to adding/subtracting with times, it'd be best to do so in UTC time rather than in local time.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
ianhays: Yeah, I struggled deciding on this one. Since I have no real idea of what the final mounted drive testing framework is going to look like, I removed these tests to be safe. Realistically, they're fine in the form:
```
Assert.Throws<DirectoryNotFoundException>(() =>
{
       Create(IOServices.GetNonExistentDrive());
});
```
They will only fail if IOServices finds no NonExistentDrives, in which case the error would be ArgumentNullException instead of DirectoryNotFoundException. Although if that is the case, then there's something wrong with the testing system that is disallowing proper drive testing <i>anyways</i>, so it should be fine to reinclude them.

src/System.IO.FileSystem/tests/Directory/GetSetTimes.cs
ianhays: Ahh. Yeah, you're right. The Task.Delay was only useful in the tests I removed that checked if writing to a file updated the LastWriteTime, since the underlying system calls took a while to actually update and a delay was necessary for that to complete. In the creation case that isn't necessary since there's no previous time to replace; either it has the right time or no time, not an old time.

Okay. Out of curiosity, why? Shouldn't DateTime.Now.AddSeconds(-1) just be an offset version of DateTime.Now.AddSeconds(-1)?

src/System.IO.FileSystem/tests/File/GetSetTimes.cs
ianhays: My justification for OuterLoop was that it was using Task.Delay and was time-sensitive. Since I've replaced Task.Delay with AddSeconds(-n), that reasoning disappears. I will remove the OuterLoops.

ianhays:
My changes to the Xplat MaxDirectory and MaxComponent values in IOInputs caused several unexpected failures in classes/tests that relied on those incorrect values.

I may abandon this PR and roll the changes into a new PR addressing this issue, as the changes will likely be moderate to high in size.

PR #2397, opened 2015-07-17
Optimizations for ImmutableArray

James-Ko:
This is a total of four commits so far to ``ImmutableArray`1.cs``.

 - Highlight: Wrapper-returning methods that used to return a `new ImmutableArray<T>(self.array)` now return `self`. ImmutableArray is a struct, overrides Equals, and is immutable, which makes 2 different ImmutableArrays pointing to the same array very very hard to distinguish from each other.

 - `InsertRange` now enumerates only once over the collection when inserting it into the new array. This also makes methods such as `InsertRange(int, ImmutableArray)` faster because that just inserts the inner array into this method, and when `ToArray()` detects that the enumerable is an array is just does a copy operation and never calls `foreach` at all.

(The downside to this is that we shouldn't even have to call `ToArray()` if the enumerable is already an `ICollection<T>`, so there should be some kind of overload of `InsertRange` for that.)

 - `ImmutableArray.Remove` and `ImmutableArray.Replace` now return `this` immediately if `this` is empty.

 - `Add` checks for empty, then calls `Insert`, which also checks for empty. Removed the check in `Add`.

 - Removed the lazy initialization of `List<int>` in `RemoveAll`: the default `List` ctor is very cheap (sets underlying array to an empty array), so it is definitely not worth the null check every iteration of the loop.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -597,11 +597,6 @@ namespace System.Collections.Immutable
         public ImmutableArray<T> Add(T item)
         {
             var self = this;
-            if (self.Length == 0)
-            {
-                return ImmutableArray.Create(item);
-            }
-
             return self.Insert(self.Length, item);
         }
 
@@ -686,6 +681,10 @@ namespace System.Collections.Immutable
         public ImmutableArray<T> Replace(T oldValue, T newValue, IEqualityComparer<T> equalityComparer)
         {
             var self = this;
+            
+            if (self.IsEmpty)
+                return self;
+            
             int index = self.IndexOf(oldValue, equalityComparer);
             if (index < 0)
             {
@@ -721,6 +720,10 @@ namespace System.Collections.Immutable
         {
             var self = this;
             self.ThrowNullRefIfNotInitialized();
+            
+            if (self.IsEmpty)
+                return self;
+            
             int index = self.IndexOf(item, equalityComparer);
             return index < 0
                 ? self
@@ -792,6 +795,9 @@ namespace System.Collections.Immutable
             self.ThrowNullRefIfNotInitialized();
             Requires.NotNull(items, "items");
             Requires.NotNull(equalityComparer, "equalityComparer");
+            
+            if (self.IsEmpty)
+                return self;
 
             var indexesToRemove = new SortedSet<int>();
             foreach (var item in items)
@@ -834,11 +840,11 @@ namespace System.Collections.Immutable
         public ImmutableArray<T> RemoveRange(ImmutableArray<T> items, IEqualityComparer<T> equalityComparer)
         {
             var self = this;
+            self.ThrowNullRefIfNotInitialized();
             Requires.NotNull(items.array, "items");
-
+            
             if (items.IsEmpty)
             {
-                self.ThrowNullRefIfNotInitialized();
                 return self;
             }
             else if (items.Length == 1)
@@ -1096,7 +1102,7 @@ namespace System.Collections.Immutable
         public IEnumerable<TResult> OfType<TResult>()
         {
             var self = this;
-            if (self.array == null || self.array.Length == 0)
+            if (self.IsDefaultOrEmpty)
             {
                 return Enumerable.Empty<TResult>();
             }
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -545,21 +545,34 @@ namespace System.Collections.Immutable
             Requires.NotNull(items, "items");
 
             if (self.Length == 0)
-            {
                 return ImmutableArray.CreateRange(items);
-            }
             
-            T[] other = items.ToArray();
-            if (other.Length == 0)
+            if (items is ICollection<T>)
             {
-                return self;
+                ICollection<T> other = (ICollection<T>)items;
+                
+                if (other.Count == 0)
+                    return self;
+                
+                T[] dest = new T[self.Length + other.Count];
+                Array.Copy(self.array, 0, dest, 0, index);
+                other.CopyTo(dest, index);
+                Array.Copy(self.array, index, dest, index + other.Count, self.Length - index);
+                return new ImmutableArray<T>(dest);
             }
-
-            T[] tmp = new T[self.Length + other.Length];
-            Array.Copy(self.array, 0, tmp, 0, index);
-            Array.Copy(other, 0, tmp, index, other.Length);
-            Array.Copy(self.array, index, tmp, index + other.Length, self.Length - index);
-            return new ImmutableArray<T>(tmp);
+            else
+            {
+                T[] other = items.ToArray();
+                
+                if (other.Length == 0)
+                    return self;
+    
+                T[] tmp = new T[self.Length + other.Length];
+                Array.Copy(self.array, 0, tmp, 0, index);
+                Array.Copy(other, 0, tmp, index, other.Length);
+                Array.Copy(self.array, index, tmp, index + other.Length, self.Length - index);
+                return new ImmutableArray<T>(tmp);
+            }
         }
 
         /// <summary>
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -890,7 +890,7 @@ namespace System.Collections.Immutable
 
             List<int> removeIndexes = null;
             
-            int i;
+            int i = 0;
             for (; i < self.array.Length; i++)
             {
                 if (match(self.array[i]))
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -888,16 +888,25 @@ namespace System.Collections.Immutable
             if (self.IsEmpty)
                 return self;
 
-            List<int> removeIndexes = new List<int>();
-            for (int i = 0; i < self.array.Length; i++)
+            List<int> removeIndexes = null;
+            
+            int i;
+            for (; i < self.array.Length; i++)
             {
                 if (match(self.array[i]))
                 {
+                    removeIndexes = new List<int>();
                     removeIndexes.Add(i);
+                    break;
                 }
             }
+            for (; i < self.array.Length; i++)
+            {
+                if (match(self.array[i]))
+                    removeIndexes.Add(i);
+            }
 
-            return removeIndexes.Count != 0 ?
+            return removeIndexes != null ?
                 self.RemoveAtRange(removeIndexes) :
                 self;
         }
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -555,7 +555,7 @@ namespace System.Collections.Immutable
                     return self;
                 
                 T[] dest = new T[self.Length + other.Count];
-                other.CopyTo(dest, index);
+                other.CopyTo(dest, index); // keep this before so any changes it makes to unintended parts of the array are overwritten by Array.Copy
                 Array.Copy(self.array, 0, dest, 0, index);
                 Array.Copy(self.array, index, dest, index + other.Count, self.Length - index);
                 return new ImmutableArray<T>(dest);
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -587,7 +587,7 @@ namespace System.Collections.Immutable
             }
             else if (items.IsEmpty)
             {
-                return new ImmutableArray<T>(self.array);
+                return self;
             }
 
             return self.InsertRange(index, items.array);
@@ -728,7 +728,7 @@ namespace System.Collections.Immutable
             self.ThrowNullRefIfNotInitialized();
             int index = self.IndexOf(item, equalityComparer);
             return index < 0
-                ? new ImmutableArray<T>(self.array)
+                ? self
                 : self.RemoveAt(index);
         }
 
@@ -875,9 +875,7 @@ namespace System.Collections.Immutable
             Requires.NotNull(match, "match");
 
             if (self.IsEmpty)
-            {
-                return new ImmutableArray<T>(self.array);
-            }
+                return self;
 
             List<int> removeIndexes = null;
             for (int i = 0; i < self.array.Length; i++)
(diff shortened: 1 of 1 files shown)


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -548,22 +548,17 @@ public ImmutableArray<T> InsertRange(int index, IEnumerable<T> items)
             {
                 return ImmutableArray.CreateRange(items);
             }
-
-            int count = ImmutableExtensions.GetCount(ref items);
-            if (count == 0)
+            
+            T[] other = items.ToArray();
```
sharwell: :exclamation: I'm not sure this is better. It only iterates once, but it allocates a new intermediate array. When `items` implements `ICollection<T>`, I don't see how this is necessarily better.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -691,6 +681,10 @@ public ImmutableArray<T> Replace(T oldValue, T newValue)
         public ImmutableArray<T> Replace(T oldValue, T newValue, IEqualityComparer<T> equalityComparer)
         {
             var self = this;
+            
+            if (self.IsEmpty)
```
sharwell: :exclamation: This changes the semantics of the method. Previously, if `self` is empty the method always threw an exception.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -875,25 +876,18 @@ public ImmutableArray<T> RemoveAll(Predicate<T> match)
             Requires.NotNull(match, "match");
 
             if (self.IsEmpty)
-            {
-                return new ImmutableArray<T>(self.array);
-            }
+                return self;
 
-            List<int> removeIndexes = null;
+            List<int> removeIndexes = new List<int>();
```
sharwell: :exclamation: I'm not convinced this will be faster than the previous lazy initialization.

sharwell:
I have a few concerns which are noted in comments in the diff. However, the wrapper elimination should be a small but easy win; I recommend separating those changes into their own pull request as I expect they would be readily merged.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: True, I'll roll this change back for `Replace`.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: @sharwell I have fixed this in a newer commit, at the cost of expanding the code by about 10 lines in a check for `ICollection<T>`. There is no longer a buffer between `items` and the dest array.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: @sharwell Why not? The [`List<T>` ctor](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L49-L51) is extremely cheap so there's no benefit to lazy initialization. We also avoid a null check for every match, and make the code more readable.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -545,26 +545,34 @@ public ImmutableArray<T> InsertRange(int index, IEnumerable<T> items)
...
             {
-                return self;
+                ICollection<T> other = (ICollection<T>)items;
+                
+                if (other.Count == 0)
+                    return self;
+                
+                T[] dest = new T[self.Length + other.Count];
+                Array.Copy(self.array, 0, dest, 0, index);
+                other.CopyTo(dest, index);
```
sharwell: :exclamation: There is potential ability for an implementation of `CopyTo` to manipulate elements of the array which it shouldn't. You can solve this by moving the call to `other.CopyTo` above the two calls to `Array.Copy`, ensuring that even if the call to `CopyTo` results in out-of-bounds array writes, the elements before and after the inserted range are fully protected against modification.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
sharwell: In the event no item is removed, this code allocates a `List<int>` instance which is never used. I believe this change deserves a much closer look during the code review, which is why I pointed it out.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: @sharwell Fixed. That's a pretty sharp insight.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: I have rewritten the method to both avoid the null check and lazily instantiate the `List<int>` in the new commit.

mikedn:
> Highlight: Wrapper-returning methods that used to return a  new ImmutableArray<T>(self.array)  now return  self . ImmutableArray is a struct, overrides Equals, and is immutable, which makes 2 different ImmutableArrays pointing to the same array very very hard to distinguish from each other.

It's not very hard, it is impossible. The value the caller gets is always a copy of whatever value you pass to return, be it `self` or `new ImmutableArray<T>(self.array)`. In terms of generated code this particular change has no effect, fortunately the JIT manages to correctly handle structs that contain only one reference field. Not to say that the change is wrong and should not be done, it's just that it doesn't do what you think it does.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -545,26 +545,34 @@ public ImmutableArray<T> InsertRange(int index, IEnumerable<T> items)
...
+                T[] other = items.ToArray();
+                
+                if (other.Length == 0)
+                    return self;
+    
+                T[] tmp = new T[self.Length + other.Length];
+                Array.Copy(self.array, 0, tmp, 0, index);
+                Array.Copy(other, 0, tmp, index, other.Length);
+                Array.Copy(self.array, index, tmp, index + other.Length, self.Length - index);
+                return new ImmutableArray<T>(tmp);
```
stephentoub: This change impacts both behavior and performance, and not necessarily in a positive way.  For example, the previous implementation was using GetCount, which is able to get a count from an ```ICollection```, an ```ICollection<T>```, or an ```IReadOnlyCollection<T>```, whereas the new version is only specializing ```ICollection<T>```.  There's too much changing as part of this for me to feel confident that it's a net benefit.  Do you have performance tests that demonstrate it is across a range of input collections and collection types?  If so, can you share?  I suggest this be reverted for now.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -602,11 +610,6 @@ public ImmutableArray<T> InsertRange(int index, ImmutableArray<T> items)
         public ImmutableArray<T> Add(T item)
         {
             var self = this;
-            if (self.Length == 0)
-            {
-                return ImmutableArray.Create(item);
-            }
-
```
stephentoub: It's true that Insert does a length check, but that doesn't necessarily mean this should be eliminated.  For example, if Add can typically be inlined, Insert can't, and the case of adding to an empty list is common (which I believe it is), then this could actually be hurting performance in a common case because it forces the calling of a non-inlined method.  We shouldn't be making/taking changes like this without a solid set of perf measurements across a range of scenarios to demonstrate that it's actually a win.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -726,9 +730,13 @@ public ImmutableArray<T> Remove(T item, IEqualityComparer<T> equalityComparer)
         {
             var self = this;
             self.ThrowNullRefIfNotInitialized();
+            
+            if (self.IsEmpty)
+                return self;
+
```
stephentoub: Is it common to Remove from an empty list?  Upfront checks like this that optimize for the empty case are great if we expect it to be empty, but if it's more likely for it not to be empty, then this actualy penalizes the common case.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -797,6 +805,9 @@ public ImmutableArray<T> RemoveRange(IEnumerable<T> items, IEqualityComparer<T>
             self.ThrowNullRefIfNotInitialized();
             Requires.NotNull(items, "items");
             Requires.NotNull(equalityComparer, "equalityComparer");
+            
+            if (self.IsEmpty)
+                return self;
```
stephentoub: Ditto.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -839,11 +850,11 @@ public ImmutableArray<T> RemoveRange(ImmutableArray<T> items)
         public ImmutableArray<T> RemoveRange(ImmutableArray<T> items, IEqualityComparer<T> equalityComparer)
         {
             var self = this;
+            self.ThrowNullRefIfNotInitialized();
```
stephentoub: This is changing behavior.  Why did you make this change?

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -875,23 +886,25 @@ public ImmutableArray<T> RemoveAll(Predicate<T> match)
             Requires.NotNull(match, "match");
 
             if (self.IsEmpty)
-            {
-                return new ImmutableArray<T>(self.array);
-            }
+                return self;
```
stephentoub: Even if we change this to ```return self;``` instead of ```return new ImmutableArray<T>(self.array);```, we should still keep the braces that keep the formatting consistent as much as possible with the rest of the project.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -875,23 +886,25 @@ public ImmutableArray<T> RemoveAll(Predicate<T> match)
...
+                    removeIndexes = new List<int>();
                     removeIndexes.Add(i);
+                    break;
                 }
             }
+            for (; i < self.array.Length; i++)
+            {
+                if (match(self.array[i]))
+                    removeIndexes.Add(i);
+            }
```
stephentoub: Could you share your performance test that demonstrates this change helps?  I understand it in theory.  In practice it makes the code more complicated and duplicates code, so there should be a measurable win to justify it.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
sharwell: :memo: The only case where this changes behavior is when *both* `this` and `items` are default instances of `ImmutableArray<T>` - the change is due to the reversal in the order of checking. Previously this threw an `ArgumentNullException` for this case, but now it throws a `NullReferenceException`.

:thought_balloon: The new behavior appears to be more consistent with other methods in this type.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
stephentoub: This change is also the cause of the test failure, I believe, as it also changes behavior, and there's a test that catches that.

stephentoub:
cc: @AArnott

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
James-Ko: @stephentoub This was the only case in the file where the init check was not at the top of the method, so I thought it may have been a mistake on the part of the original writer.

stephentoub: @James-Ko, it may very well have been a mistake, and it may very well make things more consistent with the rest of the system... I can appreciate both of those things.  It's just that at this point the library has already shipped, so we need to be cognizant of any changes in behavior and either avoid them or at least be very thoughtful about when and why they're made.  Such changes should not be made as part of a PR focused on performance improvements.  Thanks for understanding.

AArnott:
Thanks for the CC, @stephentoub. I am very interested in reviewing before this is merged. But I can't review right now. Perhaps by Monday?

stephentoub:
@AArnott, have you had a chance to look at this? I left a bunch of comments highlighting concerns, and I'm wondering if there are pieces of this PR you think should be kept.

AArnott:
Not yet, but I should be able to find some time today to.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
AArnott: +1 to both of @stephentoub's arguments: yes possibly this is a good change, and yes it should be its own PR.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
AArnott: There are a few of these changes in this file (returning an existing instance of the struct rather than creating a new one). Is this a micro-optimization where we avoid creating/copying a struct one additional time? If so, I don't object. This makes it slightly easier to read and (I guess) slightly faster. But that probably needs a new, dedicated pull request to get away from several of the more questionable changes in this PR.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
AArnott: +1

AArnott: +1
I'd also ask that comments explaining the code be added since it's not as obvious why it's written the way you've done it.

AArnott:
@James-Ko Thanks for your time in contributing this. I imagine you took quite a bit of time to look for these micro-optimizations. I agree with many of @stephentoub's comments however. I recommend we close this PR and if you'd like to send a few more focused PRs (each with one kind of optimization applied throughout the class, and without unnecessary formatting changes) for those ideas that sounded good but were lost amidst the noise of the other changes, we may be able to entertain those.
Where we asked for perf numbers to justify specific changes, I think we should get those before taking those kinds of changes.

James-Ko:
@AArnott Sure; I've been busy so I didn't have the time to separate the changes like @sharwell suggested, but I'll close this and do that right now.

PR #2401, opened 2015-07-17
Optimise First, Last, FirstOrDefault & LastOrDefault for OrderedEnumerable

### LINKED ISSUE #2400: Take advantage of algorihmic differences between first-from-ordered and ordered.
The fact of something being ordered can have implications for subsequent operations.

The ordering Linq methods (`OrderBy`, `ThenBy`, etc.) operate first in O(1) time and space to make a promise of a subsequent O(n log n) sort in O(n) space. In many case though the following operation does not need all of the promised operation.

`.OrderBy(λ).First()`, `.OrderBy(λ).First(λ)`, and so on request that a sequence be sorted (O(n log n) time/O(n) space) and then the first element taken (O(1) time and space) which is in total an O(n log n) time/O(n) space operation. However, it is reducible to the O(n) time/O(1) space operation of identifying just the item in question which can be done in O(n) time and O(1) space.

As `First`, `Last` and their variants are particularly common with ordered sequences (in some cases not having any meaning without an order), I think this would be a worth-while optimisation.

There are other optimisations available, such as all aggregate operations (`Sum` etc.) could be moved from O(n log n) time to O(n) time simply by extracting the source and ignoring the ordering, this may not be as common a case as to make it worthwhile. One could argue that the user should have not put the ordering in there in the first place, and while the cost of checking for an ordered source is slightly, it's not worth paying it on all aggregate operations to benefit a flawed case.

`Reverse` is somewhere in the middle. It's reasonable to expect someone might construct an ordered sequence and then `Reverse` in some cases. It's got subtleties (one cannot simply reverse all `OrderBy` to `OrderByDescending` and vice-versa because `OrderBy` is documented as performing stable ordering in linq-2-objects and therefore reversed must have reverse source ordering on equivalent items) but wouldn't be hard, but it's less clear whether the pay off would be worth it.

In the more useful cases of `First` etc. there is a fly in the ointment in that this would affect the number of times both key selectors for ordering and predicates for `First` etc. were called. As per #2349 this means side effects on `Func`s would be affected, though perhaps in a way that was less predictable to the user than with that issue anyway.

As such, changing the behaviour of `First` in this case would require either a decision in favour of the change that issue would allow for, or a decision that this case was less harmed by such changing side-effects.

In summary. I propose that `OrderBy(…).First(…)` and its cousins be optimised, and the trade-offs of optimising `OrderBy(…).Reverse()` be considered.

JonHanna:
Fixes #2400

If you only want the first or last element of an ordered enumerable, you do
not need to O‎(n log n) (worse case, O(n²) but rare) operation of sorting
the entire set in O(n) space, but can instead to an O(n) retrieval of just
that element in O(1) space.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 
 namespace System.Linq
@@ -832,18 +833,18 @@ namespace System.Linq
         public static IEnumerable<TSource> Take<TSource>(this IEnumerable<TSource> source, int count)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            if (count <= 0) return new EmptyPartition<TSource>();
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.Take(count);
             return TakeIterator<TSource>(source, count);
         }
 
         private static IEnumerable<TSource> TakeIterator<TSource>(IEnumerable<TSource> source, int count)
         {
-            if (count > 0)
+            foreach (TSource element in source)
             {
-                foreach (TSource element in source)
-                {
-                    yield return element;
-                    if (--count == 0) break;
-                }
+                yield return element;
+                if (--count == 0) break;
             }
         }
 
@@ -884,18 +885,15 @@ namespace System.Linq
         public static IEnumerable<TSource> Skip<TSource>(this IEnumerable<TSource> source, int count)
         {
             if (source == null) throw Error.ArgumentNull("source");
-
+            if (count < 0) count = 0;
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.Skip(count);
             IList<TSource> sourceList = source as IList<TSource>;
             return sourceList != null ? SkipList(sourceList, count) : SkipIterator<TSource>(source, count);
         }
 
         private static IEnumerable<TSource> SkipList<TSource>(IList<TSource> source, int count)
         {
-            if (count < 0)
-            {
-                count = 0;
-            }
-
             while (count < source.Count)
             {
                 yield return source[count++];
@@ -1445,6 +1443,8 @@ namespace System.Linq
         public static TSource First<TSource>(this IEnumerable<TSource> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.First();
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1464,6 +1464,8 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
+            OrderedEnumerable<TSource> ordered = source as OrderedEnumerable<TSource>;
+            if (ordered != null) return ordered.First(predicate);
             foreach (TSource element in source)
             {
                 if (predicate(element)) return element;
@@ -1474,6 +1476,8 @@ namespace System.Linq
         public static TSource FirstOrDefault<TSource>(this IEnumerable<TSource> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.FirstOrDefault();
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1493,6 +1497,8 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
+            OrderedEnumerable<TSource> ordered = source as OrderedEnumerable<TSource>;
+            if (ordered != null) return ordered.FirstOrDefault(predicate);
             foreach (TSource element in source)
             {
                 if (predicate(element)) return element;
@@ -1503,6 +1509,8 @@ namespace System.Linq
         public static TSource Last<TSource>(this IEnumerable<TSource> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.Last();
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1531,6 +1539,8 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
+            OrderedEnumerable<TSource> ordered = source as OrderedEnumerable<TSource>;
+            if (ordered != null) return ordered.Last(predicate);
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1565,6 +1575,8 @@ namespace System.Linq
         public static TSource LastOrDefault<TSource>(this IEnumerable<TSource> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.LastOrDefault();
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1593,6 +1605,8 @@ namespace System.Linq
         {
             if (source == null) throw Error.ArgumentNull("source");
             if (predicate == null) throw Error.ArgumentNull("predicate");
+            OrderedEnumerable<TSource> ordered = source as OrderedEnumerable<TSource>;
+            if (ordered != null) return ordered.LastOrDefault(predicate);
             IList<TSource> list = source as IList<TSource>;
             if (list != null)
             {
@@ -1712,23 +1726,29 @@ namespace System.Linq
         public static TSource ElementAt<TSource>(this IEnumerable<TSource> source, int index)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.ElementAt(index);
             IList<TSource> list = source as IList<TSource>;
             if (list != null) return list[index];
-            if (index < 0) throw Error.ArgumentOutOfRange("index");
-            using (IEnumerator<TSource> e = source.GetEnumerator())
+            if (index >= 0)
             {
-                while (true)
+                using (IEnumerator<TSource> e = source.GetEnumerator())
                 {
-                    if (!e.MoveNext()) throw Error.ArgumentOutOfRange("index");
-                    if (index == 0) return e.Current;
-                    index--;
+                    while (e.MoveNext())
+                    {
+                        if (index == 0) return e.Current;
+                        index--;
+                    }
                 }
             }
+            throw Error.ArgumentOutOfRange("index");
         }
 
         public static TSource ElementAtOrDefault<TSource>(this IEnumerable<TSource> source, int index)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IPartition<TSource> partition = source as IPartition<TSource>;
+            if (partition != null) return partition.ElementAtOrDefault(index);
             if (index >= 0)
             {
                 IList<TSource> list = source as IList<TSource>;
@@ -1740,9 +1760,8 @@ namespace System.Linq
                 {
                     using (IEnumerator<TSource> e = source.GetEnumerator())
                     {
-                        while (true)
+                        while (e.MoveNext())
                         {
-                            if (!e.MoveNext()) break;
                             if (index == 0) return e.Current;
                             index--;
                         }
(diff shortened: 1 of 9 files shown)


stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2958,7 +3136,67 @@ IEnumerator IEnumerable.GetEnumerator()
         }
     }
 
-    internal abstract class OrderedEnumerable<TElement> : IOrderedEnumerable<TElement>
+    internal interface IOrderedPartitonable<TElement> : IEnumerable<TElement>
```
MarcelGosselin: Typo in interface name IOrderedPartit_i_onable

stephentoub:
@dotnet-bot test this please

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -122,11 +122,6 @@ public IEnumerator<TSource> GetEnumerator()
 
             public abstract IEnumerable<TSource> Where(Func<TSource, bool> predicate);
 
-            public virtual TSource[] ToArray()
```
hackcraft: Removing this from here was mainly to be able to have both Iterator and IPartition share the same interface-path to ToArray. Some Iterators have a ToArray that ends up going through Buffer twice, so just removing them and letting Buffer do what it would do anyway is a side-effect benefit.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -657,6 +649,9 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
         public static IEnumerable<TSource> Skip<TSource>(this IEnumerable<TSource> source, int count)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            if (count < 0) count = 0;
```
hackcraft: This simplifies further logic. What I'd like to do here is `if (count <= 0) return source;` but that might raise objections in terms of exposed interfaces.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3227,25 +3243,279 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        TElement First();
+        
+        TElement FirstOrDefault();
+        
+        TElement Last();
+
+        TElement LastOrDefault();
+    }
+    
+    internal sealed class EmptyPartition<TElement> : IPartition<TElement>, IEnumerator<TElement>
```
hackcraft: Using this means that the empty-by-definition cases where e.g. a start index is greater than a end index can be considered more simply. It should also be pretty efficient in itself, but that isn't as big a deal since it should also be a relatively rare case.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3283,30 +3819,117 @@ internal override EnumerableSorter<TElement> GetEnumerableSorter(EnumerableSorte
...
+        }
+
         private void QuickSort(int[] map, int left, int right)
         {
             do
             {
                 int i = left;
                 int j = right;
-                int x = map[i + ((j - i) >> 1)];
+                int x = j - i < 8 ? map[i + ((j - i) >> 1)] : MedianOfThree(map, i, j);
```
hackcraft: I introduced MedianOfThree for the other two partitioning sort/select methods, but it may as well be used here too. It helps avoid quicksort's worse cases.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3378,30 +4103,27 @@ internal struct Buffer<TElement>
         internal TElement[] items;
         internal int count;
 
-        internal Buffer(IEnumerable<TElement> source, bool queryInterfaces = true)
```
hackcraft: A side effect of the ToArray changes above is that we can remove this case.

hackcraft:
After some further changes, checks, and squashing, the state of this PR is:

The following operations have changed time complexity on the results of an `OrderBy()`, `OrderBy().ThenBy()`, etc:

Operation|Old Time Complexity|New Time Complexity
-------------|-------------------|-------------------
First()<br>Last()<br>FirstOrDefault()<br>LastOrDefault()|O(n log n) Best<br>O(n log n) Average<br>O(n²) Worse<br>*O(n) Space Compexity*|Θ(n)<br>*O(1) Space Complexity*
ElementAt()<br>ElementAtDefault()|O(n log n) Best<br>O(n log n) Average<br>O(n²) Worse|O(n) Best<br>O(n) Average<br>O(n²) Worse

`ElementAt()` and `ElementAtDefault()` use a Θ(n) fast-path when possible (when it turns out to be the first or last element), which is sometimes O(1) in space and sometimes O(n).

There is also some work to make the worse-case less likely to be hit.

When `.Skip(j).Take(k)` is called on such results (cases of just `Skip()` or just `Take()` can be considered the same, but with `j` or `k` being `0`), then the complexity of obtaining the results for any subsquent operation (or to enumerate) is changed thus:

(`k` here is the number of results actually taken, rather than asked for, so if there are 5 items and you try to take 7, `k` is 5)

Old Time Complexity|New Time Complexity
-------------------|-------------------
O(n log n) Best<br>O(n log n) Average<br>O(n²) Worse<br>|O(n + k log k) Best<br>O(n + k log k) Average<br>O(n²) Worse

Once the results are prepared, enumerating (directly or indirectly) is also reduced from Θ(j + k) to Θ(k).

These ordered `Skip`/`Take` results also have the following changes to their complexity:

Operation|Old Time Complexity|New Time Complexity
-------------|-------------------|-------------------
First()<br>Last()†<br>FirstOrDefault()<br>LastOrDefault()†|O(n log n) Best<br>O(n log n) Average<br>O(n²) Worse<br>|O(n) Best<br>O(n) Average<br>O(n²) Worse
ElementAt()†<br>ElementAtDefault()†|O(n log n) Best<br>O(n log n) Average<br>O(n²) Worse|O(n) Best<br>O(n) Average<br>O(n²) Worse

Those marked † also had a O(n) component that was completely removed, so the lower-order costs are also improved.

All of these use a Θ(n) fast-path when possible, which is sometimes O(1) in space and sometimes O(n).

Obtaining an array of all the above results is also optimsed. Obtaining an array of those iterators that can't special-case the operation themselves is also optimised as a side-effect of that.

hackcraft:
> There is also some work to make the worse-case less likely to be hit.

Which is a complexity not directly relevant and may not be worth it. I had been doing median-of-three, but I don't think that should be done here, so I took it out again. Whether it helps or hinders should probably be considered separately.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -605,6 +594,9 @@ public override TResult[] ToArray()
         public static IEnumerable<TSource> Take<TSource>(this IEnumerable<TSource> source, int count)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            if (count <= 0) return EmptyPartition<TSource>.Instance;
```
hackcraft: Or perhaps Empty<TSource>() would be better. Since `GetEnumerator()` isn't hit in the original form, there's no meaningful observable difference this way, but there could be with Empty<TSource>().

hackcraft: Rect: It does affect garbage collection. That effect is a generally good one (faster collection and perhaps in an earlier generation), but it is an effect that can be observed with finalisers or weak references, so it's not a completely unobservable change.

JonHanna:
@dotnet-bot test this please.
If you would be so kind.

JonHanna:
@dotnet-bot test this please.

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: "return source" might not be a good idea because someone might use Skip(0) as a way to wrap an otherwise mutable source (like an array) into an iterator. I actually have seen code like that.

The idea is to make it hard to get to the underlying data and change it. It might still be possible through some reflection/unsafe code, but that woudl be considerably harder and clearly a dependency on underlying implementation compared to just casting result to T[].

JonHanna: Yep, that's pretty much my thinking as to "objections in term of exposed interfaces" which was a thinko for "objections in terms of exposed objects".

JonHanna:
A quick measure of time improvements. `Consume` here is a method that spins through the resulting enumerable as quickly as possible, calling `MoveNext()` until it returns false. The numbers given are how many times faster the operation in question is for a source enumerable of shuffled numbers of a given size with after the changes in this PR than before.

Enumerable Size|`source.OrderBy(x => x).First()`|`source.OrderBy(x => x).FirstOrDefault()`|`source.OrderBy(x => x).Last()`|`source.OrderBy(x => x).LastOrDefault()`|`Consume(source.OrderBy(x => x).Skip(5).Take(5))`|`Consume(source.OrderBy(x => x).Skip(size - 10).Take(10))`|`Consume(source.OrderBy(x => x).Skip((size - 10) / 2).Take(10))`|`source.OrderBy(x => x).ElementAt(4)`|`source.OrderBy(x => x).ElementAtOrDefault(4)`|`source.OrderBy(x => x).ElementAt(size / 2)`|`source.OrderBy(x => x).ElementAtOrDefault(size / 2)`|`source.OrderBy(x => x).Take(10).ElementAt(4)`|`source.OrderBy(x => x).Take(10).ElementAtOrDefault(4)`
----|----|----|----|----|----|----|----|----|----|----|----|----|----
0||1.51||2.48|1.57|2.56|1.57||1.64||1.33||2.15
1|2.15|1.81|2.02|1.85|3.73|2.99|1.52||2.47|2.15|1.8||2.72
10|3.75|5.97|5.08|4.27|0.85|1.24|1.03|0.87|1.42|2.27|1.41|1.42|1.44
16|6.48|6.91|7.58|5.22|1.37|1.2|1.16|1.72|1.44|1.34|1.09|1.38|1.49
100|5.74|8.08|8.49|7.49|2.19|3.09|2.3|3.72|3.24|3.17|2.79|7.15|7.68
128|6.79|8.54|9.21|9.17|3.08|2.56|2.22|3.93|3.52|3.49|3.71|3.62|3.75
1000|13.01|12.6|14.68|14.24|3.07|6.86|2.44|3.03|3.06|2.43|2.58|3.3|3.29
1024|12.61|13.4|13.66|13.89|4.76|3.89|3.79|5.04|4.65|3.88|3.83|4.19|4.93
10000|16.39|16.77|16.99|17.41|4.58|6.36|4.96|4.53|4.46|4.87|5|4.5|4.45
16381|19.58|22.61|18.99|18.75|7.46|6.97|4.89|7.35|7.26|4.8|4.82|7.46|7.38

jasonwilliams200OK:
@dotnet-bot, test this please (was giving "Jenkins is going to shut down", see => http://stackoverflow.com/a/26219545/863980).

JonHanna:

JonHanna:

JonHanna:
@dotnet-bot  test this, please.

JonHanna:
@VSadov considering your point on not doing `return source`, while I'd already ruled that out in that case there were other cases that did `return this` or otherwise avoided allocations. While those were all at least one step down the chain of calls and I had considered them therefore unlikely to cause problems, I've revised to avoid them anyway to err on the side of caution.

JonHanna:
@dotnet-bot test this, please

JonHanna:

VSadov:
LGTM

JonHanna:
@VSadov @stephentoub anything else this needs?

stephentoub:
Thanks, @JonHanna.  The reason a bunch of your PRs haven't been merged is there are some open questions on the custom license header.  @richlander is working to get those resolved asap.

JonHanna:
Cool, some clarity on that would be a good thing in any case, I don't much mind whether copyright is retained, or assigned, or who it is assigned to, but a clear rule rather than a few different comments on different PRs would be great.

Maxwe11:
Personally :-1: on custom copyright because project already has merged PRs from community with `Microsoft` copyright.

JonHanna:
In terms of what I would consider a good policy on that, I'd say assigning to MS is better than individual copyrights and everyone (including MS) assigning to the .NET Foundation or some other community group would be better still.
As a contributor I remain personally happy either way.

JonHanna:
> Personally :-1: on custom copyright because project already has merged PRs from community with Microsoft copyright.

If using that copyright is acceptable on new files (I was told I was incorrect to do so), I'd rather just go ahead and do so, all the more so with the other PRs I have open (which are going to have an increasing number of merge conflicts with each other, the more I add).

Maxwe11:
>In terms of what I would consider a good policy on that, I'd say assigning to MS is better than individual copyrights and everyone (including MS) assigning to the .NET Foundation or some other community group would be better still.
As a contributor I remain personally happy either way.

I agree with your position, but it raises a bunch of questions. 
What to do with already merged PRs from community?
Who will change them?
Will that changes accepted?
Seems to me, the better solution here is do not allow custom copyrights.

JonHanna:
@Maxwe11 yeah, there's a few different reasons why it's simpler to just assign to a project owner than have individual copyrights, and I'd rather do that, but a comment on a previous PR pointed to https://github.com/dotnet/coreclr/pull/1329#discussion_r36218185 and said it was policy not to have MS in the copyright on new files (though that comment was actually about a file that was based on existing code, so I could ignore it that time). Just assigning to MS would be much better than custom headers.

jasonwilliams200OK:
Yup and there are already nine files in source tree containing your initials: https://github.com/dotnet/corefx/search?utf8=%E2%9C%93&q=jon+hanna.
+1 for the idea of having .NET Foundation copyright (at least for all the community contributions).

JonHanna:
> Yup and there are already nine files in source tree containing your initials: https://github.com/dotnet/corefx/search?utf8=%E2%9C%93&q=jon+hanna.

Well, that's easily changed if I'm allowed to have it as something other than my name there.

> +1 for the idea of having .NET Foundation copyright (at least for all the community contributions).

Though right now I'd rather not have that discussed as just **anything** okayed as the header I'm allowed to use.

JonHanna:
Test Innerloop Windows_NT Debug Build and Test

EngrMHanif:
pulled

axljoker03 (approved):

PR #2404, opened 2015-07-17
BugFix: Process tests were accessing Modules when the process hasn't been loaded yet

Priya91:
fixes bug #2395 

Introducing a stateID, to ensure the process has been loaded before accessing the Module property. This enum will be populated, as we keep finding such state bugs in CI.

cc @stephentoub

### DIFF
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.IO.Pipes;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
 using Xunit;
 
 namespace System.Diagnostics.ProcessTests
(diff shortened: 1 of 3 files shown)


mellinoe:
Is the enum necessary? Why not just something like:

````CSharp
void EnsureModulesLoaded()
{
    while (_process.Modules.Count == 0) ;
}
````
Will there really be a lot of other enum values that would make it worth it?

File: src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs [hunk 1]
```diff
@@ -68,6 +68,16 @@ protected void StartAndKillProcessWithDelay(Process p)
...
+        protected void EnsureState(StateID state)
+        {
+            switch (state)
+            {
+                case StateID.HasLoadedModules:
+                    while (_process.Modules.Count == 0) ;
+                    break;
+            }
+        }
+
```
stephentoub: Forgetting for a moment the actual mechanism by which this is checking to ensure that modules are loaded, I'm not understanding why all of this StateID stuff is needed.  Why isn't this method just:
```C#
protected void EnsureModulesLoaded()
{
    ... // do whatever checks and loading we need for modules
}
```
?

That said, I'm not understanding the ```while (_process.Modules.Count == 0);``` part of the solution.  Won't this fail, for example, with the OS X implementation that returns an empty array?  And even if this did work, is there not a better mechanism available, or should we at least do some sort of safer spin wait in order to avoid this kind of tight loop?

Priya91:
Made the test run only on nonOSX platforms. Removed StateID, and added the check in the method itself.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -187,36 +187,31 @@ public void TestMachineName()
             Assert.NotNull(_process.MachineName);
         }
 
-        [Fact]
+        [Fact, PlatformSpecific(~PlatformID.OSX)]
         public void TestMainModule()
         {
+            // Ensure the process has loaded the modules.
+            while (_process.Modules.Count == 0) ;
```
mellinoe: Is there any chance this could infinitely loop? I guess the individual test method would timeout after a point even if it did, right? Adding an individual timeout for this might give a better error message, though.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTestBase.cs
```diff
@@ -67,7 +67,7 @@ protected void StartAndKillProcessWithDelay(Process p)
             p.Kill();
             Assert.True(p.WaitForExit(WaitInMS));
         }
-
+
```
mellinoe: Might want to undo

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -187,36 +187,31 @@ public void TestMachineName()
...
+            // Check that the mainModule is present in the modules list.
+            bool foundMainModule = false;
+            if (_process.Modules != null)
             {
-                Assert.Equal(CoreRunName, Path.GetFileNameWithoutExtension(mainModule.ModuleName));
-
-                // Check that the mainModule is present in the modules list.
-                bool foundMainModule = false;
-                if (_process.Modules != null)
+                foreach (ProcessModule pModule in _process.Modules)
```
mellinoe: Could we just do Assert.Contains if we are just trying to see if the item is in this list?

mellinoe:
Left a few comments, overall LGTM

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
stephentoub: > I guess the individual test method would timeout after a point even if it did, right?

I don't believe so. If we really need to do this kind of spinning, it'd be better to do it something like:
```C#
Assert.True(SpinWait.SpinUntil(() => _process.Modules.Count == 0, WaitInMS));
```

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -187,36 +187,31 @@ public void TestMachineName()
             Assert.NotNull(_process.MachineName);
         }
 
-        [Fact]
+        [Fact, PlatformSpecific(~PlatformID.OSX)]
```
stephentoub: If we're going to make this test not for OS X, should we add a separate test that's specific to OS X and that verifies nothing bad happens when accessing MainModule? e.g. ```Assert.Null(_process.MainModule);```

stephentoub:
Thanks, Priya.  Other than adding a short test to verify MainModule on OS X and changing how the spinning is done (assuming there's no better solution to the problem than spinning in the first place), LGTM.

stephentoub:
Looks good, thanks.

PR #2407, opened 2015-07-17
Remove unnecessary throw in ProcessManager.GetModulesInfos on Linux

stephentoub:
When I implemented ProcessManager.GetModulesInfo for Linux, I'd thought that if Windows couldn't enumerate the modules, it would always throw an exception.  As evidenced by #2395, turns out that's not the case, so I'm removing the corresponding exception on Linux.


stephentoub:
cc: @Priya91

Priya91:
:+1: LGTM

stephentoub:
@dotnet-bot test this please

Issue #645, opened 2015-02-05
Support long file names on Windows


JeremyKuhne:
Now would seem to be the time to finally tackle the MAX_PATH nightmare. With the Win32FileSystem and other abstractions we can convert incoming paths to their long path equivalent if needed before handing off to the actual Win32 APIs. It appears that all of the APIs currently in use support the extended syntax (\\?\), with the exception of Get/SetDirectory.

ebickle:
A bit of classic reading from the BCL team for anyone digging into this issue:
http://blogs.msdn.com/b/bclteam/archive/2007/02/13/long-paths-in-net-part-1-of-3-kim-hamilton.aspx
http://blogs.msdn.com/b/bclteam/archive/2007/03/26/long-paths-in-net-part-2-of-3-long-path-workarounds-kim-hamilton.aspx
http://blogs.msdn.com/b/bclteam/archive/2008/06/10/long-paths-in-net-part-3-of-3-kim-hamilton.aspx

paulcbetts:
> Now would seem to be the time to finally tackle the MAX_PATH nightmare

No, no, no, no, no. This will never happen, and should never happen. Please do not add implicit long path support to .NET, it will be a nightmare for users, who now have files that they have created in one app are inexplicably unopenable in other apps. 

MAX_PATH being a #define in Win32 means that it is hard-coded into tens of thousands of applications, in a way that is almost impossible to change after-the-fact (i.e. via AppCompat shims, etc). You can't even set this in *new* versions of the SDK because a bunch of apps have made wire / IPC protocols which would include MAX_PATH in the definition, and now they disagree and you've broken things. Using extended syntax *also* bypasses all of the path correction that Win32 does and means that path semantics are now different between apps, depending on their .NET version.  Attempting to "fix" MAX_PATH is simply not tenable in any sane way, *especially* from a non-OS component like the CLR.

smbecker:
The reality is that you already have tons if files that are unopenable by .Net applications because the framework doesn't let you bypass MAX_PATH. Or you have to resort to rewriting System.IO classes using P/Invoke like in the case of AlphaFS. This should be possible to enable long path support from within .Net. Should it be implicit? I could be convinced either way. However, it is possible in Windows and should at least be possible in .Net.

Alexx999:
@paulcbetts  I believe that this change is *really* needed because backwards-compatibility with some ancient software is actually *not* a goal. Even worse, if this kind of backwards compatibility is enforced in such ways.
[...]

PR #2410, opened 2015-07-17
Allow extended syntax for paths

JeremyKuhne:
Initial implementation of extended syntax support for paths on Windows (\\\\?\\). Basic scenarios work, I'll be adding additional tests (and any necessary fixes) in subsequent commits.

This is the first stage of #645- #2411.

### DIFF
--- a/src/Common/src/System/IO/PathInternal.Unix.cs
+++ b/src/Common/src/System/IO/PathInternal.Unix.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 
 namespace System.IO
@@ -8,6 +9,10 @@ namespace System.IO
     /// <summary>Contains internal path helpers that are shared between many projects.</summary>
     internal static partial class PathInternal
     {
+        // Therre is only one invalid path character in Unix
+        private const char InvalidPathChar = '\0';
+        internal static readonly char[] InvalidPathChars = { InvalidPathChar };
+
         /// <summary>Returns a value indicating if the given path contains invalid characters.</summary>
         internal static bool HasIllegalCharacters(string path, bool checkAdditional = false)
         {
@@ -15,17 +20,26 @@ namespace System.IO
 
             foreach (char c in path)
             {
-                // Same as Path.InvalidPathChars, unrolled here for performance
-                if (c == '\0')
+                // Same as InvalidPathChars, unrolled here for performance
+                if (c == InvalidPathChar)
                     return true;
-
-                // checkAdditional is meant to check for additional characters 
-                // permitted in a search path but disallowed in a normal path.
-                // In Windows this is * and ?,but Unix permits such characters
-                // in the filename so checkAdditional is ignored.
             }
 
             return false;
+        }
+
+        internal static int GetRootLength(string path)
+        {
+            PathInternal.CheckInvalidPathChars(path);
+            return path.Length > 0 && IsDirectorySeparator(path[0]) ? 1 : 0;
+        }
+
+        internal static bool IsDirectorySeparator(char c)
+        {
+            // The alternatie directory separator char is the same as the directory separator,
+            // so we only need to check one.
+            Debug.Assert(Path.DirectorySeparatorChar == Path.AltDirectorySeparatorChar);
+            return c == Path.DirectorySeparatorChar;
         }
     }
 }
--- a/src/Common/src/System/IO/PathInternal.Windows.cs
+++ b/src/Common/src/System/IO/PathInternal.Windows.cs
@@ -8,6 +8,29 @@ namespace System.IO
     /// <summary>Contains internal path helpers that are shared between many projects.</summary>
     internal static partial class PathInternal
     {
+        internal const string LongPathPrefix = @"\\?\";
+        internal const string UNCPathPrefix = @"\\";
+        internal const string UNCLongPathPrefixToInsert = @"?\UNC\";
+        internal const string UNCLongPathPrefix = @"\\?\UNC\";
+
+        internal static readonly char[] InvalidPathChars =
+        {
+            '\"', '<', '>', '|', '\0',
+            (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
+            (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
+            (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
+            (char)31
+        };
+
+        internal static readonly char[] InvalidPathCharsWithAdditionalChecks = // This is used by HasIllegalCharacters
+        {
+            '\"', '<', '>', '|', '\0',
+            (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
+            (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
+            (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
+            (char)31, '*', '?'
+        };
+
         /// <summary>
         /// Returns a value indicating if the given path contains invalid characters (", &lt;, &gt;, | 
         /// NUL, or any ASCII char whose integer representation is in the range of 1 through 31), 
@@ -19,19 +42,67 @@ namespace System.IO
 
             Contract.Requires(path != null);
 
-            foreach (char c in path)
-            {
-                // Note: Same as Path.InvalidPathChars, unrolled here for performance
-                if (c == '\"' || c == '<' || c == '>' || c == '|' || c < 32)
-                    return true;
+            // Question mark is a normal part of extended path syntax (\\?\)
+            int startIndex = path.StartsWith(LongPathPrefix, StringComparison.Ordinal) ? LongPathPrefix.Length : 0;
+            return path.IndexOfAny(checkAdditional ? InvalidPathCharsWithAdditionalChecks : InvalidPathChars, startIndex) >= 0;
+        }
 
-                // used when path cannot contain search strings.
-                if (checkAdditional &&
-                    (c == '?' || c == '*'))
-                    return true;
+        /// <summary>
+        /// Gets the length of the root of the path (drive, share, etc.).
+        /// </summary>
+        internal static int GetRootLength(string path)
+        {
+            CheckInvalidPathChars(path);
+
+            int i = 0;
+            int length = path.Length;
+            int volumeSeparatorLength = 2;  // Length to the colon "C:"
+            int uncRootLength = 2;          // Length to the start of the server name "\\"
+
+            bool extendedSyntax = path.StartsWith(LongPathPrefix, StringComparison.Ordinal);
+            bool extendedUncSyntax = extendedSyntax && path.StartsWith(UNCLongPathPrefix, StringComparison.Ordinal);
+            if (extendedSyntax)
+            {
+                // Shift the position we look for the root from to account for the extended prefix
+                if (extendedUncSyntax)
+                {
+                    // "C:" -> "\\?\C:"
+                    uncRootLength = UNCLongPathPrefix.Length;
+                }
+                else
+                {
+                    // "\\" -> "\\?\UNC\"
+                    volumeSeparatorLength += LongPathPrefix.Length;
+                }
             }
 
-            return false;
+            if ((!extendedSyntax || extendedUncSyntax) && length > 0 && IsDirectorySeparator(path[0]))
+            {
+                // UNC or simple rooted path (e.g. "\foo", NOT "\\?\C:\foo")
+
+                i = 1; //  Drive rooted (\foo) is one character
+                if (extendedUncSyntax || (length > 1 && (IsDirectorySeparator(path[1]))))
+                {
+                    // UNC (\\?\UNC\ or \\), scan past the next two directory separators at most
+                    // (e.g. to \\?\UNC\Server\Share or \\Server\Share\)
+                    i = uncRootLength;
+                    int n = 2; // Maximum separators to skip
+                    while (i < length && (!IsDirectorySeparator(path[i]) || --n > 0)) i++;
+                }
+            }
+            else if (length >= volumeSeparatorLength && path[volumeSeparatorLength - 1] == Path.VolumeSeparatorChar)
+            {
+                // Path is at least longer than where we expect a colon, and has a colon (\\?\A:, A:)
+                // If the colon is followed by a directory separator, move past it
+                i = volumeSeparatorLength;
+                if (length >= volumeSeparatorLength + 1 && (IsDirectorySeparator(path[volumeSeparatorLength]))) i++;
+            }
+            return i;
+        }
+
+        internal static bool IsDirectorySeparator(char c)
+        {
+            return c == Path.DirectorySeparatorChar || c == Path.AltDirectorySeparatorChar;
         }
     }
 }
--- a/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
+++ b/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Linq;
 using Xunit;
 
 namespace System.IO.FileSystem.Tests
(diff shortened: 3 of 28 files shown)


JeremyKuhne:
#2411

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -209,5 +210,11 @@ private static unsafe byte[] CreateCryptoRandomByteArray(int byteLength)
             }
             return arr;
         }
+
+        private static bool HasIllegalCharacters(string path, bool checkAdditional)
+        {
+            Contract.Requires(path != null);
+            return path.IndexOfAny(checkAdditional ? InvalidPathCharsWithAdditionalChecks : InvalidPathChars) >= 0;
```
stephentoub: Once this moves here, we can get rid of the InvalidPathCharsWithAdditionalChecks field from the Unix implementation, and just simplify this down to ```return path.IndexOfany(InvalidPathChars) >= 0;```.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -70,30 +70,45 @@ private static bool IsDirectoryOrVolumeSeparator(char c)
...
         // Gets the length of the root DirectoryInfo or whatever DirectoryInfo markers
         // are specified for the first part of the DirectoryInfo name.
-        // 
         private static int GetRootLength(string path)
         {
             CheckInvalidPathChars(path);
 
             int i = 0;
             int length = path.Length;
+            int volumeSeparator = 2;
```
stephentoub: volumeSeparatorLength?

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -70,30 +70,45 @@ private static bool IsDirectoryOrVolumeSeparator(char c)
...
+                {
+                    uncRootLength = UNCLongPathPrefix.Length;
+                }
+                else
+                {
+                    volumeSeparator += LongPathPrefix.Length;
+                }
+            }
+
+            if ((!extendedSyntax || extendedUncSyntax) && length >= 1 && IsDirectorySeparator(path[0]))
```
stephentoub: Earlier extendedUncSyntax is set to false if extendedSyntax is false, so in what circumstance here would extendedSyntax be false but extendedUncSyntax be true?

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -70,30 +70,45 @@ private static bool IsDirectoryOrVolumeSeparator(char c)
...
+            {
+                if (extendedUncSyntax)
+                {
+                    uncRootLength = UNCLongPathPrefix.Length;
+                }
+                else
+                {
+                    volumeSeparator += LongPathPrefix.Length;
+                }
+            }
```
stephentoub: Comments explaining what these manipulations are doing would be helpful.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -107,7 +122,9 @@ public static string GetFullPath(string path)
 
             // Emulate FileIOPermissions checks, retained for compatibility
             CheckInvalidPathChars(fullPath, true);
-            if (fullPath.Length > 2 && fullPath.IndexOf(':', 2) != -1)
+
+            int startIndex = fullPath.StartsWith(LongPathPrefix, StringComparison.Ordinal) ? LongPathPrefix.Length + 2 : 2;
```
stephentoub: There are lots of magic numbers floating around here (a bunch of twos so far).  It'd be helpful if these were put into named constants to help describe what they are.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -120,12 +137,87 @@ private unsafe static string NormalizePath(string path, bool fullCheck, int maxP
         {
             Contract.Requires(path != null, "path can't be null");
 
+            // If the path is in extended syntax, we don't need to normalize, but we still do some basic validity checks
+            if (path.StartsWith(LongPathPrefix, StringComparison.Ordinal))
```
stephentoub: Can all of this long-path-specific validity checking be moved into a separate helper method?

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: changed

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: This is a bit awkward to read to begin with- I was hesitant to change the existing block too much and made it even more difficult I'm afraid. I don't want to get into this block if it is *just* extendedSyntax.

I'm giving a lot of thought to rewriting this method for clarity in the future.

stephentoub: I see.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: Adding a bunch more

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: Sure, I'll do that.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: I'll add some.

src/Common/src/System/IO/PathInternal.Unix.cs
```diff
@@ -8,6 +9,9 @@ namespace System.IO
     /// <summary>Contains internal path helpers that are shared between many projects.</summary>
     internal static partial class PathInternal
     {
+        internal static readonly char[] InvalidPathChars = { '\0' };
+        internal static readonly char[] InvalidPathCharsWithAdditionalChecks = InvalidPathChars; // no additional checks on Unix
```
stephentoub: What's using these on Unix?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -8,6 +8,29 @@ namespace System.IO
...
+
+        internal static readonly char[] InvalidPathCharsWithAdditionalChecks = // This is used by HasIllegalCharacters
+        {
+            '\"', '<', '>', '|', '\0',
+            (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
+            (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
+            (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
+            (char)31, '*', '?'
+        };
+
```
stephentoub: Is there a way we can avoid defining all of these things twice, once in Path and once in PathInternal?  For example, would it make sense to have these in a shared location, with a #if used to change the name of the type they're included in?  I realize we already have some duplication, but I was unhappy with that to begin with, and this just makes it worse.  I'm ok with going ahead with this to unblock things, but I'd love to see a subsequent solution imposed.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +42,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
+        /// <summary>
+        /// Gets the length of the root of the path (drive, share, etc.).
+        /// </summary>
+        internal static int GetRootLength(string path)
+        {
+            CheckInvalidPathChars(path);
+
+            int i = 0;
+            int length = path.Length;
+            int volumeSeparatorLength = 2;  // Length to the colon "C:"
```
stephentoub: Length including the colon, right?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +42,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
+            int uncRootLength = 2;          // Length to the start of the server name "\\"
+
+            bool extendedSyntax = path.StartsWith(LongPathPrefix, StringComparison.Ordinal);
+            bool extendedUncSyntax = extendedSyntax && path.StartsWith(UNCLongPathPrefix, StringComparison.Ordinal);
```
stephentoub: If extendedSyntax is true, we already know that the first four characters of path match the first four characters of UNCLongPathPrefix.  We could compare just the next four characters and save a few comparisons (though I doubt that'd register on any benchmark).

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +42,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
+                    uncRootLength = UNCLongPathPrefix.Length;
+                }
+                else
+                {
+                    // "\\" -> "\\?\UNC\"
+                    volumeSeparatorLength += LongPathPrefix.Length;
+                }
+            }
+
+            if ((!extendedSyntax || extendedUncSyntax) && length >= 1 && IsDirectorySeparator(path[0]))
```
stephentoub: length > 0?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +42,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
+                    // (e.g. to \\?\UNC\Server\Share or \\Server\Share\)
+                    i = uncRootLength;
+                    int n = 2; // Maximum separators to skip
+                    while (i < length && (!IsDirectorySeparator(path[i]) || --n > 0)) i++;
+                }
+            }
+            else if (length >= volumeSeparatorLength && path[volumeSeparatorLength - 1] == Path.VolumeSeparatorChar)
+            {
+                // Path is at least longer than where we expect a colon, and has a colon (\\?\A:, A:)
+                // If the colon is followed by a directory separator, move past it
```
stephentoub: I have to admit, I'm finding all of this logic very convoluted.  It may be that it has to be that way, but it'd be really helpful if you could find a way to simplify it.  For example, volumeSeparatorLength is initially 2, but it gets increased earlier in a block that's only hit if extendedSyntax and !extendedUncSyntax.  Then here I need to correlate that with the fact that I'm in an else block which can only be hit if extendedSyntax and !extendedUncSyntax, but written inverted and with other conditions involved.  Could you combine that previous set of checks to only update the lengths here when it's actually needed, keeping the changes closer to where they'll actually be used?

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: (Another thing to consider would be including PathInternal* in System.Runtime.Extensions, keeping all of these shared definitions in PathInternal, and then having the public Path just delegate to PathInternal for the shared things.)

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +42,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
-                    return true;
-
-                // used when path cannot contain search strings.
-                if (checkAdditional &&
-                    (c == '?' || c == '*'))
-                    return true;
+                // UNC or simple rooted path (e.g. "\foo", NOT "\\?\C:\foo")
+
+                i = 1; //  Drive rooted (\foo) is one character
+                if (extendedUncSyntax || (length >= 2 && (IsDirectorySeparator(path[1]))))
```
stephentoub: length > 1?  I'm mentioning these because I find it easier to understand the check and the subsequent usage when the values are the same, e.g. I'm verifying that the length is greater than 1 and then I use the value at position 1.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -257,12 +275,12 @@ public void UnixWhiteSpaceAsPath_Allowed()
         }
 
         [Fact]
-        [PlatformSpecific(PlatformID.Windows)] 
+        [PlatformSpecific(PlatformID.Windows)]
         public void WindowsNonSignificantTrailingWhiteSpace()
```
stephentoub: Should the test name be changed to match?

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: (Looks like you're a step ahead of me... when I got further in the review, I see that's exactly what you did :wink:)

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -53,66 +35,99 @@ public static partial class Path
...
-                // handles UNC names and directories off current drive's root.
-                i = 1;
-                if (length >= 2 && (IsDirectorySeparator(path[1])))
+                // UNC specific checks
+                if (path.Length == PathInternal.UNCLongPathPrefix.Length || path[PathInternal.UNCLongPathPrefix.Length] == DirectorySeparatorChar)
                 {
-                    i = 2;
-                    int n = 2;
-                    while (i < length && (!IsDirectorySeparator(path[i]) || --n > 0)) i++;
+                    // Effectively empty and therefore invalid (\\?\UNC\ or \\?\UNC\\)
```
stephentoub: I'm unclear on the second part of this condition.  Can you not have a path like "\\?\UNC\\something"?

stephentoub:
cc: @ianhays, @ericstj

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: And the backslash if it exists, or the server\share with or without a slash... It gets a little verbose, which is why I kept it a little vague. Maybe putting the super explicit details in a remarks block...

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Yeah, I was thinking the same thing. I erred towards clarity for the initial implementations. My thought was it would be best to go back and fine tune for perf later based on sets of various orders of magnitude and types (100/1000 characters, relative/non-relative, 100/10K/100K paths etc.).

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Sure, I'll change it.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Changed this too.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I'm a little worried because it was so convoluted to begin with that if I simplified it I'd introduce subtle regressions (assuming I'd totally rewrite it for clarity). I do want to revisit this, but I'd rather do that along with the performance pass. Doing it in place would be easier to track a bigger change as well I think (as opposed to now, where I moved the block of code).

src/Common/src/System/IO/PathInternal.Unix.cs
JeremyKuhne: Er, nothing for the additional checks. :) Path.GetInvalidPathChars() uses the first. Missed that I didn't move the usage.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
JeremyKuhne: I'm wondering what term I should use? NonExtended? Legacy?

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: Same goes for \\server. You need at least a server\share in either case. In this check you can't have \\server\\share. Windows is very particular about UNCs- they must use backslashes and exactly the right amount (2 & 1) before you get past the server\share. (At that point slashes get normalized.)

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
stephentoub: How about ```WindowsTrailingWhitespace`` and ```WindowsExtendedSyntaxTrailingWhitespace```?  The vast majority of these FileSystem tests use normal syntax and don't call it anything special.

JeremyKuhne: That makes sense, thanks.

JeremyKuhne:
@dotnet-bot Test this please

JeremyKuhne:
Note that I'm starting to build up the follow-up PR for this.

JeremyKuhne:
@stephentoub I took my additional changes and put a combined PR in #2453. Whatever is easiest to get this through- either we take this and I rebase the other and proceed or I close and we move to the other.

I have no idea why the build isn't going through on this- it passed fine and I made a last change to rename the test methods you had pointed out.

stephentoub:
@dotnet-bot test this please

JeremyKuhne:
@dotnet-bot test this please

PR #2412, opened 2015-07-17
Fix DriveInfo.DriveType to not throw on Unix

### LINKED ISSUE #2398: DriveInfo.DriveType may throw UnauthorizedAccessException on Linux
If the associated mount point requires privileges to query, an unauthorized access exception will result.  We should verify what behavior happens in a similar situation on Windows and either maintain this behavior or switch to simply returning Unknown rather than throwing.

stephentoub:
On Windows this property never throws, but instead returns a value in the returned enum that indicates failure when failure occurs.  On Unix we were throwing in such cases, and this commit fixes that.

Fixes #2398

### DIFF
--- a/src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
+++ b/src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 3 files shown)


stephentoub:
cc: @sokket, @mspeterhon

sokket:
LGTM

stephentoub:
Thanks.

PR #2415, opened 2015-07-18
Remove some erroneous / unnecessary asserts in PipeStream on Unix

stephentoub:
The ReadAsyncCore / WriteAsyncCore methods were erroneously asserting for CanRead/CanWrite, when the implementation should instead have been failing with an exception.  It subsequently does, but in debug builds the asserts cause it to assert first.

There were also a bunch of asserts that were unnecessary because the values were just verified by the caller and are about to be verified again by the asynchronously invoked method.

I've simply removed all of these asserts.


stephentoub:
cc: @ianhays, @bpschoch

ianhays:
LGTM!

PR #2417, opened 2015-07-19
BugFix: Refresh the process component to ensure Modules is updated.

### LINKED ISSUE #2395: System.Diagnostics.ProcessTests.ProcessTests.TestMainModule failed

Priya91:
To populate the process Modules, after the first OS call, if the modules hasn't been loaded yet, then the modules is filled with an empty collection. Hence subsequent calls to Modules doesn't update with latest loaded modules, since the Modules is not null (an empty collection). Adding process.refresh to spinwait, to ensure that the Modules is updated with the OS call.

Fixes #2395 

cc @stephentoub

### DIFF
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
@@ -200,7 +200,13 @@ namespace System.Diagnostics.ProcessTests
             }
 
             // Ensure the process has loaded the modules.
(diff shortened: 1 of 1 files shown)


stephentoub:
LGTM.

PR #2418, opened 2015-07-19
Fix Process.WaitForExit premature timeout on Unix

stephentoub:
The Process.WaitForExit implementation on Unix is mixing up units, resulting in WaitForExits with timeouts exiting much earlier than the specified timeout.  The code is using Stopwatch.GetTimestamp to get a current timestamp, subtracting the current timestamp from one gathered when the wait started, and then subtracting that difference from the specified timeout.  But the specified timeout is in milliseconds, and the timestamp is in ticks, such that the difference measured in ticks is quickly larger than the timeout value in milliseconds.  The fix just converts the difference from ticks to milliseconds before subtracing from the timeout value.

### DIFF
--- a/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessWaitState.Unix.cs
+++ b/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessWaitState.Unix.cs
@@ -367,7 +367,7 @@ namespace System.Diagnostics
             Debug.Assert(!Monitor.IsEntered(_gate));
 
             // Track the time the we start waiting.
-            long startTime = GetTimestamp();
+            long startTime = Stopwatch.GetTimestamp();
 
             // Polling loop
             while (true)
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @Priya91, @pallavit

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessWaitState.Unix.cs
```diff
@@ -379,7 +379,7 @@ internal bool WaitForExit(int millisecondsTimeout)
                 // We're in a polling loop... determine how much time remains
                 int remainingTimeout = millisecondsTimeout == Timeout.Infinite ?
                     Timeout.Infinite :
-                    (int)Math.Max(millisecondsTimeout - (GetTimestamp() - startTime), 0);
+                    (int)Math.Max(millisecondsTimeout - ((Stopwatch.GetTimestamp() - startTime) / (double)Stopwatch.Frequency * 1000), 0);
```
akoeplinger: I'm curious why this doesn't just use Start/Elapsed of Stopwatch?

stephentoub: It could, but then a Stopwatch would need to be allocated.  If the code were any more complex, it'd probably be worthwhile, but as it stands, seems better to just do this manually and avoid the unnecessary allocation.

akoeplinger: Makes sense :+1:

Priya91:
LGTM I was just looking at this code, and was about to ask you about this.. :)

stephentoub:
:)

PR #2420, opened 2015-07-20
Update to build tools 59.

JeremyKuhne:
This version is needed to run tests against changes in partial
facades.


stephentoub:
If it passes CI, LGTM.

PR #2423, opened 2015-07-20
Fix currently inconsequential Path bug on Unix

stephentoub:
InvalidPathCharsWithAdditionalChecks is used by CheckInvalidPathChars when checkAdditional is true.  The only caller of this method that sets it to true is in the Windows-specific code paths, which is why none of our tests on Unix fail due to InvalidPathCharsWithAdditionalChecks accidentally being set to InvalidFileNameChars instead of to the intended InvalidPathChars.  I'm fixing it now in case this is ever used in the future, at which time the '/' check would cause real problems :)


stephentoub:
cc: @JeremyKuhne

JeremyKuhne:
LGTM

PR #2426, opened 2015-07-20
Tighter loops on Enumerable.Min, .Max & .Average.

### LINKED ISSUE #2238: Linq loops could be tighter; causes bug in one obscure case, waste in others.

hackcraft:
Fixes #2238.

Changes the flow of these aggregate operations to use tighter loops.

A test of the performance changes can be see at
https://github.com/hackcraft/Enumerable-Tester/blob/master/Enumerable%20Aggregate%20Performance.ods?raw=true

Notably:

1. Most tested uses showed an increase in performance, sometimes dramatically.
2. The only cases with a decrease in performance greater than 5% were in short
paths (e.g. single-element sequences) and so the total performance impact is
slight.
3. For other cases that showed a decrease in performance there were similar
cases (same operation on same datatype) with an increase in performance either
comparable or overwhelming the decrease.

These changes make no short-circuits as per #2350 though they do take a short-cut to the end of the sequence in those cases. #2350 still offers yet greater improvements in the applicable cases.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1756,32 +1756,48 @@ namespace System.Linq
         public static int Min(this IEnumerable<int> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            int value = 0;
-            bool hasValue = false;
-            foreach (int x in source)
+            int value;
+            using (IEnumerator<int> e = source.GetEnumerator())
             {
-                if (hasValue)
+                if (!e.MoveNext()) throw Error.NoElements();
+                value = e.Current;
+                while (e.MoveNext())
                 {
+                    int x = e.Current;
                     if (x < value) value = x;
                 }
-                else
-                {
-                    value = x;
-                    hasValue = true;
-                }
             }
-            if (hasValue) return value;
-            throw Error.NoElements();
+            return value;
         }
 
         public static int? Min(this IEnumerable<int?> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
             int? value = null;
-            foreach (int? x in source)
+            using (IEnumerator<int?> e = source.GetEnumerator())
             {
-                if (value == null || x < value)
-                    value = x;
+                // Start off knowing that we've a non-null value (or exit here, knowing we don't)
+                // so we don't have to keep testing for nullity.
+                do
+                {
+                    if (!e.MoveNext()) return value;
+                    value = e.Current;
+                } while (!value.HasValue);
+                // Keep hold of the wrapped value, and do comparisons on that, rather than
+                // using the lifted operation each time.
+                int valueVal = value.GetValueOrDefault();
+                while (e.MoveNext())
+                {
+                    int? cur = e.Current;
+                    int x = cur.GetValueOrDefault();
+                    // Do not replace & with &&. The branch prediction cost outweighs the extra operation
+                    // unless nulls either never happen or always happen. 
+                    if (cur.HasValue & x < valueVal)
+                    {
+                        valueVal = x;
+                        value = cur;
+                    }
+                }
             }
             return value;
         }
@@ -1789,70 +1805,109 @@ namespace System.Linq
         public static long Min(this IEnumerable<long> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            long value = 0;
-            bool hasValue = false;
-            foreach (long x in source)
+            long value;
+            using (IEnumerator<long> e = source.GetEnumerator())
             {
-                if (hasValue)
+                if (!e.MoveNext()) throw Error.NoElements();
+                value = e.Current;
+                while (e.MoveNext())
                 {
+                    long x = e.Current;
                     if (x < value) value = x;
                 }
-                else
-                {
-                    value = x;
-                    hasValue = true;
-                }
             }
-            if (hasValue) return value;
-            throw Error.NoElements();
+            return value;
         }
 
         public static long? Min(this IEnumerable<long?> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
             long? value = null;
-            foreach (long? x in source)
+            using (IEnumerator<long?> e = source.GetEnumerator())
             {
-                if (value == null || x < value) value = x;
+                do
+                {
+                    if (!e.MoveNext()) return value;
+                    value = e.Current;
+                } while (!value.HasValue);
+                long valueVal = value.GetValueOrDefault();
+                while (e.MoveNext())
+                {
+                    long? cur = e.Current;
+                    long x = cur.GetValueOrDefault();
+                    // Do not replace & with &&. The branch prediction cost outweighs the extra operation
+                    // unless nulls either never happen or always happen. 
+                    if (cur.HasValue & x < valueVal)
+                    {
+                        valueVal = x;
+                        value = cur;
+                    }
+                }
             }
             return value;
         }
-
+        
         public static float Min(this IEnumerable<float> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            float value = 0;
-            bool hasValue = false;
-            foreach (float x in source)
+            float value;
+            using (IEnumerator<float> e = source.GetEnumerator())
             {
-                if (hasValue)
+                if (!e.MoveNext()) throw Error.NoElements();
+                value = e.Current;
+                while(e.MoveNext())
                 {
+                    float x = e.Current;
+                    if (x < value) value = x;
                     // Normally NaN < anything is false, as is anything < NaN
                     // However, this leads to some irksome outcomes in Min and Max.
                     // If we use those semantics then Min(NaN, 5.0) is NaN, but
                     // Min(5.0, NaN) is 5.0!  To fix this, we impose a total
                     // ordering where NaN is smaller than every value, including
                     // negative infinity.
-                    if (x < value || System.Single.IsNaN(x)) value = x;
-                }
-                else
-                {
-                    value = x;
-                    hasValue = true;
+                    // Not testing for NaN therefore isn't an option, but since we
+                    // can't find a smaller value, we can short-circuit. But we consume
+                    // the rest for backwards-compatibility reasons.
+                    else if (float.IsNaN(x))
+                    {
+                        while (e.MoveNext()) {}
+                        return x;
+                    }
                 }
             }
-            if (hasValue) return value;
-            throw Error.NoElements();
+            return value;
         }
 
         public static float? Min(this IEnumerable<float?> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
             float? value = null;
-            foreach (float? x in source)
+            using (IEnumerator<float?> e = source.GetEnumerator())
             {
-                if (x == null) continue;
-                if (value == null || x < value || System.Single.IsNaN((float)x)) value = x;
+                do
+                {
+                    if (!e.MoveNext()) return value;
+                    value = e.Current;
+                } while (!value.HasValue);
+                float valueVal = value.GetValueOrDefault();
+                while (e.MoveNext())
+                {
+                    float? cur = e.Current;
+                    if (cur.HasValue)
+                    {
+                        float x = cur.GetValueOrDefault();
+                        if (x < valueVal)
+                        {
+                            valueVal = x;
+                            value = cur;
+                        }
+                        else if (float.IsNaN(x))
+                        {
+                            while (e.MoveNext()) { }
+                            return cur;
+                        }
+                    }
+                }
             }
             return value;
         }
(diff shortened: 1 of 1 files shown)


hackcraft:
The linked spreadsheet has been annotated as follows:

Relative improvement is given as a the difference between the new and old times as a percentage of the old time.

Relative improvements of 5% or more have a green background.

Relative improvements of 1–5% have a light green background.

Relative improvements of 0–1% have a very light green background.

Relative deterioration of 5% or more have a red background.

Relative deterioration of 1-5% have a pink background.

Relative deterioration of 0-1% have a light pink background.

Test cases which use a random sequence of more than 1 element are bold, as more likely to represent real data.

Those which use an increasing or decreasing sequence are not bold, as more artificial examples intended just to catch if they are a bad worse-case.

Those which use empty sequences or sequences of only one element are not bold, as less affected by performance changes.

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1756,32 +1756,50 @@ public static decimal Sum<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
             {
-                if (value == null || x < value)
-                    value = x;
+                if (!e.MoveNext()) return null;
+                value = e.Current;
+                // Start off knowing that we've a non-null value (or exit here, knowing we don't)
+                // so we don't have to keep testing for nullity.
+                while (!value.HasValue)
+                {
+                    if (!e.MoveNext()) return value;
```
stephentoub: Would it be more clear to return ```null``` here rather than returning ```value``` which we know to be ```null```?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1756,32 +1756,50 @@ public static decimal Sum<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
             if (source == null) throw Error.ArgumentNull("source");
-            int? value = null;
-            foreach (int? x in source)
+            int? value;
+            using (IEnumerator<int?> e = source.GetEnumerator())
             {
-                if (value == null || x < value)
-                    value = x;
+                if (!e.MoveNext()) return null;
+                value = e.Current;
```
stephentoub: Couldn't this be merged with the subsequent loop if a do..while was used instead? e.g.
```C#
do
{
    if (!e.MoveNext()) return null;
    value = e.Curent;
}
while (!value.HasValue);
```

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1789,70 +1807,113 @@ public static int Min(this IEnumerable<int> source)
...
+            return value;
         }
 
         public static long? Min(this IEnumerable<long?> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            long? value = null;
-            foreach (long? x in source)
+            long? value;
+            using (IEnumerator<long?> e = source.GetEnumerator())
```
stephentoub: My comments on the Min overloads for int and int? are applicable to the overloads for other types, but I'll refrain from making the same comments there. :)

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Yes, but surprisingly more expensive. I profiled a few should-make-no-difference variations and went with those that won the profiling, though I was a bit more consistent than the profiling suggested I should be, in only allowing one variant to be used across all equivalent methods.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1789,70 +1807,113 @@ public static int Min(this IEnumerable<int> source)
...
-                else
-                {
-                    value = x;
-                    hasValue = true;
+                    // Not testing for NaN therefore isn't an option, but since we
+                    // can't find a smaller value, we can short-circuit. But we consume
+                    // the rest for backwards-compatibility reasons.
+                    else if (float.IsNaN(x))
+                    {
+                        while (e.MoveNext()) {}
```
stephentoub: Nice compromise.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: As with the other line comment, I profiled a few different options that should theoretically make no difference, and went with what came out as faster.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: How much more expensive?  I personally find the do...while version to be more understandable, and this code is going to be repeated for a bunch of these overloads, so there is some benefit to clarity.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: What setup were you using for the profiling?  Were you using the new 64-bit JIT?

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: (In term of pure code, as well as preferring a `do…while` loop I'd also prefer to declare `value` closer to the first assignment, but that also had a surprising difference; not massive, but definitely there).

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Oh. Yes I should have pointed that out in the description. Shall make a note…

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2015,31 +2119,68 @@ public static decimal Min<TSource>(this IEnumerable<TSource> source, Func<TSourc
...
+                }
+                int valueVal = value.GetValueOrDefault();
+                if (valueVal >= 0)
+                {
+                    // We can fast-path this case where we know HasValue will
+                    // never affect the outcome, without constantly checking
+                    // if we're in such a state. A similar fast-path could
+                    // be done for Min on negative values, but as all-positive
+                    // or mostly-positive values are quite common in real-world
+                    // uses, it's only been done in this direction.
```
stephentoub: Clever.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2079,32 +2251,61 @@ public static long Max(this IEnumerable<long> source)
...
-                if (value == null || x > value || System.Double.IsNaN((double)value)) value = x;
+                if (!e.MoveNext()) return null;
+                value = e.Current;
+                while (!value.HasValue)
+                {
+                    if (!e.MoveNext()) return value;
+                    value = e.Current;
+                }
+                double valueVal = value.GetValueOrDefault();
+                while (double.IsNaN(valueVal))
```
stephentoub: Comments here would be helpful.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2181,32 +2420,37 @@ public static TSource Max<TSource>(this IEnumerable<TSource> source)
...
             if (value == null)
             {
-                foreach (TSource x in source)
+                using (IEnumerator<TSource> e = source.GetEnumerator())
                 {
-                    if (x != null && (value == null || comparer.Compare(x, value) > 0))
-                        value = x;
+                    if (!e.MoveNext()) return value;
+                    value = e.Current;
+                    // A do…while may seem more natural here, but profiled worse across a wide set of test data
```
stephentoub: I'm surprised it was measurable in this case.  In addition to the two interface calls for every element, we also have the interface call to the comparer now for each element below.  When you say it profiled worse, was that just in the extreme case where things are dominated by nulls in the input sequence?  I wouldn't optimize for that case.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Not a major impact, varying from the `do…` actually winning by less than 0.1% up to a bit over 1%. I was surprised, as if anything I'd have thought the `do…while` would have been the more concise in terms of generated code, as well as source code.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: I decided though that doing it everywhere it could be applied would be too clever for its own good, and restricted it to non-negative integers on `Max`.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: I tried a great many variants, but I did consider null-free and low-null to be more important. I also thought consistency between the methods more important than if a micro-opt was available by taking one approach for one datatype and another approach for another.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: I ran it on .Net 4.5.2 64-bit on Windows. I did a couple of runs and found them close to the one I turned into that spreadsheet.
I did shortened versions on mono/Linux and 32bit built, just to check nothing stood out as very different, but just checking there wasn't a surprise case were things were made worse. That was before I added in the cases of empty and single elements though.

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: As I remember a regular while{} is emitted as a do{}while  with entry point branching to a label just before the exit condition. This way the looping code path has just one branch.

It is a bit surprising that while{] cost more. It could be something with branch prediction, or perhaps the "return null" sequence causes some effects. - It would need to zero-init a temp, and then copy it into a return temp (since we are in a try/finally), not sure if JIT can follow this. If it is all about "return null", then you may be able to squeeze some cycles by pre-initing value to null and then just goto to the "return value".

VSadov: Something like:
```C#
                do
                {
                    if (!e.MoveNext()) goto exit; // goes to "return value"
                    value = e.Current;
                }
                while (!value.HasValue);
```
I did not time it. It should result in a more compact code, but only measuring can tell if that leads to faster execution.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2181,32 +2420,37 @@ public static TSource Max<TSource>(this IEnumerable<TSource> source)
...
+                    while (value == null)
+                    {
+                        if (!e.MoveNext()) return value;
+                        value = e.Current;
+                    } while (value == null);
+                    while (e.MoveNext())
```
VSadov: extra     while (value == null);

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2181,32 +2420,37 @@ public static TSource Max<TSource>(this IEnumerable<TSource> source)
...
-                        value = x;
-                        hasValue = true;
+                        TSource x = e.Current;
+                        if (comparer.Compare(x, value) > 0) value = x;
                     }
                 }
-                if (hasValue) return value;
-                throw Error.NoElements();
             }
+            return value;
```
VSadov: we could use just one using, and it could be interesting to try 

```C#
                    do
                    {
                        if (!e.MoveNext()) goto exit; // goes to "return value"
                        value = e.Current;
                    }
                    while (value == null);
```

VSadov:
Very nice. LGTM modulo couple comments.

hackcraft:
I'll experiment with the suggestions and see what the effect is

mellinoe:
How about .NET Native? We should probably do some testing to determine the effects there, as it's one of the actual real-world users of the library right now, or will be soon.

VSadov:
I think for mostly algorithmic changes, desktop CLR would be a good approximation. Since there is not much of API usage, improvements in JIT-ed scenarios should translate to improvements (or no changes) in AOT scenarios.
I think interface dispatch and EH could have different profiles on AOT systems and could be a concern if changes had introduced more exception handling or interface calls, but that would also likely be an observable change in behavior, so it is not a case here.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Pre-initialising `value` goes against my instincts, but trying it does let us reduce the duplication down to a single `do…while` with pretty negliable effects (slight lose, but in the margin-of-error territory, and a win on empty sequences).
The `goto` isn't necessary. Indeed, I found it compiled to the same IL whether there was a `goto` or a `return`. No need to risk velociraptors.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Proof that I really did try some approaches using `do…while` ;)

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Using just one using makes for a messier inside, and might even risk the jitter not realising it can cut out some null-checking code (though I doubt it). In any case, the advantage of duplicate tests over `do…while` is slight, so I'll go with `do…while` in the name of clarity and concision.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Interestingly, the two cases where the effect of switching to `do…while` was most appreciable was the `int?` and `long?` versions of `Max` where it's actually a few percentage points in some cases. That difference drops a lot if the fast-path they have is taken out, so there seems to be some branching interaction there. Still, even the more expensive `do…while` is still a win on the old version, and consistency is to be favoured, so I'm putting it in there too.

hackcraft:
I applied some changes based on the comments here, merged in the changes to master in the meantime, and squashed, but the CI is failing. It seems to be something awry in Jenkins rather than with the change. If so, is there a way to force another check? If not, am I missing something obvious?

hackcraft:
I knew I'd end up hitting that close button some time when I didn't want to.

PR #2429, opened 2015-07-20
Fix OpenCSP_Failed error string to include format parameter

stephentoub:
All of the usage sites are formatting it with an error code that's currently being ignored, but it should also help find the root cause behind #2396.


stephentoub:
cc: @bartonjs

bartonjs:
LGTM.

I checked to see what Desktop would do in this case, and it appears that it logged the HR into whatever the LOG macro did; then threw an HR-based CryptographicException.

Mainly I was concerned we were impacting whatever loc process we have by changing an existing string, but it appears to be a new string for CoreFX.

Thanks, Stephen :smile:.

stephentoub:
Thanks, Jeremy.

PR #2430, opened 2015-07-20
Reduce allocation threshold of BigIntegerCalculator

axelheer:
The aftermath of #2291 leads to the final conclusion that this is the better value. A bigger threshold doesn't have a noticeable effect.


axelheer:
cc: @stephentoub @Maoni0

stephentoub:
LGTM.  Thanks, again, @axelheer.

PR #2431, opened 2015-07-20
Disable C# 6 features until CI/build system updated

stephentoub:
With Visual Studio 2015 released today (woo hoo!), I expect we're going to see the number of contributors using VS2015 and C# 6 increase, which is a good thing.  Unfortunately, our current build / CI system is currently limited to using the C# 5 compiler.  Speaking from experience, this makes for a frustrating time when one develops a change and everything passes locally, only to submit a PR and find that it fails in CI because a C# 6 feature was being used.

Until we can get our build and CI systems updated to use C# 6 (hopefully very soon), this change simply sets the LangVersion for our C# projects to C# 5 so that, even when using VS2015, attempting to build locally with C# 6 features will fail to compile.  This won't prevent all possible dependencies on C# 6, but it will avert most.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -198,6 +198,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <!-- Temporary until build/CI system is upgraded to C# 6: disable C# 6 features -->
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj' OR '$(Language)' == 'C#'">
+    <LangVersion>5</LangVersion>
+  </PropertyGroup>
+
   <!-- Set up some common paths -->
   <PropertyGroup>
     <CommonPath>$(SourceDir)Common\src</CommonPath>


stephentoub:
cc: @jaredpar, @weshaggard, @mmitche

jaredpar:
LGTM

Note: Moving to the C# 6.0 compiler can be done as an xcopy solution.  It's just a NuGet package that needs to be referenced.  Moving this repo + Jenkins to use it would be a pretty trivial operation.  

Is it an issue with the internal build that is the blocker here?

stephentoub:
Thanks, Jared.

> Is it an issue with the internal build that is the blocker here?

@weshaggard can answer that.

weshaggard:
@jaredpar yes for things that are scoped to building with our buildtools we could attempt to enable what you are suggesting. In fact you have issue https://github.com/dotnet/corefx/issues/1316 assigned to you for doing that.

However for the larger question we do still have source code that builds in various places, including VS 2013 which cannot take a dependency on c# 6.0 yet. We do have some engineering efforts underway to help break this dependency however.

PR #2439, opened 2015-07-21
Fix small discrepancy in TaskExtensions.Unwrap

stephentoub:
This fixes a small behavioral difference between the mscorlib implementation of Unwrap and the corefx implementation of Wnwrap.  A Task can end in a canceled state but contain a CancellationToken that doesn't have cancellation requested, e.g. if TaskCompletionSource.SetCanceled() is used.  When Unwrap proxies such a task, the mscorlib implementation was appropriately reflecting that no-cancellation-requested state, whereas our corefx implementation was pretending it did have cancellation requested. This commit fixes the corefx implementation to match, and adds a few additional test cases to verify.


stephentoub:
cc: @mellinoe, @AlfredoMS

mellinoe:
LGTM

stephentoub:
Thanks, Eric.

PR #2446, opened 2015-07-22
Performance improvements to Skip and SkipWhile.

### LINKED ISSUE #2238: Linq loops could be tighter; causes bug in one obscure case, waste in others.

hackcraft:
Fixes #2238
(And I think this is the last bit to separate out of the original PR)

https://github.com/hackcraft/Enumerable-Tester/raw/master/Skip%20performance.ods
shows a comparison of the old and new approaches. Improvements are made in
the vast majority of cases, and with Skip in particular can be significant.

Edge-cases of disposal and exception ordering are dealt with; duplicating the
behaviour found with the current implementation.

(This ignores the matter of ordered skips as per #2401).

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Threading;
+using System.Runtime.ExceptionServices;
 
 namespace System.Linq
 {
@@ -676,21 +676,103 @@ namespace System.Linq
             }
         }
 
+        // Just in case somebody does something weird with the ordering of enumerator
+        // disposal, this wraps an enumerator that has been used, but not disposed,
+        // and disposes it when this is.
+        // If an exception had been found while skipping, this
+        // throws it on MoveNext.
+        private sealed class DeadEnumeratorWrapper<TSource> : IEnumerator<TSource>
+        {
+            private readonly IEnumerator<TSource> _corpse;
+            private ExceptionDispatchInfo _exception;
+
+            internal DeadEnumeratorWrapper(IEnumerator<TSource> corpse)
+                : this(corpse, null)
+            {
+            }
+
+            internal DeadEnumeratorWrapper(IEnumerator<TSource> corpse, ExceptionDispatchInfo exception)
+            {
+                _corpse = corpse;
+                _exception = exception;
+            }
+
+            public TSource Current
+            {
+                get { return default(TSource); }
+            }
+
+            object IEnumerator.Current
+            {
+                get { return Current; }
+            }
+
+            public bool MoveNext()
+            {
+                if (_exception != null)
+                {
+                    ExceptionDispatchInfo ex = _exception;
+                    _exception = null;
+                    ex.Throw();
+                }
+                return false;
+            }
+
+            public void Dispose()
+            {
+                _corpse.Dispose();
+            }
+
+            public void Reset()
+            {
+                throw new NotSupportedException();
+            }
+        }
+
         public static IEnumerable<TSource> Skip<TSource>(this IEnumerable<TSource> source, int count)
         {
             if (source == null) throw Error.ArgumentNull("source");
-            return SkipIterator<TSource>(source, count);
+            return new SkipIterator<TSource>(source, count);
         }
 
-        private static IEnumerable<TSource> SkipIterator<TSource>(IEnumerable<TSource> source, int count)
+        private sealed class SkipIterator<TSource> : IEnumerable<TSource>
         {
-            using (IEnumerator<TSource> e = source.GetEnumerator())
+            private readonly IEnumerable<TSource> _source;
+            private readonly int _count;
+
+            public SkipIterator(IEnumerable<TSource> source, int count)
             {
-                while (count > 0 && e.MoveNext()) count--;
-                if (count <= 0)
+                _source = source;
+                _count = count > 0 ? count : 0;
+            }
+
+            public IEnumerator<TSource> GetEnumerator()
+            {
+                IEnumerator<TSource> sourceEnumerator = _source.GetEnumerator();
+                try
                 {
-                    while (e.MoveNext()) yield return e.Current;
+                    for (int count = _count; count != 0; --count)
+                    {
+                        // It should be harmless to have the recipient call MoveNext again.
+                        // but we guard against strange behaviour from shortcuts in the enumerator
+                        // It should also be harmless to dispose the enumerator here if we've
+                        // exhausted it, but we guard against strange behaviour in either the
+                        // enumerator or the caller.
+                        if (!sourceEnumerator.MoveNext()) return new DeadEnumeratorWrapper<TSource>(sourceEnumerator);
+                    }
+                    return sourceEnumerator;
                 }
+                catch (Exception ex)
+                {
+                    // Force the exception to happen on a MovedNext from the caller, rather than
+                    // here.
+                    return new DeadEnumeratorWrapper<TSource>(sourceEnumerator, ExceptionDispatchInfo.Capture(ex));
+                }
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -676,38 +676,191 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
+            {
+                _source = source;
+                _predicate = predicate;
+            }
+
+            public IEnumerator<TSource> GetEnumerator()
+            {
+                // Don't short-circuit. These are almost always both true, so don't "save"
+                // time by branching after a very likely case.
+                var ret = _state == 0 & _threadId == Environment.CurrentManagedThreadId
```
hackcraft: Interestingly, I found that just allocating a new object as enumerator beat this approach slightly more often than this did. However, it was only slight and the approach that allocates the least is likely to be the approach that handles memory pressures tests don't catch the best.

stephentoub: It's really going to depend on your benchmark.  With the workstation GC, with a single-threaded benchmark, yeah, I'm not surprised that they'd be in a tight race.  The actual act of allocating an object is usually very cheap, just some pointer manipulations and whatever the cost of invoking the ctor.  The real impact comes later when the GC needs to run, causing GCs to take longer and forcing the GC to run more often because there's more garbage to be cleaned up.  Such costs are typically less noticeable when collections can be done concurrently with the running code and when machine isn't saturated.

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -676,38 +676,191 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
                 {
-                    while (e.MoveNext()) yield return e.Current;
+                    for (int count = _count; count != 0; --count)
+                    {
+                        // It should be harmless to have the recipient call MoveNext again.
+                        // but we guard against strange behaviour from shortcuts in the enumerator
+                        // It should also be harmless to dispose the enumerator here if we've
+                        // exhausted it, but we guard against strange behaviour in either the
+                        // enumerator or the caller.
+                        if (!sourceEnumerator.MoveNext()) return new DeadEnumeratorWrapper<TSource>(sourceEnumerator);
```
stephentoub: Even with the heroic acts being done here to minimize visible differences in behavior, this still has the very visible difference that where work is being done is being moved.  If I previously had:
```C#
var source = Enumerable.Range(0, Int32.MaxValue).Skip(Int32.MaxValue - 1);
var e = source.GetEnumerator();
```
previously that would have been very cheap and quick.  With this change, that's going to iterate through 2 billion elements.  That makes me nervous.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Ah, of course. And I was generally taking time to ensure GC didn't have an effect, because the difficulty in predicting could lead one or the other attempt to have unfair results.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Hmm. Calling `GetEnumerator()` and not calling `MoveNext()` even once almost immediately seems pretty obscure, even to someone who has a strange love for playing with enumerators like myself, but certainly not impossible.
The only way around that is to wrap it entirely. While that proved to be a win with `SkipWhile` I wouldn't expect it to be as big a win, and perhaps not worth the effort.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -676,38 +676,191 @@ private static IEnumerable<TSource> TakeWhileIterator<TSource>(IEnumerable<TSour
...
+            internal SkipWhileIterator(IEnumerable<TSource> source, Func<TSource, bool> predicate)
+            {
+                _source = source;
+                _predicate = predicate;
+            }
+
+            public IEnumerator<TSource> GetEnumerator()
+            {
+                // Don't short-circuit. These are almost always both true, so don't "save"
+                // time by branching after a very likely case.
```
VSadov: I'd keep this &&. 
My guess would be that CurrentManagedThreadId could cost more than a branch (not sure how it runs on Mono or things like Micro framework, whether it needs to do volatile reads etc...)  and there is no dependency between && operands so on a beefy machine we may expect that CPU will have some success at speculating/predicting.

Basically, I think if all goes well, there won't be much effect from skipping a branch here, but if CurrentManagedThreadId happens to be expensive on a given platform, it may turn against us.

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: The other problem with ExceptionDispatchInfo is debuggability. It preserves stack trace, but not stack itself. Basically if you have to do a postmortem analysis of a crash dump, you will not be able to dig through the application state at the time of the crash. You will have the state preserved at the time of ex.Throw() which is not all that useful. 
ExceptionDispatchInfo is a necessary evil in async. It is still already inconvenient enough to force us using exception filters in Roslyn to guarantee robust and immediate failfast in async code.
In iterators the dead iterator trick with ExceptionDispatchInfo would be just to gain some perf and IMO is probably not worth the overall trouble.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Can you see `_state ==0` not massively outnumbering other cases?
I'd imagine it does, and as such a small saving in that case pays for a greater cost in other cases.

VSadov: _state==0 is generally more common. However there are cases when it is not the case and in those cases the opposite is true.

I have seen cases (in Roslyn, before we switched most of the APIs to use ImmutableArray) when public object model is based on lazy-initialized IEnumerable wrappers of internal structures. That is the case where it is common for the same enumerable to be enumerated over and over and CurrentManagedThreadId was noticeable in profiles in a form of CPU cache misses. - not sure if it forces some volatile semantics or just because it touches data too far from the iterator instance. That prompted swapping the operands of && in generated code. Old compiler did the threadId compare first, for no particular reason.

I believe the     stateCheck && threadIdCheck     pattern is a nice compromise that gives up just a little in an optimal scenario (_state == 0) while does not penalize us too much in other scenarios.

hackcraft: I've come across `CurrentManagedThreadId` being expensive enough to worry about in efficiency pushes when doing some highly concurrent code in the past, but was assuming state not being zero would be rare. Since you report that not being the case I'll change this when I'm next not only on a phone.

src/System.Linq/src/System/Linq/Enumerable.cs
hackcraft: Okay. I'm inclined to abandon this chunk entirely then. Unless someone comes up with a "hey, but you could just…" that saves it in the next short time, I'll close this PR.

stephentoub:
> Unless someone comes up with a "hey, but you could just…" that saves it in the next short time, I'll close this PR

Thanks, @hackcraft.  Should we go ahead and close this?

hackcraft:
I think I'll see what happens if I redo Skip much as SkipWhile is done here (adjusted as per what @VSadov said about the `&&` in the thread check), since that it won't be a big job to get from here to there and while it may not be as dramatic an improvement it might still be worth it now I nearly have it, but as I won't have time to do that for a while, for now I'm going to just close this out.

hackcraft:
Okay, I took a look at redoing `Skip` in the same way that I had `SkipWhile` here (and also using `&&` rather than `&` on the thread test as per what @VSadov suggested), and results are not bad at all. https://github.com/hackcraft/Enumerable-Tester/raw/master/new%20version%20skip%20improvements.ods shows the improvements `Skip` now has over the current implementation, and while it compares well with my first attempt while not being too clever for its own good as my first perhaps was.
This version should also have no backwards-compatibility issues at all.

Issue #2444, opened 2015-07-22
Open items for System.Data.Common


saurabh500:
The following needs to be incorporated based on the comments in the pull request https://github.com/dotnet/corefx/pull/2197

Please note that the code is being submitted via another pull request https://github.com/dotnet/corefx/pull/2443

1. Investigate the CancellationToken for the Tasks
2. There are more comments made by Stephen Toub, which need to be taken care of.

Refer to https://github.com/dotnet/corefx/pull/2197

saurabh500:
The issues have been taken care of by PRs from contributors.

PR #2447, opened 2015-07-22
Cleanup CancellationToken and Task usage in System.Data.Common

stephentoub:
Three things addressed:

1. CancellationTokenRegistrations were only being unregistered in the case of failure, not in the case of success, so registrations were being leaked into CancellationToken instances.  Fixed to always clean up.

2. CancellationToken.Register calls were resulting in a new delegate allocation on each call.  Fixed to pass the relevant state through as object state and to let the compiler cache the delegate for the lambda.

3. The code had some legacy helpers to create already-completed tasks.  These are no longer necessary.  Fixed by removing them and replacing call sites with usage of Task.FromException/Canceled.

#2444

### DIFF
--- a/src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
+++ b/src/System.Data.Common/src/System/Data/Common/AdapterUtil.cs
@@ -16,21 +16,6 @@ namespace System.Data.Common
         // location so that the catcher of the exception will have the appropriate call stack.
         // This class is used so that there will be compile time checking of error messages.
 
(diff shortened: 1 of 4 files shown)


stephentoub:
cc: @saurabh500

saurabh500:
@stephentoub  looking into it

stephentoub:
Thanks.  The second and third listed changes are about performance and code maintainability, but the first one is about correctness.

stephentoub:
@saurabh500, just checking in on this.  Are you still evaluating?  Do you need any assistance?

PR #2451, opened 2015-07-22
Add missing unit tests for System.Environment class

Sridhar-MS:
The following public APIs from System.Enironment did not have any unit test coverage
- Environment.NewLine.cs
- Environment.ProcessorCount.cs
- Environment.TickCount.cs
- Environment.StackTrace.cs

### DIFF
--- a/src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
+++ b/src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
@@ -18,9 +18,6 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Diagnostics\Stopwatch.cs" />
-    <Compile Include="System\Environment.NewLine.cs" />
-    <Compile Include="System\Environment.ProcessorCount.cs" />
-    <Compile Include="System\Environment.StackTrace.cs" />
     <Compile Include="System\Runtime\Versioning\FrameworkName.cs" />
     <Compile Include="System\IO\Path.cs" />
     <Compile Include="System\IO\Path.Combine.cs" />
@@ -48,7 +45,11 @@
     <Compile Include="System\Convert.ToUInt64.cs" />
     <Compile Include="System\Environment.ExpandEnvironmentVariables.cs" />
     <Compile Include="System\Environment.GetEnvironmentVariable.cs" />
+    <Compile Include="System\Environment.NewLine.cs" />
+    <Compile Include="System\Environment.ProcessorCount.cs" />
     <Compile Include="System\Environment.SetEnvironmentVariable.cs" />
+    <Compile Include="System\Environment.StackTrace.cs" />
+    <Compile Include="System\Environment.TickCount.cs" />
     <Compile Include="System\Math.cs" />
     <Compile Include="System\Progress.cs" />
     <Compile Include="System\Random.cs" />
@@ -59,7 +60,6 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
-    <Compile Include="System\Environment.TickCount.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\src\System.Runtime.Extensions.CoreCLR.csproj">
--- a/src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
+++ b/src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
@@ -18,6 +18,9 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Diagnostics\Stopwatch.cs" />
+    <Compile Include="System\Environment.NewLine.cs" />
+    <Compile Include="System\Environment.ProcessorCount.cs" />
+    <Compile Include="System\Environment.StackTrace.cs" />
     <Compile Include="System\Runtime\Versioning\FrameworkName.cs" />
     <Compile Include="System\IO\Path.cs" />
     <Compile Include="System\IO\Path.Combine.cs" />
@@ -56,6 +59,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
+    <Compile Include="System\Environment.TickCount.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\src\System.Runtime.Extensions.CoreCLR.csproj">
--- a/src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
+++ b/src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
@@ -9,21 +9,18 @@ namespace System.Runtime.Extensions.Tests
 {
     public class EnvironmentNewLine
     {
+        [PlatformSpecific(PlatformID.Windows)]
         [Fact]
-        public void NewLineTest()
+        public void Windows_NewLineTest()
         {
-            //arrange
-            string expectedNewLine = Interop.IsWindows ? "\r\n" : "\n";
+            Assert.Equal("\r\n", Environment.NewLine);
+        }
 
-            //act
-            string actualNewLine = Environment.NewLine;
-
-            //assert
-            Assert.Equal(expectedNewLine, actualNewLine);
+        [PlatformSpecific(PlatformID.AnyUnix)]
+        [Fact]
+        public void Unix_NewLineTest()
+        {
+            Assert.Equal("\n", Environment.NewLine);
         }
     }
-
-
-
-
 }
--- a/src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
+++ b/src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace System.Runtime.Extensions.Tests
+{
+    public class EnvironmentNewLine
+    {
+        [Fact]
+        public void NewLineTest()
+        {
+            //arrange
+            string expectedNewLine = Interop.IsWindows ? "\r\n" : "\n";
+
+            //act
+            string actualNewLine = Environment.NewLine;
+
+            //assert
+            Assert.Equal(expectedNewLine, actualNewLine);
+        }
+    }
+
+
+
+
+}
--- a/src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
+++ b/src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace System.Runtime.Extensions.Tests
+{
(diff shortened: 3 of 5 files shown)


src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
```diff
@@ -56,6 +59,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
+    <Compile Include="System\Environment.TickCount.cs" />
```
stephentoub: This should be up with the others, right?

src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace System.Runtime.Extensions.Tests
+{
+    public class EnvironmentNewLine
+    {
+        [Fact]
+        public void NewLineTest()
+        {
+            //arrange
+            string expectedNewLine = Interop.IsWindows ? "\r\n" : "\n";
```
stephentoub: How about just having two different tests based on OS? e.g.
```C#
[Fact]
public void Windows_NewLineTest()
{
    Assert.Equal("\r\n", Environment.NewLine);
}

[Fact]
public void Unix_NewLineTest()
{
     Assert.Equal("\n", Environment.NewLine);
}
```

src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
```diff
@@ -0,0 +1,29 @@
...
+
+            //assert
+            Assert.Equal(expectedNewLine, actualNewLine);
+        }
+    }
+
+
+
+
+}
```
stephentoub: Nit: lots of blank lines

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
```diff
@@ -0,0 +1,64 @@
...
+            internal int dwPageSize;
+            internal IntPtr lpMinimumApplicationAddress;
+            internal IntPtr lpMaximumApplicationAddress;
+            internal IntPtr dwActiveProcessorMask;
+            internal int dwNumberOfProcessors;
+            internal int dwProcessorType;
+            internal int dwAllocationGranularity;
+            internal short wProcessorLevel;
+            internal short wProcessorRevision;
+            [DllImport("api-ms-win-core-sysinfo-l1-1-0.dll", SetLastError = true)]
```
stephentoub: Looks strange to have the import declared in the struct.

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
```diff
@@ -0,0 +1,64 @@
...
+            internal static extern void GetSystemInfo(ref SYSTEM_INFO lpSystemInfo);
+        }
+
+
+    }
+
+    
+
+    
+}
```
stephentoub: Nit: lots of blank lines... only one is needed

src/System.Runtime.Extensions/tests/System/Environment.StackTrace.cs
```diff
@@ -0,0 +1,70 @@
...
+            List<string> knownFrames = new List<string>()
+            {
+                "System.Runtime.Extensions.Tests.EnvironmentStackTrace.StaticFrame(Object obj)",
+                "System.Runtime.Extensions.Tests.EnvironmentStackTrace.TestClass..ctor()",
+                "System.Runtime.Extensions.Tests.EnvironmentStackTrace.GenericFrame[T1,T2](T1 t1, T2 t2)",
+                "System.Runtime.Extensions.Tests.EnvironmentStackTrace.TestFrame()"
+            };
+
+            //act
+            Task.Run(() => TestFrame()).Wait();
```
stephentoub: Why the Task.Run(() => ...).Wait() rather than just calling TestFrame() directly?

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        [Fact]
+        public void TickCountTest()
+        {
+            //arrange
+            const int milliSeconds = 1000;
+            int start = Environment.TickCount;
+
+            //act
+            Task.Delay(milliSeconds).Wait();
```
stephentoub: We've been trying hard to avoid artificial delays like this, especially in inner loop tests.  At the very least please shorten the delay significantly, e.g. 10 milliseconds rather than a second.

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
```diff
@@ -0,0 +1,30 @@
...
+            int end = Environment.TickCount;
+
+            //assert
+            Console.WriteLine("Start - " + start);
+            Console.WriteLine("End - " + end);
```
stephentoub: Please remove these Console.WriteLines.  We don't want any console output in successful test runs.

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
```diff
@@ -0,0 +1,30 @@
...
+            Assert.True(end - start >= milliSeconds);
```
stephentoub: I know you found and were fixing a bug around TickCount being the time since the system started rather than since the epoch.  Should there be a test for that here?

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            if(Interop.IsWindows)
+            {
+                SYSTEM_INFO sysInfo = new SYSTEM_INFO();
+                SYSTEM_INFO.GetSystemInfo(ref sysInfo);
+                expected = sysInfo.dwNumberOfProcessors;
+            }
+            else
+            {
+                int _SC_NPROCESSORS_ONLN = Interop.IsOSX ? 58 : 84;
```
stephentoub: Since this is now specific to a particular flavor of Unix, consider looking up the right value for FreeBSD and including it.  Alternatively, to make sure this is flagged for when the FreeBSD port team starts going through the se tests, you could assert that either IsOSX or IsLinux.

src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
stephentoub: (But it's up to you.  I personally prefer the splitting test aproach whenever there's not a lot of code being duplicated, but both approaches work.)

src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
```diff
@@ -18,6 +18,9 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System\Diagnostics\Stopwatch.cs" />
+    <Compile Include="System\Environment.NewLine.cs" />
+    <Compile Include="System\Environment.ProcessorCount.cs" />
+    <Compile Include="System\Environment.StackTrace.cs" />
```
stephentoub: All of the Environment files should ideally be grouped together.

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
```diff
@@ -0,0 +1,64 @@
...
+            //act
+            int actual = Environment.ProcessorCount;
+
+            //assert
+            Assert.True(actual > 0);
+            Assert.Equal(expected, actual);
+        }
+
+        [DllImport("libc")]
+        private static extern long sysconf(int name);
```
stephentoub: We're going to need to rationalize these test P/Invokes with the work @nguerrera is doing to P/Invoke into shims rather than directly into the underlying native library.  We can go with this for now, but we'll need to deal with this soon, as such declarations end up suffering from all of the same problems we're facing in our product code, e.g. the size of data types (the result here will probably be int on a 32-bit system, we're having to hardcode the different name values on each Unix platform, etc.)

stephentoub:
Thanks for the additional tests, Sridhar.  I left a few comments.

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
mellinoe: That seems like a good solution, given that this is a bit of a tricky thing to test to begin with.

stephentoub:
@dotnet-bot test this please

src/System.Runtime.Extensions/tests/System/Environment.StackTrace.cs
Sridhar-MS: Calling on the current thread makes the StackTrace string to contains ~50 frames including the ones from xunit. I wanted the StackTrace string to be simpler for easy diagnosis. Hence started the frame on a new Thread.

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
Sridhar-MS: Got the right value for _SC_NPROCESSORS_ONLN from - https://github.com/freebsd/freebsd/blob/master/include/unistd.h

src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
Sridhar-MS: The changes were automatically done by Visual Studio. I have fixed it.

src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
Sridhar-MS: Makes sense. Done.

src/System.Runtime.Extensions/tests/System/Environment.NewLine.cs
Sridhar-MS: Done.

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
Sridhar-MS: Moved the dllimport outside of the struct.

src/System.Runtime.Extensions/tests/System/Environment.ProcessorCount.cs
Sridhar-MS: Fixed.

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
Sridhar-MS: @stephentoub solution seems much better so copying that.

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
Sridhar-MS: Fixed

src/System.Runtime.Extensions/tests/System.Runtime.Extensions.Tests.csproj
Sridhar-MS: Fixed

src/System.Runtime.Extensions/tests/System/Environment.TickCount.cs
Sridhar-MS: Adding a test for that is a bit tricky. Environment.TickCount returns a signed Int32 in which both positive and negative numbers are valid TickCounts. The bug in PAL implementation cast the 64bit int epoch time into a negative 32bit signed int and returned it, but the value returned is in the valid range for Environment.TickCount.

PR #2452, opened 2015-07-23
Fixing the TimeZoneInfo tests on Linux

eerhardt:
Fixing the TimeZoneInfo tests on Linux now that FindSystemTimeZoneById and Local are implemented.

The TimeZone Ids are different between Windows and Linux, so I needed to adjust the tests to have the correct Id on each platform.

### DIFF
--- a/src/System.Runtime/tests/System/TimeZoneInfo.cs
+++ b/src/System.Runtime/tests/System/TimeZoneInfo.cs
@@ -2,19 +2,22 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Runtime.InteropServices;
 using Xunit;
 
 public static class TimeZoneInfoTests
 {
-    private static String s_strPacific = "Pacific Standard Time";
-    private static String s_strSydney = "AUS Eastern Standard Time";
-    private static String s_strGMT = "GMT Standard Time";
-    private static String s_strTonga = "Tonga Standard Time";
-    private static String s_strBrasil = "E. South America Standard Time";
-    private static String s_strPerth = "W. Australia Standard Time";
-    private static String s_strBrasilia = "E. South America Standard Time";
-    private static String s_strNairobi = "E. Africa Standard Time";
-    private static String s_strAmsterdam = "W. Europe Standard Time";
+    private static String s_strPacific = Interop.IsWindows ? "Pacific Standard Time" : "America/Los_Angeles";
+    private static String s_strSydney = Interop.IsWindows ? "AUS Eastern Standard Time" : "Australia/Sydney";
+    private static String s_strGMT = Interop.IsWindows ? "GMT Standard Time" : "Europe/London";
+    private static String s_strTonga = Interop.IsWindows ? "Tonga Standard Time" : "Pacific/Tongatapu";
+    private static String s_strBrasil = Interop.IsWindows ? "E. South America Standard Time" : "America/Sao_Paulo";
+    private static String s_strPerth = Interop.IsWindows ? "W. Australia Standard Time" : "Australia/Perth";
+    private static String s_strBrasilia = Interop.IsWindows ? "E. South America Standard Time" : "America/Sao_Paulo";
+    private static String s_strNairobi = Interop.IsWindows ? "E. Africa Standard Time" : "Africa/Nairobi";
+    private static String s_strAmsterdam = Interop.IsWindows ? "W. Europe Standard Time" : "Europe/Berlin";
+    private static String s_strRussian = Interop.IsWindows ? "Russian Standard Time" : "Europe/Moscow";
+    private static String s_strLibya = Interop.IsWindows ? "Libya Standard Time" : "Africa/Tripoli";
 
     private static TimeZoneInfo s_myUtc = TimeZoneInfo.Utc;
     private static TimeZoneInfo s_myLocal = TimeZoneInfo.Local;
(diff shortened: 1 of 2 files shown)


src/System.Runtime/tests/System/TimeZoneInfo.cs
```diff
@@ -159,15 +160,20 @@ public static void NearMinMaxDateTimeOffsetConvertTest()
         VerifyConvert(new DateTimeOffset(DateTime.MinValue.AddHours(5), new TimeSpan(-3, 0, 0)), s_strPacific, new DateTimeOffset(DateTime.MinValue, new TimeSpan(-8, 0, 0)));
 
         VerifyConvert(DateTime.MaxValue, s_strPacific, s_strSydney, DateTime.MaxValue);
-        VerifyConvert(DateTime.MaxValue.AddHours(-19), s_strPacific, s_strSydney, DateTime.MaxValue);
+        if (Interop.IsWindows) // [ActiveIssue(846, PlatformID.AnyUnix)]
```
stephentoub: Just so I know, what is this part blocked on for Unix?

stephentoub:
LGTM (once the underlying TimeZoneInfo changes are merged).  Thanks!

stephentoub:
Interesting; I'd have expected the CI build to fail, since the mscorlib changes to support it haven't been checked in yet.  @eerhardt, @mmitche, do you know what's going on here?  Is my understanding incorrect?

src/System.Runtime/tests/System/TimeZoneInfo.cs
eerhardt: This is blocked by the issue I raised in email regarding how to handle AdjustmentRules on Linux.

We are not handling daylight savings time after 2038 correctly for Australia/Sydney because
1. We are not reading the V2 version information.
2. We aren't taking into account the POSIX style rule on how to handle daylight savings after the last transition in the file: i.e. - AEST-10AEDT,M10.1.0,M4.1.0/3, so everything after 2038 stays on one offset for now.

I plan on fixing this issue in my next change after this round of PRs.

eerhardt:
Does the CI build run on Linux?  I guess I was under the assumption it wasn't working yet.

stephentoub:
> Does the CI build run on Linux? I guess I was under the assumption it wasn't working yet.

It does.  @mmitche is investigating what's going on... something apparently changed very recently such that now only a small portion of our tests are being run in the Linux CI, whereas they all used to (well, all that are relevant to Linux).

src/System.Runtime/tests/System/TimeZoneInfo.cs
```diff
@@ -96,7 +99,7 @@ public static void ValidateLibyaTimeZoneTest()
     [ActiveIssue(846, PlatformID.AnyUnix)]
```
bartonjs: From what I can tell, this is the last issue 846 attribute remaining in the file (aside from the partial one for Australia).  Is there a second issue here, or was this one just missed?

src/System.Runtime/tests/System/TimeZoneInfo.cs
stephentoub: Ok, thanks.  At this point, we might want a separate issue specific to cleaning up everything remaining in time zone support on Unix, and tag each remaining thing in the code with that new number.  That'll make it easier to know when we're "done" with it.  The 846 grab-bag of globalization issues is pretty big.

eerhardt: I've created a new issue for this:  https://github.com/dotnet/corefx/issues/2465

src/System.Runtime/tests/System/TimeZoneInfo.cs
eerhardt: There are still 8 tests marked with 846 attributes.  I've logged https://github.com/dotnet/corefx/issues/2465 to handle parsing the tzfile correctly, which should take care of most of them.  As I find other issues, I will log them as I go.

stephentoub:
@dotnet-bot test this please

mj1856:
Wait - Are we saying that with this change we are going to have `TimeZoneInfo` use whatever time zone information is available on the host machine?

Existing code like below will fail on Linux because there's no Windows registry data to feed `TimeZoneInfo`. 

````csharp
TimeZoneInfo tzi = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");
```

With this change we're saying we want to allow:

````csharp
TimeZoneInfo tzi = TimeZoneInfo.FindSystemTimeZoneById("America/Los_Angeles");
```

That's a breaking change IMHO.  You can't pick up code that's been running on Windows and move it to Linux.  The different hosts will have different identifiers.

Did I miss where this decision was already made?

IMHO the `TimeZoneInfo` class needs to be changed to a provider model.  One provider could pull from the host OS, but other providers may need to distribute their own data as embedded resources, or pull from online sources, etc.  Jumping right to using the host time zone info without considering this is troubling.

+CC  @tarekgh

tarekgh:
@mj1856 this change I believe is just a step to unblock the TimeZoneInfo in Linux. I agree with you we'll need to have some extra work to map the Ids if someone want to write code works for Windows and Linux at same time. I am not seeing this a breaking change though as we are supporting the Ids which is supported on the host.

mj1856:
As long as it's temporary, then sure.   It just seems like it will be a sticking point with regard to code portability.

eerhardt:
I discussed the Id difference between Windows and Linux with @joshfree, and he felt it was a secondary priority to get consistent Ids cross platforms.  The higher priority was just getting TimeZoneInfo working at all on Linux.
So, like @tarekgh said, this is just a step forward and there will be plenty more changes in this area to come.

eerhardt:
I've opened https://github.com/dotnet/corefx/issues/2538 for this issue.

tarekgh:
Thanks @eerhardt

mj1856:
Thanks.

eerhardt:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

Issue #2411, opened 2015-07-17
Allow long path syntax for Windows


JeremyKuhne:
Enable \\\\?\ formatted paths.

jakesays:
The BCL team did something like this a while ago: [LongPath](http://bcl.codeplex.com/wikipage?title=Long%20Path)

JeremyKuhne:
Part of #645.

JeremyKuhne:
We're good to go with \\\\?\\.

PR #2453, opened 2015-07-23
Allow extended syntax for paths

JeremyKuhne:
Additional work on allowing extended paths. Fixes a few issues,
adds a number of tests, normalizes naming a bit more and moves/adds
helpers to shared code.

One nice side effect of this change is that you can now delete directories
that were undeletable before (if you had subdirectories that had trailing
spaces as one example).

This is the first stage of #645- #2411.
This builds on #2410 (currently a superset of the change there).

@stephentoub 
cc: @ianhays, @ericstj


src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -2,12 +2,121 @@
...
 
 using System.Diagnostics.Contracts;
+using System.Text;
 
 namespace System.IO
 {
     /// <summary>Contains internal path helpers that are shared between many projects.</summary>
     internal static partial class PathInternal
     {
+        internal const string ExtendedPathPrefix = @"\\?\";
```
JeremyKuhne: Note that I changed the terminology here to better reflect what it does. While it does allow long paths it also allows access to additional filenames and changes resolution behavior.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -19,19 +128,67 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
...
-                    return true;
+                // Shift the position we look for the root from to account for the extended prefix
+                if (extendedUncSyntax)
+                {
+                    // "C:" -> "\\?\C:"
+                    uncRootLength = UncExtendedPathPrefix.Length;
+                }
+                else
+                {
+                    // "\\" -> "\\?\UNC\"
```
MarcelGosselin: It looks to me like this comment and the one above in the `if` branch have been inverted.

JeremyKuhne: Whoops- thanks :)

JeremyKuhne:
@stephentoub Updated for merge of #2410

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -32,6 +33,91 @@ internal static partial class PathInternal
...
         /// <summary>
+        /// Adds the extended path prefix (\\?\) if not already present.
+        /// </summary>
+        internal static string AddExtendedPathPrefix(string path)
+        {
+            if (IsExtended(path))
+                return path;
+
+            if (IsExtendedUnc(path))
+                return path.Insert(2, PathInternal.UncExtendedPrefixToInsert); // Given \\server\share in longpath becomes \\?\UNC\server\share  => UNCLongPathPrefix + path.SubString(2); => The actual command simply reduces the operation cost.
```
stephentoub: Could you move the comment to its own line / split it across lines?  It's quite long.

stephentoub: But I'm confused... aren't we at this return statement if we've already verified this is already an extended unc path?  Why are we then inserting the extended prefix again?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -32,6 +33,91 @@ internal static partial class PathInternal
...
+        /// Returns true if the path uses the extended UNC syntax (\\?\UNC\)
+        /// </summary>
+        internal static bool IsExtendedUnc(StringBuilder path)
+        {
+            return path != null && path.StartsWithOrdinal(UncExtendedPathPrefix);
+        }
+
+        private static bool StartsWithOrdinal(this StringBuilder builder, string value)
+        {
+            if (value == null || builder.Length < value.Length) return false;
```
stephentoub: nit: could you put the return on its own line?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -32,6 +33,91 @@ internal static partial class PathInternal
...
+
+            for (int i = 0; i < value.Length; i++)
+            {
+                if (builder[i] != value[i]) return false;
```
stephentoub: ditto.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -43,7 +129,7 @@ internal static bool HasIllegalCharacters(string path, bool checkAdditional = fa
             Contract.Requires(path != null);
 
             // Question mark is a normal part of extended path syntax (\\?\)
-            int startIndex = path.StartsWith(LongPathPrefix, StringComparison.Ordinal) ? LongPathPrefix.Length : 0;
+            int startIndex = PathInternal.IsExtended(path) ? ExtendedPathPrefix.Length : 0;
```
stephentoub: Are we now going to be doing many more null checks because each of these helpers does one?  Or do those get optimized away?

src/System.IO.FileSystem/src/System/IO/PathHelpers.Windows.cs
```diff
@@ -40,9 +40,16 @@ internal static string GetFullPathInternal(string path)
...
-
-            return Path.GetFullPath(Path.IsPathRooted(pathTrimmed) ? pathTrimmed : path);
+            if (PathInternal.IsExtended(path))
+            {
+                // Don't want to trim extended paths
+                return Path.GetFullPath(path);
+            }
+            else
+            {
+                string pathTrimmed = path.TrimStart(TrimStartChars).TrimEnd(TrimEndChars);
```
stephentoub: At some point (doesn't need to be now), it'd be nice to consolidate the start and end trimming into a single operation, so that if both need to trim, we create a single string rather than two.

src/System.IO.FileSystem/tests/Directory/Delete.cs
```diff
@@ -91,6 +91,15 @@ public void ShouldThrowIOExceptionDeletingCurrentDirectory()
         #region PlatformSpecific
 
         [Fact]
+        public void WindowsExtendedDirectoryWithSubdirectories()
```
stephentoub: ```[PlatformSpecific(PlatformID.Windows)]``` ?

src/System.IO.FileSystem/tests/Directory/Delete.cs
```diff
@@ -91,6 +91,15 @@ public void ShouldThrowIOExceptionDeletingCurrentDirectory()
...
+        {
+            DirectoryInfo testDir = Directory.CreateDirectory(@"\\?\" + GetTestFilePath());
```
stephentoub: How about adding a GetTestExtendedFilePath() helper?

src/System.IO.FileSystem/tests/Directory/Exists.cs
```diff
@@ -115,6 +115,42 @@ public void DirectoryLongerThanMaxPathAsPath_DoesntThrow()
 
         [Fact]
         [PlatformSpecific(PlatformID.Windows)]
+        public void ValidExtendedPathExists_ReturnsTrue()
+        {
+            Assert.All((IOInputs.GetValidPathComponentNames()), (component) =>
+            {
+                string path = @"\\?\" + Path.Combine(TestDirectory, "extended", component);
```
stephentoub: And maybe an ```ExtendedTestDirectory``` property?

stephentoub:
Left some minor comments, but otherwise LGTM.

stephentoub:
(though I thought the previous PR was about allowing extended syntax for paths... so this probably needs a revised commit title/description)

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Yeah, that's wrong. Fixing. Considering adding a test project for just the "common" bits- thoughts? Idea would be add a test project to the common directory that includes the common files and can test the internals that way.

stephentoub: Sorry, I'm not following the test project suggestion....can you elaborate?

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I did consider it- I was assuming the null check would be trivial compared to the method call and StartsWith cost. Also concerned about this being shared code and easily leaking in a rarely hit null. I'll be paying close attention when I do a perf pass through the scenarios.

src/System.IO.FileSystem/src/System/IO/PathHelpers.Windows.cs
JeremyKuhne: Agree- string op scrubbing is something I really want to do.

src/System.IO.FileSystem/tests/Directory/Delete.cs
JeremyKuhne: Yep, thanks. :) Failed the Unix tests with this one. :)

src/System.IO.FileSystem/tests/Directory/Delete.cs
JeremyKuhne: I'll add some shared test helpers with the next set of changes.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Done

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I can create a new test project in \Common\Tests that actually includes these shared files and better unit test these helpers without needing InternalsVisibleTo. I've been mulling it over for a few days and I think it makes a lot of sense.

JeremyKuhne:
Thanks. I updated the commit comments- they were still reflecting the combined nature of the first PR.

PR #2456, opened 2015-07-23
Fix warning from cmake about RPATH on Mac OS X

nguerrera:
cc @stephentoub

### DIFF
--- a/src/Native/CMakeLists.txt
+++ b/src/Native/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(CoreFX)
 
+set(CMAKE_MACOSX_RPATH ON)
 set(CMAKE_INSTALL_PREFIX $ENV{__CMakeBinDir})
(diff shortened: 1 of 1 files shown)


src/Native/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(CoreFX)
 
+set(CMAKE_MACOSX_RPATH ON)
```
stephentoub: Seems like newer cmakes have this on by default?
http://www.cmake.org/cmake/help/v3.0/policy/CMP0042.html

stephentoub:
LGTM

src/Native/CMakeLists.txt
nguerrera: Yes, the warning I get tells me to be explicit about it since the behavior is different across versions. I am using 3.x, but I inherited the minver=2.8.12 from coreclr. Not sure if we can just bump minver.

PR #2464, opened 2015-07-23
Generate and use config.h

nguerrera:
cc @stephentoub 

Previously, I had neglected to actually generate and use an appropriate config.h, which had the effect that HAVE_STAT64 and HAVE_STAT_BIRTHTIME were always treated as false.


stephentoub:
LGTM. Thanks, Nick.

PR #2478, opened 2015-07-24
Remove extraneous ProjectTypeGuids causing load failures

stephentoub:
Several projects are no longer loading correctly in VS for me.  It's happening for projects and only for projects that contain a <ProjectTypeGuids/> node in the .csproj, related to portable libraries.  These entries are unnecessary, and I'm simply deleting them, after which the projects load in VS successfully. (I'm also removing some unnecessary good in one of our .slns that's causing unnecessary hierarchy in Solution Explorer.)


bartonjs:
6896000d173190ad945fac32834c2d30815d8a29 LGTM

Issue #2243, opened 2015-07-03
Required VS 2010 build tools


colombod:
Hi, wasn't around for a while, did a full clean build on full clean Win10, VS2015RC only. On the requirements for building the repo seems this should be enough.

But when running the build script I get this

:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.Cpp.Platform.targets(55,5): error MSB8020: The build t
ools for Visual Studio 2010 (Platform Toolset = 'v100') cannot be found. To build using the v100 build tools, please in
stall Visual Studio 2010 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting t
he Project menu or right-click the solution, and then selecting "Retarget solution". [C:\Users\colombod\Documents\coref
x\src\System.ServiceProcess.ServiceController\tests\System.ServiceProcess.ServiceController.TestNativeService\System.Se
rviceProcess.ServiceController.TestNativeService.vcxproj]

stephentoub:
VS2015RC should definitely work. When you installed VS2015, did you elect to install C++?

colombod:
Yes got full install, can compile from within VS (the toolset is inherited) but if I run command line to build it seems to look for V100. I used VS cmd, not default one.

Is a fresh win10 + vs2015 setup no other things installed so far

stephentoub:
@weshaggard, any thoughts?

weshaggard:
I actually hit this myself on a new Win10 + VS 2015 machine setup. It looks like VS 2015 by default doesn't include the common C++ tools. To fix this you need modify your VS 2015 install to be sure to check the following box:

Programming Languages | Visual C++ | Common Tools for Visual C++ 2015

Once that is installed then you should be able to build this project successfully. I also submitted PR #2494 to mention this in our build instructions. 

I'm closing this issue based on the information but please do let me know if you still have trouble.

colombod:
Thanks will do immediately!

colombod:
Seems working, but anyway i noticed that part of common c++ tools is Windows Xp compatibility AKA V100 tool chain -> VS 2010 :)

PR #2494, opened 2015-07-24
Update instructions to point to VS 2015 RTM

weshaggard:
Also mention the need to include the Common C++ tools.

This should help with issue #2243

### DIFF
--- a/Documentation/building/windows-instructions.md
+++ b/Documentation/building/windows-instructions.md
@@ -11,7 +11,9 @@ Visual Studio 2013 (Update 3 or later) or Visual Studio 2015 (Preview or later) 
 
 The following free downloads are compatible:
 * [Visual Studio Community 2013 (with Update 3)](http://www.visualstudio.com/en-us/visual-studio-community-vs.aspx)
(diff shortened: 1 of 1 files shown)


Documentation/building/windows-instructions.md
```diff
@@ -11,7 +11,9 @@ Visual Studio 2013 (Update 3 or later) or Visual Studio 2015 (Preview or later)
 
 The following free downloads are compatible:
 * [Visual Studio Community 2013 (with Update 3)](http://www.visualstudio.com/en-us/visual-studio-community-vs.aspx)
```
stephentoub: Visual Studio 2013 Update 5 is available.  Should that text be updated, too?

stephentoub:
LGTM, with one question/comment.

Documentation/building/windows-instructions.md
weshaggard: I don't know much about Update 5 but if it isn't required as I don't believe it is I would just keep it as is it is for now.

PR #2495, opened 2015-07-24
Fix new c++ compiler warning after moving to C++ 14

weshaggard:
nativetestservice.cpp(140): warning C4477: 'wprintf' : format string '%s' requires an argument of type 'wchar_t *', but variadic argument 1 has type 'std::wstring *'


weshaggard:
Hmmm... good point although CI and my local builds passed so I would have expected at least some test failures. I will take a closer look.

stephentoub:
> good point although CI and my local builds passed

Don't we skip this project by default because it requires admin privileges?

weshaggard:
Yes we have disabled this set of tests from CI. I verified my changes breaks those tests and made the suggested fix and the tests pass as well as the warning is now gone.

stephentoub:
LGTM

Issue #2474, opened 2015-07-24
TestProcessStartTime is failing for me on every test run locally


stephentoub:
e.g.
```C#
Discovering: System.Diagnostics.Process.Tests
  Discovered:  System.Diagnostics.Process.Tests
  Starting:    System.Diagnostics.Process.Tests
     System.Diagnostics.ProcessTests.ProcessTests.TestProcessStartTime [FAIL]
        Assert+WrapperXunitException : File path: corefx\src\System.Diagnostics.Process\tests\System.Diagnostics.Process.Tests\Proc
  essTests.cs. Line: 393
        ---- Assert.InRange() Failure
        Range:  (635733440815048892 - 635733440815148892)
        Actual: 635733440815184323
        Stack Trace:
           corefx\src\Common\tests\System\Diagnostics\AssertWithCallerAttributes.cs(603,0): at Assert.WrapException(Exception e, St
  ring callerFilePath, Int32 callerLineNumber)
           corefx\src\Common\tests\System\Diagnostics\AssertWithCallerAttributes.cs(225,0): at Assert.InRange[T](T actual, T low, T
   high, String path, Int32 line)
           corefx\src\System.Diagnostics.Process\tests\System.Diagnostics.Process.Tests\ProcessTests.cs(393,0): at System.Diagnosti
  cs.ProcessTests.ProcessTests.TestProcessStartTime()
           ----- Inner Stack Trace -----
           corefx\src\Common\tests\System\Diagnostics\AssertWithCallerAttributes.cs(224,0): at Assert.InRange[T](T actual, T low, T
   high, String path, Int32 line)
  Finished:    System.Diagnostics.Process.Tests
```

Priya91:
should be fixed.

PR #2513, opened 2015-07-24
BugFix: Attempt fix for process tests failing in CI environment.

Priya91:
cc @stephentoub @pallavit 

for bugs #2479 #2474 #1896

### DIFF
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
@@ -6,6 +6,7 @@ using System.Security;
 using System.IO;
 using Xunit;
 using System.Threading;
+using System.ComponentModel;
 
 namespace System.Diagnostics.ProcessTests
 {
@@ -14,30 +15,17 @@ namespace System.Diagnostics.ProcessTests
         [Fact, PlatformSpecific(~PlatformID.OSX)]
         public void TestModulePropertiesExceptOnOSX()
         {
-            string fileName = CoreRunName;
-            if (global::Interop.IsWindows)
-                fileName = string.Format("{0}.exe", CoreRunName);
-
-            // Ensure the process has loaded the modules.
-            Assert.True(SpinWait.SpinUntil(() =>
-            {
-                if (_process.Modules.Count > 0)
-                    return true;
-                _process.Refresh();
-                return false;
-            }, WaitInMS));
-
-            ProcessModuleCollection moduleCollection = _process.Modules;
+            ProcessModuleCollection moduleCollection = Process.GetCurrentProcess().Modules;
             Assert.True(moduleCollection.Count > 0);
 
-            ProcessModule coreRunModule = _process.MainModule;
-            Assert.True(coreRunModule.BaseAddress.ToInt64() > 0);
-            Assert.True(coreRunModule.EntryPointAddress.ToInt64() >= 0);
-            Assert.Equal(fileName, coreRunModule.ModuleName);
-            Assert.True(coreRunModule.ModuleMemorySize > 0);
-            Assert.EndsWith(fileName, coreRunModule.FileName);
-
-            Assert.Equal(string.Format("System.Diagnostics.ProcessModule ({0})", fileName), coreRunModule.ToString());
+            for (int i = 0; i < moduleCollection.Count; i++)
+            {
+                Assert.True(moduleCollection[i].BaseAddress.ToInt64() > 0);
+                // From MSDN: Due to changes in the way that Windows loads assemblies, 
+                // EntryPointAddress will always return 0 on Windows 8 or Windows 8.1 and should not be relied on for those platforms.
+                Assert.True(moduleCollection[i].EntryPointAddress.ToInt64() >= 0);
+                Assert.True(moduleCollection[i].ModuleMemorySize > 0);
+            }
         }
 
         [Fact, PlatformSpecific(PlatformID.OSX)]
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
 using System.Threading;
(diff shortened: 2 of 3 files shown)


src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
```diff
@@ -18,26 +19,30 @@ public void TestModulePropertiesExceptOnOSX()
...
-            Assert.True(coreRunModule.EntryPointAddress.ToInt64() >= 0);
-            Assert.Equal(fileName, coreRunModule.ModuleName);
-            Assert.True(coreRunModule.ModuleMemorySize > 0);
-            Assert.EndsWith(fileName, coreRunModule.FileName);
+                ProcessModule coreRunModule = _process.MainModule;
+                Assert.True(coreRunModule.BaseAddress.ToInt64() > 0);
+                Assert.True(coreRunModule.EntryPointAddress.ToInt64() >= 0);
+                Assert.Equal(fileName, coreRunModule.ModuleName);
+                Assert.True(coreRunModule.ModuleMemorySize > 0);
+                Assert.EndsWith(fileName, coreRunModule.FileName);
```
stephentoub: Aren't these fileName checks going to suffer from the same problem as in TestMainModule?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
```diff
@@ -18,26 +19,30 @@ public void TestModulePropertiesExceptOnOSX()
...
 
-            Assert.Equal(string.Format("System.Diagnostics.ProcessModule ({0})", fileName), coreRunModule.ToString());
+                Assert.Equal(string.Format("System.Diagnostics.ProcessModule ({0})", fileName), coreRunModule.ToString());
+            }
+            catch (Win32Exception) { } // expected, known behavior of Process.MainModule to throw, when 32 bit process accesses a 64bit PEB on 64 bit machines.
```
stephentoub: So does this issue go away when the test process is 64-bit?  If so, should we assert inside the catch block that IntPtr.Size == 4 so that if we hit this catch block when we're running in a 64-bit process, we still fail?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
```diff
@@ -18,26 +19,30 @@ public void TestModulePropertiesExceptOnOSX()
...
+                ProcessModuleCollection moduleCollection = _process.Modules;
+                Assert.True(moduleCollection.Count > 0);
 
-            ProcessModule coreRunModule = _process.MainModule;
-            Assert.True(coreRunModule.BaseAddress.ToInt64() > 0);
-            Assert.True(coreRunModule.EntryPointAddress.ToInt64() >= 0);
-            Assert.Equal(fileName, coreRunModule.ModuleName);
-            Assert.True(coreRunModule.ModuleMemorySize > 0);
-            Assert.EndsWith(fileName, coreRunModule.FileName);
+                ProcessModule coreRunModule = _process.MainModule;
```
stephentoub: This may not be coreRun, right?

stephentoub: Should we be validating these properties for all modules, not just the first one?

pallavit: > This may not be coreRun, right?

True, depending on the window when we check. However for most cases this will be true. I still need to follow-up with windows on this behavior and so I would suggest keep 1896 open (we did not see a lot of hits for this before and with this iteration we should be behaving much like the old tests). Let me know if you disagree.

> Should we be validating these properties for all modules, not just the first one?

Yes Please  :)

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
pallavit: No. The underlying implementation calls EnumProcessModules (https://msdn.microsoft.com/en-us/library/windows/desktop/ms682631(v=vs.85).aspx) which has the following limitation.

*If this function is called from a 32-bit application running on WOW64, it can only enumerate the modules of a 32-bit process. If the process is a 64-bit process, this function fails and the last error code is ERROR_PARTIAL_COPY (299).*

The problem can be partially solved by using EnumProcessModulesEx (https://msdn.microsoft.com/en-us/library/windows/desktop/ms682633(v=vs.85).aspx) pending AppCompat bar-check however I would like revisit these APIs holistically due to the flakiness and cross-plat issues in general.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
pallavit: It might be best to remove MainModule checks from here (MainModule test method would cover the FileName) and as @stephentoub suggested test the generic properties of all the modules here.

Priya91: Yeah, i did that.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessModuleTests.cs
Priya91: As per the documentation, if the test process is 64 bit then this error should not happen right as @stephentoub suggests.. A 64 bit process can access 64bit TestConsoleApp.exe process modules, as well as, 32 bit TestConsoleApp.exe process modules.. Or better yet, i can test on the CurrentProcess Modules, instead of testing on another process, then this problem wouldn't occur.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -185,35 +186,18 @@ public void TestMachineName()
...
+            string fileName = "CoreRun";
+            if (global::Interop.IsWindows)
+                fileName = string.Format("{0}.exe", fileName);
 
-            // Check that the mainModule is present in the modules list.
-            IEnumerable<string> processModuleNames = _process.Modules.Cast<ProcessModule>().Select(p => Path.GetFileNameWithoutExtension(p.ModuleName));
-            Assert.Contains(CoreRunName, processModuleNames);
+            Process p = Process.GetCurrentProcess();
+            Assert.True(p.Modules.Count > 0);
+            Assert.Equal(fileName, p.MainModule.ModuleName);
```
stephentoub: This is likely to fail on Unix where "corerun" is spelled all lower-case

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -185,35 +186,18 @@ public void TestMachineName()
...
-            Assert.NotNull(mainModule);
-            Assert.Equal(CoreRunName, Path.GetFileNameWithoutExtension(mainModule.ModuleName));
+            string fileName = "CoreRun";
+            if (global::Interop.IsWindows)
+                fileName = string.Format("{0}.exe", fileName);
 
-            // Check that the mainModule is present in the modules list.
-            IEnumerable<string> processModuleNames = _process.Modules.Cast<ProcessModule>().Select(p => Path.GetFileNameWithoutExtension(p.ModuleName));
-            Assert.Contains(CoreRunName, processModuleNames);
+            Process p = Process.GetCurrentProcess();
```
stephentoub: Good idea switching to the current process.

stephentoub:
Thanks for fixing these, Priya.  Other than the casing failure on Linux, LGTM.

Priya91:
build failed with #2539 @dotnet-bot test this please

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
pallavit: @Priya91 , @stephentoub I will be a little concerned about moving to the current process. Process tests tries to be as non-invasive as possible and hence spawns a new process for every property check, I personally think that's better.

Priya91:
build failed with #2544 @dotnet-bot test this please

nguerrera:
@dotnet-bot test this please

nguerrera:
#2539 should be fixed for real now. @dotnet-bot test this please

@Priya91 Is this the fix for what #2540 hit?

Priya91:
@nguerrera : Nope, that is a new failure. Will investigate.

Priya91:
@dotnet-bot test this please

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
stephentoub: @pallavit, I'm not understanding the comment.  Can you elaborate?  ProcessTests.cs already does many different tests and property gets/sets against the same test process; it doesn't spawn a new process for each such access.

stephentoub:
> Nope, that is a new failure. Will investigate.

Priya, what was the problem?

Priya91:
@stephentoub : On Linux, When EnableRaiseEvents is set and a callback function is provided, the callback function was not called on exiting. I looked at this for sometime yesterday, couldn't figure out how that could happen.. One theory is before a thread from threadpool executes the callback, we might have done the check in the current process to verify that the callback was called. But not sure how these work, do you have any ideas ?

stephentoub:
> But not sure how these work, do you have any ideas ?

Please open an issue to track, and I can take a look when I'm back. Thanks.

Issue #2257, opened 2015-07-07
Trivial change to make Queue<T>'s Enqueue / Dequeue twice faster


omariom:
[Generic Queue](https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System/Collections/Generic/Queue.cs) doubles its capacity on each grow. And by default it starts with 4 items in its underlying array.

```C#
private const int MinimumGrow = 4; 
private const int GrowFactor = 200;  // double each time 
private const int DefaultCapacity = 4; 
```
It gives us opportunity to use well known trick - bitwise AND instead of modulus -  in Enqueue, Dequeue and Contains methods. We just need to check if size of the underlying array is power of two.

For example Dequeue could look like this:
```C#
public T Dequeue() 
{ 
    if (_size == 0) 
        throw new InvalidOperationException(SR.InvalidOperation_EmptyQueue); 

    T removed = _array[_head]; 
    _array[_head] = default(T); 
    _head = _isCapacityPowerOfTwo
                  ? (_head + 1) & (_array.Length - 1)  
                  : (_head + 1) % _array.Length; 
    _size--; 
    _version++; 
    return removed; 
} 
```

In microbechmarks  it makes Enqueue and Dequeue twice faster when default ctor is used or initial capacity (set via another ctor)  is power of two. And it doesn't degrade in opposite case because division is much heavier compared to branching on _isCapacityPowerOfTwo field.

The change doesn't make Queue consume more memory.

nguerrera:
Interesting. This would add an extra field to every instance of Queue, which might be worth weighing against the speedup. 

How does `if (++_head == array.Length) { _head = 0; }` perform in the same microbenchmark?

omariom:
@nguerrera 
> This would add an extra field to every instance of Queue, which might be worth weighing against the speedup.

Technically we don't have to use the new field if [in this ctor](https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System/Collections/Generic/Queue.cs#L48) we increase capacity to next power of 2. 

> How does  if (++_head == _array.Length) { _head = 0; }  perform in the same microbenchmark?

I will check that.

stephentoub:
> Technically we don't have to use the new field if in this ctor we increase capacity to next power of 2.

The point of that ctor, though, is for someone to say how much memory they're going to need; if we artificially increase it, that's going to consume even more memory than an extra Boolean field.

> in this ctor

[...]

PR #2514, opened 2015-07-24
Fix to issue #2257 - Trivial change to make Queue<T>'s Enqueue / Dequeue twice faster

omariom:
The issue: https://github.com/dotnet/corefx/issues/2257

This PR replaces usage of reminder operator (which is fairly slow) in Queue<T>'s Enqueue / Dequeue / Contains methods with simple boundary check.


omariom:
It added all the updates to my fork too!

PR #2515, opened 2015-07-24
Fix to issue #2257 - Trivial change to make Queue<T>'s Enqueue / Dequeue twice faster

omariom:
The issue: https://github.com/dotnet/corefx/issues/2257

This PR replaces usage of reminder operator (which is fairly slow) in Queue's Enqueue / Dequeue / Contains methods with simple boundary check.

### DIFF
--- a/src/System.Collections/src/System/Collections/Generic/Queue.cs
+++ b/src/System.Collections/src/System/Collections/Generic/Queue.cs
@@ -213,7 +213,7 @@ namespace System.Collections.Generic
             }
 
             _array[_tail] = item;
-            _tail = (_tail + 1) % _array.Length;
+            MoveNext(ref _tail);
             _size++;
             _version++;
         }
@@ -249,7 +249,7 @@ namespace System.Collections.Generic
 
             T removed = _array[_head];
             _array[_head] = default(T);
-            _head = (_head + 1) % _array.Length;
+            MoveNext(ref _head);
             _size--;
             _version++;
             return removed;
@@ -289,13 +289,13 @@ namespace System.Collections.Generic
                 {
                     return true;
                 }
-                index = (index + 1) % _array.Length;
+                MoveNext(ref index);
             }
 
             return false;
         }
 
-        internal T GetElement(int i)
+        private T GetElement(int i)
         {
             return _array[(_head + i) % _array.Length];
         }
@@ -324,7 +324,6 @@ namespace System.Collections.Generic
             return arr;
         }
 
-
         // PRIVATE Grows or shrinks the buffer to hold capacity objects. Capacity
         // must be >= _size.
         private void SetCapacity(int capacity)
@@ -347,6 +346,15 @@ namespace System.Collections.Generic
             _head = 0;
             _tail = (_size == capacity) ? 0 : _size;
             _version++;
+        }
+
+        // Increments the index wrapping it if necessary.
+        private void MoveNext(ref int index)
+        {
+            // It is tempting to use the remainder operator here but it is actually much slower 
+            // than a simple comparison and a rarely taken branch.   
+            int tmp = index + 1;
+            index = (tmp == _array.Length) ? 0 : tmp;
         }
 
         public void TrimExcess()


src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -204,16 +204,14 @@ public void Enqueue(T item)
         {
             if (_size == _array.Length)
             {
-                int newcapacity = (int)((long)_array.Length * (long)GrowFactor / 100);
-                if (newcapacity < _array.Length + MinimumGrow)
-                {
-                    newcapacity = _array.Length + MinimumGrow;
-                }
-                SetCapacity(newcapacity);
+                Grow();
```
stephentoub: Why was this refactored into a separate method?

src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -324,6 +328,15 @@ public T[] ToArray()
...
 
+        private void Grow()
+        {
+            int newcapacity = (int)((long)_array.Length * (long)GrowFactor / 100);
+            if (newcapacity < _array.Length + MinimumGrow)
+            {
+                newcapacity = _array.Length + MinimumGrow;
+            }
+            SetCapacity(newcapacity);
+        }
```
nguerrera: Factoring this out is unrelated cleanup or is there a diff I'm missing? Generally, it's better to do these separately. I don't mind having this in a helper, just having it in this same diff makes it harder to review.

In terms of helpers, maybe add one for the pattern you now have in 3 places:  `void Increment(ref int index)`?

Also, I'd add a comment about not wanting to do modulus for perf reasons.

src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -249,7 +247,10 @@ public T Dequeue()
 
             T removed = _array[_head];
             _array[_head] = default(T);
-            _head = (_head + 1) % _array.Length;
+            if (++_head == _array.Length)
```
sharwell: :exclamation: In multithreaded code which does not correctly synchronize on usages of `Queue<T>`, this change introduces a point in time where a call to `Peek()` will throw an `ArrayIndexOutOfBoundsException`, where in the previous implementation it would return `default(T)` for the same interleaving. This is not a thread-safe type, but this change in observable behavior should be explicitly confirmed as desired.

src/System.Collections/src/System/Collections/Generic/Queue.cs
omariom: No, it has nothing to do with inlining. It was just instinct :)

src/System.Collections/src/System/Collections/Generic/Queue.cs
omariom: @sharwell 
I think the classes not intended to be used in multithreaded manner don't guaranty anything. An excepton instead of returning a wrong value is not a worst thing that can happen to a  
careless dev :)
It is explicitly stated in "Any instance members are not guaranteed to be thread safe."

src/System.Collections/src/System/Collections/Generic/Queue.cs
omariom: Agree.

src/System.Collections/src/System/Collections/Generic/Queue.cs
stephentoub: Ok. In that case please revert it; if it's beneficial to separate it out, that can be done separately.

src/System.Collections/src/System/Collections/Generic/Queue.cs
sharwell: You could use a local variable to ensure `_head` is only written once by this method. This would preserve the original behavior and is highly unlikely to negatively affect performance.

omariom:
@nguerrera  @stephentoub 

Ok, this is what I've found. 
The original Enqueue does a million ops in not less than 10.5 ms.
The new one can do the same in 6.5 ms.
Using Increment(ref index) though hasn't impacted perf but  increased number of instructions
from
```ASM
lea         eax,[rdx+1]  
mov         dword ptr [rsi+1Ch],eax
cmp         r8d,eax  
jne         00007FF8F4E708E8  
xor         eax,eax  
mov         dword ptr [rsi+1Ch],eax
```
to
```ASM
lea         rax,[rsi+1Ch]  
mov         edx,dword ptr [rax]  
inc         edx  
mov         dword ptr [rax],edx  
mov         rcx,qword ptr [rsi+8]
cmp         dword ptr [rcx+8],edx
jne         00007FF8F4E80920  
xor         edx,edx  
mov         dword ptr [rax],edx  
```
That's probably because JIT is [not yet perfect at inlining methods with ref params pointing to fields](https://github.com/dotnet/coreclr/issues/1226).

But with @sharwell's suggestion, looking like:
```C#
int tail = _tail;
_array[tail] = item;
Increment(ref tail);
_tail = tail;
```
JIT does a perfect job of inlining the call to Increment:
```ASM
inc         eax  
cmp         ecx,eax  
jne         00007FF8F4E7079D  
xor         eax,eax  
mov         dword ptr [rsi+1Ch],eax
```
With the changes enqueuing a million ints takes about 5.5 ms.

The results for Dequeue: 9.3 ms vs 3 ms.

stephentoub:
Thanks, @omariom.  Sounds like adopting both Nick's and Sam's suggestions is the way to go.

sharwell:
@omariom: Keep in mind the same thing (local variable to ensure single-update to the field) can be applied to the location where `_tail` is updated. I didn't mention it before because the intermediate update was not exposed to callers by a pure method like `Peek`.

omariom:
I've made the discussed changes: https://github.com/omariom/corefx/commit/8c22635ac1e38433244b8e7351c7d653906a4db6

What to do now with Grow method? Should I create a separate issue or a PR will be enough?

nguerrera:
I have to say I'm disappointed that the code has gotten repetitive. I'd really like to capture everything in the increment helper if at all possible.

cc @CarolEidt to see if there's something else we can use to get around the sub-optimal inline...

omariom:
I can turn it back to usage of the fields (without copies). It was good enough already. 
It will get better with JIT getting better.

stephentoub:
Thanks, @omariom.  I'd suggest not worrying about the increased number of instructions for now; as you say, that's something that'll just get better as the backend improves.  We'd have the helper like:
```C#
private void MoveNext(ref int value)
{
    int tmp = value + 1;
    value = (tmp == _array.Length) ? 0 : tmp;
}
```
which would be used at call sites like:
```C#
_array[_tail] = item;
MoveNext(ref _tail); // instead of _tail = (_tail + 1) % _array.Length;
_size++;
_version++;
```
and that should provide the bulk of the wins while keeping the call sites simple.

omariom:
@stephentoub 
Ready!
And perf is good.

stephentoub:
> Ready!

Thanks, but what about the local temp in the helper?

Can you also please squash this down to a single commit?

omariom:
Fixed and squashed :ok_hand:

src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -349,6 +348,15 @@ private void SetCapacity(int capacity)
             _version++;
         }
 
+        // Increments the index wrapping it if necessary.
+        private void MoveNext(ref int index)
+        {
+            // It is tempting to use the reminder operator here but it is actually much slower
```
stephentoub: Nit: remainder, not reminder

stephentoub:
Thanks! Perf still looks good?

omariom:
Dequeue is about 3.5 ms, Enqueue 4.5-5 ms. Even better than twice.

stephentoub:
LGTM. Thanks!

src/System.Collections/src/System/Collections/Generic/Queue.cs
```diff
@@ -349,6 +348,15 @@ private void SetCapacity(int capacity)
...
 
+        // Increments the index wrapping it if necessary.
+        private void MoveNext(ref int index)
+        {
+            // It is tempting to use the remainder operator here but it is actually much slower 
+            // than a simple comparison and a rarely taken branch.   
+            int tmp = index + 1;
+            index = (tmp == _array.Length) ? 0 : tmp;
+        }
+
```
paulomorgado: Wouldn't this be better?

```c#
private void MoveNext(ref int index)
{
    index++;
    if (index == _array.Length)
    {
        index = 0;
    }
}
```

stephentoub: > Wouldn't this be better?

Have you tried it?  I would expect "no", since every read/write on `index` needs to go through that `ref` (hence the use of the `tmp` here), but if you find otherwise and have data to back it up, PRs are welcome. :smile:

paulomorgado: I was just wondering. But the `ref` access does make a difference.

fandecheng: Would the C# compiler automatically inline the MoveNext method in the release build? If so, then ref won't be slower than a local variable. Just wondering...

Issue #2516, opened 2015-07-24
FileSystem tests sometimes see wrong System.Runtime.Extensions and fail due to invalid path chars


nguerrera:
Locally, I'm seeing failures like the one logged below.

The issue appears to be that the hard-coded TestWithLocalLibraries=true has a build order dependency: if the test builds before System.Runtime.Extensions, it deploys the one from nuget that doesn't like the extended path syntax.

@weshaggard is working on a fix, cc @JeremyKuhne as FYI.

```
[0]: System.ArgumentException: Illegal characters in path.
Parameter name: path
   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.GetFullPath(String path)
   at System.IO.PathHelpers.GetFullPathInternal(String path) in C:\src\corefx\src\System.IO.FileSystem\src\System\IO\PathHelpers.Windows.cs:line 46
   at System.IO.DirectoryInfo..ctor(String path) in C:\src\corefx\src\System.IO.FileSystem\src\System\IO\DirectoryInfo.cs:line 21
   at System.IO.FileSystem.Tests.DirectoryInfo_Create.Create(String path) in C:\src\corefx\src\System.IO.FileSystem\tests\DirectoryInfo\Create.cs:line 14
   at System.IO.FileSystem.Tests.Directory_CreateDirectory.<>c__DisplayClass11_0.<ValidExtendedPathWithTrailingSlash>b__0(String component) in C:\src\corefx\src\System.IO.FileSystem\tests\Directory\CreateDirectory.cs:line 128
   at Xunit.Assert.All[T](IEnumerable`1 collection, Action`1 action)
```

JeremyKuhne:
Sorry about that- thanks for catching the issue.

nguerrera:
Fixed by #2518

PR #2518, opened 2015-07-24
Add P2P reference for new IO tests to System.Runtime.Extensions

weshaggard:
Until we can publish a new packages with the new System.Runtime.Extensions
implementation reference the locally built version. We need to change to a
P2P reference to avoid the race condition that is inherit in using
TestWithLocalLibraries. That can only be used if we know all the
dependencies are built first.

This fixes issue #2516, which I was also hitting.

### DIFF
--- a/src/System.IO.FileSystem/tests/System.IO.FileSystem.Tests.csproj
+++ b/src/System.IO.FileSystem/tests/System.IO.FileSystem.Tests.csproj
@@ -8,8 +8,6 @@
     <OutputType>Library</OutputType>
     <AssemblyName>System.IO.FileSystem.Tests</AssemblyName>
     <TestCategories>InnerLoop;OuterLoop</TestCategories>
-    <!-- Setting temporarily until we can publish a new package drop of Runtime.Extensions -->
-    <TestWithLocalLibraries>true</TestWithLocalLibraries>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


weshaggard:
cc @JeremyKuhne

src/System.IO.FileSystem/tests/System.IO.FileSystem.Tests.csproj
```diff
@@ -30,6 +28,17 @@
       <Name>System.IO.FileSystem</Name>
       <Private>True</Private>
     </ProjectReference>
+
+    <!-- 
+      Until an updated packages is published, temporarily copy the locally built System.Runtime.Extensions library
```
stephentoub: If the plan is to revert this when an updated package is available, I'd suggest including the issue number here for reference.

stephentoub:
One nit, otherwise LGTM.

weshaggard:
@stephentoub it's like I'm talking to myself :) as that is the kind of comment I usually make. I did think about that but I realized we have this in a number of other places and I was being lazy and want to get this fix out. However I will update this and add an issue to try and set the right precedence for other P2P references I don't like.

stephentoub:
:smile:

JeremyKuhne:
Thanks- LGTM. Couldn't get P2P to work- see what I was apparently missing now.

PR #2529, opened 2015-07-25
Add Linux ARM target for Native libraries.

benpye:
Simply adds the ARM target required for the native components. The compile options mirror those of coreclr, only thing I am unsure of is if the MSBuild target should be upper or lower case?


stephentoub:
Thanks, @benpye. cc: @nguerrera

stephentoub:
LGTM

nguerrera:
LGTM too. Thanks.

PR #2533, opened 2015-07-25
Adds OSPlatform.FreeBSD.

jasonwilliams200OK:
In #2068, it was decided to hold off adding public property for FreeBSD
because at the time there wasn't much progress made for the platform
support. Since then, there has been some progress made for FreeBSD.
Also with #2141 merged, I have added a missing test for S.RT.I.RI.

### DIFF
--- a/src/System.Runtime.InteropServices.RuntimeInformation/src/RuntimeInformation.FreeBSD.cs
+++ b/src/System.Runtime.InteropServices.RuntimeInformation/src/RuntimeInformation.FreeBSD.cs
@@ -1,15 +1,13 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 namespace System.Runtime.InteropServices
 {
     public static class RuntimeInformation
     {
-        private static OSPlatform s_freeBSD = OSPlatform.Create("FREEBSD");
-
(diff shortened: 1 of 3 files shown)


src/System.Runtime.InteropServices.RuntimeInformation/src/RuntimeInformation.FreeBSD.cs
```diff
@@ -1,15 +1,13 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
```
jasonwilliams200OK: I have added :bomb:, as this was the only `.cs` file in the folder missing BOM.

stephentoub:
Thanks, @jasonwilliams200OK. It's great that some progress has been made, but I don't think it's far enough along yet, and progress has slowed of late. I think we need to establish some tripwire for when we accept adding such a property, e.g. 90% of corefx tests passing on the platform, or something like that. @weshaggard? @ellismg? @joshfree?

ellismg:
I would be interested in understanding what percent of the tests that pass on Linux/OSX today are passing on FreeBSD.  I think for now we should be very close/at parity with this pass rate (which we already know is worse than the Windows pass rate, due to tests not running) before merging something like this.

stephentoub:
I'm going to close this for now.  We can revisit once the FreeBSD port is closer to being completed.  Thanks.

PR #2536, opened 2015-07-26
Fix System.IO.Native shim

janhenke:
The type of src is a const reference, not a const pointer. So use direct
member access instead of pointer access. This changes allows to build
the shim on FreeBSD. Tested on both FreeBSD and Linux.


stephentoub:
LGTM. Thanks for the fix.

cc: @nguerrera

nguerrera:
Oops. Thanks. :)

PR #2496, opened 2015-07-24
Fix serialization test failure in Minsk timezone

khdang:
Fix #2287

PR #2541, opened 2015-07-27
Serialization time zone fixes

mattjohnsonpint:
Similar to #2496, several other tests will fail when running in southern-hemisphere time zones that use daylight saving time.   The tests pick a date in January, but always use the standard offset.  Since southern hemisphere time zones are in DST during this period, the standard offset is not the correct one to use.

Calling `TimeZoneInfo.BaseUtcOffset` should always be avoided.  Use `TimeZoneInfo.GetUtcOffset` instead.

Additionally, unit tests that compare `DateTimeOffset` values should always test the value and the offset separately, due to the way that the equality comparison normalizes to UTC.   This was masking errors in the System.Runtime.Serialization.Xml tests.

### DIFF
--- a/src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
+++ b/src/System.Runtime.Serialization.Json/tests/DataContractJsonSerializer.cs
@@ -89,7 +89,7 @@ public static class DataContractJsonSerializerTests
     [Fact]
     public static void DCJS_DateTimeAsRoot()
     {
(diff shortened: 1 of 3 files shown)


stephentoub:
Thanks, Matt. Some of the var usage would ideally be cleaned up, but as that's all preexisting your change, LGTM.

cc: @khdang

mellinoe:
@dotnet-bot test this, please

nguerrera:
@dotnet-bot test this please

nguerrera:
@dotnet-bot test this please

Issue #2546, opened 2015-07-28
Many deprecation warnings in System.Security.Cryptography.Native on Mac OS X


nguerrera:
I'm about to unblock the mac build by disabling the warnings, but building System.Security.Crytpography.Native on Mac OS X produces many deprecation warnings.

bartonjs:
Really the problem is that for OSX we should be using CommonCrypto instead of OpenSSL, and that's what these deprecation warnings are trying to tell us.  This is something that we definitely need to do long-term, but we have to start somewhere.  Moving to "Future".

jasonwilliams200OK:
@bartonjs, found some references for similar CommonCrypto support work: https://github.com/search?q=openssl+commoncrypto+is%3Apr&ref=searchresults&type=Issues&utf8=✓.
Seems like mostly they have prefixed the function names with `CC_`, for which `#define`s and `static_cast`s would be required. Hopefully, there won't be a major blocker for S.S.C.N coverage.

bartonjs:
When we stopped building against 0.9.8 the warnings went away.  So this is pretty stale now :smile:

PR #2547, opened 2015-07-28
Fix Mac OS X build break due to deprecation warnings as errors

nguerrera:
Fixed the issue for System.IO.Native and disabled the warning and filed #2546 for System.Security.Cryptography.Native

cc @sokket @bartonjs

### DIFF
--- a/src/Native/System.IO.Native/nativeio.cpp
+++ b/src/Native/System.IO.Native/nativeio.cpp
@@ -7,7 +7,7 @@
 #include "nativeio.h"
 #include <sys/stat.h>
 
-#if HAVE_STAT64 && !(defined(__APPLE__) && defined(_AMD64_))
+#if HAVE_STAT64
 #    define stat_ stat64
 #    define fstat_ fstat64
 #else
--- a/src/Native/System.Security.Cryptography.Native/CMakeLists.txt
+++ b/src/Native/System.Security.Cryptography.Native/CMakeLists.txt
@@ -3,6 +3,9 @@ project(System.Security.Cryptography.Native)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+# Issue 2546 - Many deprecation warnings in System.Security.Cryptography.Native on Mac OS X
+add_compile_options(-Wno-deprecated-declarations)
+
 add_definitions(-DPIC=1)
 
 find_library(CRYPTO NAMES crypto)
--- a/src/Native/config.h.in
+++ b/src/Native/config.h.in
@@ -2,3 +2,12 @@
 
 #cmakedefine01 HAVE_STAT64
 #cmakedefine01 HAVE_STAT_BIRTHTIME
+
+// Mac OS X has stat64, but it is deprecated since plain stat now
+// provides the same 64-bit aware struct when targeting OS X > 10.5
+// and not passing _DARWIN_NO_64_BIT_INODE.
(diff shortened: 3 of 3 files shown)


nguerrera:
cc @stephentoub

sokket:
LGTM, it doesn't look like your change should cause the runtime download failure. Thanks for fixing this so quickly! :)

kangaroo:
This is wrong.  stat64 is deprecated, because _DARWIN_USE_64_BIT_INODE was introduced.  If we want 64-bit file support on x86_64 mscorlib.dll, we need to use struct stat, with _DARWIN_USE_64_BIT_INODE so we can get the correct data.

nguerrera:
@kangaroo Thanks for bringing this to my attention.

The table in the man page (quoted below) leaves me with the impression that the default configuration is basically equivalent to defining _DARWIN_USE_64_BIT_INODE on Mac OS >10.5 and I think it's safe to plan for the minimum requirement for CoreCLR to be > 10.5. 

```
      _DARWIN_FEATURE_ONLY_64_BIT_INODE not defined
-------------------------+-------------------------------
                         |      Deployment Target
     user defines:       |   < 10.5       10.5    > 10.5
-------------------------+-------------------------------
(none)                   |   32-bit      32-bit   64-bit
_DARWIN_NO_64_BIT_INODE  |   32-bit      32-bit   32-bit
_DARWIN_USE_64_BIT_INODE |   32-bit      64-bit   64-bit
-------------------------+-------------------------------
```

Am I reading this right? If so, I'm inclined to just add some clarifying remarks to the comment about Mac OS X in configure.h.in.

stephentoub:
@dotnet-bot test this please

nguerrera:
@dotnet-bot test this please

src/Native/System.IO.Native/nativeio.cpp
```diff
@@ -7,7 +7,7 @@
 #include "nativeio.h"
 #include <sys/stat.h>
```
sokket: You need the deprecated compile option for this function as well

nguerrera: I am not seeing that. Are you using my full change including the change to config.h.in?

My intent was not to use (f)stat64 on Mac OS X, which would eliminate the warning and any need to suppress it.

nguerrera:
@kangaroo I've done a quick validation that using plain stat with no special defines does give the larger stat structure on my 10.10 machine:

```
#include <stdio.h>
#include <sys/stat.h>
int main() { printf("%lu\n%lu\n", sizeof(struct stat), sizeof(struct stat64)); }
```

I can only get the smaller stat if I explicitly pass _DARWIN_NO_64_BIT_INODE, which is consistent with my interpretation of the man page above.

```
$ clang -D _DARWIN_NO_64_BIT_INODE test.c && ./a.out
120
144

$ clang -D _DARWIN_USE_64_BIT_INODE test.c && ./a.out
144
144

$ clang test.c && ./a.out
144
144
```

src/Native/System.IO.Native/nativeio.cpp
sokket: Ah sorry, I forgot to apply the config.h.in change as well

nguerrera:
I've added some clarification to the comment.

kangaroo:
LGTM with the clarified comment.

PR #2551, opened 2015-07-28
BugFix: Signature Decoder not providing ValueType or Class flag

### LINKED ISSUE #2314: Signature Decoder not providing ValueType or Class value for Types.
On the metadata reader, the signature decoder is not providing the valuetype or class value encoded on the signature blob.

SantiFdezM:
This change would fix #2314.

I just added to flags into the provider interface so that the api consumers can retrieve the information in case they need it. In order to do these they need to set the flag  ```bool NeedProjectedFlags``` to ```true``` . In this case they could know if it is a valuetype or class through the ```SignatureTypeCode TypeCode``` property added in ```ISignatureTypeProvider<TType>``` as well.

Adding @AlexGhiondea

### DIFF
--- a/src/System.Reflection.Metadata/src/System/Reflection/Metadata/Decoding/CustomAttributeDecoder.cs
+++ b/src/System.Reflection.Metadata/src/System/Reflection/Metadata/Decoding/CustomAttributeDecoder.cs
@@ -356,11 +356,11 @@ namespace System.Reflection.Metadata.Decoding
             {
                 case HandleKind.TypeDefinition:
                     TypeDefinitionHandle typeDefHandle = (TypeDefinitionHandle)typeRefOrDefHandle;
-                    return provider.GetTypeFromDefinition(typeDefHandle);
+                    return provider.GetTypeFromDefinition(typeDefHandle, null);
 
                 case HandleKind.TypeReference:
                     TypeReferenceHandle typeRefHandle = (TypeReferenceHandle)typeRefOrDefHandle;
-                    return provider.GetTypeFromReference(typeRefHandle);
+                    return provider.GetTypeFromReference(typeRefHandle, null);
 
                 default:
                     throw new BadImageFormatException();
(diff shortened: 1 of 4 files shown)


nguerrera:
cc @tmat I don't think we should communicate with the provider by mutating it. I was thinking something more along the lines of passing s struct to GetTypeFromReference that has a method for IsValueType that only does work (projection) if called. No need for NeedsProjectionInfo -- if you ask for IsValueType, we give you the correct answer always.

nguerrera:
As far as the test failure, looks like dev/metadata is referencing older packages that got deleted. A merge from master should clear that up. I'll do it soon.

tmat:
I think having a property that indicates whether or not to pass the information is ok. Then we can just have ```GetTypeFromReference(handle, bool? isValueType);```

SantiFdezM:
@nguerrera  but I need to know either if it is class or valuetype in both GetTypeFromReference and GetTypeFromDefinition because sometimes they are called just from solving types signatures that don't actually need the prefix and this information need's to be given by the Decoder everytime they call the Provider's GetTypeFromDefinition and GetTypeFromReference.

Because this flag on msil is not always exposed I should be able to know when to expose it or not by the Signature Decoder.

I would add a bool? isClass to what @tmat is proposing.

```GetTypeFromReference(handle, bool? isValueType, bool? isClass)```
```GetTypeFromDefinition(handle, bool? isValueType, bool? isClass)```

because there are cases where the flag doesn't need to be exposed, they both could be false.

nguerrera:
@tmat OK.

@SantiFdezM  Why two bool?'s There are just 3 states (not 6): class, value type, I didn't check because you don't care. So one bool? is enough.

SantiFdezM:
@nguerrera Ok, now I get your point. Sounds fair.

SantiFdezM:
Updated with commit 18815ddfe4d12b1528cd20147ba6a42806af5ad6

cc @tmat

stephentoub:
@dotnet-bot test this please

nguerrera:
This seems like an OK place to stop in the dev branch to get the ildasm work unblocked.

Follow-ups:

* Please squash the commits (since it reverts back to zero and starts again)

* File a bug that:
 a. isValueType is not set correctly when a type is projected from value type to class or vice versa.
 b. There should be an opt-in option to pas non-null value type. By default, it should always be null and when option is on, it should compute isValueType (including possible projection case)

* You'll also need to wait for #2585 to go through and then rebase on top of that.

LGTM otherwise, but give me a chance to sign off once the above are done.

nguerrera:
LGTM with #2586 to be done as follow up and pending green CI.

PR #2552, opened 2015-07-28
Add FreeBSD property to OSPlatform

khellang:
The `s_freebsd` field was never used, this exposes it through a `FreeBSD` property on `OSPlatform`.


khellang:
Oops. Didn't see #2533

PR #2558, opened 2015-07-29
Modified calls to GetTestFilePath within loops to not use the same filename

### LINKED ISSUE #1434: [FileSystem] ValidFileOptions test failure
[from CI](http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/849/console)

	Discovering: System.IO.FileSystem.Tests
	  Discovered:  System.IO.FileSystem.Tests
	  Starting:    System.IO.FileSystem.Tests
	  <Laks_jvhht23u>
	  .
		 System.IO.FileSystem.Tests.FileStream_ctor_str_fm_fa_fs_buffer_fo.ValidFileOptions [FAIL]
			System.UnauthorizedAccessException : Access to the path 'd:\j\workspace\dotnet_corefx_windows_release_prtest\bin\tests\Windows_NT.AnyCPU.Release\System.IO.FileSystem.Tests\dnxcore50\FileStream_ctor_str_fm_fa_fs_buffer_fo\ValidFileOptions_33' is denied.
			Stack Trace:
			   d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.IO.FileSystem\src\System\IO\Win32FileStream.cs(149,0): at System.IO.Win32FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs)
			   d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.IO.FileSystem\src\System\IO\Win32FileSystem.cs(433,0): at System.IO.Win32FileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
			   d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.IO.FileSystem\src\System\IO\FileStream.cs(101,0): at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
			   d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.IO.FileSystem\tests\FileStream\ctor_str_fm_fa_fs_buffer_fo.cs(19,0): at System.IO.FileSystem.Tests.FileStream_ctor_str_fm_fa_fs_buffer_fo.CreateFileStream(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
			   d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.IO.FileSystem\tests\FileStream\ctor_str_fm_fa_fs_buffer_fo.cs(33,0): at System.IO.FileSystem.Tests.FileStream_ctor_str_fm_fa_fs_buffer_fo.ValidFileOptions()
	  Finished:    System.IO.FileSystem.Tests
	  
		 System.IO.FileSystem.Tests  Total: 560, Errors: 0, Failed: 1, Skipped: 0, Time: 1.910s

ianhays:
- Added an overload to GetTestFilePath that takes an integer index parameter to add to the generated filename so that calls to the function within loops will be able to differentiate themselves
- Resolves #1434

### DIFF
--- a/src/System.IO.FileSystem/tests/FileSystemTest.cs
+++ b/src/System.IO.FileSystem/tests/FileSystemTest.cs
@@ -30,15 +30,22 @@ namespace System.IO.FileSystem.Tests
         }
 
         // Generates a test file path to use that is unique name per test case / call
-        public string GetTestFilePath([CallerMemberName]string fileName = null, [CallerLineNumber] int lineNumber = 0)
+        public string GetTestFilePath([CallerMemberName]string memberName = null, [CallerLineNumber] int lineNumber = 0)
         {
-            return Path.Combine(TestDirectory, String.Format("{0}_{1}", fileName ?? "testFile", lineNumber));
+            return Path.Combine(TestDirectory, String.Format("{0}_{1}", memberName ?? "testFile", lineNumber));
+        }
+
+        // Generates a test file path to use that is unique name per test case / call with an additional
+        // variable to specify a trailing identifier
+        public string GetTestFilePath(int index, [CallerMemberName]string memberName = null, [CallerLineNumber] int lineNumber = 0)
+        {
+            return Path.Combine(TestDirectory, String.Format("{0}_{1}_{2}", memberName ?? "testFile", lineNumber, index));
         }
 
         // Generates a test file name to use that is unique name per test case / call
-        public string GetTestFileName([CallerMemberName]string fileName = null, [CallerLineNumber] int lineNumber = 0)
+        public string GetTestFileName([CallerMemberName]string memberName = null, [CallerLineNumber] int lineNumber = 0)
         {
-            return String.Format("{0}_{1}", fileName ?? "testFile", lineNumber);
+            return String.Format("{0}_{1}", memberName ?? "testFile", lineNumber);
         }
 
         public void Dispose()
--- a/src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs.cs
+++ b/src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs.cs
@@ -61,12 +61,13 @@ namespace System.IO.FileSystem.Tests
         public void FileShareCreate()
         {
             // just check that the inputs are accepted, actual sharing varies by platform so we seperate the behavior testing
+            int i = 0;
             foreach (FileShare share in shares)
             {
-                using (FileStream fs = CreateFileStream(GetTestFilePath(), FileMode.Create, FileAccess.ReadWrite, share))
+                using (FileStream fs = CreateFileStream(GetTestFilePath(i++), FileMode.Create, FileAccess.ReadWrite, share))
                 { }
 
-                using (FileStream fs = CreateFileStream(GetTestFilePath(), FileMode.Create, FileAccess.Write, share))
+                using (FileStream fs = CreateFileStream(GetTestFilePath(i++), FileMode.Create, FileAccess.Write, share))
                 { }
             }
         }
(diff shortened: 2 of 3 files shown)


src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa.cs
```diff
@@ -30,9 +30,10 @@ public void InvalidAccessThrows()
         public void InvalidFileModeAccessReadThrows()
         {
             FileMode[] invalidModes = { FileMode.Append, FileMode.Create, FileMode.CreateNew, FileMode.Truncate };
+            int i = 0;
             foreach (FileMode invalidMode in invalidModes)
             {
-                Assert.Throws<ArgumentException>(() => CreateFileStream(GetTestFilePath(), invalidMode, FileAccess.Read));
+                Assert.Throws<ArgumentException>(() => CreateFileStream(GetTestFilePath(i++), invalidMode, FileAccess.Read));
```
stephentoub: Why does this one need to change? Since its an invalid argument, doesn't the file not actually get created?

stephentoub:
Rather than changing all these call sites, did you consider just having a long field in the base class that's incremented on each use and used in the file name (in addition to rather than instead of the caller info)?

src/System.IO.FileSystem/tests/FileSystemTest.cs
```diff
@@ -35,6 +35,13 @@ public string GetTestFilePath([CallerMemberName]string fileName = null, [CallerL
             return Path.Combine(TestDirectory, String.Format("{0}_{1}", fileName ?? "testFile", lineNumber));
         }
 
+        // Generates a test file path to use that is unique name per test case / call with an additional
+        // variable to specify a trailing identifier
+        public string GetTestFilePath(int index, [CallerMemberName]string fileName = null, [CallerLineNumber] int lineNumber = 0)
```
stephentoub: I just noticed that the parameter name is fileName on all these rather that memberName. It should probably be the latter. At some point we should also consolidate such base classes we have in a few different test projects and put it in common.

ianhays:
>Rather than changing all these call sites, did you consider just having a long field in the base class that's incremented on each use and used in the file name (in addition to rather than instead of the caller info)?

Yes, I had considered something similar to that, but I decided against implementing it that way since I don't know how stress testing is going to be done yet. There was some talk at one point about running the unit tests many times, and if a new test class isn't created each time then the index variable has the potential to grow very large, possibly overstepping MaxPath at some point. 

I also didn't want to do anything that altered the existing tests, as some of them are close to MAXPATH already because of subdirectories.

Since there were only a few places where GetTestFilePath was called within a loop anyways, this solution seemed the best.

src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa.cs
ianhays: I read it thinking it was an Assert.All and I didn't want one failed iteration to successfully create the file and cause subsequent failures for the wrong reason. 

Since it is a foreach, you are correct that it isn't a necessary change. I'll roll it back.

src/System.IO.FileSystem/tests/FileSystemTest.cs
ianhays: Yes, it should be memberName; I'll include the name change.

Agreed. The FileSystemTest class isn't specific to FileSystem, despite the name.

ianhays:
@dotnet-bot test this please

stephentoub:
LGTM

Issue #2500, opened 2015-07-24
Event logging for System.Net.*


pgavlin:
Event logging for these assemblies is currently unimplemented: we need to design and implement a solution.

davidsh:
cc: @CIPop @SidharthNabar

stephentoub:
Any reason not to just use an EventSource?

davidsh:
That is our plan.  We just haven't done it yet.  We've been getting some of it to work in System.Net.Sockets and working out issues with it with other .NET folks like @vancem

stephentoub:
Great.

omariom:
Is EventSource going to be supported on other platforms? 
For me it is a synonym for ETW.

stephentoub:
> Is EventSource going to be supported on other platforms?

Yes. EventSource has two mechanisms: the ability to target registered EventListeners, and the ability to target a platform-specific tracing scheme. The former is already working across Windows and Unix. The latter targets ETW on Windows and is currently a nop on UNIX, but the plan is to target something. @brianrob can provide more info.

omariom:
Great! No need to abstract tracing  as I planned.

davidfowl:
One concern I have (and I don't know enough about the EventSource impl) is the amount of "stuff" that gets used when using a low level APIs like sockets. It's great to have tracing at such a low level in the stack (like we do today in .NET Framework), but today creating a socket allocates a boatload of strings because -> tracing -> configuration -> xml -> rabbit hole. I'd like to see us not repeat the same mistakes. We had some socket code on the startup path in dnx and it costed us 250 ms (testing on my Macbook) of startup time. We worked around it like so https://github.com/aspnet/dnx/blob/dev/src/dnx.clr.managed/DomainManager.cs#L62

:smile:

vancem:
Two Points:


  1.  EventSource is NOT just a conduit to ETW.    While it was designed to let you log to ETW trivially, it really couples EventSources to EventListeners (there is just a built in EventListener for ETW).   Thus it is a cross-platform concept (we will support it on Linux).
  2.  EventSource can be VERY lean when it is not operating.    The check can be literally 2 instructions (compare and branch)  There are no config files to look up (unless you make listeners that do this).


That said, you can always screw up perf.   However EventSource really is designed for things like logging of the Socket library etc.   You should be using it.


If you have any questions, I can probably answer them...


Vance



________________________________

brianrob:
[...]

PR #2559, opened 2015-07-29
Initial commit of System.Net.Primitives.

pgavlin:


pgavlin:
@davidsh @CIPop @SidharthNabar @stephentoub PTAL

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+                if (value < _maxCookiesPerDomain)
+                {
+                    _maxCookiesPerDomain = value;
+                    AgeCookies(null);
+                }
+                _maxCookiesPerDomain = value;
+            }
+        }
+
+        // This method will construct a faked URI: tje Domain property is required for param.
```
CIPop: @pgavlin Spelling: _the_ Domain

src/System.Net.Primitives/src/System/Net/HttpStatusCode.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    /// <devdoc>
```
CIPop: @pallavit Nit: we could add a reference to the RFC and leave out <devdoc>

src/System.Net.Primitives/src/System/Net/IPAddress.cs
```diff
@@ -0,0 +1,535 @@
...
+                for (int i = 0; i < NumberOfLabels; i++)
+                {
+                    if (comparand._numbers[i] != _numbers[i])
+                    {
+                        return false;
+                    }
+                }
+
+                // The scope IDs must also match
+                return !compareScopeId || comparand._scopeId == _scopeId;
```
CIPop: @pgavlin Nit: (I'm very picky with this one): in the original code the first comparison was made to comparand._scopeId. To preserve the original operation order we should maybe change to `comparand._scopeId == _scopeId || !compareScopeId` ?

src/System.Net.Primitives/src/System/Net/Sockets/SocketError.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Net;
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net.Sockets
+{
+    public enum SocketError : int
+    {
+        Success = 0,
```
CIPop: @pgavlin I would prefer if the error codes are added in the winsock interop and this implementation references them. It makes a big difference in debugging issues. The error codes were present in the original file as comments such as:

`        Interrupted = (10000 + 4),      //WSAEINTR`

We could either keep the comments in or have something like:
`        Interrupted = Interop.winsock.WSAEINTR;`

Unfortunately, given that the codes don't match the NIX errno.h ones we'll need an adapter for xplat.

src/System.Net.Primitives/tests/FunctionalTests/project.lock.json
```diff
@@ -0,0 +1,1542 @@
...
+          "ref/dotnet/System.Collections.NonGeneric.dll": {}
+        },
+        "runtime": {
+          "lib/dotnet/System.Collections.NonGeneric.dll": {}
+        }
+      },
+      "System.ComponentModel.EventBasedAsync/4.0.10-beta-23127": {
+        "dependencies": {
+          "System.Runtime": "4.0.20-beta-23127",
+          "System.Threading": "4.0.10-beta-23127",
```
CIPop: @pgavlin Given that the project.json changed to use System.Threading 4.0.0, I would expect that the lock file got updated too. Any idea why this is still 4.0.10?

src/System.Net.Primitives/tests/FunctionalTests/project.lock.json
```diff
@@ -0,0 +1,1542 @@
...
+          "System.Resources.ResourceManager": "4.0.0-beta-23127",
+          "System.Threading.Tasks": "4.0.10-beta-23127"
```
CIPop: Same as above for Threading.Tasks.

src/System.Net.Primitives/tests/FunctionalTests/project.lock.json
davidsh: Other libraries may already have dependencies on System.Threading 4.0.10.0.  So, resolving all transitive dependencies results in a final dependency of 4.0.10.0 and not 4.0.0.0.

src/System.Net.Primitives/tests/FunctionalTests/project.lock.json
davidsh: @CIPop See my comment above.

CIPop:
@pgavlin Other than above (mostly nitpick) comments,
:+1: LGTM

davidsh:
LGTM

src/System.Net.Primitives/src/project.json
```diff
@@ -0,0 +1,27 @@
+{
+  "dependencies": {
+    "System.Collections": "4.0.0-beta-*",
+    "System.Diagnostics.Contracts": "4.0.0-beta-*",
+    "System.Diagnostics.Tracing": "4.0.0-beta-*",
```
CIPop: @pgavlin When I try to compile this I'm getting

Restoring all packages...
S:\c2\build.proj(24,5): error : Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
S:\c2\build.proj(24,5): error : Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
S:\c2\build.proj(24,5): error : Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
  Errors in S:\c2\src\System.Net.Primitives\src\project.json
      Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
  Errors in S:\c2\src\System.Net.Primitives\tests\FunctionalTests\project.json
      Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
  Errors in S:\c2\src\System.Net.Primitives\tests\UnitTests\project.json
      Unable to locate System.Diagnostics.Tracing 4.0.0-beta-23127
S:\c2\build.proj(24,5): error MSB3073: The command ""S:\c2\packages//dnx-coreclr-win-x86.1.0.0-beta5-12101/\bin\dnu.cmd" restore
--packages "S:\c2\packages" --source https://www.myget.org/F/dotnet-core/ --source https://www.myget.org/F/dotnet-coreclr/ --sour
ce https://www.myget.org/F/dotnet-corefx/ --source https://www.myget.org/F/dotnet-corefxtestdata/ --source https://www.myget.org/
F/dotnet-buildtools/ --source https://www.nuget.org/api/v2/ "S:\c2\src" "S:\c2\packages/Microsoft.DotNet.BuildTools.1.0.25-prerel
ease-00064/lib/"" exited with code 1.

src/System.Net.Primitives/src/System/Net/Sockets/SocketError.cs
pgavlin: I'll file an issue for this.

src/Common/src/System/Net/Logging/EventSourceLogging.cs
```diff
@@ -0,0 +1,77 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.Tracing;
+
+namespace System.Net
+{
+    [EventSource(Name = "Microsoft-System-Net-Debug")]
```
stephentoub: It's interesting that this file is in Common, as that implies this will be included in multiple assemblies.  If EventSourceLogging.cs is indeed included in multiple assemblies, doesn't that mean that multiple event sources with the same ID could end up being loaded into the same process?

src/Common/src/System/Net/Logging/GlobalLog.cs
```diff
@@ -0,0 +1,287 @@
...
+
+            if (CurrentThreadKind != threadKind)
+            {
+                Print("Thread becomes:(" + CurrentThreadKind.ToString() + ")");
+            }
+
+            return new ThreadKindFrame();
+        }
+
+        private class ThreadKindFrame : IDisposable
```
stephentoub: I'm not sure how this is actually used, as it appears to be here in support of other libraries rather than System.Net.Primitives, but should this be made internal, made a struct, and then SetThreadKind's return type could be changed to ThreadKindFrame rather than IDisposable? (The value SetThreadKind returns would also need to be changed slightly, since ThreadKindFrame couldn't define such a parameterless ctor.)  That would avoid the frame object allocation on each usage of SetThreadKind, though I don't know if/how important that is, since I'm not sure what the use looks like.

src/Common/src/System/Net/Logging/GlobalLog.cs
```diff
@@ -0,0 +1,287 @@
...
+        {
+            private int _frameNumber;
```
stephentoub: readonly?

src/Common/src/System/Net/Logging/GlobalLog.cs
```diff
@@ -0,0 +1,287 @@
...
+
+        public static void Assert(string message, string detailMessage)
+        {
+            try
+            {
+                EventSourceLogging.Log.AssertFailed(message, detailMessage);
+            }
+            finally
+            {
+                Debug.Assert(false, message, detailMessage);
```
stephentoub: Debug.Fail(message, detailedMessage)

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
```diff
@@ -0,0 +1,283 @@
...
+        {
+            Buffer[2] = (byte)(port >> 8);
+            Buffer[3] = (byte)port;
+        }
+
+        internal IPAddress GetIPAddress()
+        {
+            if (Family == AddressFamily.InterNetworkV6)
+            {
+                Contract.Assert(Size >= IPv6AddressSize);
```
stephentoub: Debug.Assert

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
```diff
@@ -0,0 +1,283 @@
...
+                long scope = (long)((Buffer[27] << 24) +
+                                    (Buffer[26] << 16) +
+                                    (Buffer[25] << 8) +
+                                    (Buffer[24]));
+
+                return new IPAddress(address, scope);
+            }
+            else if (Family == AddressFamily.InterNetwork)
+            {
+                Contract.Assert(Size >= IPv4AddressSize);
```
stephentoub: Debug.Assert

src/Common/src/System/Net/Logging/GlobalLog.cs
```diff
@@ -0,0 +1,287 @@
...
+                return;
+            }
+            if ((length < 0) || (length > buffer.Length - offset))
+            {
+                EventSourceLogging.Log.WarningDumpArray("length out of range");
+                return;
+            }
+            ArraySegment<byte> bufferSegment = new ArraySegment<byte>(buffer, offset, length);
+            byte[] bufferSegmentArray = bufferSegment.Array;
+            EventSourceLogging.Log.DebugDumpArray(bufferSegmentArray);
```
stephentoub: This sequence is odd.  We're constructing an ```ArraySegment<byte>``` from a buffer, and then immediately accessing the buffer from that segment, and passing that buffer as the argument... the ArraySegment isn't used at all.  Maybe the author of these lines thought that ArraySegment would create a sub-array based on the offset and length and return that new instance from the Array property (it doesn't)?

src/Common/src/System/Net/Logging/Logging.cs
```diff
@@ -0,0 +1,724 @@
...
+using System.Globalization;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Threading;
+
+namespace System.Net
+{
+    internal class Logging
+    {
+        private static volatile bool s_LoggingEnabled = true;
```
stephentoub: Logging is enabled by default?

src/Common/src/System/Net/Logging/Logging.cs
```diff
@@ -0,0 +1,724 @@
...
+                {
+                    InitializeLogging();
+                }
+                if (!s_LoggingEnabled)
+                {
+                    return null;
+                }
+                return s_WebSocketsTraceSource;
+            }
+        }
```
stephentoub: Would it make sense to introduce a helper for the boilerplate used in all of these TraceSource-returning properties? e.g.
```C#
private static TraceSource GetTraceSource(ref TraceSource source)
{
    if (!s_LoggingInitialized)
    {
        InitializeLogging();
    }
    return s_LoggingEnabled ? source : null;
}
```
which could then be used like:
```
internal static TraceSource WebSockets { get { return GetTraceSource(ref s_WebSocketsTraceSource); } }
```
or are there concerns around inlining or something like that?

src/Common/src/System/Net/Logging/Logging.cs
```diff
@@ -0,0 +1,724 @@
...
+        {
+            if (!ValidateSettings(traceSource, TraceEventType.Error))
+            {
+                return;
+            }
+
+            string infoLine = SR.Format(SR.net_log_exception, GetObjectLogHash(obj), method, e.Message);
+            if (!string.IsNullOrEmpty(e.StackTrace))
+            {
+                infoLine += "\r\n" + e.StackTrace;
```
stephentoub: Environment.NewLine?

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+            new RecognizedAttribute(Cookie.DomainAttributeName, CookieToken.Domain),
+            new RecognizedAttribute(Cookie.SecureAttributeName, CookieToken.Secure),
+            new RecognizedAttribute(Cookie.DiscardAttributeName, CookieToken.Discard),
+            new RecognizedAttribute(Cookie.PortAttributeName, CookieToken.Port),
+            new RecognizedAttribute(Cookie.CommentAttributeName, CookieToken.Comment),
+            new RecognizedAttribute(Cookie.CommentUrlAttributeName, CookieToken.CommentUrl),
+            new RecognizedAttribute(Cookie.HttpOnlyAttributeName, CookieToken.HttpOnly),
+        };
+
+        private static RecognizedAttribute[] s_recognizedServerAttributes = {
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+            return CookieToken.Unknown;
+        }
+    }
+
+    // CookieParser
+    //
+    // Takes a cookie header, makes cookies.
+    internal class CookieParser
+    {
+        private CookieTokenizer _tokenizer;
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+            Cookie right = (Cookie)or;
+
+            int result;
+
+            if ((result = string.Compare(left.Name, right.Name, StringComparison.OrdinalIgnoreCase)) != 0)
+            {
+                return result;
+            }
+
+            if ((result = string.Compare(left.Domain, right.Domain, StringComparison.OrdinalIgnoreCase)) != 0)
```
stephentoub: Equals instead of Compare?

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+        internal static string CheckQuoted(string value)
+        {
+            if (value.Length < 2 || value[0] != '\"' || value[value.Length - 1] != '\"')
+                return value;
+
+            return value.Length == 2 ? string.Empty : value.Substring(1, value.Length - 2);
+        }
+    }
+
+    internal class Comparer : IComparer
```
stephentoub: ```IComparer<Cookie>``` ?

stephentoub: And since this isn't nested inside of ```Cookie```, should it be named ```CookieComparer``` or something more specific than just ```Comparer```?

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+
+        public Cookie this[string name]
+        {
+            get
+            {
+                foreach (Cookie c in _list)
+                {
+                    if (string.Compare(c.Name, name, StringComparison.OrdinalIgnoreCase) == 0)
```
stephentoub: string.Equals?

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
```diff
@@ -0,0 +1,297 @@
...
+        internal enum Stamp
+        {
+            Check = 0,
+            Set = 1,
+            SetToUnused = 2,
+            SetToMaxUsed = 3,
+        }
+
+        internal int _version;
+        private ArrayList _list = new ArrayList();
```
stephentoub: ```List<Cookie>```?

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        private DateTime _timeStamp = DateTime.MinValue;
+        private bool _hasOtherVersions;
+
+        private bool _isReadOnly;
```
stephentoub: readonly? (no pun intended)

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
```diff
@@ -0,0 +1,297 @@
...
+            {
+                _list.Add(cookie);
+            }
+            else
+            {
+                _list[idx] = cookie;
+            }
+        }
+
+        public void Add(CookieCollection cookies)
```
stephentoub: How common is it to have large numbers of cookies?  As implemented, this is an ```O(N^2)``` operation in the number of cookies.

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+            {
+                return (Version == 0) ? string.Empty :
+                                       (SpecialAttributeLiteral
+                                       + VersionAttributeName
+                                       + EqualsLiteral + (IsQuotedVersion ? "\"" : string.Empty)
+                                       + _version.ToString(NumberFormatInfo.InvariantInfo) + (IsQuotedVersion ? "\"" : string.Empty));
+            }
+        }
+
+        internal static IComparer GetComparer()
```
stephentoub: If the return type were changed to the concrete implementation, and the IComparer implementation were done implicitly rather than explicitly, then call sites using the comparer directly could avoid the interface calls.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        // If isStrict == false, assumes that incoming cookie is unique.
+        // If isStrict == true, replace the cookie if found same with newest Variant.
+        // Returns 1 if added, 0 if replaced or rejected.
+        internal int InternalAdd(Cookie cookie, bool isStrict)
+        {
+            int ret = 1;
+            if (isStrict)
+            {
+                IComparer comp = Cookie.GetComparer();
```
stephentoub: It looks like all of the call sites to GetComparer then only ever compare the result of Compare to 0.  Should this be an EqualityComparer instead?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+// * A cookie with Domain=.com or Domain=.com., will always be rejected, because there is no embedded dot.
+//
+// * A cookie with Domain=ajax.com will be rejected because the value for Domain does not begin with a dot.
+
+namespace System.Net
+{
+    internal struct HeaderVariantInfo
+    {
+        private string _name;
+        private CookieVariant _variant;
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+        public const int DefaultCookieLimit = 300;
+        public const int DefaultPerDomainCookieLimit = 20;
+        public const int DefaultCookieLengthLimit = 4096;
+
+        private static readonly HeaderVariantInfo[] s_headerInfo = {
+            new HeaderVariantInfo(HttpKnownHeaderNames.SetCookie,  CookieVariant.Rfc2109),
+            new HeaderVariantInfo(HttpKnownHeaderNames.SetCookie2, CookieVariant.Rfc2965)
+        };
+
+        private Hashtable _domainTable = new Hashtable();
```
stephentoub: ```Dictionary<string, PathList>``` ?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+                throw new ArgumentNullException("cookie");
+            }
+
+            if (cookie.Domain.Length == 0)
+            {
+                throw new ArgumentException(SR.net_emptystringcall, "cookie.Domain");
+            }
+
+            Uri uri;
+            StringBuilder uriSb = new StringBuilder();
```
stephentoub: At some point we might want to look at whether ```StringBuilderCache``` would be warranted here.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+                        _count += cookies.InternalAdd(cookie, true);
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                if (e is OutOfMemoryException)
+                {
+                    throw;
+                }
```
stephentoub: Should this instead just be its own catch block? e.g.
```C#
catch (OutOfMemoryException)
{
    throw;
}
catch (Exception)
{
    if (throwOnError)
    {
        ...
    }
}
```

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+        // Also note that expired cookies are also removed during request preparation
+        // (this.GetCookies method).
+        //
+        // Param. 'domain' == null means to age in the whole container.
+        private bool AgeCookies(string domain)
+        {
+            // Border case: shrunk to zero
+            if (_maxCookies == 0 || _maxCookiesPerDomain == 0)
+            {
+                _domainTable = new Hashtable();
```
stephentoub: I'm not quite understanding the concurrency guarantees of these types.  In some places, we're locking on ```_domainTable.SyncRoot```, but here we're replacing the entire ```_domainTable``` object, without a lock, such that other threads could still be accessing the old _domainTable.  Is that ok?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+                return true;
+            }
+
+            // Test domain membership.
+            if (string.Compare(_fqdnMyDomain, 0, host, dot, _fqdnMyDomain.Length, StringComparison.OrdinalIgnoreCase) == 0)
+            {
+                return true;
+            }
+
+            // Test for "127.###.###.###" without using regex.
```
stephentoub: Is the "without using regex" important because performance is important?  If so, we probably don't want to be using String.Split either, and instead just using IndexOf to walk each part on the original string.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+                    // parameter isStrict = true below
+                    cookies.InternalAdd(cookie, true);
+                } while (true);
+            }
+            catch (Exception e)
+            {
+                if (e is OutOfMemoryException)
+                {
+                    throw;
+                }
```
stephentoub: Same catch comment as earlier.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+        internal string GetCookieHeader(Uri uri, out string optCookie2)
+        {
+            CookieCollection cookies = InternalGetCookies(uri);
+            string cookieString = String.Empty;
+            string delimiter = String.Empty;
+
+            foreach (Cookie cookie in cookies)
+            {
+                cookieString += delimiter + cookie.ToString();
+                delimiter = "; ";
```
stephentoub: Seems like this could benefit from a StringBuilder?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+            }
+            optCookie2 = cookies.IsOtherVersionSeen ?
+                          (Cookie.SpecialAttributeLiteral +
+                           Cookie.VersionAttributeName +
+                           Cookie.EqualsLiteral +
+                           Cookie.MaxSupportedVersion.ToString(NumberFormatInfo.InvariantInfo)) : String.Empty;
```
stephentoub: These all all constants, but the ToString here is going to force a string to be allocated for MaxSupportedVersion and then a string to be allocated for the concatenation.  Since MaxSupportedVersion is 1, could a MaxSupportedVersionLiteral const be added as "1" and used here?  Or if for some reason we must use ToString like this, could this whole string be cached in a readonly static string?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+            return GetCookieHeader(uri, out dummy);
+        }
+
+        internal string GetCookieHeader(Uri uri, out string optCookie2)
+        {
+            CookieCollection cookies = InternalGetCookies(uri);
+            string cookieString = String.Empty;
+            string delimiter = String.Empty;
+
+            foreach (Cookie cookie in cookies)
```
stephentoub: CookieCollection doesn't expose a struct-based, strongly-typed GetEnumerator, so every iteration through it will at least allocate an enumerator object.  Consider changing these to be for loops or adding a such a struct-based enumerator for internal usage (it would probably need some additional boilerplate since there's already a public GetEnumerator that returns IEnumerable).

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+            {
+                throw new ArgumentNullException("cookieHeader");
+            }
+            CookieCutter(uri, null, cookieHeader, true); // Will throw on error
+        }
+    }
+
+    internal class PathList
+    {
+        private SortedList _list = (SortedList.Synchronized(new SortedList(PathListComparer.StaticInstance)));
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+
+        public object this[string s]
+        {
+            get
+            {
+                return _list[s];
+            }
+            set
+            {
+                lock (SyncRoot)
```
stephentoub: _list is already a synchronized list, and SyncRoot just returns _list.SyncRoot, so this lock appears to be unnecessary.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
stephentoub: ```SortedList<string, PathList>``` ?  In which case Synchronized wouldn't be available, but PathList can provide the locking that's needed, and in my opinion that'll make the locking more clear.  For example, the fact that this instance needs to lock on SyncRoot in order to enumerate _list.Values, whereas it doesn't in order to access other properties of the list, is correct but confusing.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+using System.Globalization;
+
+namespace System.Net
+{
+    // More sophisticated password cache that stores multiple
+    // name-password pairs and associates these with host/realm.
+    public class CredentialCache : ICredentials, ICredentialsByHost, IEnumerable
+    {
+        private Hashtable _cache = new Hashtable();
+        private Hashtable _cacheForHosts = new Hashtable();
```
stephentoub: ```Dictionary<...>``` ?

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
```diff
@@ -0,0 +1,1062 @@
...
+            {
+                Cookie new_cookie = c.Clone();
+                new_cookie.VerifySetDefaults(new_cookie.Variant, uri, isLocalDomain, _fqdnMyDomain, true, true);
+                Add(new_cookie, true);
+            }
+        }
+
+        internal CookieCollection CookieCutter(Uri uri, string headerName, string setCookieHeader, bool isThrow)
+        {
+            GlobalLog.Print("CookieContainer#" + Logging.HashString(this) + "::CookieCutter() uri:" + uri + " headerName:" + headerName + " setCookieHeader:" + setCookieHeader + " isThrow:" + isThrow);
```
stephentoub: GlobalLog.Print will only trace if tracing is enabled on the underlying event source.  But regardless of whether it's enabled, these calls are going to be doing lots of work, e.g. hashing, concatenating the string to pass to Print, etc., even if it'll then be thrown away.  Seems like there should be a GlobalLog.IsEnabled property (that internally calls eventSource.Log.IsEnabled), or something like that, in order to enable using it at all of these call sites, e.g.
```C#
if (GlobalLog.IsEnabled)
{
    GlobalLog.Print(...);
}
```

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+            base(string.Empty, string.Empty, string.Empty)
+        {
+        }
+    }
+
+    internal class CredentialHostKey
+    {
+        internal string Host;
+        internal string AuthenticationType;
+        internal int Port;
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+
+        internal bool Match(string host, int port, string authenticationType)
+        {
+            if (host == null || authenticationType == null)
+            {
+                return false;
+            }
+
+            // If the protocols don't match, this credential is not applicable for the given Uri.
+            if (string.Compare(authenticationType, AuthenticationType, StringComparison.OrdinalIgnoreCase) != 0)
```
stephentoub: string.Equals?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+            {
+                return false;
+            }
+            if (string.Compare(Host, host, StringComparison.OrdinalIgnoreCase) != 0)
```
stephentoub: string.Equals?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+            if (comparand == null)
+            {
+                // This covers also the compared == null case
+                return false;
+            }
+
+            bool equals =
+                (string.Compare(AuthenticationType, comparedCredentialKey.AuthenticationType, StringComparison.OrdinalIgnoreCase) == 0) &&
+                (string.Compare(Host, comparedCredentialKey.Host, StringComparison.OrdinalIgnoreCase) == 0) &&
+                Port == comparedCredentialKey.Port;
```
stephentoub: string.Equals?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+                return false;
+            }
+            if (string.Compare(Host, host, StringComparison.OrdinalIgnoreCase) != 0)
+            {
+                return false;
+            }
+            if (port != Port)
+            {
+                return false;
+            }
```
stephentoub: These same checks are being done in Equals.  Could this just call that?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+        {
+            return "[" + Host.Length.ToString(NumberFormatInfo.InvariantInfo) + "]:" + Host + ":" + Port.ToString(NumberFormatInfo.InvariantInfo) + ":" + Logging.ObjectToString(AuthenticationType);
+        }
+    }
+
+    internal class CredentialKey
+    {
+        internal Uri UriPrefix;
+        internal int UriPrefixLength = -1;
+        internal string AuthenticationType;
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+
+        internal bool Match(Uri uri, string authenticationType)
+        {
+            if (uri == null || authenticationType == null)
+            {
+                return false;
+            }
+
+            // If the protocols don't match, this credential is not applicable for the given Uri.
+            if (string.Compare(authenticationType, AuthenticationType, StringComparison.OrdinalIgnoreCase) != 0)
```
stephentoub: string.Equals?

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
```diff
@@ -0,0 +1,523 @@
...
+                return false;
+            }
+
+            int prefixLen = prefixUri.AbsolutePath.LastIndexOf('/');
+            if (prefixLen > uri.AbsolutePath.LastIndexOf('/'))
+            {
+                return false;
+            }
+
+            return String.Compare(uri.AbsolutePath, 0, prefixUri.AbsolutePath, 0, prefixLen, StringComparison.OrdinalIgnoreCase) == 0;
```
stephentoub: string.Equals? I'll stop commenting on these.  If you choose to fix, I'd suggest just doing a search for all of them.

src/System.Net.Primitives/src/System/Net/DnsEndPoint.cs
```diff
@@ -0,0 +1,93 @@
...
+
+using System.Net.Sockets;
+
+namespace System.Net
+{
+    public class DnsEndPoint : EndPoint
+    {
+        private string _host;
+        private int _port;
+        private AddressFamily _family;
```
stephentoub: readonly?

src/System.Net.Primitives/src/System/Net/IPAddress.cs
```diff
@@ -0,0 +1,535 @@
...
+        public static readonly IPAddress IPv6Loopback = new IPAddress(new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1 }, 0);
+        public static readonly IPAddress IPv6None = new IPAddress(new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }, 0);
+
+        /// <devdoc>
+        ///   <para>
+        ///     Default to IPv4 address
+        ///   </para>
+        /// </devdoc>
+        private AddressFamily _family = AddressFamily.InterNetwork;
+        private ushort[] _numbers = new ushort[NumberOfLabels];
```
stephentoub: readonly?  I'll stop commenting on potential readonly fields as well.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+        }
+
+        /// <devdoc>
+        ///    <para>
+        ///       Initializes a new instance of the <see cref='System.Net.NetworkCredential'/>
+        ///       class with name and password set as specified.
+        ///    </para>
+        /// </devdoc>
+        public NetworkCredential(string userName, SecureString password)
```
stephentoub: @weshaggard, SecureString is showing up here in public surface area.  Is this not part of the contract?

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
```diff
@@ -0,0 +1,220 @@
...
+{
+    /// <devdoc>
+    ///    <para>
+    ///       Provides credentials for password-based authentication schemes such as basic,
+    ///       digest, NTLM and Kerberos.
+    ///    </para>
+    /// </devdoc>
+    public class NetworkCredential : ICredentials, ICredentialsByHost
+    {
+        private static readonly object s_lockingObject = new object();
```
stephentoub: Where is this used?

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
```diff
@@ -0,0 +1,220 @@
...
+            return _domain;
+        }
+
+        internal string InternalGetDomainUserName()
+        {
+            string domainUserName = InternalGetDomain();
+            if (domainUserName.Length != 0)
+                domainUserName += "\\";
+            domainUserName += InternalGetUserName();
+            return domainUserName;
```
stephentoub: If this were instead written like:
```C#
string domain = InternalGetDomain();
string userName = InternalGetUserName();
return domain.Length != 0 ?
    domain + "\\" + userName :
    userName;
```
I believe that'd save at least one string allocation.

src/System.Net.Primitives/src/System/Net/NetworkInformation/NetworkInformationException.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Runtime.InteropServices;
+
+namespace System.Net.NetworkInformation
+{
+    /// <devdoc>
+    ///    <para>
+    ///       Provides NetworkInformation exceptions to the application.
+    ///    </para>
+    /// </devdoc>
+    public class NetworkInformationException : Win32Exception
```
stephentoub: Same question as in other networking libraries about the Win32Exception base type.

src/System.Net.Primitives/src/System/Net/SocketException.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Runtime.InteropServices;
+
+namespace System.Net.Sockets
+{
+    /// <devdoc>
+    ///    <para>
+    ///       Provides socket exceptions to the application.
+    ///    </para>
+    /// </devdoc>
+    public class SocketException : Win32Exception
```
stephentoub: Ditto.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
```diff
@@ -0,0 +1,283 @@
...
+
+        public SocketAddress(AddressFamily family) : this(family, MaxSize)
+        {
+        }
+
+        public SocketAddress(AddressFamily family, int size)
+        {
+            if (size < WriteableOffset)
+            {
+                // It doesn't make sense to create a socket address with less tha
```
stephentoub: nit: tha => than

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
```diff
@@ -0,0 +1,283 @@
...
+                // 2 bytes: that's where we store the address family.
+                throw new ArgumentOutOfRangeException("size");
+            }
+            InternalSize = size;
+            Buffer = new byte[(size / IntPtr.Size + 2) * IntPtr.Size];
+
+#if BIGENDIAN
```
stephentoub: Should this  instead be based on BitConverter.IsLittleEndian?

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
```diff
@@ -0,0 +1,283 @@
...
+                    }
+                    _hash ^= remnant;
+                }
+            }
+            return _hash;
+        }
+
+        public override string ToString()
+        {
+            StringBuilder bytes = new StringBuilder();
```
stephentoub: There's enough StringBuilder usage (or potential for StringBuilder usage) that it does seem like it could be valuable to bring in StringBuilderCache.

stephentoub:
Thanks, Pat.  I left a bunch of comments/questions; up to you if/when/how you want to address them.  Otherwise, LGTM.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
davidsh: .NET Desktop has NetworkCredentials with the SecureString overloads for the constructor and the `SecurePassword` property.  However, those additional overloads and property did exist in the System.Net.Primitives 4.0.10.0 contract.  So, they should NOT be in the source code nor showing up as public methods.

weshaggard: I agree with @davidsh and the dependency on SecureString should be removed both from the public surface and the implementation.

src/System.Net.Primitives/src/System/Net/Cookie.cs
```diff
@@ -0,0 +1,1755 @@
...
+                    {
+                        if (!(variant == CookieVariant.Rfc2965 || variant == CookieVariant.Plain))
+                        {
+                            if (isThrow)
+                            {
+                                throw new CookieException(SR.Format(SR.net_cookie_attribute, DomainAttributeName, _domain));
+                            }
+                            return false;
+                        }
+                        domain = '.' + domain;
```
justinvp: This will result in an additional box allocation when compiled with the VS2013 csc (because `String.Concat(object, object)` will be used). Not a problem with Roslyn, though, as @stephentoub added an optimization that converts the `char` literal to a `string` literal so that `String.Concat(string, string)` can be used, avoiding the box. I assume corefx will be moving to the Roslyn compiler soon, so might not be worth fixing in the code.

src/Common/src/System/Net/Logging/GlobalLog.cs
pgavlin: Done.

src/Common/src/System/Net/Logging/GlobalLog.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
pgavlin: Done.

src/Common/src/System/Net/Logging/GlobalLog.cs
pgavlin: That does appear to be the case. Fixed.

src/Common/src/System/Net/Logging/Logging.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Cookie.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Cookie.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
davidsh: `CookieContainer` is a memory object.  It is NOT common for it to have tons of cookies.  It would only typically have cookies from the lifetime of the HttpClient object that was created to process one or more requests.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Cookie.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Cookie.cs
pgavlin: I've removed this method in favor of a CookieComparer singleton provided by a property on that type.

src/System.Net.Primitives/src/System/Net/CookieCollection.cs
pgavlin: Maybe. I'm going to leave this as-is for now: its equality rules differ from those of `Cookie` itself.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: Added `MaxSupportedVersionString` and an assert to validate that `MaxSupportedVersion.ToString(...)` matches its value.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: I'll make this `readonly`, but I'm going to leave the collection type the same for the time being.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
pgavlin: You're correct, but I think leaving the locking in for now will make the eventual move to SortedList<> more straightforward.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/CredentialCache.cs
pgavlin: Done. Some calls to `string.Compare` were using the variant that takes indices and a length and could not be converted.

src/System.Net.Primitives/src/System/Net/DnsEndPoint.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/IPAddress.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
pgavlin: So the password should be stored as a plaintext string?

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
pgavlin: It isn't. Removed.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
pgavlin: Done.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
davidsh: There are 2 aspects to this:

1.  The `SecureString` class is not available in NETNative (UWP) contracts.  And while it is currently available in CoreFX contracts, it is probably going away.  So, it is important that there are no public API surfaces in the `NetworkCredential` class that have `SecureString` in them.

2. As far as how/where to store the data inside the `NetworkCredential` object (implementation), you should NOT use `SecureString`.  The reality is that, yes, you will store it in memory as plaintext.  I.e. the format that it came from in the public constructor/property.

stephentoub:
Thanks for all the updates, Pat.  It looks good.  The only thing that I think must be reviewed/addressed is that comment related to the synchronization expectations of Hashtable/Dictionary.

src/Common/src/System/Net/Logging/Logging.cs
CIPop: @stephentoub This isn't yet ready but we need the class to compile. Long term ETW tracing will always be enabled similarly to how GlobalLog works now.

src/Common/src/System/Net/Logging/GlobalLog.cs
CIPop: /cc: @josguil 
@stephentoub Looks like this is only used for debugging purposes, mostly in NetworkStream, Sockets and TCPClient. Optimization is not really a goal. We may need to `#if DEBUG` this block.

src/Common/src/System/Net/Logging/Logging.cs
CIPop: @stephentoub @pgavlin I wouldn't spend too much time reviewing this code. It'll change to ETW: See #2500 . We just have these in place to make the implementation build.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
CIPop: /cc @josguil

All above performance improvements for logging mentioned by @stephentoub will be tracked by #2500.

src/System.Net.Primitives/src/System/Net/NetworkCredential.cs
CIPop: @pgavlin I agree with @weshaggard and @davidsh : SecureString should be removed. We have confirmed with the security team that there is no expected security boundary within a process. 

I believe that this code was eventually added back as SecureString was added for backward compat reasons only.

src/System.Net.Primitives/src/System/Net/NetworkInformation/NetworkInformationException.cs
CIPop: @stephentoub: I can't find your comment anywhere - sorry I'm new with GitHub... Please add it here again.

CIPop: @pgavlin: I have confirmed offline with @ericstj, @davidsh and @SidharthNabar  that, in this particular case we would like to remove NetworkInformationException from System.Net.Primitives and leave it within S.N.NetworkInformation.

Please change the code throwing the exception to throw Win32Exception instead.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
CIPop: @stephentoub I think the original dev's intent was to avoid a conditional on Windows where BIGENDIAN should always be false. 

Maybe we can postpone this until we start the x-plat PAL? (The entire BIGENDIAN block should disappear for Windows.)

src/Common/src/System/Net/Logging/GlobalLog.cs
stephentoub: As long as this is only executed in debug builds or in a particular debugging configurations, not optimizing it is fine.

src/Common/src/System/Net/Logging/Logging.cs
stephentoub: @CIPop, when you say it'll always be enabled, do you mean there will always be a listener and tracing will always output to it, or do you mean that the tracing will always be done in case there is a listener?  I'm assuming the latter, in which case I just want to make sure we follow-through on my later comment and use EventSourceWhatever.Log.IsEnabled around call sites to tracing where actual work is being done, e.g. where string concatenation is being done to formulate what's being traced.  Also, when the tracing mechanism is further designed, it'd be great if there was less of a focus on strings being output and more of a focus on integer values or enums or the like, things that are cheaper to trace out, such that enabling tracing doesn't cause a huge hit to perf.

src/Common/src/System/Net/Logging/Logging.cs
stephentoub: Ok, thanks.

src/System.Net.Primitives/src/System/Net/CookieContainer.cs
stephentoub: @CIPop, ok, thanks.

src/System.Net.Primitives/src/System/Net/NetworkInformation/NetworkInformationException.cs
stephentoub: @CIPop, sorry, I wasn't being very clear: I'd made a few comments about Win32Exception in previous networking-related PRs on GitHub, not in this one.  The comment was simply asking whether Win32Exception is the right base class.  We've made Win32Exception work on Unix, essentially wrapping an errno value and doing the appropriate strerror string translations, but it's an unfortunately named type.

src/System.Net.Primitives/src/System/Net/Sockets/SocketAddress.cs
stephentoub: Ok.

src/System.Net.Primitives/src/System/Net/NetworkInformation/NetworkInformationException.cs
CIPop: That’s what I understood as well. I completely agree on the Win32Exception name being unfortunate… Unless we decide to change (which I was told we will not for app-compat reasons), there’s nothing we could really do here.

src/Common/src/System/Net/Logging/Logging.cs
CIPop: @stephentoub /cc: @josguil, @vancem  Yes: by design all ETW tracing will be "enabled". 
I agree that we need to check that the log is enabled prior to performing all calculations. This was left out given that GlobalLog was previously only used when DEBUG was enabled. We'll follow the same pattern as with the Logging object.
Tracking this item with #2500.

Issue #2545, opened 2015-07-27
Directory.Move should either throw on all platforms or succeed on all platforms when moving cross-volume


sokket:
Currently, Directory.Move on Windows calls MoveFileEx with MOVEFILE_COPY_ALLOWED, which will succeed when the source and destination are on different volumes; on Linux, this call will fail since UnixFileSystem.cs uses rename and rename does not handle cross-volume moves. This is visible when trying to, for example, move a directory into or out of Temp since Temp (on default Ubuntu installs) is on a different mountpoint. 

MSDN states that this API should throw when moving cross-volume so we should either fix the documentation + Linux code or add an extra limitation in the Windows code.

PR #2560, opened 2015-07-29
Fixing bug #2545 where Directory.Move would not throw when it should

sokket:
Fixing a bug in the Win32 FileSystem implementation where we would 
allow moving a directory across volumes only on Windows. This
contradicts the MSDN documentation for that function and also created
a functional difference on Linux where that call would throw, as expected.
This fix makes Windows throw like the other implementations and as the
contract states should happen.

### DIFF
--- a/src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
@@ -8,12 +8,14 @@ internal partial class Interop
 {
     internal partial class mincore
     {
+        private const uint MOVEFILE_COPY_ALLOWED = 2;
+
         [DllImport(Libraries.CoreFile_L2, EntryPoint = "MoveFileExW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
         private static extern bool MoveFileEx(string src, string dst, uint flags);
 
-        internal static bool MoveFile(string src, string dst)
+        internal static bool MoveFile(string src, string dst, bool copyCrossVolume = true)
         {
-            return MoveFileEx(src, dst, 2 /* MOVEFILE_COPY_ALLOWED */);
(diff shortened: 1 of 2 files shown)


sokket:
@stephentoub @ericstj 

cc: @piotrpMSFT @ianhays

ianhays:
LGTM! An odd error popped up regarding the xunit version during build btw.

sokket:
@ianhays  yeah, not sure what that is about since none of my code touches xunit or anything related to bringing down packages...

@ellismg have you seen this Roslyn error before?

sokket:
@dotnet-bot test this please

ellismg:
Should be fixed by #2564.

@dotnet-bot test this please

ellismg:
@dotnet-bot test this please

stephentoub:
Before we go ahead with merging this, @sokket, can you confirm how it was you verified there was actually a problem?  I just tried it (prior to your fix), and on Windows I don't even get as far as a call to MoveFileEx due to this check: https://github.com/dotnet/corefx/blob/2678fc72dd27b347bb08a17fee4c6642b258cc10/src/System.IO.FileSystem/src/System/IO/Directory.cs#L561-L564

If I remove that check, then MoveFileEx fails when trying to copy the directory to a different volume.  And consulting the docs for MoveFileEx appear to imply this is by design:
```
When moving a file, the destination can be on a different file system or volume. If the destination is on another drive, you must set the MOVEFILE_COPY_ALLOWED flag in dwFlags.

When moving a directory, the destination must be on the same drive.
```
implying that this flag is just ignored for directories.

stephentoub:
(If we do actually need this, then the fix LGTM.)

sokket:
Blargh looks like I jumped the gun a little bit on this one...read too much unofficial documentation and didn't actually try it myself (on Windows at least)...my bad on this. Closing PR, not needed

PR #2566, opened 2015-07-29
Add tests for SuppressMessageAttribute in System.Diagnostics.Tools contract.

Priya91:


mellinoe:
Thanks for porting these over. I don't prefer having all of the methods start with "Test", seems unnecessary to me, especially when you added the "Tests" namespace, and when the class ends with "Tests". Might just be a preference thing, though.

Priya91:
I've be using the Test name in Process as well, and in previous team's test frameworks.. Just kind of stuck as a coding pattern. And it also gives meaning when reading the test name, as in, Test the constructor, or in other cases as in Process tests, TestStartAddressProperty, etc..

mellinoe:
Please mirror the change back into TFS whichever way we end up so they are still in sync. I'm fine either way.

stephentoub:
@dotnet-bot test this please

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [Fact]
+        public static void TestConstructorBasic()
+        {
+            string expectedCategory = "Category";
+            string expectedCheckId = "CheckId";
+            SuppressMessageAttribute sma = new SuppressMessageAttribute(expectedCategory, expectedCheckId);
+            Assert.Equal(expectedCategory, sma.Category);
+            Assert.Equal(expectedCheckId, sma.CheckId);
+        }
```
stephentoub: It'd be good to also test a few values known to sometimes cause exceptions (but shouldn't here), e.g.
```C#
[Theory]
[InlineData(null, "CheckId")]
[InlineData("", "CheckId")]
[InlineData("Category", null)]
[InlineData("Category", "")]
[InlineData("Category", "CheckId")]
public void TestConstructorBasic(string category, string checkId)
{
    var sma = new SuppressMessageAttribute(category, checkId);
    Assert.Equal(category, sma.Category);
    Assert.Equal(checkId, sma.CheckId);
}
```

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.Diagnostics.CodeAnalysis.Tests
+{
+    public static class SuppressMessageAttributeTests
+    {
+        [Fact]
+        public static void TestConstructorBasic()
```
stephentoub: Why are all of the tests static?  This will impede xunit's parallelization of tests in the assembly.

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            string expectedCategory = "Category";
+            string expectedCheckId = "CheckId";
+            SuppressMessageAttribute sma = new SuppressMessageAttribute(expectedCategory, expectedCheckId);
+            Assert.Equal(expectedCategory, sma.Category);
+            Assert.Equal(expectedCheckId, sma.CheckId);
+        }
+
+        [Fact]
+        public static void TestConstructorExtended()
```
stephentoub: This isn't really testing the constructor any more than TestConstructorBasic is.  This is really a test of the additional properties on the attribute to make sure that they can be set and then have the same value retrieved.

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
Priya91: It should not right, the class is also static, hence it can start the tests in separate threads. Or is it that, xunit's design is such, that it wont run tests in parallel if they are static?

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
Priya91: Sure.

src/System.Diagnostics.Tools/tests/System/Diagnostics/CodeAnalysis/SuppressMessageAttributeTests.cs
Priya91: Yeah this was a mirror from internal tests. I'll rewrite these.

mellinoe: I think a rename of the method is fine, given that these tests are mostly just for code coverage. There's nothing to really test on this class as it's essentially just a container for the properties here. The other option is to test each property individually in different test cases.

Issue #2414, opened 2015-07-18
Clean up Begin/End usage in System.IO.Pipes


stephentoub:
The System.IO.Pipes implementation still has a bunch of APM legacy floating around in its implementation, making it both more expensive and more complicated than if it just used Tasks directly.  This should be cleaned up, as was done for System.IO.FileSystem in #972.

Maxwe11:
Are you expecting PRs from community? :smile:

stephentoub:
If you're interested, you're welcome to it! :smile:

stephentoub:
@Maxwe11, do you want me to mark this as assigned to you?

Maxwe11:
@stephentoub I think this issue should be resolved in a couple independent PRs
- [ ] clean up BeginWaitForConnection/EndWaitForConnection
- [ ] clean up BeginRead/BeginWrite/etc.

I already have some work in private branch against first point (BeginWaitForConnection/EndWaitForConnection) so grabbing it now.

Maxwe11:
In the meantime someone else could grab the second point so leave issue as `up-for-grabs` or split it.

stephentoub:
Sounds good.  Thanks.

Maxwe11:
grabbing the rest :smile:

PR #2568, opened 2015-07-30
Cleanup Begin/End usage for connection in NamedPipeServerStream

Maxwe11:
Progress against #2414.

Note that logic  of `Cancel()` method [was taken from `IOCancellationHelper`](https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipes/src/System/IO/Pipes/IOCancellationHelper.Windows.cs#L82) class which is not presented in the diff because it was left unchanged, but it still used by BeginRead/BeginWrite.


Maxwe11:
@dotnet-bot test this please

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -258,180 +255,49 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
...
+
+                    case Interop.mincore.Errors.ERROR_PIPE_CONNECTED:
+                        // IOCompletitionCallback will not be called because we completed synchronously.
+                        completionSource.ReleaseResources();
+                        if (State == PipeState.Connected)
+                        {
+                            throw new InvalidOperationException(SR.InvalidOperation_PipeAlreadyConnected);
+                        }
+                        completionSource.SetCompletedSynchronously();
+                        return completionSource.Task;
```
stephentoub: This could just return Task.CompletedTask.  The (small) benefit of that is we don't need to artificially keep another object (the TaskCompletionSource's Task) alive.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -258,180 +255,49 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
...
-                if (afsar._errorCode == Interop.mincore.Errors.ERROR_OPERATION_ABORTED)
-                {
-                    if (cancellationHelper != null)
-                    {
-                        cancellationHelper.ThrowIOOperationAborted();
-                    }
+                    case Interop.mincore.Errors.ERROR_IO_PENDING:
+                        break;
+                    // If we are here then the pipe is already connected, or there was an error
+                    // so we should unpin and free the overlapped.
```
stephentoub: Nit: this comment should be before the ```break;```, not after.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -258,180 +255,49 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
...
             // async result
-            PipeAsyncResult asyncResult = new PipeAsyncResult();
-            asyncResult._threadPoolBinding = _threadPoolBinding;
-            asyncResult._userCallback = callback;
-            asyncResult._userStateObject = state;
+            var completionSource = new ConnectionCompletionSource(
+                _threadPoolBinding,
+                cancellationToken,
+                this,
+                onSuccess: thisRef => thisRef.State = PipeState.Connected);
```
stephentoub: We only ever construct this ConnectionCompletionSource here, and only ever with this specific callback.  Could we reduce the size of ConnectionCompletionSource by avoiding the need for this delegate and just baking the ```State = PipeState.Connected``` logic into it?

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,226 @@
...
         {
-            get { return _userStateObject; }
+            Debug.Assert(handle != null);
+            Debug.Assert(server != null);
+            Debug.Assert(onSuccess != null);
+
+            _cancellationToken = cancellationToken;
+            _onSuccess = onSuccess;
+            _serverStream = server;
+            _threadPoolBinding = handle;
```
stephentoub: Nit: could you reorder these to match the order of the parameters?

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,226 @@
...
+
+            _overlapped = _threadPoolBinding.AllocateNativeOverlapped((errorCode, numBytes, pOverlapped) =>
+            {
+                // Unpack overlapped, free the pinned overlapped, and complete the operation
```
stephentoub: I don't see the "free the pinned overlapped" being done here... ?

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,226 @@
...
 
-        public Object AsyncState
+        // Argument for onSuccess delegate.
+        private readonly NamedPipeServerStream _serverStream;
+        private readonly ThreadPoolBoundHandle _threadPoolBinding;
+
+        private CancellationTokenRegistration _cancellationRegistration;
+        private int _errorCode;
+        private NativeOverlapped* _overlapped;
+        private long _result;
```
stephentoub: Why is this (and the associated constants) a long rather than an int?

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,226 @@
...
+                }
+                else if (result != CompletedCallback)
+                {
+                    // Failed to set the result, IO is in the process of completing
+                    // Attempt to take the packed result
+                    result = Interlocked.Exchange(ref _result, NoResult);
+                }
+
+                // If we have a callback that needs to be completed
+                if ((result != NoResult) && (result != CompletedCallback) && (result != RegisteringCancellation))
```
stephentoub: Would it not be clearer to say ```if (result == ResultSuccess || result == ResultError)```?

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,226 @@
...
             }
             else
             {
-                _isComplete = true;
-                if (_waitHandle != null)
-                {
-                    _waitHandle.Set();
-                }
+                _onSuccess(_serverStream);
+                TrySetResult(default(VoidResult));
```
stephentoub: SetCompletedSynchronously()?

stephentoub:
cc: @ericstj, @ianhays

Maxwe11:
@dotnet-bot test this please

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
Maxwe11: No, this comment related to the next case `ERROR_PIPE_CONNECTED` and deafult branches.

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
Maxwe11: We call `ReleaseResources()` in `CompleteCallback(...)`

src/System.IO.Pipes/src/System/IO/Pipes/PipeAsyncResult.cs
```diff
@@ -1,88 +1,215 @@
...
+                    Debug.WriteLine("CancelIoEx finished with error code {0}.", errorCode);
+                }
+
+                // From this moment cancellation is no longer possible.
+                // This MUST happen before Overlapped is freed and Handle is disposed.
+                _cancellationRegistration.Dispose();
+            }
+        }
+
+        private void CompleteCallback(long result)
```
Maxwe11: TODO: long -> int

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
Maxwe11: At this point we already have TCS because we inherit from it. You suggest replace inheritance with composition and initialize it lazily?

stephentoub: No, I'm not suggesting changing anything else, just here returning Task.CompletedTask rather than completionSource.Task. Let's say the caller keeps a long reference to the task: it'll then be keeping alive an already cached task rather than yet another task Erich asset this point is equivalent, so we can just drop completionSource and let it be collected.

Maxwe11: makes sence :+1: , done

Maxwe11:
I've wrote the following test.

	[DllImport("api-ms-win-core-io-l1-1-0.dll", SetLastError = true)]
    internal static unsafe extern bool CancelIoEx(SafeHandle handle, NativeOverlapped* lpOverlapped);

    [Fact]
    public static async Task ServerWaitForConnectThrows_OuterCancellation()
    {
        using (NamedPipeServerStream server = new NamedPipeServerStream("unique3", PipeDirection.InOut, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous, 0, 0))
        {
            var token = new CancellationToken(canceled: false);
            Task waitForConnectionTask = server.WaitForConnectionAsync(token);
            unsafe
            {
                Assert.True(CancelIoEx(server.SafePipeHandle, null), "Outer cancellation failed");
            }

            await Assert.ThrowsAsync<IOException>(() => waitForConnectionTask);
        }
    }

It succeeds against my PR but failed against current master what seems to me incorrect.

Maxwe11:
@stephentoub could you take a look at #2640 please?

src/System.IO.Pipes/src/System/IO/Pipes/ConnectionCompletionSource.cs
```diff
@@ -0,0 +1,213 @@
...
+                if (result == ResultSuccess || result == ResultError)
+                {
+                    CompleteCallback(result);
+                }
+            }
+        }
+
+        internal void ReleaseResources()
+        {
+            if (_cancellationRegistration != null)
```
stephentoub: CancellationTokenRegistration is a struct; it can never be null.

src/System.IO.Pipes/src/System/IO/Pipes/ConnectionCompletionSource.cs
```diff
@@ -0,0 +1,213 @@
...
+                    // This case should not have any consequences although
+                    // it will be easier to debug if there exists any special case
+                    // we are not aware of.
+                    int errorCode = Marshal.GetLastWin32Error();
+                    Debug.WriteLine("CancelIoEx finished with error code {0}.", errorCode);
+                }
+
+                // From this moment cancellation is no longer possible.
+                // This MUST happen before Overlapped is freed and Handle is disposed.
+                _cancellationRegistration.Dispose();
```
stephentoub: This shouldn't be necessary.  The only way we could end up in Cancel() is from the registration on the cancellation token, so if this is invoked, that token has had cancellation requested, which will cause it to clean up all of its state.  Functionally it doesn't hurt to have this, though.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -256,182 +253,46 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
                 throw new InvalidOperationException(SR.InvalidOperation_PipeNotAsync);
             }
 
-            // Create and store async stream class library specific data in the 
-            // async result
-            PipeAsyncResult asyncResult = new PipeAsyncResult();
-            asyncResult._threadPoolBinding = _threadPoolBinding;
-            asyncResult._userCallback = callback;
-            asyncResult._userStateObject = state;
+            // Create and store async stream class library specific data in the async result
```
stephentoub: This comment seems out of date.  It can probably just be deleted.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -256,182 +253,46 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
...
+                            throw new InvalidOperationException(SR.InvalidOperation_PipeAlreadyConnected);
+                        }
+                        completionSource.SetCompletedSynchronously();
+
+                        // We return a cached task instead of TaskCompletionSource's Task allowing the GC to collect it.
+                        return Task.CompletedTask;
+
+                    default:
+                        completionSource.ReleaseResources();
+                        throw Win32Marshal.GetExceptionForWin32Error(errorCode);
```
stephentoub: Hmm.  Seems like this (and some of the above exceptions) really should have been returned via Task.FromException, but I see in the old version that we were throwing there, too, so I guess we should leave it.

stephentoub:
A few more comments, but otherwise LGTM.

@ericstj, could you review as well?

stephentoub:
@ianhays, could you review this as well?

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs
```diff
@@ -256,182 +253,45 @@ private unsafe IAsyncResult BeginWaitForConnection(AsyncCallback callback, Objec
...
-            // false here, saying the user callback was called on another thread.
-            asyncResult._completedSynchronously = false;
-            asyncResult._isComplete = true;
-
-            // The OS does not signal this event.  We must do it ourselves.
-            ManualResetEvent wh = asyncResult._waitHandle;
-            if (wh != null)
+            // If we are here then connection is pending.
+            // Check if the caller could cancel and if he could register cancellation.
+            if (cancellationToken.CanBeCanceled)
```
ianhays: This check isn't necessary since _cancellationToken.CanBeCanceled is checked at the beginning of the RegisterForCancellation method anyways on line 68 of ConnectionCompletionSource

src/System.IO.Pipes/src/System/IO/Pipes/ConnectionCompletionSource.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            // Special case for when the client has already connected to us.
+            if (errorCode == Interop.mincore.Errors.ERROR_PIPE_CONNECTED)
+            {
+                errorCode = 0;
+            }
+
+            _errorCode = (int)errorCode;
+
+            int resultState = errorCode != 0 ? ResultError : ResultSuccess;
```
ianhays: nit: ```int resultState = errorCode == 0 ? ResultSuccess : ResultError;``` would be <i>slightly</i> more readable.

src/System.IO.Pipes/tests/NamedPipesThrowsTests.cs
```diff
@@ -412,6 +412,32 @@ public static void ServerWhenDisplosedThrows()
...
+        public static object[][] OuterCancellation = new object[][]
+        {
+            new object[] { CancellationToken.None,              typeof(OperationCanceledException) },
+            new object[] { new CancellationTokenSource().Token, typeof(IOException) }
+        };
+
+        [Theory]
+        [MemberData("OuterCancellation")]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static async Task ServerWaitForConnectThrows_OuterCancellation(CancellationToken cancellationToken, Type expectedException)
```
ianhays: Should we mark this test as being related to #2640?

ianhays:
Had a few minor comments, but nothing major.

LGTM.

src/System.IO.Pipes/tests/NamedPipesThrowsTests.cs
Maxwe11: Will add comment

Maxwe11:
done

src/System.IO.Pipes/tests/System.IO.Pipes.Tests.csproj
```diff
@@ -13,8 +13,10 @@
   </PropertyGroup>
   <!-- Default configurations to help VS understand the configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
stephentoub: Nit: these can be condensed into a single AllowUnsafeBlocks entry in the general property group above.

stephentoub:
Thanks, @Maxwe11. Overall, looks good.  Looks like you're going to need to resolve some conflicts before it can be merged, and as long as you're doing that, I left a comment for one more minor nit to be addressed (you can just ammend your last commit with it).

Maxwe11:
ops...it's failed with recent replacement ThrowsAnyAsync -> ThrowsAsync. Will change it back.

Maxwe11:
Conflict resolved. Unit tests have been updated. CI is green.

stephentoub:
Thanks, @Maxwe11!

PR #2574, opened 2015-07-31
Add support for PFX files on Unix

### LINKED ISSUE #2449: Add support for reading a PFX for Unix builds

bartonjs:
With this change the following things now work on Unix systems:
* Reading a DER-encoded PKCS12 (PFX) `byte[]` via `X509Certificate2` constructors
* Reading a DER-encoded PKCS12 (PFX) `byte[]` via `X509Certificate2Collection.Import`
* Reading a DER-encoded PKCS12 (PFX) file identified by `string filename` via `X509Certificate2` 
constructors
* Reading a DER-encoded PKCS12 (PFX) file identified by `string filename` via `X509Certificate2Collection.Import`
* Reading a DER-encoded X509 (.cer) file identified by `string filename` via `X509Certificate2` 
constructors
* Reading a PEM-encoded X509 (.cer) file identified by `string filename` via `X509Certificate2` 
constructors
* Instantiating an `RSAOpenSsl` based off of having an `IntPtr` to an existing OpenSSL `RSA*` (which enables people on Unix systems to make use of hardware security keys)

The file paths and the byte[] paths don't share as much code as I would have liked, but the `BIO_s_mem` BIO didn't seem to support seeking back to the start when it read down the wrong path.

15 tests get [ActiveIssue] removed, two changed to a more specific number, and one new test was added.

Fixes #2449.

### DIFF
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeRsaHandle.Unix.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeRsaHandle.Unix.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Security;
 using System.Runtime.InteropServices;
 
@@ -25,6 +26,23 @@ namespace Microsoft.Win32.SafeHandles
         public override bool IsInvalid
         {
             get { return handle == IntPtr.Zero; }
+        }
+
+        internal static SafeRsaHandle DuplicateHandle(IntPtr handle)
+        {
+            Debug.Assert(handle != IntPtr.Zero);
+
+            // Reliability: Allocate the SafeHandle before calling RSA_up_ref so
+            // that we don't lose a tracked reference in low-memory situations.
+            SafeRsaHandle safeHandle = new SafeRsaHandle();
+
+            if (!Interop.libcrypto.RSA_up_ref(handle))
+            {
+                throw Interop.libcrypto.CreateOpenSslCryptographicException();
+            }
+
+            safeHandle.SetHandle(handle);
+            return safeHandle;
         }
     }
 }
--- a/src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
+++ b/src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
@@ -12,6 +12,8 @@ namespace System.Security.Cryptography
 {
     public sealed class RSAOpenSsl : RSA
     {
+        private const int BitsPerByte = 8;
+
         // 65537 (0x10001) in big-endian form
         private static readonly byte[] s_defaultExponent = { 0x01, 0x00, 0x01 };
 
@@ -43,6 +45,31 @@ namespace System.Security.Cryptography
         {
             _legalKeySizesValue = new[] { s_legalKeySizes };
             ImportParameters(parameters);
+        }
+
+        /// <summary>
+        /// Create an RSAOpenSsl from an existing <see cref="IntPtr"/> whose value is an
+        /// existing OpenSSL <c>RSA*</c>.
+        /// </summary>
+        /// <remarks>
+        /// This method will increase the reference count of the <c>RSA*</c>, the caller should
+        /// continue to manage the lifetime of their reference.
+        /// </remarks>
+        /// <param name="handle">A pointer to an OpenSSL <c>RSA*</c></param>
+        /// <exception cref="ArgumentException"><paramref name="handle" /> is invalid</exception>
+        public RSAOpenSsl(IntPtr handle)
+        {
+            if (handle == IntPtr.Zero)
+                throw new ArgumentException(SR.Cryptography_OpenInvalidHandle, "handle");
+
+            _legalKeySizesValue = new[] { s_legalKeySizes };
+
+            SafeRsaHandle rsaHandle = SafeRsaHandle.DuplicateHandle(handle);
+
+            // Set base.KeySize to avoid throwing an extra Lazy at the GC when
+            // using something other than the default keysize.
+            base.KeySize = BitsPerByte * Interop.libcrypto.RSA_size(rsaHandle);
+            _key = new Lazy<SafeRsaHandle>(() => rsaHandle);
         }
 
         public override int KeySize
@@ -224,7 +251,7 @@ namespace System.Security.Cryptography
 
             // Set base.KeySize directly, since we don't want to free the key
             // (which we would do if the keysize changed on import)
-            base.KeySize = 8 * Interop.libcrypto.RSA_size(key);
+            base.KeySize = BitsPerByte * Interop.libcrypto.RSA_size(key);
         }
 
         protected override void Dispose(bool disposing)
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
 
 namespace Internal.Cryptography.Pal
 {
@@ -16,9 +18,64 @@ namespace Internal.Cryptography.Pal
             return new OpenSslX509CertificateReader(Interop.libcrypto.X509_dup(handle));
         }
 
-        public static ICertificatePal FromBlob(byte[] rawData, string password, X509KeyStorageFlags keyStorageFlags)
+        public static unsafe ICertificatePal FromBlob(byte[] rawData, string password, X509KeyStorageFlags keyStorageFlags)
         {
-            return new OpenSslX509CertificateReader(rawData);
+            // If we can see a hyphen, assume it's PEM.  Otherwise try DER-X509, then fall back to DER-PKCS12.
+            SafeX509Handle cert;
+
+            // PEM
+            if (rawData[0] == '-')
+            {
+                using (SafeBioHandle bio = Interop.libcrypto.BIO_new(Interop.libcrypto.BIO_s_mem()))
+                {
+                    Interop.libcrypto.CheckValidOpenSslHandle(bio);
+
+                    Interop.libcrypto.BIO_write(bio, rawData, rawData.Length);
+                    cert = Interop.libcrypto.PEM_read_bio_X509_AUX(bio, IntPtr.Zero, IntPtr.Zero, IntPtr.Zero);
+                }
+
+                Interop.libcrypto.CheckValidOpenSslHandle(cert);
+            }
+
+            // DER-X509
+            cert = Interop.libcrypto.OpenSslD2I((ptr, b, i) => Interop.libcrypto.d2i_X509(ptr, b, i), rawData, checkHandle: false);
+
+            if (!cert.IsInvalid)
+            {
+                return new OpenSslX509CertificateReader(cert);
+            }
+
+            // DER-PKCS12
+            OpenSslPkcs12Reader pfx;
+
+            if (OpenSslPkcs12Reader.TryRead(rawData, out pfx))
+            {
+                using (pfx)
+                {
+                    pfx.Decrypt(password);
+
+                    ICertificatePal first = null;
+
+                    foreach (OpenSslX509CertificateReader certPal in pfx.ReadCertificates())
+                    {
+                        // When requesting an X509Certificate2 from a PFX only the first entry is
+                        // returned.  Other entries should be disposed.
+                        if (first == null)
+                        {
+                            first = certPal;
+                        }
+                        else
+                        {
+                            certPal.Dispose();
+                        }
+                    }
+
+                    return first;
+                }
+            }
+
+            // Unsupported
+            throw Interop.libcrypto.CreateOpenSslCryptographicException();
         }
 
         public static ICertificatePal FromFile(string fileName, string password, X509KeyStorageFlags keyStorageFlags)
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
@@ -80,7 +80,13 @@ namespace Internal.Cryptography.Pal
 
         public static ICertificatePal FromFile(string fileName, string password, X509KeyStorageFlags keyStorageFlags)
         {
-            throw new NotImplementedException();
+            // If we can't open the file, fail right away.
+            using (SafeBioHandle fileBio = Interop.libcrypto.BIO_new_file(fileName, "rb"))
+            {
+                Interop.libcrypto.CheckValidOpenSslHandle(fileBio);
+
+                return OpenSslX509CertificateReader.FromBio(fileBio, password);
+            }
         }
     }
 }
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
@@ -1,0 +1,129 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+using Microsoft.Win32.SafeHandles;
+
+namespace Internal.Cryptography.Pal
+{
+    internal sealed class OpenSslPkcs12Reader : IDisposable
+    {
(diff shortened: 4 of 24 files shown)


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
```diff
@@ -16,14 +18,74 @@ public static ICertificatePal FromHandle(IntPtr handle)
...
+        public static unsafe ICertificatePal FromBlob(byte[] rawData, string password, X509KeyStorageFlags keyStorageFlags)
         {
-            return new OpenSslX509CertificateReader(rawData);
+            // If we can see a hyphen, assume it's PEM.  Otherwise try DER-X509, then fall back to DER-PKCS12.
+            SafeX509Handle cert;
+
+            // PEM
+            if (rawData[0] == '-')
+            {
+
```
justinvp: nit: whitespace

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
```diff
@@ -0,0 +1,131 @@
...
+                out _x509Handle,
+                out _caStackHandle);
+
+            if (!parsed)
+            {
+                throw Interop.libcrypto.CreateOpenSslCryptographicException();
+            }
+        }
+
+        public IEnumerable<OpenSslX509CertificateReader> ReadCertificates()
```
justinvp: Just return `List<T>` here. That way, when you enumerate the collection below, it avoids the enumerator allocation (`List<T>` has a struct enumerator).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            if (handle != null && !handle.IsInvalid)
+            {
+                return new OpenSslPkcs12Reader(handle);
+            }
+
+            return null;
+        }
+
+        public static OpenSslPkcs12Reader TryRead(SafeBioHandle fileBio)
```
justinvp: Should this actually follow the Try pattern, returning bool with an out param?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -34,50 +35,70 @@ internal OpenSslX509CertificateReader(SafeX509Handle handle)
...
+            if (cert != null && !cert.IsInvalid)
+            {
+                return new OpenSslX509CertificateReader(cert);
+            }
 
-            if (!init)
+            // Rewind, try again.
+            Interop.NativeCrypto.BioSeek(bio, bioPosition);
+
+            using (OpenSslPkcs12Reader pfx = OpenSslPkcs12Reader.TryRead(bio))
```
justinvp: Ah, I see how this is used now. Maybe it's easier to use with `using` here without using the bool returning Try.

bartonjs:
@dotnet-bot test this please

src/Common/src/Microsoft/Win32/SafeHandles/SafeRsaHandle.Unix.cs
```diff
@@ -26,5 +27,19 @@ public override bool IsInvalid
...
+        internal static SafeRsaHandle DuplicateHandle(IntPtr handle)
+        {
+            Debug.Assert(handle != IntPtr.Zero);
+
+            if (!Interop.libcrypto.RSA_up_ref(handle))
+            {
+                throw Interop.libcrypto.CreateOpenSslCryptographicException();
+            }
+
+            SafeRsaHandle safeHandle = new SafeRsaHandle();
```
stephentoub: Small reliability suggestion: move this allocation to before the RSA_up_ref call.  That way, if we're running low on memory and the allocation were to throw, it'll happen before we increase the ref count and we won't leak the handle.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
```diff
@@ -45,6 +45,31 @@ public RSAOpenSsl(RSAParameters parameters)
...
+        /// continue to manage the lifetime of their reference.
+        /// </remarks>
+        /// <param name="handle">A pointer to an OpenSSL <c>RSA*</c></param>
+        /// <exception cref="ArgumentException"><paramref name="handle" /> is invalid</exception>
+        public RSAOpenSsl(IntPtr handle)
+        {
+            if (handle == IntPtr.Zero)
+                throw new ArgumentException(SR.Cryptography_OpenInvalidHandle, "handle");
+
+            _legalKeySizesValue = new[] { s_legalKeySizes };
```
stephentoub: Is this array exposed or modified anywhere?  If not, can you cache it in a static readonly field rather than allocated on each new RSAOpenSsl instance?

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
```diff
@@ -45,6 +45,31 @@ public RSAOpenSsl(RSAParameters parameters)
...
+
+            SafeRsaHandle rsaHandle = SafeRsaHandle.DuplicateHandle(handle);
+
+            // Set base.KeySize to avoid throwing an extra Lazy at the GC when
+            // using something other than the default keysize.
+            base.KeySize = 8 * Interop.libcrypto.RSA_size(rsaHandle);
```
stephentoub: What is 8 here, bitsPerByte?  If so, could you put it into a named constant?

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
```diff
@@ -45,6 +45,31 @@ public RSAOpenSsl(RSAParameters parameters)
...
+            _key = new Lazy<SafeRsaHandle>(() => rsaHandle);
```
stephentoub: Separate from this PR, we should consider adding another ctor to Lazy, e.g. ```public Lazy(T value)```, for these cases where the value is already available but we still need to wrap it in a ```Lazy<T>``` instance to match other uses.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
```diff
@@ -16,14 +18,74 @@ public static ICertificatePal FromHandle(IntPtr handle)
...
+                    cert = Interop.libcrypto.PEM_read_bio_X509_AUX(bio, IntPtr.Zero, IntPtr.Zero, IntPtr.Zero);
+                }
+
+                Interop.libcrypto.CheckValidOpenSslHandle(cert);
+            }
+
+            // DER-X509
+            cert = Interop.libcrypto.OpenSslD2I((ptr, b, i) => Interop.libcrypto.d2i_X509(ptr, b, i), rawData, checkHandle: false);
+
+            if (cert != null && !cert.IsInvalid)
```
stephentoub: In what situations will you get back a null SafeHandle?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
stephentoub: I was wondering the same thing.  Rather than the usage pattern currently employed:
```C#
using (OpenSslPkcs12Reader reader = TryRead(handle))
{
    if (reader != null)
    {
        ...
    }
}
```
I wonder whether it'd be cleaner / more familiar to have the reader be an out parameter and do:
```C#
OpenSslPkcs12Reader reader;
if (TryRead(handle, out reader))
{
    using (reader)
    {
        ...
    }
}
```
or something like that.  I'm ok with the current code though if it ends up working best.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
```diff
@@ -0,0 +1,131 @@
...
+                password,
+                out _evpPkeyHandle,
+                out _x509Handle,
+                out _caStackHandle);
+
+            if (!parsed)
+            {
+                throw Interop.libcrypto.CreateOpenSslCryptographicException();
+            }
+        }
```
stephentoub: Consider Debug.Assert'ing that all of the out handles are valid?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -34,50 +35,70 @@ internal OpenSslX509CertificateReader(SafeX509Handle handle)
...
-            _cert = cert;
-        }
+            // Since we aren't going to finish reading, leaving the buffer where it was when we got
+            // it seems better than leaving it in some arbitrary other position.
+            Interop.NativeCrypto.BioSeek(bio, bioPosition);
 
-        public bool HasPrivateKey
-        {
-            get { return false; }
+            // None of the formats read.  Whatever OpenSSL had to say last will be the exception.
```
stephentoub: Is the call to BioSeek going to mess with the error stack?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -298,10 +332,25 @@ public string GetNameInfo(X509NameType nameType, bool forIssuer)
...
         {
+            if (!HasPrivateKey)
+            {
+                return;
+            }
+
+            // There's nothing really to say about the key, just acknowledge there is one.
+            sb.AppendLine();
+            sb.AppendLine();
+            sb.AppendLine("[Private Key]");
```
stephentoub: Does this need to be localized?

stephentoub:
Some nits, but otherwise LGTM

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
justinvp: +1 to the `public Lazy(T value)` ctor.

In the meantime, there is a [trick](https://gist.github.com/AArnott/d285feef75c18f6ecd2b) that can be used to avoid the closure allocation.

With the following helper:

```c#
internal static class FuncHelper
{
    internal static Func<T> AsFunc<T>(this T value) where T : class
    {
        return new Func<T>(value.Return);
    }

    private static T Return<T>(this T value)
    {
        return value;
    }
}
```

You can then do this:

```c#
_key = new Lazy<SafeRsaHandle>(rsaHandle.AsFunc());
```

Which reduces the allocations from delegate + closure, to just delegate.

Not necessarily suggesting that it should be used here, but figured I'd mention it.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
bartonjs: It's actually a protected field from [RSA](https://github.com/dotnet/corefx/blob/588aea39e878f696f3e93af6a167855c5397c5f6/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/RSA.cs#L22) that the derived types need to set before exiting the constructor.

(Or, I suppose, by overriding the LegalKeySizes accessor and just cloning my existing static readonly.)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: Apparently not: http://referencesource.microsoft.com/#System/security/system/security/cryptography/x509/x509certificate2.cs,1229

But don't look at that code too long, or you'll be upset at their lack of using AppendLine and instead to make a lot of intermediate temporary strings :smile:.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
bartonjs: There are currently no cases where a null SafeHandle would be returned from the d2i helper.  I can get rid of all of the null halves of the check.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
bartonjs: Yeah, I went back and forth on this one.  I liked the way the current approach captures the disposable object and doesn't leak it into the outer scope; but I disliked that I named it Try without being bool-returning and out-value-emitting; but Try did convey the semantic of "this might not work, but it won't throw, so pay attention".

I'll just change it to
```C#
OpenSslPkcs12Reader reader;
if (TryRead(handle, out reader))
{
    using (reader)
    {
        ...
    }
}
```

and if the variable scope bothers me when I do it, there's always an extra curly brace pair.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
stephentoub: Ah, ok.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
stephentoub: Averting my eyes.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
justinvp: FYI, I added an issue to track adding the `Lazy(T value)` constructor: #2578

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: In current testing it has not; but I can snap off the exception before the final seek-restore for future-proofing.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
bartonjs: I can envision situations in which OpenSsl could emit `NULL` as the answer to any of those (single cert: don't allocate _caStackHandle; no private key, don't allocate _evpPkeyHandle; no private key, don't allocate _x509Handle either... move it into _caStack); so assuming that they just give us "some stuff" and we deal with it feels the most future-proofed.

src/Common/src/Microsoft/Win32/SafeHandles/SafeRsaHandle.Unix.cs
bartonjs: Good catch, fixed.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
bartonjs: Done, and updated the other `8 * ...` to use the constant.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificatePal.cs
bartonjs: Annihilated.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
bartonjs: Done.

bartonjs:
I snapped off the state of the code as it was into unix_pfx_reviewed, and folded the review changes into the appropriate commits; then force pushed.

It's easy for me to review (`git diff origin/unix_pfx_reviewed`); but I'm not sure the best way to advise on reading just the deltas, aside from adding bartonjs/corefx as a remote, checking out bartonjs/unix_pfx, and then using `git diff bartonjs/unix_pfx_reviewed`.

stephentoub:
LGTM. Thanks!

PR #2577, opened 2015-07-31
Win32.Registry improvements

Maxwe11:


src/Microsoft.Win32.Registry/src/Microsoft/Win32/Registry.cs
```diff
@@ -77,41 +76,39 @@ private static RegistryKey GetBaseKeyFromKeyName(string keyName, out string subK
                 throw new ArgumentNullException("keyName");
             }
 
-            string basekeyName;
-            int i = keyName.IndexOf('\\');
-            if (i != -1)
+            RegistryKey basekey = null;
+            const StringComparison comparisonType = StringComparison.OrdinalIgnoreCase;
+
+            if (keyName.StartsWith("HKEY_CURRENT_USER", comparisonType))
```
stephentoub: This doesn't look like an equivalent transformation.  If I pass in a keyName of "HKEY_CURRENT_USER_STEPHEN", previously this would throw an exception, but now it's going to match HKEY_CURRENT_USER at the start, set baseKey to Registry.CurrentUser, and set subKeyName to null.

stephentoub:
Thanks, @Maxwe11.  Most of the changes look good to me (the Array.Empty ones could end up returning a cached empty array out of public APIs, but I think that's fine for these cases).  I'm concerned about the first change around the string matching, though, and suggest that be reverted unless you have a compelling case for why such an optimization is important, in which I think the change will need to be revised further.

Maxwe11:
Sorry for the delay, updated.

>I'm concerned about the first change around the string matching, though, and suggest that be reverted

Reverted, however I've added some tests for case if you will decide make such optimization in the future.

src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
```diff
@@ -1201,10 +1205,7 @@ internal Object InternalGetValue(String name, Object defaultValue, bool doNotExp
                             try
                             {
                                 char[] newBlob = new char[checked(blob.Length + 1)];
-                                for (int i = 0; i < blob.Length; i++)
-                                {
-                                    newBlob[i] = blob[i];
-                                }
+                                Buffer.BlockCopy(blob, 0, newBlob, 0, blob.Length * sizeof(char));
```
stephentoub: Why Buffer.BlockCopy instead of Array.Copy, e.g.
```C#
Array.Copy(blob.Length, 0, newBlob, 0, blog.Length);
```
?

src/Microsoft.Win32.Registry/tests/Registry/Registry_Getvalue_str_str_obj.cs
```diff
@@ -26,6 +26,9 @@ public static void NegativeTests()
 
             // Passing a string which does NOT start with one of the valid base key names, that should throw ArgumentException.
             Assert.Throws<ArgumentException>(() => Registry.GetValue("HHHH_MMMM", null, null));
+
+            // Should throw if passed string which only starts with one of the valid base key names but actually it doesn't valid.
```
stephentoub: nit: doesn't -> isn't

src/Microsoft.Win32.Registry/tests/Registry/Registry_Getvalue_str_str_obj.cs
```diff
@@ -76,7 +86,9 @@ public static void GetValueFromDifferentKeys(RegistryKey key, string valueName)
                 key.SetValue(valueName, expectedValue);
                 try
                 {
-                    Assert.Equal(expectedValue, (int)Registry.GetValue(key.Name, valueName, defaultValue));
+                    // keyName should be case insensitive
+                    string keyName = (key.Name + (useSeparator ? "\\" : "")).ToLowerInvariant();
```
stephentoub: Why the call to ToLowerInvariant?

src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
Maxwe11: fixed

src/Microsoft.Win32.Registry/tests/Registry/Registry_Getvalue_str_str_obj.cs
Maxwe11: fixed, thanks.

src/Microsoft.Win32.Registry/tests/Registry/Registry_Getvalue_str_str_obj.cs
Maxwe11: We have no tests which test that key names are case insensitive.

stephentoub:
LGTM.  Thanks, @Maxwe11.

PR #2587, opened 2015-07-31
Adding new Unit Tests to validate the more complex FileSystemWatcher scenarios

sokket:
Adding new Unit Tests to validate the more complex FileSystemWatcher scenarios: symlinks, deep directory structures, and preseeded directories.

### DIFF
--- a/src/Common/src/Interop/Unix/libc/Interop.poll.cs
+++ b/src/Common/src/Interop/Unix/libc/Interop.poll.cs
@@ -1,0 +1,62 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class libc
+    {
+        [Flags]
+        internal enum PollFlags : short
+        {
+            POLLIN      = 0x0001,       /* any readable data available */
+            POLLERR     = 0x0008,       /* some poll error occurred */
+            POLLHUP     = 0x0010,       /* file descriptor was "hung up" */
+            POLLNVAL    = 0x0020,       /* requested events "invalid" */
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct pollfd
+        {
+            internal int fd;            /* file descriptor to poll*/
+            internal PollFlags events;  /* events to poll for */
+            internal PollFlags revents; /* events received from polling */
+        }
+
+        /// <summary>
+        /// Polls a set of file descriptors for signals and returns what signals have been set
+        /// </summary>
+        /// <param name="fds">A pointer to pollfd structs to look for</param>
+        /// <param name="count">The number of entries in fds</param>
+        /// <param name="timeout">The amount of time to wait; -1 for infinite, 0 for immediate return, and a positive number is the number of milliseconds</param>
+        /// <returns>
+        /// Returns a positive number (which is the number of structures with nonzero revent files), 0 for a timeout or no 
+        /// descriptors were ready, or -1 on error.
+        /// </returns>
+        [DllImport(Libraries.Libc, SetLastError = true)]
+        private static unsafe extern int poll(pollfd* fds, uint count, int timeout);
+
+        /// <summary>
+        /// Polls a File Descriptor for the passed in flags.
+        /// </summary>
+        /// <param name="fd">The descriptor to poll</param>
+        /// <param name="flags">The flags to poll for</param>
+        /// <param name="timeout">The amount of time to wait; -1 for infinite, 0 for immediate return, and a positive number is the number of milliseconds</param>
+        /// <param name="resultFlags">The flags that were returned by the poll call</param>
+        /// <returns>
+        /// Returns a positive number (which is the number of structures with nonzero revent files), 0 for a timeout or no 
+        /// descriptors were ready, or -1 on error.
+        /// </returns>
+        internal unsafe static int poll(int fd, PollFlags flags, int timeout, out PollFlags resultFlags)
+        {
+            pollfd pfd = default(pollfd);
+            pfd.fd = fd;
+            pfd.events = flags;
+            int result = poll(&pfd, 1, timeout);
+            resultFlags = pfd.revents;
+            return result;
+        }
+    }
+}
--- a/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
+++ b/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
@@ -299,10 +299,20 @@ namespace System.IO
             {
                 string fullPath = parent != null ? parent.GetPath(false, directoryName) : directoryName;
 
+                // inotify_add_watch will fail if this is a symlink, so check that we didn't get a symlink
+                Interop.NativeIO.FileStats stats = default(Interop.NativeIO.FileStats);
+                if ((Interop.NativeIO.LStat(fullPath, out stats) == 0) &&
+                    ((stats.Mode & (uint)Interop.NativeIO.FileTypes.S_IFMT) == Interop.NativeIO.FileTypes.S_IFLNK))
+                {
+                    return null;
+                }
+
                 // Add a watch for the full path.  If the path is already being watched, this will return 
-                // the existing descriptor.  This works even in the case of a rename.
+                // the existing descriptor.  This works even in the case of a rename. We also add the DONT_FOLLOW
+                // and EXCL_UNLINK flags to keep parity with Windows where we don't pickup symlinks or unlinked
+                // files (which don't exist in Windows)
                 int wd = (int)SysCall(
-                    (fd, path, thisRef) => Interop.libc.inotify_add_watch(fd, path, (uint)thisRef._notifyFilters),
+                    (fd, path, thisRef) => Interop.libc.inotify_add_watch(fd, path, (uint)(thisRef._notifyFilters | Interop.libc.NotifyEvents.IN_DONT_FOLLOW | Interop.libc.NotifyEvents.IN_EXCL_UNLINK)),
                     fullPath,
                     this);
 
@@ -476,7 +486,21 @@ namespace System.IO
                         string expandedName = null;
                         WatchedDirectory associatedDirectoryEntry = null;
 
-                        if (nextEvent.wd != -1) // wd is -1 for events like IN_Q_OVERFLOW that aren't tied to a particular watch descriptor
+                        // An overflow event means that we can't trust our state without restarting since we missed events and 
+                        // some of those events could be a directory create, meaning we wouldn't have added the directory to the 
+                        // watch and would not provide correct data to the caller.
+                        if ((mask & (uint)Interop.libc.NotifyEvents.IN_Q_OVERFLOW) != 0)
+                        {
+                            // Notify the caller of the error and, if the includeSubdirectories flag is set, restart to pick up any
+                            // potential directories we missed due to the overflow.
+                            watcher.NotifyInternalBufferOverflowEvent();
+                            if (_includeSubdirectories)
+                            {
+                                watcher.Restart();
+                            }
+                            break;
+                        }
+                        else
                         {
                             // Look up the directory information for the supplied wd
                             lock (SyncObj)
@@ -543,15 +567,11 @@ namespace System.IO
                         }
 
                         const Interop.libc.NotifyEvents switchMask =
-                            Interop.libc.NotifyEvents.IN_Q_OVERFLOW | Interop.libc.NotifyEvents.IN_IGNORED |
-                            Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_DELETE |
+                            Interop.libc.NotifyEvents.IN_IGNORED |Interop.libc.NotifyEvents.IN_CREATE | Interop.libc.NotifyEvents.IN_DELETE |
                             Interop.libc.NotifyEvents.IN_ACCESS | Interop.libc.NotifyEvents.IN_MODIFY | Interop.libc.NotifyEvents.IN_ATTRIB |
                             Interop.libc.NotifyEvents.IN_MOVED_FROM | Interop.libc.NotifyEvents.IN_MOVED_TO;
                         switch ((Interop.libc.NotifyEvents)(mask & (uint)switchMask))
                         {
-                            case Interop.libc.NotifyEvents.IN_Q_OVERFLOW:
-                                watcher.NotifyInternalBufferOverflowEvent();
-                                break;
                             case Interop.libc.NotifyEvents.IN_CREATE:
                                 watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Created, expandedName);
                                 break;
@@ -573,9 +593,35 @@ namespace System.IO
                                 watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Changed, expandedName);
                                 break;
                             case Interop.libc.NotifyEvents.IN_MOVED_FROM:
+                                // We need to check if this MOVED_FROM event is standalone - meaning the item was moved out
+                                // of scope. We do this by checking if we are at the end of our buffer (meaning no more events) 
+                                // and if there is data to be read by polling the fd. If there aren't any more events, fire the
+                                // deleted event; if there are more events, handle it via next pass. This adds an additional
+                                // edge case where we get the MOVED_FROM event and the MOVED_TO event hasn't been generated yet
+                                // so we will send a DELETE for this event and a CREATE when the MOVED_TO is eventually processed.
+                                if (_bufferPos == _bufferAvailable)
+                                {
+                                    bool success = false;
+                                    Interop.libc.PollFlags resultFlags;
+                                    _inotifyHandle.DangerousAddRef(ref success);
+                                    Debug.Assert(success, "Failed to add-ref inotify handle");
+                                    int result = Interop.libc.poll(_inotifyHandle.DangerousGetHandle().ToInt32(), Interop.libc.PollFlags.POLLIN, 0, out resultFlags);
+                                    _inotifyHandle.DangerousRelease();
+
+                                    // If we error or don't have any signaled handles, send the deleted event
+                                    if (result <= 0)
+                                    {
+                                        // There isn't any more data in the queue so this is a deleted event
+                                        watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Deleted, expandedName);
+                                        break;
+                                    }
+                                }
+
+                                // We will set these values if the buffer has more data OR if the poll call tells us that more data is available.
                                 previousEventName = expandedName;
                                 previousEventParent = isDir ? associatedDirectoryEntry : null;
                                 previousEventCookie = nextEvent.cookie;
+
                                 break;
                             case Interop.libc.NotifyEvents.IN_MOVED_TO:
                                 if (previousEventName != null)
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
@@ -118,4 +118,142 @@ public partial class FileSystemWatcher_4000_Tests
         },
         NotifyFilters.DirectoryName | NotifyFilters.LastAccess | NotifyFilters.FileName);
     }
+
+    [Fact]
+    public static void FileSystemWatcher_Changed_FileDataChange()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        {
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+
+            // Attach the FSW to the existing structure
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.NotifyFilter = NotifyFilters.FileName | NotifyFilters.Size;
+
+            using (var file = File.Create(Path.Combine(dir.Path, "testfile.txt")))
+            {
+                watcher.EnableRaisingEvents = true;
+
+                // Change the nested file and verify we get the changed event
+                byte[] bt = new byte[4096];
+                file.Write(bt, 0, bt.Length);
+                file.Flush();
+            }
+
+            Utility.ExpectEvent(eventOccured, "file changed");
+        }
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public static void FileSystemWatcher_Changed_PreSeededNestedStructure(bool includeSubdirectories)
+    {
+        // Make a nested structure before the FSW is setup
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        using (var dir1 = Utility.CreateTestDirectory(Path.Combine(dir.Path, "dir1")))
+        using (var dir2 = Utility.CreateTestDirectory(Path.Combine(dir1.Path, "dir2")))
+        using (var dir3 = Utility.CreateTestDirectory(Path.Combine(dir2.Path, "dir3")))
+        {
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+            byte[] bt = new byte[4096];
+
+            using (var file = File.Create(Path.Combine(dir3.Path, "testfile.txt")))
+            {
+                // Attach the FSW to the existing structure
+                watcher.Path = Path.GetFullPath(dir.Path);
+                watcher.Filter = "*";
+                watcher.NotifyFilter = NotifyFilters.Attributes | NotifyFilters.CreationTime | NotifyFilters.DirectoryName | NotifyFilters.FileName | NotifyFilters.LastAccess | NotifyFilters.LastWrite | NotifyFilters.Security | NotifyFilters.Size;
+                watcher.IncludeSubdirectories = includeSubdirectories;
+                watcher.EnableRaisingEvents = true;
+
+                // Change the nested file and verify we get (or do not get, depending on the parameter) the changed event 
+                file.Write(bt, 0, bt.Length);
+                file.Flush();
+            }
+
+            if (includeSubdirectories)
+                Utility.ExpectEvent(eventOccured, "file changed");
+            else
+                Utility.ExpectNoEvent(eventOccured, "file changed");
+
+            // Stop the FSW
+            watcher.EnableRaisingEvents = false;
+
+            using (var file2 = File.Create(Path.Combine(dir3.Path, "testfile2.txt")))
+            {
+                // Start the FSW, write to the file, expect (or don't get, depending on the parameter) to pick up the write
+                watcher.EnableRaisingEvents = true;
+                file2.Write(bt, 0, bt.Length);
+                file2.Flush();
+            }
+
+            if (includeSubdirectories)
+                Utility.ExpectEvent(eventOccured, "second file change");
+            else
+                Utility.ExpectNoEvent(eventOccured, "second file change");
+        }
+    }
+
+    [Fact, ActiveIssue(1477, PlatformID.Windows)]
+    public static void FileSystemWatcher_Changed_SymLinkFileDoesntFireEvent()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var watcher = new FileSystemWatcher())
+        {
+            AutoResetEvent are = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+
+            // Setup the watcher
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.NotifyFilter = NotifyFilters.FileName | NotifyFilters.Size;
+
+            using (var file = Utility.CreateTestFile(Path.GetTempFileName()))
+            {
+                Utility.CreateSymLink(file.Path, Path.Combine(dir.Path, "link"), false);
+                watcher.EnableRaisingEvents = true;
+
+                // Changing the temp file should not fire an event through the symlink
+                byte[] bt = new byte[4096];
+                file.Write(bt, 0, bt.Length);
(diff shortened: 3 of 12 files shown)


sokket:
@stephentoub 

cc: @piotrpMSFT

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -476,7 +478,19 @@ private void ProcessEvents()
...
+                        // Two events take priority: Overflow and SymLinks. An overflow event means that we can't trust 
+                        // our state without restarting since we missed events and some of those events could be a 
+                        // directory create, meaning we wouldn't have added the directory to the watch and would not provide
+                        // correct data to the caller. The other event, SymLinks, should be filtered out due to the FSW contract.
+                        if ((mask & (uint)Interop.libc.NotifyEvents.IN_Q_OVERFLOW) != 0)
+                        {
+                            // Notify the caller of the error and restart to pick up any potential directories we missed due
+                            // to the overflow.
+                            watcher.NotifyInternalBufferOverflowEvent();
+                            watcher.Restart();
```
stephentoub: We don't restart like this on Windows, and it's not clear to me why we're doing it here.  Couldn't we miss additional events we wouldn't otherwise miss during the time between the stopping and restarting?  And even if we do want to do this restart, shouldn't it only be if IncludeSubdirectories, since the justification in the comment above implies we'd be doing this in case we missing a directory creation notification?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -476,7 +478,19 @@ private void ProcessEvents()
                         string expandedName = null;
                         WatchedDirectory associatedDirectoryEntry = null;
 
-                        if (nextEvent.wd != -1) // wd is -1 for events like IN_Q_OVERFLOW that aren't tied to a particular watch descriptor
+                        // Two events take priority: Overflow and SymLinks. An overflow event means that we can't trust 
+                        // our state without restarting since we missed events and some of those events could be a 
+                        // directory create, meaning we wouldn't have added the directory to the watch and would not provide
+                        // correct data to the caller. The other event, SymLinks, should be filtered out due to the FSW contract.
```
stephentoub: Why is the comment discussing prioritizing symlink-related events when the code doesn't actually do anything related to them?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -492,6 +506,20 @@ private void ProcessEvents()
                                 }
                             }
                             expandedName = associatedDirectoryEntry.GetPath(true, nextEvent.name);
+
+                            // inotify picks up changes to symlinks and reports them as file changes, which breaks the contract. We need
```
stephentoub: How does that break the contract?  Aren't we excluding symlink events with the IN_DONT_FOLLOW flag, such that if an event does occur on a symlink, it's for the link file itself rather than for the target, and in that case, isn't it a valid event?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -492,6 +506,20 @@ private void ProcessEvents()
...
+                            // to filter out any symlinks and not report them; however, we also need to go through the logic below incase 
+                            // the OVERFLOW flag is set, meaning the caller will need to do a rescan. LStat should not open a file 
+                            // descriptor on the file / symlink so it should not cause a huge perf loss.
+                            Interop.NativeIO.FileStats stats = default(Interop.NativeIO.FileStats);
+                            int result = Interop.NativeIO.LStat(System.IO.Path.Combine(watcher.Path, expandedName), out stats);
```
stephentoub: This is now doing an extra syscall and an extra string allocation per event.  I've not measured, but that feels like it could add enough overhead to cause us to miss more events if we're close to a buffer overflow situation.  That in turn would cause us to restart (with the logic done earlier), which would in turn cause us to miss even more events.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Changed);
+            byte[] bt = new byte[4096];
+
+            using (FileStream file = File.Create(Path.Combine(dir3.Path, "testfile.txt")))
+            {
+                // Attach the FSW to the existing structure
+                watcher.Path = Path.GetFullPath(dir.Path);
+                watcher.Filter = "*";
+                watcher.NotifyFilter = NotifyFilters.Attributes | NotifyFilters.CreationTime | NotifyFilters.DirectoryName | NotifyFilters.FileName | NotifyFilters.LastAccess | NotifyFilters.LastWrite | NotifyFilters.Security | NotifyFilters.Size;
+                watcher.IncludeSubdirectories = true;
```
stephentoub: Do we have any tests for a similarly nested case but where IncludeSubdirectories is false and we validate that no event is fired for the nested directory/file?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+                watcher.EnableRaisingEvents = true;
+
+                // Change the nested file and verify we get the changed event
+                file.Write(bt, 0, bt.Length);
+                file.Flush();
+            }
+
+            Utility.ExpectEvent(eventOccured, "file changed", 10000);
```
stephentoub: Why such a long timeout?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+                watcher.Path = Path.GetFullPath(dir.Path);
+                watcher.Filter = "*";
+                watcher.NotifyFilter = NotifyFilters.Attributes | NotifyFilters.CreationTime | NotifyFilters.DirectoryName | NotifyFilters.FileName | NotifyFilters.LastAccess | NotifyFilters.LastWrite | NotifyFilters.Security | NotifyFilters.Size;
+                watcher.EnableRaisingEvents = true;
+
+                // Change the nested file and verify we get the changed event
+                byte[] bt = new byte[8000];
+                file.Write(bt, 0, bt.Length);
+                file.Flush();
+            }
```
stephentoub: Does the directory created by CreateTestDirectory delete itself and all contained files when 'dir' is disposed?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+            // Setup the watcher
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.NotifyFilter = NotifyFilters.FileName | NotifyFilters.Size;
+            watcher.EnableRaisingEvents = true;
+
+            // Changing the temp file should not fire an event through the symlink
+            byte[] bt = new byte[4096];
+            file.Write(bt, 0, bt.Length);
+            file.Flush();
```
stephentoub: The previous test I believe required closing the file being written to in order to trigger the event; since we're not closing it here, seems like there could be a false negative: even if the component were incorrectly raising events for the linked file's changes, those changes wouldn't have actually occurred yet, so we won't pick up any on the symlink either.  Seems like you might actually want two different watchers for this test: one monitoring the target file and one monitoring the symlink, verifying that an event fires for the target but not for the symlink.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+                watcher.EnableRaisingEvents = true;
+                file2.Write(bt, 0, bt.Length);
+                file2.Flush();
+            }
+
+            Utility.ExpectEvent(eventOccured, "second file change");
+        }
+    }
+
+    [Fact, ActiveIssue("1477", PlatformID.Windows)]
```
stephentoub: An overload of ActiveIssue's ctor accepts Int32, so you shouldn't need the quotes around the issue number.

stephentoub: That said, why is this linked to 1477?  The test is testing symlinks, which are Unix-specific, so shouldn't this test be [PlatformSpecific(PlatformID.AnyUnix)]?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+
+            // Changing the temp file should not fire an event through the symlink
+            byte[] bt = new byte[4096];
+            file.Write(bt, 0, bt.Length);
+            file.Flush();
+            Utility.ExpectNoEvent(are, "symlink'd file change");
+        }
+    }
+
+    [Fact, ActiveIssue("1477", PlatformID.Windows)]
```
stephentoub: Same question about the ActiveIssue and symlinks.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
```diff
@@ -118,4 +118,128 @@ public static void FileSystemWatcher_Changed_FileInNestedDirectory()
...
+    public static void FileSystemWatcher_Changed_SymLinkFolderDoesntFireEvent()
+    {
+        using (var dir = Utility.CreateTestDirectory())
+        using (var tempDir = Utility.CreateTestDirectory(Path.Combine(Path.GetTempPath(), "FooBar")))
+        using (var file = Utility.CreateTestFile(Path.Combine(tempDir.Path, "Test")))
+        using (var watcher = new FileSystemWatcher())
+        {
+            // Create the symlink first
+            Utility.CreateSymLink(Path.GetFullPath(file.Path), Path.Combine(dir.Path, "link"), false);
```
stephentoub: I'm probably just missing it, but I'm not understanding what this test is meant to be testing.  The watcher is set up to watch dir.Path, but the symlink is being created for a file in a temporary directory that's unrelated to dir.Path.  From the test's name, I was expecting to see this test doing the same thing as the previous test, expect with a symlink to a directory rather than to a file.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
             }
         });
     }
+
+    // This can potentially fail on Windows, depending on where the test is run from, due to 
+    // the MAX_PATH limitation in Windows. When issue 645 is closed, this shouldn't be a problem
+    [Fact, ActiveIssue("645", PlatformID.Windows)]
```
stephentoub: Nit: same quotes comment

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
...
+            }
+
+            // Put a file at the very bottom and expect it to raise an event
+            using (var file = new TemporaryTestFile(Path.Combine(lst[lst.Count - 1].Path, "temp file")))
+            {
+                Utility.ExpectEvent(eventOccured, "temp file create");
+            }
+
+            // Cleanup
+            foreach (TemporaryTestDirectory d in lst)
```
stephentoub: Consider putting this cleanup in a finally so that the directories are cleaned up even if there's a failed assert that throws.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
...
+            for (int i = 1; i < 20; i++)
+            {
+                lst.Add(new TemporaryTestDirectory(Path.Combine(lst[i - 1].Path, String.Format("dir{0}", i))));
+                Utility.ExpectEvent(eventOccured, lst[i].Path + " create");
+            }
+
+            // Put a file at the very bottom and expect it to raise an event
+            using (var file = new TemporaryTestFile(Path.Combine(lst[lst.Count - 1].Path, "temp file")))
+            {
+                Utility.ExpectEvent(eventOccured, "temp file create");
```
stephentoub: Are we validating anywhere for any/all of these tests that the path information included in the raised events is correct?  Since we're manually constructing those paths in the Linux implementation in particular, that'd be a good thing to verify.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
...
+            }
+
+            // Cleanup
+            foreach (TemporaryTestDirectory d in lst)
+                d.Dispose();
+        }
+    }
+
+    [Fact, ActiveIssue("1477", PlatformID.Windows)]
```
stephentoub: Same question about PlatformSpecific instead of ActiveIssue.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
...
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created);
+
+            // put everything in our own directory to avoid collisions
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.EnableRaisingEvents = true;
+
+            // Make the symlink in our path (to the temp file) and make sure no event is raised
+            Utility.CreateSymLink(Path.GetFullPath(temp.Path), Path.Combine(dir.Path, Path.GetFileName(temp.Path)), false);
+            Utility.ExpectNoEvent(eventOccured, "symlink created");
```
stephentoub: As I noted earlier, I actually question whether this is the behavior we want.  I would assume we actually would want an event here.  If you enumerate the directory with Directory.EnumerateFiles, for example, you'll get back one more entry after this, for the link, and so I'd expect that's something someone would want to be notified of.  It's then up to them whether they choose to follow it in response to the event.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -123,4 +124,87 @@ public static void FileSystemWatcher_Created_FileCreatedInNestedDirectory()
...
+        {
+            AutoResetEvent eventOccured = Utility.WatchForEvents(watcher, WatcherChangeTypes.Created);
+
+            // put everything in our own directory to avoid collisions
+            watcher.Path = Path.GetFullPath(dir.Path);
+            watcher.Filter = "*";
+            watcher.EnableRaisingEvents = true;
+
+            // Make the symlink in our path (to the temp folder) and make sure no event is raised
+            Utility.CreateSymLink(Path.GetFullPath(temp.Path), Path.Combine(dir.Path, Path.GetFileName(temp.Path)), true);
```
stephentoub: Ditto.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
```diff
@@ -101,4 +101,130 @@ public static void FileSystemWatcher_Renamed_FileInNestedDirectory()
             }
         });
     }
+
+    [Fact]
+    [ActiveIssue(2545)]
```
stephentoub: This issue has been closed.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
```diff
@@ -101,4 +101,130 @@ public static void FileSystemWatcher_Renamed_FileInNestedDirectory()
...
+    [PlatformSpecific(PlatformID.Windows | PlatformID.OSX)]
```
stephentoub: Why not Linux?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
```diff
@@ -101,4 +101,130 @@ public static void FileSystemWatcher_Renamed_FileInNestedDirectory()
...
+            using (var dir1 = new TemporaryTestDirectory(Path.Combine(dir.Path, "dir1")))
+            {
+                Utility.ExpectEvent(createdOccured, "dir1 created", Utility.WaitForCreationTimeoutInMs);
+
+                using (var dir2 = new TemporaryTestDirectory(Path.Combine(dir1.Path, "dir2")))
+                {
+                    Utility.ExpectEvent(createdOccured, "dir2 created", Utility.WaitForCreationTimeoutInMs);
+
+                    // Don't use the TemporaryTestFile class since it holds an open stream on the file so we can't do a move
+                    using (FileStream fs = File.Create(Path.Combine(dir2.Path, "test file"))) { };
```
stephentoub: This could just be:
```C#
File.Create(...).Dispose();
```
instead of:
```C#
using (FileStream fs = File.Create(...)) { }
```

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
```diff
@@ -101,4 +101,130 @@ public static void FileSystemWatcher_Renamed_FileInNestedDirectory()
...
+                }
+            }
+        }
+    }
+
+    [Fact]
+    [ActiveIssue(2545)]
+    [PlatformSpecific(PlatformID.Windows | PlatformID.OSX)]
+    // Note: Can't use the TestNestedDirectoriesHelper since we need access to the root
+    public static void FileSystemWatcher_Moved_NestedDirectoryRootWithoutSubdirectoriesFlag()
```
stephentoub: Same questions as above.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: We only need to restart on Linux because on Windows and OSX we only watch the root directory and we are told automatically about the subdirectories; on Linux, we need to watch the subdirectories manually, so if we got an overflow then we need to restart in order to do a full enumeration of the file system and add watches on all subfolders. You brought up a good point in that we should only do this when the bit is set to watch subdirectories

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Sorry, stale comment from an earlier revision. Fixed

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Good catch, removing this part which also should resolve your comment below about the extra alloc and syscall

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Resolved via above

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Changed this test to be a theory with InlineData to test both the positive and negative cases

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Stale parameter...good catch, removed it

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Yes, the Dispose function calls Directory.Delete and specifies to delete all children as well

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Good catch, but I think I can simplify it by using the pattern I used in the other tests; aka, using File.Create in a using, writing to the file there, then verifying the event (or lack thereof) outside of the using...that should cause the handle to be closed and all events to fire (or not).

Does that make sense?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Changed.cs
sokket: Good catch, forgot to change the link target to be the directory

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Good idea

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: See first comment for fix

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Sounds good, fixed

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
sokket: Fixed

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
sokket: Thanks

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
sokket: This was originally not linux since you can't do a Directory.Move on default Ubuntu since the system Temp dir is on a different mount point; I'll switch this to not use the system Temp and use a test-specific temp so this can be enabled on Linux

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
sokket: Fixed

sokket:
@stephentoub using this iteration as a test pass since my test machines need to be paved and reimaged. Not all comments have necessarily been fixed yet :)

stephentoub:
Thanks, Jon.  Just let me know when I should take another look.

sokket:
I didn't see these failures locally...investigating

sokket:
Merging with Ian's changes and running tests again

src/Common/src/Interop/Unix/libc/Interop.poll.cs
```diff
@@ -0,0 +1,68 @@
...
+            POLLPRI     = 0x0002,       /* OOB/Urgent readable data */
+            POLLOUT     = 0x0004,       /* file descriptor is writeable */
+            POLLRDNORM  = 0x0040,       /* non-OOB/URG data available */
+            POLLWRNORM  = POLLOUT,      /* no write type differentiation */
+            POLLRDBAND  = 0x0080,       /* OOB/Urgent readable data */
+            POLLWRBAND  = 0x0100,       /* OOB/Urgent data can be written */
+            POLLERR     = 0x0008,       /* some poll error occurred */
+            POLLHUP     = 0x0010,       /* file descriptor was "hung up" */
+            POLLNVAL    = 0x0020,       /* requested events "invalid" */
+        }
```
stephentoub: Are all of these values the same across Linux and OSX?

sokket: The one that I'm using is, the rest are kind of irrelevant for now so I'll remove them. If they are needed down the road, we can adjust then

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -573,9 +593,33 @@ private void ProcessEvents()
                                 watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Changed, expandedName);
                                 break;
                             case Interop.libc.NotifyEvents.IN_MOVED_FROM:
-                                previousEventName = expandedName;
-                                previousEventParent = isDir ? associatedDirectoryEntry : null;
-                                previousEventCookie = nextEvent.cookie;
+                                // We need to check if this MOVED_FROM event is standalone - meaning the item was moved out
+                                // of scope. We do this by checking if we are at the end of our buffer (meaning no more events) 
+                                // and if there is data to be read by polling the fd. If there aren't any more events, fire the
+                                // deleted event; if there are more events, handle it via next pass.
```
stephentoub: Maybe add an additional comment about this potentially missing the case where the FROM event arrived and a TO event was coming but hasn't arrived yet. In that case we'll end up raising a delete + create instead of a rename, but we already do that in some other cases (e.g. if there are intervening events between the FROM and TO), and this coded behavior is least worst of the options, as it means we'll always raise an event for a FROM even if it's the last one for a while.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -573,9 +593,33 @@ private void ProcessEvents()
...
+                                if (_bufferPos == _bufferAvailable)
+                                {
+                                    Interop.libc.PollFlags resultFlags;
+                                    int result = Interop.libc.poll(_inotifyHandle.DangerousGetHandle().ToInt32(), Interop.libc.PollFlags.POLLIN, 0, out resultFlags);
```
stephentoub: Why is it ok to call DangerousGetHandle() here without corresponding AddRef/ReleaseRef calls?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -573,9 +593,33 @@ private void ProcessEvents()
...
+                                    {
+                                        // There isn't any more data in the queue so this is a deleted event
+                                        watcher.NotifyFileSystemEventArgs(WatcherChangeTypes.Deleted, expandedName);
+                                    }
+                                }
+                                else
+                                {
+                                    previousEventName = expandedName;
+                                    previousEventParent = isDir ? associatedDirectoryEntry : null;
+                                    previousEventCookie = nextEvent.cookie;
```
stephentoub: This code is now duplicated in two places.  Can we avoid that? e.g. pseudo-code
```
if (AtEndOfBuffer() && !EventWaitingOnDescriptor())
{
    RaiseDeletedEvent();
}
else
{
    previousEvent* = ...;
}
```
?

src/Common/src/Interop/Unix/libc/Interop.poll.cs
stephentoub: Ok, thanks.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Good idea, added

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: Added the AddRef and Release calls

stephentoub: Ok, thanks.  I didn't know if there was a reason it was ok to not have them (e.g. we were guaranteeing through some other means it couldn't be disposed of or finalized), but adding the add/ref usage is the safe call.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
sokket: I refactored this call to remove the second call to set the locals and also avoid having another function call to set them

sokket:
@stephentoub How does this iteration look?

stephentoub:
Thanks, Jon. Looks good :)

Issue #2537, opened 2015-07-27
Win32FileSystem File.Move is inconsistent with its Exception throws


ianhays:
There is an inconsistent behavior in File.Move with regards to non existent directories in paths that is present in both the desktop and core frameworks.

If the parent directory of the source file is nonexistent, a FileNotFoundException is thrown. If the parent directory of the dest file is nonexistent, a DirectoryNotFoundException is thrown. 
Example:
```
File.Create("test1").Dispose();
File.Move(Path.Combine(Directory.GetCurrentDirectory(), "test2", "test3"), "test4"); // throws a FileNotFoundException
File.Move("test1", Path.Combine(Directory.GetCurrentDirectory(), "test2", "test3")); // throws a DirectoryNotFoundException
```

The reason for this is that File.Move makes a call to "InternalExists" to check that the sourceFile exists in full and if it doesn't then a FileNotFoundException is thrown regardless of which part of the path caused the failure. Removing this check is simple and will leave path checking to the FileSystem implementation (which would be preferable), but it will need to be mirrored in the desktop as well or ignored/worked around on both.

ianhays:
cc @ericstj @stephentoub

ericstj:
This would be a breaking change on desktop.  We could quirk it but would need to have a good reason for the change.  Why do you suggest we change the behavior?

ianhays:
The change would be for consistency with the other function calls e.g. File.Copy(source, dest) that throw DirectoryFoundExceptions when the parent directory doesn't exist. Move seems to be an exception to this rule for no particular reason.

```
File.Create("test1").Dispose();
File.Copy(Path.Combine(Directory.GetCurrentDirectory(), "test2", "test3"), "test4"); // throws a DirectoryNotFoundException
File.Copy("test1", Path.Combine(Directory.GetCurrentDirectory(), "test2", "test3")); // throws a DirectoryNotFoundException
```

I'm not certain that consistency is a good enough reason for a breaking change (since either exception can be caught as an IOException), but better to be aware of it.

ianhays:
General consensus for this issue is that this isn't a big enough issue to cause a breaking change for this fix. For situations when the distinction is relevant, catching an IOException (from which both DirectoryNotFoundException and FileNotFoundException inherit) is the better alternative.

PR #2588, opened 2015-07-31
Modified the Unix implementation of FileMove and fixed the corresponding broken FileSystemWatcher tests

### LINKED ISSUE #2460: File.Move relies on the UnixFileSystem CopyFile method that overwrites only when it shouldn't

ianhays:
- Resolves #2460 
- Changed the way the UnixFileSystem does File moving; it was previously implemented using the rename syscall but now uses a combination of link and unlink. The side effect of this change is that the FileSystemWatcher will now treat File.Move as a deletion.
- Removed all tests from FileSystemWatcher that made calls to File.Move and consolidated them into a new file: FileSystemWatcher.MoveFile.cs. This new file separates tests into Windows and Unix categories and uses Theories to supply the input WatcherChangeType and whether a move should trigger an event for the given type. If we later decide to change the Unix behavior of MoveFile, changing the FileSystemWatcher tests will be a simple matter of flipping the boolean values for the affected tests.

### DIFF
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
@@ -1,0 +1,236 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Threading;
+using Xunit;
+
+public partial class FileSystemWatcher_4000_Tests
+{
+    #region WindowsTests
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, false)]
+    [InlineData(WatcherChangeTypes.Deleted, false)]
+    [InlineData(WatcherChangeTypes.Renamed, true)]
+    [PlatformSpecific(PlatformID.Windows)]
+    public static void Windows_File_Move_To_Same_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_SameDirectory(eventType, moveRaisesEvent);
+    }
+
+    [Theory]
+    [OuterLoop]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, false)]
+    [InlineData(WatcherChangeTypes.Deleted, false)]
+    [InlineData(WatcherChangeTypes.Renamed, true)]
+    [PlatformSpecific(PlatformID.Windows)]
+    public static void Windows_File_Move_To_Different_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_DifferentDirectory(eventType, moveRaisesEvent);
+    }
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, true)]
+    [InlineData(WatcherChangeTypes.Created, true)]
+    [InlineData(WatcherChangeTypes.Deleted, false)]
+    [InlineData(WatcherChangeTypes.Renamed, true)]
+    [PlatformSpecific(PlatformID.Windows)]
+    public static void Windows_File_Move_In_Nested_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_NestedDirectory(eventType, moveRaisesEvent);
+    }
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, false)]
+    [InlineData(WatcherChangeTypes.Deleted, false)]
+    [InlineData(WatcherChangeTypes.Renamed, true)]
+    [PlatformSpecific(PlatformID.Windows)]
+    public static void Windows_File_Move_With_Set_NotifyFilter_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_WithNotifyFilter(eventType, moveRaisesEvent);
+    }
+
+    #endregion
+
+    #region UnixTests
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, true)]
+    [InlineData(WatcherChangeTypes.Deleted, true)]
+    [InlineData(WatcherChangeTypes.Renamed, false)]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public static void Unix_File_Move_To_Same_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_SameDirectory(eventType, moveRaisesEvent);
+    }
+
+    [Theory]
+    [OuterLoop]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, true)]
+    [InlineData(WatcherChangeTypes.Deleted, true)]
+    [InlineData(WatcherChangeTypes.Renamed, false)]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public static void Unix_File_Move_To_Different_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_DifferentDirectory(eventType, moveRaisesEvent);
+
+    }
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, true)]
+    [InlineData(WatcherChangeTypes.Deleted, true)]
+    [InlineData(WatcherChangeTypes.Renamed, false)]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public static void Unix_File_Move_In_Nested_Directory_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_NestedDirectory(eventType, moveRaisesEvent);
+    }
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, false)]
+    [InlineData(WatcherChangeTypes.Created, false)]
+    [InlineData(WatcherChangeTypes.Deleted, true)]
+    [InlineData(WatcherChangeTypes.Renamed, false)]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public static void Unix_File_Move_With_Set_NotifyFilter_Triggers_Event(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
+        MoveAndCheck_WithNotifyFilter(eventType, moveRaisesEvent);
+    }
+
+    #endregion
+
+    #region TestHelpers
+
+    /// <summary>
+    /// Sets up watchers for the type given before performing a File.Move operation and checking for
+    /// events. If moveRaisesEvent is true, we make sure that the given event type is observed. If false,
+    /// we ensure that it is not observed.
+    /// 
+    /// This test will move the source file to a destination file in the same directory i.e. rename it
+    /// </summary>
+    private static void MoveAndCheck_SameDirectory(WatcherChangeTypes eventType, bool moveRaisesEvent)
+    {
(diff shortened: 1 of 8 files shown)


ianhays:
cc @stephentoub @sokket

ianhays:
I'm going to change the variable names of the Theories to be "eventType", "eventName", and "moveRaisesEvent" so that it is a bit easier to understand test failure from the error.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
```diff
@@ -0,0 +1,238 @@
...
+{
+    #region WindowsTests
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, "changed", false)]
+    [InlineData(WatcherChangeTypes.Created, "created", false)]
+    [InlineData(WatcherChangeTypes.Deleted, "deleted", false)]
+    [InlineData(WatcherChangeTypes.Deleted, "changed", false)]
+    [InlineData(WatcherChangeTypes.Renamed, "renamed", true)]
+    [InlineData(WatcherChangeTypes.Renamed, "changed", true)]
```
stephentoub: There's an entry that's Deleted - "changed" and one that's Renamed - "changed", whereas all the others match up.  Is that on purpose?  Also, why not just use type.ToString() as the typeName?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        MoveAndCheck_NestedDirectory(type, typeName, observedEvent);
+    }
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, "changed", false)]
+    [InlineData(WatcherChangeTypes.Created, "created", false)]
+    [InlineData(WatcherChangeTypes.Deleted, "deleted", false)]
+    [InlineData(WatcherChangeTypes.Renamed, "renamed", true)]
+    [InlineData(WatcherChangeTypes.Renamed, "changed", true)]
```
stephentoub: Another one where type != typeName

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
```diff
@@ -0,0 +1,238 @@
...
+
+    #endregion
+
+    #region UnixTests
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, "changed", false)]
+    [InlineData(WatcherChangeTypes.Created, "created", true)]
+    [InlineData(WatcherChangeTypes.Deleted, "deleted", true)]
+    [InlineData(WatcherChangeTypes.Deleted, "changed", true)]
```
stephentoub: Another.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -54,13 +54,9 @@ public override void MoveFile(string sourceFullPath, string destFullPath)
...
             // The desired behavior for Move(source, dest) is to not overwrite the destination file
             // if it exists. Since rename(source, dest) will replace the file at 'dest' if it exists,
-            // a check is added before rename is called. Note that this is a race condition. Possible
-            // better solutions would be to use link/unlink or platform specific functions instead of
-            // rename(source, dest). Regardless of the implementation, an IOException should be thrown if 
-            // 'dest' refers to a file that already exists.
-            if (File.Exists(destFullPath))
-                throw new IOException(SR.GetResourceString(SR.IO_AlreadyExists_Name, destFullPath));
-            while (Interop.libc.rename(sourceFullPath, destFullPath) < 0)
+            // link/unlink are used instead. Note that the act of unlinking will trigger a deletion
```
stephentoub: Does the act of linking not trigger a creation notification, too?

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -54,13 +54,9 @@ public override void MoveFile(string sourceFullPath, string destFullPath)
...
+            // notice for the FileSystemWatcher
+            while (Interop.libc.link(sourceFullPath, destFullPath) < 0)
```
stephentoub: There's some commented out code below.  Can that be deleted?  Or is there some other follow-up necessary for it?

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
```diff
@@ -0,0 +1,238 @@
...
+using System.IO;
+using System.Threading;
+using Xunit;
+
+public partial class FileSystemWatcher_4000_Tests
+{
+    #region WindowsTests
+
+    [Theory]
+    [InlineData(WatcherChangeTypes.Changed, "changed", false)]
```
stephentoub: How long do these tests take to run?  For the ones where ```false``` is being passed, I believe that means by default the utility will wait for 500ms before assuming the lack of an event in that time means one wasn't raised.  If we're now doing that tens of times, that'll add quite a bit of running time to the tests.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
ianhays: It was on purpose to match the old tests that were there. I wasn't too familiar with the inner workings of the FSW so I didn't want to risk removing any coverage or realistic test cases by changing them too much. 

There were several cases where they listened for one thing and then tested for a different thing.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
ianhays: It is related to #2537, I was leaving it in place until a decision was reached on how File.Move should handle an invalid source path, but it can be removed.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
ianhays: Yes, I'll expand this comment a bit to include that.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
ianhays: Yes, they are a bit slower as there are a number of false events. On my Windows machine, it takes 8 seconds to run all of the FSW tests. 

Do you think that is fine, or should we trim it down a bit? I think the SameDirectory, NestedDirectory, and  NotifyFilter tests need to stay but I'd be okay removing the Different_Directory tests if necessary. We could outerloop them.

stephentoub: Ideally we could trim it down a bit.  As it's written, I don't think xunit will parallelize the execution of any of these, but waits are an easy thing to parallelize, so if we could make xunit parallelize these (and fix up any issues with regards to how we're managing state, temporary files that could conflict, etc.), we could probably reduce the execution time significantly without sacrificing what tests we run.  If we can't do that, I think it'd be ok to push out a few of the lower priority tests to outer loop.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -54,13 +54,9 @@ public override void MoveFile(string sourceFullPath, string destFullPath)
...
             // The desired behavior for Move(source, dest) is to not overwrite the destination file
             // if it exists. Since rename(source, dest) will replace the file at 'dest' if it exists,
-            // a check is added before rename is called. Note that this is a race condition. Possible
-            // better solutions would be to use link/unlink or platform specific functions instead of
-            // rename(source, dest). Regardless of the implementation, an IOException should be thrown if 
-            // 'dest' refers to a file that already exists.
-            if (File.Exists(destFullPath))
-                throw new IOException(SR.GetResourceString(SR.IO_AlreadyExists_Name, destFullPath));
-            while (Interop.libc.rename(sourceFullPath, destFullPath) < 0)
+            // link/unlink are used instead. Note that the Unix FileSystemWatcher will treat a Move
```
sokket: This is a problem for the FSW...we shouldn't have functional differences for the same IO operation; callers of the FSW will respond differently to Create/Delete events then they will to Move events. Can we look into a different way of implementing this function that won't cause a Delete + Create?

stephentoub: This is discussed at https://github.com/dotnet/corefx/issues/2460.  Do you have a suggestion that wasn't considered there?  This seems like the least worst of the options.

sokket: If we have a Linux-specific option (ie renameat2, as stated in that bug) then we could simply use NSFileManager on OS X to move the file (or directory). NSFileManager will abort if the item already exists, so it would have the same results as Windows and Linux, see OS X documentation on NSFileManager here: https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSFileManager_Class/index.html#//apple_ref/occ/instm/NSFileManager/moveItemAtPath:toPath:error:

Would that work? This would prevent different events on different platforms

ianhays: Renameat2 has a dependency on a the Unix kernel 3.15. Ubuntu 14.04 is running on 3.13 and is the recommended distro in the dotnet wiki.

ianhays: >Renameat2 has a dependency on a the Unix kernel 3.15

Should be more specific and say *Linux kernel 3.15

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
ianhays: Yeah, these can't be parallelized in their current state because of the usage of the Utility.CreateTestDirectory() calls. I'm going to change it to use guids for the directory/file names and only have the watchers watch their owned directory and that should solve the file concurrency issues. 

I'm not familiar with xunit parallelization of Theories; does it happen automagically or do I need to specify something?

stephentoub: > I'm not familiar with xunit parallelization of Theories; does it happen automagically or do I need to specify something?

I've never tried.  I think you'd have to change it to say that tests on the same test class could be parallelized, but I'm not 100% sure that's even possible... they'd probably need to be changed to not be theories and not be on the same class.   Might not be worth it (but the fixes to allow them to be parallelizable are still a good idea).

ianhays: I modified the tests to be thread safe but they're still being run sequentially. The FSW before my changes had 53 tests and ran in ~2.9 seconds. After my changes there are 69 tests that take ~8.4 seconds to run.

Modifying the tests to not a bunch of facts instead of theories will only take a few minutes to do since the method bodies just delegate to helper methods. Hopefully xunit will be fine with that.

ianhays: I could not find a way in the xunit docs to make the tests within a class run in parallel: they have to be in different classes. I tested them in different classes and it does run considerably faster, but it's quite a bit less readable as a result. I'm not convinced it's worth the tradeoff; I'm going to revert to the theories and ```[outerloop]``` the different directory tests.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.MoveFile.cs
stephentoub: Could you please double-check these?  I don't understand, for example, what it means in this case to have Deleted + "changed".  The name is only used as part of the assert (and filename) as far as I can see, rather than being something that's actually tested.  If it's necessary, fine, but I'm not seeing it.  The old tests may just have been bogus in this regard.

stephentoub:
Thanks, Ian.  Might want to squash the commits a bit (up to you), but otherwise LGTM.

ianhays:
I'll squash them down to one commit and merge. Thanks for the feedback!

Issue #1070, opened 2015-03-02
Improve code coverage for System.Diagnostics.Process


stephentoub:
According to the current code coverage report, this library is only at ~53%. Our general goal is closer to 90%, though this will depend on the library. We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

pallavit:
@stephentoub Does the report mention the areas with low coverage? Please share the code coverage report. Thanks.

stephentoub:
Yes.  It'll be included in the daily code coverage run once #1069 is checked in.  If you want to see the details before then, easiest way would be to just checkout the branch from the PR and do a code coverage run locally.

stephentoub:
Pallavi, now that #1069 has been checked in, the code coverage information is available in the reports at http://dotnet-ci.cloudapp.net/job/dotnet_corefx_coverage_windows/lastBuild/Code_Coverage_Report/

ssg:
@stephentoub System.Diagnostics.Process isn't in the report?

stephentoub:
@ssg, if you look back a few days you'll see it.  It's been temporarily removed: https://github.com/dotnet/corefx/pull/1104

ssg:
@stephentoub yes this link works: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_coverage_windows/67/Code_Coverage_Report/ thanks!

stephentoub:
@ssg, great; glad you found it.

Priya91:
Tests have been added to improve coverage. PerformanceCounterLib is not covered, as it requires multi machine test infrastructure. The ProcessThreadCollection and ProcessModuleCollection are not covered since it's only an implementation of ICollection.

PRs: #2513, #2421, #2417, #2404, #2311, #2283, #2262

ssg:
Hi Lakshmi,

https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessModuleCollection.cs

The code for ProcessModuleCollection seems to be able to experience
regressions in the future regardless. "It's only an implementation of
ICollection" didn't seem a good justification for me. "Not a priorty" is
but not that.

Thanks,

Sedat

stephentoub:
> PerformanceCounterLib is not covered, as it requires multi machine test infrastructure

Can we not use a machine name of "127.0.0.1" to access the local machine "remotely"?

[...]

PR #2592, opened 2015-07-31
Add more process tests.

Priya91:
#1070

### DIFF
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
@@ -1,0 +1,139 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+using System.Linq;
+using Xunit;
+
+namespace System.Diagnostics.ProcessTests
+{
+    public class ProcessCollectionTests : ProcessTestBase
+    {
+        [Fact]
+        public void TestModuleCollectionBehavior()
+        {
+            ProcessModule[] mArray = _process.Modules.Cast<ProcessModule>().ToArray();
+
+            // Constructor
+            ProcessModuleCollection moduleCollection = new ProcessModuleCollection(mArray);
+
+            // Count
+            Assert.Equal(mArray.Count(), moduleCollection.Count);
+
+            // get_item, Contains, IndexOf
+            for (int i = 0; i < mArray.Count(); i++)
+            {
+                Assert.Equal(mArray[i], moduleCollection[i]);
+                Assert.True(moduleCollection.Contains(mArray[i]));
+                Assert.Equal(i, moduleCollection.IndexOf(mArray[i]));
+            }
+
+            // CopyTo
+            ProcessModule[] moduleArray = new ProcessModule[moduleCollection.Count + 1];
+            moduleCollection.CopyTo(moduleArray, 1);
+            for (int i = 0; i < mArray.Count(); i++)
+            {
+                Assert.Equal(mArray[i], moduleArray[i + 1]);
+            }
+
+            Assert.Throws<ArgumentOutOfRangeException>(() => moduleCollection.CopyTo(moduleArray, -1));
+
+            // Explicit interface implementations
+            Assert.False(((ICollection)moduleCollection).IsSynchronized);
+            Assert.NotNull(((ICollection)moduleCollection).SyncRoot);
+
+            moduleArray = new ProcessModule[moduleCollection.Count];
+            ((ICollection)moduleCollection).CopyTo(moduleArray, 0);
+            Assert.Equal(moduleCollection.Cast<ProcessModule>().ToArray(), moduleArray);
+
+            // GetEnumerator
+            IEnumerator enumerator = moduleCollection.GetEnumerator();
+            Assert.Throws<InvalidOperationException>(() => enumerator.Current);
+
+            for (int i = 0; i < moduleCollection.Count; i++)
+            {
+                enumerator.MoveNext();
+                Assert.Equal(moduleCollection[i], enumerator.Current);
+            }
+        }
+
+        [Fact]
+        public void TestThreadCollectionBehavior()
+        {
+            ProcessThread[] tArray = _process.Threads.Cast<ProcessThread>().ToArray();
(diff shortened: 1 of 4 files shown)


Priya91:
@dotnet-bot test this please

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    public class ProcessCollectionTests : ProcessTestBase
+    {
+        [Fact]
+        public void TestModuleCollectionBehavior()
+        {
+            ProcessModule[] mArray = _process.Modules.Cast<ProcessModule>().ToArray();
+
+            ProcessModuleCollection moduleCollection = new ProcessModuleCollection(mArray);
+            Assert.True(moduleCollection.Count > 1);
```
stephentoub: Shouldn't the count equal the length of mArray?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            Assert.Equal(mArray[0], moduleCollection[0]);
```
stephentoub: Shouldn't we verify that every element in the module collection matches every corresponding element in mArray, not just the first?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
stephentoub: (Also, doesn't this suffer from the same problem we've hit elsewhere, that the mArray could be empty if we're at a bad point during the target Process' startup?)

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            Assert.True(moduleCollection.Contains(mArray[0]));
```
stephentoub: Same question; should we verify that more than just the first element can be found?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            ProcessModule[] moduleArray = new ProcessModule[moduleCollection.Count + 1];
+            moduleCollection.CopyTo(moduleArray, 1);
+            Assert.Equal(mArray[0], moduleArray[1]);
```
stephentoub: What about the rest of them?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            Assert.Equal(0, moduleCollection.IndexOf(mArray[0]));
+        }
```
stephentoub: What about the other members: GetEnumerator, ICollection.CopyTo, ICollection.IsSynchronized, ICollection.SyncRoot, InnerList, set_Item, etc.?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            Assert.Equal(0, threadCollection.IndexOf(t));
+
+            threadCollection.Remove(t);
+            Assert.Equal(-1, threadCollection.IndexOf(t));
+
+            threadCollection.Add(t);
+            Assert.Equal(threadCollection.Count - 1, threadCollection.IndexOf(t));
+        }
+    }
```
stephentoub: Same questions as for the module tests.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -544,6 +544,33 @@ public void TestGetProcessesByName()
             Assert.True(Process.GetProcessesByName(currentProcess.ProcessName, currentProcess.MachineName).Count() > 0, "TestGetProcessesByName001 failed");
         }
 
+        public static IEnumerable<object[]> GetTestProcess()
+        {
+            Process currentProcess = Process.GetCurrentProcess();
+            yield return new object[] { currentProcess, Process.GetProcessById(currentProcess.Id, "127.0.0.1") };
```
stephentoub: Great, glad this exercises what we want it to. Is there any way as part of the rest to verify that it is in fact using the remote code paths?

Priya91: NotSupportedException is thrown only when accessing MainModule property on remote process. On a process on the same machine, it is either null or a ProcessModule.

stephentoub: Ah, ok, sounds good.  Might be worth a comment.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessCollectionTests.cs
Priya91: InnerList and set_Item are not exposed for ProcessModule contract. I'll add tests for explicit interface implementations as well.

stephentoub: > I'll add tests for explicit interface implementations as well.

Great, thanks.

> InnerList and set_Item are not exposed in ProcessModuleCollection api contract.

Good to know.  InnerList should be changed to be private rather than protected, right?

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
Priya91: added in the main test method.

stephentoub:
Thanks, Priya.  LGTM.

PR #2596, opened 2015-08-01
Added Some Async Methods in XmlBaseWriter Required by WCF.

### LINKED ISSUE #2502: XmlUTF8TextWriter has not implement WriteStartElementAsync or WriteEndElementAsync
WCF's async operations need to call these methods. See https://github.com/dotnet/wcf/issues/221.

### LINKED ISSUE #2504: Add FlushAsync and WriteBase64Async in XmlDictionaryWriter
WCF needs to use FlushAsync and WriteBase64Async. See https://github.com/dotnet/wcf/issues/222.

shmao:
WCF want to use some of XmlDictionaryWriter's async methods, which were not yet implemented. The PR made these async methods available (in XmlBaseWriter) for WCF to use.

Fixes #2502 
Fixes #2504

### DIFF
--- a/src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
+++ b/src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
@@ -10,7 +10,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.Serialization;
 using System.Collections.Generic;
-
+using System.Threading.Tasks;
 
 namespace System.Xml
 {
@@ -64,6 +64,14 @@ namespace System.Xml
                 ThrowClosed();
 
             _writer.Flush();
+        }
+
+        public override Task FlushAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            return _writer.FlushAsync();
         }
 
         public override void Close()
@@ -348,6 +356,70 @@ namespace System.Xml
             }
         }
 
+        protected override Task WriteEndAttributeAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            if (_writeState != WriteState.Attribute)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlInvalidWriteState, "WriteEndAttribute", WriteState.ToString())));
+
+            return WriteEndAttributeAsyncImpl();
+        }
+
+        private async Task WriteEndAttributeAsyncImpl()
+        {
+            await FlushBase64Async().ConfigureAwait(false);
+            try
+            {
+                if (_isXmlAttribute)
+                {
+                    if (_attributeLocalName == "lang")
+                    {
+                        _nsMgr.AddLangAttribute(_attributeValue);
+                    }
+                    else if (_attributeLocalName == "space")
+                    {
+                        if (_attributeValue == "preserve")
+                        {
+                            _nsMgr.AddSpaceAttribute(XmlSpace.Preserve);
+                        }
+                        else if (_attributeValue == "default")
+                        {
+                            _nsMgr.AddSpaceAttribute(XmlSpace.Default);
+                        }
+                        else
+                        {
+                            throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentException(SR.Format(SR.XmlInvalidXmlSpace, _attributeValue)));
+                        }
+                    }
+                    else
+                    {
+                        // XmlTextWriter specifically allows for other localNames
+                    }
+                    _isXmlAttribute = false;
+                    _attributeLocalName = null;
+                    _attributeValue = null;
+                }
+
+                if (_isXmlnsAttribute)
+                {
+                    _nsMgr.AddNamespaceIfNotDeclared(_attributeLocalName, _attributeValue, null);
+                    _isXmlnsAttribute = false;
+                    _attributeLocalName = null;
+                    _attributeValue = null;
+                }
+                else
+                {
+                    await _writer.WriteEndAttributeAsync().ConfigureAwait(false);
+                }
+            }
+            finally
+            {
+                _writeState = WriteState.Element;
+            }
+        }
+
         public override void WriteComment(string text)
         {
             if (IsClosed)
@@ -456,10 +528,72 @@ namespace System.Xml
             element.LocalName = localName;
         }
 
+        private void PreStartElementAsyncCheck(string prefix, string localName, string ns, XmlDictionaryString xNs)
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            if (_documentState == DocumentState.Epilog)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlOnlyOneRoot)));
+            if (localName == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentNullException("localName"));
+            if (localName.Length == 0)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentException(SR.Format(SR.InvalidLocalNameEmpty), "localName"));
+            if (_writeState == WriteState.Attribute)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlInvalidWriteState, "WriteStartElement", WriteState.ToString())));
+        }
+
+        private async Task StartElementAndWriteStartElementAsync(string prefix, string localName, string namespaceUri)
+        {
+            prefix = await StartElementAsync(prefix, localName, namespaceUri, null).ConfigureAwait(false);
+            await _writer.WriteStartElementAsync(prefix, localName).ConfigureAwait(false);
+        }
+
+        private async Task<string> StartElementAsync(string prefix, string localName, string ns, XmlDictionaryString xNs)
+        {
+            await FlushBase64Async().ConfigureAwait(false);
+            await AutoCompleteAsync(WriteState.Element).ConfigureAwait(false);
+            Element element = EnterScope();
+            if (ns == null)
+            {
+                if (prefix == null)
+                    prefix = string.Empty;
+
+                ns = _nsMgr.LookupNamespace(prefix);
+
+                if (ns == null)
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentException(SR.Format(SR.XmlUndefinedPrefix, prefix), "prefix"));
+            }
+            else if (prefix == null)
+            {
+                prefix = _nsMgr.LookupPrefix(ns);
+
+                if (prefix == null)
+                {
+                    prefix = string.Empty;
+                    _nsMgr.AddNamespace(string.Empty, ns, xNs);
+                }
+            }
+            else
+            {
+                _nsMgr.AddNamespaceIfNotDeclared(prefix, ns, xNs);
+            }
+            element.Prefix = prefix;
+            element.LocalName = localName;
+
+            return prefix;
+        }
+
         public override void WriteStartElement(string prefix, string localName, string namespaceUri)
         {
             StartElement(ref prefix, localName, namespaceUri, null);
             _writer.WriteStartElement(prefix, localName);
+        }
+
+        public override Task WriteStartElementAsync(string prefix, string localName, string namespaceUri)
+        {
+            PreStartElementAsyncCheck(prefix, localName, namespaceUri, null);
+            return StartElementAndWriteStartElementAsync(prefix, localName, namespaceUri);
         }
 
         public override void WriteStartElement(string prefix, XmlDictionaryString localName, XmlDictionaryString namespaceUri)
@@ -489,6 +623,38 @@ namespace System.Xml
             {
                 Element element = _elements[_depth];
                 _writer.WriteEndElement(element.Prefix, element.LocalName);
+            }
+
+            ExitScope();
+            _writeState = WriteState.Content;
+        }
+
+        public override Task WriteEndElementAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            if (_depth == 0)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlInvalidDepth, "WriteEndElement", _depth.ToString(CultureInfo.InvariantCulture))));
+
+            return WriteEndElementAsyncImpl();
+        }
+
+        private async Task WriteEndElementAsyncImpl()
+        {
+            if (_writeState == WriteState.Attribute)
+                await WriteEndAttributeAsync().ConfigureAwait(false);
+
+            FlushBase64();
+            if (_writeState == WriteState.Element)
+            {
+                _nsMgr.DeclareNamespaces(_writer);
+                await _writer.WriteEndStartElementAsync(true).ConfigureAwait(false);
+            }
+            else
+            {
+                Element element = _elements[_depth];
+                await _writer.WriteEndElementAsync(element.Prefix, element.LocalName).ConfigureAwait(false);
             }
 
             ExitScope();
@@ -535,6 +701,11 @@ namespace System.Xml
             }
         }
 
+        private Task FlushElementAsync()
+        {
+            return _writeState == WriteState.Element ? AutoCompleteAsync(WriteState.Content) : Task.CompletedTask;
+        }
+
         protected void StartComment()
         {
             FlushElement();
@@ -549,6 +720,13 @@ namespace System.Xml
             FlushElement();
             if (_depth == 0)
                 throw System.Runtime.Serialization.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlIllegalOutsideRoot)));
+        }
+
+        protected async Task StartContentAsync()
+        {
+            await FlushElementAsync().ConfigureAwait(false);
+            if (_depth == 0)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlIllegalOutsideRoot)));
         }
 
         protected void StartContent(char ch)
@@ -610,10 +788,25 @@ namespace System.Xml
             _writeState = writeState;
         }
 
+        private async Task AutoCompleteAsync(WriteState writeState)
+        {
+            if (_writeState == WriteState.Element)
+            {
+                await EndStartElementAsync().ConfigureAwait(false);
+            }
+            _writeState = writeState;
+        }
+
         private void EndStartElement()
         {
             _nsMgr.DeclareNamespaces(_writer);
             _writer.WriteEndStartElement(false);
+        }
+
+        private Task EndStartElementAsync()
+        {
+            _nsMgr.DeclareNamespaces(_writer);
+            return _writer.WriteEndStartElementAsync(false);
         }
 
         public override string LookupPrefix(string ns)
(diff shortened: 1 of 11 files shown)


shmao:
@khdang @SGuyGe @mconnew @zhenlan

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -66,6 +66,14 @@ public override void Flush()
             _writer.Flush();
         }
 
+        public override Task FlushAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
```
stephentoub: Is this exception one that should propagate synchronously, or should it be encapsulated in the Task being returned?

stephentoub: Also, it looks like other Task-returning methods check whether there's currently another async operation in flight.  Do these need to do similar checks?

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -348,6 +356,65 @@ public override void WriteEndAttribute()
             }
         }
 
+        protected override async Task WriteEndAttributeAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
```
stephentoub: Here's an example where the exception is being stored into the returned Task (because this method is async).  We should be consistent, both with these additions and with what's already done in the library.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -348,6 +356,65 @@ public override void WriteEndAttribute()
...
+
+            if (_writeState != WriteState.Attribute)
+                throw System.Runtime.Serialization.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlInvalidWriteState, "WriteEndAttribute", WriteState.ToString())));
+
+            await FlushBase64Async();
```
stephentoub: All of the awaits should be done with ```.ConfigureAwait(false)```.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -348,6 +356,65 @@ public override void WriteEndAttribute()
             }
         }
 
+        protected override async Task WriteEndAttributeAsync()
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            if (_writeState != WriteState.Attribute)
+                throw System.Runtime.Serialization.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.XmlInvalidWriteState, "WriteEndAttribute", WriteState.ToString())));
```
stephentoub: There's already a ```using``` importing ```System.Runtime.Serialization```.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -535,6 +682,14 @@ protected void FlushElement()
             }
         }
 
+        protected async Task FlushElementAsync()
+        {
+            if (_writeState == WriteState.Element)
+            {
+                await AutoCompleteAsync(WriteState.Content);
+            }
+        }
```
stephentoub: It'd be more efficient if this were instead implemented like:
```C#
protected Task FlushElementAsync()
{
    return _writestate == WriteState.Element ?
        AutoCompleteAsync(WriteState.Content) :
        Task.CompletedTask;
}
```

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -1346,6 +1523,71 @@ public override void WriteBase64(byte[] buffer, int offset, int count)
...
+        {
+            if (IsClosed)
+                ThrowClosed();
+
+            if (buffer == null)
+                throw System.Runtime.Serialization.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentNullException("buffer"));
+
+            // Not checking upper bound because it will be caught by "count".  This is what XmlTextWriter does.
+            if (offset < 0)
+                throw System.Runtime.Serialization.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("offset", SR.Format(SR.ValueMustBeNonNegative)));
```
stephentoub: Argument exceptions like these should ideally propagate synchronously out of the call, but here they're getting wrapped in the returned Task.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
```diff
@@ -1376,6 +1618,14 @@ private void FlushBase64()
             }
         }
 
+        private async Task FlushBase64Async()
+        {
+            if (_trailByteCount > 0)
+            {
+                await FlushTrailBytesAsync();
+            }
+        }
```
stephentoub: This would be more efficient as:
```C#
private Task FlushBase64Async()
{
    return _trailByteCount > 0 ? 
        FlushTrailBytesAsync() :
        Task.CompletedTask;
}
```

src/System.Private.DataContractSerialization/src/System/Xml/XmlNodeWriter.cs
```diff
@@ -10,25 +10,41 @@
...
 using System.Collections.Generic;
-
+using System.Threading.Tasks;
 
 namespace System.Xml
 {
     internal abstract class XmlNodeWriter
     {
         public abstract void Flush();
+        public virtual Task FlushAsync()
```
stephentoub: Why are these virtual rather than abstract?  Since the type is internal, can't we ensure that all derived types override them appropriately?

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -111,6 +111,30 @@ protected byte[] GetBuffer(int count, out int offset)
             return _buffer;
         }
 
+        protected async Task<Tuple<byte[], int>> GetBufferAsync(int count)
```
stephentoub: I don't know whether it matters for the expected scenarios, but every call to this method is going to require allocating a tuple object; you could avoid that by using a struct type instead.

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -125,18 +149,38 @@ private void EnsureByte()
...
         {
             EnsureByte();
             _buffer[_offset++] = b;
         }
 
+        protected async Task WriteByteAsync(byte b)
+        {
+            await EnsureByteAsync();
+            _buffer[_offset++] = b;
+        }
```
stephentoub: This would be more efficient as something like:
```C#
protected Task WriteByteAsync(byte b)
{
    if (_offset >= bufferLength)
    {
        return FlushBufferAndWriteByteAsync(b);
    }
    else
    {
        _buffer[_offset++] = b;
        return Task.CompletedTask;
    }
}

private async Task FlushBufferAndWriteByteAsync(byte b)
{
    await EnsureByteAsync().ConfigureAwait(false);
    _buffer[_offset++] = b;
}
```

stephentoub:
cc: @krwq

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: You are right, I should throw an InvalidOperationException synchronously in that case.

shmao: It seems that none of XmlBaseWriter's methods checked async operation in flight because the type didn't have any async methods. Does this mean I should add this check in every public sync/async method?

stephentoub: > Does this mean I should add this check in every public sync/async method?

@krwq, what would you recommend here?  I'm not familiar with the expectations around the design for the checking of in-flight operations.

src/System.Private.DataContractSerialization/src/System/Xml/XmlNodeWriter.cs
shmao: Both XmlBinaryNodeWriter and XmlUTF8NodeWriter inherit from this class. We have no plan to add async operations for XmlBinaryNodeWriter for now.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: Changed. Thanks.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: Fixed. Thanks.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: I made these exception to be thrown synchronously.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: Added .ConfigureAwait(false) for all awaits.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
shmao: Removed.

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
shmao: Added a new struct for the result type.

shmao:
I updated the PR and added an async check wrapper on XmlDictionaryWriter which does checking of in-flight async operation. Also did changes based on the feedbacks.

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -419,12 +494,27 @@ protected virtual void FlushBuffer()
             }
         }
 
+        protected virtual async Task FlushBufferAsync()
+        {
+            if (_offset != 0)
+            {
+                await _stream.WriteAsync(_buffer, 0, _offset).ConfigureAwait(false);
+                _offset = 0;
+            }
```
mconnew: You can save some allocations by doing the following:
```csharp
if (_offset != 0)  
{
    var task = _stream.WriteAsync(_buffer, 0, _offset);
    _offset = 0;
    return task;
}
return Task.CompletedTask;
```
You could also think about removing the check for _offset!=0 and presume the caller has done this to avoid a method call/Task allocation unnecessarily. Replace it with a Debug.Assert to catch any code paths which don't do this.

shmao:
Hi, can you please have a look at the pull request? Thank you.

stephentoub:
@krwq, can you please review?

@shmao, who from WCF should review?

shmao:
I had added @mconnew from WCF to review the changes.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
krwq: I believe consistency should be our guideline so what we do in other libraries is what we should do (especially System.IO).

From a quick peek in the source code and the previous conversations (I already asked similar question before) I can see what you should do is this:
- if the arguments passed are invalid at the moment of passing (null argument, some negative length etc.) and it will be invalid even when you completely rewrite implementation then you should throw synchronously,
- If there is any slightest chance that it might be invalid only during the runtime (assume implementation may change) then throw asynchronously.

Should you do this in every public sync/async method? Depends: we should consider a risk of doing this change. I believe for the good assessment of compat risk I'd talk to @ericstj.

If the guidelines do not match what we currently have in the code and would make it inconsistent I'd start the design guidelines discussion if we should change everything or follow bad guidelines.

shmao:
@roncain , can you please help review the change? Thanks!

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseWriter.cs
stephentoub: Basically, yes.  Our guidance has been that any usage error (meaning the developer misused the method, typically by passing incorrect arguments) can/should propagate synchronously; everything else should go out through the returned Task.

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -111,6 +111,30 @@ protected byte[] GetBuffer(int count, out int offset)
...
+            DiagnosticUtility.DebugAssert(count >= 0 && count <= bufferLength, "");
+            int bufferOffset = _offset;
+            if (bufferOffset + count <= bufferLength)
+            {
+                offset = bufferOffset;
+            }
+            else
+            {
+                await FlushBufferAsync().ConfigureAwait(false);
+                offset = 0;
```
krwq: Should you throw here?

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -131,12 +155,37 @@ protected void WriteByte(byte b)
...
+
         protected void WriteByte(char ch)
         {
             DiagnosticUtility.DebugAssert(ch < 0x80, "");
             WriteByte((byte)ch);
         }
 
+        protected Task WriteByteAsync(char ch)
+        {
+            DiagnosticUtility.DebugAssert(ch < 0x80, "");
```
krwq: Nit: I know it is debug only but I'd add some meaning to the magic number

src/System.Private.DataContractSerialization/src/System/Xml/XmlUTF8TextWriter.cs
```diff
@@ -5,6 +5,7 @@
 using System.Runtime;
```
krwq: Why is there a private copy of XML here? Is there something we should add to System.Xml.* to remove it?

shmao: Sorry, which one is the private copy of XML?

krwq: Sorry - disregard my previous message, I thought this was copy of one of System.Xml.ReaderWriter files.

krwq:
LGTM

shmao:
@roncain @iamjasonp @mconnew Can someone from WCF please help review the change? We need this change to unblock 4 WCF issues. Thanks.

src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
```diff
@@ -14,6 +14,8 @@
 using SerializationTypes;
 
 using Xunit;
+using System.Threading.Tasks;
+using System.Threading;
```
iamjasonp: NIT: is this change necessary?

src/System.Runtime.Serialization.Xml/tests/XmlDictionaryWriterTest.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    public class AsyncMemoryStream : MemoryStream
+    {
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            await Task.Delay(10).ConfigureAwait(false);
+            await base.WriteAsync(buffer, offset, count, cancellationToken);
+        }
+    }
+}
```
iamjasonp: It'd be good to get an additional test to (on a basic level) ensure that the XmlDictionaryAsyncCheckWriter is indeed throwing when async behaviours happen at the same time.

src/System.Private.DataContractSerialization/src/System/Xml/XmlDictionaryAsyncCheckWriter.cs
```diff
@@ -0,0 +1,695 @@
...
+            {
+                throw new InvalidOperationException(SR.XmlAsyncIsRunningException);
+            }
+        }
+
+        private Task SetLastTask(Task task)
+        {
+            _lastTask = task;
+            return task;
+        }
```
iamjasonp: Other XmlWriters will throw when async operations are called and if XmlWriterSettings.Async is `false`, should we do the same here?

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
```diff
@@ -151,12 +200,39 @@ protected void WriteBytes(byte b1, byte b2)
...
+                return Task.CompletedTask;
+            }
+        }
+
+        private async Task FlushAndWriteBytesAsync(byte b1, byte b2)
+        {
+            await FlushBufferAsync().ConfigureAwait(false);
+            _buffer[_offset++] = b1;
+            _buffer[_offset++] = b2;
+        }
```
iamjasonp: Curious why we've done a special case with two bytes?

iamjasonp: Ah, this was hidden in the couple of lines above, there's a sync version of this too.

iamjasonp:
I'd like to see a couple more tests, especially those to test a negative case to know that `CheckAsync()` works as expected, but otherwise LGTM.

src/System.Private.DataContractSerialization/src/System/Xml/XmlDictionaryAsyncCheckWriter.cs
shmao: Yes, ideally we should do that. But XmlDictionaryWriter.CreateTextWriter does not take an XmlWriterSettings object as parameter, so we cannot set the async value for the writer. Therefore I didn't make the XmlWriterSettings.Async check here. Do you think this is ok?

src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
shmao: Yeah, just wanted to keep the async implementations consistent with the sync version.

src/System.Runtime.Serialization.Xml/tests/DataContractSerializer.cs
shmao: Nice catch. The new tests were added in this file. I forgot to remove these usings when I moved the tests to a new file.

src/System.Runtime.Serialization.Xml/tests/XmlDictionaryWriterTest.cs
shmao: Added a test for the negative case.

src/System.Private.DataContractSerialization/src/System/Xml/XmlDictionaryAsyncCheckWriter.cs
iamjasonp: Hmm, crap. I don't like it, actually - if we don't do this right the first time, it'll be a compat break to introduce the async switch later if we want to get back in line with the other readers/writers. @roncain, @mconnew, thoughts?

iamjasonp:
I think we need to close on whether or not CheckAsync should look at XmlWriterSettings.Async prior to merging this change.

jasonwilliams200OK:
> I think we need to close on whether or not CheckAsync should look at XmlWriterSettings.Async prior to merging this change. 

@iamjasonp, is the related discussion visible on GitHub? If not, would it be a good idea to merge this as the future PRs can bring prospective changes?

iamjasonp:
Much of the discussion isn't yet GitHub visible as there are implications on the desktop implementation of these APIs too. 

@shmao has been leading the discussion around this one - could you comment on where we are?

shmao:
There were several concerns regarding this PR, e.g. whether to make XmlDictionaryWriter respect XmlWriterSettings.  Now we are done with the discussion on this issue and we decided to take the changes.

 I rebased the commits. I will merge this PR once the new commit passed all checks.

stephentoub:
@dotnet-bot test this please

PR #2606, opened 2015-08-03
Add errno shim (+ existing shim cleanup precursor)

nguerrera:
1. (8d24e62) Rename System.IO.Native to System.Native. I've decided that anything provided by POSIX can go in just one bottom native lib. These libs are UNIX-specific and we aren't going to encounter a UNIX without files. :)

2. (62ee4aa) I've decided on some style changes to shim source code and applied them to System.Native before following the same style for the errno shims. I've continue to leave System.Security.Cryptography.Native alone for now as I'm waiting for feedback and convergence on the newer stuff first.

3. (f2a6aac) Add shims for converting error codes from platform-specific values to platform-agnostic values or vice versa and for invoking strerror_r with the GNU/POSIX divergence handled in the shim.

4. (b325287) Replace calls to Marshal.GetWin32Error and libc.strerror with calls to the new shims.

Some issues that I'm still contemplating and would like to hear others thoughts around:

* **Exceptions that carry error codes** (Win32Exception, IOException). 
 * I've kept the existing behavior of leaving the platform specific values in the error codes exposed by these exceptions. I could switch them to the PAL values, but that would make those externally observable and I'm not sure we want to document those as a public contract. It really depends if we intend for those error codes to
   * (a) just be extra diagnostic info in which case you can bring the context of the underlying platform to interpreting a log, debug session, etc. and converting the platform-specific code can be lossy if a non-standard error is returned
    * or (b) expect catchers to act programmatically on the error code in which case making them platform-agnostic is helpful...

* **Perf** 
 * I've checked that the switch codegen by clang is sufficiently optimized to a computed jump in the PAL -> Platform case and a jump table in the Platform -> PAL case, but there are other potential issues...
 * In order to allow exceptions to always carry the correct error message, even in the platform-specific case where we have no standard value, I've passed a (pal, platform) tuple (ErrorInfo struct) around. This is currently 64-bits but it's wasting most of those as in practice the errnos are small positive integers. There's nothing in POSIX that guarantees that they are anything but capable of being represented as ints and positive. On the PAL side, we choose the value and could use something smaller than int, but e.g. 40 bits is likely not much better than 64. I originally didn't think this would matter much guessing we'd only be on failure code paths, but there are several places in IO where we pay for this space (on the stack and in UnixFileSystem object instances) on successful code paths too.
 * I'm not currently set up to do perf testing on UNIX, so any tips on that would be much appreciated...

cc @stephentoub @ellismg @bartonjs


src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
```diff
@@ -6,11 +6,11 @@
...
+    internal static partial class System
     {
-        internal struct FileStats
+        internal struct FileStatus
         {
-            private FileStatsFlags Flags;
+            internal FileStatusFlags Flags;
             internal int Mode;
             internal int Uid;
             internal int Gid;
```
nguerrera: Note to self: forgot to rename the other members on the managed side.

src/System.Console/src/System.Console.csproj
```diff
@@ -173,4 +161,4 @@
     <None Include="project.json" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
-</Project>
+</Project>
```
nguerrera: Sigh. Final newline removal is done by VS.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -62,18 +62,17 @@ public override void MoveFile(string sourceFullPath, string destFullPath)
...
                 {
                     continue;
                 }
-                else if (errno == Interop.Errors.EXDEV) // link fails across devices / mount points
+                else if (errorInfo.Error == Interop.Error.EXDEV) // rename fails across devices / mount points
                 {
                     CopyFile(sourceFullPath, destFullPath, overwrite: false);
                     break;
                 }
-                //else if (errno == Interop.Errors.ENOENT && (!Directory.Exists(Path.GetDirectoryName(sourceFullPath)) || !Directory.Exists(Path.GetDirectoryName(destFullPath)))) // The parent directory for either source or dest can't be found
```
nguerrera: There was commented out code with no indication why and that really bothers me. Removed it assuming it's just not needed, but if there's something pending here, it should be filed as an issue and referenced as such. I'll do that if necessary if someone can enlighten me on the history here.

stephentoub: I commented on the same thing at https://github.com/dotnet/corefx/pull/2588#discussion-diff-36039386 on @ianhays's file system changes.  Please coordinate with him.

src/Native/System.Native/pal_errno.cpp
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#define _POSIX_C_SOURCE
```
nguerrera: Oops, this is not necessary and leftover from some attempts at getting the posix signature for strerror_r everywhere. In the end, I just used configure to detect which one we have, which is better because the GNU one is presumably more efficient for the fixed, recognized error case.

nguerrera:
@dotnet-bot test this please

Filed #2607 about intermittent deadlock in PLINQ on Linux. I think it's unrelated, but investigating to be sure.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
ianhays: That's my mistake and it can be removed. It was code relevant to the conversation in #2537 that I left in place pending a solution with the intention of referencing that issue in the comment trailing the line. Since it looks like we'll be keeping the existing behavior anyways it is further unnecessary and as such does not need to be kept/commented.

nguerrera:
@dotnet-bot test this please

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
```diff
@@ -32,7 +32,7 @@ internal static int inotify_rm_watch(int fd, int wd)
                 }
                 else
                 {
-                    System.Diagnostics.Debug.Fail("inotify_rm_watch failed with " + hr);
+                    global::System.Diagnostics.Debug.Fail("inotify_rm_watch failed with " + hr);
```
ellismg: Can we name things in a way such that we don't have to do this global:: dance?

nguerrera: Sure. I can use some help coming up with the right names.

Each shim library has three names associated with it.

1. The file name: e.g. System.Security.Cryptography.Native.(so|dylib)
2. The constant in Interop.Libraries: e.g. Interop.Libraries.CryptoInterop
3. The partial class inside Interop that exposes P/Invokes: Currently, Interop.NativeCrypto

Some things that are bothering about status quo that I was trying to address here:

* We have to type (3) at every call site, so shorter is better with all else equal and Native at the end is somewhat redundant with Interop at the beginning.

* (2) can be longer with less pain, and since it directly represents the file name, I think a 1:1 convention between constant and file name is better.

So I was trending towards:

* System.Native.so, Interop.Libraries.SystemNative, Interop.System.SomeExport
* System.Security.Cryptography.so, Interop.Libraries.SystemSecurityCryptographyNative, Interop.Cryptography.SomeExport

(We could also go with short names for the *.so files and then the existing conventions for e.g. libc work fine, but it's hard to come up with something systematic that works well with the long names. A lot of folks seemed to like the long names, so I'm not sure I want to revisit it again...)

nguerrera: How about we just allow abbreviations for * in Interop.*.SomeExport on the basis of having to specify it at a the call site. With a good 1:1 name in Interop.Libraries, a clarification is just one F12 away:

So we'd have:
Interop.Libraries.SystemNative
Interop.Libraries.SystemSecurityCryptographyNative
Interop.Crypto.GetX509Thumbprint(...)
Interop.Sys.Stat(...)

bartonjs: Having chosen "NativeCrypto" for the current P/Invokes into System.Security.Cryptography.Native.so, I like the shorter "Crypto", compared to burning 50 characters to say "P/Invoke". (Yeah, "Interop.Crypto" is still 14, but that's a savings of 36 characters!)

src/Native/System.Native/pal_errno.h
```diff
@@ -0,0 +1,160 @@
...
+
+#include <stdint.h>
+
+/**
+ * Error codes returned via ConvertErrno.
+ *
+ * Only the names (without the PAL_ prefix) are specified by POSIX.
+ *
+ * The values chosen below are simply assigned arbitrarily (originally
+ * in alphabetical order they appear in the spec, they can't change so
```
nguerrera: meant to say "*but* they can't..."

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
nguerrera: Are you also good with the longer names in Interop.Libraries? I think that's a good place to be explicit since it goes on its own line from the function name and parameters and actually represents the file name of the lib.

````
// Imports
internal partial class Interop {
    internal partial class Crypto {
            [DllImport(Interop.Libraries.SystemSecurityCryptographyNative)]
            internal static extern int GetX509Thumbprint(SafeX509Handle x509, byte[] buf, int cBuf); 
    }
    partial class Sys {
        [DllImport(Interop.Libraries.SystemNative, SetLastError=true, CharSet=CharSet.Ansi)]
        internal static extern int Stat(string path, out FileStatus output)]
    }
}

// Call sites
int ret = Interop.Sys.Stat(path, out status);
int ret = Interop.Crypto.GetX509Thumbprint(x509, buf, cBuf);
```

src/Common/src/Interop/Unix/Interop.Errors.cs
```diff
@@ -1,48 +1,166 @@
...
     /// <summary>Common Unix errno error codes.</summary>
-    internal static partial class Errors
+    internal enum Error
+    {
+        // These values were defined in src/Native/System.Native/fxerrno.h
+        //
+        // They compare against values obtaied via Interop.System.GetLastError() not Marshal.GetLastWin32Error()
+        // which obtains the raw errno that varies between unixes. The strong typing as an enum is meant to
+        // prevent confusing the two. Casting to or from int is suspect. Use GetLastErrorInfo() if you need to
+        // correlate these to the underlying platform values or obtain the correspodning error message.
```
nguerrera: typo: corresponding

src/Common/src/Interop/Unix/Interop.Errors.cs
```diff
@@ -1,48 +1,166 @@
...
+    // Represents a platform-agnostic Error and underlying platform-specific errno
+    internal struct ErrorInfo
+    {
+        internal readonly Error Error;
+        internal readonly int Errno;
+
+        internal ErrorInfo(Error error)
+        {
+            Error = error;
+            Errno = Interop.System.ConvertErrorPalToPlatform(error);
```
stephentoub: What would you think of exposing a function to retrieve the whole conversion table once, and then doing all of these conversions purely in managed without having to make a managed->native call each time?  I realize these are on an error path, so performance isn't as critical, but there are cases where things can fail that aren't actually failures, e.g. EINTR for I/O operations.

src/Common/src/Interop/Unix/Interop.Errors.cs
```diff
@@ -1,48 +1,166 @@
...
+        {
+            return new ErrorInfo(Marshal.GetLastWin32Error());
+        }
+
+        internal static unsafe string StrError(int platformErrno)
+        {
+            const int maxErrorMessageLength = 1024; // length long enough for most any Unix error
+            byte* buffer = stackalloc byte[maxErrorMessageLength];
+            IntPtr message = StrErrorR(platformErrno, buffer, maxErrorMessageLength);
+            return Marshal.PtrToStringAnsi(message);
```
stephentoub: I'm glad all that horrible gook I had for choosing between strerror_r's in managed goes away :)

src/Common/src/Interop/Unix/Interop.Errors.cs
```diff
@@ -1,48 +1,166 @@
...
+        // would not work on all platforms.
+        EOPNOTSUPP      = ENOTSUP,            // Operation not supported on socket
+        EWOULDBLOCK     = EAGAIN,             // Operation would block
+    }
+
+    // Represents a platform-agnostic Error and underlying platform-specific errno
+    internal struct ErrorInfo
+    {
+        internal readonly Error Error;
+        internal readonly int Errno;
```
stephentoub: The two names are so similar, it could be pretty easy to confuse them.  I realize that's one of the reasons we have the two different return types, but maybe ```RawErrno``` or something like that just to make it less attractive?

src/Common/src/Interop/Unix/Interop.IOErrors.cs
```diff
@@ -77,41 +78,41 @@ internal static Exception GetExceptionForIoErrno(int errno, string path = null,
...
-                return GetIOException(errno);
+                return GetIOException(errorInfo);
         }
     }
 
-    internal static Exception GetIOException(int errno)
+    internal static Exception GetIOException(Interop.ErrorInfo errorInfo)
     {
-        return new IOException(libc.strerror(errno), errno);
+        return new IOException(errorInfo.GetErrorMessage(), errorInfo.Errno);
```
stephentoub: I went back and forth on whether it should be erroInfo.Errno or errorInfo.Error, but I think you're right that it should be Errno.  Even though it's platform-specific, the platform's errno value will make more sense to a developer than what are effectively internal values used in Error.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
```diff
@@ -61,12 +61,12 @@ private static unsafe string getcwd(byte* ptr, int bufferSize)
 
             // Otherwise, if it failed due to the buffer being too small, return null;
             // for anything else, throw.
-            int errno = Marshal.GetLastWin32Error();
-            if (errno == Interop.Errors.ERANGE)
+            ErrorInfo errorInfo = Interop.System.GetLastErrorInfo();
```
stephentoub: My initial thought for handling errno had been that each shim function would do the conversion as needed, but I like your approach better.

src/Common/src/Interop/Unix/libc/Interop.mountpoints.cs
```diff
@@ -36,13 +36,13 @@ internal static partial class libc
         /// <param name="name">The drive name to retrieve the statfs data for.</param>
         /// <param name="data">The data retrieved from the mount point.</param>
         /// <returns>Returns true if data was filled with the results; otherwise, false.</returns>
-        internal static bool TryGetStatFsForDriveName(string name, out statfs data, out int errno)
+        internal static bool TryGetStatFsForDriveName(string name, out statfs data, out ErrorInfo errorInfo)
         {
             data = default(statfs);
-            errno = 0;
+            errorInfo = default(ErrorInfo);
```
stephentoub: Can we move this to just before the ```return true;```?

src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
```diff
@@ -44,7 +44,7 @@ internal static SafeFileHandle Open(string path, Interop.libc.OpenFlags flags, i
                 // Open the file.
                 int fd;
                 while (Interop.CheckIo(fd = Interop.libc.open(path, flags, mode), path, isDirectory: enoentDueToDirectory,
-                    errorRewriter: errno => (errno == Interop.Errors.EISDIR) ? Interop.Errors.EACCES : errno)) ;
+                    errorRewriter: e => (e.Error == Interop.Error.EISDIR) ? new Interop.ErrorInfo(Interop.Error.EACCES) : e)) ;
```
stephentoub: Per the discussion earlier about the efficiency of conversion between Error and Errno, this is another example where we end up constructing an ErrorInfo from an Error value but where (I think) we don't ever need the actual errno value, so it'd be nice not to have to pay the overheads of a managed->native call.  If you didn't want to cache the conversion table, another option to consider would be to make the conversion only happen when the alternate value is needed (changing ErrorInfo's fields to be properties).

src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
```diff
@@ -55,12 +55,12 @@ internal static SafeFileHandle Open(string path, Interop.libc.OpenFlags flags, i
...
                 {
                     handle.Dispose();
-                    throw Interop.GetExceptionForIoErrno(Marshal.GetLastWin32Error(), path);
+                    throw Interop.GetExceptionForIoErrno(Interop.System.GetLastErrorInfo(), path);
                 }
                 if ((buf.mode & Interop.libcoreclr.FileTypes.S_IFMT) == Interop.libcoreclr.FileTypes.S_IFDIR)
                 {
                     handle.Dispose();
-                    throw Interop.GetExceptionForIoErrno(Interop.Errors.EACCES, path, isDirectory: true);
+                    throw Interop.GetExceptionForIoErrno(new Interop.ErrorInfo(Interop.Error.EACCES), path, isDirectory: true);
```
stephentoub: Any readability value in adding an implicit or explicit cast from Error to ErrorInfo?

src/Native/System.Native/pal_errno.cpp
```diff
@@ -0,0 +1,249 @@
...
+        case Error::PAL_EXDEV:            return EXDEV;
+        case Error::PAL_ENONSTANDARD:     break; // fall through to assert
+    }
+
+    // We should not use this function to round-trip platform -> pal
+    // -> platform. It's here only to synthesize a platform number
+    // from the fixed set above. Note that the assert is outside the
+    // switch rather than in a default case default block because not
+    // having a default will trigger a warning (as error) if there's
+    // an enum value we haven't handled. Should that trigger, make
```
stephentoub: Nice idea.

src/Native/System.Native/pal_errno.cpp
```diff
@@ -0,0 +1,249 @@
...
+#if HAVE_GNU_STRERROR_R
+    return strerror_r(platformErrno, buffer, bufferSize);
+#else
+    // We deliberately ignore the return value here. The only three
+    // valid return values are 0 for success, EINVAL for an unknown
+    // errno value, and ERANGE if there's not enough buffer space
+    // provided. For EINVAL, it'll still fill the buffer with a
+    // reasonable error string (e.g. "Unknown error: 0x123"), and for
+    // ERANGE, it'll fill the buffer with as much of the error as it
+    // can and null-terminate it.
```
stephentoub: In the case of the XSI-compliant strerror_r, we could do better with regards to ERANGE, such as by returning NULL if we get ERANGE, at which point the managed side could allocate a larger buffer and call it again.

src/Native/System.Native/pal_stat.cpp
```diff
@@ -0,0 +1,87 @@
...
+static_assert(PAL_S_IXGRP == S_IXGRP, "");
+static_assert(PAL_S_IRWXO == S_IRWXO, "");
+static_assert(PAL_S_IROTH == S_IROTH, "");
+static_assert(PAL_S_IWOTH == S_IWOTH, "");
+static_assert(PAL_S_IXOTH == S_IXOTH, "");
+static_assert(PAL_S_ISUID == S_ISUID, "");
+static_assert(PAL_S_ISGID == S_ISGID, "");
+
+// These numeric values are not specified by POSIX, but the values
+// are common to our current targets.  If these static asserts fail,
```
stephentoub: Nice.

src/Native/System.Native/pal_stat.cpp
```diff
@@ -0,0 +1,87 @@
...
+    dst->Gid    = src.st_gid;
+    dst->Size   = src.st_size;
+    dst->ATime  = src.st_atime;
+    dst->MTime  = src.st_mtime;
+    dst->CTime  = src.st_ctime;
+
+#if HAVE_STAT_BIRTHTIME
+    dst->BirthTime = src.st_birthtime;
+    dst->Flags |= FILESTATUSFLAGS_HAS_BIRTHTIME;
+#endif
```
stephentoub: Should we zero-out dst->BirthTime otherwise?

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -290,56 +289,56 @@ private void RemoveDirectoryInternal(string fullPath, bool recursive, bool throw
...
+                        throw Interop.GetExceptionForIoErrno(errorInfo, fullPath, isDirectory: true);
                 }
             }
         }
 
         public override bool DirectoryExists(string fullPath)
         {
-            int errno;
-            return DirectoryExists(fullPath, out errno);
+            Interop.ErrorInfo errorInfo;
```
stephentoub: Maybe ```ignored``` instead of ```errorInfo```?  I'd previously named it ```erno``` because ```int``` isn't self-descriptive as to what's being ignored, but now that it's ```Interop.ErrorInfo```, it's much clearer.

src/System.IO.FileSystem/src/System/IO/UnixFileSystem.cs
```diff
@@ -290,56 +289,56 @@ private void RemoveDirectoryInternal(string fullPath, bool recursive, bool throw
...
         {
-            return FileExists(fullPath, Interop.libcoreclr.FileTypes.S_IFDIR, out errno);
+            return FileExists(fullPath, Interop.libcoreclr.FileTypes.S_IFDIR, out errorInfo);
         }
 
         public override bool FileExists(string fullPath)
         {
-            int errno;
-            return FileExists(fullPath, Interop.libcoreclr.FileTypes.S_IFREG, out errno);
+            Interop.ErrorInfo errorInfo;
```
stephentoub: Ditto.

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
```diff
@@ -20,13 +20,15 @@ internal sealed class UnixFileSystemObject : IFileSystemObject
...
             private Interop.libcoreclr.fileinfo _fileinfo;
 
-            /// <summary>
-            /// Whether we've successfully cached a stat structure.
-            /// -1 if we need to refresh _fileinfo, 0 if we've successfully cached one,
-            /// or any other value that serves as an errno error code from the
-            /// last time we tried and failed to refresh _fileinfo.
-            /// </summary>
-            private int _fileinfoInitialized = -1;
+            enum InitializationState : byte
```
stephentoub: Nit: ```private```

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
```diff
@@ -20,13 +20,15 @@ internal sealed class UnixFileSystemObject : IFileSystemObject
...
+            {
+                RefreshNeeded,     // we need to refresh _fileInfo
+                SuccessfullyCached, // we've successfully cached the stat structure
+                ErrorEncountered   // we hit an error on last try, get errno from _InitializatonErrno
```
stephentoub: What's ```_InitializationErrno```?  Is that supposed to be ```_fileInfoInitializationError```?

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
```diff
@@ -20,13 +20,15 @@ internal sealed class UnixFileSystemObject : IFileSystemObject
...
-            private int _fileinfoInitialized = -1;
+            enum InitializationState : byte
+            {
+                RefreshNeeded,     // we need to refresh _fileInfo
+                SuccessfullyCached, // we've successfully cached the stat structure
+                ErrorEncountered   // we hit an error on last try, get errno from _InitializatonErrno 
+            }
+
+            private InitializationState _fileinfoInitialization;
+            private Interop.ErrorInfo _fileinfoInitializationError;
```
stephentoub: Hmm... I don't love that this increases the size of UnixFileSystemObject.  Could you consider an alternative, even if it's a bit hacky?  e.g. having an ErrorInfo.Invalid value, or something like that?

src/Common/src/Interop/Unix/Interop.Errors.cs
nguerrera: I thought about this and was concerned about duplicating the table in each library (since we just compile the Interop.* in to each), but there aren't really that many errors so maybe it's OK after all.

src/Native/System.Native/pal_stat.cpp
nguerrera: Carried over from GetFileInformation to initial Stat in another file (rename gets lost in multiple change GitHub diff), but yes, I think that would be prudent.

src/Common/src/Interop/Unix/Interop.Errors.cs
stephentoub: That's a fair point.  Up to you.  Either way, we'll want to re-evaluate as we do more performance testing/tuning.

(Note that it could probably be done in a way that has the table shared across libraries, e.g. if the native code created the table and returned a pointer to it to managed in some format the managed code could easily use as-is.)

nguerrera: I'll file an issue to investigate, but stick to the current approach for the initial commit. You're right that it could be shared that way -- I hadn't thought of it. It is slightly tricky because we'd need to share something that is efficient to lookup in directly (whereas I was imagining building the lookup from an unordered list in managed code). The Pal -> Platform case can just be an index into an array so that's easy. However, the more interesting and common Platform -> Pal case can be sparse. I'd be inclined to use a sorted array of (int, Error) + binary search to keep the shared data structure trivial. It'll take me some time to measure, etc., and it's just complex enough to warrant its own review IMHO.

src/Common/src/Interop/Unix/libc/Interop.getcwd.cs
nguerrera: :) I started down that path and then realized that this would be much simpler and easier to maintain.

src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
nguerrera: Good idea.

src/Common/src/Interop/Unix/Interop.Errors.cs
stephentoub: Sounds good.

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
bartonjs: It feels a bit excessively long, personally.  I'd use something like `Interop.Libraries.CryptoShim` for `Interop.Crypto`.

But the total line length has a practical limit, and they should all be the same within a single interop file, so that's more an "I wouldn't have done it that way" than "I object".

Mainly I feel like SystemSecurityCryptographyNative is too long of an identifier, my brain's tokenizer doesn't quite have that much space, so it bogs down in parsing (or lexing, whatever).

src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
nguerrera: On second thought, I'll add ErrorInfo.EACCES and the others we use. Explicit conversion doesn't save many characters and implicit could lead to unintended things like calling GetLastError and converting it to ErrorInfo instead of GetLastErrorInfo.

src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
nguerrera: Yikes, I got the size of the int64_t fields wrong, but the only use so far doesn't read that far.

src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
nguerrera: Opted for an extension Info() method as explicit conversion. Doesn't require as much typing and reads better, but is still explicit, which I want.

src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
nguerrera: The byte will fit in prior padding next to existing bool, I think. But ErrorInfo is 2x errno. I opted to put the code back and reconvert the Error in the case where we throw. Seems like the right tradeoff as we get the size win for always and pay a slight time penalty only in the error case.

src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
nguerrera: Fixed.

src/Common/src/Interop/Linux/libc/Interop.inotify.cs
nguerrera: Let's settle this offline and I'll loop back to the naming as a follow-up. I've left Interop.Libraries.SystemNative for now, but change Interop.System to Interop.Sys and got rid of the "global::" shenanigans.

nguerrera:
@stephentoub I think I've addressed everything.

stephentoub:
CI failed in a FileSystemWatcher test:
```
FileSystemWatcher_4000_Tests.FileSystemWatcher_Created_Negative [FAIL]
00:30:32       Should not observe a created event
00:30:32       Expected: False
00:30:32       Actual:   True
00:30:32       Stack Trace:
00:30:32             at Utility.ExpectNoEvent(WaitHandle eventOccured, String eventName, Int32 timeout)
00:30:32             at FileSystemWatcher_4000_Tests.FileSystemWatcher_Created_Negative()
```
Not sure if that has anything to do with @sokket's FSW changes that got checked in last night.

src/Common/src/Interop/Unix/Interop.Errors.cs
```diff
@@ -1,48 +1,200 @@
...
+    internal partial class Sys
+    {
+        internal static Error GetLastError()
+        {
+            return ConvertErrorPlatformToPal(Marshal.GetLastWin32Error());
+        }
+
+        internal static ErrorInfo GetLastErrorInfo()
+        {
+            return new ErrorInfo(Marshal.GetLastWin32Error());
```
stephentoub: I know I suggested lazy initialization, and I know we have #2647 to follow up on this further, so this doesn't need to be done now.  But, it seems like any time we're getting the last errno value from the system, we're only doing so because we need to compare it against something, and that something is going to be an Error rather than a rawErrno value, so it seems like we should probably force the initialization here, e.g.
```C#
ErrorInfo ei = new ErrorInfo(Marshal.GetLastWin32Error());
var ignored = Error; // force initialize
return ei;
```
The danger with the lazy initialization on the struct is we could end up actually increasing the managed/native transition costs by passing around a copy of the struct that doesn't have the field initialized, such that every caller will end up incurring the overheads, rather than that happening just once.  By forcing it here, maybe we strike the right middle ground?

stephentoub:
Other than the unit test failure (not sure if it's related), LGTM.  Thanks, Nick.

src/Common/src/Interop/Unix/Interop.Errors.cs
nguerrera: Good point. 

I had a look and every place we were producing ErrorInfo from raw errno needed the Error a few instructions later. So I reverted lazy initialization in the case we don't need it: 5586bae. This saves an unnecessary branch and keeps the code simpler than above.

nguerrera:
FSW is passing locally, could be intermittent. Let's see how the run looks with 5586bae, which addresses the follow up to prevent risk of repeated conversion.

nguerrera:
Hmm, FSW again, but can't repro. :(

sokket:
Hmm that UT didn't change with my checkin...that looks to be an old test. 

That being said, this test looks ripe for a race condition; every OS's implementation of watching File System changes has some drawbacks, the Linux one is around directories. From looking at the failure and the test code, it looks like a timing issue around when the directory was moved and when the kernel generated the events. The test should probably be refactored to take this potential lag into account.

nguerrera:
@sokket Yes, I'm preparing a small change to the test that I think should address it.

nguerrera:
@dotnet-bot test this please

nguerrera:
@dotnet-bot test this please

PR #2609, opened 2015-08-03
Latest xUnit beta for all test projects

mellinoe:
This goes along with the buildtools PR 219 (https://github.com/dotnet/buildtools/pull/219).

* Updates all test projects to the latest xUnit beta packages
* Use a newer version of the xunit.netcore.extensions package, which depends on the latest xUnit beta
* Account for breaking changes in the xUnit API
    * MemberData requires matching parameter types (cannot just use object[]). Needed fixes in Linq.Parallel and MemoryMappedFiles tests which had tests broken by this change.
    * Assert.None removed (equivalent to Assert.DoesNotContain)
* Added some more explicit dependencies on some test projects to avoid warnings about "different versions of the same dependent assembly"

NOTE: This won't pass until the buildtools packages are published.


mellinoe:
4f5db27 shows most of the changes needed to account for the change in MemberData behavior.

mellinoe:
@dotnet-bot test this, please

mellinoe:
@dotnet-bot test this, please

mellinoe:
@dotnet-bot test this, please

mellinoe:
@dotnet-bot this this please

mellinoe:
@dotnet-bot test this, please

mellinoe:
There's a few failures in the TypeExtensions tests, not sure what's causing it exactly? Has anyone seen these failures before? They don't seem to be doing anything fancy.

    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest2
    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest3
    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest1

stephentoub:
@mellinoe, I just took a quick look.  I'm surprised these tests haven't failed before: tests that can be parallelized are accessing and mutating the same static field.

stephentoub:
Sorry, I take back my previous concurrency-related comment: I haven't tried running it, but it looks like these tests expect to be executed in a specific order, sharing state between then via a static field, and my guess is that order isn't holding as expected.

mellinoe:
I'm going to disable them for now, I am thinking that what you have said is correct. I looked and they do seem to be modifying the same static variable, incrementing it or decrementing it and then checking the value of it. Seems pretty fragile so I'll disable for now.

src/Common/tests/System/Diagnostics/AssertWithCallerAttributes.cs
```diff
@@ -277,21 +277,21 @@ internal static class Assert
     public static void None(IEnumerable collection, object expected,
         [CallerFilePath] string path = null, [CallerLineNumber] int line = 0)
     {
-        try { Xunit.Assert.None(collection, expected); }
+        try { Xunit.Assert.DoesNotContain(expected, (IEnumerable<object>)collection); }
```
stephentoub: The name of this method should change to DoesNotContain, too.

stephentoub:
Skimmed the changes; LGTM.  I left one comment, but it's minor and can be addressed separately; I'd rather get these xunit updates in.

mellinoe:
I'll merge this in and work on addressing the comment in a minute.

Issue #2610, opened 2015-08-03
FileSystem GetLastWriteTime and GetLastAccessTime don't properly convert values to Local DateTimeKind


ianhays:
The following test has different results for each platform:
```
FileInfo testFile = new FileInfo("test.txt");
testFile.Create().Dispose();

DateTime dt = new DateTime(2014, 12, 1, 12, 0, 0, DateTimeKind.Local);
Console.WriteLine("regular: " + dt.ToString());
Console.WriteLine("local: " + dt.ToLocalTime().ToString());
Console.WriteLine("utc: " + dt.ToUniversalTime().ToString());

File.SetLastWriteTime(testFile.FullName, dt);
var res = File.GetLastWriteTime(testFile.FullName);
Console.WriteLine("result: " + res.ToString());
Console.WriteLine("result.toLocal(): " + res.ToLocalTime().ToString());
Console.WriteLine("result.ToUTC(): " + res.ToUniversalTime().ToString()); 
Console.WriteLine();
```
On Desktop Framework:
```
regular: 12/1/2014 12:00:00 PM
local: 12/1/2014 12:00:00 PM
utc: 12/1/2014 8:00:00 PM
result: 12/1/2014 12:00:00 PM
result.toLocal(): 12/1/2014 12:00:00 PM
result.ToUTC(): 12/1/2014 8:00:00 PM
```
On Win32 Core:
```
regular: 12/1/2014 12:00:00 PM
local: 12/1/2014 12:00:00 PM
utc: 12/1/2014 8:00:00 PM
result: 12/1/2014 12:00:00 PM
result.toLocal(): 12/1/2014 4:00:00 AM
result.ToUTC(): 12/1/2014 8:00:00 PM
```
On Linux Core:
```
regular: 12/01/2014 12:00:00
local: 12/01/2014 12:00:00
utc: 12/01/2014 4:00:00
result: 12/01/2014 04:00:00
result.toLocal(): 12/01/2014 12:00:00
result.ToUTC(): 12/01/2014 20:00:00
```

The issue lies in the conversions from DateTimeOffsets to DateTimes: nowhere is DateTimeKind (i.e. local or UTC) considered when reading the attributes of a file. To make the behavior in line with the Desktop, we should convert all filesystem times (CreationTime, WriteTime, AccessTime) to DateTimeKind.Local when converting them from a DateTimeOffset to a DateTime.

Thanks to @eerhardt for help finding the lack of conversion.

vfrank66:
[...]

PR #2611, opened 2015-08-03
Fixed AccessTime, WriteTime, and CreationTime to return a DateTime object with properly set DateTimeKind

### LINKED ISSUE #2603: FileSystem tests failing on Linux if local timezone != UTC
```
System.IO.FileSystem.Tests.File_GetSetTimes.SettingUpdatesProperties [FAIL]
      Assert.All() Failure: 2 out of 4 items in the collection did not pass.
      [2]: Xunit.Sdk.EqualException: Assert.Equal() Failure
           Expected: 2014-12-01T12:00:00.0000000-08:00
           Actual:   2014-12-01T20:00:00.0000000
```

ianhays:
- The implementations of GetLastAccessTime, GetCreationTime, and GetLastWriteTime were not properly converting the read values to LocalTime for either Windows or Unix.
- Resolves #2603, #2610

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/File.cs
+++ b/src/System.IO.FileSystem/src/System/IO/File.cs
@@ -260,7 +260,7 @@ namespace System.IO
         public static DateTime GetCreationTime(String path)
         {
             String fullPath = PathHelpers.GetFullPathInternal(path);
-            return FileSystem.Current.GetCreationTime(fullPath).DateTime;
+            return FileSystem.Current.GetCreationTime(fullPath).LocalDateTime;
         }
 
         [System.Security.SecuritySafeCritical]  // auto-generated
@@ -286,7 +286,7 @@ namespace System.IO
         public static DateTime GetLastAccessTime(String path)
         {
             String fullPath = PathHelpers.GetFullPathInternal(path);
-            return FileSystem.Current.GetLastAccessTime(fullPath).DateTime;
+            return FileSystem.Current.GetLastAccessTime(fullPath).LocalDateTime;
         }
 
         [System.Security.SecuritySafeCritical]  // auto-generated
@@ -312,7 +312,7 @@ namespace System.IO
         public static DateTime GetLastWriteTime(String path)
         {
             String fullPath = PathHelpers.GetFullPathInternal(path);
-            return FileSystem.Current.GetLastWriteTime(fullPath).DateTime;
+            return FileSystem.Current.GetLastWriteTime(fullPath).LocalDateTime;
         }
 
         [System.Security.SecuritySafeCritical]  // auto-generated
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileSystemObject.cs
@@ -193,7 +193,7 @@ namespace System.IO
                 get
                 {
                     EnsureStatInitialized();
-                    return DateTimeOffset.FromUnixTimeSeconds(_fileinfo.atime);
+                    return DateTimeOffset.FromUnixTimeSeconds(_fileinfo.atime).ToLocalTime();
                 }
                 set { SetAccessWriteTimes((IntPtr)value.ToUnixTimeSeconds(), null); }
             }
@@ -203,7 +203,7 @@ namespace System.IO
                 get
                 {
                     EnsureStatInitialized();
-                    return DateTimeOffset.FromUnixTimeSeconds(_fileinfo.mtime);
+                    return DateTimeOffset.FromUnixTimeSeconds(_fileinfo.mtime).ToLocalTime();
                 }
                 set { SetAccessWriteTimes(null, (IntPtr)value.ToUnixTimeSeconds()); }
             }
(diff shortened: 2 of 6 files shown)


src/System.IO.FileSystem/src/System/IO/Win32FileSystemObject.cs
```diff
@@ -64,7 +64,7 @@ public DateTimeOffset CreationTime
                 {
                     EnsureDataInitialized();
                     long dt = ((long)(_data.ftCreationTimeHigh) << 32) | ((long)_data.ftCreationTimeLow);
-                    return DateTimeOffset.FromFileTime(dt);
+                    return DateTimeOffset.FromFileTime(dt).ToLocalTime();
```
stephentoub: Why is ToLocalTime necessary?  Doesn't FromFileTime already convert to local time?

stephentoub:
cc: @ericstj, @eerhardt

nguerrera:
Looks good to me, but let the I/O and timezone folks chime in.

One small nit: in the future (or now if you feel like rebasing), can you make an effort to use the first line of commits as a general subject and not the first of several changes in a bulleted list. Also make sure it's short enough to not get truncated. It renders things in many tools including GitHub much nicer and is standard git practice as well as documented in our contributing guidelines (although we've added quite a few clicks to get to it from corefx CONTRIBUTING.md of late): https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/contributing.md#commit-messages

src/System.IO.FileSystem/src/System/IO/Win32FileSystemObject.cs
eerhardt: This appears to be an inconsistency between DateTimeOffset.FromFileTime and DateTimeOffset.FromUnixTimeSeconds.  FromFileTime does "ToLocalTime()" inside of it.  But FromUnixTimeSeconds leaves the DateTime as Unspecified.  This inconsistency might be the root cause of this bug, since the Win32FileSystemObject was just calling FromFileTime - when the UnixFileSystemObject was written, it just called FromUnixTimeSeconds without doing the conversion.

It feels like these two methods should be consistent in their output - both should return the same DateTimeKind.

ianhays: The windows calls to ToLocalTime() can be safely removed after all.

It looks like the fault can be further narrowed down to FromUnixTimeSeconds/FromUnixTimeMilliseconds in DateTimeOffset.cs in mscorlib; they return a Zero Timespan when they should be returning Timespan representing the offset of the time zone from UTC.

DateTime has a FromFileTime that is the same as the one in DateTimeOffset. Would we also want a FromUnixTime*seconds in DateTime and have the DateTimeOffset delegate to it?

ianhays: Slight addition to my above comment:
The windows calls to ToLocalTime() may be removed, but <b>only</b> if the DateTimeOffsets returned use .LocalTime instead of .DateTime to convert to a DateTime object. Otherwise we get errors.

I apologize for the potentially confusing comments; there are so many places where the conversion can go wrong it's difficult to nail down exactly which parts should be fixed. I'll add some more specific tests.

ianhays: Windows DateTimeOffset.FromFileTime is working the same way that is on Desktop: FromFileTime returns a DateTimeOffset with a properly set time. The issue on Windows is only in File.cs where the DateTimeOffset is being converted into a DateTime using the .DateTime property (DateTimeKind.Unspecified) instead of the .LocalTime property (DateTimeKind.Local).

ianhays:
@nguerrera Thanks for the link! I rebased to one commit that better follows the guidelines

@stephentoub @eerhardt I spoke with @tarekgh who confirmed that FromUnixTimeSeconds should return a UTC value, so my final commit converts it specifically to LocalDateTime instead of DateTime to be in line with the Win32 FromFileTime behavior. The end result relevant to FileSystem is that the GetWriteTime and GetAccessTime accessors will return DateTimes with DateTimeKind.Local across all platforms.

Thanks for all of the assistance everyone!

src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs
```diff
@@ -368,7 +368,7 @@ public override DateTimeOffset GetCreationTime(string fullPath)
                 throw Win32Marshal.GetExceptionForWin32Error(errorCode, fullPath);
 
             long dt = ((long)(data.ftCreationTimeHigh) << 32) | ((long)data.ftCreationTimeLow);
-            return DateTimeOffset.FromFileTime(dt);
+            return DateTimeOffset.FromFileTime(dt).ToLocalTime();
```
stephentoub: FromFileTime doesn't already return local time?

eerhardt:
LGTM

src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs
ianhays: ToLocalTime() is unnecessary for FromFileTime, you are correct.

I missed CreationTime in my undoing of the ToLocalTime additions; I'll rebase again tomorrow morning and fix this.

PR #2614, opened 2015-08-04
Remove build warnings in cryptography libraries

bartonjs:
System.Security.Cryptography.Primitives uses System.IO-4.0.10.  The rest of the cryptography libraries seems to get an implicit reference to System.IO, but 4.0.0.  A post-build analysis is not pleased by this state of affairs.

So, for all libraries that currently package-to-package reference System.Security.Cryptography.Primitives; add a System.IO-4.0.10 reference.

### DIFF
--- a/src/System.Security.Cryptography.Algorithms/src/project.json
+++ b/src/System.Security.Cryptography.Algorithms/src/project.json
@@ -2,6 +2,7 @@
   "dependencies": {
     "System.Diagnostics.Contracts": "4.0.0",
     "System.Diagnostics.Debug": "4.0.10",
+    "System.IO": "4.0.10",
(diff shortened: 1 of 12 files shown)


bartonjs:
cc: @ericstj

stephentoub:
LGTM

bartonjs:
@dotnet-bot test this please.

mellinoe:
LGTM. Is this because the other libraries have a direct project reference to Cryptography.Primitives rather than a package reference?

bartonjs:
@mellinoe Yeah, the warning got introduced when I had to change the package references to project references due to the contract overhaul; but it seems like good housekeeping anyways.

ericstj:
LGTM

PR #2615, opened 2015-08-04
Implement System.Security.Cryptography.CngKey

AtsushiKan:


stephentoub:
cc: @bartonjs, @nguerrera

src/System.Security.Cryptography.Cng/System.Security.Cryptography.Cng.sln
```diff
@@ -0,0 +1,28 @@
...
+# Visual Studio 14
+VisualStudioVersion = 14.0.23107.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Security.Cryptography.Cng", "src\System.Security.Cryptography.Cng.csproj", "{4C1BD451-6A99-45E7-9339-79C77C42EE9E}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Security.Cryptography.Cng.Tests", "tests\System.Security.Cryptography.Cng.Tests.csproj", "{FF53459F-66F7-4F00-8D36-DF440CE18419}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
stephentoub: This is a Windows-only project, right?  For consistency with other such projects, we should name the configurations Windows_Debug, Windows_Release, etc.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
```diff
@@ -0,0 +1,145 @@
...
+
+using Microsoft.Win32.SafeHandles;
+
+using ErrorCode = Interop.NCrypt.ErrorCode;
+
+namespace Internal.Cryptography
+{
+    internal static class Helpers
+    {
+        public static byte[] CloneByteArray(this byte[] src)
```
stephentoub: Doesn't need to be as part of this PR, but we should really look into consolidating a bunch of these Helpers.  For example, I think I've seen this Clone method duplicated in at least a handful of the other crypto libs.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            return (byte[])(src.Clone());
+        }
+
+        //
+        // The C# construct
+        //   
+        //    fixed (byte* p = new byte[0])
```
stephentoub: @jaredpar, is this by design?

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
```diff
@@ -0,0 +1,145 @@
...
+        // This helper enables the syntax:
+        //
+        //    fixed (byte* p = new byte[0].MapZeroLengthArrayToNonNullPointer())
+        //
+        // which always sets "p" to a non-NULL pointer for a non-null byte array. 
+        //
+        public static byte[] MapZeroLengthArrayToNonNullPointer(this byte[] src)
+        {
+            if (src != null && src.Length == 0)
+                return new byte[1];
```
stephentoub: How often is this helper used?  If it's common, could we cache such an array in a read-only static?

AtsushiKan: The 1-element array only gets created in corner cases where someone sets a 0-length property on a key.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/KeyPropertyName.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        //
+        // Properties defined by the CLR
+        //
+
+        /// <summary>
+        ///     Is the key a CLR created ephemeral key, it will contain a single byte with value 1 if the
+        ///     key was created by the CLR as an ephemeral key.
+        /// </summary>
+        internal const string ClrIsEphemeral = "CLR IsEphemeral";
```
stephentoub: Now that all of the crypto support for corefx is implemented outside of the runtime, is this still relevant?

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
```diff
@@ -53,42 +56,46 @@ private enum OwnershipState {
...
-        private SafeNCryptHandle Holder {
-            get {
-                Contract.Requires((m_ownershipState == OwnershipState.Duplicate && m_holder != null) ||
-                                  (m_ownershipState != OwnershipState.Duplicate && m_holder == null));
-                Contract.Requires(m_holder == null || m_holder.m_ownershipState == OwnershipState.Holder);
+        private SafeNCryptHandle Holder
+        {
+            get
+            {
+                Contract.Requires((_ownershipState == OwnershipState.Duplicate && _holder != null) ||
```
stephentoub: We're not using the contract rewriter, nor are we compiling with CONTRACTS_FULL.  At least for now, should we make all of these Contract.Requres into Debug.Asserts?

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
```diff
@@ -53,42 +56,46 @@ private enum OwnershipState {
...
                 Contract.Ensures(IsValidOpenState);
                 Contract.Assert(value.IsValidOpenState);
 #endif
-                Contract.Assert(m_ownershipState != OwnershipState.Duplicate);
-                Contract.Assert(value.m_ownershipState == OwnershipState.Holder);
-                
-               
-                m_holder = value;
-                m_ownershipState = OwnershipState.Duplicate;
+                Contract.Assert(_ownershipState != OwnershipState.Duplicate);
```
stephentoub: We've standardized across corefx on using Debug.Assert instead of Contract.Assert.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
jaredpar: @stephentoub it is explicitly undefined :smile:.  From section 18.6 in C# 3.5 spec: 

> The behavior of the fixed statement is implementation-defined if the array expression is null or if the array has zero elements

The IL though seems to dynamically check the length of arrays and use `null` in case of `0` or a `null` array.

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
```diff
@@ -194,17 +211,18 @@ private enum OwnershipState {
         ///     See code:Microsoft.Win32.SafeHandles.SafeNCryptHandle#NCryptHandleDuplicationAlgorithm for
         ///     details about the algorithm.
         /// </summary>
-        private T DuplicateDuplicatedHandle<T>() where T : SafeNCryptHandle, new() {
-            Contract.Ensures(m_ownershipState == OwnershipState.Duplicate);
+        private T DuplicateDuplicatedHandle<T>() where T : SafeNCryptHandle, new()
```
stephentoub: I'm trying to understand why it's ok that we're addref'ing and suppressing finalization (seems like that's a recipe for leaking these handles), so I want to look at the algorithm mentioned in the comment, but I can't find NCryptHandleDuplicationAlgorithm... where is that?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithm.cs
```diff
@@ -0,0 +1,290 @@
...
+        {
+            if (object.ReferenceEquals(left, null))
+            {
+                return object.ReferenceEquals(right, null);
+            }
+
+            return left.Equals(right);
+        }
+
+        [Pure]
```
stephentoub: If we're going to use [Pure], seems like we should use it consistently... why is != Pure but == isn't?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithm.cs
```diff
@@ -0,0 +1,290 @@
...
+                Contract.Ensures(Contract.Result<CngAlgorithm>() != null);
+
+                if (s_sha512 == null)
+                {
+                    s_sha512 = new CngAlgorithm("SHA512"); // BCRYPT_SHA512_ALGORITHM
+                }
+
+                return s_sha512;
+            }
+        }
```
stephentoub: We could simplify all of these properties by introducing a helper, e.g.
```C#
private static CngAlgorithm GetCachedAlgorithm(ref CngAlgorithm cache, string algorithm)
{
    return cache ?? (cache = new CngAlgorithm(algorithm));
}
```
which we then use from the various properties, e.g.
```C#
public static CngAlgorithm Sha512 { get { return GetCachedAlgorithm(ref s_sha512, "SHA512"); } }
```

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithm.cs
AtsushiKan: That [Pure] stuff carried over from the original desktop code. Pure is defined in S.D.Contracts so if we're getting rid of the other contract stuff, I'll get rid of these too.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithm.cs
stephentoub: (Just noticed you have the fields all marked as volatile.  You could do the above using Volatile.Read/Write instead.  And as long as we're making these volatile, do you want to use an Interlocked to ensure only a single instance is ever returned?)

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/KeyPropertyName.cs
AtsushiKan: We still have to track the ephemeral state for third-party KSP's and the name "CLR Ephemeral" is our reserved property name (and visible to apps.) So we have to keep the name, anachronistic as it is.

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
AtsushiKan: I didn't want to throw out the Contract info unnecessarily as this seems like goodness for something as tricky as safehandles. Still, I don't object to getting rid of the contracts entirely, given that we're not testing them anyway.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            return key;
+        }
+
+        /// <summary>
+        ///     Setup the key properties specified in the key creation parameters
+        /// </summary>
+        private static void InitializeKeyProperties(SafeNCryptHandle keyHandle, CngKeyCreationParameters creationParameters)
+        {
+            unsafe
```
stephentoub: If the whole method is unsafe, you could choose to move the ```unsafe``` into the signature; doing so would just save the extra level of indenting.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithm.cs
AtsushiKan: I think apps expect the static properties to have perf characteristics like using a enum literal. Actually, we could probably replace all this lazy initialization code with static field initializers. 
That would also guarantee a single-instance (though I don't believe single instancing is actually required here.)

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
AtsushiKan: I could but I stick consistently to the rule that "unsafe" only goes into the signature if one of the parameter types is unsafe. Otherwise, it's an implementation detail and it rubs me the wrong way to put the keyword into the signature line.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
AtsushiKan: CloneByteArray() is the one that's really common across the crypto libraries, IIRC. By itself, it doesn't merit creating a source coupling between these libraries.

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
AtsushiKan: Jeremy ported the safe handle code, best to ask him on this one. I only made changes to fit the coding style rules.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.OpenHandle.cs
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    throw new ArgumentException(SR.Cryptography_OpenEphemeralKeyHandleWithoutEphemeralFlag, "keyHandleOpenOptions");
+                }
+
+                keyFullySetup = true;
+            }
+            finally
+            {
+                // Make sure that we don't leak the handle the CngKey duplicated
+                if (!keyFullySetup && key != null)
```
stephentoub: If this finally were changed to a catch, couldn't we eliminate this keyFullySetup mechanism? e.g.
```C#
catch
{
    if (key != null)
        key.Dispose();
    throw;
}
```

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.StandardProperties.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        /// <summary>
+        ///     Size, in bits, of the key
+        /// </summary>
+        public int KeySize
+        {
+            get
+            {
+                int keySize = _keyHandle.GetPropertyAsDword(KeyPropertyName.Length, CngPropertyOptions.None);
+                return keySize;
```
stephentoub: Why this pattern of store into a local and then returning the local, rather than just returning the result directly?  To help with debugging or something?

jaredpar: That pattern shouldn't be used anymore, even for debugging.  The latest release of Visual Studio supports return value in the autos window making this pattern obsolete.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKeyBlobFormat.cs
```diff
@@ -0,0 +1,197 @@
...
+            get
+            {
+                Contract.Ensures(Contract.Result<CngKeyBlobFormat>() != null);
+
+                if (s_eccPrivate == null)
+                {
+                    s_eccPrivate = new CngKeyBlobFormat("ECCPRIVATEBLOB");      // BCRYPT_ECCPRIVATE_BLOB
+                }
+
+                return s_eccPrivate;
```
stephentoub: FWIW, even if you don't factor this out into a helper, I personally like the more concise:
```C#
return s_eccPrivate ?? (s_eccPrivate = new CngKeyBlobFormat("ECCPRIVATEBLOB"));
```
I realize it's a matter of taste.  Just sharing :)

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.StandardProperties.cs
AtsushiKan: > To help with debugging?

Exactly.

AtsushiKan: Not changing my stance on this one. I've seen enough cases where the "latest and greatest" features are either unknown, non-functional or not available at the time (think .NET native compiled apps being debugged with windbg.) There's no harm in using this pattern.

src/System.Security.Cryptography.Cng/System.Security.Cryptography.Cng.sln
AtsushiKan: Fixed with updated PR (though I neglected to note that the commit description.)

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs
AtsushiKan: I'm going to offline this one. Maybe there's a cleanness issue here, maybe not but if there is, we want to look at this more holistically (i.e. X509Certificates also has this pattern) rather than throwing in patches into this PR.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Diagnostics;
+using System.IO;
+
+namespace System.Security.Cryptography
+{
+    public sealed class RSACng : RSA
+    {
+        public RSACng()
+        {
+            throw new NotImplementedException(SR.WorkInProgress);
```
stephentoub: I don't remember the current state of things internally: are these NotImplementedExceptions going to cause build warnings internally such that this should instead use the NotImplemented.ByDesign function?

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
```diff
@@ -0,0 +1,49 @@
...
+                Assert.Equal(true, key.IsEphemeral);
+                Assert.Equal(false, key.IsMachineKey);
+                Assert.Equal(null, key.KeyName);
+                Assert.Equal(0x100, key.KeySize);
+                Assert.Equal(CngKeyUsages.AllUsages, key.KeyUsage);
+                Assert.Equal(IntPtr.Zero, key.ParentWindowHandle);
+                Assert.Equal(CngProvider.MicrosoftSoftwareKeyStorageProvider, key.Provider);
+
+                CngUIPolicy policy = key.UIPolicy;
+                Assert.Equal(null, policy.CreationTitle);
```
stephentoub: Assert.Null

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace System.Security.Cryptography.Cng.Tests
+{
+    public static class ImportExportTests
+    {
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void TestImportExport1()
+        {
+
+            byte[] reExported;
```
stephentoub: Nit: why is this declared all the way up here?  Couldn't the assert be done inside the using as well?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
AtsushiKan: I already cross-built this on the K branch to check this out - Open-source assemblies don't trigger the NI cop.

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
AtsushiKan: Probably inheriting the pattern of other (private) tests where the "using" line actually creates a key on disk, so I want to get rid of it as fast as I can because I get tired of cleaning up turd off my machine.

Not a factor here since I'm only checking in non-invasive tests. I'll put this on the "clean up if I rev the PR again" list.

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
AtsushiKan: The test is checking a bunch of properties with a mix of null and non-null values so I'd rather use Assert.Equal() for all of them for consistency.

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
AtsushiKan: Final PR moves reExported back into using block. Do you have any other blocking issues?

src/System.Security.Cryptography.Cng/src/Interop/Interop.Libraries.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+internal static partial class Interop
+{
+    internal static class Libraries
+    {
+        internal const string NCrypt = "ncrypt.dll";
```
bartonjs: Why not use src/Common/src/Interop/Windows/Interop.Libraries.cs?

src/System.Security.Cryptography.Cng/src/Interop/NCrypt/NCrypt.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
```
bartonjs: Why not src/Common/src/Interop/Windows/NCrypt/... ?  Since CNG is ultimately used in the Algorithms library (which has outstanding issues for having bad Interop style) there likely is P/Invoke signature sharing that could go on here.

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
```diff
@@ -53,42 +55,43 @@ private enum OwnershipState {
...
-            set {
-                Contract.Ensures(m_holder.m_ownershipState == OwnershipState.Holder);
-                Contract.Ensures(m_ownershipState == OwnershipState.Duplicate);
+            set
+            {
 #if DEBUG
-                Contract.Ensures(IsValidOpenState);
-                Contract.Assert(value.IsValidOpenState);
+                Debug.Assert(value.IsValidOpenState);
 #endif
```
bartonjs:
```C#
#if DEBUG
Debug.Assert(...);
#endif
```

Feels a bit redundant; could the `#if` be removed?

src/System.Security.Cryptography.Cng/src/Interop/Interop.Libraries.cs
AtsushiKan: The CNG interop class is unlikely to be used by any other library (and would require the caller to consume the SSC.Cng contract anyway) so I chose to keep it local. It still follows the conventions so it can be easily merged into common if later appropriate.




In src/System.Security.Cryptography.Cng/src/Interop/Interop.Libraries.cs<https://github.com/dotnet/corefx/pull/2615#discussion_r36243338>:

> @@ -0,0 +1,10 @@

> +// Copyright (c) Microsoft. All rights reserved.

> +// Licensed under the MIT license. See LICENSE file in the project root for full license information.

> +

> +internal static partial class Interop

> +{

> +    internal static class Libraries

> +    {

> +        internal const string NCrypt = "ncrypt.dll";

Why not use src/Common/src/Interop/Windows/Interop.Libraries.cs?

src/System.Security.Cryptography.Cng/src/Microsoft/Win32/SafeHandles/NCryptSafeHandles.cs
AtsushiKan: No, because IsValidOpenState itself is #if DEBUG only.

src/System.Security.Cryptography.Cng/src/Resources/Strings.resx
```diff
@@ -0,0 +1,186 @@
...
+  <data name="Cryptography_NonCompliantFIPSAlgorithm" xml:space="preserve">
+    <value>This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.</value>
+  </data>
+  <data name="Cryptography_TlsRequiresLabelAndSeed" xml:space="preserve">
+    <value>The TLS key derivation function requires both the label and seed properties to be set.</value>
+  </data>
+  <data name="Cryptography_UnsupportedPaddingMode" xml:space="preserve">
+    <value>The specified PaddingMode is not supported.</value>
+  </data>
+  <data name="Cryptography_ECXmlSerializationFormatRequired" xml:space="preserve">
```
bartonjs: Do we even have EC XML Serialization?  Are the strings here the minimal required set?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngAlgorithmGroup.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        public static bool operator ==(CngAlgorithmGroup left, CngAlgorithmGroup right)
+        {
+            if (object.ReferenceEquals(left, null))
+            {
+                return object.ReferenceEquals(right, null);
+            }
+
+
```
bartonjs: Nit: excess blank lines

src/System.Security.Cryptography.Cng/src/Resources/Strings.resx
AtsushiKan: We can prune them after the library is complete. For productivity efficiency, I hoovered up every string I could find the CNG-related desktop code. Nothing more annoying than being yanked out of the zone because of a missing SR string.



In src/System.Security.Cryptography.Cng/src/Resources/Strings.resx<https://github.com/dotnet/corefx/pull/2615#discussion_r36243957>:

> +  <data name="Cryptography_NotValidPublicOrPrivateKey" xml:space="preserve">

> +    <value>Key is not a valid public or private key.</value>

> +  </data>

> +  <data name="Cryptography_InvalidSignatureAlgorithm" xml:space="preserve">

> +    <value>The hash algorithm is not supported for signatures. Only MD5, SHA1, SHA256,SHA384, and SHA512 are supported at this time.</value>

> +  </data>

> +  <data name="Cryptography_NonCompliantFIPSAlgorithm" xml:space="preserve">

> +    <value>This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.</value>

> +  </data>

> +  <data name="Cryptography_TlsRequiresLabelAndSeed" xml:space="preserve">

> +    <value>The TLS key derivation function requires both the label and seed properties to be set.</value>

> +  </data>

> +  <data name="Cryptography_UnsupportedPaddingMode" xml:space="preserve">

> +    <value>The specified PaddingMode is not supported.</value>

> +  </data>

> +  <data name="Cryptography_ECXmlSerializationFormatRequired" xml:space="preserve">

Do we even have EC XML Serialization? Are the strings here the minimal required set?

src/System.Security.Cryptography.Cng/src/Interop/NCrypt/NCrypt.cs
AtsushiKan: Algorithms uses BCrypt calls, not NCrypt calls so there’s no overlap. The RSACng code might have more overlap – will keep that in mind. The Cng contract surfaces so many enums that map directly to NCRYPT constants that the interop layer is tied to the Cng contract anyway, unless we want a whole bunch of enum->int casts in the code.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                fixed (char* pinnedCreationTitle = uiPolicy.CreationTitle, 
+                             pinnedFriendlyName = uiPolicy.FriendlyName,
+                             pinnedDescription = uiPolicy.Description)
+                {
+                    NCRYPT_UI_POLICY ncryptUiPolicy = new NCRYPT_UI_POLICY()
+                    {
+                        dwVersion = 1,
```
bartonjs: Is there an associated named constant for version 1?  I feel like Windows is usually pretty good at not requiring magic numbers to float around.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
```diff
@@ -0,0 +1,159 @@
...
+                string useContext = uiPolicy.UseContext;
+                if (useContext != null)
+                {
+                    int useContextByteLength = checked((useContext.Length + 1) * sizeof(char));
+                    fixed (char* pinnedUseContext = useContext)
+                    {
+                        ErrorCode errorCode = Interop.NCrypt.NCryptSetProperty(keyHandle, KeyPropertyName.UseContext, pinnedUseContext, useContextByteLength, CngPropertyOptions.Persist);
+                        if (errorCode != ErrorCode.ERROR_SUCCESS)
+                            throw errorCode.ToCryptographicException();
+                    }
```
bartonjs: I understand the `+1` (`\0`), and the `char*`, but this feels complicated enough that maybe an NCryptSetStringProperty (or just an overload) should exist in the Interop.NCrypt type as a helper method; for any other place that string properties need to be set.

If this is set differently than other string properties would be, a comment explaining how it's different would go a long way.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
AtsushiKan: Yes, the official named constant is apparently “1” ☺

Quoting from MSDN page for NCRYPT_UI_POLICY:

dwVersion
    The version number of the structure. This member must contain 1.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Create.cs
AtsushiKan: This is the only place that sets a string key property. That’s why I didn’t add a helper. I’ve found that the helpers in the original code hid a lot of quirky edge-case behavior that then spread virally throughout the code so I didn’t want to go down that path. As it turned out, a single “void*” overload actually worked nicely here compared to the alternatives.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.StandardProperties.cs
```diff
@@ -0,0 +1,302 @@
...
+                        // and Win32 layer does not enforce error return contract.
+                        // Therefore, they can return whatever error code they think appropriate.
+                        return false;
+                    }
+
+                    if (cbResult != sizeof(byte))
+                        return false;
+
+                    if (propertyValue != 1)
+                        return false;
```
bartonjs: Would we want to `Debug.Assert(propertyValue == 0)` here?  Desktop currently returns true if this is 1, this returns false if it's not 1, so they'd differ on a seemingly-illegal value of 2.

AtsushiKan: We can’t Debug.Assert the value since an app could set the “CLR IsEphemeral” property with an illegal value himself (ah, the joys of in-band signaling.) So an illegal value isn’t
necessarily a sign of a bug inside CoreFx. The desktop and CoreFx code use the same logic – they both return “IsEphemeral = false” for 2 and other illegal values. I just verified this on the desktop and CoreFx.

bartonjs: Right, I can do Boolean conversions.
```C#
bool b0 = !(v != 1);
bool b1 = v == 1;
```
Those do, in fact, say the same thing.  My bad.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKeyCreationParameters.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+
+namespace System.Security.Cryptography
+{
+    /// <summary>
+    ///     Settings to be applied to a CNG key before it is finalized.
```
bartonjs: On MSDN we have this as "Contains advanced properties for key creation.", which doesn't introduce confusion around the word "finalized".

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKeyHandleOpenOptions.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+
+namespace System.Security.Cryptography
+{
+    /// <summary>
+    ///     Parameter to CngKey.Open(SafeNCryptKeyHandle,...). This is not a mapping of a native NCrypt value.
```
bartonjs: I don't know if we'll ever have MSDN documentation generated from in-source comments, but "This is not a mapping of a native NCrypt value. " feels more like a `// Note: This is NOT a mapping of a native NCrypt value.` comment than a `/// <summary />` one.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKeyCreationParameters.cs
AtsushiKan: Sounds good. Will add the change to the list for the next PR revision.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngProperty.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Runtime.InteropServices;
+
+using Internal.Cryptography;
+
+namespace System.Security.Cryptography
+{
+    /// <summary>
+    ///     Wrapper represeting an arbitrary property of a CNG key or provider
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]  // The [StructLayout] is here to prevent a spurious ApiReviewer alert. We do not actually depend on the layout of this struct.
```
bartonjs: If you haven't already, please track things like this as you build out the library, and then we can do a pass over the contract generator to fix these issues.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKeyHandleOpenOptions.cs
AtsushiKan: I could go either way on this, but I’ll make the suggested change.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngPropertyOptions.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+
+namespace System.Security.Cryptography
+{
+    /// <summary>
+    ///     Maps to the "dwFlags" parameter of NCryptGetProperty() and NCryptSetProperty().
```
bartonjs: An awful lot of the enum `<summary>` values seem to be notes to a reader/editor of the code, rather than notes to the consumer of the code.

AtsushiKan: If we ever use these to generate MSDN docs, I assume they’ll have to go through a thorough sweep so I’m not too stressed about that now.

(Given that SSC.Cng is a lightweight interop layer and all but demands knowledge of CNG to use properly, I actually do think that noting the correspondence between the managed enum and the ncrypt.h #define is useful info for callers.)

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngUIPolicy.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        public string Description { get; private set; }
+
+        /// <summary>
+        ///     Description of how the key will be used
+        /// </summary>
+        public string UseContext { get; private set; }
+
+        /// <summary>
+        ///     Title of the dialog box displayed when a newly created key is finalized, null for the default title
```
bartonjs: I'm not sure what "finalized" means in this context...

src/System.Security.Cryptography.Cng/tests/CreateTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using Test.Cryptography;
+using Xunit;
+
+namespace System.Security.Cryptography.Cng.Tests
+{
+    public static class CreateTests
+    {
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void CreateEmphemeral1()
```
bartonjs: Rather than just "1" and "2", can the test names describe what they're doing (in particular, different from each other?)

src/System.Security.Cryptography.Cng/tests/CreateTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            CngAlgorithm alg = CngAlgorithm.ECDiffieHellmanP256;
+            using (CngKey key = CngKey.Create(alg, null, null))
+            {
+            }
+        }
+
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void CreateEphemeral2()
```
bartonjs: Maybe this is CreateEphemeral_WithParameters ?

src/System.Security.Cryptography.Cng/tests/CreateTests.cs
bartonjs: Should these tests be [Theory] over various CngAlgorithms?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngUIPolicy.cs
AtsushiKan: Here’s the MSDN doc for the native CNG version of this struct member. Not sure it’s an improvement. The docs on the BCL crypto apis in general assume a knowledge of the underlying Windows crypto apis.


A pointer to a null-terminated Unicode string that contains the text that will be used in the title of the strong key dialog box when the key is completed. If this member is NULL, a default creation title will be used in the strong key dialog box. This member is only used on key finalization.

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using Test.Cryptography;
+using Xunit;
+
+namespace System.Security.Cryptography.Cng.Tests
+{
+    public static class ImportExportTests
+    {
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void TestImportExport1()
```
bartonjs: This test feels like it's at least three things:
1) key.Handle.Dispose() doesn't affect reading properties (maybe this isn't important, but I can't think of why else it's the first thing done)
2) That the default import results in all those properties being what you expect
3) That Export produces the same answer that Import did.

Can you please break them up into three named tests?

And not that it belongs here, but somewhere we should have a test that does
```C#
SafeNCryptKeyHandle keyHandle1 = key.Handle;
SafeNCryptKeyHandle keyHandle2 = key.Handle;

Assert.NotSame(keyHandle1, keyHandle2);
```

src/System.Security.Cryptography.Cng/tests/OpenTests.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public static class OpenTests
+    {
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void OpenFromHandle1()
+        {
+
+            byte[] reExported;
+            using (CngKey keyOriginal = CngKey.Import(TestData.Key1, CngKeyBlobFormat.GenericPublicBlob))
```
bartonjs: Based on reading the code here I'm assuming that `TestData.Key1` is an ECDH-P256 key; but it'd be nice if that was included in the name.

src/System.Security.Cryptography.Cng/tests/CreateTests.cs
AtsushiKan: I don’t know what the [Theory] attribute does but given that this is a lightweight interop layer over CNG, I think it’s a second order priority for the CoreFx tests to test all the algorithms (we’d be testing the underlying CNG platform rather than getting code coverage on CoreFx itself.)

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Win32.SafeHandles;
+using Test.Cryptography;
+using Xunit;
+
+namespace System.Security.Cryptography.Cng.Tests
+{
+    public static class ImportExportTests
+    {
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
```
bartonjs: You can leave the [PlatformSpecific] attributes if you like; but the best bet would be to mark the whole test project with the UnsupportedPlatforms attribute, like its CSP brother.

src/System.Security.Cryptography.Cng/tests/OpenTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                SafeNCryptKeyHandle keyHandle = keyOriginal.Handle;
+                Assert.NotNull(keyHandle);
+                CngKey key = CngKey.Open(keyHandle, CngKeyHandleOpenOptions.EphemeralKey);
```
bartonjs: Why isn't `key` in a `using` statement?

src/System.Security.Cryptography.Cng/tests/OpenTests.cs
```diff
@@ -0,0 +1,51 @@
...
+                keyHandle.Dispose();
+
+                Assert.Equal(CngAlgorithm.ECDiffieHellmanP256, key.Algorithm);
```
bartonjs: It feels like really these should be comparing keyOriginal to key, not key to "the right answer" (since "the right answer" was already a different test)

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
AtsushiKan: The test project is marked with UnsupportedPlatforms. I was following the *.Csp test code pattern here.

src/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -0,0 +1,67 @@
...
+               key.SetProperty(p);
+
+               CngProperty p2 = key.GetProperty(propertyName, CngPropertyOptions.CustomProperty);
+               Assert.Equal(propertyName, p2.Name);
+               Assert.Equal(CngPropertyOptions.CustomProperty, p2.Options);
+
+               // This one is odd. CNG keys can have properties with zero length but CngKey.GetProperty() transforms this into null.
+               Assert.Equal(null, p2.GetValue());
+
+               Assert.Throws<CryptographicException>(() => key.GetProperty("DOES NOT EXIST", CngPropertyOptions.CustomProperty));
```
bartonjs: This feels like a distinct test (GetNotSpecifiedProperty / GetProperty_NotSpecified)

src/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -0,0 +1,67 @@
...
+               bool hasProperty;
+
+               hasProperty = key.HasProperty(propertyName, CngPropertyOptions.CustomProperty);
+               Assert.False(hasProperty);
+
+               key.SetProperty(new CngProperty(propertyName, new byte[0], CngPropertyOptions.CustomProperty));
+               hasProperty = key.HasProperty(propertyName, CngPropertyOptions.CustomProperty);
+               Assert.True(hasProperty);
+
+               key.SetProperty(new CngProperty(propertyName, new byte[500], CngPropertyOptions.CustomProperty));
```
bartonjs: Is this testing overwrite property? Should that be its own case?

src/System.Security.Cryptography.Cng/tests/ImportExportTests.cs
bartonjs: Oh, that just means that I forgot to remove those attributes when I broke it up from the unified RSA library.  That's my bad.  Opened #2626 against myself to do that cleanup.

src/System.Security.Cryptography.Cng/System.Security.Cryptography.Cng.sln
bartonjs: @stephentoub That was my bad.  Presumably I should do the equivalent to the CSP and OpenSSL libraries?

bartonjs:
@dotnet-bot test this please

bartonjs:
Through https://github.com/dotnet/corefx/commit/617e040bc361661006b21db220102a13a67e4831 LGTM; but you should probably squash this into a single commit before merging.

The failed test was a Linux process start test, which already has an issue (#2613).

PR #2620, opened 2015-08-04
When building other platforms, skip tests

JeremyKuhne:
When building other platforms on Windows, skip tests.

@weshaggard, @stephentoub

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -236,6 +236,11 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <!-- Don't run tests if we're building another platform's binaries on Windows -->
+    <SkipTests Condition="'$(SkipTests)'=='' and ('$(OsEnvironment)'=='Windows_NT' and '$(TargetsWindows)'!='true')">true</SkipTests>
+  </PropertyGroup>
+
+  <PropertyGroup>
     <!-- Work around known Dev14 bug - see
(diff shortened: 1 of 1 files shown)


mellinoe:
LGTM, this is always kind of annoying

stephentoub:
LGTM

ellismg:
Is the test layout still built?

stephentoub:
I was assuming it would be if you specified ```/p:SkipTests=false``` explicitly.  Jeremy?

JeremyKuhne:
Test binaries are still built.

ellismg:
Given that, I think after this lands we should move run-tests.sh to start consuming the test builds from the Linux and OSX builds.  Then we can remove Pat's hack to get OS specific test binaries.  Thoughts?

JeremyKuhne:
Could you clarify why it's pulling from the Windows to begin with? I'm not following why that was done.

ellismg:
When we first stood this up, test layouts were only built when running the tests, so building a Linux or OSX layout would cause test failures (since we'd try to run the tests on Windows, at build time).  I think when Nick got tests working in VS again, a few weeks ago, he moved the layout generation to earlier in the build process, so now they are always built, even if we don't want to run them.

JeremyKuhne:
Ah, I see. So that could be removed even without this.

JeremyKuhne:
@dotnet-bot test this please

PR #2623, opened 2015-08-04
Delete 'Assert.None' helper methods

mellinoe:
This helper class contains wrappers for all of the assert methods. "None" was removed in the latest version of xUnit because it was identical to "DoesNotContain". To fix our helper code here, I just switched the calls to "DoesNotContain". Instead, it makes more sense to just get rid of the "None" helpers altogether and only wrap "DoesNotContain."

EDIT: I just removed them, all of the "DoesNotcontain" overloads are covered by existing wrappers.

### DIFF
--- a/src/Common/tests/System/Diagnostics/AssertWithCallerAttributes.cs
+++ b/src/Common/tests/System/Diagnostics/AssertWithCallerAttributes.cs
@@ -274,27 +274,6 @@ internal static class Assert
         catch (Exception e) { throw WrapException(e, path, line); }
     }
 
-    public static void None(IEnumerable collection, object expected,
-        [CallerFilePath] string path = null, [CallerLineNumber] int line = 0)
-    {
-        try { Xunit.Assert.DoesNotContain(expected, (IEnumerable<object>)collection); }
-        catch (Exception e) { throw WrapException(e, path, line); }
-    }
-
-    public static void None<T>(IEnumerable<T> collection, Predicate<T> predicate,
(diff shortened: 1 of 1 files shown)


mellinoe:
@stephentoub

src/Common/tests/System/Diagnostics/AssertWithCallerAttributes.cs
```diff
@@ -115,6 +115,20 @@ internal static class Assert
...
     }
 
+    public static void DoesNotContain(IEnumerable collection, object expected,
+        [CallerFilePath] string path = null, [CallerLineNumber] int line = 0)
+    {
+        try { Xunit.Assert.DoesNotContain(expected, (IEnumerable<object>)collection); }
+        catch (Exception e) { throw WrapException(e, path, line); }
+    }
+
+    public static void DoesNotContain<T>(IEnumerable<T> collection, T expected,
```
stephentoub: The purpose of these wrappers is to have the exact same signature as the xunit ones so they can be drop-in replacements.  The order of the arguments should be the same, there should be a one-to-one correspondence between the wrapper methods and the xunit methods being called, etc.

mellinoe: I think there may have been a couple of "Assert.None" overloads that had parameters in different orders. I deleted the ones that didn't directly wrap DoesNotContain calls.

stephentoub:
LGTM. Thanks.

mellinoe:
Thanks, Stephen

PR #2625, opened 2015-08-04
Simplify traversal targets

JeremyKuhne:
Create a helper target to simplify traversal targets.

@weshaggard

### DIFF
--- a/dir.traversal.targets
+++ b/dir.traversal.targets
@@ -1,93 +1,68 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Target Name="BuildAllProjects">
+  <Target Name="IterateProjects">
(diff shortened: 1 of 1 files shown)


dir.traversal.targets
```diff
@@ -1,93 +1,68 @@
...
-
-    <!-- Given we ErrorAndContinue we need to propagate the error if the overall task failed -->
-    <Error Condition="'$(MSBuildLastTaskResult)'=='false'" />
   </Target>
 
   <PropertyGroup>
     <TraversalBuildDependsOn>
-      BuildAllProjects;
+      BuildAllProperties;
+      IterateProjects;
```
weshaggard: If someone does a rebuild which will trigger all of these targets I don't believe IterateProjects will get called for anything beyond the first one that gets triggered.

JeremyKuhne: I'll verify & update, thanks.

JeremyKuhne:
Meh, in order to call more than once you need to MSBuild yourself. Shelving this idea for now.

PR #2628, opened 2015-08-05
Fix UnixFileSystem.MaxDirectoryName to be MaxPath

### LINKED ISSUE #2627: CreateDirectory on Unix is failing for directory paths longer than MaxDirectoryName
We have a mixup of terminology leading to the bug.  MaxDirectoryName is defined to be MaxName (i.e. the max size of a component in a path), but then in CreateDirectory we're using it to mean the maximum  full path of a directory, causing us to throw when creating directory paths longer than ~255 characters.

stephentoub:
UnixFileSystem's override of MaxDirectoryName was assuming the property meant the maximum size of the file name of a directory, i.e. the component length, not the full path length of everything from the beginning of the root.  As a result, checks against MaxDirectoryName, such as in CreateDirectory, were failing incorrectly with longer path lengths.

The fix is just to return MaxPath from MaxDirectoryName; there's no difference on Unix.  I also added a test to cover this case, included a previously excluded test, and deleted a faulty test.

Fixes #2627

### DIFF
--- a/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
+++ b/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
@@ -233,7 +233,6 @@ namespace System.IO.FileSystem.Tests
 
         [Fact]
         [PlatformSpecific(PlatformID.AnyUnix)]
-        [ActiveIssue(645)]
         public void UnixPathLongerThan256_Allowed()
         {
             DirectoryInfo testDir = Create(GetTestFilePath());
@@ -245,11 +244,14 @@ namespace System.IO.FileSystem.Tests
 
         [Fact]
         [PlatformSpecific(PlatformID.AnyUnix)]
-        public void UnixPathLongerThan256_Throws()
+        public void UnixPathWithDeeplyNestedDirectories()
         {
-            DirectoryInfo testDir = Create(GetTestFilePath());
-            PathInfo path = IOServices.GetPath(testDir.FullName, 257, IOInputs.MaxComponent);
-            Assert.Throws<PathTooLongException>(() => Create(path.FullPath));
+            DirectoryInfo parent = Create(GetTestFilePath());
+            for (int i = 1; i <= 100; i++) // 100 == arbitrarily large number of directories
+            {
+                parent = Create(Path.Combine(parent.FullName, "dir" + i));
+                Assert.True(Directory.Exists(parent.FullName));
+            }
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @sokket, @ianhays, @JeremyKuhne

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -245,11 +244,14 @@ public void UnixPathLongerThan256_Allowed()
...
         [Fact]
         [PlatformSpecific(PlatformID.AnyUnix)]
-        public void UnixPathLongerThan256_Throws()
+        public void UnixPathWithDeeplyNestedDirectories()
         {
-            DirectoryInfo testDir = Create(GetTestFilePath());
-            PathInfo path = IOServices.GetPath(testDir.FullName, 257, IOInputs.MaxComponent);
-            Assert.Throws<PathTooLongException>(() => Create(path.FullPath));
+            DirectoryInfo parent = Create(GetTestFilePath());
+            for (int i = 1; i <= 200; i++) // 200 == arbitrarily large number of directories
```
JeremyKuhne: BSD/Mac is 1024- this will bust I assume?

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -245,11 +244,14 @@ public void UnixPathLongerThan256_Allowed()
 
         [Fact]
         [PlatformSpecific(PlatformID.AnyUnix)]
-        public void UnixPathLongerThan256_Throws()
+        public void UnixPathWithDeeplyNestedDirectories()
```
JeremyKuhne: Do we have a test for directory with MaxName? (/<255chars>/)

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
stephentoub: I can make it 100 instead of 200.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
stephentoub: Yes, there's a test named something like DirectoryWithComponentLongerThanMaxComponent.

sokket:
LGTM

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
JeremyKuhne: Great- LGTM with that change

stephentoub: Thanks for the catch.

stephentoub:
Thanks, guys.

PR #2629, opened 2015-08-05
Remove [PlatformSpecific] from the Windows-only CSP library

### LINKED ISSUE #2626: Remove redundant [PlatformSpecific] constraints in CSP tests
The System.Security.Cryptography.Csp library is Windows-only, and the test project is marked as Windows-Only; it doesn't need to also say [PlatformSpecific] on every test (and it doesn't on "every" test, since it includes the RSA suite)

bartonjs:
The library is marked as being unsupported for platforms other than Windows, the [PlatformSpecific] attributes were holdovers from when this was in the former unified RSA library.

Fixes #2626.


stephentoub:
LGTM

PR #2630, opened 2015-08-05
Change System.IO.FileSystem.Tests to use FileCleanupTestBase

stephentoub:
When I rewrote the MemoryMappedFiles tests, I cribbed from the
System.IO.FileSystem test's base class and created a Common base class
that could be used by test projects that had temporary files to be cleaned up.
At the time, however, FileSystem's tests were undergoing heavy changes and
so I didn't update FileSystem's tests to use the common base.  Now that
things have quieted down, I've made that change.

In doing so, I made a few tweaks to the FileSystem tests, where tests were
assuming that the test directory was always in the current directory, and
the consolidated base class puts temporary files in the user's temporary
directory instead.

### DIFF
--- a/src/Common/tests/System/IO/FileCleanupTestBase.cs
+++ b/src/Common/tests/System/IO/FileCleanupTestBase.cs
@@ -6,13 +6,17 @@ using System.Runtime.CompilerServices;
 namespace System.IO
 {
     /// <summary>Base class for test classes the use temporary files that need to be cleaned up.</summary>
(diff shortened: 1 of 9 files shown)


stephentoub:
cc: @ianhays

stephentoub:
@dotnet-bot test this please

ianhays:
LGTM!

stephentoub:
Thanks, Ian.

PR #2631, opened 2015-08-05
Fix TestUserCredentialsPropertiesOnWindows without admin privileges

stephentoub:
The System.Diagnostics.Process test TestUserCredentialsPropertiesOnWindows has two problems:
1) It requires admin privileges, and it looks like the intention was to silently bail from the test if such privileges weren't available, but instead it's causing the test to fail, which means outer loop tests fail when run without admin privileges.
2) When the exception is thrown, it's getting the error code from the wrong location: NetUserAdd returns the error code rather than having it available from GetLastWin32Error.

This commit addresses (2) simply by passing the result to Win32Exception, and it addresses (1) by catching the exception and returning instead of allowing it to propagate.

Eventually we'll need a better solution for such tests, but for now this should suffice.


stephentoub:
cc: @Priya91

Priya91:
LGTM

stephentoub:
Thanks, Priya.

PR #2633, opened 2015-08-05
Add tests for Process.WaitForExit

stephentoub:
Our Unix implementation of Process.WaitForExit is relatively complicated, as we need to support functionality implicitly supported by Windows that's not built-in on Unix.  This commit adds some tests to more heavily exercise Process waiting; I plan to add some more subsequently, but I wanted to get these in now.


stephentoub:
cc: @Priya91

Priya91:
LGTM

PR #2637, opened 2015-08-05
Add back support for reading PEM files as an X509Certificate2.

### LINKED ISSUE #2634: new X509Certificate2(byte[]) no longer works for PEM data on Linux
https://github.com/dotnet/corefx/commit/a8f70ba3814aaf6f324af278306a4ad4222bc376 added support for PFX, but the same change fails to return the ICertificatePal after successfully decoding as PEM; so it falls back to DER and PFX, both of which fail to decode, because it was PEM.

bartonjs:
The code did all of the work for reading the PEM-formatted certificate, but then didn't return the object.

Added TestByteArrayConstructor_PEM, and broke the existing TestByteArrayConstructor into TestByteArrayConstructor_DER and UseAfterDispose; because it was really doing two separate things.

Fixes #2634.


stephentoub:
LGTM

PR #2638, opened 2015-08-05
Fix for Process.TestStartTime flakiness.

Priya91:
#2613 

cc @stephentoub

### DIFF
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
@@ -133,10 +133,12 @@ namespace System.Diagnostics.ProcessTests
         {
             DateTime timeBeforeProcessStart = DateTime.UtcNow;
             Process p = CreateProcessInfinite();
-            StartAndKillProcessWithDelay(p);
+            p.Start();
+            Assert.Throws<InvalidOperationException>(() => p.ExitTime);
+            p.Kill();
+            Assert.True(p.WaitForExit(WaitInMS));
             Assert.True(p.ExitTime.ToUniversalTime() > timeBeforeProcessStart, "TestExitTime is incorrect.");
         }
-
 
         [Fact]
         public void TestId()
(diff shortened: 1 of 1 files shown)


stephentoub:
Thanks, Priya. Can you elaborate on what the problem was and how this fixes it?

Priya91:
> Can you elaborate on what the problem was and how this fixes it?

Sure. The test was always failing with the Process.StartTime value being greater that the afterTicks value. This could be because:

1. p.Start()
2. long afterTicks = DateTime.Now.Ticks
3. OS actually starts process now
4. Check p.StartTime.Ticks < afterTicks => which will be false.

With the proposed change I am ensuring the Process was actually started by the OS, by checking if there is a running process with p.Id using Process.GetProcessById().

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
```diff
@@ -133,8 +133,21 @@ public void TestExitTime()
...
-            Assert.True(p.ExitTime.ToUniversalTime() > timeBeforeProcessStart, "TestExitTime is incorrect.");
+            try
+            {
+                p.Start();
+                Assert.Throws<InvalidOperationException>(() => p.ExitTime);
+                p.Kill();
+                Assert.True(p.WaitForExit(WaitInMS));
+                Assert.True(p.ExitTime.ToUniversalTime() > timeBeforeProcessStart, "TestExitTime is incorrect.");
+            }
+            finally
```
stephentoub: What's the purpose of this finally block?  I thought we weren't doing things like this in individual tests and letting the test class' Dispose clean them up.  Does this also need to be done here for some reason?

stephentoub:
One question, otherwise LGTM.

src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
Priya91: Yes, we don't need this. CreateProcessInfinite adds the process to the list of processes to be disposed.

PR #2641, opened 2015-08-05
 Fix all TimeZoneInfo tests on Linux. 

### LINKED ISSUE #2465: Update tzfile parsing in TimeZoneInfo to V2 and create AdjustmentRules correctly for Linux
I’ve come across a big difference in the Time Zone adjustment rules between Windows and Linux.
 
On Windows, the public type TimeZoneInfo.AdjustmentRule defines the structure of the data stored in the registry.
 
Basically, time zone adjustment rules are 100% geared around daylight savings time.
The AdjustmentRule defines:
1.	A span of time (DateStart to DateEnd) that the rule is valid
2.	An offset (DaylightDelta) from the BaseUtcOffset that should be used when daylight savings time is in effect.
3.	And then 2 “TransitionTime” structures DaylightTransitionStart and DaylightTransitionEnd, that describe when inside the span of time (1.) that Daylight savings time is in effect.

However, on Linux, time zone information is stored differently and isn’t based solely on daylight savings time.  It shows every single time the time changed in the time zone.  For the US, you get 2 entries for every year in time zones that support DST.  The last 3 time transitions in Moscow haven't been logged as daylights savings transitions, as they were supposed to be permanent transitions:

Transition Date | Transition Offset
---------------------|---------------------
03/27/2010 23:00:00 | 04:00:00 DST
10/30/2010 23:00:00 | 03:00:00
03/26/2011 23:00:00 | 04:00:00
10/25/2014 22:00:00 | 03:00:00

The structure of data in the tzfile is:
1.	A “Transition Date”, which is the date and time that the time offset changed
2.	A “Transition Offset”, which is the offset from UTC that went into effect on the Transition Date
3.	A value whether this offset is to be considered “daylight savings time” or not.

The issue is, given the structure of AdjustmentRule, there is no way to express these types of offset transitions, since AdjustmentRule is based on daylight savings time.

To solve this, we need to add a "StandardDelta" property to AdjustmentRule that can be used when a non-daylight savings transition was made.  Then we need to update the AdjustmentRule creation code for Linux to correctly fill out a new AdjustmentRule for each transition.
To handle DateTimes less than the first transition time, we should make an AdjustmentRule based on the first standard-time ttinfo structure in the file.
To handle DateTimes greater than the last transition time, we need to parse the V2 information, which has the POSIX style rule:
http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html
format: std offset dst [offset],start[/time],end[/time].
ex. EST+5EDT,M3.2.0/2,M11.1.0/2
For V1 tzfiles, we should just use the last transition noted in the file for dates greater than what is expressed.

eerhardt:
Now that TimeZoneInfo is correctly parsing and creating adjustment rules on Linux, remove all [ActiveIssue] attributes in the TimeZoneInfo tests and make them pass on Linux.

There are a few cases where Windows and Linux rules differ.  For instance, Linux will use local mean time for DateTimes before standard time came into affect.  Also, there are some Australia time zones that changed rules that Windows couldn't/didn't support.

Also added a new test for "multi-year daylight savings" in Argentina from 1946-1963.

Fix #2465

@ellismg, @tarekgh

### DIFF
--- a/src/System.Runtime/tests/System/TimeZoneInfo.cs
+++ b/src/System.Runtime/tests/System/TimeZoneInfo.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Globalization;
+using System.Runtime.InteropServices;
 using Xunit;
 
 public static class TimeZoneInfoTests
@@ -20,8 +21,8 @@ public static class TimeZoneInfoTests
     private static String s_strRussian = s_isWindows ? "Russian Standard Time" : "Europe/Moscow";
     private static String s_strLibya = s_isWindows ? "Libya Standard Time" : "Africa/Tripoli";
     private static String s_strCatamarca = s_isWindows ? "Argentina Standard Time" : "America/Catamarca";
-    private static String s_strLisbon = Interop.IsWindows ? "GMT Standard Time" : "Europe/Lisbon";
-    private static String s_strNewfoundland = Interop.IsWindows ? "Newfoundland Standard Time" : "America/St_Johns";
+    private static String s_strLisbon = s_isWindows ? "GMT Standard Time" : "Europe/Lisbon";
+    private static String s_strNewfoundland = s_isWindows ? "Newfoundland Standard Time" : "America/St_Johns";
 
     private static TimeZoneInfo s_myUtc = TimeZoneInfo.Utc;
     private static TimeZoneInfo s_myLocal = TimeZoneInfo.Local;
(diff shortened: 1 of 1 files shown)


stephentoub:
Once the coreclr changes go in and the tests pass, LGTM.

eerhardt:
FYI - I added another commit containing one more test that uncovered a bug I found in TimeZoneInfo on Linux.

stephentoub:
@dotnet-bot test this please

stephentoub:
@eerhardt, some conflicts need to be resolved before this can be merged.  Thanks.

PR #2642, opened 2015-08-06
Improve process launching support in Process tests

stephentoub:
This started out with my wanting to add several more complicated Process.WaitForExit tests.  However, in order to do so I overhauled how we handle process launching in our System.Diagnostics.Process.Tests.

Today in src\System.Diagnostics.Process\tests, there are two subfolders, one for a normal tests assembly, and another that houses a console app that's invoked by the Process tests.  The logic for the tests is therefore split across these two projects, and yet for the tests to be successful, the logic must be carefully coordinated.  A similar situation arose when I was writing the interprocess communication tests, and for that I put together a remote executor console app that can execute arbitrary code, specified to it via assembly/type/method/argument information passed to it as command-line arguments.

The first commit in this PR moves that console app and associated launcher out of the interprocess communication tests and into Common, making it available for any project that wants to use it.  In the process, I cleaned it up a bit further, and added some additional functionality, such as in addition to support for static method, (basic) support for instance methods, which enables lambdas without closures (the VS2015 C# compiler generates such lambdas as instance methods rather than as statics).

The second commit then takes advantage of that in the Process tests.  It deletes the old test console app used by the Process tests, and updates the tests to use the common launcher.  In doing, I moved all relevant logic close together, got rid of the strings that were used to coordinate across the assemblies, and in some cases took advantage of lambdas to make it very simple to specify the remoted logic.

The third commit then finally adds some of the additional Process.WaitForExit tests I wanted to add.


stephentoub:
cc @Priya91, @pallavit, @mellinoe

stephentoub:
@dotnet-bot test this please

pallavit:
@stephentoub LGTM. Thanks for doing this :).

stephentoub:
Thanks for the feedback, Pallavi and Priya.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #2643, opened 2015-08-06
Use CopyTo where possible in InsertRange

jamesqo:

The logic has been changed to use `CopyTo` for enumerables that impl `ICollection` and `ICollection<T>`, which includes arrays. This also affects enumerables that don't inherit from them, since while getting the count of an `IEnumerable` we capture it into a `List<T>` (which impls `ICollection<T>`) to avoid enumerating over it more than once.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -549,6 +549,9 @@ namespace System.Collections.Immutable
                 return ImmutableArray.CreateRange(items);
             }
 
+            // If items is not any of the interfaces below, it will
+            // be 'captured' in a List<T> here and travel
+            // down the ICollection<T> path.
             int count = ImmutableExtensions.GetCount(ref items);
             if (count == 0)
             {
@@ -556,13 +559,11 @@ namespace System.Collections.Immutable
             }
 
             T[] tmp = new T[self.Length + count];
-            Array.Copy(self.array, 0, tmp, 0, index);
-            int sequenceIndex = index;
-            foreach (var item in items)
-            {
-                tmp[sequenceIndex++] = item;
-            }
 
+            // CopyTo first in case bad interface impl
+            // tries to modify other parts of tmp
+            items.CopyTo(tmp, index);
+            Array.Copy(self.array, 0, tmp, 0, index);
             Array.Copy(self.array, index, tmp, index + count, self.Length - index);
             return new ImmutableArray<T>(tmp);
         }
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
@@ -87,6 +87,41 @@ namespace System.Collections.Immutable
         }
 
         /// <summary>
+        /// Copies part of an <see cref="IEnumerable{T}"/> to an array, starting at the specified index.
+        /// </summary>
+        /// <typeparam name="T">The type of element.</typeparam>
+        /// <param name="sequence">The sequence of elements you want to copy.</param>
+        /// <param name="dest">The destination array.</param>
+        /// <param name="index">The index in the array to start copying to.</param>
+        internal static void CopyTo<T>(this IEnumerable<T> sequence, T[] dest, int index)
+        {
+            Requires.NotNull(sequence, "sequence");
+            Requires.NotNull(dest, "dest");
+            Requires.Range(index >= 0 && index < dest.Length, "index");
+            
+            var collectionOfT = sequence as ICollection<T>;
+            if (collectionOfT != null)
+            {
+                collectionOfT.CopyTo(dest, index);
+                return;
+            }
+            
+            var collection = sequence as ICollection;
+            if (collection != null)
+            {
+                collection.CopyTo(dest, index);
+                return;
+            }
+            
+            // Fallback to manual iteration over dest and sequence
+            int sequenceIndex = index;
+            foreach (var item in sequence)
+            {
+                dest[sequenceIndex++] = item;
+            }
+        }
+
+        /// <summary>
         /// Gets a copy of a sequence as an array.
         /// </summary>
         /// <typeparam name="T">The type of element.</typeparam>
--- a/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
+++ b/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
@@ -1375,9 +1375,46 @@ namespace System.Collections.Immutable.Test
             Assert.Throws<NotSupportedException>(() => c.SyncRoot);
         }
 
+        ////[Fact] // Perf test
+        public void InsertRange_Perf_Arrays()
+        {
+            TestInsertRangePerfWith(Enumerable.Range(1, 10000).ToArray());
+        }
+        
+        ////[Fact] // Perf test
+        public void InsertRange_Perf_ListOfT()
+        {
+            TestInsertRangePerfWith(Enumerable.Range(1, 10000).ToList());
+        }
+        
+        ////[Fact] // Perf test
+        public void InsertRange_Perf_NonCollectionEnumerable()
+        {
+            // IEnumerable being tested here must not inherit from
+            // ICollection<T>, ICollection, or IReadOnlyCollection<T>.
+            
+            TestInsertRangePerfWith(Enumerable.Range(1, 10000));
+        }
+
         protected override IEnumerable<T> GetEnumerableOf<T>(params T[] contents)
         {
             return ImmutableArray.Create(contents);
+        }
+
+        private static void TestInsertRangePerfWith<T>(IEnumerable<T> toInsert)
+        {
+            var iarray = ImmutableArray<T>.Empty;
+            
+            var watch = new Stopwatch();
+            
+            for (int i = 0; i < 3; i++)
+            {
+                watch.Start();
+                iarray = iarray.InsertRange(0, toInsert);
+                watch.Stop();
+                Debug.WriteLine(watch.Elapsed);
+                watch.Reset();
+            }
         }
 
         /// <summary>


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -442,7 +442,7 @@ public int LastIndexOf(T item, int startIndex, int count, IEqualityComparer<T> e
             }
             else
             {
-                for (int i = startIndex; i >= startIndex - count + 1; i--)
+                for (int i = startIndex; i > startIndex - count; i--)
```
AArnott: What does this change have to do with the perf improvement? If it is just style, it should probably be separated. It may be easier reading, but it takes time to validate that this doesn't introduce a subtle regression so in the interest of increasing the chances of taking the pull request on its perf gains it's best to avoid distractions

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -556,13 +559,25 @@ public ImmutableArray<T> InsertRange(int index, IEnumerable<T> items)
...
+            if (items is ICollection<T>)
             {
-                tmp[sequenceIndex++] = item;
+                ((ICollection<T>)items).CopyTo(tmp, index);
             }
-
+            else
+            {
+                // items is IReadOnlyCollection<T>
+
```
AArnott: stylecop: there shouldn't be a blank line after a single-line comment.
(corefx folks can comment as to whether this is an enforced rule)

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -556,13 +559,25 @@ public ImmutableArray<T> InsertRange(int index, IEnumerable<T> items)
...
-            Array.Copy(self.array, 0, tmp, 0, index);
-            int sequenceIndex = index;
-            foreach (var item in items)
+            
+            // CopyTo first in case bad interface impl
+            // tries to modify other parts of tmp
+            if (items is ICollection<T>)
             {
-                tmp[sequenceIndex++] = item;
+                ((ICollection<T>)items).CopyTo(tmp, index);
```
AArnott: You now have two type checks here: the one in the if condition and the type cast here. While there is a *small* perf improvement for only doing the type check once, I doubt it's going to be significant for this method. But I'm more interested in the readability/maintainability of the code if you use `as` for the cast above the `if` line, and then the `if` check can be whether the result is non-null. That way one line of code being changed can't result in accidental `InvalidCastException`s down the road.

AArnott:
What is the perf improvement from this change? Can you share your test cases and measurements with us? It seems the significant change (if any) would come from the fact that a collection's implementation of `ICollection<T>.CopyTo` would be significantly faster than enumerating using `IEnumerable<T>` to enumerate the collection. Is this actually significant in some case?

James-Ko:
@AArnott Arrays implement `ICollection<T>`, so if `items` happens to be an array, then it would essentially be the same thing as an `Array.Copy`, which is much much faster than `foreach`.

If the enumerable inherits from none of these collection interfaces, then it will get converted to a `List<T>`, and calling `CopyTo` on one of those does a straight `Array.Copy` operation [as well](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L388).

AArnott:
Thanks. Did you take any perf measurements? Can you share your findings?
While I believe you know what you're talking about, we've learned that optimizing for perf really should be backed up with real measurements because unintended side-effects of your change can negate your intended improvement; or some unusual trait about the code makes the normally faster approach not actually faster. Finally, when you measure, and collect perf traces, you can not only see how much faster you may have made it, but also if there are any other obvious wins to be had.

James-Ko:
@AArnott Sure, I'll write up some tests next thing in the morning.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -85,6 +85,38 @@ internal static int GetCount<T>(ref IEnumerable<T> sequence)
...
         }
+        
+        /// <summary>
+        /// Copies part of an IEnumerable&lt;T&gt; to an array, starting at the specified index.
+        /// </summary>
+        /// <typeparam name="T">The type of element.</typeparam>
+        /// <param name="sequence">The sequence of elements you want to copy.</param>
+        /// <param name="dest">The destination array.</param>
+        /// <param name="index">The index in the array to start copying to.</param>
+        internal static void CopyTo(this IEnumerable<T> sequence, T[] dest, int index)
```
stephentoub: This is missing a generic type parameter for T and will fail to compile.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -85,6 +85,38 @@ internal static int GetCount<T>(ref IEnumerable<T> sequence)
...
+        {
+            var collectionOfT = items as ICollection<T>;
```
stephentoub: What is ```items```?  Did you mean ```sequence```?  This won't compile either.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -85,6 +85,38 @@ internal static int GetCount<T>(ref IEnumerable<T> sequence)
...
+            
+            if (collectionOfT != null)
+            {
+                collectionOfT.CopyTo(tmp, index);
```
stephentoub: Do all relevant collections implement CopyTo to copy in the same order as enumerating those collections?  If not, this would be a behavioral change.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -85,6 +85,38 @@ internal static int GetCount<T>(ref IEnumerable<T> sequence)
...
+        
+        /// <summary>
+        /// Copies part of an IEnumerable&lt;T&gt; to an array, starting at the specified index.
+        /// </summary>
+        /// <typeparam name="T">The type of element.</typeparam>
+        /// <param name="sequence">The sequence of elements you want to copy.</param>
+        /// <param name="dest">The destination array.</param>
+        /// <param name="index">The index in the array to start copying to.</param>
+        internal static void CopyTo(this IEnumerable<T> sequence, T[] dest, int index)
+        {
```
AArnott: Please add argument validation for all three parameters.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -85,6 +85,38 @@ internal static int GetCount<T>(ref IEnumerable<T> sequence)
...
+            var collectionOfT = items as ICollection<T>;
+            
+            if (collectionOfT != null)
+            {
+                collectionOfT.CopyTo(tmp, index);
+                return;
+            }
+            
+            var collection = items as ICollection;
+
```
AArnott: style: IMO it's better to omit this blank line between the conditional cast and the if that will use the result. It makes reading the method more obvious that you have three chunks of work.

src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
```diff
@@ -1374,6 +1374,42 @@ public void ICollectionSyncRoot_NotSupported()
             ICollection c = ImmutableArray.Create(1, 2, 3);
             Assert.Throws<NotSupportedException>(() => c.SyncRoot);
         }
+        
+        //[Fact] Perf test
```
AArnott: StyleCop: leave a single space after the `//` or use `////`. 
Also it would be helpful if you put a `//` in between the attribute and the `Perf test` so that folks can uncomment the attribute without introducing a compile error. So in short, can you change each of these lines to:
```csharp
    ////[Fact] // Perf test
```

src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
```diff
@@ -1374,6 +1374,42 @@ public void ICollectionSyncRoot_NotSupported()
...
+        public void InsertRange_Perf_NonCollectionEnumerable()
+        {
+            // IEnumerable being tested here must not inherit from
+            // ICollection<T>, ICollection, or IReadOnlyCollection<T>.
+            
+            TestInsertRangePerfWith(Enumerable.Range(1, 10000));
+        }
+        
+        private void TestInsertRangePerfWith(IEnumerable<T> toInsert)
```
AArnott: Can this be made `static`?
Also (stylecop) please move the private methods below the protected methods.

AArnott:
I like your recent iterations with the `CopyTo` extension method, BTW. It makes for much easier reading in both files. :+1: 
Change LGTM once you resolve Stephen's identified issues.

src/System.Collections.Immutable/tests/ImmutableArrayTest.cs
James-Ko: Sure, but all the other methods are non-static?

AArnott:
Thank you @James-Ko for those follow-up fixes.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
James-Ko: I'm sorry, but I don't understand you there; you're asking if the impl of `CopyTo` is unreliable and will result in a behavioral change from manually enumerating?

stephentoub: Not unreliabile, just with a different ordering between what CopyTo does and what enumeration does.  Consider a collection like (pseudo-code):
```C#
public class MyStack<T> : IEnumerable<T>, ICollection
{
    private T[] _array;
    private int _count;
    ...
    public void CopyTo(T[] array, int index)
    {
        Array.Copy(_array, 0, array, index, _array.Length);
    }

    public IEnumerable<T> GetEnumerable()
    {
        for (int i = _count - 1; i >= 0; i--)
        {
            yield return _array[i];
        }
    }
}
```
It's implemented so that the order of the elements that end up in the array is different if you use CopyTo versus enumerating it to fill in the array.  I don't see anything in the documentation for ICollection that dictates the order in which the elements are copied, that it must match the ordering provided by the IEnumerable implementation.

Now, maybe in practice this doesn't happen, or maybe even if it does it's so rare that it's acceptable risk.  I was pointing out that it can be an observable behavior change.

James-Ko:
@AArnott Whoops, thought you were referring to the test methods as well; I'm not familiar with XUnit, and NUnit does allow static `[TestCases]`. I'll revert those changes right now.

AArnott:
@James-Ko The rate of your compile errors that you're pushing to the server, and the test method mistakes concerns me. Are you testing these before pushing? We expect you to build and test your changes -- not just to submit PRs with changes that you haven't personally vetted.

James-Ko:
@AArnott Right. I'll try to push a little less next time; thanks for the tip.

stephentoub:
@James-Ko, can you please revert the whitespace changes and squash this?

@AArnott, you're ok with the CopyTo vs enumerable ordering issue I called out?  I'm ok with it if you are; I believe those orderings are the same for all of the collections in corefx, or at least all of the main ones, and in the rare case where there is a difference it seems unlikely to cause an issue for anyone.

AArnott:
@stephentoub I would suspect the risk is very low that CopyTo doesn't use the same order as enumeration. My one pause is `Dictionary<K,V>` itself, since its data structures are quite unusual so who knows what `CopyTo` vs. its `IEnumerator<T>` does. Do you happen to know?

The whitespace changes I was divided on. They aren't relevant, but the git community in general doesn't like trailing whitespace and our contribution guidelines forbid formatting pull requests. So when do these issues get fixed (probably all at once, I'd prefer, but who's going to do that?). It's not a big deal IMO either way. I'm just curious.

AArnott:
BTW if we are too concerned about the behavioral change of `CopyTo` rather than enumeration order, we could special case the path for those collection types that we know are the same and that would significantly benefit from the change (such as arrays).

stephentoub:
> My one pause is Dictionary<K,V> itself, since its data structures are quite unusual so who knows what CopyTo vs. its IEnumerator<T> does. Do you happen to know?

I believe it's the same, in both cases just walking from beginning to end its internal entries array yielding/copying filled cells.

stephentoub:
>  our contribution guidelines forbid formatting pull requests. So when do these issues get fixed

@weshaggard should comment on that.  I know there was talk of running some tools over the code to address this kind of thing.

weshaggard:
As for the clean-up we have various issues filed to clean-up the libraries, they just aren't very high on our priority list right now.

stephentoub:
LGTM

stephentoub:
@weshaggard, are you ok with the small potential behavioral change?  Seems low risk.

James-Ko:
Update: I've pushed a new commit with similar changes for ImmutableArray.Builder. This would be part of #2707, except that my extension method isn't available yet on other branches.

weshaggard:
@stephentoub yes I'm OK with the using CopyTo instead of the enumerator, as I agree the risk seems low.

stephentoub:
@James-Ko, did you run the tests locally?  Two tests are failing:
```
 System.Collections.Immutable.Test.ImmutableArrayBuilderTest.AddRangeIEnumerable [FAIL]
        Assert.Equal() Failure
        Expected: 1
        Actual:   0
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Collections.Immutable\tests\ImmutableArrayBuilderTest.cs(72,0): at System.Collections.Immutable.Test.ImmutableArrayBuilderTest.AddRangeIEnumerable()
     System.Collections.Immutable.Test.ImmutableArrayBuilderTest.CountContract [FAIL]
        Assert.Equal() Failure
        Expected: <ConcatIterator>d__1<Int32> [1, 2, 3, 4, 5, ...]
        Actual:   Builder<Int32> [1, 2, 3, 4, 5, ...]
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Collections.Immutable\tests\ImmutableArrayBuilderTest.cs(381,0): at System.Collections.Immutable.Test.ImmutableArrayBuilderTest.CountContract()
```

James-Ko:
@stephentoub Alright, I apologize, I haven't been building the tests locally. For some reason, building with Visual Studio 2015 results in obscure errors every time while restoring packages, and searching for them turns up nothing. Additionally, it would modify hundreds of lines of my project.json && project.lock.json while doing this, to the point where `git diff` was completely unreadable and I would have to keep resetting the changes to both files.

Also, following the [Unix build instructions](https://github.com/dotnet/corefx/blob/master/Documentation/building/unix-instructions.md) for my armhf Linux laptop results in 196 compilation errors.

stephentoub:
> For some reason, building with Visual Studio 2015 results in obscure errors every time while restoring packages, and searching for them turns up nothing. Additionally, it would modify hundreds of lines of my project.json && project.lock.json while doing this, to the point where git diff was completely unreadable and I would have to keep resetting the changes to both files.

I believe that's because your branch is significantly behind dotnet/master: "407 commits behind dotnet:master."  Things have been moving fairly quickly in the repo, in particular with regards to packages lately, so it's a good idea to keep in sync.

Thanks.

James-Ko:
@stephentoub Thanks, I'll be rebasing my `master` fairly more often after this.

I'll fix everything in ~5 days when I have access to my desktop so I can build everything before submitting.

James-Ko:
Alright, I finally managed to get the repo building, so here it is.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -246,15 +246,13 @@ public void AddRange(IEnumerable<T> items)
             {
                 Requires.NotNull(items, "items");
 
-                int count;
-                if (items.TryGetCount(out count))
+                int count = ImmutableExtensions.GetCount(ref items);
```
stephentoub: If items isn't a list, isn't this going to allocate a list and copy it all to one?  That's very expensive.

James-Ko: I was just following the same strategy as in InsertRange, but I can see your point because here it's not as necessary to know the size beforehand. I've updated my commit to use the original implementation as a fallback.

stephentoub:
@AArnott, @jaredpar, how concerned are we with code being able to get at the array that underlies an ImmutableArray without using unsafe code?  This change would seem to enable that, at least in certain cases.  Code could construct a custom ICollection whose CopyTo implementation cached away the array it's handed by the builder, which may also be the array that's then used to construct the ImmutableArray.

AArnott:
@stephentoub Ooh, good catch, Stephen. I think we've thus far avoided trusting non-sealed members with references to the T[]. For example we avoided passing arrays to Streams (at least back when ImmutableArray did some of that) since that would allow customers access to it.
However since then there has been a lot of discussion around providing a `Dangerous*` method that provides access, but I can't see any of them so perhaps we decided not to. @terrajobst do you recall what the decision there was?

If we don't have such methods, we should avoid this. We can enable it as a fast path when we are confident that the implementation is safe (such as another T[] or a specific concrete type, or a sealed type or method perhaps).

nguerrera:
We pulled the `Dangerous*` methods after much deliberation.

James-Ko:
Although it's certainly true a bad interface implementation could get at the underlying array, consider this:

- the vast majority of cases will be "safe" collections from mscorlib (e.g. arrays, lists)
- the collection would have no way to check whether the array it was copying to was a field of ImmutableArray, Builder, or just a regular array
- the reference to the array is passed by value, so a bad implementation can't set the field to null, change its length, etc. etc. to make the Builder misbehave
- if we can't trust CopyTo to do the right thing, can we trust GetEnumerator? or Dispose? Then why not stop accepting interfaces altogether if their methods can't be trusted?

Also FYI: Builder has this weakness as well if `AddRange` + `MoveToImmutable` is called, since it passes the same reference to the new ImmutableArray.

stephentoub:
> the vast majority of cases will be "safe" collections from mscorlib (e.g. arrays, lists)
> the collection would have no way to check whether the array it was copying to was a field of ImmutableArray, Builder, or just a regular array
> the reference to the array is passed by value, so a bad implementation can't set the field to null, change its length, etc. etc. to make the Builder misbehave

The concern isn't random collections, it's that a backdoor like this essentially exposes state that's meant to never be exposed, e.g.
```C#
public static class NefariousCode
{
    public static T[] ExtractUnderlyingArray<T>(ImmutableArray<T>.Builder builder)
    {
        var thief = new ThievingCollection<T>();
        builder.AddRange(thief);
        return thief.StolenArray;
    }

    private class ThievingCollection<T> : ICollection<T>
    {
        internal T[] StolenArray;
        public void CopyTo(T[] array, int arrayIndex) { StolenArray = array; }
        ...
    }
}
```

> if we can't trust CopyTo to do the right thing, can we trust GetEnumerator? or Dispose? Then why not stop accepting interfaces altogether if their methods can't be trusted?

As shown above, the problem is that the code in this PR is giving to an interface method of unknown origin state that's never meant to be given out.

hackcraft:
One matter that affects the balance of risk, is just how nefarious `NefariousCode` would have to be to do this. If there's a reason, even a very foolish reason, why good-faith code might do something like that then the risk is much greater.

On the other hand, if the only reason for doing such a thing is to deliberately mess things up, then it could also do worse things through reflection, so perhaps a check for `AppDomain.IsFullyTrusted` could allow the optimisation on the basis that any code that is abusing it isn't given any opportunities it doesn't already have.

(But in changing the number of checks made before attempting the fastpath the performance examination would have to be done again).

James-Ko:
@hackcraft Unfortunately, `AppDomain` Isn't available in WinRT so it can't be used in .NET Core.

James-Ko:
Alright, finished. I did about 4 type checks in my new SafeCopyTo extension method though, so for collections that do not implement the list interfaces I don't know if it's worth it.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs
```diff
@@ -37,17 +37,17 @@ internal static bool TryGetCount<T>(this IEnumerable<T> sequence, out int count)
...
         {
-            var collection = sequence as ICollection;
-            if (collection != null)
+            var collectionOfT = sequence as ICollection<T>;
+            if (collectionOfT != null)
             {
-                count = collection.Count;
+                count = collectionOfT.Count;
                 return true;
             }
```
James-Ko: Not necessarily related to this PR, but I couldn't help but changing the order here since the generic ICollection is by far more common than non-generic. Is this OK?

hackcraft: IIRC, If you have an array, accessing it's implementation if `ICollection` is faster than accessing its' `ICollection<T>`, and faster still than access its `ICollection<T>` if it is implementing it due to array variance (it isn't actually a a `T[]` but something assignable to `T[]`). Therefore unless you are going to make use of another method or property of `ICollection<T>` at this point (like it's typed `CopyTo`), then the order that was there is probably better in that case.

If something implemented `ICollection<T>` and also `ICollection` it makes no odds.

If something only implements `ICollection<T>` it's probably still faster to find out that it doesn't, than to get `ICollection<T>` for an array.

So that's the details I can think of affecting the balance. Lacking data on actual performance, I'd lean toward having it as it was.

James-Ko:
Good to go? The builds are failing but that's because MSBuild isn't in their PATH for some reason.

stephentoub:
> Good to go?

I'm not convinced at this point that it's really a net win, but I'll let @AArnott chime in on whether he things it's worth taking or whether there are other changes to be made.

jamesqo:
@stephentoub Seeing as you were OK with the original changes and the extension method is no longer needed for `Builder`, I've decided to pull the new changes and separate them out into a new PR.

AArnott:
The latest change doesn't eliminate the risk of exposing the inner array. I'm not saying that's the wrong design, I'm just calling it out. If the perf benefit is substantial enough, but we want to mitigate the risk, we might consider creating an intermediate array that we give out in order to use the `CopyTo` method on the source collection and then use `Array.Copy` to copy it again from the temporary collection to the collection we'll actually use for the new `ImmutableArray<T>`. That of course will eat away at least some of the pert benefit, and will cause a bit more GC pressure given we're allocating more than we were. That said, if we're eliminating allocating the `IEnumerator<T>` (which is of smaller, fixed size) then the additional object we're allocating for the temporary array is offset at least somewhat.

I don't think the added complexity from this change is too severe. So if the perf wins are there, and useful to someone, I'm in favor.

stephentoub:
> @stephentoub Seeing as you were OK with the original changes

That was before I realized we were exposing the inner array.

> we might consider creating an intermediate array that we give out in order to use the CopyTo method on the source collection and then use Array.Copy to copy it again from the temporary collection to the collection we'll actually use for the new ImmutableArray<T>.

At that point it becomes much less obvious whether there's actually a benefit or not, and we'd need to see some extensive perf tests verifying the net wins on a large number of situations and confirming no regressions, or if there are regressions, justifying why the wins outweigh the losses.

From my perspective, while I appreciate the effort that's gone into this, at this point I personally think it makes sense to close out this PR.

jamesqo:
Great, it's been more than a month so TBH I've been waiting for this. Thanks AArnott/Stephen.

stephentoub:
Thanks, @jamesqo.

PR #2644, opened 2015-08-06
Remove O(n) null check in ImmutableArray.RemoveAll

James-Ko:

Previously the `List<int>` in this method was lazily instantiated. I don't see the point of this because the default `List<T>` ctor is extremely cheap, and lazy instantiation just makes the code less readable.


stephentoub:
> I don't see the point of this because the default  List<T>  ctor is extremely cheap, and lazy instantiation just makes the code less readable.

It's not the cost of the ctor itself that drives the desire for laziness here:  it's the allocation and the cost that imposes on garbage collection. At a very high level, the more allocations there are, the more GC has to run and the longer GC takes.

stephentoub:
Thank you for submitting this, but we would need very compelling performance data to take a change like this. I'm going to close this out for now.

PR #2646, opened 2015-08-06
Don't wrap with ImmutableArray

James-Ko:

Some methods now `return array` instead of `return new ImmutableArray<T>(array)`, which is essentially the same thing.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -40,7 +40,7 @@ namespace System.Collections.Immutable
         public static readonly ImmutableArray<T> Empty = new ImmutableArray<T>(new T[0]);
 
         /// <summary>
-        /// The backing field for this instance. References to this value should never be shared with outside code. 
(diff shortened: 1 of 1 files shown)


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -942,14 +942,14 @@ public ImmutableArray<T> Sort(int index, int count, IComparer<T> comparer)
...
 
-            if (comparer == null)
-            {
-                comparer = Comparer<T>.Default;
-            }
-
             // 0 and 1 element arrays don't need to be sorted.
             if (count > 1)
             {
+                if (comparer == null)
```
stephentoub: This change to move the comparer check and setting is fine from my perspective, but it doesn't have anything to do with wrapping/unwrapping immutable arrays, right?

stephentoub:
As was mentioned in https://github.com/dotnet/corefx/pull/2397#issuecomment-122183453, the generated code is the same in both cases, so this shouldn't affect throughput positively or negatively.  But I do think it makes the code a bit easier to read, and it does result in smaller IL, both of which are good things.  So, LGTM.

@AArnott?

AArnott:
:+1:

PR #2649, opened 2015-08-06
Use RuntimeInformation across corefx for platform detection

stephentoub:
This commit deletes our temporary Interop.PlatformDetection.cs mechanism and replaces usage of it with usage of System.Runtime.InteropServices.RuntimeInformation.dll.


stephentoub:
cc: @Priya91, @ellismg 

@ianhays, there was one usage of Interop.IsLinux in a FileSystem tests file that appeared to be stale (it wasn't included in the .csproj, and it was still following the old style).  I've deleted it as part of this change; please let me know if that's not the right answer.

bartonjs:
LGTM

ianhays:
LGTM!

@stephentoub FileInfo/Create_str.cs can be safely removed. I renamed it to Create.cs and I suppose the deletion didn't get committed along with the addition.

Priya91:
LGTM

stephentoub:
Thanks, all.

PR #2597, opened 2015-08-01
HttpClient xplat: Basic request and response processing

vijaykota:
- Also addressed some of the comments in https://github.com/dotnet/corefx/issues/2555

PR #2651, opened 2015-08-06
Automatic Content Decompression

kapilash:
Modified CurlHandler to set the curl options for automatic response content decompression.

### DIFF
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.Net.Http;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -21,6 +22,8 @@ namespace System.Net.Http
         #region Constants
 
         private const string UriSchemeHttps = "https";
+        private const string EncodingNameGzip = "gzip";
+        private const string EncodingNameDeflate = "deflate";
         private readonly static string[] AuthenticationSchemes = { "Negotiate", "Digest", "Basic" }; // the order in which libcurl goes over authentication schemes
         #endregion
 
@@ -32,6 +35,7 @@ namespace System.Net.Http
         private IWebProxy _proxy = null;
         private ICredentials _serverCredentials = null;
         private ProxyUsePolicy _proxyPolicy = ProxyUsePolicy.UseDefaultProxy;
+        private DecompressionMethods _automaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
 
         #endregion
 
@@ -120,6 +124,27 @@ namespace System.Net.Http
                 {
                     throw new PlatformNotSupportedException(SR.net_http_unix_invalid_client_cert_option);
                 }
+            }
+        }
+
+        internal bool SupportsAutomaticDecompression
+        {
+            get
+            {
+                return true;
+            }
+        }
+
+        internal DecompressionMethods AutomaticDecompression
+        {
+            get
+            {
+                return _automaticDecompression;
+            }
+            set
+            {
+                CheckDisposedOrStarted();
+                _automaticDecompression = value;
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -245,11 +270,37 @@ private SafeCurlHandle CreateRequestHandle(RequestCompletionSource state)
...
+                    {
+                        curlEncoding.Append(',');
+                    }
+
+                    curlEncoding.Append(EncodingNameDeflate);
+                }
+                Interop.libcurl.curl_easy_setopt(requestHandle, CURLoption.CURLOPT_ACCEPTENCODING, curlEncoding.ToString());
+            }
+        }
+
```
stephentoub: FWIW, you could avoid the per-call StringBuilder here and simplify the logic by changing this to be something like:
```C#
private void SetRequestHandleDecompressionOptions(SafeCurlHandle requestHandle)
{
    bool gzip = (AutomaticDecompression & DecompressionMethods.GZip) != 0;
    bool deflate = (AutomaticDecompression & DecompressionMethods.Deflate) != 0;
    if (gzip || deflate)
    {
        string encoding = 
            gzip && deflate ? EncodingNameGzip + ", " + EncodingNameDeflate :
            gzip ? EncodingNameGzip :
            EncodingNameDeflate;
        Interop.libcurl.curl_easy_setopt(requestHandle, CURLoption.CURLOPT_ACCEPTENCODING, encoding);
}
```

stephentoub: Also, it looks like no verification is done when AutomaticDecompression is set.  Does any need to be done here?  The Windows implementation looks like there might be some validation done in winhttp that causes an exception to be thrown; if so, we should stive to match that same behavior here, whatever the behavior is.

stephentoub:
cc: @davidsh, @CIPop, @pgavlin

stephentoub:
A few comments, but otherwise LGTM.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
justinvp: It's pretty common to accept both gzip and deflate... You could go even further and avoid the allocations associated with the string concatenation, something like:

```c#
private const string EncodingNameGzipAndDeflate = EncodingNameGzip + ", " + EncodingNameDeflate;

...

private void SetRequestHandleDecompressionOptions(SafeCurlHandle requestHandle)
{
    bool gzip = (AutomaticDecompression & DecompressionMethods.GZip) != 0;
    bool deflate = (AutomaticDecompression & DecompressionMethods.Deflate) != 0;
    if (gzip || deflate)
    {
        string encoding = 
            gzip && deflate ? EncodingNameGzipAndDeflate :
            gzip ? EncodingNameGzip :
            EncodingNameDeflate;
        Interop.libcurl.curl_easy_setopt(requestHandle, CURLoption.CURLOPT_ACCEPTENCODING, encoding);
    }
}
```

stephentoub: There shouldn't be any allocations in my example, as all of the strings in the concat are const; the compiler will effectively do the same transformation you did.

justinvp: Actually, disregard my comment... Since `EncodingNameGzip` and `EncodingNameDeflate` are already constants, my suggestion is really no different than what @stephentoub suggests.

justinvp: Yeah, I realized that after I already commented. Comment deleted.

kapilash: Sending out a PR to remove the string builder allocation. Thank you, @stephentoub  and @justinvp 

@stephentoub :
> The Windows implementation looks like there might be some validation done in winhttp that causes an exception to be thrown

Winhttp throws error when the native library does not support the feature. Curl-Option validation is getting checked-in via other PR #2597.  Depending on which of these go first, one of us will incorporate the error checking here.

stephentoub:
LGTM

PR #2654, opened 2015-08-06
Implement System.Security.Cryptography.RSACng

AtsushiKan:
RSACng implemented with one open issue (tracked internally as 1208349) - SignData and VerifyData throw NI for any hash algorithm other than MD5/SHA1/SHA256/SHA384/SHA512. I want to peel that off for a later commit as it's likely to create a lot of debate over how/if to share the CNG hash provider code with SSC.Algorithms (these also involve the only P/Invokes that any assembly outside of SSC.Cng has any reasonable excuse to call.) 

Since the rest of the class is in good shape, I'd like to get it put away as I'll probably be randomized next week.

### DIFF
--- a/src/System.Security.Cryptography.Cng/src/Interop/BCrypt/Interop.AsymmetricEncryption.Types.cs
+++ b/src/System.Security.Cryptography.Cng/src/Interop/BCrypt/Interop.AsymmetricEncryption.Types.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    /// <summary>
+    /// BCrypt types related to asymmetric encryption algorithms
+    /// </summary>
+    internal partial class BCrypt
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct BCRYPT_OAEP_PADDING_INFO
+        {
+            /// <summary>
+            ///     Null-terminated Unicode string that identifies the hashing algorithm used to create the padding.
+            /// </summary>
+            internal IntPtr pszAlgId;
+
+            /// <summary>
+            ///     Address of a buffer that contains the data used to create the padding.
+            /// </summary>
+            internal IntPtr pbLabel;
+
+            /// <summary>
+            ///     Number of bytes in the pbLabel buffer.
+            /// </summary>
+            internal int cbLabel;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct BCRYPT_PKCS1_PADDING_INFO
+        {
+            /// <summary>
+            ///     Null-terminated Unicode string that identifies the hashing algorithm used to create the padding.
+            /// </summary>
+            internal IntPtr pszAlgId;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct BCRYPT_PSS_PADDING_INFO
+        {
+            /// <summary>
+            ///     Null-terminated Unicode string that identifies the hashing algorithm used to create the padding.
+            /// </summary>
+            internal IntPtr pszAlgId;
+
+            /// <summary>
+            ///     The size, in bytes, of the random salt to use for the padding.
+            /// </summary>
+            internal int cbSalt;
+        }
+    }
+}
--- a/src/System.Security.Cryptography.Cng/src/Resources/Strings.resx
+++ b/src/System.Security.Cryptography.Cng/src/Resources/Strings.resx
@@ -132,6 +132,9 @@
   <data name="Cryptography_ArgRSAaRequiresRSAKey" xml:space="preserve">
     <value>Keys used with the RSACng algorithm must have an algorithm group of RSA.</value>
   </data>
+  <data name="Cryptography_HashAlgorithmNameNullOrEmpty" xml:space="preserve">
+    <value>The hash algorithm name cannot be null or empty.</value>
+  </data>
   <data name="Cryptography_InvalidAlgorithmGroup" xml:space="preserve">
     <value>The algorithm group '{0}' is invalid.</value>
   </data>
@@ -182,5 +185,8 @@
   </data>
   <data name="WorkInProgress" xml:space="preserve">
     <value>WorkInProgress.</value>
+  </data>
+  <data name="WorkInProgress_UnsupportedHash" xml:space="preserve">
+    <value>Unsupported hash algorithm. RSACng currently supports only MD5, SHA1, SHA256, SHA384 and SHA512.</value>
   </data>
 </root>
--- a/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
+++ b/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
@@ -1,0 +1,106 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+
+using Microsoft.Win32.SafeHandles;
+
+using Internal.Cryptography;
+
+using ErrorCode = Interop.NCrypt.ErrorCode;
+using AsymmetricPaddingMode = Interop.NCrypt.AsymmetricPaddingMode;
+using BCRYPT_OAEP_PADDING_INFO = Interop.BCrypt.BCRYPT_OAEP_PADDING_INFO;
+
+namespace System.Security.Cryptography
+{
+    public sealed partial class RSACng : RSA
+    {
+        /// <summary>
+        ///     Encrypts data using the public key.
+        /// </summary>
+        public override byte[] Encrypt(byte[] data, RSAEncryptionPadding padding)
+        {
+            unsafe
+            {
+                return EncryptOrDecrypt(data, padding, Interop.NCrypt.NCryptEncrypt);
+            }
+        }
+
+        /// <summary>
+        ///     Decrypts data using the private key.
+        /// </summary>
+        public override byte[] Decrypt(byte[] data, RSAEncryptionPadding padding)
+        {
+            unsafe
+            {
+                return EncryptOrDecrypt(data, padding, Interop.NCrypt.NCryptDecrypt);
+            }
+        }
+
+        //
+        // Conveniently, Encrypt() and Decrypt() are identical save for the actual P/Invoke call to CNG. Thus, both
+        // api's invoke this common helper with the "transform" parameter determining whether encryption or decryption is done.
+        //
+        private byte[] EncryptOrDecrypt(byte[] data, RSAEncryptionPadding padding, EncryptOrDecryptAction encryptOrDecrypt)
+        {
+            if (data == null)
+                throw new ArgumentNullException("data");
+
+            if (padding == null)
+                throw new ArgumentNullException("padding");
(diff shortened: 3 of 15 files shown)


src/System.Security.Cryptography.Cng/src/Interop/BCrypt/Interop.AsymmetricEncryption.Types.cs
```diff
@@ -0,0 +1,36 @@
...
+        internal struct BCRYPT_PKCS1_PADDING_INFO
+        {
+            internal IntPtr pszAlgId;  // null-terminated Unicode string that identifies the hashing algorithm used to create the padding.
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct BCRYPT_PSS_PADDING_INFO
+        {
+            internal IntPtr pszAlgId;  // null-terminated Unicode string that identifies the hashing algorithm used to create the padding.
+            internal int cbSalt;       // The size, in bytes, of the random salt to use for the padding.
```
stephentoub: If we're going to add comments for every field like this, should we do it as XML comments instead?  That helps to not only document it in code, but provides the information to the dev via IntelliSense while working on the code that uses the fields.

src/System.Security.Cryptography.Cng/src/Resources/Strings.resx
```diff
@@ -183,4 +186,7 @@
   <data name="WorkInProgress" xml:space="preserve">
     <value>WorkInProgress.</value>
   </data>
+  <data name="WorkInProgress_UnsupportedHash" xml:space="preserve">
+    <value>Unsupported hash algorithm. RSACng currently supports only MD5, SHA1, SHA256, SHA384 or SHA512.</value>
```
stephentoub: Nit: "or" => "and"

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace System.Security.Cryptography
+{
+    public sealed partial class RSACng : RSA
+    {
+        /// <summary>
+        ///     Encrypts data using the public key.
+        /// </summary>
+        public override byte[] Encrypt(byte[] data, RSAEncryptionPadding padding)
+        {
+            unsafe
```
stephentoub: Why is unsafe needed here and in the next function?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
```diff
@@ -0,0 +1,106 @@
...
+        //
+        private byte[] EncryptOrDecrypt(byte[] data, RSAEncryptionPadding padding, EncryptOrDecryptAction encryptOrDecrypt)
+        {
+            if (data == null)
+                throw new ArgumentNullException("data");
+
+            if (padding == null)
+                throw new ArgumentNullException("padding");
+
+            unsafe
```
stephentoub: It looks like the only place that needs this is the call to EncryptOrDecrypt (L74).  If that's true, consider wrapping just that line in unsafe.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
```diff
@@ -0,0 +1,106 @@
...
+        private static unsafe byte[] EncryptOrDecrypt(SafeNCryptKeyHandle key, byte[] input, AsymmetricPaddingMode paddingMode, void* paddingInfo, EncryptOrDecryptAction encryptOrDecrypt)
+        {
+            int numBytesNeeded;
+            ErrorCode errorCode = encryptOrDecrypt(key, input, input.Length, paddingInfo, null, 0, out numBytesNeeded, paddingMode);
+            if (errorCode != ErrorCode.ERROR_SUCCESS)
+                throw errorCode.ToCryptographicException();
+
+            byte[] output = new byte[numBytesNeeded];
+            errorCode = encryptOrDecrypt(key, input, input.Length, paddingInfo, output, numBytesNeeded, out numBytesNeeded, paddingMode);
+            if (errorCode != ErrorCode.ERROR_SUCCESS)
```
stephentoub: Is it ever possible at this point for numBytesNeeded to be less than output.Length?  i.e. could the first call tell us a certain number of bytes were needed, but actually use fewer on the next call?  I assume not, but maybe an assert would be helpful.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
stephentoub: Nevermind; I see that the NCryptEncrypt/Decrypt functions being passed as delegates have pointers in their signatures.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
```diff
@@ -0,0 +1,106 @@
...
+    public sealed partial class RSACng : RSA
+    {
+        /// <summary>
+        ///     Encrypts data using the public key.
+        /// </summary>
+        public override byte[] Encrypt(byte[] data, RSAEncryptionPadding padding)
+        {
+            unsafe
+            {
+                return EncryptOrDecrypt(data, padding, Interop.NCrypt.NCryptEncrypt);
```
stephentoub: This is going to allocate a delegate for each call.  Consider caching it.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.HashData.cs
```diff
@@ -0,0 +1,61 @@
...
+            return hash;
+        }
+
+        private static HashAlgorithm GetHasher(HashAlgorithmName hashAlgorithm)
+        {
+            // @todo B#1208349:  This is a temporary implementation that should nevertheless handle most real-world cases. To fully implement this method,
+            //   it needs to be able to handle arbitrary hash algorithms on the local CNG primitive provider's menu. There are some interop-cleanup/layering decisions
+            //   that need to made first.
+
+            if (hashAlgorithm == HashAlgorithmName.MD5)
```
stephentoub: Did you consider a switch here instead of a bunch of ifs?

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
AtsushiKan: The method it’s calling is unsafe. It’s just well-hidden because it’s inside a delegate parameter which has a void*.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.HashData.cs
AtsushiKan: HashAlgorithmName isn’t a type that’s eligible for a switch. This is throwaway code in any case.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
AtsushiKan: There are two call-sites to EncryptOrDecrypt (L60 and L74), and more if we add support for new padding modes.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.HashData.cs
stephentoub: Ah, sure, I thought a string was exposed.

stephentoub:
cc: @bartonjs

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
AtsushiKan: All the better to slow down dictionary attacks? ☺ I don’t know, I think given that asymmetric crypto is highly expensive to begin with, not having the delegate stay around in a static field forever is better.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
```diff
@@ -0,0 +1,212 @@
...
+            //     byte[cbPrime1]       exponent1           - DP
+            //     byte[cbPrime2]       exponent2           - DQ
+            //     byte[cbPrime1]       coefficient         - InverseQ
+            //     byte[cbModulus]      privateExponent     - D
+            //
+            byte[] tempMagic = new byte[4];
+            tempMagic[0] = rsaBlob[0]; tempMagic[1] = rsaBlob[1]; tempMagic[2] = rsaBlob[2]; tempMagic[3] = rsaBlob[3];
+            KeyBlobMagicNumber magic = (KeyBlobMagicNumber)BitConverter.ToInt32(tempMagic, 0);
+            //Check the magic value in key blob header. If blob does not have required magic 
+            // then it trhows Cryptographic exception
```
bartonjs: Typo: trhows => throws

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        /// <summary>
+        ///     This function checks the magic value in the key blob header
+        /// </summary>
+        /// <param name="includePrivateParameters">Private blob if true else public key blob</param>
+        private static void CheckMagicValueOfKey(KeyBlobMagicNumber magic, bool includePrivateParameters)
+        {
+            if (includePrivateParameters)
+            {
+                if (magic != KeyBlobMagicNumber.BCRYPT_RSAPRIVATE_MAGIC && magic != KeyBlobMagicNumber.BCRYPT_RSAFULLPRIVATE_MAGIC)
```
bartonjs: It seems a little weird that we'd accept BCRYPT_RSAPRIVATE_MAGIC; because we've assumed the structure should be the size of BCRYPT_RSAFULLPRIVATE_BLOB (though we called it BCRYPT_RSAKEY_BLOB).

Since I know that's what full framework is already doing I'm not sure what the implications are; just that the code smells a little fishy.

src/System.Security.Cryptography.Cng/tests/RsaCngTests.cs
```diff
@@ -0,0 +1,283 @@
...
+
+using Test.Cryptography;
+using Xunit;
+
+namespace System.Security.Cryptography.Cng.Tests
+{
+    public static class RsaCngTests
+    {
+        [Fact]
+        public static void EncryptDecryptRoundtrip()
```
bartonjs: The RSA algorithm tests should just source-include the versions in Common\test\Cryptography\AlgorithmImplementations\RSA; and you'll need to define a tiny bit of boilerplate to let the tests know how to build an RSACng.  https://github.com/dotnet/corefx/commit/dc473a18dcc13b845c08ceda2c98b9ac4854e064 is where I did it for OpenSSL.

bartonjs: What I'd like to see in this class is any functionality that is unique to RSACng; like the CngKey ctor.

bartonjs:
All in all, LGTM.  It'd be nice to get the unified test suite incorporated; but if it's giving you strife it's fine as a followup change.

PR #2655, opened 2015-08-06
Fix FileSystemWatcher test race and allow process exit time == start time

nguerrera:
Addressing the issue that was hit by PR #2606 trying to get through CI.

cc @stephentoub @sokket 

On Linux, we can sometimes generate delete and create
instead of rename, depending on the timing of events.
As such, we can't perform renames and test that no
create or delete is raised when we're running on Linux.

The issue had so far only arisen on the negative create
test and not the negative delete test, which turned out
to be because the negative delete test was using the wrong
event name.

### DIFF
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Xunit;
 
@@ -101,7 +102,14 @@ public partial class FileSystemWatcher_4000_Tests
                 testFile.Flush();
 
                 // renaming a directory
-                testDir.Move(testDir.Path + "_rename");
+                //
+                // We don't do this on Linux because depending on the timing of MOVED_FROM and MOVED_TO events,
+                // a rename can trigger delete + create as a deliberate handling of an edge case, and this
+                // test is checking that no create events are raised.
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    testDir.Move(testDir.Path + "_rename");
+                }
 
                 // deleting a file & directory by leaving the using block
             }
--- a/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests/ProcessTests.cs
@@ -137,7 +137,7 @@ namespace System.Diagnostics.ProcessTests
             Assert.Throws<InvalidOperationException>(() => p.ExitTime);
             p.Kill();
             Assert.True(p.WaitForExit(WaitInMS));
-            Assert.True(p.ExitTime.ToUniversalTime() > timeBeforeProcessStart, "TestExitTime is incorrect.");
+            Assert.True(p.ExitTime.ToUniversalTime() >= timeBeforeProcessStart, "TestExitTime is incorrect.");
         }
 
         [Fact]
--- a/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
+++ b/src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Deleted.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Xunit;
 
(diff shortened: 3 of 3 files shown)


stephentoub:
LGTM.

I'm still slightly skeptical of the decision we made in https://github.com/dotnet/corefx/pull/2587 around this.  The whole problem stems from an imperfect design we're building on top of: inotify handles renames via a pair of MOVED_FROM and MOVED_TO events that are potentially delivered separately and that for some moves may not have a MOVED_TO at all (such as if moving a file out of the watched path).  This means that if we get a MOVED_FROM and we haven't yet received any further events, we have a decision to make:
- Do we wait until we get another event to make a decision about how to process the MOVED_FROM, seeing if the next event is a MOVED_TO (in which case the MOVED_FROM is treated as part of the rename) or if it's something else (in which case the MOVED_FROM is treated as a deletion).
- Do we assume we won't get an associated MOVED_TO, and immediately treat the MOVED_FROM as a deletion.

I'd originally implemented the former, as it doesn't suffer from the race condition the latter suffers from, which might treat some MOVED_FROM/MOVED_TO pairs as a rename and others as a creation/deletion, based on timing.  But it suffers from another problem, that if you just get a MOVED_FROM and no other event arrives (or arrives for a while) the event associated with the MOVED_FROM could be delayed (indefinitely if another event never arrives).  https://github.com/dotnet/corefx/pull/2587 switched from the former to the latter, under the assumption that it's the least bad of the two options.  That's probably true, as most listeners are likely to handle a rename similarly to a delete+create, but I haven't yet let my mind rest about it. :smile:

sokket:
@stephentoub Agreed, it's very much an imperfect design...the OS FileSystemWatcher APIs are very imperfect on every platform, unfortunately. There are probably some ways we could make our classes more reliable later down the line (for instance, by making a FileSystem specific watcher instead of an OS specific watcher) once we have everything x-play and stable, but hopefully this will do until then.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -101,7 +102,14 @@ public static void FileSystemWatcher_Created_Negative()
...
+                //
+                // We don't do this on Linux because depending on the timing of MOVED_FROM and MOVED_TO events,
+                // a rename can trigger delete + create as a deliberate handling of an edge case, and this
+                // test is checking that no delete events are raised.
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    testDir.Move(testDir.Path + "_rename");
+                }
 
                 // deleting a file & directory by leaving the using block
```
nguerrera: Hmm, this will trigger a delete. I got overconfident and made the change to check delete instead of changed without re-testing. :(

akoeplinger:
@stephentoub not sure if you've read it, but a section in [this article on LWN](https://lwn.net/Articles/605128/) ("Challenge 3: handling rename events") talks about the rename problem. They claim that adding a small timeout for waiting on the second MOVED_TO results in a very good rename detection rate:

> It employs a 2-millisecond timeout for the second read(), which was found to be sufficient to catch around 99.8% of the true renames even when the monitored directory tree was subject to a high level of rename activity.

nguerrera:
I don't see anything wrong in the log for the failure (disregard earlier deleted message, I was looking the code wrong). I'm going to rebase to fix a copy/paste error (I wrote deleted in comment that should have been created) and use that to try again...

nguerrera:
Earlier, I misread the log where xunit had crashed and my grep for "Failed.[0-9]" failed me. It was issue #2653.

stephentoub:
@akoeplinger, I hadn't read that; thanks for sharing.  Just skimmed it.  We hit up against most of the same issues described there, and ended up with most of the same solutions.  I'd considered a timeout, but didn't do the due diligence to analyze various real-world cases to determine whether it'd really help and what the optimal value would be.  If we think a 2ms timeout would help, that'd be an easy addition to the existing implementation and should have relatively little downside.

nguerrera:
CC @pgavlin 

failed with system.net.primitives looking for ntdll on linux. Not my CI day.

nguerrera:
cc @mmitche some sort of infrastructure race? My PR was submitted before SNP changes were merged...

nguerrera:
Making note of previous run before I retry: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_prtest/2954/

nguerrera:
@dotnet-bot test this please

mmitche:
@nguerrera I'm not sure.  The root job as well as the corefx Linux testing built the same hash.  The question is where the merge is done.   Whether there is a merge on the CI side from master, or on Github's side.  If I look at the logs, they are all testing the same hash, but the final checkout hash appears to be different for the Linux debug tst leg.  I'll probably need to dig through the source of the pull request builder to find out what it's attempting to do.

nguerrera:
Something else now #2660 

sad

nguerrera:
#2653 again. Sigh. I'm looking at it.

nguerrera:
@dotnet-bot test this please

mmitche:
@nguerrera That was painful.  I'll try to look into that odd merge issue today.

PR #2656, opened 2015-08-06
ZipFile: Reduce allocations in IsDirEmpty

justinvp:
- Use `Directory.EnumerateFileSystemEntries` to avoid the extra allocations associated with the `FileInfo`/`DirectoryInfo` instance created by `DirectoryInfo.EnumerateFileSystemInfos`.

- Use the default overload of `EnumerateFileSystemEntries`, which uses `SearchOption.TopDirectoryOnly`, as `SearchOption.AllDirectories` is not necessary here and `AllDirectories` currently has more allocations than `TopDirectoryOnly` on Windows.

- Use the enumerator directly to avoid the unnecessary call to `Current`.

(Use of uppercase `String` is deliberate to match the other formatting in this file.)

### DIFF
--- a/src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
+++ b/src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
 
@@ -569,10 +570,8 @@ namespace System.IO.Compression
 
         private static Boolean IsDirEmpty(DirectoryInfo possiblyEmptyDir)
         {
-            foreach (FileSystemInfo fi in possiblyEmptyDir.EnumerateFileSystemInfos("*", SearchOption.AllDirectories))
-                return false;
-
-            return true;
+            using (IEnumerator<String> enumerator = Directory.EnumerateFileSystemEntries(possiblyEmptyDir.FullName).GetEnumerator())
+                return !enumerator.MoveNext();
         }
     }  // class ZipFile
 }  // namespace


stephentoub:
LGTM.

src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
```diff
@@ -569,8 +570,11 @@ private static string EntryFromPath(string entry, int offset, int length)
...
         {
-            foreach (FileSystemInfo fi in possiblyEmptyDir.EnumerateFileSystemInfos("*", SearchOption.AllDirectories))
-                return false;
+            using (IEnumerator<String> enumerator = Directory.EnumerateFileSystemEntries(possiblyEmptyDir.FullName).GetEnumerator())
+            {
+                if (enumerator.MoveNext())
+                    return false;
+            }
 
             return true;
```
stephentoub: It occurs to me that now that this is using MoveNext instead of foreach, it could be simplified slightly:
```C#
private static bool IsDirEmpty(DirectoryInfo possiblyEmptyDir)
{
    using (IEnumerator<string> enumerator = Directory.EnumerateFileSystemEntries(possiblyEmptyDir.FullName).GetEnumerator())
        return !enumerator.MoveNext();
}
```

justinvp: I had it like the way you suggest before I submitted the PR, but I checked http://sourceof.net quickly to see the style used for [`Enumerable.Any`](https://github.com/Microsoft/referencesource/blob/74706335e3b8c806f44fa0683dc1e18d3ed747c2/System.Core/System/Linq/Enumerable.cs#L1168), so I changed it to match that. However, I just checked the latest corefx source, and `Enumerable.Any` was [simplified](https://github.com/dotnet/corefx/blob/6b5a8aed6b8cc87f074024d7947236ca2a769c2d/src/System.Linq/src/System/Linq/Enumerable.cs#L1463) as you suggest, so I'll go ahead and simplify it here as well.

ellismg:
LGTM.  Thanks, Justin.

PR #2661, opened 2015-08-06
Make it easier to use run-test.sh outside CI.

ellismg:
Update run-test.sh so that it deletes mscorlib.ni.dll (instead of relying on it to not be copied to the test machine, as we do in CI) and write a simple document on how to use it outside of CI/

/cc @mmitche @nguerrera @JeremyKuhne

### DIFF
--- a/Documentation/building/cross-platform-testing.md
+++ b/Documentation/building/cross-platform-testing.md
@@ -1,0 +1,58 @@
+# Running XUnit tests cross platform
+
+Unlike Windows, where we run tests as part of the build, we have a seperate
+explicit testing step on Linux and OSX.  Over time, this special step will go
+away in favor of a similar "run tests during the build" model.
+
+`run-test.sh` is the shell script used by Jenkins in order to run all the XUnit
+tests cross platform. It combines the cross platform CoreCLR and CoreFX builds
+together into a test layout and then runs each test project from CoreFX.
+
+In order to run tests, you need to build a bunch of different projects.  The
+instructions assume you are building for Linux, but are easily modifiable for OSX.
+
+1. Release or debug CoreCLR.  In Jenkins we use a release CoreCLR build instead
+   of debug CoreCLR since it is much faster at actually running XUnit, but debug
+   will work if you have the time.
+
+   From the root of your CoreCLR enlistment on Linux, run `./build.sh Release` in
+   order to build.
+2. A coresponding version of mscorlib.dll, built on Windows but targeting
+   Linux.  This can be produced by running `build.cmd linuxmscorlib Release` from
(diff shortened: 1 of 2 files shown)


nguerrera:
LGTM. Thanks!

Consider adding a note that using release CoreCLR is **much** faster.

Documentation/building/cross-platform-testing.md
```diff
@@ -0,0 +1,58 @@
...
+   From the root of your CoreCLR enlistment on Linux, run `./build.sh Release` in
+   order to build.
+2. A coresponding version of mscorlib.dll, built on Windows but targeting
+   Linux.  This can be produced by running `build.cmd linuxmscorlib Release` from
+   a CoreCLR enlistment on Windows.  Remember that the runtime and mscorlib are
+   tightly coupled with respect to object sizes and layout so you need to ensure
+   you have either a release coreclr and release mscorlib or debug coreclr and
+   debug mscorlib.
+3. A Linux build of CoreFX.  On Windows, run `build.cmd /p:OSGroup=Linux
+   /p:SkipTests=true`.  It is okay to build a Debug version of CoreFX and run it
```
JeremyKuhne: SkipTests is no longer necessary

ellismg: Thanks.  Removed.

JeremyKuhne:
LGTM

Documentation/building/cross-platform-testing.md
```diff
@@ -0,0 +1,58 @@
...
+instructions assume you are building for Linux, but are easily modifiable for OSX.
+
+1. Release or debug CoreCLR.  In Jenkins we use a release CoreCLR build instead
+   of debug CoreCLR since it is much faster at actually running XUnit, but debug
+   will work if you have the time.
+
+   From the root of your CoreCLR enlistment on Linux, run `./build.sh Release` in
+   order to build.
+2. A coresponding version of mscorlib.dll, built on Windows but targeting
+   Linux.  This can be produced by running `build.cmd linuxmscorlib Release` from
```
stephentoub: Or ```build.cmd linuxmscorlib Release vs2015``` if using VS2015 (or maybe that extra arg is no longer necessary?)

stephentoub:
LGTM. Thanks, Matt.

Documentation/building/cross-platform-testing.md
ellismg: I think it is still required :-(.  I think we should just look at updating build.cmd in coreclr to behave more like the version in CoreFX, which does some probing to figure out what versions of VS you have an uses the best one.

I will merge this as is, since I don't want to to balloon into a "here's all the different ways to build everything" document, but if you feel strongly that we should list both and want to do a PR, feel free.

PR #2318, opened 2015-07-11
Many optimizations for System.Linq.Enumerable.

James-Ko:
Just sat down for 5 hours looking at every method in `Enumerable.cs` to make every optimization I possibly could.

Here's a summary of the changes:

 - `WhereEnumerableIterator<TSource>` and `WhereSelectEnumerableIterator<TSource>` no longer check if `IEnumerator<T>` implements `IDisposable` (it does)
 - `SkipWhileIterator`: remove redundant checks on `yielding`
 - `RangeIterator` has been optimized to avoid double adding
 - Avoid null checks on `int?` and `long?` and etc. before calling `GetValueOrDefault()` (you **can** call a method on `null` if it's a nullable)
 - Optimize every single `Min` and `Max` call:
  - No more first-time loop checking.
  - Nullables like `int?` and `long?` avoid many null-checks after "scrolling forward" to the first non-null element in the collection.
  - `float?` and `double?` avoid another null-check with a trick using `GetValueOrDefault()`

PR #2662, opened 2015-08-06
Remove casts to IDisposable where unneeded

James-Ko:
Subissue of PR #2318, doesn't necessarily increase perf but it should for readability.

### DIFF
--- a/src/System.Console/src/System/IO/SyncTextReader.cs
+++ b/src/System.Console/src/System/IO/SyncTextReader.cs
@@ -16,7 +16,7 @@ namespace System.IO
         public static SyncTextReader GetSynchronizedTextReader(TextReader reader)
         {
             Debug.Assert(reader != null);
-            return reader as SyncTextReader ?? 
+            return reader as SyncTextReader ??
                 new SyncTextReader(reader);
         }
 
@@ -31,8 +31,7 @@ namespace System.IO
             {
                 lock (_methodLock)
                 {
-                    // Explicitly pick up a potentially methodimpl'ed Dispose
-                    ((IDisposable)_in).Dispose();
+                    _in.Dispose();
                 }
             }
         }
@@ -87,7 +86,7 @@ namespace System.IO
 
         //
         // On SyncTextReader all APIs should run synchronously, even the async ones.
-        // No explicit locking is needed, as they all just delegate 
+        // No explicit locking is needed, as they all just delegate
         //
 
         public override Task<String> ReadLineAsync()
--- a/src/System.Console/src/System/IO/SyncTextWriter.cs
+++ b/src/System.Console/src/System/IO/SyncTextWriter.cs
@@ -15,7 +15,7 @@ namespace System.IO
         internal static SyncTextWriter GetSynchronizedTextWriter(TextWriter writer)
         {
             Debug.Assert(writer != null);
-            return writer as SyncTextWriter ?? 
+            return writer as SyncTextWriter ??
                 new SyncTextWriter(writer);
         }
 
@@ -47,8 +47,7 @@ namespace System.IO
             {
                 lock (_methodLock)
                 {
-                    // Explicitly pick up a potentially methodimpl'ed Dispose
-                    ((IDisposable)_out).Dispose();
+                    _out.Dispose();
                 }
             }
         }
--- a/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
+++ b/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
@@ -1359,7 +1359,7 @@ namespace System.Collections.Concurrent
         {
             if (key == null) throw new ArgumentNullException("key");
 
-            return (key is TKey) && ((ConcurrentDictionary<TKey, TValue>)this).ContainsKey((TKey)key);
+            return (key is TKey) && this.ContainsKey((TKey)key);
         }
 
         /// <summary>Provides an <see cref="T:System.Collections.Generics.IDictionaryEnumerator"/> for the
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -184,8 +184,11 @@ namespace System.Linq
 
             public override void Dispose()
             {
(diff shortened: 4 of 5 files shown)


James-Ko:
Sorry, my editor has a tendency to trim whitespace sometimes. For the second commit, go to line 1362.

khellang:
Thank goodness for GitHub's well-hidden `?w=1` switch :wink:

sharwell:
@James-Ko It would be a good idea to either disable that feature in your editor, or *really* get used to only staging the lines in Git which you actually intended to change. With this pull request, Visual Studio will now say (in CodeLens) the last reason `ConcurrentDictionary<TKey, TValue>.CopyToPairs` changed was "Remove unneeded cast to ConcurrentDictionary".

stephentoub:
I agree with @sharwell.  Please only submit changes relevant to the issue at hand.  The real meat of this PR looks reasonable and like something we should take, but I'll wait to review until the whitespace changes are undone.  Thanks, @James-Ko.

James-Ko:
Alright, the whitespace problems should be fixed now. Git undid some of the indentation changes as well, though, so I had to address that too.

src/System.Console/src/System/IO/SyncTextReader.cs
```diff
@@ -31,8 +31,8 @@ protected override void Dispose(bool disposing)
             {
                 lock (_methodLock)
                 {
-                    // Explicitly pick up a potentially methodimpl'ed Dispose
-                    ((IDisposable)_in).Dispose();
+                    // This will always pick up TextReader.Dispose
```
stephentoub: This comment is unnecessary.  It can be deleted entirely rather than modified.

src/System.Console/src/System/IO/SyncTextWriter.cs
```diff
@@ -47,8 +47,8 @@ protected override void Dispose(bool disposing)
             {
                 lock (_methodLock)
                 {
-                    // Explicitly pick up a potentially methodimpl'ed Dispose
-                    ((IDisposable)_out).Dispose();
+                    // This will always pick up TextWriter.Dispose
```
stephentoub: Ditto.

stephentoub:
Thanks for undoing the whitespace changes.  I left a couple of additional comments to be addressed, but otherwise LGTM.  Please remove those comments and then squash the commits.  Thanks, @James-Ko.

jasonwilliams200OK:
In addition to what @sharwell suggested, you may want to install [EditorConfig plugin/extension](http://editorconfig.org/) for your editor (available for almost all major editors). 
This repository has `.editorconfig` file at root. With that in place and EditorConfig plugin installed, it automagically overrides your editor settings for that particular instance of your editor, in which the project is loaded and prevents misformatting.

src/System.Console/src/System/IO/SyncTextReader.cs
nguerrera: I wonder why the author of the original code was worried about the case where a subclass had re-implemented IDisposable. I much prefer what you have, but this is technically an externally observable in that pathological case and there's a cryptic comment showing intention for the old behavior. cc @davkean 

I wouldn't bother commenting the plain Dispose at all.

James-Ko:
Got it. Also thanks @jasonwilliams200OK for the suggestion.

src/System.Console/src/System/IO/SyncTextReader.cs
stephentoub: @nguerrera, @davkean, @weshaggard, so is this not a change we should take?  I'm not sure why we'd be using this pattern to pick up the custom IDisposable implementation here but not elsewhere.

davkean: I don't know the history behind this particular code, but think about the time it was written:

TextReader only had Close and explicitly implemented Dispose. Close was virtual and there was a virtual Dispose(bool). Perhaps they were attempting to skip some overriden Close behavior?

James-Ko: @nguerrera I don't see any effect this would have, actually. Neither `TextWriter` or `TextReader` explicitly impl IDisposable, and Dispose is resolved during compilation to pick up on those implementations of IDisposable. The original authors of the code may have made a mistake.

stephentoub: @James-Ko, it's a situation like this:
```C#
using System;
using System.IO;
using System.Text;

class Program
{
    static void Main()
    {
        TextWriter writer = new MyTextWriter();
        ((IDisposable)writer).Dispose();
        writer.Dispose();
    }
}

class MyTextWriter : TextWriter, IDisposable
{
    void IDisposable.Dispose()
    {
        Console.WriteLine("IDisposable.Dispose()");
    }

    public override Encoding Encoding { get { return Encoding.UTF8; } }
}
```
Run that, and you'll see the output message only output for the call via ((IDisposable)writer).Dispose, so the point @nguerrera was raising is that if such an instance was passed in, changing how we invoke Dispose could stop invoking that custom Dispose implementation.

davkean: @James-Ko At the time the code was written, TextWriter and TextReader explicitly implemented IDisposable. Theoretically, also I can have my own TextReader/TextWriter implement IDisposable again, and completely skip the base's implementation. Such a class would experience a behavior change in above. However, the likelihood of such a thing is probably pretty low and such an implementation would have different broken behavior with other things that called Dispose directly.

stephentoub: Thanks, David.  Agreed.

sharwell: > I'm not sure why we'd be using this pattern to pick up the custom `IDisposable` implementation here but not elsewhere.

Technically this is the behavior of a `using` statement so it gets used frequently.

stephentoub:
@James-Ko, there are still a few places where code is being modified with whitespace-only changes.  As it's just a few, I'll go ahead and merge this, but in future PRs please avoid that.

Thanks for the contribution!

src/System.Console/src/System/IO/SyncTextReader.cs
stephentoub: @sharwell, good point.

weshaggard: I'm also fine with this change.

PR #2663, opened 2015-08-06
Remove null checks before GetValueOrDefault

James-Ko:
Separated from PR #2318. Free perf.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -136,7 +136,7 @@ namespace System.Linq
                 //
                 // Normally, this happens when you chain two consecutive Select<,>. There may be
                 // some clever way to handle that second generic parameter within the limitations of the
-                // static type system but it's a lot simpler just to break the chain by inserting 
+                // static type system but it's a lot simpler just to break the chain by inserting
                 // a dummy Where(x => y) in the middle.
                 //
                 return new WhereEnumerableIterator<TSource>(this, x => true).SelectImpl<TResult>(selector);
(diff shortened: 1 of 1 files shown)


stephentoub:
> Free perf

Can you please share your perf tests and results?  What's the impact when all values are null?  When some are null?  When none are null?  Etc.

stephentoub:
cc: @VSadov

mikedn:
> Can you please share your perf tests and results? What's the impact when all values are null? When some are null? When none are null? Etc.

This approach should be faster if the number of nulls in the source is low. `GetValueOrDefault` "cheats" and returns the `value` field without actually checking `hasValue`, that avoids a bunch of things: a memory load, a 0 test and a branch.

If the number of nulls is high then this will likely be slower, especially in the decimal case. `add` operations that previously were simply skipped now how to be done.

Let's see what benchmarks will say :)

James-Ko:
Just finished writing the test. [Here](https://gist.github.com/James-Ko/3741a5024ffbc6e8b3ce) is the test ([ideone[(http://ideone.com/nicJBg)), [here](https://gist.github.com/James-Ko/9ff5f1b166f77e839d75) are the old results, [here](https://gist.github.com/James-Ko/99ec130bb94073ca0d9f) are the new ones.

stephentoub:
Thanks for sharing.  Based purely on the data you shared, I don't think this is a change we should take.  Where there was an improvement, it was typically significantly less than 1%, and there were plenty of regressions, including several greater than 1%, and one almost 5%.

James-Ko:
Alright, I understand. Thanks.

PR #2665, opened 2015-08-06
Small optimization for Range

James-Ko:
Separated from PR #2318. I thought this would be optimized out by the JIT in the first place, but my [perf](https://gist.github.com/James-Ko/d4f60cdc188666927eae) tests ([ideone](http://ideone.com/n2F97s)) show that the new impl is consistently faster than the original, albeit by very little.


stephentoub:
cc: @VSadov 

LGTM

PR #2668, opened 2015-08-07
Optimize ToArray for non-empty enumerables

James-Ko:
Separated from PR #2318. This does *not* negatively impact empty collections, and behavior for `Buffer<T>.ToArray()` shouldn't change.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -3437,18 +3437,20 @@ namespace System.Linq
 
             if (items == null)
             {
-                foreach (TElement item in source)
+                using (IEnumerator<TElement> e = source.GetEnumerator())
                 {
-                    if (items == null)
+                    if (!e.MoveNext()) return;
+                    
+                    items = new TElement[4];
+                    do
                     {
-                        items = new TElement[4];
-                    }
-                    else if (items.Length == count)
-                    {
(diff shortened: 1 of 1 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3110,9 +3110,9 @@ internal struct Buffer<TElement>
 
         internal Buffer(IEnumerable<TElement> source, bool queryInterfaces = true)
         {
-            TElement[] items = null;
-            int count = 0;
-
+            items = null;
+            count = 0;
```
stephentoub: With this change, all of the accesses in this method are now field accesses rather than local accesses.  You've verified that doesn't harm the generated assembly?

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: +1 here.
Keeping items and count as locals until the copying is done can be beneficial, so I'd preserve that. The change removes one null check (that would seem to be very predictable), so perf benefit of that might be very small and I wonder if introducing field accesses may actually make things slower.

* can we keep items/count as locals until done copying?
* what speedup are we looking at here overall?

James-Ko:
Closed because of dupe @ #2871.

PR #2669, opened 2015-08-07
Fewer loop checks in SkipWhile

jamesqo:
Separated from PR #2318. Sorry if I seem to be making a lot of PRs today haha, I just want to finish up with all of the changes I've made.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -709,11 +709,14 @@ namespace System.Linq
 
         private static IEnumerable<TSource> SkipWhileIterator<TSource>(IEnumerable<TSource> source, Func<TSource, bool> predicate)
         {
-            bool yielding = false;
-            foreach (TSource element in source)
+            using (IEnumerator<TSource> e = source.GetEnumerator())
             {
-                if (!yielding && !predicate(element)) yielding = true;
-                if (yielding) yield return element;
+                do
+                    if (!e.MoveNext()) yield break;
+                while (predicate(e.Current));
+                do
+                    yield return e.Current;
+                while (e.MoveNext());
             }
         }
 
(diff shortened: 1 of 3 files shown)


James-Ko:
Err... looks like something with `ProcessTests.ProcessThreadTests.TestStartTimeProperty` is failing? This is only happening on [Linux debug](http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/1583/).

stephentoub:
> looks like something with ProcessThreadTests.TestStartTimeProperty is failing?

Yeah, https://github.com/dotnet/corefx/issues/2613.  @Priya91, FYI.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -703,11 +703,14 @@ public static IEnumerable<TSource> SkipWhile<TSource>(this IEnumerable<TSource>
...
+            using (IEnumerator<TSource> e = source.GetEnumerator())
             {
-                if (!yielding && !predicate(element)) yielding = true;
-                if (yielding) yield return element;
+                do
+                    if (!e.MoveNext()) yield break;
+                while (predicate(e.Current));
+                do
+                    yield return e.Current;
+                while (e.MoveNext());
```
nguerrera: Hmm, the old code was easier to read and maintain IMHO. What kind of speedup do you see here? I'd imagine it wouldn't be very significant vs. general overhead of iterators, but I'm prepared to be proven wrong.

stephentoub:
cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
Maxwe11: +1

VSadov: The new code removes the need for a lifted "yielding" flag an multiple reads of that field. Instead the state machine gets one more state.
This should, in theory, lead to some improvement, but it is indeed curious to know if it indeed made things faster and by how much?

stephentoub:
@dotnet-bot test this please

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: @jamesqo, do you have measurements to go along with this change?

jamesqo: @stephentoub Yep, [here they are](https://gist.github.com/jamesqo/87e3ab99180f4f01cc30) ([results](https://gist.github.com/jamesqo/05aed6873cbd6812b406)). Unfortunately, it looks like nguerrera/Maxwe11 are correct; the new code is barely faster than the old code.

JonHanna: I've found doing similar work that you really have to do these tests across a range of element types, a range of enumerable types and a range of enumerable sizes. The results can vary considerably.

stephentoub:
Thanks, @jamesqo.  Given the discussion thus far and the lack of demonstrable improvements, I'm going to close this out for now.  If you're able to show that this has a real positive impact, please do reopen. Thanks!

PR #2671, opened 2015-08-07
Add small timeout to FSW MOVED_FROM polling on Linux

stephentoub:
inotify handles renames via a pair of MOVED_FROM and MOVED_TO events that are potentially delivered separately and that for some moves may not have a MOVED_TO at all (such as if moving a file out of the watched path). This means that if we get a MOVED_FROM and we haven't yet received any further events, we have a choice:
- Do we wait until we get another event to make a decision about how to process the MOVED_FROM, seeing if the next event is a MOVED_TO (in which case the MOVED_FROM is treated as part of the rename) or if it's something else (in which case the MOVED_FROM is treated as a deletion).
- Do we assume we won't get an associated MOVED_TO, and immediately treat the MOVED_FROM as a deletion.

We originally implemented the former, as it doesn't suffer from the race condition the latter suffers from, which might treat some MOVED_FROM/MOVED_TO pairs as a rename and others as a creation/deletion, based on timing. But it suffers from another problem, that if you just get a MOVED_FROM and no other event arrives (or arrives for a while) the event associated with the MOVED_FROM could be delayed (indefinitely if another event never arrives). #2587 switched from the former to the latter, under the assumption that it's the least bad of the two options.

As discussed at https://github.com/dotnet/corefx/pull/2655#issuecomment-128484923, a compromise can be to wait a few milliseconds for the MOVED_TO to arrive, as that's been shown to handle a vast majority of the cases where a MOVED_TO is on its way.  This commit simply adds that timeout to our call to poll.  

(As long as this code was being modified, the commit also adds a try/finally around the handle manipulation, just in case an exception happens to emerge.)

### DIFF
--- a/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
+++ b/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
@@ -601,15 +601,34 @@ namespace System.IO
                                 // so we will send a DELETE for this event and a CREATE when the MOVED_TO is eventually processed.
                                 if (_bufferPos == _bufferAvailable)
                                 {
-                                    bool success = false;
-                                    Interop.libc.PollFlags resultFlags;
-                                    _inotifyHandle.DangerousAddRef(ref success);
-                                    Debug.Assert(success, "Failed to add-ref inotify handle");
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @sokket, @akoeplinger

sokket:
LGTM, thanks for pushing out the fix so quickly!

nguerrera:
LGTM

PR #2673, opened 2015-08-07
Make AES work on Windows 7

### LINKED ISSUE #2672: AES doesn't work on Windows 7
The AES tests are failing on Windows 7 with a not-very-expressive CryptographicException.

bartonjs:
Creating a key from an AES algorithm then changing the cipher mode on the key is a Windows 8 feature; Windows 7 require that the algorithm be set correctly then the key built from the correct one.

Fixes #2672.

### DIFF
--- a/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/AesCngCryptoTransform.cs
+++ b/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/AesCngCryptoTransform.cs
@@ -20,8 +20,9 @@ namespace Internal.Cryptography
                 _iv = cipherIv.CloneByteArray();
                 _currentIv = new byte[cipherIv.Length];
             }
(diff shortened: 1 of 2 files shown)


bartonjs:
cc: @morganbr

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/AesCngCryptoTransform.cs
```diff
@@ -72,6 +73,29 @@ protected override void Reset()
             }
         }
 
+        private static SafeAlgorithmHandle GetCipherAlgorithm(CipherMode cipherMode)
```
stephentoub: Consider naming this GetCachedCipherAlgorithm, or something that hints that this can only return algorithms that are rooted and kept alive, since the caller's not doing that.

stephentoub:
LGTM

morganbr:
LGTM

PR #2675, opened 2015-08-07
ZipFile: Reduce string allocations

justinvp:
When creating entry names, reuse a single `char[]` buffer across all entries to avoid intermediary `char[]` array allocations for each entry. Append the `PathSeparator` (if necessary) as part of this operation to avoid additional string concatenation allocations.

### DIFF
--- a/src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
+++ b/src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
@@ -500,6 +500,12 @@ namespace System.IO.Compression
                 if (includeBaseDirectory && di.Parent != null)
                     basePath = di.Parent.FullName;
 
+                // Windows' MaxPath (260) is used as an arbitrary default capacity, as it is likely
+                // to be greater than the length of typical entry names from the file system, even
+                // on non-Windows platforms. The capacity will be increased, if needed.
+                const int DefaultCapacity = 260;
+                char[] entryNameBuffer = new char[DefaultCapacity];
+
                 foreach (FileSystemInfo file in di.EnumerateFileSystemInfos("*", SearchOption.AllDirectories))
                 {
                     directoryIsEmpty = false;
@@ -507,11 +513,10 @@ namespace System.IO.Compression
                     Int32 entryNameLength = file.FullName.Length - basePath.Length;
                     Debug.Assert(entryNameLength > 0);
 
-                    String entryName = EntryFromPath(file.FullName, basePath.Length, entryNameLength);
-
                     if (file is FileInfo)
                     {
                         // Create entry for file:
+                        String entryName = EntryFromPath(file.FullName, basePath.Length, entryNameLength, ref entryNameBuffer);
                         ZipFileExtensions.DoCreateEntryFromFile(archive, file.FullName, entryName, compressionLevel);
                     }
                     else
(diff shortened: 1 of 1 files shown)


src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
```diff
@@ -550,21 +558,34 @@ private static string EntryFromPath(string entry, int offset, int length)
...
 
-            // '/' is a more broadly recognized directory separator on all platforms (eg: mac, linux)
-            // We don't use Path.DirectorySeparatorChar or AltDirectorySeparatorChar because this is
-            // explicitly trying to standardize to '/'
-            for(int i = 0; i < chars.Length; i++)
+            unsafe
             {
-                if (chars[i] == Path.DirectorySeparatorChar || chars[i] == Path.AltDirectorySeparatorChar)
-                    chars[i] = PathSeparator;
+                fixed (char* pResult = result)
```
stephentoub: @ellismg, what are your thoughts on this kind of string mutation being done in corefx?

nguerrera: Not speaking for @ellismg, but I would not like this to spread. Perhaps we can do something along the lines of #180 for string creation? I'd want to contain the unsafe code to one place. This is a mundane thing to want to do: create and populate a string of known size, and avoid unnecessary allocation/copying. I should not have to resort to unsafe code to do it.

Putting that aside for a minute, I believe you'd get a speedup by passing '\0' to the constructor and filling the '/''s in a loop. We've actually tweaked that constructor to take into account that the memory will already have been zeroed out: https://github.com/dotnet/coreclr/blob/f618412e9fc6ff82416943c7c3e8b0199582106d/src/mscorlib/src/System/String.cs#L1578

ellismg: +1 to Nick's comments.  I love the desire to reduce allocations and wish there was a better way to do this but this pattern makes me really nervous.  I would prefer for us to use this as an example of while we need to solve this problem in a safe manner.

If this was a real bottleneck, the maybe we could do this.  I'd rather we just explore figuring out if we could defer this normalization until we were actually writing the entry into the zip file itself.  Ideally at that point we would not need to allocate a string at all.

In general, I like the desire to write tight, low allocation code all the time to avoid a peanut butter effect across the platform just because we don't have specific metrics on a key scenario.  However, in this case, unless we knew this was really killing us from real world profiles in main line scenarios, I think the cure is worse than the disease.

stephentoub: Thanks, guys. +1

justinvp: Thanks for the feedback, guys. In lieu of having "something along the lines of #180 for string creation" readily available, I'll update this PR with the alternate approach I mentioned, which doesn't use unsafe code.

> I believe you'd get a speedup by passing '\0' to the constructor

Now that you mention it, this change to string sounds familiar. Thanks for the reminder!

stephentoub:
@dotnet-bot test this please

stephentoub:
LGTM.  Thanks for the update, Justin.  I like this better.

src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
```diff
@@ -550,23 +557,40 @@ private static string EntryFromPath(string entry, int offset, int length)
             }
 
             if (length == 0)
-                return String.Empty;
+                return appendPathSeparator ? PathSeparator.ToString() : String.Empty;
```
stephentoub: Presumably the length == 0 && appendPathSeparator case isn't very common?

justinvp: I haven't been able to even exercise the `length == 0` case. I don't think it's possible. *Maybe* on Unix when `sourceDirectoryName` is `"/"`, `includeBaseDirectory` is `true`, and `"/"` doesn't contain any sub directories or files, but a root dir without any sub directories or files seems extremely unlikely (if not impossible), and zipping such a directory seems just as improbable.

stephentoub: Makes sense.  Thanks.

Issue #1942, opened 2015-06-05
Proposal: Dictionary<TKey, TValue>.TryAdd(TKey, TValue)


stephentoub:
It's very common to see code that does this:

``` C#
Dictionary<TKey, TValue> data = ...;
...
if (!data.ContainsKey(key))
    data.Add(key, value);
```

This forces the dictionary to lookup the key twice when the key isn't yet in the dictionary.

It'd be great if we had an additional method on the type:

``` C#
public bool TryAdd(TKey key, TValue value);
```

Functionally it would behave exactly as if it were implemented like:

``` C#
public bool TryAdd(TKey key, TValue value)
{
    if (ContainsKey(key)) return false;
    Add(key, value);
    return true;
}
```

but the implementation internally would be done more efficiently, essentially implemented by calling a variant of Insert but that returns false instead of throwing when a duplicate key is found.

svick:
I wonder whether other functions from `ConcurrentDictionary` could make sense on `Dictionary` as convenience functions/performance improvements.

For example if you need something like `MultiValueDictionary`, you would normally write:

```c#
List<T> values;
if (!dict.TryGetValue(key, out values))
{
    values = new List<T>();
    dict.Add(key, values);
}

// use values here
```

Or you could simplify this and also potentially make it more efficient into:

```c#
var values = dict.GetOrAdd(key, _ => new List<T>());
```

I'm focusing on this being simpler code, because it's not clear to me whether it would actually be more efficient

Though this specific example might become moot if [`MultiValueDictionary`](https://www.nuget.org/packages/Microsoft.Experimental.Collections) ever leaves alpha.

---

Regarding `TryAdd()`: what about a delegate overload, for the cases where `value` is freshly created and expensive to create? Or does "expensive to create" imply "the overhead of the additional lookup is insignificant, so the `ContainsKey()`/`Add()` combo is sufficient"?

richlander:
@stephentoub I suspect a lot of people would appreciate this feature. It's conceptually more intuitive.

Do you have a sense of the performance benefit? How many times would you need execute this pattern for it to matter?

whoisj:
This is better. The entire test and set model most developers use with `Dictionary<K,V>` is flawed from a thread safety point of view anyways. This allows a (in theory) thread-safe to test and set new values into the dictionary; in addition to saving on typing. :+1: 

What about methods like `.AddOrUpdate` and '.TryRemove' or even `.TryUpdate'.

[...]

PR #2677, opened 2015-08-07
Replace ContainsKey with TryGetValue in Regex

James-Ko:
I did a recursive search for `".ContainsKey("` and sifted though the matches; unfortunately, the only places I could optimize were `WinRTFileSystem` and `Regex`. Here is the PR.

That being said, many of the matches I found looked like `if (!dict.ContainsKey(key) dict.Add(key, value)`. If we [add a `TryAdd` method](https://github.com/dotnet/corefx/issues/1942) like we have for ConcurrentDictionary, around ~40 different places could be optimized.

### DIFF
--- a/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -362,10 +362,8 @@ namespace System.Text.RegularExpressions
             {
                 if (_caps != null)
                 {
-                    if (!_caps.ContainsKey(i))
+                    if (!_caps.TryGetValue(i, out i))
                         return String.Empty;
-
-                    i = _caps[i];
                 }
 
                 if (i >= 0 && i < _capslist.Length)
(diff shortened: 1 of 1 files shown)


src/System.IO.FileSystem/src/System/IO/WinRTFileSystem.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: Please revert the changes to this file.  We don't build or test it with the infrastructure we have right now in the corefx repo, making it difficult to validate changes.

src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -395,10 +393,11 @@ public int GroupNumberFromName(String name)
             // look up name if we have a hashtable of names
             if (_capnames != null)
             {
-                if (!_capnames.ContainsKey(name))
+                int result;
```
stephentoub: This will not compile.  Please make sure you're building and testing locally before submitting any changes in a PR.  We appreciate your contributions, but this has been happening a fair amount, and it makes it difficult for us to continue to evaluate such changes.

stephentoub:
LGTM.  Thanks for the fix.

PR #2681, opened 2015-08-07
Adding new Unit Tests for the System.Threading.Timer implementation

### LINKED ISSUE #848: Open Source System.Threading.Timer Tests

sokket:
Extends the Test Coverage of the System.Threading.Timer implementation and moves the test coverage into CoreFX (a separate PR on CoreCLR will be done to remove the current test files there). 

@stephentoub 
/cc: @piotrpMSFT @joshfree 

Fixes #848

### DIFF
--- a/src/System.Threading.Timer/System.Threading.Timer.sln
+++ b/src/System.Threading.Timer/System.Threading.Timer.sln
@@ -1,0 +1,34 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{ac20a28f-fda8-45e8-8728-058ead16e44c}") = "System.Threading.Timer.Tests", "tests\System.Threading.Timer.Tests.csproj", "{ac20a28f-fda8-45e8-8728-058ead16e44c}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Linux_Debug|Any CPU = Linux_Debug|Any CPU
+		Linux_Release|Any CPU = Linux_Release|Any CPU
+		OSX_Debug|Any CPU = OSX_Debug|Any CPU
+		OSX_Release|Any CPU = OSX_Release|Any CPU
+		Windows_Debug|Any CPU = Windows_Debug|Any CPU
+		Windows_Release|Any CPU = Windows_Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Linux_Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Linux_Debug|Any CPU.Build.0 = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Linux_Release|Any CPU.ActiveCfg = Release|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Linux_Release|Any CPU.Build.0 = Release|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.OSX_Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.OSX_Debug|Any CPU.Build.0 = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.OSX_Release|Any CPU.ActiveCfg = Release|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.OSX_Release|Any CPU.Build.0 = Release|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Windows_Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Windows_Debug|Any CPU.Build.0 = Debug|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Windows_Release|Any CPU.ActiveCfg = Release|Any CPU
+		{ac20a28f-fda8-45e8-8728-058ead16e44c}.Windows_Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+EndGlobal
--- a/src/System.Threading.Timer/tests/TimerChangeTests.cs
+++ b/src/System.Threading.Timer/tests/TimerChangeTests.cs
@@ -1,0 +1,97 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using Xunit;
+
+public class TimerChangeTests
+{
+    private void EmptyTimerTarget(object o) { }
+
+    [Fact]
+    public void Timer_Change_NegativeTimeSpan_DueTime_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() =>t.Change(TimeSpan.FromMilliseconds(-2), new TimeSpan(1) /* not relevant */));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_NegativeTimeSpan_Period_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(new TimeSpan(1) /* not relevant */, TimeSpan.FromMilliseconds(-2)));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_NegativeInt_DueTime_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(-2, 1 /* not relevant */));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_NegativeInt_Period_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(1 /* not relevant */, -2));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_TooLongTimeSpan_DueTime_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(TimeSpan.FromMilliseconds((long)0xFFFFFFFF), new TimeSpan(1) /* not relevant */));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_TooLongTimeSpan_Period_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(new TimeSpan(1) /* not relevant */, TimeSpan.FromMilliseconds((long)0xFFFFFFFF)));
+        }
+    }
+
+    [Fact]
+    public void Timer_Change_TimeSpan_AfterDispose_Throws()
+    {
+        var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1);
+        t.Dispose();
+        Assert.Throws<ObjectDisposedException>(() => t.Change(TimeSpan.FromMilliseconds(1), TimeSpan.FromMilliseconds(1)));
+    }
+
+    [Fact]
+    public void Timer_Change_Int_AfterDispose_Throws()
+    {
+        var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1);
+        t.Dispose();
+        Assert.Throws<ObjectDisposedException>(() => t.Change(1, 1));
+    }
+
+    [Fact]
+    public void Timer_Change_BeforeDueTime_ChangesWhenTimerWillFire()
+    {
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            are.Set();
+        }), null, TimeSpan.FromSeconds(500), TimeSpan.FromMilliseconds(50)))
+        {
+            Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(100)), "The reset event should not have been set yet");
+            t.Change(TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(-1));
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(500)), "Should have received a timer event after this new duration");
+        }
+    }
+}
--- a/src/System.Threading.Timer/tests/TimerFiringTests.cs
+++ b/src/System.Threading.Timer/tests/TimerFiringTests.cs
@@ -1,0 +1,204 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+using System.Threading;
+using Xunit;
+
+public class TimerFiringTests
+{
+    [Fact]
+    public void Timer_Fires_After_DueTime_Ellapses()
+    {
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            are.Set();
+        }), null, TimeSpan.FromMilliseconds(250), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */))
+        {
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(750)));
+        }
+    }
+
+    [Fact]
+    public void Timer_Fires_AndPassesStateThroughCallback()
+    {
+        object state = new object();
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            Assert.Same(s, state);
+        }), state, TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */)) { }
+    }
+
+    [Fact]
+    public void Timer_Fires_AndPassesNullStateThroughCallback()
+    {
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            Assert.Null(s);
+        }), null, TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */)) { }
+    }
+
+    [Fact]
+    public void Timer_Fires_After_DueTime_AndOn_Period()
+    {
+        int count = 0;
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            count++;
+            if (count >= 2)
+            {
+                are.Set();
+            }
+        }), null, TimeSpan.FromMilliseconds(250), TimeSpan.FromMilliseconds(50)))
+        {
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(2000)));
+        }
+    }
+
+    [Fact]
+    public void Timer_FiresOnlyOnce_OnDueTime_With_InfinitePeriod()
+    {
+        int count = 0;
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            count++;
+            if (count >= 2)
+            {
+                are.Set();
+            }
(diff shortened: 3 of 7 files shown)


src/System.Threading.Timer/tests/TimerChangeTests.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            are.Set();
+        }), null, TimeSpan.FromMilliseconds(500), TimeSpan.FromMilliseconds(50)))
+        {
+            t.Change(TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1));
+            Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(600)), "Should not have had a timer event since we changed the DueTime");
```
stephentoub: These values will make the test take a while and also seem a bit prone to random fluctations in the CI system.  What if you instead set the timeout to something extremely large, verified that the event isn't set, then change it to something extremely small, and verify that it fires?

src/System.Threading.Timer/tests/TimerChangeTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using System;
+using System.Threading;
+using Xunit;
+
+public class TimerChangeTests
+{
+    private static void EmptyTimerTarget(Object o) { }
+
+    [Fact]
+    public static void Timer_Change_NegativeTimeSpan_DueTime_Throws()
```
stephentoub: Any reason these need to be static?  I'm not 100% sure, but I seem to remember that xunit wouldn't parallelize static [Fact] methods like this... I could be wrong.

src/System.Threading.Timer/tests/TimerChangeTests.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    [Fact]
+    public static void Timer_Change_NegativeInt_DueTime_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(new Action(() =>
+            {
+                t.Change(-2, 1 /* not relevant */);
+            }));
```
stephentoub: FWIW, how you've written these is fine, but you shouldn't need as much ceremony, e.g. you should be able to write:
```C#
Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(-2 1 /* not relevant */));
```
without explicitly constructing the Action delegate and with using an expression rather than than a statement block.

src/System.Threading.Timer/tests/TimerChangeTests.cs
sokket: Good catch, side effect from an initial version....removed all instances of static

src/System.Threading.Timer/System.Threading.Timer.sln
```diff
@@ -0,0 +1,34 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
```
stephentoub: You should upgrade ;)

src/System.Threading.Timer/tests/TimerChangeTests.cs
Maxwe11: http://xunit.github.io/docs/running-tests-in-parallel.html
`static` not mentioned

src/System.Threading.Timer/System.Threading.Timer.sln
sokket: Copied the csproj and sln from a different project...strange the VS 2015 didn't update this value when I launched this sln

src/System.Threading.Timer/tests/TimerChangeTests.cs
sokket: Fixed

src/System.Threading.Timer/tests/TimerChangeTests.cs
sokket: Fixed

src/System.Threading.Timer/tests/TimerChangeTests.cs
stephentoub: Yeah, like I said, I could be wrong :)  Not sure where I got that idea from, but I just tried it and it doesn't appear to matter.  Still, I prefer xunit unit tests as instance rather than static methods.

stephentoub:
I get a warning:
```
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1819,5): warning MSB3277: Found conflicts between different versions of the s
ame dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [\corefx\src\System.Threading.Timer\tests\System.Threading.Timer.Tests.csproj]
```
when I try to compile these.  Can anything be done about that?

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(750)));
+        }
+    }
+
+    [Fact]
+    public void Timer_Fires_AndPassesStateThroughCallback()
+    {
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            Assert.NotNull(s);
```
stephentoub: I'd suggest verifying it's the same object.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        AutoResetEvent are = new AutoResetEvent(false);
+
+        using (var t = new Timer(new TimerCallback((object s) =>
+        {
+            count++;
+            if (count >= 2)
+            {
+                are.Set();
+            }
+        }), null, TimeSpan.FromMilliseconds(500), TimeSpan.FromMilliseconds(50)))
```
stephentoub: Can you use a smaller value than half a second?  I realize it's not much, but with lots of tests, these values add up and make the tests take longer.

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Good catch

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Sure, I'll cut it to 250ms. Any smaller and I'm concerned that we could hit flakiness if the thread is context switched out or the system is slow.

sokket:
Fixed the build warning

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        });
+
+        using (t1 = new Timer(tc, t1State, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1)))
+        using (t2 = new Timer(tc, t2State, TimeSpan.FromMilliseconds(0), TimeSpan.FromMilliseconds(-1)))
+        {
+            // Wait for both events to fire
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(1500 /*1.5 seconds*/)));
+        }
+    }
+}
```
stephentoub: Some other tests to consider:
- An outer loop test that verifies the Timer can be garbage collected when no one's referencing it, and that it stops firing in that case
- An outer loop test that verifies a repeated timer still fires/executes subsequent invocations even if the first one is blocked.
- A test that verifies the Timer can dispose itself from its own callback
- A test that verifies the Timer can be disposed multiple times
- A test that verifies we can create lots (thousands) of timers and have many timers firing at the same time, e.g. create ten thousand timers that signal a countdown event and that have a one millisecond timeout
- A test that verifies a non-repeating timer that's already fired can have its timeout changed to be restarted

sokket:
@stephentoub Added the cases you mentioned

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        }), null, TimeSpan.FromMilliseconds(250), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */))
+        {
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(750)));
+        }
+    }
+
+    [Fact]
+    public void Timer_Fires_AndPassesStateThroughCallback()
+    {
+        Object state = new Object();
```
stephentoub: Nit: Object => object

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    [Fact]
+    public void Timer_CanDisposeSelfInCallback()
+    {
+        // There's nothing to validate besides that we don't throw an exception, so rely on xunit 
+        // to catch any exception that would be thrown and signal a test failure
+        Timer t = null;
+        TimerCallback tc = new TimerCallback((object o) => t.Dispose());
+        t = new Timer(tc, null, 100, -1);
```
stephentoub: You still need to wait for the callback to be invoked.  The tests could have ended by the time the callback eventually runs, and the callback is happening on a thread pool thread xunit has no knowledge of.  Plus, it's quite likely that the Timer will get GC'd and it'll never fire.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    [Fact]
+    public void Timer_CanBeDisposedMultipleTimes()
+    {
+        // There's nothing to validate besides that we don't throw an exception, so rely on xunit 
+        // to catch any exception that would be thrown and signal a test failure
+        TimerCallback tc = new TimerCallback((object o) => { });
+        var t = new Timer(tc, null, 100, -1);
+        for (int i = 0; i < 10; i++)
```
stephentoub: Wow, you're really serious about disposing this thing :)

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            Assert.True(are.WaitOne(50), "Should have received first timer event");
+            Assert.False(are.WaitOne(20), "Should not have received a second timer event");
+            t.Change(1, -1);
+            Assert.True(are.WaitOne(50), "Should have received a second timer event after changing it");
+        }
+    }
+
+    [Fact, OuterLoop]
+    public void Running_Timer_CanBeDisposedAndStopsFiring()
```
stephentoub: Nit: Disposed => Finalized

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+    {
+        AutoResetEvent are = new AutoResetEvent(false);
+        TimerCallback tc = new TimerCallback((object o) => are.Set());
+        var t = new Timer(tc, null, 100, 500);
+        Assert.True(are.WaitOne(250), "Failed to get first timer fire");
+        t = null; // Remove our refence so the timer can be GC'd 
+        GC.Collect();
```
stephentoub: Does this test pass locally for you consistently?  I'd have expected this could need:
```C#
GC.Collect();
GC.WaitForPendingFinalizers();
GC.Collect();
```
You ran the outer loop tests locally?

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        Assert.False(are.WaitOne(750), "Should not have received a timer fire after it was collected");
+    }
+
+    [Fact, OuterLoop]
+    public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
```
stephentoub: The body of this test isn't doing what I was expecting it to do based on the name.  I was expecting the timer callback to block, but that's not happening.  What's this test trying to validate?

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Fixed here and everywhere else

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Good catch, fixed

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Might as well hit it with a big hammer to make sure it doesn't break :)

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Fixed

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: I ran every test and it passed...could have a timing issue so I'll add those calls to alleviate that

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,226 @@
...
+                Assert.True(success, String.Format("Not all timers fired: only {0} of {1}", count, maxTimers));
+            }
+        }
+        finally
+        {
+            foreach (Timer t in arr)
+                if (t != null)
+                    t.Dispose();
+        }
+    }
```
stephentoub: This could be simplified using a CountdownEvent, e.g.
```C#
public void ManyTimers_EachTimerDoesFire()
{
    const int NumTimers = 10000;
    CountdownEvent ce = new CountdownEvent(NumTimers);
    Timer[] timers = Enumerable.Range(0, 10000).Select(_ => new Timer(s => ce.Signal(), null, 20, -1))).ToArray();
     try { Assert.True(ce.Wait(250)); }
     finally { foreach (Timer t in timers) t.Dispose(); }
}
```
or something like that.

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: The test is setting up a periodic timer so it should be firing every 500ms. We wait for the first timer to validate that one happens and then release our reference to the timer itself. We tell the GC to collect and then validate that the timer was collected by ensuring we don't receive another callback. The second callback will happen if we don't cleanup properly since the timer is periodic; if we cleanup properly, we should not get another are.Set call and therefore the Assert.False should pass.

src/System.Threading.Timer/tests/project.lock.json
```diff
@@ -0,0 +1,935 @@
+{
+  "locked": false,
```
stephentoub: Needs to be true

src/System.Threading.Timer/tests/TimerFiringTests.cs
stephentoub: I don't think we're talking about the same test.  Sorry if my comment wasn't clear, but I was referring to PeriodicTimerIsntBlockedByABlockedTimer.  That's not what you're describing, right?

src/System.Threading.Timer/tests/TimerFiringTests.cs
stephentoub: Interesting.  Ok, thanks.

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Ah gotcha, it wasn't clear which test...sorry about that. I misunderstood your comment...fixed the test to add a block in the callback

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Ah, I've never used a CountdownEvent before...useful!

src/System.Threading.Timer/tests/project.lock.json
sokket: Fixed

sokket:
The failed test doesn't repro locally so it could be a timing issue...increasing timeouts on that test

sokket:
@stephentoub any more feedback? :)

src/System.Threading.Timer/tests/TimerChangeTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    [Fact]
+    public void Timer_Change_NegativeInt_Period_Throws()
+    {
+        using (var t = new Timer(new TimerCallback(EmptyTimerTarget), null, 1, 1))
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => t.Change(1 /* not relevant */, -2));
+        }
+    }
+    [Fact]
```
stephentoub: Nit: missing blank line

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            else
+            {
+                areT2.Set();
+            }
+        });
+
+        using (var t1 = new Timer(tc, t1State, 50, -1))
+        using (var t2 = new Timer(tc, t2State, 100, 100))
+        {
+            Assert.True(areT1.WaitOne(500), "Blocking callback prevented a second timer from firing");
```
stephentoub: I'm not sure this is really testing much.  The first callback ill wait for 250 milliseconds, this will wait for 500, and the second timer will fire after 100 milliseconds, so isn't there an extremely high likelihood that the test would pass even if callbacks were serialized?

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    [Fact, OuterLoop]
```
stephentoub: Does this ManyTimers test take a long time?  We might be able to get away with it as inner loop.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,211 @@
...
+    public void ManyTimers_EachTimerDoesFire()
+    {
+        int maxTimers = 10000;
+        CountdownEvent ce = new CountdownEvent(maxTimers);
+        Timer[] timers = System.Linq.Enumerable.Range(0, maxTimers).Select(_ => new Timer(s => ce.Signal(), null, 20 /* enough time to wait on the are */, -1)).ToArray();
+        try
+        {
+            Assert.True(ce.Wait(250), String.Format("Not all timers fired, {0} left of {1}", ce.CurrentCount, maxTimers));
```
stephentoub: I'd be ok with timeouts like this being longer.  Basically, anything of the form Assert.True(something.Wait(...)) can have a long timeout, and anything of the form Assert.False(something.Wait(...)) needs as short a timeout as possible.

stephentoub:
Pretty close; left a few more comments.

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: The whole suite takes ~5 seconds, inner and outer

stephentoub: Ok, cool.  I'd make this one inner then.

src/System.Threading.Timer/tests/TimerChangeTests.cs
sokket: Fixed

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: They're all quick so I'll make them all inner :)

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Okay, I bumped them up

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Fixed

sokket:
Updated with fixes from feedback

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    [Fact]
+    public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
+    {
+        AutoResetEvent are1 = new AutoResetEvent(false);
+        AutoResetEvent are2 = new AutoResetEvent(false);
+        TimerCallback tc = new TimerCallback((object o) =>
+        {
+            using (var t1 = new Timer((object obj) => are1.Set(), null, 100, -1))
```
stephentoub: There's no need for this to be as large as 100... it could just be 1, basically anything bigger than 0.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            {
+                Assert.True(are1.WaitOne(250), "Should have received a second callback while blocked");
+                are2.Set();
+            }
+        });
+
+        using (var t1 = new Timer(tc, null, 50, -1))
```
stephentoub: Similarly, this can also just be 1 instead of 50.  There's no need to delay artificially.

stephentoub:
I'd recommend changing the two timeout values I mentioned, but other than that, LGTM.

sokket:
Changing the values to 1 made the test execute ~100ms slower, but I'll make the change and update the revision

stephentoub:
> Changing the values to 1 made the test execute ~100ms slower

Really?  Consistently across many runs?  I find that very surprising.

sokket:
The test ran at ~5.8 seconds pre-change and now runs around ~5.9 consistently.

stephentoub:
> The test ran at ~5.8 seconds pre-change and now runs around ~5.9 consistently.

That's very counterintuitive.  It's also not what I see when I run the tests locally.  I pulled down your branch and tried both with the original 100/50 values and the recommended 1/1 values, and as I'd expect the latter is consistently ~150ms faster on my Windows machine.

sokket:
Good to know :) glad it's just my machine then

sokket:
Merging, thanks @stephentoub!

PR #2682, opened 2015-08-07
Improve FileSystemWatcher implementation on Windows, and fix tests

### LINKED ISSUE #2659: PreSeededNestedStructure FSW test failing locally
This test is consistently failing for me when run locally (Win10 Enterprise RTM).
```
FileSystemWatcher_4000_Tests.FileSystemWatcher_Changed_PreSeededNestedStructure(includeSubdirectories: False) [FAIL]
        Should not observe a second file change event
        Expected: False
        Actual:   True
        Stack Trace:
           \corefx\src\System.IO.FileSystem.Watcher\tests\Utility\Utility.cs(109,0): at Utility.ExpectNoEvent(WaitHandle eventOccure
  d, String eventName, Int32 timeout)
           \corefx\src\System.IO.FileSystem.Watcher\tests\FileSystemWatcher.Changed.cs(197,0): at FileSystemWatcher_4000_Tests.FileS
  ystemWatcher_Changed_PreSeededNestedStructure(Boolean includeSubdirectories)
```

stephentoub:
The Windows implementation has some ineffiiciences, as well as some issues that look like they could be reliability problems.  Fixes:
- Every iteration of calling ReadDirectoryChangesW was allocating a bunch of state (an FSWAsyncResult, a delegate, a new overlapped, etc).  Fixed to allocate that state once and store it for the duration of the watching operation.
- Every iteration of calling ReadDirectoryChangesW was pinning/unpinning the buffer.  Fixed to use a PreAllocatedOverlapped so that this happens just once for the whole operation.
- Centralized all of the cleanup code of the thread pool binding, and removed the thread pool binding field from the watcher as it's no longer necessary and it risks regressing to the code where we accidentally used the wrong one.
- Removed confusing and duplicative flags around whether the implementation is currently running.
- Removed duplicate interop code and moved the rest to the Common\src\Interop folder to follow the convention for the rest of the repo
- Made more robust the case where a failure occurs after creating the directory handle and before starting monitoring.
- Removed unnecessary locking.
- Reorganized some code to make the implementation easier to read, and added comments.

@ericeil, can you review the overlapped changes?
@ericstj, @chcosta, @sokket, can you review everything?

Tests:
- Fixes #2659.  The test was failing, at least on my Windows 10 machine, because the watcher was picking up changes to one of the ancestor directories of the file being changed (that file's changes were causing modifications to the ancestor directory as well).
- Reduced the running time of the tests on my machine from ~40 seconds to ~3.5 seconds.


src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
```diff
@@ -11,184 +10,167 @@ namespace System.IO
...
-                                new SafeFileHandle(IntPtr.Zero, false)             // file with attributes to copy
-                            );
-
-            if (IsHandleInvalid)
+            var defaultSecAttrs = default(Interop.mincore.SECURITY_ATTRIBUTES);
+            _directoryHandle = Interop.mincore.CreateFile(
+                lpFileName: _directory,
+                dwDesiredAccess: Interop.mincore.FileOperations.FILE_LIST_DIRECTORY,
+                dwShareMode: FileShare.Read | FileShare.Delete | FileShare.Write,
+                securityAttrs: ref defaultSecAttrs,
```
sokket: Out of curiosity, why are you using named parameters for parameter?

stephentoub: No strong reason: the author of the original code apparently thought it was important to comment what these things were, so instead of comments I just used the parameter names (whenever possible I prefer using things the compiler will verify).

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
```diff
@@ -11,184 +10,167 @@ namespace System.IO
...
             finally
             {
-                if (!ok)
+                // At this point the operation has either been initiated and we'll let the callback
+                // handle things from here, or the operation has been stopped or failed, in which case
+                // we need to cleanup because we're no longer executing.
+                if (!continueExecuting)
                 {
-                    threadPoolBinding.FreeNativeOverlapped(overlappedPointer);
+                    // Clean up the overlapped pointer created for this iteration
```
sokket: Can we make the state var implement IDisposable so we can call state.Dispose and it'll clean up any internal vars correctly without needed to make 3 calls to it?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
```diff
@@ -196,234 +178,206 @@ private unsafe void Monitor(byte[] buffer)
...
+                {
+                    Debug.Assert(oldName == null, "Two FILE_ACTION_RENAMED_OLD_NAME in a row!  [" + oldName + "], [ " + name + "]");
+                    oldName = name;
+                }
+                else if (action == Interop.mincore.FileOperations.FILE_ACTION_RENAMED_NEW_NAME)
+                {
+                    if (oldName != null)
                     {
-                        NotifyInternalBufferOverflowEvent();
+                        NotifyRenameEventArgs(WatcherChangeTypes.Renamed, name, oldName);
```
sokket: Duped code...

Debug.Assert(oldName != null,...)
Notify(..)
oldName = null;

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Created.cs
```diff
@@ -8,7 +8,7 @@
 using System.Threading;
 using Xunit;
 
-public partial class FileSystemWatcher_4000_Tests
+public class CreatedTests
```
sokket: Woohoo, thanks for making these have useful names :)

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
```diff
@@ -150,7 +150,7 @@ public static void FileSystemWatcher_Moved_NestedDirectoryRootWithoutSubdirector
 
                 using (var dir2 = new TemporaryTestDirectory(Path.Combine(dir1.Path, "dir2")))
                 {
-                    Utility.ExpectNoEvent(createdOccured, "dir2 created", Utility.WaitForCreationTimeoutInMs);
+                    Utility.ExpectNoEvent(createdOccured, "dir2 created");
```
sokket: Why was the creation timeout removed? I thought that was needed for reliability on the Linux implementation

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
stephentoub: Good catch.  It was there in the original that I moved; will fix.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
stephentoub: It needed 30 seconds?

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
```diff
@@ -11,184 +10,167 @@ namespace System.IO
...
+                // Start ignoring all events that were initiated before this, and 
+                // allocate the buffer to be pinned and used for the duration of the operation
+                int session = Interlocked.Increment(ref _currentSession);
+                byte[] buffer = AllocateBuffer();
+
+                // Store all state, including a preallocated overlapped, into the state object that'll be
+                // passed from iteration to iteration during the lifetime of the operation.  The buffer will be pinned
+                // from now until the end of the operation.
+                state = new AsyncReadState(session, buffer, _directoryHandle, ThreadPoolBoundHandle.BindHandle(_directoryHandle));
+                unsafe { state.PreAllocatedOverlapped = new PreAllocatedOverlapped(ReadDirectoryChangesCallback, state, buffer); }
```
Maxwe11: Why not cached? e.g.

    new PreAllocatedOverlapped((errorCode, numBytes, overlappedPtr) =>
    {
            var state = (AsyncReadState)ThreadPoolBoundHandle.GetNativeOverlappedState(overlappedPtr);
    state.Watcher.ReadDirectoryChangesCallback(errorCode, numBytes, state);
    }, state, buffer);

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
stephentoub: The overlappedPointer isn't in the state object, so that would still need to be separate.  And this is the only code that disposes these other two things, so we could add a function and make this the body of that function, then add a call to it here, or we could just have the two lines here.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
stephentoub: I'd thought about it but it didn't seem worth the code complication.  It's one delegate created per run of the FSW, and usually with an FSW you create one, start it, and that's it.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
sokket: Cool, makes sense to keep as-is then

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
sokket: Hmm it might not need 30 seconds but the default 500ms is too slow; I saw the Linux implementation miss actual creation events at 500ms so we could get false negatives leaving it at 500ms

stephentoub: How often?  Any increase in these NoEvent waits makes a real impact on how long it takes to run the tests... prior to this change, the FSW tests were taking ~40 seconds.  If there's a false negative now and again but we'll do the right thing most of the time, I'd prefer to stick with the shorter timeout, and we could add an outer loop test to wait for whatever longer period of time we wanted to have utmost confidence.

sokket: It varies a ton...sometimes it took 300ms and sometimes it took 15 seconds; it was very strange. If a little amount of false negative flakiness is okay, then works for me :)

ericeil:
The Overlapped changes LGTM.

src/System.IO.FileSystem.Watcher/tests/FileSystemWatcher.Renamed.cs
akoeplinger: Flaky tests are very annoying IMHO. Unless this only happens *very* infrequently like once a month, I'd prefer keeping the timeout (we can make it conditional on `OsPlatform.Linux`).

stephentoub: How is it annoying if it's flaky in the direction of not asserting?

akoeplinger: Ah sorry, I assumed it'd fail - that's what I typically see when a test is "flaky" :smile: not worried about it then.

stephentoub: Cool :smile: And I completely agree: as someone who spends a lot of time fighting failing tests on github, non-deterministically failing tests really get on my nerves. In this case, I don't love it, but it just means the test will pass even though it maybe should have failed. In general we have a fair amount of that in these FSW tests, as there are a bunch of tests that try to verify the absence of something happening at some point in the future, and so at some point we have to say "well, if it hasn't happened yet, it probably won't" and declare success, even though it could still happen.

stephentoub:
Thanks, Eric and Jon.

stephentoub:
Just to put some numbers on this, I ran a test where I repeatedly used File.SetAttributes to twiddle the attributes on a file, with a FileSystemWatcher listening for attribute change notifications.  For the whole test (including both the FSW code and the code using File.SetAttributes, which itself allocates), the number of GCs dropped with this change by ~30%.

PR #2687, opened 2015-08-07
Update BuildTools version to uptake _AssemblyInfo.cs changes

naamunds:
Update the BuildTools version in CoreFx to pick up the changes to the
generation of _AssemblyInfo.cs which, are necessary for the localization
work.


stephentoub:
LGTM

PR #763, opened 2015-02-13
Cache evaluated console color strings

stephentoub:
Each time a call to Console.ForegroundColor or Console.BackgroundColor is made on Unix, we access the terminfo database to get the right evaluation string, evaluate it, and then write out the result.  But there are only 16 possible colors for each of foreground and background, and these strings are idempotent.  This commit just adds a small cache to avoid recomputing these strings every time.

Fixes #761 

cc: @pgavlin, @ellismg

PR #2689, opened 2015-08-08
Use rectangular arrays in System.Console

James-Ko:
Readability improvements to Console.

### DIFF
--- a/src/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/System.Console/src/System/ConsolePal.Unix.cs
@@ -181,7 +181,7 @@ namespace System
             // See if we've already cached a format string for this foreground/background
             // and specific color choice.  If we have, just output that format string again.
             int fgbgIndex = foreground ? 0 : 1;
-            string evaluatedString = s_fgbgAndColorStrings[fgbgIndex][ccValue]; // benign race
+            string evaluatedString = s_fgbgAndColorStrings[fgbgIndex, ccValue]; // benign race
             if (evaluatedString != null)
             {
                 Console.Write(evaluatedString);
@@ -200,7 +200,7 @@ namespace System
 
                     Console.Write(evaluatedString);
 
-                    s_fgbgAndColorStrings[fgbgIndex][ccValue] = evaluatedString; // benign race
+                    s_fgbgAndColorStrings[fgbgIndex, ccValue] = evaluatedString; // benign race
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
Maxwe11: revert this change

src/System.Console/src/System/ConsolePal.Windows.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
Maxwe11: Ditto.

stephentoub:
At the time this code was written (https://github.com/dotnet/corefx/pull/763), multidimensional arrays could not be instantiated on Unix, hence why the code was written to use a jagged array.  That was subsequently fixed (https://github.com/dotnet/coreclr/pull/376).  As such, I'm fine with it being changed to use a multidimensional array now.  But please do revert the unrelated whitespace changes, and please update the commit description; it's not really about readability... I'd prefer to see a description like "Replace jagged with multidim array in Console on Unix", or something like that.

James-Ko:
Looks like my editor stripped the BOM from each of the files- this may be why the first line keeps showing up as part of the commit when there's no visible change.

jasonwilliams200OK:
Regarding BOM, I think we should have a rule at the end of [`.editorconfig`](https://github.com/dotnet/corefx/blob/646cbde026d7b9c2e59374c33f404ac6f38a4643/.editorconfig) file to avoid it in future:

```powershell
# UTF-8 with BOM  
[*.{c,cs,cpp,h,def,txt,xml,json}]  
charset=utf-8-bom 
```

sharwell:
For C# files, you could use the new [SA1412 Store Files as UTF-8](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1412.md) rule in StyleCopAnalyzers (new for Beta 5 which I'm in the process of releasing). :smile:

Maxwe11:
@James-Ko squash into single commit and what about Stephen's comment?

James-Ko:
@sharwell Thanks; I don't often use Visual Studio for minor changes, though, so I created a [shell script](https://gist.github.com/James-Ko/d65e10f6a2aa3fc6ffe0) for people who don't have an IDE.

@Maxwe11 I realize, I was busy for the past few hours; here it is.

stephentoub:
@dotnet-bot test this please

stephentoub:
Thanks for updating the PR. LGTM.

PR #2691, opened 2015-08-09
Eliminate char buffer allocations in ConsolePal.Unix

James-Ko:
Instead of creating an intermediary `char[]` buffer, we now call `GetString`. This internally calls `string.CreateStringFromEncoding`, copying the characters directly into the string.

### DIFF
--- a/src/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/System.Console/src/System/ConsolePal.Unix.cs
@@ -321,32 +321,6 @@ namespace System
             }
         }
 
-        /// <summary>Creates a string from an array of ASCII bytes.</summary>
-        /// <param name="buffer">The byte buffer.</param>
-        /// <param name="offset">The starting location in the buffer from which to begin the string.</param>
-        /// <param name="length">The length of the resulting string.</param>
-        /// <returns>
-        /// A string containing characters copied from the buffer, one character per byte starting
-        /// from <paramref name="offset"/> and going for <paramref name="length"/> bytes.
-        /// </returns>
-        private static string StringFromAsciiBytes(byte[] buffer, int offset, int length)
-        {
-            // Special-case for empty strings
-            if (length == 0)
-            {
-                return string.Empty;
-            }
-
-            // new string(sbyte*, ...) doesn't exist in the targeted reference assembly,
-            // so we first copy to an array of chars, and then create a string from that.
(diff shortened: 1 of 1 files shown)


stephentoub:
Thanks, @James-Ko.  This Console code was written before the Encoding.ASCII property was available to it; now that it is, it's a fine thing to use.  However, it'd be better to just delete the whole StringFromAsciiBytes function and replace the two call sites with calls to Encoding.ASCII.GetBytes directly.

James-Ko:
I held off from doing that, actually, because neither `GetString` ([here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/ASCIIEncoding.cs#L369)) or `CreateStringFromEncoding` ([here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs#L1365)) check if `length == 0`, and we end up calling deep into that function if we eliminate the check. Should I go ahead regardless?

stephentoub:
> I held off from doing that, actually, because neither GetString (here) or CreateStringFromEncoding (here) check if length == 0

ASCIIEncoding does in its GetString here:
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/ASCIIEncoding.cs#L387

stephentoub:
(Also, CreateStringFromEncoding actually does check: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs#L1377-L1378)

James-Ko:
@stephentoub Neither of them check `length`, and in fact the check in `CreateStringFromEncoding` gets as far as calling `GetCharsCount` on the `Encoding` before stopping, but I agree that it's relatively rare and the cost for readability is small.

stephentoub:
> Neither of them check length, and in fact the check in CreateStringFromEncoding gets as far as calling GetCharsCount on the Encoding before stopping

Sorry, I'm not quite understanding what point you're trying to make.  So just to state my comments about the ASCIIEncoding implementation more clearly, ```Encoding.ASCII.GetString(byte[] bytes, int index, int count)``` is non-allocating if ```bytes.Length == 0``` or ```count == 0```; in such cases it'll return ```string.Empty``` rather than allocating a new string.

stephentoub:
Regardless, the change LGTM now.  Thanks.

PR #2694, opened 2015-08-09
Avoid char buffer allocations in System.IO.Compression

James-Ko:
These changes avoid creating an intermediate `char[]` buffer when creating a string.

### DIFF
--- a/src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
@@ -453,7 +453,7 @@ namespace System.IO.Compression
                 readEntryNameEncoding = Encoding.UTF8;
             }
 
-            return new string(readEntryNameEncoding.GetChars(entryNameBytes)); //readEntryNameEncoding.GetString(entryNameBytes);
+            return readEntryNameEncoding.GetString(entryNameBytes);
         }
 
         private Byte[] EncodeEntryName(String entryName, out bool isUTF8)


Maxwe11:
1. your fork is out of date
2. similar change (unsafe string mutation) was recently rejected (#2675)

src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
```diff
@@ -551,19 +551,24 @@ private static string EntryFromPath(string entry, int offset, int length)
...
+                for (int i = 0; i < result.Length; i++)
+                {
+                    if (pResult[i] == Path.DirectorySeparatorChar || pResult[i] == Path.AltDirectorySeparatorChar)
+                        pResult[i] = PathSeparator;
+                }
             }
-
-            return new string(chars);
+            
+            return result;
```
stephentoub: As @Maxwe11 stated, we don't want to be doing String mutation like this, and @justinvp already changed this implementation in https://github.com/dotnet/corefx/pull/2675.  Please revert the changes to this file.

src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
```diff
@@ -453,7 +453,8 @@ private String DecodeEntryName(Byte[] entryNameBytes)
                 readEntryNameEncoding = Encoding.UTF8;
             }
 
-            return new string(readEntryNameEncoding.GetChars(entryNameBytes)); //readEntryNameEncoding.GetString(entryNameBytes);
+            fixed (byte* bytePtr = entryNameBytes)
+                return readEntryNameEncoding.GetString(bytePtr, entryNameBytes.Length);
```
stephentoub: Why are you using unsafe code here?  It's good to make a change to avoid the intermediate chars[], but why not just do ```return readEntryNameEncoding.GetString(entryNameBytes);``` as was in the commented out code?

James-Ko: Because for some bizarre reason the default implementation of `GetString(byte[])` calls the exact same code I've just removed, while `GetString(byte*, int)` does not (refer [here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/Encoding.cs#L1492-L1496) and [here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/Encoding.cs#L1322-L1336)). I kept the change in my earlier PR #2691 because ASCIIEncoding overrides this, but here we don't know the type of the encoding beforehand.

src/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFile.cs
James-Ko: Don't we do this all the time in String and StringBuilder for performance reasons, though? I realize that mundane tasks like this should avoid pointer arithmetic, but say if we had this in a performance-critical piece of code, then what are the guidelines?

stephentoub: > Don't we do this all the time in String and StringBuilder for performance reasons, though? 

Yes, but that's inside String, the same type, and StringBuilder, which is effectively an extension of that type. To code outside of such types, string should be considered immutable, except in very rare circumstances. The mutability is effectively an internal implementation detail.

> but say if we had this in a performance-critical piece of code, then what are the guidelines?

The guideline is don't do it. If there's an extremely compelling case for doing it in a particular situation, complete with scenarios and perf tests and no alternatives, then it's possible an exception would be made.

src/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
stephentoub: While I can appreciate that argument, all of the important Encoding-derived types override that GetString overload.  It'd be extremely rare to end up in a situation here where an encoding was being used that didn't, and as such I'd much rather avoid unsafe code.

James-Ko:
@Maxwe11 I've updated the fork to resolve these problems.

stephentoub:
Thanks for updating it, @James-Ko. LGTM.

PR #2695, opened 2015-08-09
Fix TODO in CapiHelper around Encoding.ASCII

stephentoub:
The new System.Security.Cryptography.Csp assembly was depending on the older System.Text.Encoding 4.0.0 contract, which lacks Encoding.ASCII.  Updated it to use 4.0.10, which does contain Encoding.ASCII, and fixed an associated TODO.

cc: @AtsushiKan, @bartonjs, @ericstj


bartonjs:
LGTM. Thanks, @stephentoub.

stephentoub:
Thanks, Jeremy.

PR #2697, opened 2015-08-09
More char buffer elimination in CapiHelper

James-Ko:
The comment said the compiler wouldn't recognize `Encoding.ASCII`, but when I hover over it in VS it appears to work fine.


justinvp:
This is a dupe of #2695

James-Ko:
Don't know how I didn't see that; I guess I should close this, then.

PR #2698, opened 2015-08-09
Avoid the byte[] alloc in Win32FileStream.VerifyHandleIsSync

justinvp:
Updated:

Call the native `ReadFile` and `WriteFile` interop functions directly, passing a stack allocated array, instead of allocating a `byte[]` on the heap.

---

Previously:

> When creating a `FileStream` from a `SafeFileHandle` and `isAsync` is `false` (the default value), the handle is verified to make sure it is synchronous. As part of the verification, a new `byte[]` of length 1 is allocated and used to test the handle. This change uses a lazy static cached `byte[]` to avoid repeated allocations.

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -332,44 +332,33 @@ namespace System.IO
         [System.Security.SecuritySafeCritical]  // auto-generated
         private unsafe void VerifyHandleIsSync()
         {
+            Debug.Assert(!_isAsync);
+
             // Do NOT use this method on pipes.  Reading or writing to a pipe may
             // cause an app to block incorrectly, introducing a deadlock (depending
             // on whether a write will wake up an already-blocked thread or this
             // Win32FileStream's thread).
+            Debug.Assert(Interop.mincore.GetFileType(_handle) != Interop.mincore.FileTypes.FILE_TYPE_PIPE);
 
-            // Do NOT change this to use a byte[] of length 0, or test test won't
-            // work.  Our ReadFile & WriteFile methods are special cased to return
-            // for arrays of length 0, since we'd get an IndexOutOfRangeException 
-            // while using C#'s fixed syntax.
-            byte[] bytes = new byte[1];
-            int errorCode = 0;
-            int r = 0;
+            byte* bytes = stackalloc byte[1];
+            int numBytesReadWritten;
+            int r = -1;
 
             // If the handle is a pipe, ReadFile will block until there
             // has been a write on the other end.  We'll just have to deal with it,
             // For the read end of a pipe, you can mess up and 
             // accidentally read synchronously from an async pipe.
             if (_canRead)
-            {
-#if USE_OVERLAPPED
-                r = ReadFileNative(_handle, bytes, 0, 0, null, out errorCode);
-#else
-                r = ReadFileNative(_handle, bytes, 0, 0, out errorCode);
-#endif
-            }
+                r = Interop.mincore.ReadFile(_handle, bytes, 0, out numBytesReadWritten, IntPtr.Zero);
             else if (_canWrite)
-            {
-#if USE_OVERLAPPED
-                r = WriteFileNative(_handle, bytes, 0, 0, null, out errorCode);
-#else
-                r = WriteFileNative(_handle, bytes, 0, 0, out errorCode);
-#endif
-            }
+                r = Interop.mincore.WriteFile(_handle, bytes, 0, out numBytesReadWritten, IntPtr.Zero);
 
-            if (errorCode == ERROR_INVALID_PARAMETER)
-                throw new ArgumentException(SR.Arg_HandleNotSync, "handle");
-            if (errorCode == Interop.mincore.Errors.ERROR_INVALID_HANDLE)
-                throw Win32Marshal.GetExceptionForWin32Error(errorCode, "<OS handle>");
+            if (r == 0)
+            {
+                int errorCode = GetLastWin32ErrorAndDisposeHandleIfInvalid(throwIfInvalidHandle: true);
+                if (errorCode == ERROR_INVALID_PARAMETER)
+                    throw new ArgumentException(SR.Arg_HandleNotSync, "handle");
+            }
         }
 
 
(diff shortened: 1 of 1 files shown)


stephentoub:
Hmm, I'm torn about this one.  @justinvp, how did you come across this: was it via code inspection, or did you notice it in some profile / performance report?  I ask because I'd expect constructing a FileStream via the SafeHandle ctor is done relatively infrequently, and this byte[] is very small, allocated only once per stream (which typically has lots of other allocations involved, including a fairly large buffer array), and is thrown away fairly quickly, so it's likely it'll never make it past gen0.  I'm wondering if it's really worth being cached.  I don't have a strong opinion about it either way, though.

justinvp:
I just noticed this while reviewing the code. I'm also torn for the same reasons you mention. Though, I tend to lean towards not allocating if it doesn't negatively impact the readability/complexity of the code, even if its a short-lived object.

Thinking about this more, what about just stack allocating this `byte[]` and calling the `ReadFile`/`WriteFile` interop methods directly? Then there's no heap allocation at all and no need to cache. And it'd actually simplify the code in some ways, as the #ifdefs could be removed (because we know we're only calling the synchronous methods). Something like (comments removed here for brevity):

```c#
private unsafe void VerifyHandleIsSync()
{
    byte* bytes = stackalloc byte[1];
    int r = 0;
    int numBytesReadWritten;

    if (_canRead)
        r = Interop.mincore.ReadFile(_handle, bytes, 0, out numBytesReadWritten, IntPtr.Zero);
    else if (_canWrite)
        r = Interop.mincore.WriteFile(_handle, bytes, 0, out numBytesReadWritten, IntPtr.Zero);

    if (r == 0)
    {
        int errorCode = Marshal.GetLastWin32Error();
        
        if (errorCode == ERROR_INVALID_PARAMETER)
            throw new ArgumentException(SR.Arg_HandleNotSync, "handle");

        if (errorCode == Interop.mincore.Errors.ERROR_INVALID_HANDLE)
        {
            _handle.Dispose();
            throw Win32Marshal.GetExceptionForWin32Error(errorCode, "<OS handle>");
        }
    }
}
```

justinvp:
I went ahead and updated the PR to use a stack allocated array.

Note: I also tested passing `null` for the bytes and that works just as well on my machine (Windows 10 Pro). However, the documentation for `ReadFile` and `WriteFile` doesn't indicate it's OK to use a null `lpBuffer`, so I've kept the stack allocated array instead of passing null.

stephentoub:
Thanks, Justin.  I'd considered this revised approach as well, but it ends up increasing the IL (and presumably the resulting assembly code) by more than the cached one-element byte array would consume, so it's not clear to me it's a win either.  If the error handling code could be factored out of this function and Read/WriteFileNative into a shared helper, this could be reasonable (I'd also hope to see a Debug.Assert at the beginning of this Verify function asserting that isAsync is false).

justinvp:
(Thanks, for taking a look again, Steve. I pushed some new changes a little too soon... Please ignore until I leave another comment saying its ready for review.)

justinvp:
I consolidated the error handling around `ERROR_INVALID_HANDLE` and disposing the handle, and added some asserts to the top of `VerifyHandleIsSync`. PTAL.

As part of the consolidation, I removed some superfluous `if` checks and comments in Read/WriteFileNative about pipe errors not being errors as these checks don't actually do anything useful other than returning -1 early before another `if` check for `ERROR_INVALID_HANDLE`, and the callers already have comments about the pipe errors not being errors.

Note that the check in `ReadFileNative` is for `ERROR_BROKEN_PIPE` or `ERROR_PIPE_NOT_CONNECTED`, but only `ERROR_BROKEN_PIPE` is being checked by the callers. I'm not sure if the lack of checking for `ERROR_PIPE_NOT_CONNECTED` is a bug or not, but it's unrelated to this change, and it doesn't look like `ERROR_PIPE_NOT_CONNECTED` was checked as far back as .NET 2.0 (based on the SSCLI 2.0 ROTOR source).

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -901,32 +890,11 @@ private long SeekCore(long offset, SeekOrigin origin)
...
-                // connection was dropped and the server closed the handle. However, 
-                // the client side handle is still open and even valid for certain 
-                // operations.
-                //
-                // Note that _parent.Dispose doesn't throw so we don't need to special case. 
-                // SetHandleAsInvalid only sets _closed field to true (without 
-                // actually closing handle) so we don't need to call that as well.
-                if (errorCode == Interop.mincore.Errors.ERROR_INVALID_HANDLE)
-                    _handle.Dispose();
+                int errorCode = GetLastWin32ErrorAndDisposeHandleIfInvalid(throwIfInvalidHandle: true);
```
stephentoub: I'm curious why you chose to pass ```true``` here for throwIfInvalidHandle.  It doesn't really matter, as we'll end up throwing the same exception regardless, just inside that method instead of from the next line here.

stephentoub:
Thanks for the revisions, Justin.   Looks much better now.

@ericstj, look ok to you?

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
justinvp: I went back and forth here as it doesn't matter either way.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -332,44 +332,33 @@ private static Interop.mincore.SECURITY_ATTRIBUTES GetSecAttrs(FileShare share)
...
+            int numBytesReadWritten;
+            int r = -1;
 
             // If the handle is a pipe, ReadFile will block until there
             // has been a write on the other end.  We'll just have to deal with it,
             // For the read end of a pipe, you can mess up and 
             // accidentally read synchronously from an async pipe.
             if (_canRead)
-            {
-#if USE_OVERLAPPED
```
ericstj: We should probably scrub the codebase of USE_OVERLAPPED now that we have an overlapped implementation for UWP.  I filed https://github.com/dotnet/corefx/issues/2722

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
justinvp: OK, I removed `throwIfInvalidHandle: true` here, to avoid future developers from asking the same question.

ericstj:
LGTM :shipit:

stephentoub:
@dotnet-bot test this please

stephentoub:
Thanks, Justin.

Issue #2581, opened 2015-07-31
Change path length check to only throw past extended length


JeremyKuhne:
Part of #2580.  Should only throw if resolved length + extended syntax is over 32,767.

JeremyKuhne:
#3001

PR #2700, opened 2015-08-10
Allow extended long path directories

JeremyKuhne:
Add methods for checking max length as it varies on path format.

Most code paths hit PathTooLong with GetFullPath- directories are a special
case as Win32 apis have a special limit (248) for directory creation.

#2581, #2411, #2579
@stephentoub, @mmitche, @weshaggard

### DIFF
--- a/src/Common/src/System/IO/PathInternal.Unix.cs
+++ b/src/Common/src/System/IO/PathInternal.Unix.cs
@@ -41,5 +41,22 @@ namespace System.IO
             Debug.Assert(Path.DirectorySeparatorChar == Path.AltDirectorySeparatorChar);
             return c == Path.DirectorySeparatorChar;
         }
+
+
+        /// <summary>
+        /// Returns true if the path is too long
+        /// </summary>
+        internal static bool IsPathTooLong(string fullPath)
+        {
+            return fullPath.Length >= Interop.libc.MaxPath;
+        }
+
+        /// <summary>
+        /// Returns true if the directory is too long
+        /// </summary>
+        internal static bool IsDirectoryTooLong(string fullPath)
+        {
+            return fullPath.Length >= Interop.libc.MaxPath;
+        }
     }
 }
--- a/src/Common/src/System/IO/PathInternal.Windows.cs
+++ b/src/Common/src/System/IO/PathInternal.Windows.cs
@@ -13,6 +13,9 @@ namespace System.IO
         internal const string UncPathPrefix = @"\\";
         internal const string UncExtendedPrefixToInsert = @"?\UNC\";
         internal const string UncExtendedPathPrefix = @"\\?\UNC\";
+        internal const int MaxShortPath = 260;
+        internal const int MaxShortDirectoryPath = 248;
+        internal const int MaxExtendedPath = short.MaxValue;
 
         internal static readonly char[] InvalidPathChars =
         {
@@ -31,6 +34,40 @@ namespace System.IO
             (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
             (char)31, '*', '?'
         };
+
+        /// <summary>
+        /// Returns true if the path is too long
+        /// </summary>
+        internal static bool IsPathTooLong(string fullPath)
+        {
+            if (IsExtended(fullPath))
+            {
+                return fullPath.Length >= MaxExtendedPath;
+            }
+            else
+            {
+                // Will need to be updated with #2581 to allow all paths to MaxExtendedPath
+                // minus legth of extended local or UNC prefix.
+                return fullPath.Length >= MaxShortPath;
+            }
+        }
+
+        /// <summary>
+        /// Returns true if the directory is too long
+        /// </summary>
+        internal static bool IsDirectoryTooLong(string fullPath)
+        {
+            if (IsExtended(fullPath))
+            {
+                return fullPath.Length >= MaxExtendedPath;
+            }
+            else
+            {
+                // Will need to be updated with #2581 to allow all paths to MaxExtendedPath
+                // minus legth of extended local or UNC prefix.
+                return fullPath.Length >= MaxShortDirectoryPath;
+            }
+        }
 
         /// <summary>
         /// Adds the extended path prefix (\\?\) if not already present.
--- a/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
+++ b/src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
@@ -124,7 +124,7 @@ namespace System.IO.FileSystem.Tests
             var components = IOInputs.GetValidPathComponentNames();
             Assert.All(components, (component) =>
             {
-                string path = @"\\?\" + IOServices.AddTrailingSlashIfNeeded(Path.Combine(testDir.FullName, component));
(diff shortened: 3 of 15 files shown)


src/Common/src/System/IO/PathInternal.Unix.cs
```diff
@@ -41,5 +41,22 @@ internal static bool IsDirectorySeparator(char c)
...
             return c == Path.DirectorySeparatorChar;
         }
+
+
+        /// <summary>
+        /// Returns true if the path is too long
+        /// </summary>
+        internal static bool IsPathTooLong(string fullPath)
+        {
+            return fullPath.Length >= Interop.libc.MaxPath;
```
stephentoub: Why is == considered too long?  Is MaxPath defined to be one larger than the longest path allowed?

stephentoub: I see from reading further in the changes that we were already doing this comparison, so this isn't new.  Is it correct though, on both Unix and Windows?

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
```diff
@@ -221,6 +221,17 @@ public void DirectoryLongerThanMaxPathAsPath_ThrowsPathTooLongException()
 
         [Fact]
         [PlatformSpecific(PlatformID.Windows)]
+        public void ExtendedDirectoryLongerThanLegacyMaxPathSucceeds()
+        {
+            var paths = IOInputs.GetPathsLongerThanMaxPath(useExtendedSyntax: true, includeExtendedMaxPath: false);
+            Assert.All(paths, (path) =>
+            {
+                Assert.NotNull(Create(path));
```
stephentoub: There's nothing else to validate about the created DirectoryInfo beyond that it's not null?

stephentoub:
Other than a few minor comments and the Unix compilation failures (looks like you're missing a .cs file in one or more .csproj files), LGTM.

cc: @ianhays

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
```diff
@@ -21,7 +21,11 @@ internal static class IOInputs
     // Unix values vary system to system; just using really long values here likely to be more than on the average system
     public static readonly int MaxDirectory = 247; // Does not include trailing \0. This the maximum length that can be passed to APIs taking directory names, such as Directory.CreateDirectory, Directory.Move
     public static readonly int MaxPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 259 : 10000;      // Does not include trailing \0.
-    public static readonly int MaxComponent = 255;
+    public static readonly int MaxExtendedPath = short.MaxValue - 1;
```
akoeplinger: The -1 is because of trailing \0 right? I'd suggest adding a comment like above that highlights this.

JeremyKuhne: Yes, as above. I'll add the comment.

src/System.IO.FileSystem/tests/Directory/CreateDirectory.cs
JeremyKuhne: I could check exists instead and let it null-ref.  I'll update later.

src/Common/src/System/IO/PathInternal.Unix.cs
JeremyKuhne: I believe so.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -33,6 +36,40 @@ internal static partial class PathInternal
...
+        internal static bool IsPathTooLong(string fullPath)
+        {
+            if (IsExtended(fullPath))
+            {
+                return fullPath.Length >= MaxExtendedPath;
+            }
+            else
+            {
+                // Will need to be updated with #2581 to allow all paths to MaxExtendedPath
+                // minus legth of extended local or UNC prefix.
```
MarcelGosselin: typo: le_n_gth

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -33,6 +36,40 @@ internal static partial class PathInternal
...
+        internal static bool IsDirectoryTooLong(string fullPath)
+        {
+            if (IsExtended(fullPath))
+            {
+                return fullPath.Length >= MaxExtendedPath;
+            }
+            else
+            {
+                // Will need to be updated with #2581 to allow all paths to MaxExtendedPath
+                // minus legth of extended local or UNC prefix.
```
MarcelGosselin: ditto

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
```diff
@@ -18,10 +18,20 @@ internal static class IOInputs
     public static bool CaseSensitive { get { return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) | RuntimeInformation.IsOSPlatform(OSPlatform.OSX); } }
     public static bool CaseInsensitive { get { return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) | RuntimeInformation.IsOSPlatform(OSPlatform.OSX); } }
 
-    // Unix values vary system to system; just using really long values here likely to be more than on the average system
-    public static readonly int MaxDirectory = 247; // Does not include trailing \0. This the maximum length that can be passed to APIs taking directory names, such as Directory.CreateDirectory, Directory.Move
-    public static readonly int MaxPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 259 : 10000;      // Does not include trailing \0.
-    public static readonly int MaxComponent = 255;
+    // Max path length (minus trailing \0). Unix values vary system to system; just using really long values here likely to be more than on the average system.
+    public static readonly int MaxPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 259 : 10000;
```
ianhays: Doesn't the current Unix implementation fail with any Unix paths longer than the Windows MAX_PATH or has that been fixed already? If it does still fail as Windows fails, then we could simplify this to 259 for both cases for a tighter test. I seem to recall you mentioning Unix long paths work normally now though.

stephentoub: > Doesn't the current Unix implementation fail with any Unix paths longer than the Windows MAX_PATH or has that been fixed already?

That should have been fixed several weeks ago.  If you find anything to the contrary, please shout.  The only such issues that are known are those that hit the Path implementation in mscorlib and Path stuff in the libcoreclr PAL, and those should be restricted to support implemented in mscorlib and the runtime, e.g. assembly loading, resource loading, etc.

ianhays: Thanks for the confirmation! In that case, LGTM.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Thanks- going to ignore this one for now as the change fixes a build issue. I'll be changing this comment when I add the implicit support.

PureKrome:
:tada: :sparkles: :balloon:

niemyjski:
YAY!

nzthiago:
image

PR #2702, opened 2015-08-10
Change Mac OSX to "OS X"

jakepetroules:
OS X is the official, canonical name of Apple's desktop operating system.

It has not been called "Mac OS X" since 2012, and there is always a space between "OS" and "X" when written as a proper name.

"OS X is the operating system that powers every Mac." - http://www.apple.com/osx/what-is/


stephentoub:
LGTM. Thanks.

PR #2707, opened 2015-08-10
Use Array.Copy in ImmutableArray.Builder

James-Ko:
Replaced some for-loops I found with Array.Copy and Array.Clone.

I couldn't optimize the `AddRange(IEnumerable<T>)` overload because it requires use of the extension method made in PR #2643, so I'll submit a new commit there soon.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
@@ -269,11 +269,7 @@ namespace System.Collections.Immutable
                 var offset = this.Count;
                 this.Count += items.Length;
 
-                var nodes = _elements;
-                for (int i = 0; i < items.Length; i++)
-                {
-                    nodes[offset + i] = items[i];
-                }
+                Array.Copy(items, 0, _elements, offset, items.Length);
             }
 
             /// <summary>
@@ -287,11 +283,7 @@ namespace System.Collections.Immutable
                 var offset = this.Count;
                 this.Count += items.Length;
 
-                var nodes = _elements;
-                for (int i = 0; i < items.Length; i++)
-                {
-                    nodes[offset + i] = items[i];
-                }
+                Array.Copy(items, 0, _elements, offset, items.Length);
             }
 
             /// <summary>
@@ -302,16 +294,12 @@ namespace System.Collections.Immutable
             public void AddRange(T[] items, int length)
             {
                 Requires.NotNull(items, "items");
-                Requires.Range(length >= 0, "length");
+                Requires.Range(length >= 0 && length <= items.Length, "length");
 
                 var offset = this.Count;
                 this.Count += length;
 
-                var nodes = _elements;
-                for (int i = 0; i < length; i++)
-                {
-                    nodes[offset + i] = items[i];
-                }
+                Array.Copy(items, 0, _elements, offset, length);
             }
 
             /// <summary>
@@ -420,14 +408,9 @@ namespace System.Collections.Immutable
             /// </summary>
             public T[] ToArray()
             {
-                var tmp = new T[this.Count];
-                var elements = _elements;
-                for (int i = 0; i < tmp.Length; i++)
-                {
-                    tmp[i] = elements[i];
-                }
-
-                return tmp;
+                T[] result = new T[this.Count];
+                Array.Copy(_elements, 0, result, 0, this.Count);
+                return result;
             }
 
             /// <summary>
--- a/src/System.Collections.Immutable/tests/ImmutableArrayBuilderTest.cs
+++ b/src/System.Collections.Immutable/tests/ImmutableArrayBuilderTest.cs
@@ -133,7 +133,7 @@ namespace System.Collections.Immutable.Test
             Assert.Throws<ArgumentNullException>(() => builder1.AddRange((int[])null));
             Assert.Throws<ArgumentNullException>(() => builder1.AddRange(null, 42));
             Assert.Throws<ArgumentOutOfRangeException>(() => builder1.AddRange(new int[0], -1));
-            Assert.Throws<IndexOutOfRangeException>(() => builder1.AddRange(new int[0], 42));
+            Assert.Throws<ArgumentOutOfRangeException>(() => builder1.AddRange(new int[0], 42));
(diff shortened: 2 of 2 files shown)


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
```
stephentoub: Why is this using necessary?  The code is in the System.Collections.Immutable namespace, so the System namespace is implicitly available.

James-Ko: I never knew that, actually. I'll revert that change, then.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -420,14 +409,7 @@ public bool Contains(T item)
...
             {
-                var tmp = new T[this.Count];
-                var elements = _elements;
-                for (int i = 0; i < tmp.Length; i++)
-                {
-                    tmp[i] = elements[i];
-                }
-
-                return tmp;
+                return (T[])_elements.Clone();
```
stephentoub: This is not a safe/equivalent change.  _elements.Length may be larger than Count, with unfilled indices at the end.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -307,11 +300,7 @@ public void AddRange(T[] items, int length)
                 var offset = this.Count;
                 this.Count += length;
 
-                var nodes = _elements;
-                for (int i = 0; i < length; i++)
-                {
-                    nodes[offset + i] = items[i];
-                }
+                Array.Copy(items, 0, _elements, offset, length);
```
stephentoub: This is going to be a behavioral change in the case of some invalid input, as an ArgumentException will be generated instead of an IndexOutOfRangeExceptions in the case where length is larger than items.Length.

stephentoub:
cc: @AArnott

James-Ko:
Fixed some changes, everything should be working now.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -304,14 +296,13 @@ public void AddRange(T[] items, int length)
                 Requires.NotNull(items, "items");
                 Requires.Range(length >= 0, "length");
 
+                if (length > items.Length)
+                    throw new IndexOutOfRangeException("length");
```
stephentoub: The parameter to the ctor is a message rather than a parameter name; this should probably just use the parameterless ctor and let it use the default message.  I'd suggest also adding a comment to indicate this is being done for backwards compat.

This is, of course, assuming that we care about this exception not changing.  If we're ok with it changing, since it probably should have been an argument exception in the first place, then we could instead do argument validation with Requires.Range and remove this (tests would also need to be updated, as there are tests validating this behavior).  @terrajobst?  @AArnott?

AArnott: Personally I'd be OK with the change in behavior by throwing the more appropriate exception. Folks shouldn't be catching the old exception anyway.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -420,14 +411,9 @@ public bool Contains(T item)
...
             {
-                var tmp = new T[this.Count];
-                var elements = _elements;
-                for (int i = 0; i < tmp.Length; i++)
-                {
-                    tmp[i] = elements[i];
-                }
-
-                return tmp;
+                T[] result = new T[_count];
```
AArnott: I think we try to reuse our own properties rather than fields when available. Can you change this `_count` to `this.Count` as it was before?

AArnott:
If we're confident that Array.Copy is *always* faster than our own loop, this LGTM. 
What about ref types? or structs? or very large structs? Or blittable types? (just to through some types out there to consider when making this optimization... I'd hate to unknowingly tank perf for one scenario to optimize for another without weighing the pros and cons)

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
stephentoub: @terrajobst, ok change to make?

James-Ko:
@AArnott I made a perf test for all the scenarios [here](https://dotnetfiddle.net/PUtuxk). It's not uncommon for the test to print all `True` in a row, so I'd say these changes are valid.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
stephentoub: @weshaggard, are you ok with changing the exception that emerges here?

weshaggard: I agree with @AArnott that people should not be catching this anyway. Assuming failures Requires.Range checks throw and ArgumentOutOfRangeException then I'm OK with this change.

James-Ko: Fixed this and the associated test in the new commit.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
James-Ko: Done.

stephentoub:
LGTM. Thanks, @James-Ko.

PR #2708, opened 2015-08-10
Reduce allocations in XmlStreamNodeWriter

jamesqo:
Save a few allocations whenever `UnsafeGetUTF8Length` or `UnsafeGetUTF8Chars` are called. The second method also only allocates a byte buffer if its contents are too large to be written directly to the destination array.

### DIFF
--- a/src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
+++ b/src/System.Private.DataContractSerialization/src/System/Xml/XmlStreamNodeWriter.cs
@@ -76,16 +76,17 @@ namespace System.Xml
             }
         }
 
-        private int GetByteCount(char[] chars)
+        private Encoding EncodingOrDefault
         {
-            if (_encoding == null)
+            get
             {
-                return s_UTF8Encoding.GetByteCount(chars);
+                return _encoding ?? s_UTF8Encoding;
             }
-            else
-            {
-                return _encoding.GetByteCount(chars);
-            }
+        }
+
+        private unsafe int GetByteCount(char* pChar, int length)
+        {
+            return EncodingOrDefault.GetByteCount(pChar, length);
         }
 
         protected byte[] GetBuffer(int count, out int offset)
@@ -348,12 +349,8 @@ namespace System.Xml
             if (chars == charsMax)
                 return charCount;
 
-            char[] chArray = new char[charsMax - chars];
-            for (int i = 0; i < chArray.Length; i++)
-            {
-                chArray[i] = chars[i];
-            }
-            return (int)(chars - (charsMax - charCount)) + GetByteCount(chArray);
+            int charLength = (int)(charsMax - chars);
+            return charCount - charLength + GetByteCount(chars, charLength);
         }
 
         /// <SecurityNote>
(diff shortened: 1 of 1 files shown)


stephentoub:
@James-Ko, thank you for continuing to submit improvements to corefx.  As we've previously asked, we really need you to do some basic validation locally before submitting a PR, including compiling and running tests with the changes.  This PR doesn't compile.  I'll leave this PR open, but in the future I plan to close any such PRs that don't compile, after which point you're welcome to address the issues and resubmit.

stephentoub:
cc: @krwq

James-Ko:
@stephentoub: I understand. I only have access to an ARM Linux laptop with me at the moment, so I won't push anything for the next week so I can get back and build/test/submit. Thanks for the tip.

James-Ko:
Pushed a new commit letting it build, with a few readability improvements. `UnsafeGetUTF8Chars` is still very, very hard to read however.

James-Ko:
Looks like a test in System.Threading.Tasks.WaitAllAny is failing? I don't think this is related to my PR.

stephentoub:
> Looks like a test in System.Threading.Tasks.WaitAllAny is failing? I don't think this is related to my PR.

https://github.com/dotnet/corefx/pull/2884

stephentoub:
@dotnet-bot test this please

jasonwilliams200OK:
@stephentoub, is there currently a mechanism to submit a benchmark test as part of PR, which will run on CI build and compare the stats with previous commit (master `HEAD`)?

stephentoub:
> is there currently a mechanism to submit a benchmark test as part of PR, which will run on CI build and compare the stats with previous commit (master HEAD)?

Not yet.

stephentoub:
@jamesqo, this is adding some non-trivial additional complexity to the implementation.  Could you share some details on the performance benefit this makes on the overall operation being performed?

jamesqo:
@stephentoub Sorry, but I can't see how we would write perf tests here when the class is internal. All I can say is that large chunks of text are usually common for such a method, so to avoid newing a `string` or potentially a `byte[]` could save a lot of memory.

stephentoub:
That's why I asked about the overall operation. What publicly exposed APIs use this, and how do your changes affect the performance of such methods?

stephentoub:
@krwq, any thoughts on this?  @jamesqo, were you able to demonstrate an improvement from this?

jamesqo:
@stephentoub Hey, really sorry I didn't respond to you earlier but I've been busy the last few weeks. I'll post a performance test by next morning if I have time.

jamesqo:
Not sure how to put this but... the test I wrote seems to be causing strange behavior? It's reporting that `00:00:00` has elapsed after the test has been going on for 2 seconds, and the debugger says it's "unable to read memory." I stepped through `UnsafeGetUTF8Chars` to make sure it was going down the correct branch, so I'm not exactly sure why this is happening.

[**Here's**](https://gist.github.com/jamesqo/fb0a0133a6772d6a37a1) the test, I didn't screw up on pointer arithmetic or anything did I? @krwq, can you check if the changes made here are valid?

stephentoub:
> the test

Thanks, @jamesqo.  Sorry if I wasn't clear, but when I asked about the overall operation, what I meant was if you could write a test (not using reflection) that exercised some public entrypoint that would see a benefit from this.

@khdang, opinions on this change?

khdang:
@jamesqo I agree with Stephen that we would like to understand the related scenarios and more details of the improvement after the changes. Otherwise it would be unclear whether the benefit is worth the complexity of the new code.

stephentoub:
@jamesqo, I'm going to close this PR out for now.  If you can share an example where this provides a measurable improvement (via public APIs), whether in throughput or memory reductions, please resubmit.  We just want to make sure there's value being gained from the extra complexity, and there's no test coverage for some of the new code paths introduced, which is concerning anytime but especially when unsafe code is involved. Regardless, thanks for your continued interest in contributing.

jamesqo:
@stephentoub I understand, no problem.

PR #2712, opened 2015-08-10
Add the common RSA tests to RSACng.Test.

AtsushiKan:
This required some fixes to RSACng that will need to be
ported to 4.6.

- Set the export policies so that ImportParameters(true)
  works.

- Throw the right (or wrong, depending on your pov)
  exception if ImportParameters() is given an
  RSAParameters with null fields.

- VerifyHash(). Don't throw CryptographicException
  if the hash doesn't verify. Just return false like
  the nice docs say.

We have two tests that still fail and are being
investigated with the Win Crypto guys. Since it is
likely the bug is on their side, we'll split these
off into their own file so that RSACng can exclude them.

### DIFF
--- a/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
+++ b/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
@@ -38,9 +38,21 @@ namespace System.Security.Cryptography
             unsafe
             {
                 if (parameters.Exponent == null || parameters.Modulus == null)
-                    throw new ArgumentException(SR.Cryptography_InvalidRsaParameters);
+                    throw new CryptographicException(SR.Cryptography_InvalidRsaParameters);
 
-                bool includePrivate = parameters.P != null && parameters.Q != null;
+                bool includePrivate;
+                if (parameters.D == null)
+                {
+                    includePrivate = false;
+                    if (parameters.P != null || parameters.DP != null || parameters.Q != null || parameters.DQ != null || parameters.InverseQ != null)
+                        throw new CryptographicException(SR.Cryptography_InvalidRsaParameters);
+                }
+                else
+                {
+                    includePrivate = true;
+                    if (parameters.P == null || parameters.DP == null || parameters.Q == null || parameters.DQ == null || parameters.InverseQ == null)
+                        throw new CryptographicException(SR.Cryptography_InvalidRsaParameters);
+                }
 
                 //
                 // We need to build a key blob structured as follows:
@@ -94,7 +106,11 @@ namespace System.Security.Cryptography
                     Debug.Assert(offset == blobSize, "offset == blobSize");
                 }
                 CngKeyBlobFormat blobFormat = includePrivate ? s_rsaPrivateBlob : s_rsaPublicBlob;
-                Key = CngKey.Import(rsaBlob, blobFormat);
+
+                CngKey newKey = CngKey.Import(rsaBlob, blobFormat);
+                newKey.ExportPolicy |= CngExportPolicies.AllowPlaintextExport;
+
+                Key = newKey;
             }
         }
 
--- a/src/System.Security.Cryptography.Cng/tests/RsaCngTests.cs
+++ b/src/System.Security.Cryptography.Cng/tests/RsaCngTests.cs
@@ -16,88 +16,11 @@ namespace System.Security.Cryptography.Cng.Tests
     public static class RsaCngTests
     {
         [Fact]
-        public static void EncryptDecryptRoundtrip()
-        {
-            {
-                byte[] plainText = "87abc91203fa927823".HexToByteArray();
-                TestEncryptDecryptRoundTrip(plainText, RSAEncryptionPadding.Pkcs1, 0x100);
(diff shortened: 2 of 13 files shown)


src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
```diff
@@ -94,7 +97,15 @@ public override void ImportParameters(RSAParameters parameters)
                     Debug.Assert(offset == blobSize, "offset == blobSize");
                 }
                 CngKeyBlobFormat blobFormat = includePrivate ? s_rsaPrivateBlob : s_rsaPublicBlob;
-                Key = CngKey.Import(rsaBlob, blobFormat);
+
+                CngKey newKey = CngKey.Import(rsaBlob, blobFormat);
+                CngExportPolicies exportPolicy = newKey.ExportPolicy | CngExportPolicies.AllowPlaintextExport;
+                using (SafeNCryptKeyHandle newKeyHandle = newKey.Handle)
```
stephentoub: It's ok to Dispose of newKey.Handle but then assign newKey to Key?

stephentoub:
cc: @bartonjs

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
AtsushiKan: Yeah, the Handle property duplicates the SafeHandle so it has to be disposed anyway, so it was better to do it deterministically.

Having said that, I think I’m going to get rid of the need to Duplicate entirely and tighten the code by exposing an internal ExportPolicy setter in CngKey. I’ve just been waiting to see if there’s other feedback to incorporate.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
```diff
@@ -38,10 +38,13 @@ public override void ImportParameters(RSAParameters parameters)
             unsafe
             {
                 if (parameters.Exponent == null || parameters.Modulus == null)
-                    throw new ArgumentException(SR.Cryptography_InvalidRsaParameters);
+                    throw new CryptographicException(SR.Cryptography_InvalidRsaParameters);
 
                 bool includePrivate = parameters.P != null && parameters.Q != null;
 
+                if (includePrivate && parameters.DP == null)
```
bartonjs: Why just DP? Why not also D, DQ, and InverseQ?

The state is that D, P, Q, DP, DQ, and InverseQ must either all be null, or all be not-null.

AtsushiKan: Errm… probably because the null-DP case was the only one we were testing. We’re only writing out P and Q, so strictly speaking, there’s no reason to enforce rules on any of DQ/DQ/D/IQ – but maybe we should be enforcing them anyway to be consistent with OpenSSL/Csp. Does a null-check on each of these sound like a reasonably coherent behavior wrt the RSA family?

src/System.Security.Cryptography.Cng/tests/RsaCngTests.cs
```diff
@@ -237,20 +237,6 @@ private static void TestSignAndVerifyDataFromStream(int messageSize)
         }
 
         [Fact]
-        public static void ImportExport()
```
bartonjs: The only test that I see in this file that doesn't seem to currently be covered by the RSA algorithm tests is `VerifyHashPss`, which is just because the IRSAProvider interface doesn't yet expose support levels (CSP doesn't support PSS, and our OpenSSL wrapper doesn't yet).

So I'd expect pretty much everything except the PSS test to be deleted, not just this one.

AtsushiKan: Focus was on the breaking changes from 4.6, not on deduplicating the tests. But since I’m updating the PR anyway, I’ve gone ahead and nuked everything except the PSS tests.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.ImportExport.cs
bartonjs: The CSP implementation enforces that they're all present; OpenSSL would function just fine with only D, but since this is a data exchange type/contract it went with the "specify everything" approach; and I feel it's better to have them all match here with the most stringent requirements; otherwise you get weird cross-plat behaviors with different minimum requirements.

AtsushiKan: I was trying not to change 4.6 behavior any more than necessary, but I suspect we’re losing that battle anyway…

stephentoub:
@dotnet-bot test this please

bartonjs:
With some appropriate squashing, the changes look good to me.  Just be sure to wait to see that CI agrees :smile:.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #2713, opened 2015-08-10
Localize Microsoft.CSharp with machine translations

naamunds:
This change adds localization to the Microsoft.CSharp project by using
the Multilingual App Toolkit (MAT). It includes machine translations as
generated by MAT. This produces localized satellite assemblies when
building.

When building on a machine without MAT, there will be a warning that MAT
is not installed.  This means that the .xlf files (translation files)
and localized resources files will not be synced with the neutral-
language resource files during that build; they can be synced by
building on a machine with MAT installed.

Localization to the other projects in fxcore will be added in the
future.

### DIFF
--- a/src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
+++ b/src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
@@ -1,0 +1,520 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+  <file datatype="xml" source-language="en-US" target-language="de" original="MICROSOFT.CSHARP/RESOURCES/STRINGS.RESX" tool-id="MultilingualAppToolkit" product-name="n/a" product-version="n/a" build-num="n/a">
+    <header>
+      <tool tool-id="MultilingualAppToolkit" tool-name="Multilingual App Toolkit" tool-version="4.0.1387.0" tool-company="Microsoft" />
+    </header>
+    <body>
+      <group id="MICROSOFT.CSHARP/RESOURCES/STRINGS.RESX" datatype="resx">
+        <trans-unit id="InternalCompilerError" translate="yes" xml:space="preserve">
+          <source>An unexpected exception occurred while binding a dynamic operation</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Unerwartete Ausnahme beim Binden eines dynamischen Vorgangs</target>
+        </trans-unit>
+        <trans-unit id="BindRequireArguments" translate="yes" xml:space="preserve">
+          <source>Cannot bind call with no calling object</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Ein Aufruf ohne aufrufendes Objekt kann nicht gebunden werden.</target>
+        </trans-unit>
+        <trans-unit id="BindCallFailedOverloadResolution" translate="yes" xml:space="preserve">
+          <source>Overload resolution failed</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Fehler bei der Überladungsauflösung</target>
+        </trans-unit>
+        <trans-unit id="BindBinaryOperatorRequireTwoArguments" translate="yes" xml:space="preserve">
+          <source>Binary operators must be invoked with two arguments</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Binäre Operatoren müssen mit zwei Argumenten aufgerufen werden.</target>
+        </trans-unit>
+        <trans-unit id="BindUnaryOperatorRequireOneArgument" translate="yes" xml:space="preserve">
+          <source>Unary operators must be invoked with one argument</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Unäre Operatoren müssen mit einem Argument aufgerufen werden.</target>
+        </trans-unit>
+        <trans-unit id="BindPropertyFailedMethodGroup" translate="yes" xml:space="preserve">
+          <source>The name '{0}' is bound to a method and cannot be used like a property</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Der Name "{0}" ist an eine Methode gebunden und kann nicht wie eine Eigenschaft verwendet werden.</target>
+        </trans-unit>
+        <trans-unit id="BindPropertyFailedEvent" translate="yes" xml:space="preserve">
+          <source>The event '{0}' can only appear on the left hand side of +</source>
+          <target state="needs-review-translation" state-qualifier="mt-suggestion">Das Ereignis "{0}" kann nur verwendet werden, auf der linken Seite des +</target>
+        </trans-unit>
+        <trans-unit id="BindInvokeFailedNonDelegate" translate="yes" xml:space="preserve">
+          <source>Cannot invoke a non-delegate type</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Ein Nichtdelegattyp kann nicht aufgerufen werden.</target>
+        </trans-unit>
+        <trans-unit id="BindImplicitConversionRequireOneArgument" translate="yes" xml:space="preserve">
+          <source>Implicit conversion takes exactly one argument</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Eine implizite Konvertierung akzeptiert genau ein Argument.</target>
+        </trans-unit>
+        <trans-unit id="BindExplicitConversionRequireOneArgument" translate="yes" xml:space="preserve">
+          <source>Explicit conversion takes exactly one argument</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Eine explizite Konvertierung akzeptiert genau ein Argument.</target>
+        </trans-unit>
+        <trans-unit id="BindBinaryAssignmentRequireTwoArguments" translate="yes" xml:space="preserve">
+          <source>Binary operators cannot be invoked with one argument</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Binäre Operatoren können nicht mit einem Argument aufgerufen werden.</target>
+        </trans-unit>
+        <trans-unit id="BindBinaryAssignmentFailedNullReference" translate="yes" xml:space="preserve">
+          <source>Cannot perform member assignment on a null reference</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Für einen NULL-Verweis kann keine Memberzuweisung ausgeführt werden.</target>
+        </trans-unit>
+        <trans-unit id="NullReferenceOnMemberException" translate="yes" xml:space="preserve">
+          <source>Cannot perform runtime binding on a null reference</source>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">Die Laufzeitbindung kann für einen NULL-Verweis nicht ausgeführt werden.</target>
+        </trans-unit>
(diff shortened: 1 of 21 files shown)


stephentoub:
@dotnet-bot test this please

stephentoub:
cc: @weshaggard, @jhendrixMSFT, @tarekgh

stephentoub:
> When building on a machine without MAT, there will be a warning that MAT is not installed

Can we avoid that?  The vast majority of folks building the corefx repo will not have these tools, and we're trying to keep the number of warnings during a run at 0... otherwise real warnings get lost in the noise.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
```diff
@@ -15,6 +15,10 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
+  <PropertyGroup Label="MultilingualAppToolkit">
+    <MultilingualAppToolkitVersion>4.0</MultilingualAppToolkitVersion>
+    <MultilingualFallbackLanguage>en-US</MultilingualFallbackLanguage>
+  </PropertyGroup>
```
stephentoub: If we're going to have all of this boilerplate in each project, can we do it in buildtools instead?

src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
```diff
@@ -0,0 +1,520 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
stephentoub: This is a substantial amount of machine-generated XML!  Could we instead adopt an approach where if the file isn't available, it's generated as a temporary during build and then thrown away?  That way we'd only need to check in the files for which manual editing was necessary.  We could potentially even make it so that you only had to check in override content, with everything else filled in by the build system automatically.  I'm nervous about the fact that, for a single project, this change is adding 8000 lines of code.  We have ~100 projects in corefx.  Back of the napkin estimate, and this means ~800,000 lines of boilerplate eventually checked in to the repo.

akoeplinger:
@naamunds I skimmed over the German translations (I'm a native speaker) and some of them sound a bit weird. Should I file improvements as part of this PR or in a follow-up PR? Don't you already have the localized messages from Desktop .NET, why use machine translation?

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
tarekgh: I second Stephen here. in build tools we can have the defaults for all projects and each project can customize it if need to.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
```diff
@@ -216,9 +220,33 @@
...
     <EmbeddedResource Include="Resources\$(AssemblyName).rd.xml" />
+    <EmbeddedResource Include="Resources\Strings.de.resx" />
+    <EmbeddedResource Include="Resources\Strings.es.resx" />
+    <EmbeddedResource Include="Resources\Strings.fr.resx" />
+    <EmbeddedResource Include="Resources\Strings.it.resx" />
+    <EmbeddedResource Include="Resources\Strings.ja.resx" />
+    <EmbeddedResource Include="Resources\Strings.ko.resx" />
+    <EmbeddedResource Include="Resources\Strings.ru.resx" />
+    <EmbeddedResource Include="Resources\Strings.zh-Hans.resx" />
+    <EmbeddedResource Include="Resources\Strings.zh-Hant.resx" />
```
tarekgh: are we going to embed all these resources in Microsoft.CSharp? or these will be embedded in the satellite assembly?

naamunds:
Yeah, that makes sense.  I’ll change the Warning in the MATPrerequisite target to a Message with high importance.

naamunds:
@dotnet-bot test this please

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
naamunds: That's a good idea.  @weshaggard would you suggest making a new .targets file in buildtools\src\Microsoft.DotNet.Build.Tasks\Targets for localization?  I'm thinking it would include the two properties at this point in the code, and then also the Import and MATPrerequisite target at lines 248-251.

src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
naamunds: I suspect that many of the translation files will need manual editing, because the machine translations typically aren't that great, as Alexander noted is the case with these translations.  So to get much value out of this approach, I think we'd have to support the ability to provide override content and otherwise dynamically use machine translations.

I see a few drawbacks to that approach:
- The machine translations could change over time if they're being generated dynamically.  I think MAT uses an online service like Bing Translate or Google Translator, and if the translations change, then something that was reviewed and considered correct initially may become incorrect later without us knowing.
- It would make it more difficult for people to review the translations if they're not checked in.  Rather than just reviewing the translations in source control, they would have to locally generate the machine translations themselves.
- Setting up the infrastructure to generate the .xlf files during the build using dynamically-generated machine translations combined with the override content would require substantial investment.  As a rough initial estimate I would put it at 2 weeks, but it could even be more than that.

I agree that the translations are a lot of code, but checking in translations is standard practice (as far as I know), and is what we do in the main .NET Framework code in TFS.

naamunds:
> Should I file improvements as part of this PR or in a follow-up PR? Don't you already have the localized messages from Desktop .NET, why use machine translation?

@akoeplinger I think any improvements should be done in a separate PR.  We do have translations for some of the fxcore projects (including Microsoft.CSharp) in our legacy code in TFS.  @markwilkie do we have a plan for porting over those translations?  I was thinking that would be done in a separate commit, but maybe it would make sense to do it now.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
naamunds: Each resource will be embedded in a satellite assembly.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
weshaggard: Yes I think we should have a localization targets file that contains the boilerplate.

src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
weshaggard: @naamunds there was some talk about potentially having these in another parallel repo, something to consider. Not even sure if MAT actually supports that scenario. Beyond the actual size of the content do we have any ideas on the impact of the build times that this adds?

weshaggard:
>@stephentoub Yeah, that makes sense.  I’ll change the Warning in the MATPrerequisite target to a Message with high importance.
The ideas would be to cut down on the amount of text going to the output. Do we think it is wroth even having a message that is high importance? Perhaps just a normal or low importance so it shows in the log file but not on the console output. What do we expect people to do with a bunch of messages from all the projects if the tools aren't installed? We also need to take our x-plat builds into account which will not even have the ability to install the MAT tools.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
```diff
@@ -216,9 +220,33 @@
...
+    <XliffResource Include="MultilingualResources\Microsoft.CSharp.ru.xlf" />
+    <XliffResource Include="MultilingualResources\Microsoft.CSharp.zh-Hans.xlf" />
+    <XliffResource Include="MultilingualResources\Microsoft.CSharp.zh-Hant.xlf" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+  <Import Project="$(MSBuildExtensionsPath)\Microsoft\Multilingual App Toolkit\Microsoft.Multilingual.ResxResources.targets" Label="MultilingualAppToolkit" Condition="Exists('$(MSBuildExtensionsPath)\Microsoft\Multilingual App Toolkit\Microsoft.Multilingual.ResxResources.targets')" />
+  <Target Name="MATPrerequisite" BeforeTargets="PrepareForBuild" Condition="!Exists('$(MSBuildExtensionsPath)\Microsoft\Multilingual App Toolkit\Microsoft.Multilingual.ResxResources.targets')" Label="MultilingualAppToolkit">
+    <Warning Text="$(MSBuildProjectFile) is Multilingual build enabled, but the Multilingual App Toolkit is unavailable during the build. If building with Visual Studio, please check to ensure that toolkit is properly installed." />
```
weshaggard: We should definitely avoid this warning as there will be builds, i.e. x-plat builds, that will not even have the ability to fix this warning. What happens if the tools are not actually present?

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
weshaggard: Do we expect all the libraries to have the same languages? If so we could consider adding these lists to the localization targets file to reduce the boilerplate more as well.

How did you chose the languages? I assume that list is subject to change?

src/Microsoft.CSharp/src/MultilingualResources/Microsoft.CSharp.de.xlf
naamunds: I didn't notice a major impact on build time, but I haven't actually timed it.  For the official builds, where we want localization, I don't see a way that we could avoid this impact (even if it is significant).  I think that the approach used in this PR, with the .xlf and localized .resx files directly in the project, would be the most performant approach.  That's actually another downside to the approach of dynamically generating machine translations with the build; when I did it for just the Microsoft.CSharp project, it took somewhere around 30 seconds.  I think putting them in a separate repo and dynamically compiling them into the project with the build would add significant time to the build also.

One thing that I think we should consider is supporting the ability to not build localization when building locally, in order to reduce build time.  I was thinking that that would be something that we would consider after getting the initial localization infrastructure in place, because it's lower priority.

markwilkie:
There is no plan (yet) regarding the specifics of porting the existing translations over.  Yes, I agree that doing this as part of a separate commit is good.  I recommend validating the tooling and process and then tactically tackling getting the translations themselves up to date.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
naamunds: I think that we do expect them all to have the same languages.  I chose these languages, which are called the VS9, because they are what @markwilkie said we should use for now.  My impression from the conversation with him is that that's subject to change.

How are you thinking I would add this list to the localization.targets file?  Would I put the EmbeddedResource and XliffResource elements directly in there, using something like "$(MSBuildProjectName).de.xlf" for the name of the XliffResources?

weshaggard: Yes that is how I was envisioning it.

src/Microsoft.CSharp/src/Microsoft.CSharp.csproj
naamunds: I'll change this to a default-importance Message, which will show up in the build log but not on the console.  I'll also change the text of it to be tailored to our team.

Also, Cameron from the MAT team said that they would like to allow MAT to be installed on Linux and Mac at some point in the future, so hopefully it will eventually be possible for developers to do x-plat builds with MAT.

naamunds:
I've moved the loc code from Microsoft.CSharp.csproj to buildtools in https://github.com/dotnet/buildtools/pull/234.

naamunds:
Based on feedback in https://github.com/dotnet/buildtools/pull/234, I'm removing the <IncludeLocalizationImport> property from Microsoft.CSharp.csproj and instead conditioning the localization target on the existence of an XLF file.  This means that there won't be any changes to Microsoft.CSharp.csproj with this pull request.

naamunds:
I've submitted https://github.com/dotnet/buildtools/pull/241 to address the build failures here.  When that pull request is merged in, I'll uptake the new version of BuildTools here, which should fix these build failures.

jhendrixMSFT:
LGTM, although I believe you'll want to submit this new content as dotnet-bot.

PR #2729, opened 2015-08-11
Fix very sporadically failing Complex test

stephentoub:
I'd previously "fixed" this same test by incorporating infinity support, but I needed to do the comparisons based on IsNegative/PositiveInfinity rather than by comparing the values directly.

cc: @mellinoe


joshfree:
:+1:

PR #2730, opened 2015-08-11
Reworked the tests for System.IO.Pipes

ianhays:
I made a number of changes to our current Pipes tests and added several hundred new tests for edge cases that were previously ignored, particularly with Named Pipes defined with an InOut PipeDirection

- Refactored the great Throws tests currently in place to allow testing pipes with all in/out direction combinations for both Named pipes as well as Anonymous pipes
- Replaced all relevant hardcoded pipe names with unique values
- Reorganized the tests to use more inheritance and share more code for easy extensibility
- All test classes are now in the same namespace: System.IO.Pipes.Tests
- Expanded tests for InOut NamedPipes and made all existing In/Out tests also test all valid InOut combinations.
- Split the Named and Anonymous tests into two separate folders: NamedPipeTest and AnonymousPipeTest
- Increased code coverage across all test classes


src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+        {
+            using (var pair = CreateNamedPipePair())
+            {
+                var pipe = pair.serverStream;
+                pipe.Dispose();
+                byte[] buffer = new byte[] { 0, 0, 0, 0 };
+
+                Assert.Throws<ObjectDisposedException>(() => pipe.Disconnect());
+                Assert.Throws<ObjectDisposedException>(() => pipe.WaitForConnection());
+                Assert.ThrowsAsync<ObjectDisposedException>(async () => { await pipe.WaitForConnectionAsync(); });
```
Maxwe11: This should be

    await Assert.ThrowsAsync<ObjectDisposedException>(() => pipe.WaitForConnectionAsync());

`ThrowsAsync` doesn't unwrap Task<Task> but return Task which could be awaited.

src/System.IO.Pipes/tests/PipeTest.Read.cs
```diff
@@ -0,0 +1,256 @@
...
+                var pipe = pair.readablePipe;
+                byte[] buffer11 = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+                Assert.True(pipe.IsConnected);
+                Assert.True(pipe.CanRead);
+
+                var ctx = new CancellationTokenSource();
+                Task readToken = pipe.ReadAsync(buffer11, 0, buffer11.Length, ctx.Token);
+                ctx.Cancel();
+                Assert.ThrowsAsync<TimeoutException>(() => readToken);
+                Assert.ThrowsAsync<TimeoutException>(() => pipe.ReadAsync(buffer11, 0, buffer11.Length, ctx.Token));
```
Maxwe11: We should `await` here.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
ianhays: Thanks, will fix! I'll search and make sure I don't do this elsewhere.

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.Write.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.IO.Pipes.Tests
+{
+    public class AnonymousPipeTest_Write_ServerIn_ClientOut : PipeTest_Read
```
ianhays: Should inherit from PipeTest_Write

ianhays:
cc: @stephentoub  @bpschoch

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.cs
```diff
@@ -108,6 +108,9 @@ public NamedPipeServerStream(String pipeName, PipeDirection direction, int maxNu
             {
                 throw new ArgumentOutOfRangeException("maxNumberOfServerInstances", SR.ArgumentOutOfRange_MaxNumServerInstances);
             }
+            // inheritability will always be None since this private constructor is only called from other constructors from which
```
stephentoub: Interesting.  @ericstj, was the removal of this ctor from the public surface area for .NET Core by design?

If so, we should remove the associated code from the implementation.

If not, we should look at putting the ctor back into the contract.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+            using (var pair = CreateNamedPipePair())
+            {
+                var server = pair.serverStream;
+                var ctx = new CancellationTokenSource();
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) // [ActiveIssue(812, PlatformID.AnyUnix)] - cancellation token after the operation has been initiated
+                {
+                    Task serverWaitTimeout = server.WaitForConnectionAsync(ctx.Token);
+                    ctx.Cancel();
+                    await Assert.ThrowsAnyAsync<OperationCanceledException>(() => serverWaitTimeout);
```
Maxwe11: :question:  `Assert.ThrowsAsync`
I understand that was taken from original test but `Assert.ThrowsAsync` should be used here, at least on Windows `Assert.ThrowsAsync` should pass.

Maxwe11: I suspect `Assert.ThrowsAnyAsync` was accidently used in original test.

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.CreateClient.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        [Fact]
+        public static void NullParameters_Throws_ArgumentNullException()
+        {
+            Assert.Throws<ArgumentNullException>("pipeHandleAsString", () => new AnonymousPipeClientStream((string)null));
+            Assert.Throws<ArgumentNullException>("pipeHandleAsString", () => new AnonymousPipeClientStream(PipeDirection.Out, (string)null));
+            Assert.Throws<ArgumentNullException>("safePipeHandle", () => new AnonymousPipeClientStream(PipeDirection.In, (SafePipeHandle)null));
+        }
+
+        [Fact]
```
stephentoub: Theory? e.g.
```C#
[Theory]
[InlineData("abc")]
[InlineData("-1")]
public static void InvalidStringParameters_Throws_ArgumentException(string handle)
{
    Assert.Throws<ArgumentException>("pipeHandleAsString", () => new AnonymousPipeClientStream(handle));
    Assert.Throws<ArgumentException>("pipeHandleAsString", () => new AnonymousPipeClientStream(PipeDirection.Out, handle));
}
```

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
ianhays: Good catch; several of the original tests were using ThrowsAnyAsync and in one location were using it to catch the derived Exception TaskCanceledException.

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.CreateClient.cs
ianhays: Yup!

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.CreateServer.cs
```diff
@@ -0,0 +1,75 @@
...
+        [InlineData(PipeDirection.In, -500)]
+        [InlineData(PipeDirection.Out, -500)]
+        [InlineData(PipeDirection.InOut, -500)] //bufferSize will cause an exception before InOut will
+        public static void InvalidBufferSize_Throws_ArgumentOutOfRangeException(PipeDirection direction, int bufferSize)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("bufferSize", () => new AnonymousPipeServerStream(direction, HandleInheritability.None, bufferSize));
+        }
+
+        [Theory]
+        [InlineData((PipeDirection)123)]
```
stephentoub: Doesn't need to be a theory if there's only one input.

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.Read.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace System.IO.Pipes.Tests
+{
+    public class AnonymousPipeTest_Read_ServerIn_ClientOut : PipeTest_Read
```
stephentoub: What does the "Read" in the name here mean?

stephentoub: Oh, nevermind, I see the pattern you're using.

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.Specific.cs
```diff
@@ -0,0 +1,172 @@
...
+            using (var server = new AnonymousPipeServerStream(PipeDirection.In, HandleInheritability.None, desiredBufferSize))
+            using (var client = new AnonymousPipeClientStream(PipeDirection.Out, server.ClientSafePipeHandle))
+            {
+                Assert.Equal(desiredBufferSize, server.InBufferSize);
+                Assert.Equal(desiredBufferSize, client.OutBufferSize);
+            }
+        }
+
+        [Fact]
+        [PlatformSpecific(~PlatformID.Linux)]
```
stephentoub: Could you add a comment about why this is is not for Linux?

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            Assert.Throws<ArgumentNullException>("pipeName", () => new NamedPipeClientStream(null));
+            Assert.Throws<ArgumentNullException>("pipeName", () => new NamedPipeClientStream(".", null));
+        }
+
+        [Fact]
+        public static void EmptyStringPipeName_Throws_ArgumentException()
+        {
+            Assert.Throws<ArgumentException>(() => new NamedPipeClientStream(""));
+            Assert.Throws<ArgumentException>(() => new NamedPipeClientStream(".", ""));
```
stephentoub: Are there parameter names associated with these exceptions that should be validated?

ianhays: Interestingly enough, there are not parameter names associated with the exceptions for a zero-length string. I confirmed that this behavior is present on Desktop as well with the following test:

```
try {
    var ccli = new NamedPipeClientStream("");
}
catch (ArgumentException e)
{
    Console.WriteLine(e.ParamName.Length); // throws ArgumentNullException - e.ParamName is null
}
```

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.Specific.cs
ianhays: Linux_BufferSizeRoundtrips:
```
// On Linux, setting the buffer size of the server will also set the buffer size of the
// client, regardless of the direction of the flow
```
BufferSizeRoundTripping:
```
// On systems other than Linux, setting the buffer size of the server will only set
// set the buffer size of the client if the flow of the pipe is towards the client i.e.
// the client is defined with PipeDirection.In
```
How's that?

src/System.IO.Pipes/tests/AnonymousPipeTests/AnonymousPipeTest.CreateServer.cs
ianhays: Fixed!

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        [Theory]
+        [InlineData(PipeDirection.In)]
+        [InlineData(PipeDirection.InOut)]
+        [InlineData(PipeDirection.Out)]
+        public static void InvalidPipeOptions_Throws_ArgumentOutOfRangeException(PipeDirection direction)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("options", () => new NamedPipeClientStream(".", "client1", direction, (PipeOptions)255));
+            Assert.Throws<ArgumentOutOfRangeException>("options", () => new NamedPipeClientStream(".", "client1", direction, (PipeOptions)255, Security.Principal.TokenImpersonationLevel.None));
```
stephentoub: I'd suggest adding ```using System.Security.Principal;``` at the top of the file, and then just using ```TokenImpersonationLevel.None``` here.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -0,0 +1,127 @@
...
+            Assert.Throws<ArgumentException>("safePipeHandle", () => new NamedPipeClientStream(direction, false, true, pipeHandle));
+        }
+
+        [Theory]
+        [InlineData(PipeDirection.In)]
+        [InlineData(PipeDirection.InOut)]
+        [InlineData(PipeDirection.Out)]
+        public static void BadHandleKind_Throws_IOException(PipeDirection direction)
+        {
+            using (FileStream fs = new FileStream(Guid.NewGuid().ToString(), FileMode.Create, FileAccess.Write, FileShare.None, 8, FileOptions.DeleteOnClose))
```
stephentoub: Can we use the same temporary file mechanism we've used in the FileSystem and MMF suites?

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                SafeFileHandle safeHandle = fs.SafeFileHandle;
+
+                bool gotRef = false;
+                try
+                {
+
```
stephentoub: Nit: unnecessary blank line

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
ianhays: Good idea!

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateServer.cs
```diff
@@ -0,0 +1,219 @@
...
+        public static void Windows_CreateFromAlreadyBoundHandle_Throws_ArgumentException(PipeDirection direction)
+        {
+            // The pipe is already bound
+            var pipe = new NamedPipeServerStream(GetUniquePipeName(), direction, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous);
+            Assert.Throws<ArgumentException>(() => new NamedPipeServerStream(direction, true, true, pipe.SafePipeHandle));
+            pipe.Dispose();
+        }
+
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)] // NumberOfServerInstances > 1 isn't supported and has undefined behavior on Unix
```
stephentoub: We don't throw for NumberOfServerInstances > 1 on Unix?

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+        [InlineData(PipeOptions.None, PipeOptions.Asynchronous, false, true)]
+        [InlineData(PipeOptions.None, PipeOptions.Asynchronous, true, true)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.None, false, false)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.None, true, false)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.None, false, true)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.None, true, true)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.Asynchronous, false, false)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.Asynchronous, true, false)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.Asynchronous, false, true)]
+        [InlineData(PipeOptions.Asynchronous, PipeOptions.Asynchronous, true, true)]
```
stephentoub: This looks like it's every combination of each of the four binary values?  If so, you might consider making that more explicit via MemberData instead of InlineData, e.g.
```C#
public static IEnumerable<object[]> OneWayReadWritesMemberData()
{
    var options = new [] { PipeOptions.None, PipeOptions.Asynchronous };
    var bools = new [] { false, true };
    foreach (PipeOption serverOption in options)
    foreach (PipeOption clientOption in options)
    foreach (bool asyncServerOps in bools)
    foreach (bool asyncClientOps in bools)
        yield return new object[] { serverOption, clientOption, asyncServerOps, asyncClientOps };
}
```

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+                byte[] received = new byte[] { 0 };
+                Task clientTask = Task.Run(async () =>
+                {
+                    if (asyncClientOps)
+                    {
+                        await client.ConnectAsync();
+                        if (pair.writeToServer)
+                        {
+                            received = await ReadBytesAsync(client, sendBytes.Length);
+
```
stephentoub: Nit: unnecessary blank line

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+                        }
+                    }
+                });
+                if (asyncServerOps)
+                {
+                    await server.WaitForConnectionAsync();
+                    if (pair.writeToServer)
+                    {
+                        await WriteBytesAsync(server, sendBytes);
+
```
stephentoub: Nit: unnecessary blank line

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateServer.cs
ianhays: Nope. Oddly enough, we do the same validity checks on it that we do for Windows: check that it is between 1 and 255.

ianhays: I'm going to change the tests in NamedPipeTest.CreateServer to test some different MaxServerInstances. They currently only test "0".

stephentoub: I'm not sure if that's "odd" or a bug... we should think through whether that's correct or whether we should throw a PlatformNotSupportedException.  I don't remember if I did that on purpose or if it was an oversight on my part.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
ianhays: This is much better, thank you.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateServer.cs
ianhays: I'll open an issue for further discussion: #2764

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
stephentoub: Not pushing one way or the other, but why do we care whether it's an OperationCanceledException or the more derived TaskCanceledException?

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -0,0 +1,573 @@
...
+                Assert.Throws<InvalidOperationException>(() => client.IsMessageComplete);
+            }
+        }
+
+        [Fact]
+        public void OperationsOnUnconnectedServer()
+        {
+            using (var pair = CreateNamedPipePair())
+            {
+                var server = pair.serverStream;
```
stephentoub: There's a fair amount of ```var``` being used when the type isn't clear from the right-hand side... can we change those to use the actual type name?

ianhays: Sure thing. If the right side doesn't explicitly state the type inline, I will change var to the type.

stephentoub:
Thanks, Ian.  I skimmed through and left a few comments.  Overall, looks better, thanks.  

Do you know why line coverage drops with this change? Looks like it was 75.9% before and with this change is 74.6%.

And regarding coverage, it looks like we're instantiating a StreamAsyncHelper, but then none of its members are being exercised.  Do you know why? Are you planning to do a subsequent pass through to look at test gaps identified via code coverage?

Maxwe11:
>And regarding coverage, it looks like we're instantiating a StreamAsyncHelper, but then none of its members are being exercised. Do you know why?

StreamAsyncHelper used in "dead" branches inside BeginRead/BeginWrite. It will be eliminated with cleanup #2414 :smile:

ianhays:
I'm actually finishing up an update right now that brings coverage up and above its previous levels :) Some lines were being tested in unexpected places and my refactoring removed those scenarios from being hit, so I'm re-adding tests specifically for those lines.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
ianhays: We can if you'd prefer; it didn't seem worthwhile to make all tests inherit from FileCleanupTestBase and create all of those TestDirectories since we only create a file in two places across all Pipes tests.

What I instead change the file name that is being created to be similar to as if it <i>were</i> inheriting from FileCleanupTestBase? Something like:
```
using (FileStream fs = new FileStream(Path.Combine(Path.GetTempPath(), "_BadHandleKind_Throws_IOException_" + Path.GetRandomFileName()), FileMode.Create, FileAccess.Write, FileShare.None, 8, FileOptions.DeleteOnClose))
```

stephentoub: Sold.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
ianhays: To me it seems that in general we want to test as precisely as possible regardless of how important the distinctions may be just so that if we decide to change the behavior later, the tests will fail and let us know that we're breaking something, somewhere. For example, in the FS tests we wanted to always catch FileNotFoundExceptions or DirectoryNotFoundExceptions instead of just catching an IOException for the same reason. It's not a perfect example since those two exceptions were functionally different (although loosely so), but I think the spirit of the argument is valid.

I'm with you though Steve that we probably don't care either way; it's a very edge case that won't be encountered frequently. However, I went ahead and changed all ThrowsAnyAsync instances to ThrowsAsync for consistency's sake (since we generally don't use ThrowsAnyAsync).

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
```diff
@@ -0,0 +1,318 @@
...
+            {
+                var ctx = new CancellationTokenSource();
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) // [ActiveIssue(812, PlatformID.AnyUnix)] - Unix implementation currently ignores timeout and cancellation token once the operation has been initiated
+                {
+                    Task clientConnectToken = client.ConnectAsync(ctx.Token);
+                    ctx.Cancel();
+                    await Assert.ThrowsAsync<TaskCanceledException>(() => clientConnectToken);
+                }
+                ctx.Cancel();
+                await Assert.ThrowsAsync<TaskCanceledException>(() => client.ConnectAsync(ctx.Token));
```
stephentoub: This sporadically fails for me now that it's been changed away from ThrowsAnyAsync:
```
System.IO.Pipes.Tests.NamedPipeTest_Specific.CancelConnectToNonExistentServer_Throws_TaskCanceledException [FAIL]
        Assert.Throws() Failure
        Expected: typeof(System.Threading.Tasks.TaskCanceledException)
        Actual:   typeof(System.OperationCanceledException): The operation was canceled.
        Stack Trace:
              at System.Threading.CancellationToken.ThrowOperationCanceledException()
              at System.Threading.CancellationToken.ThrowIfCancellationRequested()
           \corefx\src\System.IO.Pipes\src\System\IO\Pipes\NamedPipeClientStream.cs(152,0): at System.IO.Pipes.NamedPipeClientStream
  .ConnectInternal(Int32 timeout, CancellationToken cancellationToken, Int32 startTime)
           \corefx\src\System.IO.Pipes\src\System\IO\Pipes\NamedPipeClientStream.cs(210,0): at System.IO.Pipes.NamedPipeClientStream
  .<>c__DisplayClass20_0.<ConnectAsync>b__0()
              at System.Threading.Tasks.Task.InnerInvoke()
              at System.Threading.Tasks.Task.Execute()
           --- End of stack trace from previous location where exception was thrown ---
              at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
              at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
              at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
              at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
              at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
```

stephentoub:
Other than the failure (which I'm able to repro fairly frequently, especially when doing code coverage), LGTM.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
ianhays: Huh, interesting... I did not see that error in my ~100 runs. Back to ThrowsAnyAsync!

PR #2734, opened 2015-08-11
Increasing timeouts on System.Threading.Timer tests to prevent flakiness

### LINKED ISSUE #2717: Four failures in new System.Threading.Timer tests in CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release/946/console
```
TimerFiringTests.Timer_CanDisposeSelfInCallback [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Threading.Timer\tests\TimerFiringTests.cs(126,0): at TimerFiringTests.Timer_CanDisposeSelfInCallback()
     TimerFiringTests.NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain [FAIL]
        Should have received a second timer event after changing it
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Threading.Timer\tests\TimerFiringTests.cs(150,0): at TimerFiringTests.NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain()
     TimerFiringTests.Running_Timer_CanBeFinalizedAndStopsFiring [FAIL]
        Failed to get first timer fire
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Threading.Timer\tests\TimerFiringTests.cs(160,0): at TimerFiringTests.Running_Timer_CanBeFinalizedAndStopsFiring()
     TimerFiringTests.MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer [FAIL]
        Blocking callback prevented a second timer from firing
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Threading.Timer\tests\TimerFiringTests.cs(184,0): at TimerFiringTests.MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
  
  Unhandled Exception: Xunit.Sdk.TrueException: Should have received a second callback while blocked
  Expected: True
  Actual:   False
     at Xunit.Assert.True(Nullable`1 condition, String userMessage)
     at TimerFiringTests.<>c__DisplayClass22.<MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer>b__20(Object o)
     at System.Threading.TimerQueueTimer.CallCallbackInContext(Object state)
     at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
     at System.Threading.TimerQueueTimer.CallCallback()
     at System.Threading.TimerQueueTimer.Fire()
     at System.Threading.TimerQueue.FireNextTimers()
     at System.Threading.TimerQueue.AppDomainTimerCallback()
  Finished:    System.Threading.Timer.Tests
```

sokket:
@stephentoub This passed on my local machines multiple times but I'll do a couple lab runs on here to see if it fixes it during CI runs

This PR fixes #2717

### DIFF
--- a/src/System.Threading.Timer/tests/TimerFiringTests.cs
+++ b/src/System.Threading.Timer/tests/TimerFiringTests.cs
@@ -8,6 +8,8 @@ using Xunit;
 
 public class TimerFiringTests
 {
+    private const int MaxPositiveTimeoutInMs = 30000;
+
     [Fact]
     public void Timer_Fires_After_DueTime_Ellapses()
     {
@@ -18,7 +20,7 @@ public class TimerFiringTests
             are.Set();
         }), null, TimeSpan.FromMilliseconds(250), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */))
         {
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(750)));
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(MaxPositiveTimeoutInMs)));
         }
     }
 
@@ -56,7 +58,7 @@ public class TimerFiringTests
             }
         }), null, TimeSpan.FromMilliseconds(250), TimeSpan.FromMilliseconds(50)))
         {
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(2000)));
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(MaxPositiveTimeoutInMs)));
         }
     }
 
@@ -75,46 +77,13 @@ public class TimerFiringTests
             }
         }), null, TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(Timeout.Infinite) /* not relevant */))
         {
-            Assert.False(are.WaitOne(TimeSpan.FromSeconds(1)));
-        }
-    }
-
-    [Fact]
-    public void MultipleTimers_ShouldFire_InCorrectOrder()
-    {
-        object t1State = new object();
-        object t2State = new object();
-        Timer t1 = null;
-        Timer t2 = null;
-        int count = 0;
-        AutoResetEvent are = new AutoResetEvent(false);
-        TimerCallback tc = new TimerCallback((object o) =>
-        {
-            if (count == 0)
-            {
-                Assert.Same(t2State, o);
-                count++;
-            }
-            else
-            {
-                Assert.Same(t1State, o);
-                are.Set();
-            }
-        });
-
-        using (t1 = new Timer(tc, t1State, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1)))
-        using (t2 = new Timer(tc, t2State, TimeSpan.FromMilliseconds(0), TimeSpan.FromMilliseconds(-1)))
-        {
-            // Wait for both events to fire
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(1500 /*1.5 seconds*/)));
+            Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(250 /*enough for 2 fires + buffer*/)));
         }
     }
 
     [Fact]
     public void Timer_CanDisposeSelfInCallback()
     {
-        // There's nothing to validate besides that we don't throw an exception, so rely on xunit 
-        // to catch any exception that would be thrown and signal a test failure
         Timer t = null;
         AutoResetEvent are = new AutoResetEvent(false);
         TimerCallback tc = new TimerCallback((object o) =>
@@ -123,7 +92,7 @@ public class TimerFiringTests
             are.Set();
         });
         t = new Timer(tc, null, 1, -1);
-        Assert.True(are.WaitOne(50));
+        Assert.True(are.WaitOne(MaxPositiveTimeoutInMs));
     }
 
     [Fact]
@@ -144,10 +113,10 @@ public class TimerFiringTests
         TimerCallback tc = new TimerCallback((object o) => are.Set());
         using (var t = new Timer(tc, null, 1, Timeout.Infinite))
         {
-            Assert.True(are.WaitOne(50), "Should have received first timer event");
-            Assert.False(are.WaitOne(20), "Should not have received a second timer event");
+            Assert.True(are.WaitOne(MaxPositiveTimeoutInMs), "Should have received first timer event");
+            Assert.False(are.WaitOne(500), "Should not have received a second timer event");
             t.Change(10, Timeout.Infinite);
-            Assert.True(are.WaitOne(50), "Should have received a second timer event after changing it");
+            Assert.True(are.WaitOne(MaxPositiveTimeoutInMs), "Should have received a second timer event after changing it");
         }
     }
 
@@ -156,33 +125,32 @@ public class TimerFiringTests
     {
         AutoResetEvent are = new AutoResetEvent(false);
         TimerCallback tc = new TimerCallback((object o) => are.Set());
-        var t = new Timer(tc, null, 100, 500);
-        Assert.True(are.WaitOne(250), "Failed to get first timer fire");
+        var t = new Timer(tc, null, 1, 500);
+        Assert.True(are.WaitOne(MaxPositiveTimeoutInMs), "Failed to get first timer fire");
         t = null; // Remove our refence so the timer can be GC'd 
         GC.Collect();
         GC.WaitForPendingFinalizers();
         GC.Collect();
-        Assert.False(are.WaitOne(750), "Should not have received a timer fire after it was collected");
+        Assert.False(are.WaitOne(500), "Should not have received a timer fire after it was collected");
     }
 
     [Fact]
-    public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
+    public void MultpleTimers_PeriodicTimerIsntBlockedByBlockedCallback()
     {
-        AutoResetEvent are1 = new AutoResetEvent(false);
-        AutoResetEvent are2 = new AutoResetEvent(false);
-        TimerCallback tc = new TimerCallback((object o) =>
+        int callbacks = 2;
+        Barrier b = new Barrier(callbacks + 1);
+        Timer t = null;
+        t = new Timer(_ =>
         {
-            using (var t1 = new Timer((object obj) => are1.Set(), null, 1, -1))
+            if (Interlocked.Decrement(ref callbacks) >= 0)
             {
-                Assert.True(are1.WaitOne(250), "Should have received a second callback while blocked");
-                are2.Set();
+                Assert.True(b.SignalAndWait(MaxPositiveTimeoutInMs));
             }
-        });
+            t.Dispose();
+        }, null, -1, -1);
+        t.Change(1, 50);
 
-        using (var t1 = new Timer(tc, null, 1, -1))
-        {
-            Assert.True(are2.WaitOne(500), "Blocking callback prevented a second timer from firing");
-        }
+        Assert.True(b.SignalAndWait(MaxPositiveTimeoutInMs));
     }
 
     [Fact]
@@ -193,7 +161,7 @@ public class TimerFiringTests
         Timer[] timers = System.Linq.Enumerable.Range(0, maxTimers).Select(_ => new Timer(s => ce.Signal(), null, 100 /* enough time to wait on the are */, -1)).ToArray();
         try
         {
-            Assert.True(ce.Wait(500), String.Format("Not all timers fired, {0} left of {1}", ce.CurrentCount, maxTimers));
+            Assert.True(ce.Wait(MaxPositiveTimeoutInMs), String.Format("Not all timers fired, {0} left of {1}", ce.CurrentCount, maxTimers));
         }
         finally
         {


src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -122,8 +120,8 @@ public void Timer_CanDisposeSelfInCallback()
             t.Dispose();
             are.Set();
         });
-        t = new Timer(tc, null, 1, -1);
-        Assert.True(are.WaitOne(50));
+        t = new Timer(tc, null, 10, -1);
```
stephentoub: Changing this from 1 to 10 could actually make things worse.  I'd leave it at 1.  If this test is going to fail, it's because the callback didn't complete within the time that we're waiting on the event... but making this dueTime larger actually delays when the timer invokes the callback. (In this case, the change is probably a nop due to rounding of timeout values.)

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -122,8 +120,8 @@ public void Timer_CanDisposeSelfInCallback()
...
+        Assert.True(are.WaitOne(250));
```
stephentoub: This could be much larger, like many seconds.  This is a case where we're waiting for something to happen, and that thing must happen or the test should fail.  The only reason we want a timeout here at all is so that the test doesn't hang indefinitely if there is a bug.  This is the same case as when we've used 30 seconds in some of the other test suites, like FSW.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -142,12 +140,12 @@ public void NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain()
     {
         AutoResetEvent are = new AutoResetEvent(false);
         TimerCallback tc = new TimerCallback((object o) => are.Set());
-        using (var t = new Timer(tc, null, 1, Timeout.Infinite))
+        using (var t = new Timer(tc, null, 10, Timeout.Infinite))
```
stephentoub: Another case where we should leave it as 1.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -142,12 +140,12 @@ public void NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain()
...
         {
-            Assert.True(are.WaitOne(50), "Should have received first timer event");
-            Assert.False(are.WaitOne(20), "Should not have received a second timer event");
+            Assert.True(are.WaitOne(250), "Should have received first timer event");
```
stephentoub: Can be very large.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -142,12 +140,12 @@ public void NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain()
...
+            Assert.False(are.WaitOne(100), "Should not have received a second timer event");
             t.Change(10, Timeout.Infinite);
-            Assert.True(are.WaitOne(50), "Should have received a second timer event after changing it");
+            Assert.True(are.WaitOne(250), "Should have received a second timer event after changing it");
```
stephentoub: Can be very large.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -142,12 +140,12 @@ public void NonRepeatingTimer_ThatHasAlreadyFired_CanChangeAndFireAgain()
     {
         AutoResetEvent are = new AutoResetEvent(false);
         TimerCallback tc = new TimerCallback((object o) => are.Set());
-        using (var t = new Timer(tc, null, 1, Timeout.Infinite))
+        using (var t = new Timer(tc, null, 10, Timeout.Infinite))
         {
-            Assert.True(are.WaitOne(50), "Should have received first timer event");
-            Assert.False(are.WaitOne(20), "Should not have received a second timer event");
+            Assert.True(are.WaitOne(250), "Should have received first timer event");
+            Assert.False(are.WaitOne(100), "Should not have received a second timer event");
```
stephentoub: These Assert.False(are.WaitOne(...)) cases are the ones we need to be more stringent with, because we're waiting for something that we expect to never happen, so in the success case it's going to take the full amount of time we specify.  100ms may still be too small though for all of the fluctations possible in the CI environment.  I'd suggest going to at least 250.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -156,7 +154,7 @@ public void Running_Timer_CanBeFinalizedAndStopsFiring()
     {
         AutoResetEvent are = new AutoResetEvent(false);
         TimerCallback tc = new TimerCallback((object o) => are.Set());
-        var t = new Timer(tc, null, 100, 500);
+        var t = new Timer(tc, null, 10, 500);
```
stephentoub: Can make this 1 instead of 10, though as noted above I expect they'll actually behave the same.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -172,16 +170,16 @@ public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
...
+            using (var t1 = new Timer((object obj) => are1.Set(), null, 10, -1))
             {
-                Assert.True(are1.WaitOne(250), "Should have received a second callback while blocked");
+                Assert.True(are1.WaitOne(500), "Should have received a second callback while blocked");
                 are2.Set();
             }
         });
 
-        using (var t1 = new Timer(tc, null, 1, -1))
+        using (var t1 = new Timer(tc, null, 10, -1))
```
stephentoub: Can remain 1.

sokket:
Put all timeouts at 10 to 1 and increased positive assert timeouts to at least 1 second

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -174,14 +172,14 @@ public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
...
-                Assert.True(are1.WaitOne(250), "Should have received a second callback while blocked");
+                Assert.True(are1.WaitOne(1000), "Should have received a second callback while blocked");
                 are2.Set();
             }
         });
 
         using (var t1 = new Timer(tc, null, 1, -1))
         {
-            Assert.True(are2.WaitOne(500), "Blocking callback prevented a second timer from firing");
+            Assert.True(are2.WaitOne(2000), "Blocking callback prevented a second timer from firing");
```
stephentoub: I just realized this test's name doesn't match what it's testing.  I'm also still a bit concerned about potential flakiness.  Here's a suggestion for how to rewrite it to hopefully be more reliable (pseudo-code):
```C#
[Fact]
public void MultipleTimers_PeriodTimerIsntBlocked()
{
    const int FailureTimeout = 10 * 1000;
    var mres = new ManualResetEventSlim();
    int callbacks = 2;
    var b = new Barrier(callbacks + 1);

    Timer t = null;
    t = new Timer(_ =>
    {
        if (Interlocked.Decrement(ref callbacks) >= 0)
        {
            Assert.True(b.SignalAndWait(FailureTimeout));
        }
        t.Dispose();
    }, null, 1, 50);
    Assert.True(b.SignalAndWait(FailureTimeout));
}
```

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -93,6 +93,7 @@ public void MultipleTimers_ShouldFire_InCorrectOrder()
             if (count == 0)
             {
                 Assert.Same(t2State, o);
+                Interlocked.Increment(ref count);
```
stephentoub: Why do you have both the interlocked and regular increment?

stephentoub: And thinking about it more, you might just want to delete this whole test.  Technically there's nothing guaranteeing that t2 will fire before t1.

stephentoub: Additionally, GitHub won't let me comment on them because they weren't changed as part of this PR, but I just noticed that the PassesStateThroughCallback tests are likely not testing what you want them to be testing, because you're constructing the timer and then immediately disposing it... there's a good chance it'll be disposed before the callback gets invoked, at which point we'll never run the asserts in the callback.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -106,15 +107,13 @@ public void MultipleTimers_ShouldFire_InCorrectOrder()
         using (t2 = new Timer(tc, t2State, TimeSpan.FromMilliseconds(0), TimeSpan.FromMilliseconds(-1)))
         {
             // Wait for both events to fire
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(1500 /*1.5 seconds*/)));
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(3000 /*3 seconds*/)));
```
stephentoub: For the cases of ```Assert.True(are.WaitOne(...))``` where we expect the event to be set and the only reason we have a timeout at all is to avoid hanging the tests for a prolonged period of time (Jenkins will wait a few hours to kill a hung test), I'd suggest having a constant ExpectedCallbackWaitTimeout, or something like that, which is used across all of the tests, and which is set to 30 seconds or something obscenely long.

sokket:
Looks like a FSW failure not a timer one

@dotnet-bot test this please

sokket:
@dotnet-bot test this please

stephentoub:
> Looks like a FSW failure not a timer one

Yeah, it's the FileSystemWatcher_Changed_PreSeededNestedStructure test you added and that's been failing intermittently.  I'd tried to fix it, but apparently I only fixed some of the problems with it.  I'm not sure what's going on with at the moment... needs investigation (#2740).  In the meantime I'm going to disable it until you or I can investigate.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -75,46 +77,13 @@ public void Timer_FiresOnlyOnce_OnDueTime_With_InfinitePeriod()
...
-                are.Set();
-            }
-        });
-
-        using (t1 = new Timer(tc, t1State, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1)))
-        using (t2 = new Timer(tc, t2State, TimeSpan.FromMilliseconds(0), TimeSpan.FromMilliseconds(-1)))
-        {
-            // Wait for both events to fire
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(1500 /*1.5 seconds*/)));
+            Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(MaxPositiveTimeoutInMs)));
```
stephentoub: I was suggesting MaxPositiveTimeoutInMs be used in places of the form ```Assert.True(are.WaitOne(...))``` because in those cases we're expecting an event and the only reason we have a timeout at all is to avoid hanging the tests in the case of failure.  But this is ```Assert.False(are.WaitOne(...))``` because we're expecting that we *won't* get an event, which means that in a successful run we'll end up waiting for the entire timeout. Such cases should not use a long value, but should use the shortest value that's reasonable.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -156,33 +125,32 @@ public void Running_Timer_CanBeFinalizedAndStopsFiring()
...
     }
 
     [Fact]
-    public void MultpleTimers_PeriodicTimerIsntBlockedByABlockedTimer()
+    public void MultpleTimers_PeriodicTimerIsntBlockedByBlockedCallback()
     {
-        AutoResetEvent are1 = new AutoResetEvent(false);
-        AutoResetEvent are2 = new AutoResetEvent(false);
-        TimerCallback tc = new TimerCallback((object o) =>
+        ManualResetEventSlim mres = new ManualResetEventSlim();
```
stephentoub: I realize this was in my pseudo-code example, but it's not necessary (I think I was originally intending to use it to signal the calling thread, but I ended up using the same Barrier for that); you can delete this line, as the mres is never used.

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -156,33 +125,32 @@ public void Running_Timer_CanBeFinalizedAndStopsFiring()
...
-            using (var t1 = new Timer((object obj) => are1.Set(), null, 1, -1))
+            if (Interlocked.Decrement(ref callbacks) >= 0)
             {
-                Assert.True(are1.WaitOne(250), "Should have received a second callback while blocked");
-                are2.Set();
+                Assert.True(b.SignalAndWait(MaxPositiveTimeoutInMs));
             }
-        });
+            t.Dispose();
+        }, null, 1, 50);
```
stephentoub: In my code, I had:
```C#
}, null, -1, -1);
t.Change(1, 50);
```
(though I think I wrote it like the above initially, so maybe you got an email or something that contained the old version and based this on that).  This is subtly different from what you have here:
```C#
}, null, 1, 50);
```
but in an important way.  The version that just sets the times in the ctor is subject to a race condition, where the callback could execute in another thread before the calling thread is able to assign the constructed Timer instance to ```t```, such that the callback then tries to execute ```t.Dispose()``` and gets a NullReferenceException.  By instead passing -1,-1 to the constructor and then subsequently calling t.Change(1, 50), we guarantee that ```t``` has been assigned by the time the callback is executed.

stephentoub:
Left a few more comments to be addressed.  Otherwise LGTM.  Thanks.

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Ah my mistake, I put that there accidentally. Good catch :)

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Will do

src/System.Threading.Timer/tests/TimerFiringTests.cs
sokket: Ah really good catch, thanks! Made the fix

src/System.Threading.Timer/tests/TimerFiringTests.cs
```diff
@@ -75,46 +77,13 @@ public void Timer_FiresOnlyOnce_OnDueTime_With_InfinitePeriod()
...
-                are.Set();
-            }
-        });
-
-        using (t1 = new Timer(tc, t1State, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1)))
-        using (t2 = new Timer(tc, t2State, TimeSpan.FromMilliseconds(0), TimeSpan.FromMilliseconds(-1)))
-        {
-            // Wait for both events to fire
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(1500 /*1.5 seconds*/)));
+            Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(250 /*enough for 2 fires + buffer*/)));
```
stephentoub: I think this is going to be cutting it very close, and we'll likely see some failures.  But in the name of getting these other fixes in to stop the rash of issues we're hitting in CI, we can wait until/if we see such failures and then address it.

stephentoub:
Thanks.

PR #2744, opened 2015-08-12
Adding terminfo tests to the System.Console test area

sokket:
@stephentoub

### DIFF
--- a/src/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/System.Console/src/System/ConsolePal.Unix.cs
@@ -18,6 +18,10 @@ namespace System
     // e.g. for reading/writing /dev/stdin, /dev/stdout, and /dev/stderr,
     // for getting environment variables for accessing charset information for encodings, 
     // and terminfo databases / strings for manipulating the terminal.
+    // NOTE: The test class reflects over this class to run the tests due to limitations in
+    //       the test infrastructure that prevent OS-specific builds of test binaries. If you
+    //       change any of the class / struct / function names, parameters, etc then you need
+    //       to also change the test class.
     internal static class ConsolePal
     {
         public static Stream OpenStandardInput()
@@ -251,26 +255,27 @@ namespace System
             /// <summary>The cached instance.</summary>
             public static TerminalColorInfo Instance { get { return _instance.Value; } }
 
+            private TerminalColorInfo(TermInfo.Database db)
+            {
+                ForegroundFormat = db != null ? db.GetString(TermInfo.Database.SetAnsiForegroundIndex) : string.Empty;
+                BackgroundFormat = db != null ? db.GetString(TermInfo.Database.SetAnsiBackgroundIndex) : string.Empty;
+                ResetFormat = db != null ?
+                    db.GetString(TermInfo.Database.OrigPairsIndex) ??
+                    db.GetString(TermInfo.Database.OrigColorsIndex)
+                    : string.Empty;
+
+                int maxColors = db != null ? db.GetNumber(TermInfo.Database.MaxColorsIndex) : 0;
+                MaxColors = // normalize to either the full range of all ANSI colors, just the dark ones, or none
+                    maxColors >= 16 ? 16 :
+                    maxColors >= 8 ? 8 :
+                    0;
+            }
+
             /// <summary>Lazy initialization of the terminal color information.</summary>
             private static Lazy<TerminalColorInfo> _instance = new Lazy<TerminalColorInfo>(() =>
             {
-                TermInfo.Database db = TermInfo.Database.Instance;
-                TerminalColorInfo tci = new TerminalColorInfo();
-                if (db != null)
-                {
-                    tci.ForegroundFormat = db.GetString(TermInfo.Database.SetAnsiForegroundIndex);
-                    tci.BackgroundFormat = db.GetString(TermInfo.Database.SetAnsiBackgroundIndex);
-                    tci.ResetFormat =
-                        db.GetString(TermInfo.Database.OrigPairsIndex) ??
-                        db.GetString(TermInfo.Database.OrigColorsIndex);
-
-                    int maxColors = db.GetNumber(TermInfo.Database.MaxColorsIndex);
-                    tci.MaxColors = // normalize to either the full range of all ANSI colors, just the dark ones, or none
-                        maxColors >= 16 ? 16 :
-                        maxColors >= 8 ? 8 :
-                        0;
-                }
-                return tci;
+                TermInfo.Database db = TermInfo.Database.Instance; // Could be null if TERM is set to a file that doesn't exist
+                return new TerminalColorInfo(db);
             }, isThreadSafe: true);
         }
 
@@ -902,10 +907,16 @@ namespace System
                                     ~value);
                                 break;
 
-                            // Augment first two parameters by 1
+                                // Some terminfo files appear to have a fairly liberal interpretation of %i. The spec states that %i increments the first two arguments, 
+                                // but some uses occur when there's only a single argument. To make sure we accomodate these files, we increment the values 
+                                // of up to (but not requiring) two arguments.
                             case 'i':
-                                args[0] = 1 + args[0].Int32;
-                                args[1] = 1 + args[1].Int32;
+                                if (args.Length > 0)
+                                {
+                                    args[0] = 1 + args[0].Int32;
+                                    if (args.Length > 1)
+                                        args[1] = 1 + args[1].Int32;
+                                }
                                 break;
 
                             // Conditional of the form %? if-part %t then-part %e else-part %;
--- a/src/System.Console/tests/System.Console.Tests.csproj
+++ b/src/System.Console/tests/System.Console.Tests.csproj
@@ -26,6 +26,7 @@
     <Compile Include="Timeout.cs" />
     <Compile Include="ThreadSafety.cs" />
     <Compile Include="XunitAssemblyAttributes.cs" />
+    <Compile Include="TermInfo.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\src\System.Console.csproj">
--- a/src/System.Console/tests/TermInfo.cs
+++ b/src/System.Console/tests/TermInfo.cs
@@ -1,0 +1,263 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+using Xunit;
+
+public class TermInfo
+{
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void VerifyInstalledTermInfosParse()
+    {
+        string[] locations = GetFieldValueOnObject<string[]>("_terminfoLocations", null, typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TermInfo+Database"));
+        foreach (string location in locations)
+        {
+            if (!Directory.Exists(location))
+                continue;
+
+            foreach (string term in Directory.EnumerateFiles(location, "*", SearchOption.AllDirectories))
+            {
+                if (term.ToUpper().Contains("README")) continue;
+                object info = CreateTermColorInfo(ReadTermInfoDatabase(Path.GetFileName(term)));
+
+                if (!string.IsNullOrEmpty(GetForegroundFormat(info)))
+                {
+                    Assert.NotEmpty(EvaluateParameterizedStrings(GetForegroundFormat(info), 0 /* irrelevant, just an integer to put into the formatting*/));
+                }
+
+                if (!string.IsNullOrEmpty(GetBackgroundFormat(info)))
+                {
+                    Assert.NotEmpty(EvaluateParameterizedStrings(GetBackgroundFormat(info), 0 /* irrelevant, just an integer to put into the formatting*/));
+                }
+            }
+        }
+    }
+
+    // Note that we cannot use a Theory here due using unprintable characters and a bug in xunit where the runner errors
+    // out while trying to generate a report and printing the character (bug #380 on xunit)
+    private void TermInfoVerification(string termToTest, string expectedForeground, string expectedBackground, int colorValue)
+    {
+        object info = CreateTermColorInfo(ReadTermInfoDatabase(termToTest));
+        Assert.Equal(expectedForeground, EvaluateParameterizedStrings(GetForegroundFormat(info), colorValue));
+        Assert.Equal(expectedBackground, EvaluateParameterizedStrings(GetBackgroundFormat(info), colorValue));
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Xterm256ColorDefault()
+    {
+        TermInfoVerification("xterm-256color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Xterm256ColorLight()
+    {
+        TermInfoVerification("xterm-256color", "\u001B\u005B\u00331m", "\u001B\u005B\u00341m", 1);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Xterm256ColorDark()
+    {
+        TermInfoVerification("xterm-256color", "\u001B\u005B90m", "\u001B\u005B100m", 8);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void ScreenTermDefault()
+    {
+        TermInfoVerification("screen", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void ScreenTermLight()
+    {
+        TermInfoVerification("screen", "\u001B\u005B\u00332m", "\u001B\u005B\u00342m", 2);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void ScreenTermDark()
+    {
+        TermInfoVerification("screen", "\u001B\u005B\u00339m", "\u001B\u005B\u00349m", 9);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void EtermDefault()
+    {
+        TermInfoVerification("Eterm", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void EtermLight()
+    {
+        TermInfoVerification("Eterm", "\u001B\u005B\u00333m", "\u001B\u005B\u00343m", 3);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void EtermDark()
+    {
+        TermInfoVerification("Eterm", "\u001B\u005B\u003310m", "\u001B\u005B\u003410m", 10);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Wsvt25TermDefault()
+    {
+        TermInfoVerification("wsvt25", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Wsvt25TermLight()
+    {
+        TermInfoVerification("wsvt25", "\u001B\u005B\u00334m", "\u001B\u005B\u00344m", 4);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void Wsvt25TermDark()
+    {
+        TermInfoVerification("wsvt25", "\u001B\u005B\u003311m", "\u001B\u005B\u003411m", 11);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void MachColorTermDefault()
+    {
+        TermInfoVerification("mach-color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void MachColorTermLight()
+    {
+        TermInfoVerification("mach-color", "\u001B\u005B\u00335m", "\u001B\u005B\u00345m", 5);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void MachColorTermDark()
+    {
+        TermInfoVerification("mach-color", "\u001B\u005B\u003312m", "\u001B\u005B\u003412m", 12);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.OSX)]
+    public void EmuTermInfoDoesntBreakParser()
+    {
+        // This file (available by default on OS X) is called out specifically since it contains a format where it has %i
+        // but only one variable instead of two. Make sure we don't break in this case
+        TermInfoVerification("emu", "\u001Br1", "\u001Bs1", 0);
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void TryingToLoadTermThatDoesNotExistDoesNotThrow()
+    {
+        object db = ReadTermInfoDatabase("foobar____");
+        object info = CreateTermColorInfo(db);
+        Assert.Null(db);
+        Assert.NotNull(GetBackgroundFormat(info));
+        Assert.NotNull(GetForegroundFormat(info));
+        Assert.NotNull(GetMaxColors(info));
+        Assert.NotNull(GetResetFormat(info));
+    }
+
+    private object ReadTermInfoDatabase(string term)
+    {
+        MethodInfo readDbMethod = typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TermInfo+Database").GetTypeInfo().GetDeclaredMethods("ReadDatabase").Where(m => m.GetParameters().Count() == 1).Single();
+        return readDbMethod.Invoke(null, new object[] { term });
+    }
+
+    private object CreateTermColorInfo(object db)
+    {
+        return typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TerminalColorInfo").GetTypeInfo().DeclaredConstructors
+                              .Where(c => c.GetParameters().Count() == 1).Single().Invoke(new object[] { db });
+    }
+
+    private string GetForegroundFormat(object colorInfo)
+    {
+        return GetFieldValueOnObject<string>("ForegroundFormat", colorInfo, typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TerminalColorInfo"));
+    }
+
+    private string GetBackgroundFormat(object colorInfo)
+    {
+        return GetFieldValueOnObject<string>("BackgroundFormat", colorInfo, typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TerminalColorInfo"));
+    }
+
+    private int GetMaxColors(object colorInfo)
+    {
+        return GetFieldValueOnObject<int>("MaxColors", colorInfo, typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TerminalColorInfo"));
+    }
+
+    private string GetResetFormat(object colorInfo)
+    {
+        return GetFieldValueOnObject<string>("ResetFormat", colorInfo, typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal+TerminalColorInfo"));
+    }
+
+    private T GetFieldValueOnObject<T>(string name, object instance, Type baseType)
+    {
+        return (T)baseType.GetTypeInfo().GetDeclaredField(name).GetValue(instance);
+    }
+
+    private object CreateFormatParam(object o)
+    {
+        Assert.True((o.GetType() == typeof(Int32)) || (o.GetType() == typeof(string)));
+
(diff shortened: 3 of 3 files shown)


src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -251,25 +254,37 @@ private struct TerminalColorInfo
...
+            }
+
+            /// <summary>Lazy initialization of the terminal color information.</summary>
+            private static Lazy<TerminalColorInfo> _instance = new Lazy<TerminalColorInfo>(() =>
+            {
+                TermInfo.Database db = TermInfo.Database.Instance;
+                TerminalColorInfo tci = new TerminalColorInfo(db);
                 return tci;
             }, isThreadSafe: true);
         }
```
stephentoub: Leave a line note

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -7,6 +7,9 @@
 using System.Text;
 using System.Runtime.InteropServices;
 
+// Let the test classes talk to internals
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("System.Console.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007D1FA57C4AED9F0A32E84AA0FAEFD0DE9E8FD6AEC8F87FB03766C834C99921EB23BE79AD9D5DCC1DD9AD236132102900B723CF980957FC4E177108FC607774F29E8320E92EA05ECE4E821C0A5EFE8F1645C4C0C93C1AB99285D622CAA652C1DFAD63D745D6F2DE5F17E5EAF0FC4963D261C8A12436518206DC093344D5AD293")]
```
stephentoub: I'd suggest putting this in a separate file that's then only included in the Unix build.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -251,25 +254,37 @@ private struct TerminalColorInfo
...
                     int maxColors = db.GetNumber(TermInfo.Database.MaxColorsIndex);
-                    tci.MaxColors = // normalize to either the full range of all ANSI colors, just the dark ones, or none
+                    MaxColors = // normalize to either the full range of all ANSI colors, just the dark ones, or none
                         maxColors >= 16 ? 16 :
                         maxColors >= 8 ? 8 :
                         0;
                 }
+                else
+                {
+                    Debug.Assert(false);
```
stephentoub: Debug.Fail?

stephentoub: That said, we do we need this else at all?  Can we just assert at the beginning of the ctor that db is not null?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -251,25 +254,37 @@ private struct TerminalColorInfo
...
+                    ResetFormat = default(string);
+                    MaxColors = default(int);
+                }
+            }
+
+            /// <summary>Lazy initialization of the terminal color information.</summary>
+            private static Lazy<TerminalColorInfo> _instance = new Lazy<TerminalColorInfo>(() =>
+            {
+                TermInfo.Database db = TermInfo.Database.Instance;
+                TerminalColorInfo tci = new TerminalColorInfo(db);
```
stephentoub: Assuming you change the ctor to assert if db is null, then you'd change this to:
```C#
TermInfo.Database db = TermInfo.Database.Instance;
return db != null ?
    new TerminalColorInfo(db) :
    default(TerminalColorInfo);
```

src/System.Console/tests/Color.cs
```diff
@@ -92,6 +92,68 @@ public static void NonRedirectedOutputDoesUseAnsiSequences()
...
     }
 
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void VerifyInstalledTermInfosParse()
+    {
+        foreach (string location in ConsolePal.TermInfo.Database._terminfoLocations)
+        {
+            // Skip directories that dont exist
+            if (Directory.Exists(location) == false)
```
stephentoub: ```if (!Directory.Exists(location))``` ?

src/System.Console/tests/Color.cs
```diff
@@ -92,6 +92,68 @@ public static void NonRedirectedOutputDoesUseAnsiSequences()
...
+                continue;
+
+            foreach (string term in Directory.EnumerateFiles(location, "*", SearchOption.AllDirectories))
+            {
+                if (term.ToUpper().Contains("README")) continue;
+                ConsolePal.TermInfo.Database db = ConsolePal.TermInfo.Database.ReadDatabase(Path.GetFileName(term));
+                Assert.True(db != null, string.Format("Could not parse terminfo file {0}", term)); // Don't use Assert.NotNull so we can print out the term info file on failure
```
stephentoub: The comment is probably unnecessary, though I agree with the sentiment.  It should be obvious from the string being passed that it contains useful debug info.

src/System.Console/tests/Color.cs
```diff
@@ -92,6 +92,68 @@ public static void NonRedirectedOutputDoesUseAnsiSequences()
...
+            }
+        }
+    }
+
+    [Fact]
```
stephentoub: Would this be better as a [Theory] with multiple [InlineData]s?  You've got the same test four times, just with different inputs.

src/System.Console/tests/Color.cs
```diff
@@ -92,6 +92,68 @@ public static void NonRedirectedOutputDoesUseAnsiSequences()
...
+    {
+        TermInfoVerification("mach-color", "\u001B\u005B\u0033\u0025p1%dm", "\u001B\u005B\u0034\u0025p1%dm");
+    }
+
+    // We need to compare bytes since the strings output to the console are different than the underlying bytes.
+    private void TermInfoVerification(string termToTest, string expectedForeground, string expectedBackground)
+    {
+        ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(ConsolePal.TermInfo.Database.ReadDatabase(termToTest));
+        Assert.True(expectedForeground == info.ForegroundFormat, "Foreground format does not match");
+        Assert.True(expectedBackground == info.BackgroundFormat, "Background format does not match");
```
stephentoub: These can just be Assert.Equal; there's no extra useful info in the message.

src/System.Console/tests/Color.cs
```diff
@@ -92,6 +92,68 @@ public static void NonRedirectedOutputDoesUseAnsiSequences()
...
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void VerifyMachColorTermInfoParsesCorrectly()
+    {
+        TermInfoVerification("mach-color", "\u001B\u005B\u0033\u0025p1%dm", "\u001B\u005B\u0034\u0025p1%dm");
+    }
+
+    // We need to compare bytes since the strings output to the console are different than the underlying bytes.
+    private void TermInfoVerification(string termToTest, string expectedForeground, string expectedBackground)
+    {
+        ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(ConsolePal.TermInfo.Database.ReadDatabase(termToTest));
```
stephentoub: A few things:
1. What if this termToTest doesn't exist on the system on which the test is running?
2. The strings being passed in here are only part of what needs to be tested.  These are the strings that are pulled out of the terminfo file, but they're not what's written out to stdout.  Rather, these strings are actually an interpreted language that we need to evaluate with inputs (in the case of Foreground/BackgroundColor, the numeric values of the selected colors) in order to determine what strings to actually write out to the console: https://github.com/dotnet/corefx/blob/master/src/System.Console/src/System/ConsolePal.Unix.cs#L199-L201 .  We should be doing some level of verification on that evaluation process, at the very least to ensure we don't throw exceptions, but ideally something beyond that.  I suggest that in these tests for known terminfos, you verify the output that gets generated for a variety of color values, and in the overall test that parses all of the terminfo files, you do more basic verification, e.g. if the foreground/background strings exist, that running them through the evaluator doesn't throw and produces a non-empty string.

src/System.Console/tests/Color.cs
stephentoub: As an alternative, you could consider exposing the results of enumerating the files as a MemberData function, e.g. (pseudo-code)
```C#
public static IEnumerable<object[]> Terms()
{
    var results = new List<object[]>();
    foreach (string location in ConsolePal.TermInfo.Database._termInfoLocations)
    foreach (string termPath in Directory.EnumerateFiles(location, "*", SearchOptions.AllDirectories))
    {
        string term = Path.GetFileName(termPath);
        string dirName = Path.GetFileName(Path.GetDirectoryName(termPath));
        if (ter[0] == dirName[0]) // terminfo files are stored in a directory beginning with the same letter
            results.Add(new object[] { term });
    }
    Assert.True(results.Count > 0);
    return results;
}
```
which you can then consume in a theory:
```C#
[Theory]
public void VerifyInstalledTermInfosParse(string term)
{
    ConsolePal.TermInfo.Database db = ConsolePal.TermInfo.Database.ReadDatabase(term);
    Assert.NotNull(db);
    ... // additional verification on db here (see later comment)
}
```

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Fixed

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Fixed

src/System.Console/tests/Color.cs
sokket: Is that the style recommendation? I personally like the == false since it's easy to miss the ! prefix but if the ! prefix is the way to do it, I'll gladly change it

stephentoub: We don't have a style guideline on it as far as I'm aware.  I personally prefer the ```!Directory.Exists(location)``` approach because it reads better to me, i.e. "if not exists" rather than "if exists is false".  Not a big deal.

src/System.Console/tests/Color.cs
sokket: Fixed

src/System.Console/tests/Color.cs
sokket: Good idea

src/System.Console/tests/Color.cs
sokket: Does enumerating the files in a separate function and passing that in via a Theory give us a benefit over enumerating in the function itself? I only went with this direction because no other function needed to enumerate over the files, so it seemed contained. If you think it's best to use a Theory then I'm all for changing it :)

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Good idea

src/System.Console/tests/Color.cs
sokket: Okay, I'll switch to use ! :)

sokket:
@stephentoub FYI I had to remove a Debug.Assert in the constructor of TerminalColorInfo since it would assert during Debug tests when testing that we don't fall over when trying to load a terminfo that doesn't exist. I changed the TerminalColorInfo ctor to make it default to empty strings when trying to load a terminfo that doesn't exist so that if someone's TERM variable is corrupt or out-of-date with the on-disk file, they can still run a console app and not blow up.

src/System.Console/src/System.Console.csproj
```diff
@@ -93,6 +93,7 @@
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetsUnix)' == 'true' ">
     <Compile Include="System\ConsolePal.Unix.cs" />
+    <Compile Include="System\InternalsVisibleTo.Unix.cs" />
```
stephentoub: This is looking for InternalsVisibleTo.Unix.cs, but the file is named InternalsVisibleTo.cs.  I'd leave it as InternalsVisibleTo.cs and update this .csproj line accordingly.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -162,7 +162,7 @@ private static string GetCharset()
         /// <summary>Outputs the format string evaluated and parameterized with the color.</summary>
         /// <param name="foreground">true for foreground; false for background.</param>
         /// <param name="color">The color to store into the field and to use as an argument to the format string.</param>
-        private static void ChangeColor(bool foreground, ConsoleColor color)
+        internal static void ChangeColor(bool foreground, ConsoleColor color)
```
stephentoub: Why did the visibility here need to change?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -902,10 +903,14 @@ public static string Evaluate(string format, params FormatParam[] args)
                                     ~value);
                                 break;
 
-                            // Augment first two parameters by 1
+                            // Some terminfo files make liberal use of %i to mean increment only one argument, not two.
+                            // To make sure we don't break on these files, check if we have one or two variables and increment
+                            // the proper number according to how many are passed.
```
stephentoub: I'd suggest phrasing this a bit differently, just to match a tad better to what we're doing in the revised implementation:
Some terminfo files appear to have a fairly liberal interpretation of %i.  The spec states that %i increments the first two arguments, but some uses occur when there's only a single argument.  To make sure we accomodate these files, we increment the values of up to (but not requiring) two arguments.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -902,10 +903,14 @@ public static string Evaluate(string format, params FormatParam[] args)
...
                             case 'i':
-                                args[0] = 1 + args[0].Int32;
-                                args[1] = 1 + args[1].Int32;
+                                if (args.Length > 0)
+                                    args[0] = 1 + args[0].Int32;
+                                if (args.Length > 1)
+                                    args[1] = 1 + args[1].Int32;
```
stephentoub: I know I scribbled it out this way in email, but we could make it slightly better by doing:
```C#
if (args.Length > 0)
{
    args[0] = 1 + args[0].Int32;
    if (args.Length > 1)
        args[1] = 1 + args[1].Int32;
}

File: src/System.Console/src/System.Console.csproj (hunk 10 above)
Comment (sokket) [inline src/System.Console/src/System.Console.csproj]: Gah I forgot to fix that csproj...thanks

File: src/System.Console/src/System/ConsolePal.Unix.cs (hunk 11 above)
Comment (sokket) [inline src/System.Console/src/System/ConsolePal.Unix.cs]: I was originally using this in a previous version, reverting

File: src/System.Console/tests/System.Console.Tests.csproj [hunk 14]
```diff
@@ -27,6 +27,9 @@
     <Compile Include="ThreadSafety.cs" />
     <Compile Include="XunitAssemblyAttributes.cs" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetsUnix)' == 'true' ">
+    <Compile Include="TermInfo.cs" />
+  </ItemGroup>
```
stephentoub: Hmm, I forgot this would be an issue.  Have you verified that the tests are actually running as part of CI?  I don't remember now if we assume the test assemblies are the same for both Windows and Unix.  If this is an issue, you could switch to using reflection to access internals rather than using InternalsVisibleTo.

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Sure thing

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            if (!Directory.Exists(location))
+                continue;
+
+            foreach (string term in Directory.EnumerateFiles(location, "*", SearchOption.AllDirectories))
+            {
+                if (term.ToUpper().Contains("README")) continue;
+                ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(ConsolePal.TermInfo.Database.ReadDatabase(Path.GetFileName(term)));
+
+                if (string.IsNullOrEmpty(info.ForegroundFormat) == false)
```
stephentoub: I know you prefer the ```== false```... that's fine, we should just be consistent, especially in the same function.... above you're using ```!``` with Directory.Exists.

sokket: Good point, making consistent :)

src/System.Console/tests/System.Console.Tests.csproj
sokket: Not sure and I can't seem to find my old builds...I'll check in the version I'm pushing out now

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+    [InlineData("xterm-256color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m")]
+    [InlineData("screen", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m")]
+    [InlineData("Eterm", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m")]
+    [InlineData("wsvt25", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m")]
+    [InlineData("mach-color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m")]
+    public void TermInfoVerification(string termToTest, string expectedForeground, string expectedBackground)
+    {
+        ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(ConsolePal.TermInfo.Database.ReadDatabase(termToTest));
+        Assert.Equal(expectedForeground, ConsolePal.TermInfo.ParameterizedStrings.Evaluate(info.ForegroundFormat, 0));
+        Assert.Equal(expectedBackground, ConsolePal.TermInfo.ParameterizedStrings.Evaluate(info.BackgroundFormat, 0));
```
stephentoub: Consider using a value other than 0, something that's not likely to be confused with a default value.  Some of the evaluation strings also do special processing that distinguishes between the bold and not-bold color values, so you might consider using a value from each range.

src/System.Console/tests/Color.cs
stephentoub: The value of the [Theory] approach is that a failed assert for one input doesn't prevent subsequent inputs from being tested, since each is logically considered its own test case.  It also has the benefit of keeping the generation of the inputs separate from the code actually doing the testing/verification. But it's not a big deal.

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Good idea

src/System.Console/tests/TermInfo.cs
sokket: Good idea, added

src/System.Console/tests/System.Console.Tests.csproj
stephentoub: Doesn't look like they're getting run.

sokket: Agreed. I'll switch the tests over to use reflection to invoke the methods rather than InternalsVisibleTo; I'll do the switch tomorrow morning before I go OOF

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,83 @@
...
+    [InlineData("xterm-256color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("xterm-256color", "\u001B\u005B\u00331m", "\u001B\u005B\u00341m", 1)]
+    [InlineData("screen", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("screen", "\u001B\u005B\u00332m", "\u001B\u005B\u00342m", 2)]
+    [InlineData("Eterm", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("Eterm", "\u001B\u005B\u00333m", "\u001B\u005B\u00343m", 3)]
+    [InlineData("wsvt25", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("wsvt25", "\u001B\u005B\u00334m", "\u001B\u005B\u00344m", 4)]
+    [InlineData("mach-color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("mach-color", "\u001B\u005B\u00335m", "\u001B\u005B\u00345m", 5)]
```
stephentoub: Nit: Could you also add some that use a value in the range [8,15] ?  The color values are broken into two groups: dark colors in the range [0,7] and light colors in the range [8,15], and some of these strings have special conditional logic for evaluating the cases where the values are greater than 7... I'd like to make sure we handle those correctly.

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,83 @@
...
+        Assert.Equal(expectedBackground, ConsolePal.TermInfo.ParameterizedStrings.Evaluate(info.BackgroundFormat, colorValue));
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.OSX)]
+    public void EmuTermInfoDoesntBreakParser()
+    {
+        // This file (available by default on OS X) is called out specifically since it contains a format where it has %i
+        // but only one variable instead of two. Make sure we don't break in this case
+        ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(ConsolePal.TermInfo.Database.ReadDatabase("emu"));
```
stephentoub: Rather than having this code, could you jut call TermInfoVerification("emu", ...)?  If we had the ability to have a PlatformSpecific InlineData, seems like this would be that, but since we don't, we're adding a fact that's platform-specific and then delegating.

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void TryingToLoadTermThatDoesNotExistDoesNotThrow()
+    {
+        ConsolePal.TermInfo.Database db = ConsolePal.TermInfo.Database.ReadDatabase("foobar____");
+        ConsolePal.TerminalColorInfo info = new ConsolePal.TerminalColorInfo(db);
+        Assert.Null(db);
+        Assert.NotNull(info);
```
stephentoub: TerminalColorInfo is a struct... it can never be null.

src/System.Console/tests/System.Console.Tests.csproj
stephentoub: Thanks.

src/System.Console/tests/TermInfo.cs
sokket: Ah good to know, done

src/System.Console/tests/TermInfo.cs
sokket: Good idea

src/System.Console/tests/TermInfo.cs
sokket: Ah right right

src/System.Console/src/System.Console.csproj
```diff
@@ -93,6 +93,7 @@
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetsUnix)' == 'true' ">
     <Compile Include="System\ConsolePal.Unix.cs" />
+    <Compile Include="System\InternalsVisibleTo.cs" />
```
stephentoub: InternalsVisibleTo is no longer needed, right?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -251,26 +251,27 @@ private struct TerminalColorInfo
...
-                        db.GetString(TermInfo.Database.OrigColorsIndex);
-
-                    int maxColors = db.GetNumber(TermInfo.Database.MaxColorsIndex);
-                    tci.MaxColors = // normalize to either the full range of all ANSI colors, just the dark ones, or none
-                        maxColors >= 16 ? 16 :
-                        maxColors >= 8 ? 8 :
-                        0;
-                }
-                return tci;
+                TermInfo.Database db = TermInfo.Database.Instance; // Could be null if TERM is set to a file that doesnt exist
```
stephentoub: Nit: missing apostrophe in "doesnt"

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -902,10 +903,16 @@ public static string Evaluate(string format, params FormatParam[] args)
                                     ~value);
                                 break;
 
-                            // Augment first two parameters by 1
+                                // Some terminfo files appear to have a fairly liberal interpretation of % i.The spec states that % i increments the first two arguments,
```
stephentoub: Nit: "%i. The" instead of "% i.The"

src/System.Console/src/System/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,2 @@
+// Let the test classes talk to internals
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("System.Console.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007D1FA57C4AED9F0A32E84AA0FAEFD0DE9E8FD6AEC8F87FB03766C834C99921EB23BE79AD9D5DCC1DD9AD236132102900B723CF980957FC4E177108FC607774F29E8320E92EA05ECE4E821C0A5EFE8F1645C4C0C93C1AB99285D622CAA652C1DFAD63D745D6F2DE5F17E5EAF0FC4963D261C8A12436518206DC093344D5AD293")]
```
stephentoub: Seems like this file can be deleted now.

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Fixed

src/System.Console/src/System.Console.csproj
sokket: Ah right, deleting

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Text;
+using Xunit;
+
+public class TermInfo
+{
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public void VerifyInstalledTermInfosParse()
+    {
+        string[] locations = GetFieldValueOnObject<string[]>("_terminfoLocations", null, Type.GetType("ConsolePal.TermInfo.Database"));
```
stephentoub: I'm surprised this works; I'd have thought a) you'd need to specify the full type name, including "System.", and b) that Type.GetType would only be looking at the current assembly and the core platform assembly, but not System.Console.dll.  I'd have expected you'd instead have needed something like:
```C#
typeof(Console).GetTypeInfo().Assembly.GetType("System.ConsolePal.TermInfo.Database")
```

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        }
+    }
+
+    [Theory]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    [InlineData("xterm-256color", "\u001B\u005B\u00330m", "\u001B\u005B\u00340m", 0)]
+    [InlineData("xterm-256color", "\u001B\u005B\u00331m", "\u001B\u005B\u00341m", 1)]
+    [InlineData("xterm-256color", "\u001B\u005B\u00338m", "\u001B\u005B\u00348m", 8)]
```
stephentoub: How did you arrive at the output strings for the new 8-15 versions?  They don't look right to me (but I haven't tried them).

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private string GetResetFormat(object colorInfo)
+    {
+        return GetFieldValueOnObject<string>("ResetFormat", colorInfo, Type.GetType("ConsolePal.TerminalColorInfo"));
+    }
+
+    private T GetFieldValueOnObject<T>(string name, object instance, Type baseType)
+    {
+        return (T)baseType.GetTypeInfo().DeclaredFields.Where(f => f.Name == name).Select(g => g).Single().GetValue(instance);
```
stephentoub: Why is the Select needed here?

src/System.Console/src/System/ConsolePal.Unix.cs
sokket: Fixed

src/System.Console/src/System/InternalsVisibleTo.cs
sokket: Yup!

src/System.Console/tests/TermInfo.cs
```diff
@@ -0,0 +1,132 @@
...
+        Assert.Null(db);
+        Assert.NotNull(GetBackgroundFormat(info));
+        Assert.NotNull(GetForegroundFormat(info));
+        Assert.NotNull(GetMaxColors(info));
+        Assert.NotNull(GetResetFormat(info));
+    }
+
+    private object ReadTermInfoDatabase(string term)
+    {
+        MethodInfo readDbMethod = Type.GetType("ConsolePal.TermInfo.Database").GetTypeInfo().DeclaredMethods.Where(m => m.Name == "ReadDatabase").Where(m1 => m1.GetParameters().Count() == 1).Select(m2 => m2).Single();
```
stephentoub: For all of these, same question/comment as the GetType one earlier.  If it works, great, I just wasn't expecting it to.

src/System.Console/tests/TermInfo.cs
sokket: I haven't had a chance to test since I'm WFH without a Linux Box available...hoping Jenkins tells me if everything is working or what needs to be fixed :)

src/System.Console/tests/TermInfo.cs
sokket: See above :)

src/System.Console/tests/TermInfo.cs
stephentoub: Rather than iterating through DeclaredMethods, can you just use GetDeclaredMethod?

sokket: Ah that is more useful

src/System.Console/tests/TermInfo.cs
sokket: Fixed

sokket:
@dotnet-bot test this please

stephentoub:
Thanks, Jon.  If it passes CI, LGTM.

sokket:
Forgot Unix tag on new tests, rerunning with them

sokket:
CI passed, merging. Thanks!

Issue #1993, opened 2015-06-10
[Feature] Finish X509Certificate support for Unix


bartonjs:
The current Unix X509 code was to the level that it needed to be to allow presenting an SSL wire cert for verification.  This needs to be completed.

The functions which currently throw a `NotImplementedException` are
```
Internal.Cryptography.Pal.CertificatePal::CertificatePal.FromFile(string, string, X509KeyStorageFlags)
Internal.Cryptography.Pal.CertificatePal::CertificatePal.FromHandle(IntPtr)
Internal.Cryptography.Pal.OpenSslX509CertificateReader::OpenSslX509CertificateReader.Archived.set(bool)
Internal.Cryptography.Pal.OpenSslX509CertificateReader::OpenSslX509CertificateReader.FriendlyName.set(string)
Internal.Cryptography.Pal.OpenSslX509CertificateReader::OpenSslX509CertificateReader.SetPrivateKey(AsymmetricAlgorithm, AsymmetricAlgorithm)
Internal.Cryptography.Pal.OpenSslX509ChainProcessor::OpenSslX509ChainProcessor.ChainElements.get()
Internal.Cryptography.Pal.OpenSslX509ChainProcessor::OpenSslX509ChainProcessor.ChainStatus.get()
Internal.Cryptography.Pal.OpenSslX509ChainProcessor::OpenSslX509ChainProcessor.Verify(X509VerificationFlags, Exception)
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.ComputeCapiSha1OfPublicKey(PublicKey)
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.DecodePublicKey(Oid, byte[], byte[])
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.DecodeX509BasicConstraintsExtension(byte[], bool, bool, int)
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.EncodeX509BasicConstraints2Extension(bool, bool, int)
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.EncodeX509EnhancedKeyUsageExtension(OidCollection)
Internal.Cryptography.Pal.OpenSslX509Encoder::OpenSslX509Encoder.EncodeX509KeyUsageExtension(X509KeyUsageFlags)
[...]

PR #2747, opened 2015-08-12
Add support for exporting as certificates on Unix.

bartonjs:
Adds support for exporting an X509Certificate2 or an X509Certificate2Collection as a PFX.

PFX export is limited to a single private key (issue 2743), and fails on an empty collection (issue 2746).


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
+
+namespace Internal.Cryptography.Pal
+{
+    internal class CollectionBackedStoreProvider : IStorePal
+    {
+        private readonly X509Certificate2Collection _certs;
```
justinvp: Could this be `List<X509Certificate2>`?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+                    {
+                        return Interop.libcrypto.OpenSslI2D(
+                            (handle, b) => Interop.libcrypto.i2d_PKCS12(handle, b),
+                            pkcs12);
+                    }
+                }
+            }
+        }
+
+        public IEnumerable<X509Certificate2> Certificates
```
justinvp: Could the interface be changed to return `List<X509Certificate2>`?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            get
+            {
+                foreach (X509Certificate2 cert in _certs)
+                {
+                    yield return cert;
```
justinvp: If possible, I would avoid `yield` and just return the collection. Even better if the signature and underlying collection can be changed to `List<T>`.

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -848,3 +848,69 @@ int BioSeek(
...
+    EVP_PKEY* pkey)
+{
+    if (!pkey)
+    {
+        return 0;
+    }
+
+    int i = CRYPTO_add(&pkey->references, 1, CRYPTO_LOCK_EVP_PKEY);
+
+    return (i > 1);
```
stephentoub: So CRYPTO_add returns the new number of references, and it could never be 1 because that would mean it was previously 0 at which point it would have been released?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+            _certs = new X509Certificate2Collection(certs);
+        }
+
+        public void Dispose()
+        {
+        }
+
+        public IEnumerable<X509Certificate2> Find(X509FindType findType, object findValue, bool validOnly)
+        {
+            return Array.Empty<X509Certificate2>();
```
stephentoub: It's not possible that the thing being searched for could match _certs[0]?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+        private readonly X509Certificate2Collection _certs;
+
+        internal CollectionBackedStoreProvider(X509Certificate2 cert)
+        {
+            _certs = new X509Certificate2Collection(cert);
+        }
+
+        internal CollectionBackedStoreProvider(X509Certificate2Collection certs)
+        {
+            _certs = new X509Certificate2Collection(certs);
```
stephentoub: I'm a little confused about what's supported.  Add throws an InvalidOperationException, and ExportX509Der only uses _certs[0], so I was assuming that this currently only supports wrapping a single certificate.  Is that true?  If so, should we assert that certs.Count <= 1?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+        public byte[] Export(X509ContentType contentType, string password)
+        {
+            switch (contentType)
+            {
+                case X509ContentType.Cert:
+                    return ExportX509Der();
+                case X509ContentType.Pfx:
+                    return ExportPfx(password);
+                default:
+                    throw new CryptographicException();
```
stephentoub: Is this Export function exposed publically such that something other than Cert or Pfx could be passed in?

src/System.Security.Cryptography.X509Certificates/tests/CollectionTests.cs
```diff
@@ -360,6 +376,136 @@ public static void X509CertificateCollectionSyncRoot()
...
+                // If Windows gets here, or any exception other than PlatformNotSupportedException is raised,
+                // let that fail the test.
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    throw;
+                }
+
+                return;
+            }
+
```
stephentoub: Do we need a check here that RuntimeInformation.IsOSPlatform(OSPlatform.Windows) is true, to verify that we did get a PNSE on Unix?

stephentoub:
A few comments, but otherwise LGTM.

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: Yep, that's my understanding, too.  But reading deeper I think that's you suggesting that it is comment-worthy :).

stephentoub: :wink: And maybe assert-worthy.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
bartonjs: I think this is just telling me that I need to reconstruct the PAL interface here a bit.  In Windows the export operations happen on the same data structure as the X509Store; so the export code says "make a memory store, then call export on it"; and this is an implementation of that memory store just for the duration of a call to Export; so calls to Find shouldn't happen (which means it should probably be an Exception; but I copy/pasted this from "Find is valid, but doesn't work yet" in the system stores).

But, that said, I'd prefer to do that reshaping to a "Here, export this collection" instead of "Make a store from a collection; hey store, go export yourself" as a followup, so I can try to get the changes that depend on this (black box) in for beta7 :).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
bartonjs: Windows compat.  Calling Export(X509ContentType.Cert) on a collection returns _certs[0].RawData (or null); and ignores _certs[1..] outright.  I'll add some comments :).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
bartonjs: Ish.  Values other than that can get into here; but this object is never exposed outside of the library plumbing.  I can switch this back to yet another NotImplementedException to be cleaned up in the #1993 cleanup pass.  (The Serialized* forms are probably going to be PlatformNotSupportedException, because they're CAPI structures, not standards).

PKCS#7 will probably get around to appearing in beta8/rc.

src/System.Security.Cryptography.X509Certificates/tests/CollectionTests.cs
bartonjs: Good call

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
bartonjs: I can certainly change the field to be a List; and could change the internal interface to say that it returns a list; but I really don't want to give out the collection (you'll note that Add and Remove both throw); so at best this would be `return new List<X509Certificate2>(_certs)`.

Or I guess I could just store it as `_certs = new List<X509Certificate2>(ctorCerts).AsReadOnly();`, and return that here.

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: Assert in the calling managed code that true was returned, or some sort of native assert?  (If "some sort of native assert" is the right answer, could you point it out to me? I don't know of any)

stephentoub: How about just returning ```i``` rather than the result of the comparison, and then doing the validation and the comparison in the calling managed code?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs
justinvp: Yeah, I see why you'd want to avoid exposing the collection to callers (even if the callers are internal). Not as part of this PR, but maybe revise the `IStorePal` interface at some point in the future... It looks like all the callers of `Certificates` and `Find` just end up adding each item to a `X509Certificate2Collection`. Perhaps changing `Certificates` and `Find` to methods that take a `X509Certificate2Collection` to add to, and have the `StorePal` handle enumerating an adding to the collection. This would avoid the enumerator allocations (and any intermediate read-only collections).

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: I did it the way I did because that's how OpenSSL's `RSA_up_ref` works, but I'm happy enough with the argument of "but since OpenSSL didn't write EVP_PKEY_up_ref, we did, we'll do it our way".

bartonjs:
Letting CI run on the incremental change, will squash before merging.

PR #2749, opened 2015-08-12
Add performance tests for BigInteger

axelheer:
These unit tests should help measuring performance, until we have some kind of real performance-measurement system in place.

#1307

### DIFF
--- a/src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
+++ b/src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
@@ -1,0 +1,198 @@
+﻿using System;
+using System.Diagnostics;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Numerics.Tests
+{
+    public class PerformanceTests
+    {
+        private readonly Random _random;
+        private readonly ITestOutputHelper _output;
+
+        public PerformanceTests(ITestOutputHelper output)
+        {
+            _random = new Random(1138);
+            _output = output;
+        }
+
+        [Theory]
+        [ActiveIssue("PerformanceTest")]
+        [InlineData(1000000, 16, 16)]
+        [InlineData(1000000, 64, 64)]
+        [InlineData(1000000, 256, 256)]
+        [InlineData(1000000, 1024, 1024)]
+        [InlineData(100000, 4096, 4096)]
+        [InlineData(100000, 16384, 16384)]
+        [InlineData(100000, 65536, 65536)]
+        public void Add(int count, int leftBits, int rightBits)
+        {
+            RunBenchmark(count, leftBits, rightBits, (l, r) => BigInteger.Add(l, r));
+        }
+
+        [Theory]
+        [ActiveIssue("PerformanceTest")]
+        [InlineData(1000000, 16, 16)]
+        [InlineData(1000000, 64, 64)]
+        [InlineData(1000000, 256, 256)]
+        [InlineData(1000000, 1024, 1024)]
+        [InlineData(100000, 4096, 4096)]
+        [InlineData(100000, 16384, 16384)]
+        [InlineData(100000, 65536, 65536)]
+        public void Subtract(int count, int leftBits, int rightBits)
+        {
+            RunBenchmark(count, leftBits, rightBits, (l, r) => BigInteger.Subtract(l, r));
+        }
+
+        [Theory]
+        [ActiveIssue("PerformanceTest")]
+        [InlineData(1000000, 16, 16)]
+        [InlineData(1000000, 64, 64)]
+        [InlineData(1000000, 256, 256)]
+        [InlineData(100000, 1024, 1024)]
+        [InlineData(10000, 4096, 4096)]
+        [InlineData(1000, 16384, 16384)]
+        [InlineData(100, 65536, 65536)]
+        public void Multiply(int count, int leftBits, int rightBits)
+        {
+            RunBenchmark(count, leftBits, rightBits, (l, r) => BigInteger.Multiply(l, r));
+        }
+
+        [Theory]
+        [ActiveIssue("PerformanceTest")]
+        [InlineData(1000000, 16)]
+        [InlineData(1000000, 64)]
+        [InlineData(1000000, 256)]
+        [InlineData(100000, 1024)]
+        [InlineData(10000, 4096)]
+        [InlineData(1000, 16384)]
+        [InlineData(100, 65536)]
+        public void Square(int count, int bits)
+        {
+            RunBenchmark(count, bits, v => BigInteger.Multiply(v, v));
+        }
+
+        [Theory]
+        [ActiveIssue("PerformanceTest")]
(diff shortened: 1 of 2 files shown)


axelheer:
cc: @stephentoub

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        [InlineData(1000000, 16)]
+        [InlineData(1000000, 64)]
+        [InlineData(1000000, 256)]
+        [InlineData(100000, 1024)]
+        [InlineData(10000, 4096)]
+        [InlineData(1000, 16384)]
+        [InlineData(100, 65536)]
+        public void Square(int count, int bits)
+        {
+            RunBenchmark(count, bits, v => BigInteger.Multiply(v, v));
```
stephentoub: Are there special code paths in place for when the two values are the same?  I'm just wondering why Square is it's own perf test when it just calls Multiply and we have a separate Multiply benchmark above.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        [InlineData(1000000, 16, 16)]
+        [InlineData(1000000, 64, 16)]
+        [InlineData(1000000, 256, 128)]
+        [InlineData(100000, 1024, 512)]
+        [InlineData(10000, 4096, 2048)]
+        [InlineData(1000, 16384, 8192)]
+        [InlineData(100, 65536, 32768)]
+        public void Remainder(int count, int leftBits, int rightBits)
+        {
+            RunBenchmark(count, leftBits, rightBits, (l, r) => BigInteger.Remainder(l, r));
```
stephentoub: For Divide and Remainder, is there any concern that we could be getting unlucky with the chosen values so that the numerator is frequently smaller than the denominator and we don't really exercise the main code paths?

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        private const int RUN_COUNT = 3;
+
+        private long RunBenchmark(int count, Action<int> operation)
+        {
+            operation(0);
+
+            long result = long.MaxValue;
+            for (int j = 0; j < RUN_COUNT; j++)
+            {
+                Stopwatch watch = new Stopwatch();
```
stephentoub: You can move this to be outside of the loop, and then just use Restart instead of Start.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+                watch.Start();
+                for (int i = 0; i < count; i++)
+                    operation(i % MAX_SEED);
+                watch.Stop();
+
+                result = Math.Min(result, watch.ElapsedMilliseconds);
```
stephentoub: Interesting. Why min rather than mean or something else?  Mostly just curious.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            return seed;
+        }
+
+        private BigInteger CreateRandomInteger(int bits)
+        {
+            byte[] value = new byte[(bits + 8) / 8];
+            BigInteger result = BigInteger.Zero;
+
+            while (result.IsZero)
```
stephentoub: I assume this IsZero check is to avoid problems with division/remainder?

stephentoub:
Thanks, @axelheer.  A few comments, but otherwise LGTM.  We'll eventually move these to whatever perf testing system we land on for the repro, but it's great to have these here for now so that we can all compare based on the same benchmarks as improvements are being made.

@ericeil, @pharring, just FYI.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: Yeah, if the two internal arrays are the same, an optimized algorithm is being used, which takes some advantage of that fact.

stephentoub: Makes sense then.  Thanks.  Maybe just add a comment here to that effect so that future passerby's understand why this may appear duplication with the above.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: Dividing two values of roughly the same length wouldn't execute much of the code, the resulting quotient would be very small. If we choose a numerator half the length, the resulting quotient should have roughly half the length too, which leads to a well-balanced exercise, I think.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: Ah, I'll amend that.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
stephentoub: Ok.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: The idea is, that one run isn't deterministic enough, since there could be some kind of noise. So each benchmark gets three opportunities to prove its point.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: Yep, a value greater zero should satisfy the requirements of every operation.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
stephentoub: In terms of determinism and being able to compare results across runs, should we really be relying on randomly-generated values, or should we hardcode the specific values we want to test on each run?  I realize you're providing a hardcoded seed to Random, but I'm not sure that'll actually guarantee determinism across runs, e.g. if xunit runs the tests in a different order each time.

stephentoub:
Thanks.

src/System.Runtime.Numerics/tests/BigInteger/PerformanceTests.cs
axelheer: I agree that every run should use the same values. Since xunit creates a fresh instance for every run and every instance starts with a hardcoded seed, that should be the case. I haven't verified it, but I'm quite sure about it.

PR #2752, opened 2015-08-12
Performance improvements to Skip and SkipWhile.

### LINKED ISSUE #2238: Linq loops could be tighter; causes bug in one obscure case, waste in others.

JonHanna:
Fixes #2238
(And I think this is the last bit to separate out of the original PR)

https://github.com/hackcraft/Enumerable-Tester/raw/master/Skip%20performance.ods and https://github.com/hackcraft/Enumerable-Tester/raw/master/new%20version%20skip%20improvements.ods
show comparisons of the old and new approaches. Improvements are made in
the vast majority of cases.

(This ignores the matter of ordered skips as per #2401).

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Threading;
 
 namespace System.Linq
 {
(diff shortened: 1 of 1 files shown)


hackcraft:
When I closed #2446 I said:

> I think I'll see what happens if I redo Skip much as SkipWhile is done here (adjusted as per what @VSadov said about the && in the thread check), since that it won't be a big job to get from here to there and while it may not be as dramatic an improvement it might still be worth it now I nearly have it, but as I won't have time to do that for a while, for now I'm going to just close this out.

I've done exactly that, and results are not bad at all. https://github.com/hackcraft/Enumerable-Tester/raw/master/new%20version%20skip%20improvements.ods shows the improvements Skip now has over the current implementation, and while it compares well with my first attempt while not being too clever for its own good as my first perhaps was.
This version should also have no backwards-compatibility issues at all.

hackcraft:
Marking this [WIP] as there's some more I want to do on this, I think.

stephentoub:
cc: @VSadov 

@hackcraft, can you elaborate on the key ideas behind this implementation, what it is you're doing that's making things work better in some cases?  It's increasing the amount of code in these implementations by almost 12x, replacing ~40 lines of code with ~460 lines of code, and adding a significant amount of complexity.

hackcraft:
The key idea **had** been to pass through on the source enumerator to the caller in the skip, in #2446 

The result of removing all the potential issues with that produced what is close what you would have if you hand-optimised the compiler-generated iterator, albeit arrived at from a different direction. I wouldn't have considered it worth trying, but that I arrived at it through trying something else.

The one genuine idea it does have now, and which was only added at the very end, is of optimising the combination `.Skip().Take()`, which is a very common combo (`.Skip(pageNum * pageSize).Take(pageSize)`.

Maybe I should cut this back to just what's needed to support that idiom, and see how well that does.

PR #2753, opened 2015-08-12
Fixed interpreter support for properties on Nullable<T>

VSadov:
such as HasValue / Value

Also fixed a problem with invoking of GetValueOrDefault on null receiver discovered while testing.
We should return a default  value in such case, not the boxed receiver which is null and would cause NRE.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
@@ -1925,14 +1925,10 @@ namespace System.Linq.Expressions.Interpreter
         {
             public override int Run(InterpretedFrame frame)
             {
-                var obj = frame.Pop();
-                if (obj == null)
+                if (frame.Peek() == null)
                 {
+                    frame.Pop();
                     throw new InvalidOperationException();
-                }
-                else
-                {
-                    frame.Push(obj);
                 }
                 return +1;
             }
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
@@ -1899,7 +1899,7 @@ namespace System.Linq.Expressions.Interpreter
 
     internal abstract class NullableMethodCallInstruction : Instruction
     {
-        private static NullableMethodCallInstruction s_hasValue,s_value,s_equals,s_getHashCode,s_getValueOrDefault1,s_toString;
+        private static NullableMethodCallInstruction s_hasValue, s_value, s_equals, s_getHashCode, s_getValueOrDefault1, s_toString;
 
         public override int ConsumedStack { get { return 1; } }
         public override int ProducedStack { get { return 1; } }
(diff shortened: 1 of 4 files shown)


VSadov:
@AlekseyTs - please review

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
```diff
@@ -1899,7 +1899,7 @@ public override string ToString()
 
     internal abstract class NullableMethodCallInstruction : Instruction
     {
-        private static NullableMethodCallInstruction s_hasValue,s_value,s_equals,s_getHashCode,s_getValueOrDefault,s_getValueOrDefault1,s_toString;
+        private static NullableMethodCallInstruction s_hasValue,s_value,s_equals,s_getHashCode,s_getValueOrDefault1,s_toString;
```
stephentoub: Nit: I realize you only changed the name of one of these, but it'd be nice if there were spaces between each field.

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
```diff
@@ -1940,9 +1940,20 @@ public override int Run(InterpretedFrame frame)
 
         private class GetValueOrDefault : NullableMethodCallInstruction
         {
+            private Type defaultValueType;
```
stephentoub: readonly?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
```diff
@@ -1940,9 +1940,20 @@ public override int Run(InterpretedFrame frame)
...
+            {
+                defaultValueType = mi.ReturnType;
+            }
+
             public override int Run(InterpretedFrame frame)
             {
-                // we're already unboxed in the interpreter, so this is a nop.
+                if (frame.Peek() == null)
+                {
+                        frame.Pop();
```
stephentoub: Nit: indentation looks off here

src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
```diff
@@ -4266,6 +4266,31 @@ public static void TestConvertToNullable()
             Assert.Equal(2, d(2));
         }
 
+        [Fact]
```
stephentoub: Remind me: will these tests run with both the compiled and interpreted versions?

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/TypeOperations.cs
VSadov: no prob

src/System.Linq.Expressions/tests/SequenceTests/SequenceTests.cs
VSadov: I could not yet figure how to run tests in both modes. 
The problem is that interpretation is not a runtime switch, it requires rebuild. It is easy to run test in interpretation mode locally by setting an environment variable and forcing rebuild.

I will be porting these fixes to ProjectN branch where they will definitely run in interpreter mode, but yes, we need a way to run interpreter in CI.

VSadov: Entered https://github.com/dotnet/corefx/issues/2756 to track this test hole

PR #2754, opened 2015-08-12
More robust RsaCng.Encrypt/Decrypt/SignHash.

AtsushiKan:
This addresses two pieces of feedback from WinCrypto:

- NCryptDecrypt() can potentially ask for more buffer space than
  he ends up needing (saves him from having to do a extra decrypt
  just to calculate the size.) We need to guard against that
  and shrink the array to fit, in case the api overbudgeted.

  I've added the resize to other places where we do this
  "okay, here's 0 bytes of space. What, you want more? Okay,
  here it is" API pattern. It's cheap insurance.

- Try to avoid the double-call in some cases by estimating
  a buffer size upfront (Encrypt/Decrypt/SignHash).

### DIFF
--- a/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
+++ b/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
@@ -26,17 +26,17 @@ namespace System.Security.Cryptography
             if (format == null)
                 throw new ArgumentNullException("format");
 
-            int numBytesRequired;
-            ErrorCode errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, null, 0, out numBytesRequired, 0);
+            int numBytesNeeded;
+            ErrorCode errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, null, 0, out numBytesNeeded, 0);
             if (errorCode != ErrorCode.ERROR_SUCCESS)
                 throw errorCode.ToCryptographicException();
 
-            byte[] buffer = new byte[numBytesRequired];
-            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, buffer.Length, out numBytesRequired, 0);
+            byte[] buffer = new byte[numBytesNeeded];
+            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, buffer.Length, out numBytesNeeded, 0);
             if (errorCode != ErrorCode.ERROR_SUCCESS)
                 throw errorCode.ToCryptographicException();
 
-            Array.Resize(ref buffer, numBytesRequired);
+            Array.Resize(ref buffer, numBytesNeeded);
             return buffer;
         }
     }
--- a/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
+++ b/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
@@ -32,10 +32,11 @@ namespace System.Security.Cryptography
                 throw errorCode.ToCryptographicException();
 
             byte[] buffer = new byte[numBytesRequired];
-            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, numBytesRequired, out numBytesRequired, 0);
+            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, buffer.Length, out numBytesRequired, 0);
             if (errorCode != ErrorCode.ERROR_SUCCESS)
                 throw errorCode.ToCryptographicException();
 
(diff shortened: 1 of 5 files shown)


src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
```diff
@@ -32,10 +32,11 @@ public byte[] Export(CngKeyBlobFormat format)
                 throw errorCode.ToCryptographicException();
 
             byte[] buffer = new byte[numBytesRequired];
-            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, numBytesRequired, out numBytesRequired, 0);
+            errorCode = Interop.NCrypt.NCryptExportKey(_keyHandle, IntPtr.Zero, format.Format, IntPtr.Zero, buffer, buffer.Length, out numBytesRequired, 0);
```
stephentoub: Here you changed numBytesRequired to buffer.Length, but in the previous file you left it as numBytesRequired.  I realize they're the same value, but any reason for the discrepancy?

AtsushiKan: Which file are you referring to? The only one’s I could find were the ones where I was passing a pinned pointer down (and thus has no Length property, and passing down the source array’s Length didn’t seem any more robust than passing down “numBytesRequired.”)

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.EncryptDecrypt.cs
```diff
@@ -84,18 +84,25 @@ private byte[] EncryptOrDecrypt(byte[] data, RSAEncryptionPadding padding, Encry
...
         // Now that the padding mode and information have been marshaled to their native counterparts, perform the encryption or decryption.
         //
-        private static unsafe byte[] EncryptOrDecrypt(SafeNCryptKeyHandle key, byte[] input, AsymmetricPaddingMode paddingMode, void* paddingInfo, EncryptOrDecryptAction encryptOrDecrypt)
+        private unsafe byte[] EncryptOrDecrypt(SafeNCryptKeyHandle key, byte[] input, AsymmetricPaddingMode paddingMode, void* paddingInfo, EncryptOrDecryptAction encryptOrDecrypt)
         {
-            int numBytesNeeded;
-            ErrorCode errorCode = encryptOrDecrypt(key, input, input.Length, paddingInfo, null, 0, out numBytesNeeded, paddingMode);
-            if (errorCode != ErrorCode.ERROR_SUCCESS)
-                throw errorCode.ToCryptographicException();
+            int estimatedSize = KeySize / 8;
```
stephentoub: How likely is it that this is too small?  If it's unlikely, makes sense.  If it's somewhat likely, I'm wondering if it's really better than starting with 0 and avoiding the estimated-size allocation.

AtsushiKan: I obtained this my experimenting with a bunch of key sizes and padding algorithms. This was always on the nose for Encrypts and Signs and just a little above the needed size (due to padding) for Decrypts.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/CngKey.Export.cs
stephentoub: https://github.com/AtsushiKan/corefx/blob/rsaencrypt/src/System.Security.Cryptography.Cng/src/Internal/Cryptography/Helpers.cs#L80

AtsushiKan: Ok, I think misread your question, then. Since I was calling the Length property pretty much everywhere else, I decided to make this one consistent with that too, as it’s a more robust pattern. I hate debugging native buffer overruns.

stephentoub: Ok. Doesn't the same reasoning apply to that other call site?  I'm just going for consistency.

AtsushiKan: The other call site passes a byte* rather than a byte[]. So I can’t compute the length of it. I could pass down the Length of the original array, but this doesn’t seem any more robust than what’s there now. Either way, the callsite isn’t self-correcting in the light of surrounding changes.

stephentoub: You're trying to keep the buffer argument and the length argument as in sync as possible.  In the other call site, the numBytesNeeded is used to create the propertyValue array which is used to create the pPropertyValue pointer, so two hops between the numBytesNeeded and the pPropertyValue, whereas there's only one hop between propertyValue (and propertyValue.Length) and pPropertyValue.  Plus the names are almost identical, so there's a clear connection between them: looking just at the call site, I get the connection between "pPropertyValue" and "propertyValue.Length" much more so than I do between "pPropertyValue" and "numBytesNeeded".

If it were me and I were making this change, I'd change both.  But it's not a big deal.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

AtsushiKan:
cc: @bartonjs

AtsushiKan:
cc: @bartonjs

bartonjs:
LGTM

PR #2755, opened 2015-08-12
Downgrade Reflection.Metadata to Immutable Collections 1.1.36

tmat:
Can't use the latest version w/o updating VS.


tmat:
@nguerrera @davkean

davkean:
Why the change looks good, can we wait for Nick? I have no idea why 1.1.37 was picked up - it may have been for a real reason.

tmat:
@davkean Nick is on vacation. We need this to unblock a customer. I believe it was just a result of an auto-merge.

nguerrera:
This is OK by me. I had taken care to keep it 1.1.36 in dev/metadata for this very reason, but seems a merge from master was done that breaks that.

AlexGhiondea:
@tmat @nguerrera That was probably me -- I wasn't aware of this restriction. Thanks for fixing it!

PR #2766, opened 2015-08-12
BugFix: Test modules from CurrentProcess to avoid cross bit access on 64bit machines

### LINKED ISSUE #2750: Two Process tests failed in CI due to GetModuleInfos failing
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release/961/console
```
System.Diagnostics.ProcessTests.ProcessTests.TestModules [FAIL]
        System.ComponentModel.Win32Exception : Only part of a ReadProcessMemory or WriteProcessMemory request was completed
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\src\System\Diagnostics\ProcessManager.Windows.cs(406,0): at System.Diagnostics.NtProcessManager.GetModuleInfos(Int32 processId, Boolean firstModuleOnly)
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\src\System\Diagnostics\Process.cs(277,0): at System.Diagnostics.Process.get_Modules()
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\tests\ProcessTests.cs(278,0): at System.Diagnostics.ProcessTests.ProcessTests.TestModules()
```
and
```
System.Diagnostics.ProcessTests.ProcessCollectionTests.TestModuleCollectionBehavior [FAIL]
        System.ComponentModel.Win32Exception : Only part of a ReadProcessMemory or WriteProcessMemory request was completed
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\src\System\Diagnostics\ProcessManager.Windows.cs(406,0): at System.Diagnostics.NtProcessManager.GetModuleInfos(Int32 processId, Boolean firstModuleOnly)
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\src\System\Diagnostics\Process.cs(277,0): at System.Diagnostics.Process.get_Modules()
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Diagnostics.Process\tests\ProcessCollectionTests.cs(0,0): at System.Diagnostics.ProcessTests.ProcessCollectionTests.TestModuleCollectionBehavior()
```

Priya91:
fixes #2750

### DIFF
--- a/src/System.Diagnostics.Process/tests/ProcessCollectionTests.cs
+++ b/src/System.Diagnostics.Process/tests/ProcessCollectionTests.cs
@@ -12,7 +12,7 @@ namespace System.Diagnostics.ProcessTests
         [Fact]
         public void TestModuleCollectionBehavior()
         {
-            ProcessModule[] mArray = _process.Modules.Cast<ProcessModule>().ToArray();
+            ProcessModule[] mArray = Process.GetCurrentProcess().Modules.Cast<ProcessModule>().ToArray();
 
             // Constructor
             ProcessModuleCollection moduleCollection = new ProcessModuleCollection(mArray);
--- a/src/System.Diagnostics.Process/tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/ProcessTests.cs
@@ -275,7 +275,8 @@ namespace System.Diagnostics.ProcessTests
         [Fact]
         public void TestModules()
         {
-            foreach (ProcessModule pModule in _process.Modules)
+            ProcessModuleCollection moduleCollection = Process.GetCurrentProcess().Modules;
+            foreach (ProcessModule pModule in moduleCollection)
             {
                 // Validated that we can get a value for each of the following.
                 Assert.NotNull(pModule);


src/System.Diagnostics.Process/tests/ProcessCollectionTests.cs
```diff
@@ -12,7 +12,7 @@ public class ProcessCollectionTests : ProcessTestBase
         [Fact]
         public void TestModuleCollectionBehavior()
         {
-            ProcessModule[] mArray = _process.Modules.Cast<ProcessModule>().ToArray();
+            ProcessModule[] mArray = Process.GetCurrentProcess().Modules.Cast<ProcessModule>().ToArray();
```
stephentoub: @Priya91, can you remind me how we'd end up in a situation where _process is a different bitness than Process.GetCurrentProcess()?

Priya91: ```_process.Start``` is starting ```RemoteExecutorConsoleApp.exe```, which is built for ```Any CPU```, and ```Prefer 32bit``` property is not set on the project. So while running on 64 bit machines, it will start as 64 bit process, and the ```xunit.console.netcore.exe``` is a 32bit process. ```Process.GetCurrentProcess()``` will access the 32 bit ```xunit.console.netcore.exe``` process so there should be no bitness problem.

stephentoub: Got it. Thanks.

stephentoub:
LGTM

src/System.Diagnostics.Process/tests/ProcessCollectionTests.cs
mellinoe: Not saying the fix is wrong, but regarding this:

> RemoteExecutorConsoleApp.exe, which is built for Any CPU, and Prefer 32bit property is not set on the project. So while running on 64 bit machines, it will start as 64 bit process

Won't the bitness of the process just be based on which runtime we are using? Our tests are all using a 32-bit runtime (at least for now). That's why the xUnit console process is 32-bit (it's actually compiled as AnyCPU 64-bit preferred). I would guess that the Remote Executor process is using the same runtime and therefore is probably also 32-bit.

Priya91: Hmm, taking a closer look at the line where this is thrown, it could be because even after 50 tries, the process modules weren't loaded or the OS loader is touching the module information during this time. For this case too, win32 api throws an exception, with the same error message that is ```Only part of a ReadProcessMemory or WriteProcessMemory request was completed``` as the bitness problem.

Either way, using the CurrentProcess Modules should solve both these problems. One scenario where the CurrentProcess may fail will be when the OSLoader is touching the module information of the current process. But in the 50 iterations we should have a success hit, as another test with similar failure before has stopped failing with this change.

PR #2775, opened 2015-08-12
Add additional System.Threading.Tasks tests

mellinoe:
I've ported the remainder of our System.Threading.Tasks tests from our internal sources. A lot of these are moved over unchanged (except for porting to xUnit).

Some of the tests were not really designed to be run one after another, and left a lot of "orphaned" tasks doing a whole bunch of work even when the test case finished running. This was causing unpredictable performance, as some test would take a really long time in a "full run", but very little time if run individually, because there were tons of leftover tasks still running. I've attempted to clean those cases up, so that each test case cleans up after the tasks it starts. I have also marked a decent chunk of these tests as OuterLoop because they tend to run for a very long time (some just by their nature, some probably can be optimized if we want to make them inner loop tests).

@stephentoub


mellinoe:
Fixes #702

mellinoe:
Hit a failure in a crypto test

mellinoe:
@dotnet-bot test this, please

src/System.Threading.Tasks/tests/AggregateExceptionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Diagnostics;
+
+namespace System.Threading.Tasks.Tests
+{
+    public class AggregateExceptionTests
+    {
+        [Fact]
+        public static void RunAggregateException_Constructor()
+        {
+            AggregateException ex = new AggregateException();
```
stephentoub: I was about to say that it's strange to have AggregateException tests here rather than in the contract from which AggregateException is exposed... and then I discovered that AggregateException is in fact exposed from System.Threading.Tasks.  That's odd, but... oh well.

src/System.Threading.Tasks/tests/AggregateExceptionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Threading.Tasks.Tests
+{
+    public class AggregateExceptionTests
+    {
+        [Fact]
+        public static void RunAggregateException_Constructor()
```
stephentoub: The class is named "AggregateExceptionTests"... we probably don't need "RunAggregateException_" as a prefix on the name of every test it contains.

src/System.Threading.Tasks/tests/AggregateExceptionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            // > Flattening (no recursion)...
+
+            AggregateException flattened1 = aggExceptionBase.Flatten();
+            Exception[] expected1 = new Exception[] {
+                exceptionA, exceptionB, exceptionC
+            };
+
+            if (expected1.Length != flattened1.InnerExceptions.Count)
+            {
+                Assert.True(false, string.Format("RunAggregateException_Flatten: > error: expected count {0} differs from actual {1}",
```
stephentoub: Assert.Equal(expected1.Length, flattened1.InnerExceptions.Count);

src/System.Threading.Tasks/tests/AggregateExceptionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+                    expected1.Length, flattened1.InnerExceptions.Count));
+            }
+
+            for (int i = 0; i < flattened1.InnerExceptions.Count; i++)
```
stephentoub: Could this whole loop (and the previous assert) just be Assert(expected1, flattened1.InnerExceptions)?

src/System.Threading.Tasks/tests/AggregateExceptionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            if (expected2.Length != flattened2.InnerExceptions.Count)
+            {
+                Assert.True(false, string.Format("RunAggregateException_Flatten:  FAILED.  > error: expected count {0} differs from actual {1}",
+                    expected2.Length, flattened2.InnerExceptions.Count));
+            }
+
+            for (int i = 0; i < flattened2.InnerExceptions.Count; i++)
+            {
+                if (expected2[i] != flattened2.InnerExceptions[i])
```
stephentoub: Ditto.

src/System.Threading.Tasks/tests/CESchedulerPairTests_Yeti.cs
```diff
@@ -0,0 +1,674 @@
...
+        protected override IEnumerable<Task> GetScheduledTasks() { return null; }
+        private Object _lockObj = new Object();
+        private int _counter = 1; //This is used ot keep track of how many scheduler tasks were created
+        public ThreadLocal<int> SchedulerID = new ThreadLocal<int>(); //This is the ID of the scheduler. 
+
+        /// <summary>The maximum concurrency level for the scheduler.</summary>
+        private readonly int _maxConcurrencyLevel;
+        public override int MaximumConcurrencyLevel { get { return _maxConcurrencyLevel; } }
+    }
+    public class CESchedulerPairTests
```
stephentoub: Nit: blank line

src/System.Threading.Tasks/tests/CESchedulerPairTests_Yeti.cs
```diff
@@ -0,0 +1,674 @@
...
+        [Fact]
+        public static void TestCreationOptions()
+        {
+            string[] input = new string[] { "default", "scheduler", "maxconcurrent", "all" };
+            foreach (string s in input)
+            {
+                Debug.WriteLine("***** Starting {0}", s);
+                TestCreationOptions(s);
+            }
+        }
```
stephentoub: This can just be a Theory.

src/System.Threading.Tasks/tests/CESchedulerPairTests_Yeti.cs
```diff
@@ -0,0 +1,674 @@
...
+        [Fact]
+        [OuterLoop]
+        public static void TestMaxItemsPerTask()
+        {
+            TestMaxItemsPerTask(4, 1, true);
+            TestMaxItemsPerTask(1, 4, true);
+            TestMaxItemsPerTask(4, 1, false);
+            TestMaxItemsPerTask(1, 4, false);
+        }
+        public static void TestMaxItemsPerTask(int maxConcurrency, int maxItemsPerTask, bool completeBeforeTaskWait)
```
stephentoub: Could just be a theory.

src/System.Threading.Tasks/tests/CESchedulerPairTests_Yeti.cs
```diff
@@ -0,0 +1,674 @@
...
+        /// Test to ensure that when exclusive tasks are scheduled no concurrent tasks are run
+        /// </summary>
+        [Fact]
+        public static void TestConcurrentBlockage()
+        {
+            TestConcurrentBlockage(true);
+            TestConcurrentBlockage(false);
+        }
+
+        public static void TestConcurrentBlockage(bool useReader)
```
stephentoub: Can be a Theory.  I'll stop commenting on these.

src/System.Threading.Tasks/tests/System.Threading.Tasks.Tests.csproj
```diff
@@ -26,5 +23,54 @@
...
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="XunitAssemblyAttributes.cs" />
+    <Compile Include="UnwrapTests.cs" />
+    <Compile Include="AggregateExceptionTests.cs" />
+    <Compile Include="OperationCanceledExceptionTests.cs" />
+    <Compile Include="CancellationTokenTests.cs" />
+    <Compile Include="MethodCoverage.cs" />
+    <Compile Include="CESchedulerPairTests_Yeti.cs" />
```
stephentoub: Can we remove _Yeti from the file name?

src/System.Threading.Tasks/tests/MethodCoverage.cs
```diff
@@ -0,0 +1,256 @@
...
+            Task t1 = Task.Factory.StartNew(() => { mre.WaitOne(); taskId1 = Task.CurrentId; });
+            Task t2 = Task.Factory.StartNew(() => { mre.WaitOne(); taskId2 = Task.CurrentId; cts.Cancel(); });
+
+            List<Task<int?>> whenAllTaskResult = new List<Task<int?>>();
+            List<Task> whenAllTask = new List<Task>();
+            whenAllTask.Add(t1); whenAllTask.Add(t2);
+            Task<int> contTask = Task.WhenAll(whenAllTask).ContinueWith<int>(
+                (task) =>
+                {
+                    Console.WriteLine("In ContinueWith delegate");
```
stephentoub: Please remove the WriteLine.

stephentoub:
@mellinoe, thanks.  There's an enormous amount of code here.  At some point I'd like to review it all, but for now I just did a few spot checks to look for any glaring errors.  I'm fine with you checking in once CI consistently passes.  Note that the current failure that shows on Linux is also reproduceable on Windows... I was able to trigger it a handful of times by running that one test repeatedly (I made it a [Theory] that took a single int parameter and used a MemberData that yielded a thousand integers).

mellinoe:
Thanks for the first pass, Stephen. I will check out the flaky test tomorrow and see what is up. That was one of the last files I included so I didn't run those tests as many times as the others.

I'll address the low-hanging fruit you mentioned, like the tests that can obviously be easily converted to [Theory]'s and the bad Assertion styles. I also plan on doing a more thorough cleanup / rewrite of some of these tests next week when I'm TestCop. There's a few obvious things I want to at least accomplish at that time:

* Convert even more test cases to Theory's when useful
* Share more code between test cases: There is a lot of code that is duplicated or only slightly different between test classes
* Improve test speed: The tests are running usually in 5-8 seconds on my machine, but sometimes they will take as long as 30 seconds for some reason. It's probably due to what I described above (test cases leaving tasks running for a long time). It's hard to pinpoint the cause of this.
* Check out the code coverage numbers and see where to improve coverage

src/System.Threading.Tasks/tests/Task/TaskRtTests.cs
```diff
@@ -0,0 +1,1144 @@
...
+                        Assert.False(tasks[i] == tasks[j], "TaskRtTests.RunFromResult:    > FAILED: " + i + " and " + j + " created tasks should not be equal (ref)");
+                    }
+                }
+
+                // Make sure they all have the correct results
+                for (int i = 0; i < tasks.Length; i++)
+                    Assert.True(tasks[i].Result == results[i], "TaskRtTests.RunFromResult:    > FAILED: Task " + i + " had the wrong result (ref)");
+            }
+
+            // Uncomment if/when we want Task.FromException
```
akoeplinger: I suppose we do want them now? :wink:

mellinoe: Good catch, added back and fixed u

stephentoub:
Great that you'll be spending some more time on these next week.  In the meantime, as an initial port, LGTM.

src/System.Threading.Tasks/tests/Task/TaskRtTests.cs
```diff
@@ -0,0 +1,1135 @@
...
+              () => future.ContinueWith((t, s) => { }, null, CancellationToken.None, TaskContinuationOptions.None, null));
+
+            Assert.ThrowsAsync<ArgumentNullException>(
+               () => future.ContinueWith<int>((Func<Task<int>, Object, int>)null, null, CancellationToken.None));
+            Assert.ThrowsAsync<ArgumentNullException>(
+              () => future.ContinueWith<int>((Func<Task<int>, Object, int>)null, null, TaskContinuationOptions.None));
+            Assert.ThrowsAsync<ArgumentNullException>(
+              () => future.ContinueWith<int>((Func<Task<int>, Object, int>)null, null, CancellationToken.None, TaskContinuationOptions.None, TaskScheduler.Default));
+            Assert.ThrowsAsync<ArgumentNullException>(
+              () => future.ContinueWith<int>((t, s) => 2, null, CancellationToken.None, TaskContinuationOptions.None, null));
```
jnm2: Ran across these in master. None of these are awaited. Looks like a bug both now and at the time.
https://github.com/dotnet/corefx/blob/3efd19931a28e0d978fb273d20d857f597fca93c/src/System.Threading.Tasks/tests/project.json#L15
https://github.com/xunit/xunit/blob/2.1-beta3/src/xunit.assert/Asserts/ExceptionAsserts.cs#L50

stephentoub: Ugh. Yeah, that's busted.  I'll fix it.  Thanks for pointing it out.

PR #2776, opened 2015-08-12
Make RSA tests' AssertKeyEquals tolerate D changing.

### LINKED ISSUE #2741: [RSA Tests] Update RSAParameters validation to ensure D is valid, as opposed to roundtrips
RSA `D` values can be generated in one of two ways:
* `D = ModInv(E, (p - 1) * (q - 1))` (totient/psi)
* `D = ModInv(E, LCM(p - 1, q - 1))` (lambda)

Windows CAPI and CNG both generate `D` using the lambda method (which matches the PKCS#1 algorithm).  OpenSSL seems to generate it using the totient method.  Both approaches will produce the same `DP`, `DQ`, and `InverseQ` values.

OpenSSL and Windows CAPI both will roundtrip an arbitrary `D` value across import/export.  Windows CNG always exports the lambda-method `D`.

Since both values for `D` will produce functioning answers under the RSA algorithm, the two expanded keys can be said to be equivalent.  The tests therefore should only ensure that the `D` value produces a valid computation of `(D * E) % lambda == 0x01`; rather than that it was the same as the input value.

As a part of that change, ImportExportUnusual.cs should be merged back into ImportExport.cs; since the CNG tests should pass under that constraint.

bartonjs:
Windows CNG always exports the D value they would have generated (using the lambda modulus), rather than the value that was given to them.  This is particularly relevant for OpenSSL-generated keys, because they produce a D value using the totient modulus.

Now the test accepts equal valued arrays (like it did), or if they're not SequenceEqual the new D value will be tested for legality; and a legal value will be accepted even if it's different.

Fixes #2741.


src/System.Security.Cryptography.Cng/tests/project.lock.json
```diff
@@ -30,23 +33,26 @@
           "lib/DNXCore50/System.Console.dll": {}
         }
       },
-      "System.Diagnostics.Debug/4.0.10-beta-22703": {
+      "System.Diagnostics.Debug/4.0.10-beta-23127": {
```
bartonjs: I double checked, all the beta upgrades in this lock file are bringing it in line with the rest of the modules.  This addition and the last upgrade must have crossed in the night.

bartonjs:
cc: @AtsushiKan

src/Common/tests/Cryptography/AlgorithmImplementations/RSA/ImportExport.cs
```diff
@@ -189,12 +242,28 @@ internal static void AssertKeyEquals(ref RSAParameters expected, ref RSAParamete
...
+            {
+                Assert.Null(actual.D);
+            }
+            else
+            {
+                Assert.NotNull(actual.D);
+
+                // If the value matched expected, take that as valid and shortcut the math.
+                // If it didn't, we'll test that the value is at least legal.
+                if (!expected.D.SequenceEqual(actual.D))
```
bartonjs: I had a Console.WriteLine in `VerifyDValue` during development, so I know that the CNG tests got in there exactly twice, which is what I'd expect.

AtsushiKan:
LGTM

PR #2781, opened 2015-08-13
Fix a few more issues in Timer tests

stephentoub:
- Several tests were disposing the Timer before the callback with Asserts could be invoked, such that the tests were nops rather than validating what they were trying to validate.
- Several tests were modifying shared state without synchronization; very unlikely to have caused a problem, but better safe than sorry.
- One test had a race condition that could have resulted in a NullReferenceException, with a callback that attempted to access a captured local before it was initialized.
- Several tests were susceptible to the Timer being tested getting finalized in release builds, which would cause the test to fail

cc: @sokket


stephentoub:
@dotnet-bot test this please

sokket:
LGTM

PR #2789, opened 2015-08-13
Avoid intermediate collection/enumerator allocations in X509 IStorePal

justinvp:
All callers of `IStorePal.Certificates` and `IStorePal.Find` enumerate the resulting `IEnumerable<X509Certificate2>` and add each item to a `X509Certificate2Collection`. This PR changes `IStorePal` to instead accept a `X509Certificate2Collection` to add items to, avoiding any intermediate collection and enumerator allocations.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/StorePal.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/StorePal.cs
@@ -21,19 +21,15 @@ namespace Internal.Cryptography.Pal
 {
     internal sealed partial class StorePal : IDisposable, IStorePal
     {
-        public IEnumerable<X509Certificate2> Certificates
+        public void CopyTo(X509Certificate2Collection collection)
         {
-            get
-            {
-                LowLevelListWithIList<X509Certificate2> certificates = new LowLevelListWithIList<X509Certificate2>();
+            Debug.Assert(collection != null);
 
-                SafeCertContextHandle pCertContext = null;
-                while (Interop.crypt32.CertEnumCertificatesInStore(_certStore, ref pCertContext))
-                {
-                    X509Certificate2 cert = new X509Certificate2(pCertContext.DangerousGetHandle());
-                    certificates.Add(cert);
-                }
-                return certificates;
+            SafeCertContextHandle pCertContext = null;
+            while (Interop.crypt32.CertEnumCertificatesInStore(_certStore, ref pCertContext))
+            {
+                X509Certificate2 cert = new X509Certificate2(pCertContext.DangerousGetHandle());
+                collection.Add(cert);
             }
         }
 
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/IStorePal.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/IStorePal.cs
@@ -9,9 +9,9 @@ namespace Internal.Cryptography.Pal
 {
     internal interface IStorePal : IDisposable
     {
-        IEnumerable<X509Certificate2> Find(X509FindType findType, Object findValue, bool validOnly);
+        void FindAndCopyTo(X509FindType findType, object findValue, bool validOnly, X509Certificate2Collection collection);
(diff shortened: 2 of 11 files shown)


justinvp:
cc @bartonjs

bartonjs:
You changed some `object`s to `Object`, but the other direction is preferred.  If you feel the burning desire to fix that, that's great; otherwise it'll get cleaned up eventually.

Aside from the casing on the word object, LGTM.

Thanks :smile:.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/StorePal.cs
```diff
@@ -21,19 +21,15 @@ namespace Internal.Cryptography.Pal
...
-                {
-                    X509Certificate2 cert = new X509Certificate2(pCertContext.DangerousGetHandle());
-                    certificates.Add(cert);
-                }
-                return certificates;
+            SafeCertContextHandle pCertContext = null;
+            while (Interop.crypt32.CertEnumCertificatesInStore(_certStore, ref pCertContext))
+            {
+                X509Certificate2 cert = new X509Certificate2(pCertContext.DangerousGetHandle());
+                collection.Add(cert);
```
stephentoub: The only danger I see here is if it's possible for an exception to get thrown such that we would have added only some of the certs to the output collection and then if there's code that tries to recover. I don't know if that's a real concern.

stephentoub: (And indeed that appears to be what's happening, so ignore this :smile:)

justinvp:
> You changed some objects to Object

Ah, it was an unintended copy/paste change. I'll fix. (In the `IStorePal` definition it was `Object`, which I just left to be cleaned up later. Then just copy/pasted that definition into the implementations, but in some of the implementations it was the correct `object`.)

Also, will fix the build issue on Unix. (I had been verifying Linux builds before submitting, but I added some Debug.Asserts at the last minute and forgot to double-check Linux, which isn't compiling.

stephentoub:
Other than the casing and the missing import for System.Diagnostics, LGTM2.

KrzysztofCwalina:
Looks good to me

stephentoub:
@dotnet-bot test this please

justinvp:
@dotnet-bot test this please

stephentoub:
The tests are failing due to a segfault on Linux:
```
Starting:    System.Security.Cryptography.X509Certificates.Tests
./run-test.sh: line 156: 31144 Segmentation fault      (core dumped) ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCategory
```

justinvp:
Thanks for finding that, Steve. I think I know what the issue is: a problem with `X509CertificateCollection.CopyTo`. Looking into it.

justinvp:
Yeah, `CopyTo` is broken. Right now, it's implemented like this:

```c#
public partial class X509CertificateCollection : ICollection, IEnumerable, IList
{
    private LowLevelListWithIList<Object> _list;
    
    internal IList List
    {
        get { return this; }
    }
    
    public void CopyTo(X509Certificate[] array, int index)
    {
        List.CopyTo(array, index);
    }
    
    void ICollection.CopyTo(Array array, int index)
    {
        List.CopyTo(array, index);
    }

    ...
}
```

So it'll just end up recursively calling itself.

`LowLevelListWithIList<T>` doesn't have an implementation of `ICollection.CopyTo`. Before I go add an implementation to it to fix this, is there any reason why this can't be switched to `List<T>`? This library doesn't have an explicit dependency on `System.Collections`, but its dependencies (e.g. `System.IO.FileSystem`) already do depend on `System.Collections`.

bartonjs:
@justinvp Switching to List should be fine.  I've certainly used it a fair bit recently within that library.  If there's a reason not to, I don't know what it is.

justinvp:
I added a separate commit that fixes `X509CertificateCollection.CopyTo` (includes tests). CI passes now.

stephentoub:
Thanks for the fix.  LGTM.

PR #2790, opened 2015-08-13
Implement full CNG hash support for RSACng.SignData

AtsushiKan:
- This unifies the CNG hash support for Algorithms and Cng.


AtsushiKan:
cc @bartonjs

src/Common/src/Internal/Cryptography/BCryptAlgorithmCache.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Diagnostics;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+
+using Microsoft.Win32.SafeHandles;
+using NTSTATUS = Interop.BCrypt.NTSTATUS;
+using BCryptOpenAlgorithmProviderFlags = Interop.BCrypt.BCryptOpenAlgorithmProviderFlags;
+
+namespace Internal.Cryptography
```
bartonjs: I'd have put this under Common\src\Interop\Windows\BCrypt, personally.  But I'll defer to @stephentoub on the pathing.

src/Common/src/Internal/Cryptography/BCryptAlgorithmCache.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (ntStatus != NTSTATUS.STATUS_SUCCESS)
+                throw Interop.BCrypt.CreateCryptographicException(ntStatus);
+
+            Entry[] newCache = new Entry[cache.Length + 1];
+            Entry newEntry = new Entry(hashAlgorithmId, flags, safeBCryptAlgorithmHandle);
+            Array.Copy(cache, newCache, cache.Length);
+
+            // Check again in case some other thread added the hash we need just now. It's fine if he did, but we don't want to create
+            // a duplicate entry in the new cache as that'll slow down all future lookups.
+            foreach (Entry entry in newCache)
```
bartonjs: I don't see how this could hit.  You're looking at the same contents as the previous foreach due to the method-local reference and atomic replacement.

src/Common/src/Internal/Cryptography/BCryptAlgorithmCache.cs
stephentoub: That'd be my preference as well.  Eventually we want all of this interop stuff to end up in src\Common\src\Interop\Windows\libraryname\etc.

src/Common/src/Internal/Cryptography/BCryptAlgorithmCache.cs
AtsushiKan: Yep.

src/Common/src/Microsoft/Win32/SafeHandles/SafeBCryptAlgorithmHandle.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+using NTSTATUS = Interop.BCrypt.NTSTATUS;
+
+namespace Microsoft.Win32.SafeHandles
+{
+    internal sealed class SafeBCryptAlgorithmHandle : SafeBCryptHandle
```
bartonjs: SafeBCryptAlgorithmHandle.Windows.cs; if I understand the naming convention in this directory :).

AtsushiKan: SafeLocalAllocHandle.cs suggests otherwise.

bartonjs:
Deferring to Stephen on all of the pathing things (in particular, I'm not sure where I'd put the PAL HashAlgorithmProvider base class).

Aside from the one suspect foreach the semantics of the change look good to me.

src/Common/src/Internal/Cryptography/HashProviderCng.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+        }
+
+        public sealed override byte[] FinalizeHashAndReset()
+        {
+            byte[] hash = new byte[_hashSize];
+            NTSTATUS ntStatus = Interop.BCrypt.BCryptFinishHash(_hHash, hash, hash.Length, 0);
+            if (ntStatus != NTSTATUS.STATUS_SUCCESS)
+                throw Interop.BCrypt.CreateCryptographicException(ntStatus);
```
stephentoub: Should we be resetting the hash object even if this fails?

src/Common/src/Internal/Cryptography/HashProviderCng.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            ResetHashObject();
+            return hash;
+        }
+
+        public sealed override void Dispose(bool disposing)
+        {
+            if (disposing)
```
stephentoub: I know you have a comment to this effectin GetCachedBCryptAlgorihmHandle, but I think it'd be worth adding a comment here that we explicitly don't dispose of _hAlgorithm because it's a cached object used by other instances.

src/Common/src/Internal/Cryptography/BCryptAlgorithmCache.cs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: It looks like your most recent update added a copy of this file but didn't delete this one.

stephentoub:
A few comments but otherwise LGTM.

src/Common/src/Internal/Cryptography/HashProviderCng.cs
AtsushiKan: Not sure there’s a good answer for this one. On Win8 and forward, ResetHash() is a nop, so it boils down to do apps who allocate a HashAlgorithm object have a reasonable expectation of continuing to use that object if a ComputeHash() throws an exception? I don’t know that we have a good answer to this or that I want to try to figure that out now, given that this particular commit isn’t changing this behavior.

PR #2794, opened 2015-08-13
Fixed test that was leaving temp directories in root folder

ianhays:
One of the Extended Path FileSystem tests was leaving temporary test directories in the root folder of the running directory. This small change modified the tests to instead treat the user's temp directory as the root.


ianhays:
cc: @JeremyKuhne @stephentoub

stephentoub:
LGTM

JeremyKuhne:
LGTM

PR #2796, opened 2015-08-13
Throw PlatformNotSupportedException for named Semaphore on Unix

stephentoub:
libcoreclr's synchronization primitives implementation currently supports names, but those names have process-wide rather than system-wide scope.  This is very dangerous for most code that would want names, as they're typically used for cross-process synchronization, and as such the current behavior could lead to bad race conditions difficult to diagnose.  Until a better solution is available, we will throw PlatformNotSupportedException when trying to create such named primitives.

Semaphore's code is in the corefx repo, so this commit both updates the semaphore implementation and updates the tests for all of the named primitives.  A separate commit in coreclr will address the primitives implemented there.  A few of the modified tests that are expected PlatformNotSupportedException are currently marked as [ActiveIssue], as they will fail until the coreclr changes are merged.


stephentoub:
Corresponding PR in coreclr: https://github.com/dotnet/coreclr/pull/1387

stephentoub:
cc: @kouvel, @jkotas, @ellismg

weshaggard:
LGTM

stephentoub:
Thanks, Wes.

Issue #1533, opened 2015-04-27
FileStream constructor should allow bufferSize = 0 to indicate non-buffered stream


WalkingCat:
FileStream.Init will throw ArgumentOutOfRangeException if (bufferSize <-0), IMHO it should allow bufferSize = 0 value, indicating this FileStream object nerver uses buffer on this level, provides more flexibility and control for app developer, current workaround is set bufferSize to 1, which is ugly.

stephentoub:
> current workaround is set bufferSize to 1, which is ugly

Can you clarify why that's ugly?

WalkingCat:
because it still has possibility to allocate memory for buffer array when Read/Write actually called with argument numBytes = 1, its not guaranteed non-buffered.

stephentoub:
Right.  I was just clarifying that your concern is purely around a byte[1] getting allocated per FileStream; there's no other difference you're concerned about, right?

WalkingCat:
its not just about memory allocation, buffering can affect the behavior/logic, like at what moment the native IO calls are actually called etc, when I say I want a non-buffered stream, I expect exactly non-buffered behavior, otherwise its suprising.

stephentoub:
I agree that if that were the case, it would be troubling to claim that it never uses the buffer.  I'm missing from the code where that's happening, though.  If the buffer size is 1, then any attempt to write any amount of data will skip the buffer, due to checks that the amount of data is greater than or equal to the buffer size.  Is there something in the code I'm missing where that's not the case?  Can you help me by pointing to a specific example?  It may very well be there, I'm just missing it.

stephentoub:
Actually, on closer look I do see that happening in the async code path here https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs#L1384.  I don't see it happening in the sync code path.

WalkingCat:
yes I was talking about WriteInternalAsync, it looks like sync methods handle buffering subtly differently, which is just weird implementation details and should not be relied on anyway.

stephentoub:
Right.  So my top-level point is that if/when https://github.com/dotnet/corefx/issues/1530 is fixed such that the async-buffering behavior matches the sync-buffering behavior, then this request is really about   a slight usability benefit (being able to say 0 instead of 1 but to effectively mean the same thing) that also avoids a byte[1] array allocation per FileStream, correct?  I'm not trying to downplay the benefits, just quantify them.

[...]

PR #2802, opened 2015-08-14
Handle internal overflow of the write buffer in WriteAsync by introdu…

### LINKED ISSUE #1531: Win32FileStream buffer-full flushes are also done in a blocking fashion
Win32FileStream flushes its buffer synchronously from the caller's perspective, even when using WriteAsync.

tymlipari:
Handle internal overflow of the write buffer in WriteAsync asynchronously by introducing new cases:

Introduces new cases to handle overflow of the internal write buffer rather than synchronously flushing the buffer. In the best case, it'll create a new buffer and write the original one to disk (single write). In the worst case, it will chain the incoming write to the original flush operation (double write). All writes are now done asynchronously. Additionally, adds a new unit test to ensure the different buffering cases are being covered.

Fix #1531

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -562,6 +562,16 @@ namespace System.IO
                 {
                     writeTask.GetAwaiter().GetResult();
                 }
+
+                // If not completing synchronously, ensure the async flush operation
+                // is tracked in case of incoming WriteAsync calls.
+                else
+                {
+                    if (ActiveBufferOperation())
+                        _activeBufferOperation = Task.WhenAll(_activeBufferOperation, writeTask);
+                    else
+                        _activeBufferOperation = writeTask;
+                }
             }
             else
             {
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -1344,7 +1344,7 @@ namespace System.IO
 
             // Case 1 - no active buffer op & fits in remaining size
             int remainingBuffer = _bufferSize - _writePos;
-            if (!ActiveBufferOperation() && numBytes <= remainingBuffer)
+            if (!HasActiveBufferOperation && numBytes < remainingBuffer)
             {
                 if (_buffer == null) _buffer = new byte[_bufferSize];
 
@@ -1354,7 +1354,7 @@ namespace System.IO
             }
 
             // Case 2 - active buffer op & fits in new buffer
-            if (ActiveBufferOperation() && numBytes <= _bufferSize)
+            if (HasActiveBufferOperation && numBytes < _bufferSize)
             {
                 _buffer = new byte[_bufferSize];
 
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -520,6 +520,19 @@ namespace System.IO
             _readLen = 0;
         }
 
+        // Returns a task that flushes the internal write buffer
+        private Task FlushWriteAsync(CancellationToken cancellationToken)
+        {
+            Debug.Assert(_isAsync);
+            Debug.Assert(_readPos == 0 && _readLen == 0, "FileStream: Read buffer must be empty in FlushWriteAsync!");
+
+            Task flushTask = WriteInternalCoreAsync(_buffer, 0, _writePos, CancellationToken.None);
+
+            _writePos = 0;
+
+            return flushTask;
+        }
+
         // Writes are buffered.  Anytime the buffer fills up 
         // (_writePos + delta > _bufferSize) or the buffer switches to reading
         // and there is left over data (_writePos > 0), this function must be called.
@@ -529,7 +542,7 @@ namespace System.IO
 
             if (_isAsync)
             {
-                Task writeTask = WriteInternalCoreAsync(_buffer, 0, _writePos, CancellationToken.None);
+                Task writeTask = FlushWriteAsync(CancellationToken.None);
                 // With our Whidbey async IO & overlapped support for AD unloads,
                 // we don't strictly need to block here to release resources
                 // since that support takes care of the pinning & freeing the 
@@ -1277,8 +1290,9 @@ namespace System.IO
                 // pipes.   
                 Debug.Assert(_readPos == 0 && _readLen == 0, "Win32FileStream must not have buffered data here!  Pipes should be unidirectional.");
 
-                if (_writePos > 0)
-                    FlushWrite(false);
+                if(_writePos > 0)
+                    return FlushWriteAsync(cancellationToken)
+                        .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
 
                 return WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
             }
@@ -1291,19 +1305,45 @@ namespace System.IO
                 _readLen = 0;
             }
 
-            int n = _bufferSize - _writePos;
-            if (numBytes <= n)
+            int remainingBuffer = _bufferSize - _writePos;
+
+            // if the incoming write fits in the remaining buffer, treat
+            // as synchronous since memcopy is so fast
+            if (numBytes <= remainingBuffer)
             {
                 if (_writePos == 0) _buffer = new byte[_bufferSize];
                 Buffer.BlockCopy(array, offset, _buffer, _writePos, numBytes);
                 _writePos += numBytes;
-                
+
                 return Task.CompletedTask;
             }
 
-            if (_writePos > 0)
-                FlushWrite(false);
+            // if the incoming write fits in the current buffer + next buffer,
+            // only perform a single write to flush the existing buffer and fill
+            // the next buffer with the remainder
+            if (numBytes <= (_bufferSize + remainingBuffer))
+            {
+                if (_writePos == 0) _buffer = new byte[_bufferSize];
 
+                Buffer.BlockCopy(array, offset, _buffer, _writePos, remainingBuffer);
+
+                byte[] oldBuffer = _buffer;
+                _buffer = new byte[_bufferSize];
+
+                _writePos = (numBytes - remainingBuffer);
+                Buffer.BlockCopy(array, offset + remainingBuffer, _buffer, 0, _writePos);
+
+                return WriteInternalCoreAsync(oldBuffer, 0, _bufferSize, cancellationToken);
+            }
+
+            // If there's data in the existing buffer, and the incoming write would overflow it and
+            // the subsequent buffer, chain the flush operation with the subsequent write
+            else if(_writePos > 0)
+                return FlushWriteAsync(cancellationToken)
+                    .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
+
+            // Otherwise, there's no buffered data, and the incoming write is too big to fit
+            // in the buffer, so write it directly to disk
             return WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
         }
 
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -526,10 +526,11 @@ namespace System.IO
             Debug.Assert(_isAsync);
             Debug.Assert(_readPos == 0 && _readLen == 0, "FileStream: Read buffer must be empty in FlushWriteAsync!");
 
+            // If the buffer is already flushed, don't spin up the OS write
+            if (_writePos == 0) return Task.CompletedTask;
+
             Task flushTask = WriteInternalCoreAsync(_buffer, 0, _writePos, cancellationToken);
-
             _writePos = 0;
-
             return flushTask;
         }
 
@@ -1264,6 +1265,14 @@ namespace System.IO
             return result;
         }
 
+        private async Task FlushAndWriteAsync(byte[] array, int offset, int numBytes, CancellationToken cancellationToken)
+        {
+            Debug.Assert(_isAsync);
+
+            await FlushWriteAsync(cancellationToken);
+            await WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
+        }
+
         [System.Security.SecuritySafeCritical]  // auto-generated
         private Task WriteInternalAsync(byte[] array, int offset, int numBytes, CancellationToken cancellationToken)
         {
@@ -1290,9 +1299,8 @@ namespace System.IO
                 // pipes.   
                 Debug.Assert(_readPos == 0 && _readLen == 0, "Win32FileStream must not have buffered data here!  Pipes should be unidirectional.");
 
-                if(_writePos > 0)
-                    return FlushWriteAsync(cancellationToken)
-                        .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
+                if (_writePos > 0)
+                    return FlushAndWriteAsync(array, offset, numBytes, cancellationToken);
 
                 return WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
             }
@@ -1338,9 +1346,8 @@ namespace System.IO
 
             // If there's data in the existing buffer, and the incoming write would overflow it and
             // the subsequent buffer, chain the flush operation with the subsequent write
-            else if(_writePos > 0)
-                return FlushWriteAsync(cancellationToken)
-                    .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
+            else if (_writePos > 0)
+                return FlushAndWriteAsync(array, offset, numBytes, cancellationToken);
 
             // Otherwise, there's no buffered data, and the incoming write is too big to fit
             // in the buffer, so write it directly to disk
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
@@ -61,6 +61,7 @@ namespace System.IO
         private long _appendStart;// When appending, prevent overwriting file.
         
         private Task<int> _lastSynchronouslyCompletedTask = null;
+        private Task _activeBufferOperation = null;
 
         [System.Security.SecuritySafeCritical]
         public Win32FileStream(String path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options, FileStream parent) : base(parent)
@@ -1265,12 +1266,9 @@ namespace System.IO
             return result;
         }
 
-        private async Task FlushAndWriteAsync(byte[] array, int offset, int numBytes, CancellationToken cancellationToken)
+        private bool ActiveBufferOperation()
         {
-            Debug.Assert(_isAsync);
-
-            await FlushWriteAsync(cancellationToken);
-            await WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
+            return _activeBufferOperation != null && !_activeBufferOperation.IsCompleted;
         }
 
         [System.Security.SecuritySafeCritical]  // auto-generated
@@ -1299,13 +1297,20 @@ namespace System.IO
                 // pipes.   
                 Debug.Assert(_readPos == 0 && _readLen == 0, "Win32FileStream must not have buffered data here!  Pipes should be unidirectional.");
 
+                // If there's data in the write buffer, flush it before starting a new write
                 if (_writePos > 0)
-                    return FlushAndWriteAsync(array, offset, numBytes, cancellationToken);
+                {
+                    Task flushTask = FlushWriteAsync(cancellationToken);
+                    if (ActiveBufferOperation())
+                        _activeBufferOperation = Task.WhenAll(_activeBufferOperation, flushTask);
+                    else
+                        _activeBufferOperation = flushTask;
+                }
 
                 return WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
             }
 
-            // Handle buffering.
+            // Ensure the buffer is clear for writing
             if (_writePos == 0)
             {
                 if (_readPos < _readLen) FlushRead();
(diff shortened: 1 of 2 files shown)


src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -520,6 +520,19 @@ private void FlushRead()
             _readLen = 0;
         }
 
+        // Returns a task that flushes the internal write buffer
+        private Task FlushWriteAsync(CancellationToken cancellationToken)
+        {
+            Debug.Assert(_isAsync);
+            Debug.Assert(_readPos == 0 && _readLen == 0, "FileStream: Read buffer must be empty in FlushWriteAsync!");
+
+            Task flushTask = WriteInternalCoreAsync(_buffer, 0, _writePos, CancellationToken.None);
```
piotrpMSFT: Why are we not passing through the cancellationToken param?

tymlipari: Whoops, likely a copy-paste error. Fixed in updated commit.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1277,8 +1290,9 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
                 // pipes.   
                 Debug.Assert(_readPos == 0 && _readLen == 0, "Win32FileStream must not have buffered data here!  Pipes should be unidirectional.");
 
-                if (_writePos > 0)
-                    FlushWrite(false);
+                if(_writePos > 0)
+                    return FlushWriteAsync(cancellationToken)
+                        .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
```
stephentoub: Chaining like this is the right idea.  However, using ContinueWith as is done is going to be problematic.  ContinueWith executes the continuation regardless of whether the antecedent task completes successfully, due to cancellation, or due to faulting with an exception, and we'd only want to do the subsequent write if the flush completed successfully.  If it didn't, we want to return the status from the FlushWriteAsync task instead of that of executing the write.  The semantics of await are more appropriate here.  As such, you probably want a helper async method that awaits the flush and then awaits the write, and that helper can be invoked here.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,19 +1305,45 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
             }
 
-            if (_writePos > 0)
-                FlushWrite(false);
+            // if the incoming write fits in the current buffer + next buffer,
+            // only perform a single write to flush the existing buffer and fill
+            // the next buffer with the remainder
+            if (numBytes <= (_bufferSize + remainingBuffer))
+            {
+                if (_writePos == 0) _buffer = new byte[_bufferSize];
```
stephentoub: Hmm, I'm not sure why the existing code has this line above; seems fishy, in particular because we could already have a valid buffer, so I'm not sure why we're overwriting it.  @ericstj, do you know why the code above in the ```numBytes <= remainingBuffer``` block checks ```if (_writePos == 0)``` rather than ```if (_buffer == null)```.?

tymlipari: Wouldn't it be possible that a flush operation is potentially long running in the background when the new write comes in? In that case, _buffer wouldn't be null, but _writePos would be 0 and a new buffer should be allocated since the existing one is being referenced by the flush.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,19 +1305,45 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+
+                Buffer.BlockCopy(array, offset, _buffer, _writePos, remainingBuffer);
+
+                byte[] oldBuffer = _buffer;
+                _buffer = new byte[_bufferSize];
```
stephentoub: So does this mean that every time we hit this path of going past the end of the current buffer, we're going to allocate a new one?  That seems quite costly.

This would seem to put us between and a rock and a hard place.  The caller of WriteAsync needs to be able to use the byte[] they provide when the call returns, so we can't just asynchronously flush the buffer and then asynchronously write the data they provide, because as soon as we start the asynchronous flush, we'll return to them, but an asynchronous operation using their buffer will still be active.  An alternative is what the code already employs: do the flush synchronously, but that then mostly negates the value of using WriteAsync in the first place (any writes of a size less than the buffer will end up either synchrously writing to the buffer or synchronously flushing), and that's what this PR is trying to address. Ugh.

I wonder if some small buffer pool is warranted, e.g. we allow only up to N buffer allocations, which are pooled, and if we try to get a buffer from the pool and we've already allocated N, then we fall back to synchronous flushing.

Or maybe in the case of overflowing the buffer, we support growing the buffer used to be the size of what's already cached plus the size of whatever the developer supplied (we could still do flushes when we hit the original _bufferSize, but we'd have enough room to store all of the user supplied data).  In degenerate situations, the buffer could get extremely large, but maybe for such cases we set of a threshold and do synchronous flushes beyond that.

@ericstj, you've probably thought about this before...?

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,19 +1305,45 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                Buffer.BlockCopy(array, offset + remainingBuffer, _buffer, 0, _writePos);
+
+                return WriteInternalCoreAsync(oldBuffer, 0, _bufferSize, cancellationToken);
+            }
+
+            // If there's data in the existing buffer, and the incoming write would overflow it and
+            // the subsequent buffer, chain the flush operation with the subsequent write
+            else if(_writePos > 0)
+                return FlushWriteAsync(cancellationToken)
+                    .ContinueWith((_) => WriteInternalCoreAsync(array, offset, numBytes, cancellationToken));
```
stephentoub: This is unfortunately going to have the problems mentioned in my previous: the supplied array could be overwritten by the caller time the flush completes.  This is a harder problem to solve than I'd originally thought through.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: It depends on what guarantees we make about using the async APIs on FileStream.  If we say that it's valid to do:
```C#
Task t1 = fileStream.WriteAsync(data, 0, data.Length);
Task t2 = fileStream.WriteAsync(data, 0, data.Length);
await Task.WhenAny(t1, t2);
```
then, yeah, that would explain it.  Many stream say that's not valid, and either throw if you try to do that or serialize the calls rather than try to overlap them.  @ericstj, what guarantees do we make about FileStream in this regard?

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: (That said, there are streams in the framework that say you're not allowed to reuse the same buffer you provided until the async operation has completed... including the base Stream class, so maybe FileStream does the same thing?)

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: Actually, I think I may be making this more complicated than it needs to be (Friday afternoon brain):
- If the write is truly asynchronous, the caller can't reuse the buffer until the Task is completed.  So in general they can't assume the buffer is reusable until then.
- Given that, worst case we can do something akin to what you've done below, where you just hold onto the user's buffer until it's time to actually write it out.  This is essentially what the base Stream does when there are other async operations in progress, chaining them all together.

As such, what you have is probably fine, except that I'm not sure it's worth the extra buffer allocation for this case.  What if we just said:
- When the data fits entirely in the current buffer, write it to the buffer and return synchronously.
- When it overflows the current buffer, write out the current buffer (if there is one) and then write out the user's buffer.

Essentially I think that's exactly what you have, except just deleting this additional case you added.

Thoughts?

tymlipari: That's totally fine with me, I added that case to try and avoid a second write operation, and figured the allocation is fine since it's already reallocated in the synchronous case above when the buffer is empty.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
tymlipari: Given that the design above results in potentially two OS writes per call to WriteAsync, supporting what you've written above could result in confusing output to consumers. Even if there's a valid case for supporting non-deterministic writing of the inputs, each WriteAsync doesn't result in an atomic OS write, so the data could be interleaved in the output stream.

And if it's fine to require that you can't have multiple active async writes to the stream, we can change the condition to a null check which will result in less allocations. Right now when the buffer is empty, it'll still allocate a new buffer on the incoming write despite _buffer not being in use.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: That's true. Let me think about this a bit over the weekend. This would all get much simpler if we said that all async operations are serialized and you can't do synchronous operations while an async operation is in flight. Then we shouldn't need any additional buffer allocations.

stephentoub:
I thought more about the whole buffering issue.

Seems like we have a few goals here:
A. Minimize the blocking employed, e.g. not synchronously flushing during an asynchronous write
B. Minimize the allocation employed, e.g. not creating a new buffer every time we flush

We also need to maintain existing behaviors as much as possible.  For example, in going over the code, it’s clear that it goes out of its way to maintain the current Position, so that multiple asynchronous writes can be issued one after the other to execute concurrently, such that the resulting sequence will end up being correct. (In other words, we don't need to worry about multiple threads concurrently accessing the FileStream, but we do need to worry about the same thread issuing multiple async operations to potentially run concurrently.)

However, the common case is non-overlapped calls to th epublic XxAsync methods, e.g. the common pattern is ```await fileStream.XxAsync(…)```, such that no additional operations are issued on the FileStream instance until the previous operation completes.  That’s what we should be optimizing for, while still supporting the ```Task t1 = fileStream.XxAsync(); Task t2 = fileStream.XxAsync(); await Task.WhenAll(t1, t2);``` case.

So, here’s my thinking on what we do:
1. We keep track of a single Task field on the FileStream used to determine whether any async operation is still in progress that may involve the internal buffer.  Any time an async flush  operation is issued, this field gets updated:
- If the field is null or if the Task in the field has completed (doesn't matter if successfully or not), we simply store the created task into the field.
- If the Task in the field is not completed, we store “Task.WhenAll(_field, newTask)” into the field.
At least for now, this serves purely to help with buffer management: if we know there’s no asynchronous flush in progress, then we don’t need to allocate a new internal buffer.
2. When we WriteAsync:
- If the data fits in the remaining space in the internal buffer, and if f the tracking task has completed, we simply store it into the buffer synchronously, as we do today.  This is very efficient.  No extra allocations, no extra async ops.
- If the data fits in the remaining space in the internal buffer, but the tracking task has not yet completed, there’s a flush async operation still in progress, which means we can’t touch the internal buffer.  So we allocate a new buffer to overwrite the previous one and store the data into the buffer.  This does involve an extra buffer allocation, but it’s the uncommon case of scheduling an async write when there's a previously issued async flush still in progress.
- If the data doesn’t fit in the remaining space in the internal buffer and the buffer is empty, the data being written is larger than the whole internal buffer, and we simply do the asynchronous write, as we do today.
- If the data doesn't fit in the remaining space in the internal buffer and the buffer has data in it, we issue the asynchronous flush to empty the buffer.  Once we have that flush’s task (and we know the Position has been updated accordingly), assuming it didn’t synchronously fail, we issue the internal WriteAsync for the user provided data.  These operations are now running concurrently, but doing it this way (rather than using a continuation off of the flush task) means that when returning from this write operation, the Position will have been updated accordingly.  The only difficulty here is that we now have two tasks.  We can return a Task.WhenAll that combines the two operations.  We don't need to proactively allocate a new buffer in this case, becuase the write operation isn't touching the buffer: we simply pass the user's data directly through.

The only really interesting case here is the last one, where there’s data in the buffer and not enough remaining space for the data being written.  As I see it, there are only three possible options for how to handle this:
1. Synchronously flush the buffer, then do the write asynchronously.  This is what the code does today.  This has the significant downside that all WriteAsync calls with data <= bufferLength essentially become Write calls, making everything synchronous.  Ugh!
2. Issue the flush asynchronously, and when it completes, issue the write asynchronously.  This would be ideal, except that when the WriteAsync call returns to its synchronous caller, the Position may not accurately reflect where the next operation should be writing data.  This breaks the previously discussed invariant.
3. Issue the flush asynchronously, and then regardless of whether it’s completed, issue the write asynchronously.  This has the downside that in the case where the flush fails asynchronously, we may have updated the position for the subsequent write in a case where today we wouldn’t have.

I’ve suggested doing (3) above because it only negatively impacts the case where flushes fail (which is rare), we can mitigate it for the case where the flush fails synchronously (in which case we can detect that and can avoid starting the write), and in the case of failure the data in the file probably can’t be considered coherent anyway.

@ericstj, @tymlipari, @ianhays, thoughts?

ianhays:
>1  Synchronously flush the buffer, then do the write asynchronously. This is what the code does today. This has the significant downside that all WriteAsync calls with data <= bufferLength essentially become Write calls, making everything synchronous.

Delayed synchronous operations at that. The initial WriteAsync with data <= bufferLength when the buffer is empty will complete synchronously after the memcpy into the buffer, yes? But then when a subsequent call comes in that requires a flush, the actual writing of that first WriteAsync is performed synchronously.

That an asynchronous write uses the buffer at all seems foreign to me, but to use it <i>synchronously</i> feels downright incorrect.

> 2 ...except that when the WriteAsync call returns to its synchronous caller, the Position may not accurately reflect where the next operation should be writing data.

Why is this the case here but not in (3)? Wouldn't Position be set to the location of the successful flush + length of the write?

>3 Issue the flush asynchronously, and then regardless of whether it’s completed, issue the write asynchronously. This has the downside that in the case where the flush fails asynchronously, we may have updated the position for the subsequent write in a case where today we wouldn’t have.

To clarify, would the scenario you mention look something like this:

buffer size 10 bytes
position = 0
WriteAsync(4 bytes @ position 0) - copies to buffer and returns completed Task; 
position = 4
WriteAsync(4 bytes @ position 4) - copies to buffer and returns completed Task
position = 8
WriteAsync(4 bytes @ position 8) - runs asynchronously alongside the flushing of the buffer filled with data from the first two WriteAsync calls. But if the Flush fails, then the position of this third write will be different here than if the first two Writes were already flushed before the call.

stephentoub:
> The initial WriteAsync with data <= bufferLength when the buffer is empty will complete synchronously after the memcpy into the buffer, yes? 

Correct.  Which I think is fine.  The buffer is an optimization that alleviates the need to make sys calls and hit the disk on every write operation.

> But then when a subsequent call comes in that requires a flush, the actual writing of that first WriteAsync is performed synchronously.

Correct.  That's the problem.  Flushing of the buffer is done by writing it out synchronously rather than asynchronously.

> Why is this the case here but not in (3)? Wouldn't Position be set to the location of the successful flush + length of the write?

It's the difference between (pseudo-code):
```
Task flushTask = WriteInternalAsync(_buffer, 0, _bufferedData);
Task writeTask = WriteInternalAsync(data, offset, count);
```
and
```
Task flushTask = WriteInternalAsync(_buffer, 0, _bufferedData);
Task writeTask = flushTask.ContinueWith(t => WriteinternalAsync(data, offset, count);
```
By the time the first block completes, Position has been updated to reflect the position after the second write.  But by the first block completes, flushTask may not have completed and thus we may not have actually called WriteInternalAsync the second time yet, so Position may not be correct.

> But if the Flush fails, then the position of this third write will be different here than if the first two Writes were already flushed before the call.

We'd issue two writes: one for the buffered 8 bytes and one for the caller's 4 bytes.  If the one for the buffered 8 bytes fails synchronously (e.g. something goes wrong attempting to initiate the overlapped I/O), we don't try to do the second write, the position remains unchanged, no problem.  If the one for the buffered 8 bytes succeeds, it's exactly as it is is today, no problem.  If the one for the buffered 8 bytes fails asynchronously, however, then we would have already have issued the second 4 byte write, so the position would not be 12 bytes ahead of where we started rather than the 8 bytes ahead of where we'd started if we'd done the flushing synchronously.

ianhays:
>But by the first block completes, flushTask may not have completed and thus we may not have actually called WriteInternalAsync the second time yet, so Position may not be correct.

I see, I was mixing up my sequence of events with (1). 

I'd argue that having a potentially incorrect Position is a big enough problem to seek a different solution - to not do so would make concurrent async IO operations far more difficult to be sure about. 

Between (1) and (3), I begrudgingly cast my vote for (3). The edge case in which it fails is narrow enough that I think the benefits gained from asynchronous writes with data.Length < bufferSize are worth the tradeoff.

tymlipari:
@stephentoub I've pushed a new commit that I think handles your cases above. Let me know what you think.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -547,6 +562,16 @@ private void FlushWrite(bool calledFromFinalizer)
...
                 }
+
+                // If not completing synchronously, ensure the async flush operation
+                // is tracked in case of incoming WriteAsync calls.
+                else
+                {
+                    if (ActiveBufferOperation())
+                        _activeBufferOperation = Task.WhenAll(_activeBufferOperation, writeTask);
+                    else
+                        _activeBufferOperation = writeTask;
```
stephentoub: Every time we call FlushWriteAsync, we do these four lines to update _activeBufferOperation.  Should they just be moved into FlushWriteAsync?  This is the only place where there's the potential for avoiding it, and it's an uncommon case (an async FileStream being finalized with data in the buffer) and a cheap enough operation that the overhead doesn't seem problematic even if it's unnecessary.  Worst case, it could be moved into its own helper rather than repeating it each time.

stephentoub: This is just personal preference, so feel free to leave it as-is, but I prefer to write such things as:
```C#
_activeBufferOperation = ActiveBufferOperation() ?
    Task.WhenAll(writeTask, _activeBufferOperation) :
    _writeTask;
```
That way it's clear that we're always assigning into the field, and the conditional just determines exactly what we're assigning.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1251,6 +1276,11 @@ public override int ReadByte()
             return result;
         }
 
+        private bool ActiveBufferOperation()
```
stephentoub: I think this would be better as a property, e.g.
```C#
private bool HasActiveBufferOperation
{
    get { return _activeBufferOperation != null && !_activeBufferOperation.IsCompleted; }
}
```

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+            // so the state of any async operation that touches the internal buffer must be tracked
+            // to ensure there are no data races at play.
+            //
+            //   1. No active async buffer op and numBytes fits in existing buffer:
+            //       Simply perform a memory copy operation and return synchronously.
+            //
+            //   2. Active async buffer op and numBytes <= _bufferSize
+            //       Since there's an active buffer operation, a new buffer needs to be allocated
+            //       to write to in order to ensure the existing buffer isn't modified mid-write.
+            //       The incoming write can then be copied into the new buffer.
```
stephentoub: It occurred to me as I was reading how you wrote this out that there's another option for this case: simply skip the buffer entirely and just issue the async write, rather than allocating a new buffer and copying into it.  In other words, we use the buffer if there's room and there's no active buffer operation, otherwise we issue the write directly.  Might be worth noting that, and we could experiment with the change separately to see which approach yields the best results.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+            //
+            //   3. No active async buffer and numBytes too large for buffer
+            //       If there's buffered data, issue a flush operation and then directly issue the
+            //       incoming write since it can't be buffered.
```
stephentoub: The way this is phrased makes me as a reader think that the write is only issued if there's buffered data.  I'd suggest rephrasing it just slightly:
```
If there's buffered data, issue a flush operation.  Then, regardless, issue the incoming write, since it can't be buffered.
```

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+            //       If there's buffered data, attach it to the existing async operation using Task.WhenAll
+            //       and then directly issue the incoming write.
+            //
+            //       Note: since _writePos is reset synchronously when calling FlushWriteAsync, it can be
+            //       assumed that if _writePos > 0 and there's an active async flush operation, the buffer
+            //       was reallocated since that operation was initiated.
+
+            // Case 1 - no active buffer op & fits in remaining size
+            int remainingBuffer = _bufferSize - _writePos;
+            if(!ActiveBufferOperation() && numBytes <= remainingBuffer)
```
stephentoub: Nit: space needed before the open paren

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                if (_buffer == null) _buffer = new byte[_bufferSize];
+
                 Buffer.BlockCopy(array, offset, _buffer, _writePos, numBytes);
                 _writePos += numBytes;
-                
                 return Task.CompletedTask;
             }
 
+            // Case 2 - active buffer op & fits in new buffer
+            else if (ActiveBufferOperation() && numBytes <= _bufferSize)
```
stephentoub: Consider getting rid of the ```else``` here for case 2: it's not needed because case 1 ends with a return, and case 3/4 don't have an else, so this (or those) are inconsistent.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+            }
+
+            // Case 3 & 4 - The incoming write can't be handled with a synchronous
+            // memcpy, so issue a write to disk, flushing if necessary
+
             if (_writePos > 0)
-                FlushWrite(false);
+            {
+                Task flushTask = FlushWriteAsync(cancellationToken);
+                if (ActiveBufferOperation())
```
stephentoub: I suggest adding a check here for if the flushTask completed synchronously and unsuccessfully: if it did, we can avoid issuing the write, e.g.:
```C#
Task flushTask = FlushWriteAsync(cancellationToken);
if (flushTask.IsCanceled || flushTask.IsFaulted)
    return flushTask;
```

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: A question for a subsequent investigation/change perhaps, as it would impact usage of the buffer elsewhere, but I wonder if we should actually be doing the comparison ```< remainingBuffer``` rather than ```<= remainingBuffer```? If the data being written fills the buffer, then we know we're going to flush the buffer out, and the question just becomes do we do it now (```< remainingBuffer```) or do we do it the next time we try to write anything (```<= remainingBuffer```).  Seems like if we know we're going to need to do it, we might as well kick it off now.  But maybe there are reasons not to?

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,18 +1328,63 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                    _activeBufferOperation = Task.WhenAll(_activeBufferOperation, flushTask);
+                else
+                    _activeBufferOperation = flushTask;
+            }
```
stephentoub: If there was buffered data that we flush, I think we need to return a Task.WhenAll that includes the flushTask and the WriteInternalCoreAsync task.  That way, if either the write or the flush (which is logically part of the write given how this is structured) fail, the caller will know that there was some sort of problem and the stream may not represent what they hoped it would.

src/System.IO.FileSystem/tests/FileStream/WriteAsync.cs
```diff
@@ -270,6 +270,30 @@ public void WriteAsyncBufferedCompletesSynchronously()
...
+
+                // Overflow bufferSize * 2
+                await fs.WriteAsync(TestBuffer, 0, 5);
+                Assert.True(fs.Length == 8);
+
+                // Overflow bufferSize * 2 with empty buffer
+                await fs.WriteAsync(TestBuffer, 0, 5);
+                Assert.True(fs.Length == 13);
+            }
+        }
```
stephentoub: It'd be good to have a test that tries to issue many writes/flushes to be run concurrently. The test you have and this should also verify that Position reflects the correct value immediately after the synchronous call to the XxAsync method returns.  Note that the OS may actually choose to issue writes synchronously, and I believe it'll do so if the write would require extending the size of the file, so it'd be good to have the test verify behavior both with a zero-length file to start and with a file presized for all data needed (this coudl be done as a theory with a Boolean argument that controls whether you start by setting the length).  I believe it'll also do writes synchronously for small writes, so it'd be good to also have a test that does large writes, like 100K at a time.

stephentoub:
This is looking good, I think.  Nice job, @tymlipari.

@ericstj, could you please weigh in on the approach and the change overall?

@ianhays, thoughts as well?

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
tymlipari: I sort of agree, I think that it should be `(< remainingBuffer)`. Right now it just delays the flush until the next incoming 

The only case I can see for it being `(<= remainingBuffer)` is for something like `_bufferSize == 1` where it'd effectively never be buffered in that case. Although I see in #1533, you comment that buffer size of 1 is conceptually non-buffered.

stephentoub: This might just be a bug, one that's been around for a long time.  Note that the synchronous Write path does skip the buffer in the == case: https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs#L947-L952

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1291,19 +1317,68 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                if (_buffer == null) _buffer = new byte[_bufferSize];
+
                 Buffer.BlockCopy(array, offset, _buffer, _writePos, numBytes);
                 _writePos += numBytes;
-                
                 return Task.CompletedTask;
             }
 
+            // Case 2 - active buffer op & fits in new buffer
+            if (ActiveBufferOperation() && numBytes <= _bufferSize)
```
stephentoub: These two calls to ActiveBufferOperation() should now be to HasActiveBufferOperation.

tymlipari: Yup, looks like I was a bit too eager to push the change. I'm running a quick build and test locally and then I'll push the fix.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
tymlipari: Changed to `>` in latest revision to be consistent with the synchronous Write.

stephentoub:
Thanks for doing this, @tymlipari.

This is a worthwhile change to take, especially since useAsync==true is the default now for FileStream in .NET Core.  At the moment, though, this area doesn't have enough test coverage for us to feel confident in merging it (I discovered at least one bug for which I already have a fix locally, specifically the case where multiple concurrent writes are issued while there's an outstanding flush).

I want to write a bunch of additional tests for this and do some more inspection of it before we incorporate it.  As such, I think the best thing to do at this point is for me to take this over.  Once I'm ready, I'll submit a new PR that includes your commits plus additional ones with tests and any necessary code fixes; I'll cc you on that PR.  We can leave this PR open as a placeholder until then.

Sound ok to you?  If you'd prefer an alternate plan, please do let me know.

stephentoub:
Opened #2929 to replace this (squashing and keeping the commits from this PR).  Thanks, @tymlipari!

PR #2803, opened 2015-08-14
Allow a SwitchExpression that has no SwitchCase

### LINKED ISSUE #1877: [System.Linq.Expressions] Allow Switch with empty cases list
Now we can not create an expression, that represents this code:
```cs
switch(x) {
default:
    return 0;
}
```
But c# allows this and this is can be very helpful - users do not needs a checks and  brunches in own code.

Problem with a next line: https://github.com/dotnet/corefx/blob/master/src/System.Linq.Expressions/src/System/Linq/Expressions/SwitchExpression.cs#L205

I have asked this question on StackOverflow a few months ago:  http://stackoverflow.com/questions/28360782/switch-without-cases-but-with-default-in-system-linq-expressions

Why this is not allowed? Can we change it?

JonHanna:
Fixes #1877

Currently, we do not allow for a switch expression with no SwitchCases. This
though can be useful if the SwitchCases are determined dynamically, in not
having to special-case the case where there are zero such SwitchCases.

However, merely changing this method to allow the zero-cases condition may have
compatibility issues with code acting on the expression tree assuming that this
condition cannot occur. Therefore we produce a switch with a case for the
default value of the expression matching that of the the default case.

By analogy to C#, we turn the equivalent to the (valid) C#:

    switch(expr)
    {
        default:
            doSomething();
            break;
    }

into:

    switch(expr)
    {
        case 0:
            doSomething();
            break;
        default:
            doSomething();
            break;
    }

but where the `0` is typed according to the type of `expr` as `default(T)` for whatever that type is.

If there is no `defaultBody` either we create a noop body. This will mean the
type of the expression is `void`.

If there is no `defaultBody` and the type is explicitly set to something other
than void, an `ArgumentException` is thrown.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
@@ -781,12 +781,14 @@ namespace System.Linq.Expressions
 
             switch (expressions.Length)
             {
+                case 0:
+                    expressions = new []{ Empty() };
+                    goto default;
                 case 2: return Block(expressions[0], expressions[1]);
                 case 3: return Block(expressions[0], expressions[1], expressions[2]);
                 case 4: return Block(expressions[0], expressions[1], expressions[2], expressions[3]);
                 case 5: return Block(expressions[0], expressions[1], expressions[2], expressions[3], expressions[4]);
                 default:
-                    ContractUtils.RequiresNotEmpty(expressions, "expressions");
                     RequiresCanRead(expressions, "expressions");
                     return new BlockN(expressions.Copy());
             }
@@ -857,8 +859,12 @@ namespace System.Linq.Expressions
         public static BlockExpression Block(IEnumerable<ParameterExpression> variables, IEnumerable<Expression> expressions)
         {
             ContractUtils.RequiresNotNull(expressions, "expressions");
+            var variableList = variables.ToReadOnly();
             var expressionList = expressions.ToReadOnly();
-            ContractUtils.RequiresNotEmpty(expressionList, "expressions");
+            if (variableList.Count != 0)
+                ContractUtils.RequiresNotEmpty(expressionList, "expressions");
+            else if (expressionList.Count == 0)
+                expressionList = new ReadOnlyCollection<Expression>(new []{ Empty() });
             RequiresCanRead(expressionList, "expressions");
 
             return Block(expressionList.Last().Type, variables, expressionList);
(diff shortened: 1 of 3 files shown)


stephentoub:
cc: @VSadov

hackcraft:
I also added allowing blocks with no expressions. It came up in the course of coding this that they aren't allowed (it had seemed an obvious approach to producing an nop expression), and an analogous change.

stephentoub:
@dotnet-bot test this please

DinoV:
I think the idea that being able to build up the switch statement dynamically and not worry about when you have zero cases makes sense.  So in generally I like the idea.  

Personally I’m -1 on the proposal to generate a case statement for default(T) – I think we should just relax the checks here.  IMHO this is going to cause more trouble than it’s worth.  When people go to re-write these trees they’re going to be different then the tree they expected to generate, and I suspect that will cause problems.  In general with the ETs we always sided on doing exactly what you told us to do.

Further by allowing this no one is actually going to break until someone starts generating a tree which includes these semantics – so it’s not a breaking change to allow this.  If someone runs into a consumer of their tree which cannot handle the empty switch statement they can always go and write a helper which does this transform for them (again, we previously encouraged helpers when you wanted specialized semantics).  Once we introduce this transform though we’ll never be able to remove it because then we may truly break someone who is now relying upon trees being constructed in a certain manner.

Of course, we should update our compiler and interpreter if it doesn't support the empty expression blocks.  But I briefly looked at them and it seems like they should be ok.

hackcraft:
Relaxing the check is definitely the preferable approach if it's not considered breaking. My suggestion is only for a workaround if that is considered breaking, and shouldn't be considered otherwise.

One thing that occurs to me now is that there are certainly current cases where people build expressions that then fails at the next step (just try doing a `Last()` on EF or a `First()` internal to a query on EF with some databases, and the expression that results will then fail). Since not all valid expressions can be used in all cases already, it seems to me that allowing a currently-invalid expression to be valid isn't as breaking a change as it would be otherwise.

VSadov:
After some discussing it seems we are coming to a conclusion that it is better to just allow switches with no cases.

- it seems that ET compiler and interpreter already can handle such constructs, although tests need to be added
- it is unlikely that it will be a backward compatibility issue since relying on something _not_ working would seem uncommon.

The proposal is to:
1) allow empty case list
2) when there are no cases and type is not provided, we should use the type of "default", if "default" is also not provided, we should use Void.

The reason why another approach with generating a single case same as default does not seem good is mostly that once it is implemented, someone can easily take dependency on the shape of the tree and we can never take it back. In this sense if we want to allow switches with no cases, it is better to just do it in as most direct and final way as possible.

JonHanna:
Well, I'd have rather have written a PR for such a change than for this workaround anyway, so cheerfully closing.
Did empty `BlockExpression`s come up in the same discussion?

PR #2805, opened 2015-08-14
Introducing checks for libcurl features

kapilash:
Runtime and load-time checks to ensure that the features user is relying
on are available in the libcurl library loaded

### DIFF
--- a/src/Common/src/Interop/Unix/libcurl/Interop.libcurl_types.cs
+++ b/src/Common/src/Interop/Unix/libcurl/Interop.libcurl_types.cs
@@ -125,6 +125,30 @@ internal static partial class Interop
             internal const int CURLMSG_DONE = 1;
         }
 
+        internal static partial class CURL_VERSION_Features
+        {
+            internal const int CURL_VERSION_IPV6         = (1<<0);
+            internal const int CURL_VERSION_KERBEROS4    = (1<<1);
+            internal const int CURL_VERSION_SSL          = (1<<2);
+            internal const int CURL_VERSION_LIBZ         = (1<<3);
+            internal const int CURL_VERSION_NTLM         = (1<<4);
+            internal const int CURL_VERSION_GSSNEGOTIATE = (1<<5);
+            internal const int CURL_VERSION_DEBUG        = (1<<6);
+            internal const int CURL_VERSION_ASYNCHDNS    = (1<<7);
+            internal const int CURL_VERSION_SPNEGO       = (1<<8);
+            internal const int CURL_VERSION_LARGEFILE    = (1<<9);
+            internal const int CURL_VERSION_IDN          = (1<<10);
+            internal const int CURL_VERSION_SSPI         = (1<<11);
+            internal const int CURL_VERSION_CONV         = (1<<12);
+            internal const int CURL_VERSION_CURLDEBUG    = (1<<13);
+            internal const int CURL_VERSION_TLSAUTH_SRP  = (1<<14);
+            internal const int CURL_VERSION_NTLM_WB      = (1<<15);
+            internal const int CURL_VERSION_HTTP2        = (1<<16);
+            internal const int CURL_VERSION_GSSAPI       = (1<<17);
+            internal const int CURL_VERSION_KERBEROS5    = (1<<18);
+            internal const int CURL_VERSION_UNIX_SOCKETS = (1<<19);
+        }
+
         // Type definition of CURLMsg from multi.h
         [StructLayout(LayoutKind.Explicit)]
         internal struct CURLMsg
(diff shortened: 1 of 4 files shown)


kapilash:
cc @davidsh @vijaykota

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_types.cs
```diff
@@ -139,6 +163,16 @@ internal struct CURLMsg
...
 
+        // Curl Version Information
+        [StructLayout(LayoutKind.Explicit)]
+        internal struct curl_version_info_data
+        {
+            [FieldOffset(0)]
+            internal int age;
+            [FieldOffset(32)]
+            internal int features;
+        }
```
stephentoub: I see this in the libcurl docs:
```C
typedef struct {
  CURLversion age;          /* see description below */
 
  /* when 'age' is 0 or higher, the members below also exist: */
  const char *version;      /* human readable string */
  unsigned int version_num; /* numeric representation */
  const char *host;         /* human readable string */
  int features;             /* bitmask, see below */
  char *ssl_version;        /* human readable string */
  long ssl_version_num;     /* not used, always zero */
  const char *libz_version; /* human readable string */
  const char * const *protocols; /* protocols */
 
  /* when 'age' is 1 or higher, the members below also exist: */
  const char *ares;         /* human readable string */
  int ares_num;             /* number */
 
  /* when 'age' is 2 or higher, the member below also exists: */
  const char *libidn;       /* human readable string */
 
  /* when 'age' is 3 or higher (7.16.1 or later), the members below also
     exist  */
  int iconv_ver_num;       /* '_libiconv_version' if iconv support enabled */
 
  const char *libssh_version; /* human readable string */
 
} curl_version_info_data;
```
This looks quite different.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -34,13 +37,19 @@ internal partial class CurlHandler : HttpMessageHandler
...
+        private const int CurlAge = 5;
+        private const int MinCurlAge = 3;
 
         #endregion
 
         #region Fields
-  
+
+        private static bool _supportsAutomaticDecompression;
+        private static bool _supportsSSL;
```
stephentoub: These can both be readonly, right?

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_types.cs
kapilash: Yes. We are interested only in the `features` field. In the odd case, where the libcurl installed is very old, we do not want `Marshal.PtrToStructure` (the structure will not contain all the fields). That is why we are deliberately using `LayoutKind.Explicit`.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
kapilash: True. My bad. They are initialized in the static constructor only.

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_types.cs
stephentoub: I see.  In that case I think it'd be better to use the default of SequentialLayout and include all the fields up through the one you care about, in particular because pointer sizes can change based on 32-bit/64-bit.  Now, with the work @nguerrera's pioneering to introduce native shims, this will end up being in native anyway, and we'll only surface the information we care about, but in the meantime, I think it'd be good to see it something like:
```C#
// Note that this struct is larger than is what's defined here.
// This definition is only valid for use with Marshal.PtrToStructure
// and not for general use in P/Invoke signatures.
internal struct curl_version_info_data
{
    internal CURLversion age;
    private unsafe char* version;
    private int version_num;
    private unsafe char* host;
    internal int features;
    // ... ommitting additional fields here
}
```

kapilash: Ack. But I thought this would be costlier - runtime would try to marshal the two pointers and one integer as well. And in case of explicit, it is just a pointer addition. But fair point about arch differences. 
Sending out a PR with your comments.

kapilash: fixed.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
kapilash: Fixed as well. Thank you

stephentoub:
@dotnet-bot test this please

PR #2806, opened 2015-08-14
Enable CurrentUser X509Stores on Unix.

### LINKED ISSUE #2206: Add support for CurrentUser X509Stores on Unix
LM\Root is effectively the only store coming online with the initial X509Chain implementation (and as ReadOnly).

CU\My is the primary location to expect private keys, and so we really need a story for this.

bartonjs:
The store is implemented as PFX files in ~/.microsoft/netfx/cryptography/x509stores/{storename}, and utilizes a FileSystemWatcher after first read to synchronize changes across other instances (and processes).

Fixes #2206.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
+++ b/src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
@@ -8,6 +8,12 @@ internal static partial class Interop
 {
     internal static partial class Sys
     {
+        // Even though csc will by default use a sequential layout, a CS0649 warning as error
+        // is produced for un-assigned fields when no StructLayout is specified.
+        //
+        // Explicitly saying Sequential disables that warning/error for consumers which only
+        // use Stat in debug builds.
+        [StructLayout(LayoutKind.Sequential)]
         internal struct FileStatus
         {
             internal FileStatusFlags Flags;
--- a/src/Common/src/System/IO/PersistedFiles.Unix.cs
+++ b/src/Common/src/System/IO/PersistedFiles.Unix.cs
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+namespace System.IO
+{
+    internal static class PersistedFiles
+    {
+        // If we ever need system persisted data, /etc/dotnet/corefx/
+        private const string TopLevelDirectory = "dotnet";
+        // User persisted data, ~/.dotnet/corefx/
+        private const string TopLevelUserDirectory = "." + TopLevelDirectory;
+        private const string SecondLevelDirectory = "corefx";
+
+        private static string s_userProductDirectory;
+
+        /// <summary>
+        /// Get the location of where to persist information for a particular aspect of the framework,
+        /// such as "cryptography".
+        /// </summary>
+        /// <param name="featureName">The directory name for the feature</param>
+        /// <returns>A path within the user's home directory for persisting data for the feature</returns>
+        internal static string GetUserFeatureDirectory(string featureName)
+        {
+            if (s_userProductDirectory == null)
+            {
+                EnsureUserDirectories();
+            }
+
+            return Path.Combine(s_userProductDirectory, featureName);
+        }
+
+        /// <summary>
+        /// Get the location of where to persist information for a particular aspect of a feature of
+        /// the framework, such as "x509stores" within "cryptography".
+        /// </summary>
+        /// <param name="featureName">The directory name for the feature</param>
+        /// <param name="subFeatureName">The directory name for the sub-feature</param>
+        /// <returns>A path within the user's home directory for persisting data for the sub-feature</returns>
+        internal static string GetUserFeatureDirectory(string featureName, string subFeatureName)
+        {
+            if (s_userProductDirectory == null)
+            {
+                EnsureUserDirectories();
+            }
+
+            return Path.Combine(s_userProductDirectory, featureName, subFeatureName);
+        }
+
+        /// <summary>
+        /// Get the location of where to persist information for a particular aspect of the framework,
+        /// with a lot of hierarchy, such as ["cryptography", "x509stores", "my"]
+        /// </summary>
+        /// <param name="featurePathParts">A non-empty set of directories to use for the storage hierarchy</param>
+        /// <returns>A path within the user's home directory for persisting data for the feature</returns>
+        internal static string GetUserFeatureDirectory(params string[] featurePathParts)
+        {
+            Debug.Assert(featurePathParts != null);
+            Debug.Assert(featurePathParts.Length > 0);
+
+            if (s_userProductDirectory == null)
+            {
+                EnsureUserDirectories();
+            }
+
+            return Path.Combine(s_userProductDirectory, Path.Combine(featurePathParts));
+        }
+
+        private static void EnsureUserDirectories()
+        {
+            // TODO (2820): Use getpwuid_r(geteuid(), ...) instead of the HOME environment variable
+            string userHomeDirectory = Environment.GetEnvironmentVariable("HOME");
+
+            if (string.IsNullOrEmpty(userHomeDirectory))
+            {
+                throw new InvalidOperationException(SR.PersistedFiles_NoHomeDirectory);
+            }
+
+            s_userProductDirectory = Path.Combine(
+                userHomeDirectory,
+                TopLevelUserDirectory,
+                SecondLevelDirectory);
+        }
+    }
+}
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
@@ -1,0 +1,447 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Internal.Cryptography.Pal
+{
+    /// <summary>
+    /// Provides an implementation of an X509Store which is backed by files in a directory.
+    /// </summary>
+    internal class DirectoryBasedStoreProvider : IStorePal
+    {
+        // {thumbprint}.1.pfx to {thumbprint}.9.pfx
+        private const int MaxSaveAttempts = 9; 
+        private const string PfxExtension = ".pfx";
+        // *.pfx ({thumbprint}.pfx or {thumbprint}.{ordinal}.pfx)
+        private const string PfxWildcard = "*" + PfxExtension;
+        // .*.pfx ({thumbprint}.{ordinal}.pfx)
+        private const string PfxOrdinalWildcard = "." + PfxWildcard;
+
+        private static string s_userStoreRoot;
+
+        private readonly string _storePath;
+        private readonly object _fileWatcherLock = new object();
+        private List<X509Certificate2> _certificates;
+        private FileSystemWatcher _watcher;
+
+        private readonly bool _readOnly;
+
+#if DEBUG
+        static DirectoryBasedStoreProvider()
+        {
+            Debug.Assert(
+                0 == OpenFlags.ReadOnly,
+                "OpenFlags.ReadOnly is not zero, read-only detection will not work");
+        }
+#endif
+
+        internal DirectoryBasedStoreProvider(string storeName, OpenFlags openFlags)
+        {
+            if (string.IsNullOrEmpty(storeName))
+            {
+                throw new CryptographicException(SR.Arg_EmptyOrNullString);
+            }
+
+            string directoryName = GetDirectoryName(storeName);
+
+            if (s_userStoreRoot == null)
+            {
+                // Do this here instead of a static field initializer so that
+                // the static initializer isn't capable of throwing the "home directory not found"
+                // exception.
+                s_userStoreRoot = PersistedFiles.GetUserFeatureDirectory("cryptography", "x509stores");
+            }
+
+            _storePath = Path.Combine(s_userStoreRoot, directoryName);
+
+            if (0 != (openFlags & OpenFlags.OpenExistingOnly))
+            {
+                if (!Directory.Exists(_storePath))
+                {
+                    throw new CryptographicException(SR.Cryptography_X509_StoreNotFound);
+                }
+            }
+
+            // ReadOnly is 0x00, so it is implicit unless either ReadWrite or MaxAllowed
+            // was requested.
+            OpenFlags writeFlags = openFlags & (OpenFlags.ReadWrite | OpenFlags.MaxAllowed);
+
+            if (writeFlags == OpenFlags.ReadOnly)
+            {
+                _readOnly = true;
+            }
+        }
+        
+        public void Dispose()
+        {
+            if (_watcher != null)
+            {
+                _watcher.Dispose();
+                _watcher = null;
+            }
+        }
+
+        public void FindAndCopyTo(
+            X509FindType findType,
+            object findValue,
+            bool validOnly,
+            X509Certificate2Collection collection)
+        {
+        }
+
+        public byte[] Export(X509ContentType contentType, string password)
+        {
+            // Export is for X509Certificate2Collections in their IStorePal guise,
+            // if someone wanted to export whole stores they'd need to do
+            // store.Certificates.Export(...), which would end up in the
+            // CollectionBackedStoreProvider.
+            Debug.Fail("Export was unexpected on a DirectoryBasedStore");
+            throw new InvalidOperationException();
+        }
+
+        public void CopyTo(X509Certificate2Collection collection)
+        {
+            Debug.Assert(collection != null);
+
+            // Copy the reference locally, any directory change operations
+            // will cause the field to be reset to null.
+            List<X509Certificate2> certificates = _certificates;
+
+            if (certificates == null)
+            {
+                // ReadDirectory will both load _certificates and return the answer, so this call
+                // will have stable results across multiple adds/deletes happening in parallel.
+                certificates = ReadDirectory();
+                Debug.Assert(certificates != null);
+            }
+
+            foreach (X509Certificate2 cert in certificates)
+            {
+                collection.Add(cert);
+            }
+        }
+
+        private List<X509Certificate2> ReadDirectory()
+        {
+            if (!Directory.Exists(_storePath))
+            {
+                // Don't assign the field here, because we don't have a FileSystemWatcher
+                // yet to tell us that something has been added.
+                return new List<X509Certificate2>(0);
+            }
+
+            List<X509Certificate2> certs = new List<X509Certificate2>();
+
+            lock (_fileWatcherLock)
+            {
+                if (_watcher == null)
+                {
+                    _watcher = new FileSystemWatcher(_storePath, PfxWildcard)
+                    {
+                        NotifyFilter = NotifyFilters.LastWrite,
+                    };
+
+                    FileSystemEventHandler handler = FlushCache;
+                    _watcher.Changed += handler;
+                    _watcher.Created += handler;
+                    _watcher.Deleted += handler;
+                    // The Renamed event has a different delegate type
+                    _watcher.Renamed += FlushCache;
+                    _watcher.Error += FlushCache;
+                }
+
+                // Start watching for change events to know that another instance
+                // has messed with the underlying store.  This keeps us aligned
+                // with the Windows implementation, which opens stores with change
+                // notifications.
+                _watcher.EnableRaisingEvents = true;
+
+                foreach (string filePath in Directory.EnumerateFiles(_storePath, PfxWildcard))
+                {
+                    X509Certificate2 cert;
+
+                    try
+                    {
+                        cert = new X509Certificate2(filePath);
+                    }
+                    catch (CryptographicException)
+                    {
+                        // The file wasn't a certificate, move on to the next one.
+                        continue;
+                    }
+
+                    certs.Add(cert);
+                }
+
+                // Don't release _fileWatcherLock until _certificates is assigned, otherwise
+                // we may be setting it to a stale value after the change event said to clear it
+                _certificates = certs;
+            }
+
+            return certs;
+        }
+
+        public void Add(ICertificatePal certPal)
+        {
+            // Save the collection to a local so it's consistent for the whole method
+            List<X509Certificate2> certificates = _certificates;
+            OpenSslX509CertificateReader cert = (OpenSslX509CertificateReader)certPal;
+
+            using (X509Certificate2 copy = new X509Certificate2(cert.DuplicateHandles()))
+            {
+                // certificates will be null if anything has changed since the last call to
+                // get_Certificates; including Add being called without get_Certificates being
+                // called at all.
+                if (certificates != null)
+                {
+                    foreach (X509Certificate2 inCollection in certificates)
+                    {
+                        if (inCollection.Equals(copy))
+                        {
+                            if (!copy.HasPrivateKey || inCollection.HasPrivateKey)
+                            {
+                                // If the existing store only knows about a public key, but we're
+                                // adding a public+private pair, continue with the add.
+                                //
+                                // So, therefore, if we aren't adding a private key, or already have one,
+                                // we don't need to do anything.
+                                return;
+                            }
+                        }
+                    }
+                }
+
+                // This may well be the first time that we've added something to this store.
+                Directory.CreateDirectory(_storePath);
+
+                string thumbprint = copy.Thumbprint;
+                bool findOpenSlot;
+
+                // The odds are low that we'd have a thumbprint colission, but check anyways.
+                string existingFilename = FindExistingFilename(copy, _storePath, out findOpenSlot);
+
+                if (existingFilename != null)
+                {
+                    bool dataExistsAlready = false;
+
+                    // If the file on disk is just a public key, but we're trying to add a private key,
+                    // we'll want to overwrite it.
+                    if (copy.HasPrivateKey)
+                    {
+                        try
+                        {
+                            using (X509Certificate2 fromFile = new X509Certificate2(existingFilename))
+                            {
+                                if (fromFile.HasPrivateKey)
+                                {
+                                    // We have a private key, the file has a private key, we're done here.
+                                    dataExistsAlready = true;
+                                }
+                            }
+                        }
+                        catch (CryptographicException)
+                        {
+                            // We can't read this file anymore, but a moment ago it was this certificate,
+                            // so go ahead and overwrite it.
+                        }
+                    }
+                    else
+                    {
+                        // If we're just a public key then the file has at least as much data as we do.
+                        dataExistsAlready = true;
+                    }
+
+                    if (dataExistsAlready)
+                    {
+                        // The file was added but our collection hasn't resynced yet.
+                        // Set _certificates to null to force a resync.
+                        _certificates = null;
+                        return;
+                    }
+                }
+
+                if (_readOnly)
+                {
+                    // Windows compatibility: Don't throw until it has been established that
+                    // the certificate is not present, and we need to modify the filesystem.
+                    throw new CryptographicException(SR.Cryptography_X509_StoreReadOnly);
+                }
+
+                string destinationFilename;
+                FileMode mode = FileMode.CreateNew;
+
+                if (existingFilename != null)
+                {
+                    destinationFilename = existingFilename;
+                    mode = FileMode.Create;
+                }
+                else if (findOpenSlot)
+                {
+                    destinationFilename = FindOpenSlot(thumbprint);
+                }
+                else
+                {
+                    destinationFilename = Path.Combine(_storePath, thumbprint + PfxExtension);
+                }
+
+                using (FileStream stream = new FileStream(destinationFilename, mode))
+                {
+                    byte[] pkcs12 = copy.Export(X509ContentType.Pkcs12);
+                    stream.Write(pkcs12, 0, pkcs12.Length);
+                }
+
+#if DEBUG
+                // Verify that we're creating files with u+rw and o-rw, g-rw.
+                const Interop.libc.Permissions requiredPermissions =
+                    Interop.libc.Permissions.S_IRUSR |
+                    Interop.libc.Permissions.S_IWUSR;
+
+                const Interop.libc.Permissions forbiddenPermissions =
+                    Interop.libc.Permissions.S_IROTH |
+                    Interop.libc.Permissions.S_IWOTH |
+                    Interop.libc.Permissions.S_IRGRP |
+                    Interop.libc.Permissions.S_IWGRP;
+
+                Interop.Sys.FileStatus stat;
+
+                Debug.Assert(Interop.Sys.Stat(destinationFilename, out stat) == 0);
+                Debug.Assert((stat.Mode & (int)requiredPermissions) != 0, "Created PFX has insufficient permissions to function");
+                Debug.Assert((stat.Mode & (int)forbiddenPermissions) == 0, "Created PFX has too broad of permissions");
+#endif
+            }
+
+            // Null out _certificates so the next call to get_Certificates causes a re-scan.
+            _certificates = null;
+        }
+
+        public void Remove(ICertificatePal certPal)
+        {
+            OpenSslX509CertificateReader cert = (OpenSslX509CertificateReader)certPal;
+
+            using (X509Certificate2 copy = new X509Certificate2(cert.DuplicateHandles()))
+            {
+                bool hadCandidates;
+                string currentFilename = FindExistingFilename(copy, _storePath, out hadCandidates);
+
+                if (currentFilename != null)
+                {
+                    if (_readOnly)
+                    {
+                        // Windows compatibility, the readonly check isn't done until after a match is found.
+                        throw new CryptographicException(SR.Cryptography_X509_StoreReadOnly);
+                    }
+
+                    File.Delete(currentFilename);
+                }
+            }
+
+            // Null out _certificates so the next call to get_Certificates causes a re-scan.
+            _certificates = null;
+        }
+
+        private static string FindExistingFilename(X509Certificate2 cert, string storePath, out bool hadCandidates)
+        {
+            hadCandidates = false;
+
+            foreach (string maybeMatch in Directory.EnumerateFiles(storePath, cert.Thumbprint + PfxWildcard))
+            {
+                hadCandidates = true;
+
+                try
+                {
+                    using (X509Certificate2 candidate = new X509Certificate2(maybeMatch))
+                    {
+                        if (candidate.Equals(cert))
+                        {
+                            return maybeMatch;
+                        }
+                    }
+                }
+                catch (CryptographicException)
+                {
+                    // Contents weren't interpretable as a certificate, so it's not a match.
+                }
+            }
+
+            return null;
+        }
+
+        private string FindOpenSlot(string thumbprint)
+        {
+            // We already know that {thumbprint}.pfx is taken, so start with {thumbprint}.1.pfx
+
+            // We need space for {thumbprint} (thumbprint.Length)
+            // And ".0.pfx" (6)
+            // If MaxSaveAttempts is big enough to use more than one digit, we need that space, too (MaxSaveAttempts / 10)
+            StringBuilder pathBuilder = new StringBuilder(thumbprint.Length + PfxOrdinalWildcard.Length + (MaxSaveAttempts / 10));
+
+            pathBuilder.Append(thumbprint);
+            pathBuilder.Append('.');
+            int prefixLength = pathBuilder.Length;
+
+            for (int i = 1; i <= MaxSaveAttempts; i++)
+            {
+                pathBuilder.Length = prefixLength;
+                
+                pathBuilder.Append(i);
+                pathBuilder.Append(PfxExtension);
+
+                string builtPath = Path.Combine(_storePath, pathBuilder.ToString());
+
+                if (!File.Exists(builtPath))
+                {
+                    return builtPath;
+                }
+            }
+
+            throw new CryptographicException(SR.Cryptography_X509_StoreNoFileAvailable);
+        }
+
+        private void FlushCache(object sender, EventArgs e)
+        {
+            lock (_fileWatcherLock)
+            {
+                // Events might end up not firing until after the object was disposed, which could cause
+                // problems consistently reading _watcher; so save it to a local.
+                FileSystemWatcher watcher = _watcher;
+
+                if (watcher != null)
+                {
+                    // Stop processing events until we read again, particularly because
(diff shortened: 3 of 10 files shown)


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+
+namespace Internal.Cryptography.Pal
+{
+    /// <summary>
+    /// Provides an implementation of an X509Store which is backed by files in a directory.
+    /// </summary>
+    internal class DirectoryBasedStoreProvider : IStorePal
+    {
+        // {thumbprint}.1.pfx to {thumbprint}.9.pfx
+        private const int MaxSaveAttempts = 9;
```
stephentoub: Is 9 a sufficient number?  Why have a limit at all?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+        {
+            if (string.IsNullOrEmpty(storeName))
+            {
+                throw new CryptographicException(SR.Arg_EmptyOrNullString);
+            }
+
+            string directoryName = GetDirectoryName(storeName);
+         
+            _storePath = Path.Combine(
+                Environment.GetEnvironmentVariable("HOME"),
```
stephentoub: What if "HOME" isn't set and this comes back as null?  Won't that blow up?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+            }
+        }
+
+        private List<X509Certificate2> ReadDirectory()
+        {
+            if (!Directory.Exists(_storePath))
+            {
+                // Don't assign the field here, because we don't have a FileSystemWatcher
+                // yet to tell us that something has been added.
+                return new List<X509Certificate2>(0);
```
stephentoub: I assume this is pretty rare?  If not, consider caching the empty list.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+            {
+                throw new CryptographicException(SR.Arg_EmptyOrNullString);
+            }
+
+            string directoryName = GetDirectoryName(storeName);
+         
+            _storePath = Path.Combine(
+                Environment.GetEnvironmentVariable("HOME"),
+                ".microsoft",
+                "netfx",
```
stephentoub: I believe this is now the first time we'll have a need for persisting non-temporary files.  If we have additional need to do so, we'll want them to all be in the same general place.  (And even for temporary files, we should probably have a consistent location in which we store them, e.g. Path.Combine(Path.GetTempPath(), ".microsoft", "netfx", ...)).  Given that, what would you think of adding some constants to a file somewhere in Common, e.g. PersistedFiles.TopLevelFolderName = ".microsoft", PersistedFiles.SecondLevelFolderName = "netfx", and then we'd use those here and anywhere else we need to write out files. e.g. Path.Combine(homeDirectory, PersistedFiles.TopLevelFolderName, PersistedFiles.SecondLevelFolderName, "cryptography", ...), and in other libraries where we write out temporaries, we could use Path.Combine(Path.GetTempPath(), PersistedFiles.TopLevelFolderName, ...), etc.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+                certs.Add(cert);
+            }
+
+            if (_watcher == null)
+            {
+                _watcher = new FileSystemWatcher(_storePath, PfxWildcard);
+                _watcher.Changed += DirectoryChanged;
+                _watcher.Created += DirectoryChanged;
+                _watcher.Deleted += DirectoryChanged;
+                _watcher.Renamed += DirectoryChanged;
```
stephentoub: Do we need to watch for errors, too?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+
+                certs.Add(cert);
+            }
+
+            if (_watcher == null)
+            {
+                _watcher = new FileSystemWatcher(_storePath, PfxWildcard);
+                _watcher.Changed += DirectoryChanged;
+                _watcher.Created += DirectoryChanged;
+                _watcher.Deleted += DirectoryChanged;
```
stephentoub: As a small optimization, it'd be better to do:
```C#
FileSystemEventHandler handler = DirectoryChanged;
_watcher.Changed += handler;
_watcher.Created += handler;
_watcher.Deleted += handler;
```

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+                    // The file wasn't a certificate, move on to the next one.
+                    continue;
+                }
+
+                certs.Add(cert);
+            }
+
+            if (_watcher == null)
+            {
+                _watcher = new FileSystemWatcher(_storePath, PfxWildcard);
```
stephentoub: Can we limit the NotifyFilters in any way, or we need to be notified of any possible change to any of the files?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+                catch (CryptographicException)
+                {
+                    // The file wasn't a certificate, move on to the next one.
+                    continue;
+                }
+
+                certs.Add(cert);
+            }
+
+            if (_watcher == null)
```
stephentoub: Do we need to be concerned with files that get added/modified between the time that we enumerate the directory and the time we create/start the watcher?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+
+        private static string GetDirectoryName(string storeName)
+        {
+            Debug.Assert(storeName != null);
+
+            try
+            {
+                string fileName = Path.GetFileName(storeName);
+
+                if (!StringComparer.Ordinal.Equals(storeName, fileName))
```
justinvp: Why not `!string.Equals(storeName, fileName, StringComparison.Ordinal)`?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+                    destinationFilename = FindOpenSlot(thumbprint);
+                }
+                else
+                {
+                    destinationFilename = Path.Combine(_storePath, thumbprint + PfxExtension);
+                }
+
+                // With no explicit permissions dictated these are being created with Unix permissions of
+                // -rwx------
+                // which leaves the private keys scoped to the user, and makes us happy.
```
stephentoub: If this is important, should we assert that it remains true?  These were the permissions I used at https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs#L88, but if that's important for this code, we'll want to make sure we don't inadvertently breaking it by changing those permissions.  Plus, it's possible for the system to use a umask to further constrain the permissions when new files are created; not sure if that negatively affects this.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,400 @@
...
+
+                string destinationFilename;
+
+                if (existingFilename != null)
+                {
+                    destinationFilename = existingFilename;
+                }
+                else if (findOpenSlot)
+                {
+                    destinationFilename = FindOpenSlot(thumbprint);
```
stephentoub: Do we need to be concerned about another process racing to find a slot and ending up with the same one?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: I don't like the idea of `Add` having a theoretically infinite loop due to someone poisoning the directory.

Since the base portion of the filename is the SHA-1 hash of the certificate (aka the thumbprint/fingerprint) the chance of needing to use this limiter at all is low.  So this is allowing for 10 different certificates all with the same thumbprint (or up to 10 poison files) before we give up and tell someone that the universe stopped making sense.

And I'll admit that I went with 9 because it allowed for a nice pattern of {stem}.pfx and {stem}.1.pfx to {stem}.9.pfx in decimal formatting.

stephentoub: Ok, sounds reasonable.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: The POSIX rules say that a conforming shell will set HOME.  Technically a user could unset it, or redirect it; but a lot of Unix software would get really cranky at that point.

For maximal defense we could add something which uses `getpwuid(geteuid())` to read the `pw_dir` entry; but I'm not sure that's warranted here.  (Since it's non-sealed struct reading we'd want a native shim over it, so we'd probably want something like a resurgence of Environment.GetFolderPath for the `SpecialFolder.UserProfile` value... but that feels a bit heavy-handed.)

bartonjs: Forgot my supporting documentation link: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html (section 8.3)

http://pubs.opengroup.org/onlinepubs/009695399/basedefs/pwd.h.html (says "at least the following members" for `struct passwd`)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Sounds perfectly reasonable.  I think I'd want to take it a step further and add `PersistedFiles.GetUserFeatureDirectory(string feature)`, which returns ~/.microsoft/netfx/{feature}; and would leave the `${HOME}` vs more complicated code aspect centralized.

And maybe even
```C#
PersistedFiles.GetUserFeatureDirectory(params string[] featurePathParts)
{
    Debug.Assert(featurePathParts != null);
    Debug.Assert(featurePathParts.Length > 0);
    ...
}
```

And then we know everyone's being good about being under ~/.microsoft/netfx.  Thoughts?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Yeah, it's a pretty convoluted situation, but would work on Windows:

```C#
using (X509Store storeA = new X509Store("LetsMakeANewStore", StoreLocation.CurrentUser))
using (X509Store storeB = new X509Store("LetsMakeANewStore", StoreLocation.CurrentUser))
{
    storeA.Open(OpenFlags.ReadWrite);
    storeB.Open(OpenFlags.ReadOnly);

    // Directory doesn't exist yet

    // Make people in the know sad at how wasteful this line is, since Certificates is rehydrated each request. Also, 0 is the answer
    int storeBCount = storeB.Certificates.Count;

    storeA.Add(someMagicCert);

    // That add caused the directory to exist

    // Now this is 1.
    storeBCount = storeB.Certificates.Count;
}
```

It's "you asked for the contents of a store that you've never written to" (being just empty isn't sufficient, we won't go back and delete the old directory); and I don't really forsee that being regular.

Though if you just mean make the line be akin to `return s_emptyList ?? (s_emptyList = new List<X509Certificate2>(0));`, that's something which wouldn't have visible effect (even convolutedly), so I'd be happy doing that.  (The list itself being a "cache" I parsed out two meanings of the comment)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Quite prossibly.  (prossibly: (n) The fine line between "possibly" and "probably")

I was hoping to avoid locking; but don't think I can.

Create FSW
Activate FSW
Read directory
FSW fires
assign _certs

Already have bad data.  So I guess I need to lock this method and the FSW event; and flip the activation to before the read.

Alternatively, I could dump the cache altogether; but multiple enumerations (such as multiple calls to `Find` (once I write that)) feels like it'll be more common than concurrent writes.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: LastWriteTime is probably sufficient.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Will do

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: It couldn't really hurt

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
stephentoub: Sounds good :smile:

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
stephentoub: > Though if you just mean make the line be akin to

Yeah, I was referring just to caching s_emptyList.  But if this isn't common, it's not worth the extra caching field.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
stephentoub: > prossibly

I like it; I've now added it to my vocabulary.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: What I care about is u+rw, o-rw, and probably g-rw (execute is a complete don't care, though for cleanliness it shouldn't really be set... but CIFS/SMB writes files with +x, so it's not important).

Since that permissions set is pretty well hard-coded I'm not sure how I'd go about asserting that here (either the forward assert of "make this true" or the backward assert of "ack! the universe stopped making sense"); but it would certainly please me to learn how...

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
stephentoub: Cool.  In particular I was thinking we'd at least want to avoid having LastAccessTime set, as we'd then disable watching any time someone read a cert.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: I find this way easier to read, and it has the benefit of dealing with nulls (which doesn't apply here... but it drives my motor memory)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
stephentoub: > Since that permissions set is pretty well hard-coded I'm not sure how I'd go about asserting that here

I was thinking you would do (pseudo code):
```C#
File.WriteAllBytes(destinationFileName, ...);
#if DEBUG
FileStatus stat;
Debug.Assert(Interop.libcoreclr.GetFileInformationFromPath(destinationFileName, out stat) == 0);
Debug.Assert((stat.mode & permissionsYouNeed) != 0);
Debug.Assert((stat.mode & permissionsYouDontWant) == 0);
#endif
```
or something like that.  Is that not valid?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: The odds of two writes to the same file being different certificates is Really Low (because certificates are all about the same size, and hash collisions usually require significantly disparate sizing).

A more plausible, but still weird, case would be object A is writing a public key, and object B is writing a public/private pair.  If A happens to be the last writer the private key went bye-bye.

I think I've convinced myself that I need to open the stream myself to ensure that it opens in CreateNew when it thinks it's writing a new file (the public key to public+private key upgrade is Create).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Danke.  Intentionally broke the second one to verify it worked, and happiness was had.

```
Created PFX has too broad of permissions
---- Assert Long Message ----

   at Internal.Cryptography.Pal.DirectoryBasedStoreProvider.Add(ICertificatePal certPal)
   at System.Security.Cryptography.X509Certificates.X509Store.Add(X509Certificate2 certificate)
```

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
akoeplinger: @bartonjs Note that there **are** valid cases where HOME isn't set, e.g. in init scripts. We should definitely make sure we don't just blow up when HOME isn't there or have a fallback.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
akoeplinger: +1 to what Stephen said. Additionally, has there been any discussion about what the folder should be called? Forgive my honesty, but I'm not fond of `~/.microsoft` given that this is a .NET Foundation project. Why not something like `~/.dotnet` or `~/.corefx`?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: @akoeplinger Presumably by "init scripts" you mean `/etc/rc/` (or your distro's preferred mechanism), and not `~/.bashrc`?  This is for user-scoped certificates, system scoped is a different issue.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: I'd be fine with ~/.dotnet (and a parallel of /etc/dotnet if/when we do system-level persistence).  @stephentoub? @terrajobst?

stephentoub: I like ".dotnet" instead of ".microsoft" as well.  I'd still have "corefx" as a subdirectory, so as to avoid conflicts with coreclr, dnx, or anything else that decided to go to the same location.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
akoeplinger: Yeah. Even init scripts can be ran as a different user than root (i.e. a specific service user), so having "user-scoped" certificates is still a valid case there. If we want to rely on HOME, then that's fine as long as it throws a proper exception when it's not set :smile:

bartonjs: Okay, with the delta commit that's about to appear, if I blank $HOME in a shell script before launching the tests I get:
```
   System.Security.Cryptography.X509Certificates.Tests.X509StoreTests.ReadMyCertificates [FAIL]
      System.InvalidOperationException : The home directory of the current user could not be determined.
      Stack Trace:
            at System.IO.PersistedFiles.EnsureUserDirectories()
            at System.IO.PersistedFiles.GetUserFeatureDirectory(String featureName, String subFeatureName)
            at Internal.Cryptography.Pal.DirectoryBasedStoreProvider..ctor(String storeName, OpenFlags openFlags)
            at Internal.Cryptography.Pal.StorePal.FromSystemStore(String storeName, StoreLocation storeLocation, OpenFlags openFlags)
            at System.Security.Cryptography.X509Certificates.X509Store.Open(OpenFlags flags)
            at System.Security.Cryptography.X509Certificates.Tests.X509StoreTests.ReadMyCertificates()
```

bartonjs:
Compile breaks in release were because Stat was only called in debug.  `[StructLayout(LayoutKind.Sequential)]` solved that problem, and I'll leave that as a separate commit when this is all said and done. (@nguerrera and I verified in the emitted IL that this wasn't actually changing anything)

CI doesn't have the HOME environment variable set.  Good prescience, @akoeplinger. #2820 is tracking fixing this for real.

"Round 3, fight!"

stephentoub:
@dotnet-bot test this please

bartonjs:
@dotnet-bot test this please (without intermittent failures, if you don't mind)

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
```diff
@@ -8,6 +8,7 @@ internal static partial class Interop
 {
     internal static partial class Sys
     {
+        [StructLayout(LayoutKind.Sequential)]
```
stephentoub: Can you add a comment for why this is being specified explicitly even though it's the default?

src/Common/src/System/IO/PersistedFiles.Unix.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    internal static class PersistedFiles
+    {
+        // If we ever need system persisted data, /etc/dotnet/corefx/
+        private const string TopLevelDirectory = "dotnet";
+        // User persisted data, ~/.dotnet/corefx/
+        private const string TopLevelUserDirectory = "." + TopLevelDirectory;
+        private const string SecondLevelDirectory = "corefx";
+
+        // Do not use the user variables in the static initializer in case $HOME didn't resolve.
```
stephentoub: You mean we avoid setting these in the static cctor in case it doesn't resolve, results in exception, and makes the type unusable, right?

src/Common/src/System/IO/PersistedFiles.Unix.cs
```diff
@@ -0,0 +1,88 @@
...
+            Debug.Assert(featurePathParts != null);
+            Debug.Assert(featurePathParts.Length > 0);
+
+            if (s_userProductDirectory == null)
+            {
+                EnsureUserDirectories();
+            }
+
+            return Path.Combine(s_userProductDirectory, Path.Combine(featurePathParts));
+        }
```
stephentoub: Just FYI, after this change goes in, I plan to add ```GetUserTempFeatureDirectory``` methods, and then use them in the libraries that generate temporary files, e.g. Pipes, MemoryMappedFiles, etc.

src/Common/src/System/IO/PersistedFiles.Unix.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        private static void EnsureUserDirectories()
+        {
+            // TODO (2820): Use getpwuid_r(geteuid(), ...) instead of the HOME environment variable
+            s_userHomeDirectory = Environment.GetEnvironmentVariable("HOME");
```
stephentoub: It seems the only place this is needed is in subsequently computing s_userProductDirectory.  Could it be a local instead of a field?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+                // the static initializer isn't capable of throwing the "home directory not found"
+                // exception.
+                s_userStoreRoot = PersistedFiles.GetUserFeatureDirectory("cryptography", "x509stores");
+            }
+
+            _storePath = Path.Combine(s_userStoreRoot, directoryName);
+
+            if (0 != (openFlags & OpenFlags.OpenExistingOnly))
+            {
+                if (!Directory.Exists(_storePath))
```
stephentoub: Nit: could just be an && in the previous if rather than a nested if

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+                else if (findOpenSlot)
+                {
+                    destinationFilename = FindOpenSlot(thumbprint);
+                }
+                else
+                {
+                    destinationFilename = Path.Combine(_storePath, thumbprint + PfxExtension);
+                }
+
+                using (FileStream stream = new FileStream(destinationFilename, mode))
```
stephentoub: If this results in an exception due to a race in targeting the same slot (e.g. another process adding the same cert), should we retry? Catch the IOException and throw a CryptographicException?  Just let this exception propagate?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+            OpenSslX509CertificateReader cert = (OpenSslX509CertificateReader)certPal;
+
+            using (X509Certificate2 copy = new X509Certificate2(cert.DuplicateHandles()))
+            {
+                bool hadCandidates;
+                string currentFilename = FindExistingFilename(copy, _storePath, out hadCandidates);
+
+                if (currentFilename != null)
+                {
+                    if (_readOnly)
```
stephentoub: Should we throw on attempted removal even if we can't find the cert in the store, e.g. moving this if up to the beginning of the method?  In Add, it looks like we throw in the similar case where we're adding a cert that already exists and thus it would be a nop.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+            // If MaxSaveAttempts is big enough to use more than one digit, we need that space, too (MaxSaveAttempts / 10)
+            StringBuilder pathBuilder = new StringBuilder(thumbprint.Length + PfxOrdinalWildcard.Length + (MaxSaveAttempts / 10));
+            HashSet<string> existingFiles = new HashSet<string>(Directory.EnumerateFiles(_storePath, thumbprint + PfxOrdinalWildcard));
+
+            for (int i = 1; i <= MaxSaveAttempts; i++)
+            {
+                pathBuilder.Clear();
+
+                pathBuilder.Append(thumbprint);
+                pathBuilder.Append('.');
```
stephentoub: This is minor, especially since you've stated that FindOpenSlot will be called rarely, and the chances of this iterating more than once are even rarer, but... rather than doing:
```C#
pathBuilder.Clear();
pathBuilder.Append(thumbprint);
pathBuilder.Append('.');
```
as part of the loop, you could do:
```C#
pathBuilder.Append(thumbprint);
pathBuilder.Append('.');
```
before the loop, and then in the loop do:
```C#
pathBuilder.Length = thumbprint.Length + 1;
```

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+                pathBuilder.Append(i);
+                pathBuilder.Append(PfxExtension);
+
+                string builtPath = pathBuilder.ToString();
+
+                if (!existingFiles.Contains(builtPath))
```
stephentoub: Rather than enumerating the directory, building up a hash set from that, and doing contains checks against it, since you expect multiple iterations to be rare, did you consider just doing a File.Exists check here?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,454 @@
...
+
+        private void WatcherError(object sender, ErrorEventArgs e)
+        {
+            FlushCache();
+        }
+
+        private void DirectoryChanged(object sender, FileSystemEventArgs e)
+        {
+            FlushCache();
+        }
```
stephentoub: Since you don't need the strongly typed event args, you can collapse both WatcherError and DirectoryChanged into a single:
```C#
private void WatcherChanged(object sender, EventArgs e)
{
    FlushCache();
}
```
and then use that one method for all of the callbacks.  Delegates support contravariance in this regard.

src/System.Security.Cryptography.X509Certificates/src/project.json
```diff
@@ -5,6 +5,7 @@
     "System.Diagnostics.Debug": "4.0.10",
     "System.Globalization.Calendars": "4.0.0",
     "System.IO.FileSystem": "4.0.0",
+    "System.IO.FileSystem.Watcher": "4.0.0-beta-*",
```
stephentoub: Is "-beta-*" the right thing here?

stephentoub:
Thanks, Jeremy,  Left some minor comments, but in general LGTM.

src/Common/src/Interop/Unix/System.Native/Interop.Stat.cs
bartonjs: Done.

src/Common/src/System/IO/PersistedFiles.Unix.cs
bartonjs: Yeah, that was the implicit remainder of that comment.  But since I've removed the field it's a bit less of a trap now.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: So, originally I had made the trivial tests of Add and Remove while ReadOnly.  Add passed on Windows (threw the exception), Remove didn't.  So I made the test with lots of comments about how the test can't really work happily, yadda yadda.  Didn't occur at all to me that "add when it already exists" would be a similar no-op.  But it is.  So, grr.  Making `Add` now only throw when it thinks it needs to do work.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: I'm sure what I did made sense at the time that I wrote it, but I agree I was doing way too much work... made the change.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: Nice.  Made the change.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: For some reason that didn't occur to me.  I would have expected it for passing to a parameter, but I think the fact I was interacting with an event threw me off.  Changed, thanks :).

src/System.Security.Cryptography.X509Certificates/src/project.json
bartonjs: System.IO.FileSystem.Watcher doesn't have a stable version published, so the -beta-* is required.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
bartonjs: I started down the path of wrapping this in a `for { try { find_filename_and_write(); } catch (IOException e) { maybe_wrap_and_throw(); } }` construct, and was upset at the amount of unverifiable code that I was writing.

I think I want to just leave it as the IOException (though could be easily convinced) due to this reasoning:
* On Windows the exception has the `HResult` property which could be inspected to act on the failure reason.
* On Unix we won't have that, and emitting the IOException would at least allow the caller code to handle the failure case semantically.

The counter argument would be "Yeah, but they can just read `e.InnermostException()` and act accordingly", which I'm behind like 49.98%.  Partially because they wouldn't know to write that code.

bartonjs:
Incremental diffs are available at https://github.com/bartonjs/corefx/commits/X509Store_Unix_CurrentUser_reviewed, in particular https://github.com/bartonjs/corefx/commit/46ee2d8001e033ada2dab9a2ea2e39b5f9e72868 and https://github.com/bartonjs/corefx/commit/61d671f12df325cff3cd9e79ab9f2e80fd897070.

Squashed down to two commits (the FileStatus layout attribute and "the rest").

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
```diff
@@ -0,0 +1,447 @@
...
+            // If MaxSaveAttempts is big enough to use more than one digit, we need that space, too (MaxSaveAttempts / 10)
+            StringBuilder pathBuilder = new StringBuilder(thumbprint.Length + PfxOrdinalWildcard.Length + (MaxSaveAttempts / 10));
+
+            pathBuilder.Append(thumbprint);
+            pathBuilder.Append('.');
+            int prefixLength = pathBuilder.Length;
+
+            for (int i = 1; i <= MaxSaveAttempts; i++)
+            {
+                pathBuilder.Length = prefixLength;
```
stephentoub: Nit (not worth fixing unless you're going to make additional changes): this could move to the end of the loop, since the first time in it's already the right length.

stephentoub:
LGTM.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/DirectoryBasedStoreProvider.cs
justinvp: Fair enough. FWIW, the static `String.Equals` does handle nulls, and may be slightly faster, as the `StringComparer` may not have been allocated yet (though, it probably has by the time this code runs) and `StringComparer` just ends up eventually calling `String.Equals`.

PR #2809, opened 2015-08-14
Make Expression.Empty() use a single instance.

hackcraft:
Reduces number of calls and allocations.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
@@ -49,6 +49,8 @@ namespace System.Linq.Expressions
 
     public partial class Expression
     {
+        private static readonly DefaultExpression s_empty = new DefaultExpression(typeof(void));
+
         /// <summary>
         /// Creates an empty expression that has <see cref="System.Void"/> type.
         /// </summary>
@@ -58,7 +60,7 @@ namespace System.Linq.Expressions
         /// </returns>
         public static DefaultExpression Empty()
         {
-            return new DefaultExpression(typeof(void));
+            return s_empty;
         }
 
         /// <summary>


src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
```diff
@@ -49,6 +49,7 @@ protected internal override Expression Accept(ExpressionVisitor visitor)
 
     public partial class Expression
     {
+        private static readonly DefaultExpression _empty = new DefaultExpression(typeof(void));
```
stephentoub: Nit: add a blank line after this one

stephentoub: ```s_empty``` ?

stephentoub:
cc: @VSadov 

Assuming the base Expression type is immutable, LGTM.

src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
hackcraft: Is `s_` the general convention? In any case, there's certainly a few using it in this assembly, so I'll do that in a minute.

stephentoub: In the corefx repo, for static fields, yup.

hackcraft:
The base is immutable. Indeed, the base has no non-static state at all.

VSadov:
This would be a breaking change to anyone expecting that Expression.Empty creates a new instance.

In ET v2 we had implemented a fairly elaborate node deduplication logic that would also reuse trivial things like numerical constants as well as complete subtrees that could be proven semantically equivalent to cached node. Just to rip the whole thing out for compat reasons. :-(

Interestingly, the same approach works nicely at the green layer of Roslyn syntax trees. Those are immutable and not public, so noone can tell the difference when nodes are reused.

hackcraft:
Is there a case where such reference non-identity is going to have an impact? If so I'll change this PR to one that just adds a comment warning against it.

VSadov:
The most trivial breakage is when customer uses nodes as keys in a hashtable. With this change client will crash with "adding a duplicate key". 
Customer's reaction will not be rewriting of his, possibly years old, code, but instead a strong demand to revert to pre-existing behavior.
Besides, he can chose to create new instances or to reuse the same one. With the change, there will be no choice so he will see it as a take-back. These things are hard to change once shipped.

VSadov:
Re: documenting the behavior and rationale behind that

I believe most if not all factory methods create new instances. Those which don't, if there are such, must have been doing so since v1 and hopefully say so in documentation. 
Otherwise, creating new nodes is the default behavior since deduping can be added externally, but cannot be undone if factories themselves start doing it.

hackcraft:
Hmm. Looking at the implementation of Default(), there's a branch there that makes sense after this PR but before it is just wasted work for the same result.

I'm going to take that as evidence that at some point the same idea as this had been tried and found wanting, and close this.

When I'm not on a phone I think I will add that comment; I'm not the only person who when they see multiple equivalent immutable objects produced, will want to make them single instances, so someone else may well try the same thing.

PR #2811, opened 2015-08-14
Fix and enable a disabled Task test

### LINKED ISSUE #2797: TestSchedulerNesting occasionally fails
The test for nested ConcurrentExclusiveSchedulerPair's seems to fail very occasionally, about 1-5 times per 1000 runs in my machine. The root cause seems to be that some of the tasks are being scheduled to run after the scheduler pair has been completed (after the Complete() method is called). Based on the structure of the test, this shouldn't necessarily be possible, as all of the previous tasks are Wait()'d before the call to Complete(). There may be a product bug here.

stephentoub:
The test was assuming that if all tasks queued to a scheduler have completed that the scheduler won't be making further use of whatever underlying scheduler it's targeting.  That assumption is false, however, as the scheduler may still queue additional housekeeping tasks, and it requires that the underlying scheduler remain alive until the overlaying scheduler declares completion via its Completion task.

Fixes #2797 
cc: @mellinoe

### DIFF
--- a/src/System.Threading.Tasks/tests/CESchedulerPairTests.cs
+++ b/src/System.Threading.Tasks/tests/CESchedulerPairTests.cs
@@ -420,12 +420,8 @@ namespace System.Threading.Tasks.Tests
 
         /// <summary>
         /// Ensure that CESPs can be layered on other CESPs.
-        /// There is and assert in src\Threading\System\Threading\Tasks\ConcurrentExclusiveSchedulerPair.cs
-        /// line 355 - the scheduler enters in exclusive processing mode when  should not
-        /// it needs more investigations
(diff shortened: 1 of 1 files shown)


mellinoe:
Can't say I am incredibly familiar with this part of the code, but LGTM :+1:

stephentoub:
@dotnet-bot test this please

(hung restoring packages)

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:
(builds are failing due to GitHub and package restore timeout issues... seems like there might be some networking problems...)

PR #2814, opened 2015-08-14
Reduce running time of Tasks tests

stephentoub:
@mellinoe, I know you're planning to spend some time fixing up the task tests.  In the meantime, they're taking between 40-80 seconds to run on my machine, and they appear to be taking an equivalent amount of time on the CI servers.  This change makes a few tweaks to bring that time down:
- Marks the worst offending tests as [OuterLoop]
- Changes a few places that used to be Environment.ProcessorCount and were apparently at some point changed to be a hardcoded number back to being Environment.ProcessorCount

The tests now take ~10 seconds on my machine, though still with quite a bit of variability.

### DIFF
--- a/src/System.Threading.Tasks/tests/CESchedulerPairTests.cs
+++ b/src/System.Threading.Tasks/tests/CESchedulerPairTests.cs
@@ -101,7 +101,7 @@ namespace System.Threading.Tasks.Tests
             ConcurrentExclusiveSchedulerPair schedPair = null;
             //Need to define the default values since these values are passed to the verification methods
             TaskScheduler scheduler = TaskScheduler.Default;
-            int maxConcurrentLevel = 4;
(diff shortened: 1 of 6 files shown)


mellinoe:
LGTM. 40 seconds is definitely way too long. On my machine they are a bit faster than that, around ~10 seconds, but the CI server seems to be taking a very long time as well.

Do you have any ideas to reduce the variability of the run time? I don't believe there is any randomness to the tests from what I saw. Is it just the nature of the scheduling of the tasks that makes it so different? It will be easier to trim the time down further if we can get consistent timings.

stephentoub:
> Do you have any ideas to reduce the variability of the run time? 

I haven't spent much time looking at them. My guess is that it's very much based on how things are getting scheduled and how that interacts with thread pool thread injection, combined with what appears to be a fair amount of blocking happening in the test code.

PR #2824, opened 2015-08-15
Cleanup and harden X509Certificate collections

justinvp:
This PR is broken up into three separate commits (details below). The latter two commits have minor behavior changes that I think are worth doing for .NET Core.

cc: @bartonjs, @weshaggard 

@weshaggard, are the behavior changes in commits 2 and 3 acceptable for .NET Core? If not, I can remove or modify these commits.

#### 1. Cleanup X509Certificate collections

- Avoid enumerator allocations when enumerating the collections.
 - Either enumerate the underlying list directly, or use a for loop and and the indexer, to avoid the enumerator allocation.
- Reduce the allocations associated with creating an enumerator.
 - Unfortunately these already use public reference type enumerators. Internally we can have the enumerators use `List<T>`'s struct enumerator instead of a boxed `IEnumerator`, reducing the allocations associated with the `X509CertificateEnumerator` from 2 allocations to 1. `X509Certificate2Enumerator` was allocating and using the `X509CertificateEnumerator`, and we can just have it use the list's struct enumerator directly, so its allocations are down from 3 allocations to 1.
- Throw `ArgumentNullException` from `Insert` and `Remove` to match the full .NET Framework (desktop) behavior. Add tests to verify.
- Remove some `ArgumentOutOfRangeException` checks as the underlying `List<T>` already handles the same checks. Add tests to verify.
- Use `List<T>`'s `ICollection.SyncObject`, allowing the field to be removed.
- Remove unused usings.
- Move fields to the top of class definitions.
- `Object` -> `object`; `String` -> `string`.
- Minor whitespace cleanup.

#### 2. Harden X509Certificate collections

This commit introduces minor behavior changes in order to harden the `X509Certificate` collections in .NET Core:

- Using `X509CertificateCollection` or `X509Certificate2Collection` as an `IList` allows adding/inserting non-`X509Certificate` objects into the collection. Subsequent calls to enumerate or access these items results in `InvalidCastException` being thrown. This matches the full .NET Framework (desktop) behavior. Clearly this is not the intended behavior of these collections, which are meant to hold `X509Certificate` and `X509Certificate2` objects. This commit changes `X509CertificateCollection` and `X509Certificate2Collection` to throw `ArgumentException` when non-`X509Certificate` objects are attempted to be added/inserted (this behavior comes "for-free" by switching the underlying list from `List<object>` to `List<X509Certificate>`).

- `X509CertificateEnumerator` has a public constructor with a collection parameter that it uses without first verifying the parameter is non-null, which could result in a `NullReferenceException`. This commit adds a null check, throwing `ArgumentNullException` if the parameter is null.

#### 3. X509Certificate2Collection.Contains: Allow null certificates

`X509Certificate2Collection.Contains` throws `ArgumentNullException` when a null value is passed in, but its `IList.Contains` implementation and `X509CertificateCollection.Contains` do not throw. This commit changes `X509Certificate2Collection.Contains` to not throw for a null value to be consistent with its `IList.Contains` implementation and `X509CertificateCollection.Contains`.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
@@ -30,14 +30,11 @@ namespace System.Security.Cryptography.X509Certificates
         {
             get
             {
-                return (X509Certificate2)(List[index]);
+                return (X509Certificate2)(base[index]);
             }
             set
             {
-                if (value == null)
-                    throw new ArgumentNullException("value");
-
-                List[index] = value;
+                base[index] = value;
             }
         }
 
@@ -46,7 +43,7 @@ namespace System.Security.Cryptography.X509Certificates
             if (certificate == null)
                 throw new ArgumentNullException("certificate");
 
-            return List.Add(certificate);
+            return base.Add(certificate);
         }
 
         public void AddRange(X509Certificate2[] certificates)
@@ -100,7 +97,7 @@ namespace System.Security.Cryptography.X509Certificates
             if (certificate == null)
                 throw new ArgumentNullException("certificate");
 
-            return List.Contains(certificate);
+            return base.Contains(certificate);
         }
 
         public byte[] Export(X509ContentType contentType)
@@ -171,7 +168,7 @@ namespace System.Security.Cryptography.X509Certificates
             if (certificate == null)
                 throw new ArgumentNullException("certificate");
 
-            List.Insert(index, certificate);
+            base.Insert(index, certificate);
         }
 
         public void Remove(X509Certificate2 certificate)
@@ -179,7 +176,7 @@ namespace System.Security.Cryptography.X509Certificates
             if (certificate == null)
                 throw new ArgumentNullException("certificate");
 
-            List.Remove(certificate);
+            base.Remove(certificate);
         }
 
         public void RemoveRange(X509Certificate2[] certificates)
--- a/src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
@@ -1,14 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.IO;
-using System.Text;
-using System.Diagnostics;
-using System.Globalization;
-using System.Runtime.InteropServices;
-
-using Internal.Cryptography;
 using Internal.Cryptography.Pal;
 
 namespace System.Security.Cryptography.X509Certificates
@@ -44,6 +36,7 @@ namespace System.Security.Cryptography.X509Certificates
             {
                 if (value == null)
                     throw new ArgumentNullException("value");
+
                 List[index] = value;
             }
         }
@@ -87,10 +80,9 @@ namespace System.Security.Cryptography.X509Certificates
             int i = 0;
             try
             {
-                foreach (X509Certificate2 certificate in certificates)
+                for (; i < certificates.Count; i++)
                 {
-                    Add(certificate);
-                    i++;
+                    Add(certificates[i]);
                 }
             }
             catch
@@ -116,7 +108,7 @@ namespace System.Security.Cryptography.X509Certificates
             return Export(contentType, password: null);
         }
 
-        public byte[] Export(X509ContentType contentType, String password)
+        public byte[] Export(X509ContentType contentType, string password)
         {
             using (IStorePal storePal = StorePal.LinkFromCertificateCollection(this))
             {
@@ -124,7 +116,7 @@ namespace System.Security.Cryptography.X509Certificates
             }
         }
 
-        public X509Certificate2Collection Find(X509FindType findType, Object findValue, bool validOnly)
+        public X509Certificate2Collection Find(X509FindType findType, object findValue, bool validOnly)
         {
             if (findValue == null)
                 throw new ArgumentNullException("findValue");
@@ -139,8 +131,7 @@ namespace System.Security.Cryptography.X509Certificates
 
         public new X509Certificate2Enumerator GetEnumerator()
         {
-            X509CertificateEnumerator baseEnumerator = base.GetEnumerator();
-            return new X509Certificate2Enumerator(baseEnumerator);
+            return new X509Certificate2Enumerator(this);
         }
 
         public void Import(byte[] rawData)
@@ -148,7 +139,7 @@ namespace System.Security.Cryptography.X509Certificates
             Import(rawData, password: null, keyStorageFlags: X509KeyStorageFlags.DefaultKeySet);
         }
 
-        public void Import(byte[] rawData, String password, X509KeyStorageFlags keyStorageFlags)
+        public void Import(byte[] rawData, string password, X509KeyStorageFlags keyStorageFlags)
         {
             if (rawData == null)
                 throw new ArgumentNullException("rawData");
@@ -159,12 +150,12 @@ namespace System.Security.Cryptography.X509Certificates
             }
         }
 
-        public void Import(String fileName)
+        public void Import(string fileName)
         {
             Import(fileName, password: null, keyStorageFlags: X509KeyStorageFlags.DefaultKeySet);
         }
 
-        public void Import(String fileName, String password, X509KeyStorageFlags keyStorageFlags)
+        public void Import(string fileName, string password, X509KeyStorageFlags keyStorageFlags)
         {
             if (fileName == null)
                 throw new ArgumentNullException("fileName");
@@ -222,10 +213,9 @@ namespace System.Security.Cryptography.X509Certificates
             int i = 0;
             try
             {
-                foreach (X509Certificate2 certificate in certificates)
+                for (; i < certificates.Count; i++)
                 {
-                    Remove(certificate);
-                    i++;
+                    Remove(certificates[i]);
                 }
             }
             catch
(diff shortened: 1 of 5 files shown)


stephentoub:
@dotnet-bot test this please

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
```diff
@@ -105,9 +97,6 @@ public void AddRange(X509Certificate2Collection certificates)
 
         public bool Contains(X509Certificate2 certificate)
         {
-            if (certificate == null)
```
bartonjs: It might be worth a comment reminding someone not to put the nullcheck back.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509CertificateCollection.cs
```diff
@@ -68,33 +52,26 @@ bool IList.IsReadOnly
...
             }
             set
             {
-                if (index < 0 || index >= Count)
-                    throw new ArgumentOutOfRangeException("index", SR.ArgumentOutOfRange_Index);
                 if (value == null)
                     throw new ArgumentNullException("value");
 
-                _list[index] = value;
+                ((IList)_list)[index] = value;
```
bartonjs: Why use the IList version? The only difference I see is that the IList version throws on null arguments, but you left the null-check here.

Oh... right, the IList version takes `object`, but the `List<T>` version takes a `T`.

A comment to the effect of "List already has an IList indexer which handles the type and range checks, let it do all the work rather than have it replicated here" would save on some thinking for the next guy.

bartonjs: Or, I guess the answer is an easy "Because this way it compiles"; and it's only a problem of reading the diff.

I wonder why the indexer is up here when the rest of the IList explicit implementations are later... because when you see lots of the casting it makes more sense :).

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
bartonjs: Since you added a test, someone would have to go really out of their way to put it back; so probably not worth messing with the commit integrity.

bartonjs:
The code and behavior changes seem reasonable to me.  Your call, @weshaggard.

weshaggard:
I think the behavior changes are reasonable (i.e. changes in using IList and the case where you now throw ArgumentNullException and the place we now don't throw it for the Contains).

The changes look good to me I only had the one question about using the List property instead of casting to IList everywhere. I don't know if make much of a difference performance wise but it would help with the readability of the code.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509CertificateCollection.cs
justinvp: > I wonder why the indexer is up here when the rest of the IList explicit implementations are later.

I had the same thought, and considered moving it down, but didn't want to make the diff harder to read.

justinvp:
> Why not use the List property in these cases instead of casting?

The `List` property is unusual in that it returns `this` *not* the underlying `_list`. If I changed those places to use the `List` property, they'd just be calling themselves recursively until it stack overflows. Before I submitted the PR, I had originally changed the `List` property to return the underlying `_list` and be typed as `List<object>` (and then later `List<X509Certificate>`), and was using it everywhere I needed the list. But this ended up resulting in a much larger diff as the places that were using the `List` property were counting on it throwing `ArgumentNullException` where it needed to. I had to update a lot more places to throw `ArgumentNullException` (e.g. the indexer, `Add`, `Insert`, `Remove`, for both the public methods and non-generic explicit versions, across both collections). Then to simplify that, I added a bunch of private "internal" Add/Insert/Remove methods that were the single place the argument null checking was happening, and update all the relevant callers to call those. But then, with the change to hardening to only accept `X509Certificate` objects, the explicit implementations of `IList` need to call the non-generic, explicitly implemented implementations on the underlying `_list` to get the `ArgumentException` behavior we want for bogus types. In the end, it just made the diff look much crazier.

All that said, I still think there's a way to avoid the casts for the public methods and simplify things, so I'll take another crack at it.

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2Collection.cs
justinvp: I added a comment here and in the tests.

stephentoub:
LGTM, Justin, thanks.

One additional thought: I took a look at the code coverage report for your PR, and although you added a bunch of tests (thanks!), some of the code paths you touched don't have any coverage.  Would you mind adding a few more tests to take care of that?  I don't expect you'll find any problems, but just in case...

justinvp:
While we're already discussing these collections, there's one more minor behavior change that I think would be worth making for .NET Core... `X509CertificateEnumerator.Current` and `X509Certificate2Enumerator.Current` are both documented as throwing `InvalidOperationException` when the enumerator is positioned before the first element of the collection or after the last element. This aligns with the documented behavior of `IEnumerator.Current`, but is inconsistent with the documented behavior of `IEnumerator<T>.Current`, which states that `Current` is undefined in these cases (but does not need to throw).

I went ahead and added another commit that changes the behavior of the public `Current` to align with `IEnumerator<T>.Current`, to no longer throw for these cases. The explicit implementations of `IEnumerator.Current` remain the same (they will throw). This has a minor perf improvement when accessing the public `Current` and aligning with `IEnumerator<T>.Current` paves the way for eventually implementing the generic interfaces at some point in the future.

@weshaggard, do you think this is a reasonable change? If not, I will remove this commit.

@weshaggard, regarding your previous comment about using the `List` property to avoid the casting, I cleaned things up as follows:

 - The `List` property is now private, renamed `NonGenericList`, and returns the underlying `_list`. Explicit `IList` and `ICollection` implementations now use this instead of casting themselves.
 - The public methods can just call `_list` directly, avoiding any casts. `X509Certificate2Collection` calls through to the `base` methods. (The null checks remain in `X509Certificate2Collection because the parameters are named differently--`certificate` vs `value`).
 - Null checks use a common helper `Requires.NotNull(...)` just to streamline the throwing of `ArgumentNullException` within these collections.

@stephentoub, I added more tests (didn't find any problems). Coverage is now 100% for `X509CertificateCollection` and the enumerators, and
99.3% for `X509Certificate2Collection`. I wanted to get `X509Certificate2Collection` to 100%, but the final thing that needs more coverage is around importing a file (unrelated to my changes), and I didn't have an appropriate test file readily available to make use of.

stephentoub:
Thanks, Justin. LGTM.

@weshaggard, you ok with the final commit?

src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509CertificateCollection.cs
```diff
@@ -173,68 +146,83 @@ public override int GetHashCode()
...
-        internal IList List
+        internal void GetEnumerator(out List<X509Certificate>.Enumerator enumerator)
         {
-            get { return this; }
+            enumerator = _list.GetEnumerator();
         }
 
-        private List<Object> _list;
-        private Object _syncRoot;
+        internal static class Requires
```
weshaggard: I'm not sure I like introducing this Requires class, even if we do add such a helper, which I'm not convinced we should do it should go somewhere under Common.  What value do you believe this brings?

justinvp: The reason I added it was because I was adding the same "not null" check to a bunch of methods in these collections. If you'd rather just go with duplicating the if checks, that's fine with me, I don't have a preference either way.

Even if we keep it, I'm not sure it should be Common... Seems like each corefx library has its own take on these precondition helpers. I chose the variant that seemed most common. Is it worth standardizing on one, and if so, would all libraries eventually be updated to use the common one?

bartonjs: I'm with @weshaggard on this one; I'm not sure the helper really adds much.  If there were a unified helper-for-throwing pattern that was consistent across corefx (or even the assembly) that'd be one thing; but this is something only consistent across these two classes.

justinvp:
I removed the helper.

Wes, you OK with the behavior change in the last commit?

nguerrera:
IMHO, Current being "undefined" means it can return anything or throw. Callers should just never call Current in that situation so it doesn't matter. I dislike the differences between the non-generic and generic interface docs.

For new code, I'm fine with IEnumerator<T>.Current doing whatever is optimal and IEnumerator.Current just calling Enumerator<T>.Current. We should update the docs to expressly allow it.

However, I wouldn't bother changing the behavior for existing classes for this.

weshaggard:
@nguerrera just bet me too the same conclusion, I think we should keep the Current property as it was and undo your last commit. I agree with Nick that I much prefer consistency between the to implementations and "undefined" can throw.

The rest of the changes look good to me. Thanks @justinvp.

justinvp:
OK, I removed the last commit (CI still running).

This means we have to go out of our way to maintain the existing behavior of throwing: both the public `Current` and explicitly implemented `IEnumerator.Current` have to call `List<T>.Enumerator`'s non-generic `IEnumerator.Current` implementation through the generic helper (to avoid boxing since `List<T>.Enumerator` explicitly implements `IEnumerator.Current`) and cast the return type for both collections. The public `List<T>.Enumerator.Current` can't be used here because it doesn't throw. (Or, alternatively, call the public `List<T>.Enumerator.Current`, but add/modify a field and add code to do the before/after->throw check ourselves).

stephentoub:
Thanks, Justin!

PR #2825, opened 2015-08-15
Deleting System.Runtime test failing after sync primitives change

stephentoub:
One System.Runtime test is testing the disposal of a Mutex.  Besides that already being covered by multiple System.Threading tests, the test is creating the Mutex with a name, which after recent mscorlib changes fails on Unix, and is preventing successful CI runs.  I'm simply deleting the test.


stephentoub:
Merging without review as it's simply deleting a test, and without this nothing will pass CI builds.

PR #2829, opened 2015-08-15
Fix System.IO.Pipes test on 64-bit

### LINKED ISSUE #2818: Pipes tests fail on x64
I hit this when integrating a new drop of build tools, I've temporarily rolled the tests back to x86 until this is fixed.

```
  Starting:    System.IO.Pipes.Tests
     System.IO.Pipes.Tests.NamedPipeTest_Specific.Windows_MessagePipeTransissionMode [FAIL]
        System.IO.IOException : The operation completed successfully
        Stack Trace:
           E:\dotnet-bot\corefx\src\System.IO.Pipes\src\System\IO\Pipes\PipeStream.Windows.cs(969,0): at System.IO.Pipes.PipeStream.WinIOError(Int32 errorCode)
           E:\dotnet-bot\corefx\src\System.IO.Pipes\src\System\IO\Pipes\NamedPipeServerStream.Windows.cs(152,0): at System.IO.Pipes.NamedPipeServerStream.GetImpersonationUserName()
           E:\dotnet-bot\corefx\src\System.IO.Pipes\tests\NamedPipeTests\NamedPipeTest.Specific.cs(129,0): at System.IO.Pipes.Tests.NamedPipeTest_Specific.Windows_MessagePipeTransissionMode()
  Finished:    System.IO.Pipes.Tests
```

stephentoub:
A test is trying to use NamedPipeServerStream.GetImpersonationUserName and is failing when run in a 64-bit process.  The client stream is being created with TokenImpersonalLevel.Identification, but MSDN states that the underlying GetNamedPipeHandleState being used to get the name will only allow retrieval of the name if SECURITY_IMPERSONATION is used, e.g. TokenImpersonationLevel.Impersonation.  Changing the TokenImpersonationLevel allows the test to pass, on both 32-bit and 64-bit.  This commit does that.

Fixes #2818 
cc: @pallavit, @ericstj

What's strange is that the test passes as-is on 32-bit with Identification rather than Impersonation.  This is true on desktop as well, e.g. this program runs fine in a 32-bit process but fails with an exception in a 64-bit process:
```C#
using System;
using System.IO.Pipes;
using System.Security.Principal;
using System.Threading.Tasks;

class Program
{
    static void Main()
    {
        string pipeName = Guid.NewGuid().ToString("N");
        using (var server = new NamedPipeServerStream(pipeName, PipeDirection.InOut))
        using (var client = new NamedPipeClientStream(".", pipeName, PipeDirection.InOut, PipeOptions.None, TokenImpersonationLevel.Identification))
        {
            Task t = server.WaitForConnectionAsync();
            client.Connect();
            t.Wait();

            Console.WriteLine(server.GetImpersonationUserName()); // succeeds 32-bit, fails 64-bit
        }
    }
}
```
Something to follow-up on separately.

### DIFF
--- a/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
+++ b/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Runtime.InteropServices;
+using System.Security.Principal;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
@@ -89,7 +90,7 @@ namespace System.IO.Pipes.Tests
 
             using (NamedPipeServerStream server = new NamedPipeServerStream(pipeName, PipeDirection.InOut, 1, PipeTransmissionMode.Message))
             {
-                using (NamedPipeClientStream client = new NamedPipeClientStream(".", pipeName, PipeDirection.InOut, PipeOptions.None, Security.Principal.TokenImpersonationLevel.Identification))
+                using (NamedPipeClientStream client = new NamedPipeClientStream(".", pipeName, PipeDirection.InOut, PipeOptions.None, TokenImpersonationLevel.Impersonation))
                 {
                     server.ReadMode = PipeTransmissionMode.Message;
                     Assert.Equal(PipeTransmissionMode.Message, server.ReadMode);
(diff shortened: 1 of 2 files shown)


ericstj:
The weird thing was that the test was throwing an error on ERROR_SUCCESS, it made me think we actually had a PInvoke signature incorrect somewhere in the product code.  That's a more serious bug than the test just passing an invalid arg.

That said, I agree that the MSDN docs are straightforward on this.  I'm OK with this change.

stephentoub:
> it made me think we actually had a PInvoke signature incorrect somewhere in the product code

That was my initial thought as well, but I scrutinized it and couldn't find anything wrong.  So I wrote out a server and client in C++, and I was able to reproduce the same behavior: passing SECURITY_IMPERSONATION works for both x86 and x64, but passing SECURITY_IDENTIFICATION only works for x86... for x64, GetNamedPipeHandleStateW returns 0 but then subsequently so does GetLastError.  So this isn't a problem with the .NET code.

PR #2830, opened 2015-08-15
Avoid test deadlocks in threading WaitHandle tests

stephentoub:
A bunch of the Mutex/EventWaitHandle/Semaphore tests are waiting indefinitely for a signal.  If that signal never arrives due to the test failing, CI will end up waiting for several hours and then aborting the whole run.  This change adds timeouts to those waits.  It also changes a name used in one of the semaphore tests to be randomly generated; this should hopefully fix a pervasive failure we're seeing in CI right now.

Edit: updated with an additional commit.  The timeouts added in the first commit identified the failing test in CI, and the second commit should fix the hang.

### DIFF
--- a/src/System.Threading/tests/AutoResetEventTests.cs
+++ b/src/System.Threading/tests/AutoResetEventTests.cs
@@ -8,6 +8,8 @@ using System.Threading.Tasks;
 
 public class AutoResetEventTests
 {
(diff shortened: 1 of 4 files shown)


bartonjs:
LGTM.  Thanks for driving down the instability.

PR #2831, opened 2015-08-15
Remove redundant check.

hackcraft:
Simple removal of a check-and-branch that has the same nett result in either
case.

Also added a comment to warn off anyone tempted to do the same dangerous
optimisation I attempted in #2809

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/DefaultExpression.cs
@@ -58,7 +58,7 @@ namespace System.Linq.Expressions
         /// </returns>
         public static DefaultExpression Empty()
         {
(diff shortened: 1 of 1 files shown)


stephentoub:
@dotnet-bot test this please

stephentoub:
LGTM

cc: @VSadov

hackcraft:
BTW, does that comment to the bot work for anyone? There's been a couple of times there's been a failure that seemed to be unrelated to the actual commit, and I've wondered if there was a way to trigger another test.

stephentoub:
> does that comment to the bot work for anyone? 

It does. We had a bunch of issues since yesterday that caused lots of PR failures, where the change to a new tool set, updating dependencies, and running tests as 64-bit exposed some latent issues, but I believe I've now fixed them all (or at least the most common problems) and things should be running smoothly again.

VSadov:
LGRM

PR #2834, opened 2015-08-17
Use Stat from System.Native shim throughout

nguerrera:
Replace all usage of GetFileInformation from libcoreclr with newer
Stat from System.Native shim.


nguerrera:
cc @stephentoub

stephentoub:
LGTM. Thanks.

PR #2835, opened 2015-08-17
Fix behavior and test for X509Store.Add when store is readonly.

bartonjs:
Despite what was concluded from the earlier testing, Windows does actually report an access denied error when calling X509Store.Add on a store instance that was opened as read-only.  This fixes the test, and the Unix implementation.


bartonjs:
Either I didn't run it after the last time I made changes to it, or I had invalidated my own test conditions; but after pulling down the newest master and building clean I failed my own test deterministically.  Good old HR `0x80070005` (`E_ACCESSDENIED`).

I might be the only person with a cert sans private key in my CurrentUser\My store, so maybe no one would ever see the failure; but it was wrong nonetheless.

Verified this behavior against CoreFX\Windows and full framework.

nguerrera:
LGTM

PR #2839, opened 2015-08-17
Additional String.Join tests

bbowyersmyth:
Added more test cases for:

    String.Join(String, String[])
    String.Join(String, String[], int, int)
    String.Join(String, Object[])
    String.Join(String, IEnumerable<String>)
    String.Join(String, IEnumerable<T>)

### DIFF
--- a/src/System.Runtime/tests/System/String.cs
+++ b/src/System.Runtime/tests/System/String.cs
@@ -704,11 +704,17 @@ public static unsafe class StringTests
         String s;
 
         // String Array
+        s = String.Join("$$", new String[] { }, 0, 0);
+        Assert.Equal("", s);
+
         s = String.Join("$$", new String[] { null }, 0, 1);
         Assert.Equal("", s);
 
         s = String.Join("$$", new String[] { null, "Bar", null }, 0, 3);
         Assert.Equal("$$Bar$$", s);
+
+        s = String.Join(null, new String[] { "Foo", "Bar", "Baz" }, 0, 3);
+        Assert.Equal("FooBarBaz", s);
 
         s = String.Join("$$", new String[] { "Foo", "Bar", "Baz" }, 0, 3);
         Assert.Equal("Foo$$Bar$$Baz", s);
@@ -730,11 +736,17 @@ public static unsafe class StringTests
         Assert.Throws<ArgumentOutOfRangeException>(() => s = String.Join("$$", new String[] { "Foo" }, 2, 1));
 
         // Object Array
+        s = String.Join("@@", new object[] { });
+        Assert.Equal("", s);
+
         s = String.Join("@@", new object[] { "Red" });
         Assert.Equal("Red", s);
 
         s = String.Join("@@", new object[] { "Red", "Green", "Blue" });
         Assert.Equal("Red@@Green@@Blue", s);
+
+        s = String.Join(null, new object[] { "Red", "Green", "Blue" });
+        Assert.Equal("RedGreenBlue", s);
 
         s = String.Join("@@", new object[] { null, "Green", "Blue" }); // Feature of object[] overload to exit if [0] is null
         Assert.Equal("", s);
@@ -754,6 +766,9 @@ public static unsafe class StringTests
         s = String.Join("|", new List<string>() { "Red" });
         Assert.Equal("Red", s);
 
+        s = String.Join(null, new List<string>() { "Red", "Green", "Blue" });
+        Assert.Equal("RedGreenBlue", s);
+
         s = String.Join("|", new List<string>() { "Red", "Green", "Blue" });
         Assert.Equal("Red|Green|Blue", s);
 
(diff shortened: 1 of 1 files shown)


src/System.Runtime/tests/System/String.cs
```diff
@@ -702,6 +703,13 @@ public static void TestJoin()
     {
         String s;
 
+        // String Array
+        s = String.Join("$$", new String[] { null }, 0, 1);
+        Assert.True(s == "");
+
+        s = String.Join("$$", new String[] { null, "Bar", null }, 0, 3);
+        Assert.True(s == "$$Bar$$");
```
akoeplinger: Instead of `Assert.True` you can use `Assert.Equal` here and below. It prints a much better message when the test case fails.

bbowyersmyth: Sure thing. Updated this file to be consistent.

src/System.Runtime/tests/System/String.cs
```diff
@@ -19,7 +20,7 @@ public static void TestLastIndexOfWithEmptyString()
     {
         string s = "Dill Guv Dill Guv Dill";
         int lastIndex = s.LastIndexOf("", StringComparison.OrdinalIgnoreCase);
-        Assert.True(lastIndex == s.Length - 1);
+        Assert.Equal(lastIndex, s.Length - 1);
```
stephentoub: Thanks for changing all of these to be Equal instead of True.  However, the first argument to Equal is meant to be the expected value, the second the actual value, so most (all?) of these need to have their arguments swapped.  Otherwise, if the assert fails, the error message that gets output ends up being confusing.  Could you swap them where appropriate as well?

src/System.Runtime/tests/System/String.cs
```diff
@@ -702,22 +703,82 @@ public static void TestJoin()
...
-        Object[] o = { "Red", "Green", "Blue" };
-        s = String.Join("@@", o);
-        Assert.True(s == "Red@@Green@@Blue");
+        s = String.Join("--", new List<Object>() { null, "Green", null });
+        Assert.Equal(s, "--Green--");
 
-        String[] ss = { "Red", "Green", "Blue" };
-        s = String.Join("@@", ss);
-        Assert.True(s == "Red@@Green@@Blue");
+        Assert.Throws<ArgumentNullException>(() => s = String.Join("--", (IEnumerable<Object>)null));
```
stephentoub: Thanks for the extra tests.  While you're adding these, could you also cover a few more cases, e.g.:
- That Join throws when the separator string is null
- That Join works correctly when an empty array/enumerable is supplied

stephentoub:
Thanks for the updates, @bbowyersmyth.  I know you've submitted changes to the corefx repo in the past, but the bot is saying that you still need to sign the CLA.  Maybe your previous changes were small enough that this wasn't needed?  Or maybe you've changed your account in the interim or something?

bbowyersmyth:
@stephentoub The CLA was signed on the 1st of Jan (probably your 31st Dec). I can forward you the confirmation email if that helps match things up?

stephentoub:
@joshfree, can you help sort out why the CLA bot is marking @bbowyersmyth's PRs with "CLA required"?

bbowyersmyth:
@stephentoub @joshfree It does say it is for project "ASP.NET MVC". Is another CLA required for corefx or is this for all MS Open Tech?

stephentoub:
@bbowyersmyth, ah, thanks, I believe that's a different CLA.  You need to complete the one specifically mentioned in the comment from the bot earlier.  Thanks!

bbowyersmyth:
@stephentoub Ok will do

src/System.Runtime/tests/System/String.cs
bbowyersmyth: Swapped the arguments + updated some of the existing ones that were in the wrong order.

src/System.Runtime/tests/System/String.cs
bbowyersmyth: null separator is a valid value. Added tests for those.
Added empty array tests. Empty enumerables were there.

joshfree:
@stephentoub @bbowyersmyth yup - everyone signs the specific CLA for dotnetfoundation; this will cover you for /dotnet/corefx, /dotnet/coreclr, /dotnet/wcf...  /Microsoft/* may use different CLAs.

src/System.Runtime/tests/System/String.cs
stephentoub: > null separator is a valid value

Ah, right, thanks, I forgot that gets normalized to an empty string.

stephentoub:
Thanks, @bbowyersmyth.  There's probably a more consolidated way to write some of these new tests (e.g. with theories), but the style you used matches what's already there, so it's fine to stick with for now.  LGTM.

bbowyersmyth:
Yeah I tried to match the style that was there.

PR #2840, opened 2015-08-17
Enabling POST method options

shrutigarg:


shrutigarg:
cc: @stephentoub, @davidsh ,@CIPop, @SidharthNabar , @pgavlin

shrutigarg:
cc: @vijaykota

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -434,14 +434,24 @@ private SafeCurlHandle CreateRequestHandle(RequestCompletionSource state, GCHand
                 // Set maximum automatic redirection option
                 SetCurlOption(requestHandle, CURLoption.CURLOPT_MAXREDIRS, _maxAutomaticRedirections);
             }
-            if (state.RequestMessage.Content != null)
+
+            if (state.RequestMessage.Method == HttpMethod.Put)
```
stephentoub: The previous code used Content != null here.  Is that expected for this case?  Should we assert it?

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -434,14 +434,24 @@ private SafeCurlHandle CreateRequestHandle(RequestCompletionSource state, GCHand
...
             {
                 SetCurlOption(requestHandle, CURLoption.CURLOPT_UPLOAD, 1L);
             }
-            if (state.RequestMessage.Method == HttpMethod.Head)
+            else if (state.RequestMessage.Method == HttpMethod.Head)
             {
                 SetCurlOption(requestHandle, CURLoption.CURLOPT_NOBODY, 1L);
             }
+            else if (state.RequestMessage.Method == HttpMethod.Post)
```
stephentoub: Rather than this chain of if/elses, did you consider a switch?

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
hackcraft: PUT or POST without a body is valid (and equivalent to sending a zero-octet message) though people have coded to the assumption that it's not valid in the past. On the one hand this could mean null content with zero-length content would be safer, but on the other hand there could well be a server out there that (incorrectly) treats no body as allowed by different to zero-octet body, and such a "safety" mechanism would prevent working to be bug-compatible with that server.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
shrutigarg: You think its better to use switch in this case where all "if"s in this method are not a part of a single condition ?

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
shrutigarg: Curl do support a Put/Post with no content , so the change here is to follow that. checking content if null or not, is not correct here.

vijaykota: @stephentoub , just to clarify, the earlier check for non-null Content was a bug

stephentoub: Ok, thanks.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
stephentoub: I think that's fine, especially if we're eventually going to have more conditions here. I personally like:
```C#
switch (state.RequestMessage.Method)
{
    case HttpMethod.Put:
       ...
    case HttpMethod.Head:
        ...
    case HttpMethod.Post:
        ...
}
```
better than:
```C#
if (state.RequestMessage.Method == HttpMethod.Put)
{
    ...
}
else if (state.RequestMessage.Method == HttpMethod.Head)
{
    ...
}
else if (state.RequestMessage.Method == HttpMethod.Post)
{
    ...
}
```
but it's not a big deal.  Whichever you prefer.

shrutigarg: Changing it to  Switch as per your recommendation.

shrutigarg: Tried changing to switch as per your recommendation but getting error " A switch expression or case label must be a bool, char, string, integral, enum, or corresponding nullable type"
while compiling. Seems i cant apply switch statement on HttpMethod Type.

shrutigarg:
@dotnet-bot test this please

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
stephentoub: I just assumed HttpMethod was an enum. Sorry for misleading.

PR #2841, opened 2015-08-17
Add more pipe tests

### LINKED ISSUE #2640: [IO.Pipes] Outer cancellation of WaitForConnectionAsync/ReadAsync/WriteAsync
Consider the following code
```c#
[DllImport("api-ms-win-core-io-l1-1-0.dll", SetLastError = true)]
internal static unsafe extern bool CancelIoEx(SafeHandle handle, NativeOverlapped* lpOverlapped);

public static async Task CanBeCanceled_False()
{
    using (NamedPipeServerStream server = new NamedPipeServerStream("unique3", PipeDirection.InOut, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous, 0, 0))
    {
        Task waitForConnectionTask = server.WaitForConnectionAsync(CancellationToken.None); // cannot be canceled
       
        unsafe
        {
            CancelIoEx(server.SafePipeHandle, null);
        }

        await waitForConnectionTask; // throws OperationCanceledException
    }
}

public static async Task CanBeCanceled_True()
{
    using (NamedPipeServerStream server = new NamedPipeServerStream("unique3", PipeDirection.InOut, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous, 0, 0))
    {
        var ctx = new CancellationTokenSource();
        Task waitForConnectionTask = server.WaitForConnectionAsync(ctx.Token); // could be canceled
       
        unsafe
        {
            CancelIoEx(server.SafePipeHandle, null);
        }

        await waitForConnectionTask; // throws IOException: IO operation was aborted unexpectedly.
    }
}
```

Is it intentional behavior or it's a bug that we receive `OperationCanceledException` even if we couldn't cancel in a normal way (without interop) ?

Maxwe11:
closes #2640 as wontfix

### DIFF
--- a/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
+++ b/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
@@ -51,6 +51,33 @@ namespace System.IO.Pipes.Tests
         }
 
         [Theory]
+        [InlineData(PipeDirection.In)]
+        [InlineData(PipeDirection.InOut)]
+        [InlineData(PipeDirection.Out)]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void ReservedPipeName_Throws_ArgumentOutOfRangeException(PipeDirection direction)
+        {
+            const string serverName = ".";
+            const string reservedName = "anonymous";
+            Assert.Throws<ArgumentOutOfRangeException>("pipeName", () => new NamedPipeClientStream(reservedName));
+            Assert.Throws<ArgumentOutOfRangeException>("pipeName", () => new NamedPipeClientStream(serverName, reservedName));
+            Assert.Throws<ArgumentOutOfRangeException>("pipeName", () => new NamedPipeClientStream(serverName, reservedName, direction));
+            Assert.Throws<ArgumentOutOfRangeException>("pipeName", () => new NamedPipeClientStream(serverName, reservedName, direction, PipeOptions.None));
+            Assert.Throws<ArgumentOutOfRangeException>("pipeName", () => new NamedPipeClientStream(serverName, reservedName, direction, PipeOptions.None, TokenImpersonationLevel.Impersonation));
+        }
+
+        [Fact]
+        [PlatformSpecific(PlatformID.AnyUnix)]
+        public static void NotSupportedPipePath_Throws_PlatformNotSupportedException()
+        {
+            string hostName;
+            Assert.True(Interop.TryGetHostName(out hostName));
+
+            Assert.Throws<PlatformNotSupportedException>(() => new NamedPipeClientStream("foobar" + hostName, "foobar"));
+            Assert.Throws<PlatformNotSupportedException>(() => new NamedPipeClientStream(hostName, "foobar" + Path.GetInvalidFileNameChars()[0]));
+        }
+
+        [Theory]
         [InlineData((PipeDirection)123)]
         public static void InvalidPipeDirection_Throws_ArgumentOutOfRangeException(PipeDirection direction)
         {
--- a/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
+++ b/src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
@@ -235,23 +235,16 @@ namespace System.IO.Pipes.Tests
             }
         }
 
-        [DllImport("api-ms-win-core-io-l1-1-0.dll", SetLastError = true)]
-        private static unsafe extern bool CancelIoEx(SafeHandle handle, NativeOverlapped* lpOverlapped);
-
         [Fact]
         [PlatformSpecific(PlatformID.Windows)]
-        // [ActiveIssue(2640)]
         public async Task CancelTokenOff_ServerWaitForConnectionAsyncWithOuterCancellation_Throws_OperationCanceledException()
         {
             using (NamedPipePair pair = CreateNamedPipePair())
             {
                 NamedPipeServerStream server = pair.serverStream;
                 Task waitForConnectionTask = server.WaitForConnectionAsync(CancellationToken.None);
-                unsafe
-                {
-                    Assert.True(CancelIoEx(server.SafePipeHandle, null), "Outer cancellation failed");
-                }
 
+                Assert.True(Interop.CancelIoEx(server.SafePipeHandle), "Outer cancellation failed");
                 await Assert.ThrowsAnyAsync<OperationCanceledException>(() => waitForConnectionTask);
                 Assert.True(waitForConnectionTask.IsCanceled);
             }
@@ -259,7 +252,6 @@ namespace System.IO.Pipes.Tests
 
         [Fact]
         [PlatformSpecific(PlatformID.Windows)]
-        // [ActiveIssue(2640)]
         public async Task CancelTokenOn_ServerWaitForConnectionAsyncWithOuterCancellation_Throws_IOException()
         {
             using (NamedPipePair pair = CreateNamedPipePair())
@@ -267,13 +259,9 @@ namespace System.IO.Pipes.Tests
                 var cts = new CancellationTokenSource();
                 NamedPipeServerStream server = pair.serverStream;
                 Task waitForConnectionTask = server.WaitForConnectionAsync(cts.Token);
-                unsafe
-                {
-                    Assert.True(CancelIoEx(server.SafePipeHandle, null), "Outer cancellation failed");
-                }
 
+                Assert.True(Interop.CancelIoEx(server.SafePipeHandle), "Outer cancellation failed");
                 await Assert.ThrowsAsync<IOException>(() => waitForConnectionTask);
-                Assert.False(waitForConnectionTask.IsCanceled);
             }
         }
 
@@ -310,6 +298,7 @@ namespace System.IO.Pipes.Tests
 
                 Assert.Throws<InvalidOperationException>(() => server.Flush());
                 Assert.Throws<InvalidOperationException>(() => server.IsMessageComplete);
+                Assert.Throws<InvalidOperationException>(() => server.GetImpersonationUserName());
             }
         }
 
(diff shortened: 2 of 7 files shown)


src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -50,6 +51,38 @@ public static void EmptyStringServerName_Throws_ArgumentException(PipeDirection
             Assert.Throws<ArgumentException>(() => new NamedPipeClientStream("", "client1", direction, PipeOptions.None, TokenImpersonationLevel.None));
         }
 
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void ReservedPipePath_Throws_ArgumentOutOfRangeException()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => new NamedPipeClientStream(".", "anonymous"));
```
stephentoub: The server version of this test has "anonymous" in a local const and tests other overloads of NamedPipeClientStream's ctor as well.  It'd be nice to be consistent.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -50,6 +51,38 @@ public static void EmptyStringServerName_Throws_ArgumentException(PipeDirection
...
+            const int HOST_NAME_MAX = 255; // man gethostname
+            const int ArrLength = HOST_NAME_MAX + 1;
+
+            int err;
+            string hostName;
+
+            unsafe
+            {
+                byte* name = stackalloc byte[ArrLength];
+                err = gethostname(name, ArrLength);
```
stephentoub: We shoudl Assert.Equal(0, err), right?

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -513,6 +520,82 @@ public void OperationsOnUnconnectedClient()
...
+                byte[] buffer = new byte[] { 0, 0, 0, 0 };
+
+                pair.Connect();
+
+                if (server.CanRead)
+                {
+                    Task serverReadToken = server.ReadAsync(buffer, 0, buffer.Length, CancellationToken.None);
+                    unsafe
+                    {
+                        Assert.True(CancelIoEx(server.SafePipeHandle, null), "Outer cancellation failed");
```
stephentoub: Why are you using Assert.True rather than Assert.Equal / Assert.Null / etc. for a bunch of these asserts?

Maxwe11: `Assert.True` has message argument

stephentoub:
cc: @ianhays

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
stephentoub: Nevermind; I misread this code... I missed that you were asserting on the result of CancelIoEx... I thought you were asserting that server.SafePipeHandle == null.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.CreateClient.cs
```diff
@@ -51,6 +52,44 @@ public static void EmptyStringServerName_Throws_ArgumentException(PipeDirection
...
+        public static void NotSupportedPipePath_Throws_PlatformNotSupportedException()
+        {
+            const int HOST_NAME_MAX = 255; // man gethostname
+            const int ArrLength = HOST_NAME_MAX + 1;
+            string hostName;
+
+            unsafe
+            {
+                byte* name = stackalloc byte[ArrLength];
+                Assert.True(gethostname(name, ArrLength) == 0, "gethostname failed");
```
stephentoub: Why not Assert.Equal here?  This seems like a typical case where the message doesn't add additional value.

src/System.IO.Pipes/tests/PipeTest.Write.cs
```diff
@@ -133,6 +133,23 @@ public virtual void ReadOnWriteOnlyPipe_Throws_NotSupportedException()
...
+            {
+                PipeStream pipe = pair.writeablePipe;
+
+                // Shouldn't throw
+                pipe.Write(Array.Empty<byte>(), 0, 0);
+
+                Task writeTask = pipe.WriteAsync(Array.Empty<byte>(), 0, 0);
+                await writeTask;
+
+                Assert.Equal(TaskStatus.RanToCompletion, writeTask.Status);
```
stephentoub: This assert is superfluous.  If writeTask ended as anything other than RanToCompletion, we'll never get to this assert because the await on writeTask will throw.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -540,6 +623,80 @@ public void OperationsOnUnconnectedClient()
...
+                {
+                    var cts = new CancellationTokenSource();
+                    Task clientWriteToken = client.WriteAsync(buffer, 0, buffer.Length, cts.Token);
+                    unsafe
+                    {
+                        Assert.True(CancelIoEx(client.SafePipeHandle, null), "Outer cancellation failed");
+                    }
+
+                    await Assert.ThrowsAnyAsync<IOException>(() => clientWriteToken);
+                    Assert.False(clientWriteToken.IsCanceled);
```
stephentoub: This assert is really a test of Tasks, not of pipes.  The previous await Assert.ThrowsAnyAsync could only complete successfully if clientWriteToken ended as Faulted, such that Assert.False(clientWriteToken.IsCanceled) can never fail.

stephentoub: (I see a few more cases of this; I won't comment on those.)

Maxwe11: I intentionally add this Assert for catching bugs like this one:

    var tcs = new TaskCompletionSource<int>();
    tcs.SetException(new OperationCanceledException());
    // not tcs.Task.IsCanceled is false

stephentoub: I'm not following.  In what situation could the ```await Assert.ThrowsAnyAsync<IOException>(() => clientWriteToken);`` pass successfully but then the ```Assert.False(clientWriteToken.IsCanceled)``` could fail?  In the code snippet you shared, we'd never get to the Assert.False because the Assert.ThrowsAnyAsync would fail due to the thrown OperationCanceledException not being an IOException.

Maxwe11: Make sense, will remove this Assert.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -473,12 +484,14 @@ public void OperationsOnUnconnectedClient()
         {
             using (NamedPipePair pair = CreateNamedPipePair())
             {
+                pair.Connect();
```
ianhays: This seems like two separate but very similar scenarios:

1. Make a connected pair, dispose the client, then try to perform operations on it.
2. Make a clientstream, dispose it, then try to perform operations on it.

I suppose, though, that the behavior is the same in either case and the connected case is more specific so we should be fine to just use that one. For example, pipe.Connect() throws ObjectDisposedException in either case but it would throw InvalidOperationException if it wasnt disposed but was connected, so having the connected case shows that ObjectDisposedException is going to be thrown before InvalidOperationException.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
```diff
@@ -118,16 +128,42 @@ public void Windows_MessagePipeTransissionMode()
...
+                    }
 
-                    int len4 = server.Read(received3, len3, received3.Length - len3);
+                    expectedRead = msg1.Length - expectedRead;
+                    Assert.Equal(expectedRead, server.Read(received3, len3, expectedRead));
                     Assert.True(server.IsMessageComplete);
-                    Assert.Equal(msg1.Length, len3 + len4);
                     Assert.Equal(msg1, received3);
 
-                    string userName = server.GetImpersonationUserName();    // not sure what to test here that will work in all cases
```
ianhays: I believe this is the only place where we test a successful call to GetImpersonationUserName, so it's removal would result in a code coverage decrease. The other tests you added for the exception throwing on GetImpersonationUserName probably balanced out the decrease on the successful case.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
```diff
@@ -346,6 +346,14 @@ public void WaitForConnectionOnAlreadyConnectedServer_Throws_InvalidOperationExc
                 }
 
                 Assert.Throws<InvalidOperationException>(() => client.IsMessageComplete);
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    Assert.Throws<IOException>(() => client.NumberOfServerInstances);
+                }
+                else
+                {
+                    Assert.Throws<PlatformNotSupportedException>(() => client.NumberOfServerInstances);
```
ianhays: This line (355 in NamedPipeTest.Simple.cs) is throwing an "IOException: Pipe is broken" on Linux for:

- Server: InOut, Client: Out
- Server: In, Client: Out

ianhays:
Some really great additions here! Thanks @Maxwe11

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
Maxwe11: Good catch, will move it into separate test and compare with value retrieved from interop call.

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
Maxwe11: thanks

src/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Simple.cs
Maxwe11: @ianhays we already have test case for `InvalidOperationException` (see line 458 at this file) so could you clarify what should I do?

ianhays: My apologies, I forgot about that one. Your changes are good.

Maxwe11:
Updated

stephentoub:
LGTM

PR #2842, opened 2015-08-17
[WIP] Reduce range checks.

hackcraft:
When a signed value LIMIT is known to be greater than zero, then:

    bool exclusive = value >= 0 && value < LIMIT;
    bool inclusive = value >= 0 && value <= LIMIT;

Is equivalent to:

    bool exclusive = (uint)value < (uint)LIMIT;
    bool inclusive = (uint)value <= (uint)LIMIT;

And so on.

While conceptually in C# replacing the former with the latter would mean
replacing two comparisons with two casts and a comparison, in terms of
the CIL produced there is no cast; rather blt.s is replaced with blt.un.s
clt replaced with clt.un etc.

Changing the former for the latter can therefore not just remove a branch
(in cases where it isn't removed on being jitted) but also increase the odds
of a method being considered "small" by the jitter when it optimises.

Experiment with making such changes.

### DIFF
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Win32.SafeHandles
             get
             {
                 long h = (long)handle;
-                return h < 0 || h > int.MaxValue;
+                return (ulong)h > (ulong)int.MaxValue;
             }
         }
     }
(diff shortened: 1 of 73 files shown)


stephentoub:
As I noted on the associated issue, I personally find this makes the code harder to understand and more error-prone, and I've not seen any perf data that demonstrates it's measurable/important to do across the board like this.  If there are specific instances where it does make an impactful difference, I'm fine with those changing on a case-by-case basis; better yet would be, if there are enough such cases, for the backend to learn how to optimize those better without changing the code.

PR #2845, opened 2015-08-17
Make results implement IList<T> when possible.

### LINKED ISSUE #2844: Implement IList<T> in Linq responses when possible
There is code throughout .NET that accepts arguments of `IEnumerable<T>` and has an optimised path for `IList<T>` and/or `ICollection<T>`. Some methods of `System.Linq.Enumerable` would be examples, but also elsewhere such as some collections' constructors. The effect upon `.ToList()`, `.ToArray()` and any operation in `Enumerable` that requires buffering are perhaps particularly of note here.

Some methods of `Enumerable` have results which are conceptually lists. E.g. `Repeat` returns a sequence of a specified size where all items are the same, which can be modelled as a read-only list.

Some methods of `Enumerable` have results which can be conceptually lists in certain conditions. E.g. If `Select` is called on a source that is itself a list, then the result would be a read-only list of the same size as the source list, and a different type.

Returning such list types when possible would increase the times that the existing fast-paths for list types are made use of. At the same time, some such changes would add further such cases, potentially allowing chains of operations that cannot currently hit such fast-paths to do so.

hackcraft:
Fixes #2844

Range, Repeat always
Concat, Reverse, Select, Zip when source allows.

Also, make DefaultIfEmpty make use of type of source to shortcut.

Some WhereSelectXXXIterator types now never have a null predicate.
Take advantage of this by removing dead branches. Assert this too.

WhereSelectEnumerableIterator can still have a null predicate, but
this need only be checked for once per enumeration.


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -42,8 +42,15 @@ private static IEnumerable<TSource> WhereIterator<TSource>(IEnumerable<TSource>
             if (source == null) throw Error.ArgumentNull("source");
             if (selector == null) throw Error.ArgumentNull("selector");
             if (source is Iterator<TSource>) return ((Iterator<TSource>)source).Select(selector);
-            if (source is TSource[]) return new WhereSelectArrayIterator<TSource, TResult>((TSource[])source, null, selector);
-            if (source is List<TSource>) return new WhereSelectListIterator<TSource, TResult>((List<TSource>)source, null, selector);
+            IList<TSource> ilist = source as IList<TSource>;
+            if (ilist != null)
```
hackcraft: This reduces the cost of testing for `TSource[]` or `List<TSource>` in the case where the source is not such a type, by one check, increases by one check when they are of such a type, but also adds support for other `IList<TSource>` types.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -364,13 +389,26 @@ public override bool MoveNext()
                 {
                     case 1:
                         _enumerator = _source.GetEnumerator();
-                        state = 2;
-                        goto case 2;
+                        if (_predicate == null)
```
hackcraft: This is never going to change for the object's lifetime, so there's no need to check it more than once per enumeration.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -528,6 +572,668 @@ public override TResult[] ToArray()
...
+                throw Error.NotSupported();
+            }
+    
+            public bool IsReadOnly
+            {
+                get { return true; }
+            }
+        }
+
+        internal sealed class SelectListIterator<TSource, TResult> : Iterator<TResult>, IList<TResult>
```
hackcraft: This is an optimised class that could also be served by `SelectIListIterator<TSource, TResult>`. My instincts would have been not to include it, but as it replaces `WhereSelectListIterator<TSource, TResult>` it seemed perhaps best not to remove the degree to which that class optimises by using `List<TSource>.Enumerator` directly.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -889,6 +1595,13 @@ public static IEnumerable<TSource> Concat<TSource>(this IEnumerable<TSource> fir
         {
             if (first == null) throw Error.ArgumentNull("first");
             if (second == null) throw Error.ArgumentNull("second");
+            IList<TSource> firstList = first as IList<TSource>;
+            if (firstList != null)
+            {
+                IList<TSource> secondList = second as IList<TSource>;
+                if (secondList != null)
+                    return new ListConcat<TSource>(firstList, secondList);
+            }
```
hackcraft: Slight constant cost to `Concat` of non-list types here.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -897,12 +1610,115 @@ private static IEnumerable<TSource> ConcatIterator<TSource>(IEnumerable<TSource>
...
         {
             if (first == null) throw Error.ArgumentNull("first");
             if (second == null) throw Error.ArgumentNull("second");
             if (resultSelector == null) throw Error.ArgumentNull("resultSelector");
+            IList<TFirst> firstList = first as IList<TFirst>;
+            if (firstList != null)
+            {
+                IList<TSecond> secondList = second as IList<TSecond>;
+                if (secondList != null) return new ZipListIterator<TFirst, TSecond, TResult>(firstList, secondList, resultSelector);
+            }
```
hackcraft: Again, slight constant cost in the case where this change doesn't help.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1012,6 +1942,105 @@ private static IEnumerable<TSource> ReverseIterator<TSource>(IEnumerable<TSource
             Buffer<TSource> buffer = new Buffer<TSource>(source);
             for (int i = buffer.count - 1; i >= 0; i--) yield return buffer.items[i];
         }
+        
+        private class ReverseList<TSource> : ReadOnlyList<TSource>, IListIterator<TSource>
```
hackcraft: This does change the buffering behaviour. Breaking?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1465,26 +2499,173 @@ public static TSource ElementAtOrDefault<TSource>(this IEnumerable<TSource> sour
...
         public static IEnumerable<int> Range(int start, int count)
         {
-            long max = ((long)start) + count - 1;
-            if (count < 0 || max > Int32.MaxValue) throw Error.ArgumentOutOfRange("count");
-            return RangeIterator(start, count);
+            return new RangeIterator(start, count);
         }
 
-        private static IEnumerable<int> RangeIterator(int start, int count)
+        private sealed class RangeIterator : ReadOnlyList<int>, IListIterator<int>
```
hackcraft: Range and Repeat are perhaps the cleanest here, with no costs in any case over the previous.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -528,6 +572,668 @@ public override TResult[] ToArray()
...
+                {
+                    current = _selector(_source[_index++]);
+                    return true;
+                }
+                Dispose();
+                return false;
+            }
+
+            // Once we have generic virtual support back, rename this back to Select and turn it into an override of the parent's abstract Select() method.
+            public IEnumerable<TResult2> SelectImpl<TResult2>(Func<TResult, TResult2> selector)
```
hackcraft: Currently, this won't be hit at all, but it's worth thinking of the future.

stephentoub:
cc: @VSadov

hackcraft:
I've now hidden the `IList<T>` interfaces. This has the disadvantage of blocking the interface from a consumer that is using it carefully, and the advantage of blocking the interface from a consumer that is using it carelessly. This hence addresses the concern at https://github.com/dotnet/apireviews/blob/master/2015-04-28-misc/README.md#1182-systemlinq-performance-improvement-suggestions

If it's later decided that it's not a strong enough concern to prevent the interface being discoverable, it will be relatively easy to expose them again.

The concern about `Count` is harder, though perhaps less significant; the use of `Count` + `CopyTo` behaves as-if atomically if the calling thread is the only one that accesses the source lists, while other operations can currently fail on other lists (there are after all methods with the same issue already). This increases the frequency, but not the gist of that issue. If it's not a ByDesign and not a WontFix then any fix will also affect the code changed here anyway. (And if a future fix can propogate a guarantee to maintain count, this code could be a starting point for it).

The call cost is now greater though, which complicates the cost/benefit.

hackcraft:
The cost of hiding the `IList` is definitely too high for this to be useful as-is, but there's potential for improvement. Closing for now though.

PR #2846, opened 2015-08-17
Make registry tests more robust

### LINKED ISSUE #2819: Microsoft.Win32.Registry.Tests fail if machine is dirty
I hit this failure after breaking into the tests. 
```
Starting:    Microsoft.Win32.Registry.Tests
     Microsoft.Win32.RegistryTests.RegistryKey_CreateSubKey_str.NegativeTests [  FAIL]
        Assert.Null() Failure
        Expected: (null)
        Actual:   HKEY_CURRENT_USER\REG_TEST_1
        Stack Trace:
           E:\dotnet-bot\corefx\src\Microsoft.Win32.Registry\tests\TestSubKey.cs(19,0): at Microsoft.Win32.RegistryTests.TestSubKey..ctor(String testKeyName)
           E:\dotnet-bot\corefx\src\Microsoft.Win32.Registry\tests\RegistryKey\RegistryKey_CreateSubKey_str.cs(16,0): at Microsoft.Win32.RegistryTests.RegistryKey_CreateSubKey_str..ctor()
```
The only way to workaround was to manually clean up the key the tests left behind.
The tests should ensure a clean state before running.

stephentoub:
Currently if the registry tests are interrupted during their execution, they leave behind state in the registry that causes all subsequent runs of the tests to fail. This commit cleans up the registry tests to avoid this.  The logic for creating a name for the test class is moved into the base, which also deletes any existing key of that name prior to running a test.

Fixes #2819
cc: @ericstj, @pallavit


src/Microsoft.Win32.Registry/tests/RegistryTestsBase.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private string CreateUniqueKeyName()
+        {
+            // Create a name to use for this class of tests. The name includes:
+            // - A "corefxtest" prefix to help make it clear to anyone looking at the registry
+            //   that these keys are test-only and can be deleted, in case the tests crash and 
+            //   we end up leaving some keys behind.
+            // - The name of this test class, so as to avoid problems with tests on different test 
+            //   classes running concurrently
```
Maxwe11: Currently parallelism is off but I think it could be enabled.

stephentoub: Yup.  Even if it's left off, though, I think there's value in including the class name in the test key.

src/Microsoft.Win32.Registry/tests/RegistryTestsBase.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using Xunit;
+
+namespace Microsoft.Win32.RegistryTests
+{
+    public abstract class RegistryTestsBase : IDisposable
+    {
+        protected readonly string _testRegistryKeyName;
+        protected readonly RegistryKey _testRegistryKey;
+
```
nguerrera: Doesn't really matter for tests, but I'd expect PascalCased properties not underscore fields for protected members.

nguerrera:
LGTM

stephentoub:
Thanks, Nick.

PR #2847, opened 2015-08-17
Add a gitattribute rule for .sh files

gregg-miskelly:
This forces bash scripts to always use lf line endings so that if a
repro is accessed in Unix via a file share from Windows, the scripts
will work.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,3 +22,9 @@ indent_size = 2
 # Xml config files
 [*.{props,targets,config,nuspec}]
 indent_size = 2
+
+# Shell scripts
+[*.sh]
+end_of_line = lf
+[*.{cmd, bat}]
+end_of_line = crlf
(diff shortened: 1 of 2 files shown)


jasonwilliams200OK:
:+1: 
Please also add rule in https://github.com/dotnet/corefx/blob/master/.editorconfig:

```yml
[*.sh]
end_of_line: lf
```

Manual: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#end_of_line

gregg-miskelly:
Added the .editorconfig change also.

jasonwilliams200OK:
Thank you @gregg-miskelly. At some later point, we may even have cmd/bat rule (`end_of_line = crlf`) under `[*]`, but currently I am not sure if all the rest of file types mandate "editable with notepad", as the rest of the Windows editors are lf-tolerant. :)

stephentoub:
cc: @ellismg

ellismg:
LGTM.  Thanks for doing this.

Issue #2820, opened 2015-08-14
Use getpwuid_r to read the user's home directory, not the HOME environment variable


bartonjs:
https://github.com/dotnet/corefx/pull/2806 is adding something which needs to determine where the user's home directory is.  There are system contexts where the HOME environment variable isn't set, but there's still a user, and those situations should work.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/pwd.h.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/getpwuid_r.html

https://github.com/dotnet/corefx/pull/2806/files#diff-97146b006d1e7892c2cc8390444738bdR74 for an approximate target point for the consumer.

bartonjs:
https://github.com/dotnet/corefx/blob/cda0ae17da76af6bd726c5cbd9c3c55fa4b881de/src/System.Console/src/System/ConsolePal.Unix.cs#L493 is also calling GetEnvironmentVariable("HOME"), and might want to switch to using this same thing.

PR #2850, opened 2015-08-17
Consolidate temp files in Unix implementations

stephentoub:
System.IO.Pipes and System.IO.MemoryMappedFiles both create temporary files, and end up doing so in inconsistent locations.  This commit changes them to use a consistently named location under the temporary path, e.g. /tmp/.dotnet/corefx/pipes/* and /tmp/.dotnet/corefx/maps/*

cc: @bartonjs, @ianhays

### DIFF
--- a/src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
+++ b/src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
@@ -9,8 +9,9 @@ namespace System.IO.MemoryMappedFiles
     {
         private static FileStream CreateSharedBackingObject(Interop.libc.MemoryMappedProtections protections, long capacity)
         {
-            string path = TmpPathPrefix + Guid.NewGuid().ToString("N") + ".tmp";
-
+            Directory.CreateDirectory(s_tempMapsDirectory);
+            string path = Path.Combine(s_tempMapsDirectory, Guid.NewGuid().ToString("N"));
+            
             FileAccess access =
                 (protections & (Interop.libc.MemoryMappedProtections.PROT_READ | Interop.libc.MemoryMappedProtections.PROT_WRITE)) != 0 ? FileAccess.ReadWrite :
                 (protections & (Interop.libc.MemoryMappedProtections.PROT_WRITE)) != 0 ? FileAccess.Write :
@@ -37,6 +38,10 @@ namespace System.IO.MemoryMappedFiles
         // ---- PAL layer ends here ----
         // -----------------------------
 
-        private static readonly string TmpPathPrefix = Path.Combine(Path.GetTempPath(), MemoryMapObjectFilePrefix);
+        private static readonly string s_tempMapsDirectory = Path.Combine(
+            Path.GetTempPath(),
+            PersistedFiles.TopLevelHiddenDirectory,
+            PersistedFiles.SecondLevelDirectory,
+            "maps");
     }
 }
--- a/src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
+++ b/src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
@@ -38,10 +38,6 @@ namespace System.IO.MemoryMappedFiles
         // ---- PAL layer ends here ----
         // -----------------------------
 
-        private static readonly string s_tempMapsDirectory = Path.Combine(
-            Path.GetTempPath(),
-            PersistedFiles.TopLevelHiddenDirectory,
-            PersistedFiles.SecondLevelDirectory,
-            "maps");
+        private static readonly string s_tempMapsDirectory = PersistedFiles.GetTempFeatureDirectory("maps");
     }
 }
--- a/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
@@ -37,7 +37,7 @@ namespace System.IO.Pipes
             }
 
             // Return the pipe path
-            return Path.Combine(GetPipeDirectoryPath(), pipeName);
+            return Path.Combine(EnsurePipeDirectoryPath(), pipeName);
         }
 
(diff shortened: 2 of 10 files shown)


src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
```diff
@@ -37,6 +38,10 @@ private static FileStream CreateSharedBackingObject(Interop.libc.MemoryMappedPro
         // ---- PAL layer ends here ----
         // -----------------------------
 
-        private static readonly string TmpPathPrefix = Path.Combine(Path.GetTempPath(), MemoryMapObjectFilePrefix);
+        private static readonly string s_tempMapsDirectory = Path.Combine(
+            Path.GetTempPath(),
```
bartonjs: Why not create a method for this in `PersistedFiles`? (`GetHiddenTempFeaturePath`?)

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
```diff
@@ -255,6 +226,59 @@ public virtual PipeTransmissionMode ReadMode
...
+                // level individually (though we assume that the temp path already exists).
+
+                path = Path.GetTempPath();
+
+                path = Path.Combine(path, PersistedFiles.TopLevelHiddenDirectory);
+                CreateDirectory(path);
+
+                path = Path.Combine(path, PersistedFiles.SecondLevelDirectory);
+                CreateDirectory(path);
+
```
bartonjs: CreateDirectory should create all intermediate directories it needs, and Path.Combine doesn't care about existence...

But also a "why isn't this using a method off of PersistedFiles"?  (I like the idea of the directory names staying private, vs internal; maybe it's just me)

bartonjs: Oh... are you too low level here to use Directory.CreateDirectory?  That makes some reluctant sense.

src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
stephentoub: Because System.IO.Pipes doesn't/can't depend on System.IO.FileSystem and thus can't use Directory.CreateDirectory, which supports creating all of the non-existing directories in the path.  That means for the would-be primary consumer of the API (currently), we'd be building up the path, only to then force System.IO.Pipes to need to parse that string in order to create each individual directory in the path.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
stephentoub: Yup :(

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
```diff
@@ -255,6 +226,59 @@ public virtual PipeTransmissionMode ReadMode
         // ---- PAL layer ends here ----
         // -----------------------------
 
+        private static string s_pipeDirectoryPath;
+
+        private static string GetPipeDirectoryPath()
+        {
+            string path = s_pipeDirectoryPath;
+            if (path == null)
```
ianhays: So in some cases we will have path == null but the directory will actually exist because it was created (but not cleaned up) by an earlier running instance. In this case, we will go through three calls to CreateDirectory that that return Interop.Error.EEXIST to ensure the directory exists.

Is it worthwhile to instead start with a check to see if the directory already exists a la Directory.Exists (using GetFileInformationFromPath for Unix)?

stephentoub: Sure, I'll add the check.

src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
bartonjs: Right now PersistedFiles only uses Path.Combine; though I do understand the argument about Pipes not being able to use Directory.CreateDirectory (therefore not wanting to go reparse the path to do the equivalent to `mkdir -p`).

This call could use the helper method, as witnessed by the single Path.Combine here.
Pipes could accomplish it by either exposing something like `CreateTempFeatureDirectoryRecursive(Action<string> mkdir)`; or exposing `GetTempTopLevelPath()` and `GetTempProductPath()`.  Or MemoryMappedFiles using the helper, and Pipes having a comment that it's really `PersistedFiles.GetTempFeaturePath`, but written longhand because it can't use Directory.CreateDirectory.

Certainly not a "must" comment; just trying to avoid a followup change where we change this to use a helper if we see it coming in the short-to-mid-term.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
stephentoub: Done.  I also changed it to recreate the directory each time if it didn't exist, rather than assuming that just because it existed when the static was set that it still exists.

src/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedFile.Unix.BackingObject_File.cs
stephentoub: Ok.  I added the GetTempFeaturePath function and used it from MemoryMappedFiles.  I still don't use it from Pipes.

bartonjs:
LGTM.  If you wanted to avoid pulling in the "no home directory" string by breaking it up into PersistedFiles.User.Unix.cs and PersistedFiles.Temp.Unix.cs that might be reasonable.  Or, since that exception could end up going away with https://github.com/dotnet/corefx/issues/2820, maybe it is short term pain that's endurable.

stephentoub:
> maybe it is short term pain that's endurable

Yeah, it's a bit annoying, but I'm not going to worry about it for now.  With how we do the IO error handling on both Windows and Unix with the helper function that spits out the right exception message for a variety of error conditions, we already pull in more error strings into some assemblies than we actually need... this is just one more case of that.

Thanks.

PR #2853, opened 2015-08-17
Revert range optimization

### LINKED ISSUE #2852: Enumerable.Range bug for boundary values
I believe that this commit (https://github.com/dotnet/corefx/commit/31599c59cb601459982be0aa78f7f8381b53566c) introduced a bug in boundary conditions for Enumerable.Range

The below program exhibits this:

````CSharp
{
    int from = int.MaxValue - 9;
    int count = 10;
    Console.WriteLine("Enumerable.Range:");
    foreach (int i in Enumerable.Range(from, count))
    {
        Console.WriteLine(i);
    }
}
````

On the Desktop library, this prints: 

````
Enumerable.Range:
2147483638
2147483639
2147483640
2147483641
2147483642
2147483643
2147483644
2147483645
2147483646
2147483647
````

With the latest version of System.Linq here, it prints no elements:

````
Enumerable.Range:
````

Here is the new iteration logic for Enumerable.Range:

````CSharp
private static IEnumerable<int> RangeIterator(int start, int count)
{
    for (int i = 0; i < count; i++) yield return start + i;
    for (int end = start + count; start < end; start++)
        yield return start;
}
````

The loop condition will fail for any range that will reach Int32.MaxValue.

NOTE: This doesn't seem to be covered by any of our Linq tests. I only discovered it because one of our internal Linq.Parallel tests failed, which was trying to compare ParallelEnumerable.Range to Enumerable.Range, and I noticed that Enumerable.Range was returning nothing for this case. We should improve coverage here on top of fixing this bug.

### PR DESCRIPTION
**Pull request by mellinoe**
The new implementation does not handle the edge case where Enumerable.Range enumerates up to Int32.MaxValue. If values are given such that the end value will be Int32.MaxValue, no values are enumerated. This simply reverts that change.

I've added a simple case that will catch this regression in the future.

Fixes #2852 .

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1472,8 +1472,7 @@ namespace System.Linq
 
         private static IEnumerable<int> RangeIterator(int start, int count)
         {
-            for (int end = start + count; start < end; start++)
-                yield return start;
+            for (int i = 0; i < count; i++) yield return start + i;
         }
 
         public static IEnumerable<TResult> Repeat<TResult>(TResult element, int count)
--- a/src/System.Linq/tests/RangeTests.cs
+++ b/src/System.Linq/tests/RangeTests.cs
@@ -92,5 +92,18 @@ namespace System.Linq.Tests
                 Assert.NotSame(enum1, enum2);
             }
         }
+
+        [Fact]
+        public void Range_ToInt32MaxValue()
+        {
+            int from = Int32.MaxValue - 3;
+            int count = 4;
+            var rangeEnumerable = Enumerable.Range(from, count);
+
+            Assert.Equal(count, rangeEnumerable.Count());
+
+            int[] expected = { Int32.MaxValue - 3, Int32.MaxValue - 2, Int32.MaxValue - 1, Int32.MaxValue };
+            Assert.Equal(expected, rangeEnumerable);
+        }
     }
 }

### CODE REVIEW

**Comment by stephentoub** [comment]
LGTM.  Thanks for fixing it, Eric.

**Comment by VSadov** [comment]
The unported Range test has scenario for this break:
https://github.com/VSadov/corefx/blob/UnportedLegacyTests/src/System.Linq/LegacyTestsToPort/LINQ/Range/Range2/Range2.cs

We should gate changes on legacy tests being ported in given area.

**Comment by VSadov** [comment]
LGTM, thanks, Eric

File: src/System.Linq/src/System/Linq/Enumerable.cs [hunk 1]
```diff
@@ -1472,8 +1472,7 @@ public static IEnumerable<int> Range(int start, int count)
 
         private static IEnumerable<int> RangeIterator(int start, int count)
         {
-            for (int end = start + count; start < end; start++)
```
hackcraft: Changing this to start != end should also work, as the addition is unchecked.

James-Ko:
Thanks for fixing it, sorry about that.

akoeplinger:
@VSadov now that the ported tests are in, I decided to run with the misbehaving change applied to see if the Range2 test breaks. It does not, however, and I think it didn't even before the port (i.e. it only verifies that an ArgumentOutOfRangeException is thrown, not that the collection has the expected elements). Can you verify this should actually break that test?

hackcraft:
@akoeplinger none of the tests in the legacy tests @VSadov brought over and you integrated would have caught this, the condition simply isn't looked for.

PR #2854, opened 2015-08-17
Bring back range optimisation, fixed for boundary case.

hackcraft:
31599c5 introduced a boundary error, and was hence reverted in b75ffa4.

This reintroduces the idea of 31599c5, but fixed for the boundary case.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1472,7 +1472,7 @@ namespace System.Linq
 
         private static IEnumerable<int> RangeIterator(int start, int count)
         {
-            for (int i = 0; i < count; i++) yield return start + i;
+            for (int end = start + count; start != end; start++) yield return start;
         }
 
         public static IEnumerable<TResult> Repeat<TResult>(TResult element, int count)


stephentoub:
cc: @VSadov, @mellinoe

VSadov:
I think, it would be right thing to do to port legacy Range tests with this change. 
There are not that many of them, but they do cover the overflow issue and will work as a regression scenarios as well.

hackcraft:
Good idea.

VSadov:
@hackcraft - you may wait for https://github.com/dotnet/corefx/pull/2874 to go in.

VSadov:
retest this please

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1472,7 +1472,7 @@ public static IEnumerable<int> Range(int start, int count)
 
         private static IEnumerable<int> RangeIterator(int start, int count)
         {
-            for (int i = 0; i < count; i++) yield return start + i;
+            for (int end = start + count; start != end; start++) yield return start;
```
weshaggard: Is it ever possible for count to come in here as negative? If so this loop will do bad things. Why do we think this is more so much more efficient?

stephentoub: > Is it ever possible for count to come in here as negative?

No, it's checked a few lines above:
https://github.com/dotnet/corefx/pull/2854/files#diff-95a14d3b4ae14339b623dd95c7537fd0R1469

hackcraft: A negative count getting to here would be a bug in either version, as the calling method is specified as throwing in that case.

PR #2855, opened 2015-08-17
Support writing colored messages to console in a multi-threaded environment

### LINKED ISSUE #2808: Synchronized colors in the console
TLDR: I believe that there is currently no way to correctly write colored messages to the console in a multi-threaded environment. There should be one.

Long version:

When you want to write a colored message to the console, the common pattern looks something like:

``` c#
Console.ForegroundColor = ConsoleColor.Red;
Console.WriteLine("ERROR!");
Console.ResetColor();
```

This approach is not thread-safe, since other writers can also write while the color is set.

An improvement is [to lock on a private object](http://stackoverflow.com/a/13861466/41071):

``` c#
static object lockObj = new object();

...

lock (lockObj)
{
    Console.ForegroundColor = ConsoleColor.Red;
    Console.WriteLine("ERROR!");
    Console.ResetColor();
}
```

This will synchronize correctly among writers that use the same lock object, but other writers can still end up with incorrect color.

On the Desktop .Net, writing to the console uses [`SyncTextWriter`](http://referencesource.microsoft.com/#mscorlib/system/io/textwriter.cs,9e3dd0323cf4ee8b) as `Console.Out` which locks on itself. This means that locking on `Console.Out` will synchronize correctly with other non-colored writers (those that just call `Console.WriteLine` without any additional synchronization) and colored writers which also lock on `Console.Out`:

``` c#
lock (Console.Out)
{
    Console.ForegroundColor = ConsoleColor.Red;
    Console.WriteLine("ERROR!");
    Console.ResetColor();
}
```

On CoreFX, [a different version of `SyncTextWriter`](https://github.com/dotnet/corefx/blob/fa6e6a1e6300cdbffa31998a8c70d3bb9b7d06c3/src/System.Console/src/System/IO/SyncTextWriter.cs) is used, which locks on its own private lock object. This is generally considered a better practice than locking on `this`, but it also means there is no way to write a colored message to the console that is correctly synchronized with non-colored writers.

It would be nice if there was an officially supported pattern to write a colored message to the console with correct synchronization on both Desktop and Core. Failing that, I think that `SyncTextWriter` on Core should be implemented the same way as the Desktop version, so that the `lock (Console.Out)` version also worked on Core, event if it's relying on an implementation detail.

pallavit:
Fix #2808  

In the System.Console portable library SyncTextReader/SyncTextWriter (the reader and writer used by the Console) locked on a private lock object which prevented users from writing colored console messages in sync with non-colored writers. Note this behavior was a regression from the Desktop
where this could be achieved by locking against Console.Out which resulted in locking the writer APIs since the SyncTextWriter internally locked on itself.

Although this approach comes from an implementation detail, the scenario is useful and there exists no work-around for it in the portable library.So, moving back to the earlier implementation, where the readers and writers locked against themselves instead of a private object.

### DIFF
--- a/src/System.Console/src/System/IO/SyncTextReader.cs
+++ b/src/System.Console/src/System/IO/SyncTextReader.cs
@@ -8,10 +8,12 @@ using System.Threading.Tasks;
 
 namespace System.IO
 {
+    /* SyncTextReader intentionally locks on itself rather than a private lock object.
+     * This is done to synchronize different console readers(Issue#2855).
+     */
     internal sealed class SyncTextReader : TextReader
     {
         private readonly TextReader _in;
-        private readonly object _methodLock = new object();
 
         public static SyncTextReader GetSynchronizedTextReader(TextReader reader)
         {
@@ -29,7 +31,7 @@ namespace System.IO
         {
             if (disposing)
             {
-                lock (_methodLock)
+                lock (this)
                 {
                     _in.Dispose();
                 }
@@ -38,7 +40,7 @@ namespace System.IO
 
         public override int Peek()
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.Peek();
             }
@@ -46,7 +48,7 @@ namespace System.IO
 
         public override int Read()
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.Read();
             }
@@ -54,7 +56,7 @@ namespace System.IO
 
         public override int Read([In, Out] char[] buffer, int index, int count)
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.Read(buffer, index, count);
             }
@@ -62,7 +64,7 @@ namespace System.IO
 
         public override int ReadBlock([In, Out] char[] buffer, int index, int count)
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.ReadBlock(buffer, index, count);
             }
@@ -70,7 +72,7 @@ namespace System.IO
 
         public override String ReadLine()
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.ReadLine();
             }
@@ -78,7 +80,7 @@ namespace System.IO
 
         public override String ReadToEnd()
         {
-            lock (_methodLock)
+            lock (this)
             {
                 return _in.ReadToEnd();
             }
--- a/src/System.Console/src/System/IO/SyncTextWriter.cs
+++ b/src/System.Console/src/System/IO/SyncTextWriter.cs
@@ -7,9 +7,13 @@ using System.Threading.Tasks;
 
 namespace System.IO
 {
+    /* SyncTextWriter intentionally locks on itself rather than a private lock object.
+     * This is done to synchronize different console writers.
+     * For example - colored console writers can be synchronized with non-colored
+     * writers by locking on Console.On (Issue#2855).
+     */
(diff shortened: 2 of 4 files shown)


src/System.Console/src/System/IO/SyncTextReader.cs
```diff
@@ -11,7 +11,6 @@ namespace System.IO
     internal sealed class SyncTextReader : TextReader
     {
         private readonly TextReader _in;
-        private readonly object _methodLock = new object();
```
Maxwe11: :bulb:  I'd rather assign `_methodLock = this;` with comment explaining why we lock on this.

stephentoub: Personally I don't think it's worth an extra field for that.  If you did want something like that, you could do:
```C#
private object MethodLock { get { return this; } } // appropriate comment
```
and then use MethodLock everything instead of _methodLock.  That said, I'm also fine with it as it is.

pallavit: @Maxwe11 Thanks for the feedback. I agree with @stephentoub and would prefer to use lock(this). It gives an additional benefit of being consistent with the desktop code as both code basis now have same implementation.

akoeplinger: Can you add a comment to at least one of the `lock(this)` that it is intentional so nobody tries the same change in the future? Or is there a way to assert this in a unit test?

stephentoub: We should definitely add a comment to the implementation.

We should also add a test, e.g.
```C#
[Fact]
public void SyncTextWriterLockedOnThis()
{
    TextWriter oldWriter = Console.Out;
    try
    {
        var newWriter = new CallbackTextWriter();
        Console.SetOut(newWriter);
        TextWriter syncWriter = Console.Out;

        bool called = false;
        newWriter.Callback = _ =>
        {
            Assert.True(Monitor.IsEntered(syncWriter));
            called = true;
        };
        Console.Write("c");
        Assert.True(called);
    }
    finally
    {
        Console.SetOut(oldWriter);
    }
}

private sealed class CallbackTextWriter : TextWriter
{
    internal Action<char> Callback;

    public override Encoding Encoding { get { return Encoding.UTF8; } }

    public override void Write(char value) { Callback(value); }
}
```

stephentoub:
With a comment and a test as @akoeplinger suggests, LGTM.

pallavit:
Thanks for the review. I have incorporated the feedback. Let me know if you have any further comments, else I would like to squash the commits and merge.

src/System.Console/src/System/IO/SyncTextReader.cs
```diff
@@ -8,10 +8,13 @@
 
 namespace System.IO
 {
+    /* SyncTextReader intentionally locks on itself rather than a private lock object.
+     * This is done to ensure colored console readers can be synchronized with non-colored
+     * readers (Issue#2855).
+     */
```
stephentoub: Does the reader change really have to do with color?

stephentoub:
One nit, but otherwise LGTM.

src/System.Console/src/System/IO/SyncTextReader.cs
pallavit: Umm...you are right. In fact even for the writer's it can potentially impact other states (for example, encoding, although it is not exposed right now). I will remove the color from the comments.

PR #2856, opened 2015-08-17
 Move EnsureOpenSslInitialized shim from coreclr to corefx  

nguerrera:
Moved EnsureOpenSslInitialized from [coreclr/src/pal/src/misc/corefx.cpp](https://github.com/dotnet/coreclr/blob/dc0fca478639add871244ad106500ae0f5ff676c/src/pal/src/misc/corefx.cpp#L46)

Before doing so, I went ahead and renamed the crypto shim API along the lines we discussed in #2606 and going with Jeremy's preferred shorter name for Interop.Libraries.*. That's in the separate commit abaeaa8.

There are still some style differences between S.S.C.Native and S.Native that I'd like to address, but for now stopping at getting the rename in and EnsureOpenSslInitialized over the fence.

cc @bartonjs @stephentoub


bartonjs:
LGTM

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Initialization.cs
```diff
@@ -33,6 +56,16 @@ static libcrypto()
             ERR_load_crypto_strings();
         }
 
+        internal static void Initialize()
+        {
```
stephentoub: I'd suggest adding a short comment here that this is a temporary function, meant to be a nop just to give multiple other static cctors a hook to ensure CryptoInitializer's static cctor is invoked.

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -911,3 +915,114 @@ UpRefEvpPkey(
...
+
+/*
+Function:
+_LockingCallback
+
+Called back by OpenSSL to lock or unlock.
+*/
+static
+void
+_LockingCallback(int mode, int n, const char* file, int line)
```
stephentoub: Just curious: why the underscore prefix?  Is this a style we're adopting for static functions?

stephentoub:
I think it'd be worth mentioning in the commit message that, as part of the port, the original implementation was simplified due to not needing dynamic loading (we couldn't take a static dependency on OpenSSL in libcoreclr, but we can from this shim).

Other than that and the two minor comments, LGTM.

src/Native/System.Security.Cryptography.Native/openssl.c
nguerrera: I was following Jeremy's lead with _MakeTimeT in same file, but I think we can just mark the functions static, _MakeTimeT is not currently. I'll go in the other direction (to static without _) as it's what I have everywhere else in the shims.

nguerrera:
@stephentoub Updated the commit message and I think I addressed everything else. Mind taking another look?

stephentoub:
Thanks, Nick. LGTM.

PR #2859, opened 2015-08-18
Move ForkAndExecProcess shim from coreclr to corefx

nguerrera:
Moved from [coreclr/src/pal/src/misc/corefx.cpp](https://github.com/dotnet/coreclr/blob/b4333eff478fd48292d350f675be2ba916ea8271/src/pal/src/misc/corefx.cpp#L206)

cc @stephentoub

### DIFF
--- a/src/Native/System.Native/pal_exec.cpp
+++ b/src/Native/System.Native/pal_exec.cpp
@@ -1,0 +1,144 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "../config.h"
+#include "pal_exec.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+enum
+{
+    READ_END_OF_PIPE = 0,
+    WRITE_END_OF_PIPE = 1,
+};
+    
+static void CloseIfOpen(int fd)
+{
+    // Ignoring errors from close is a deliberate choice and we musn't
+    // let close during cleanup on a failure path disturb errno.
+    if (fd >= 0)
+    {
+        int priorErrno = errno;
+        close(fd);
+        errno = priorErrno;
+    }
+}
+
+extern "C"
+int32_t ForkAndExecProcess(
+    const char* filename,
+    char* const argv[],
+    char* const envp[],
+    const char* cwd,
+    int32_t redirectStdin,
+    int32_t redirectStdout,
+    int32_t redirectStderr,
+    int32_t* childPid,
+    int32_t* stdinFd,
+    int32_t* stdoutFd,
+    int32_t* stderrFd)
+{
+    int success = true;
+    int stdinFds[2] = { -1, -1 }, stdoutFds[2] = { -1, -1 }, stderrFds[2] = { -1, -1 };
+    int processId = -1;
+
+    // Validate arguments
+    if (nullptr == filename || nullptr == argv || nullptr == envp ||
+        nullptr == stdinFd || nullptr == stdoutFd || nullptr == stderrFd ||
+        nullptr == childPid)
+    {
+        assert(!"null argument.");
+        errno = EINVAL;
+        success = false;
+        goto done;
+    }
+
+    if ((redirectStdin  & ~1) != 0 ||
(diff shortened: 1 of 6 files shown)


src/Native/System.Native/pal_exec.cpp
```diff
@@ -0,0 +1,142 @@
...
+    if ((processId = fork()) == -1)
+    {
+        assert(!"fork() failed.");
+        success = false;
+        goto done;
+    }
+
+    /* From the time the child process (processId == 0) begins running from fork to when
+     * it reaches execve, the child process must not touch anything in the PAL.  Doing so
+     * is not safe. The parent process (processId >= 0) may continue to use the PAL.
```
stephentoub: This PAL comment is no longer relevant, right?

src/Native/System.Native/pal_exec.cpp
```diff
@@ -0,0 +1,142 @@
...
+enum
+{
+    READ_END_OF_PIPE = 0,
+    WRITE_END_OF_PIPE = 1,
+};
+    
+inline static void CloseIfOpen(int fd)
+{
+    if (fd >= 0)
+        close(fd);
```
stephentoub: Maybe add a comment that we're explicitly choosing to ignore any close failures.

src/Native/System.Native/pal_exec.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+extern "C"
```
stephentoub: Lest anyone think it'd be a good idea to move this function back to managed code,I think a small comment here would be helpful, indicating why we can't do the fork/exec combination from managed.  Essentially, forking only clones the current thread, not all threads in the process, which makes it very unsafe given things like GC, finalization, etc.  We could, for example, end up in deadlock situations by trying to run managed code in the child process between the fork and the exec.

src/Native/System.Native/pal_exec.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /* From the time the child process (processId == 0) begins running from fork to when
+     * it reaches execve, the child process must not touch anything in the PAL.  Doing so
+     * is not safe. The parent process (processId >= 0) may continue to use the PAL.
+     */
+
+    if (processId == 0) // processId == 0 if this is child process
+    {
+        // Close the parent end of any open pipes
```
stephentoub: I realize I wrote this comment in the first place :), but we should probably clarify it slightly to say that we're closing the child's copy of the parent's end of any open pipes, not that we're actually closing the parent's end.

src/Native/System.Native/pal_exec.cpp
```diff
@@ -0,0 +1,142 @@
...
+        CloseIfOpen(stdinFds[WRITE_END_OF_PIPE]);
+        CloseIfOpen(stdoutFds[READ_END_OF_PIPE]);
+        CloseIfOpen(stderrFds[READ_END_OF_PIPE]);
+
+        *stdinFd  = -1;
+        *stdoutFd = -1;
+        *stderrFd = -1;
+        *childPid = -1;
+    }
+
```
stephentoub: The intention here was that if the fork failed, errno would still be set to whatever it was from the fork call.  But I just noticed we're making intervening calls to close, and it's possible close is changing errno.  Consider capturing errno after the fork and resetting it here.

src/Native/System.Native/pal_exec.cpp
stephentoub: (I realize I already had such a comment in the .h file :) Nevermind.)

stephentoub:
A few minor comments.  Otherwise, LGTM.

src/Native/System.Native/pal_exec.cpp
nguerrera: :) I've gone in the other direction and removed the pasted comments from the header for the parameters (did not mean to duplicate them) so the impl signature looks more bare and hints more at needing to look at the header for docs...

src/Native/System.Native/pal_exec.cpp
nguerrera: I opted to make CloseIfOpen preserve errno which fits well with earlier suggestion to document that ignoring close errors is deliberate.

stephentoub:
Updates LGTM.

PR #2861, opened 2015-08-18
Combine is-operator-then-cast sequences.

hackcraft:
Convert some cases of the form:

    if (value is T) ((T)value).DoSomething();

Etc. Into:

    T tValue = value as T;
    if (tValue != null) tValue.DoSomething();

Slightly smaller IL output, and removes double-cast in the case that uses it.


stephentoub:
LGTM, but @VSadov should take a look, in particular because I believe some of this code doesn't have tests that run in the repo.  Additionally, I'm not sure what should be done about the spacing (or lack thereof) in the new code; I find it a bit difficult to read as-is, but then I feel the same way about the lack of spacing in some of the other LINQ code of this ilk.

hackcraft:
I'm not fond of the conventions here either, but I've been trying to copy the style already used in the file affected. Not that you wouldn't perhaps find my normal style worse.

VSadov:
LGTM

PR #2862, opened 2015-08-18
Add HashSet<T> ctors with capacity

### LINKED ISSUE #382: Add ctor to HashSet<T> that allows the initial capacity to be specified

stephentoub:
Replaces PR #2122.  I squashed and fixed up @MarkusSintonen's commits from that PR.
Fixes #382.

New tests are not included due to System.Collections.dll being a partial facade, and the repo not currently having the infrastructure to enable adding tests for new surface area in partial facades.

cc: @terrajobst, @KrzysztofCwalina

### DIFF
--- a/src/System.Collections/src/System/Collections/Generic/HashSet.cs
+++ b/src/System.Collections/src/System/Collections/Generic/HashSet.cs
@@ -92,6 +92,10 @@ namespace System.Collections.Generic
             _version = 0;
         }
 
+        public HashSet(int capacity)
+            : this(capacity, EqualityComparer<T>.Default)
+        { }
+
         public HashSet(IEnumerable<T> collection)
             : this(collection, EqualityComparer<T>.Default)
         { }
@@ -128,6 +132,21 @@ namespace System.Collections.Generic
                 (_count > 0 && _slots.Length / _count > ShrinkThreshold))
             {
                 TrimExcess();
+            }
+        }
+
+        public HashSet(int capacity, IEqualityComparer<T> comparer)
+            : this(comparer)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException("capacity");
+            }
+            Contract.EndContractBlock();
+
+            if (capacity > 0)
+            {
+                Initialize(capacity);
             }
         }
 


src/System.Collections/src/System/Collections/Generic/HashSet.cs
```diff
@@ -131,6 +135,21 @@ public HashSet(IEnumerable<T> collection, IEqualityComparer<T> comparer)
...
         }
 
+        public HashSet(int capacity, IEqualityComparer<T> comparer)
+            : this(comparer)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException("capacity");
+            }
+            Contract.EndContractBlock();
```
nguerrera: I wonder (not tied to this commit) if we should strip Contract.EndContractBlock et al since we're not using the code contract tooling.

nguerrera:
LGTM

It might be worth merging master -> future and updating the contracts under the new ref/ now that they're part of the repo. It would also be nice to make sure there are (internal) tracking issues to get new surface area backported to desktop. It's not necessary to block on that for this commit, but perhaps we should open an issue to do those things for any changes already in future. cc @weshaggard

stephentoub:
CI failed due to future needing to be updated with some recent commits to master. I'll take care of that tomorrow if no one beats me to it.

src/System.Collections/src/System/Collections/Generic/HashSet.cs
weshaggard: It is something definitely worth considering, to this point I've not seen the value of ripping it all out yet. It may still be somewhat useful if we do get the C# vNext contracts feature in providing some hints about the contracts.

weshaggard:
@nguerrera @stephentoub it is my expectation that at some point when we decide to merge future back to master someone will need to pay the debt of ensuring the new APIs are added to the correct contracts and supported on the appropriate platforms. It is probably worth creating a github issue tracking this merge and listing pointers to the commits/issues that we have accepted in future for easier tracking. I also agree with @nguerrera that we can start adding these APIs into the ref for the given contract which will also aid in identifying public API changes.

stephentoub:
> It is probably worth creating a github issue tracking this merge and listing pointers to the commits/issues that we have accepted in future for easier tracking

Great. Please do :wink:

weshaggard:
@stephentoub I've created item https://github.com/dotnet/corefx/issues/2869 to track the merge of future to master. I will try and do a pass over the existing PR's and add links into the main item there, but please feel free to assist with the links and keeping it updated.

stephentoub:
@dotnet-bot test this please

PR #2864, opened 2015-08-18
Fix System.Native shim on FreeBSD

janhenke:
The error codes ENODATA, ENOSR, ENOSTR and ETIME are not defined on
FreeBSD. Guard these codes with conditional compilation and return
EINVAL.

### DIFF
--- a/src/Common/src/Interop/Unix/Interop.Errors.cs
+++ b/src/Common/src/Interop/Unix/Interop.Errors.cs
@@ -61,7 +61,6 @@ internal static partial class Interop
         ENETUNREACH      = 0x10028,           // Network unreachable.
         ENFILE           = 0x10029,           // Too many files open in system.
         ENOBUFS          = 0x1002A,           // No buffer space available.
-        ENODATA          = 0x1002B,           // No message is available on the STREAM head read queue.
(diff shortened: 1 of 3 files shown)


janhenke:
@nguerrera can you review this change?

nguerrera:
Thanks, @janhenke

I see that I missed that these are annotated as obsolescent and optional: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html

We should just remove them as we don't need them and it will keep the code tighter. If they were to show up with them removed, they'd return PAL_ENONSTANDARD in one direction, which is OK by me. The other direction is for Interop.Error.EXXX.Info(), which we would not even be able to do if they were removed. (Make sure to remove them from Interop.Errors.cs as well.)

cc @sokket

janhenke:
Okay, I'll rework the PR to completely remove those error codes.

janhenke:
@nguerrera I have committed and squashed the change. It no drops those 4 error codes.

stephentoub:
LGTM

jasonwilliams200OK:
Thanks @janhenke! :+1: :100: :rocket:

nguerrera:
LGTM. Thanks again!

PR #2867, opened 2015-08-18
Add Pipes test for chain of writers/readers

stephentoub:
Adds a test that does lots of async writes/reads with various size buffers and verifies that the right data comes through.

cc: @Maxwe11, @ianhays

### DIFF
--- a/src/System.IO.Pipes/tests/PipeTest.Read.cs
+++ b/src/System.IO.Pipes/tests/PipeTest.Read.cs
@@ -251,5 +251,43 @@ namespace System.IO.Pipes.Tests
                 Assert.Equal(123, pair.readablePipe.ReadByte());
             }
         }
+
+        [Theory]
+        [InlineData(5000, 1, 1)]       // very small buffers
+        [InlineData(500, 21, 18)]      // lots of iterations, with read buffer smaller than write buffer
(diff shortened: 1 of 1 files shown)


src/System.IO.Pipes/tests/PipeTest.Read.cs
```diff
@@ -251,5 +251,43 @@ public void ValidWriteByte_ValidReadByte()
...
+                            new ArraySegment<byte>(readBuffer, 0, numRead));
+                        totalRead += numRead;
+                    }
+                    Assert.Equal(writeBuffer.Length, totalRead);
+
+                    await writerTask;
+                }
+            }
+        }
+
```
ianhays: nit: extra line

ianhays:
Good addition! LGTM

Maxwe11:
:+1:

src/System.IO.Pipes/tests/PipeTest.Read.cs
stephentoub: Thanks; this one was actually on purpose... personal preference, but I like having a blank line between the end of the methods and the end of the class so that it's easier for me to see where to add more methods.

PR #2868, opened 2015-08-18
Fix hashing on Windows 7.

bartonjs:
GetHashSizeInBytes got inlined into HashProviderCng..ctor, but it did so using the local variable hHash, not the field _hHash.  The Windows 7 execution path results in the field being set, but not the local, resulting in STATUS_INVALID_HANDLE on that call, and results in a CryptographicException.


bartonjs:
cc: @AtsushiKan

stephentoub:
LGTM

bartonjs:
@AtsushiKan Yeah, I sort of wanted to break it back out into a method for computing the size, and extract the central body of the ctor out to `OpenHashHandle`; but resisted the temptation for a bugfix change.

While the braces can be easily overlooked/edited away later, they add enough protection that I'll go ahead and reset my CI clock :).

PR #2870, opened 2015-08-18
Implement ECDsa and ECDsaCng

AtsushiKan:
cc @bartonjs

### DIFF
--- a/src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaStub.cs
+++ b/src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaStub.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Security.Cryptography;
+
+namespace System.Security.Cryptography.EcDsa.Tests
+{
+    // Stub out the last remaining abstract members to throw NotImplementedException
+    internal class ECDsaStub : ECDsa
+    {
+        public override byte[] SignHash(byte[] hash)
+        {
+            return Array.Empty<byte>();
+        }
+
+        public override bool VerifyHash(byte[] hash, byte[] signature)
+        {
+            return false;
+        }
+
+        protected override byte[] HashData(byte[] data, int offset, int count, HashAlgorithmName hashAlgorithm)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override byte[] HashData(Stream data, HashAlgorithmName hashAlgorithm)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
+++ b/src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
@@ -1,0 +1,440 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using Xunit;
+
+namespace System.Security.Cryptography.EcDsa.Tests
+{
+    public partial class ECDsaTests
+    {
+        public static IEnumerable<object[]> AllImplementations()
+        {
+            return new[] { 
+                new ECDsa[] { ECDsaFactory.Create() },
+                new ECDsa[] { new ECDsaStub() },
+            };
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArray_NullData_ThrowsArgumentNullException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentNullException>("data",
+                () => ecdsa.SignData((byte[])null, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArray_DefaultHashAlgorithm_ThrowsArgumentException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentException>("hashAlgorithm",
+                () => ecdsa.SignData(new byte[0], default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_NullData_ThrowsArgumentNullException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentNullException>("data",
+                () => ecdsa.SignData(null, -1, -1, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_NegativeOffset_ThrowsArgumentOutOfRangeException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("offset",
+                () => ecdsa.SignData(new byte[0], -1, -1, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_OffsetGreaterThanCount_ThrowsArgumentOutOfRangeException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("offset",
+                () => ecdsa.SignData(new byte[0], 2, 1, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_NegativeCount_ThrowsArgumentOutOfRangeException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("count",
+                () => ecdsa.SignData(new byte[0], 0, -1, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_CountGreaterThanLengthMinusOffset_ThrowsArgumentOutOfRangeException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>("count",
+                () => ecdsa.SignData(new byte[0], 0, 1, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_DefaultHashAlgorithm_ThrowsArgumentException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentException>("hashAlgorithm",
+                () => ecdsa.SignData(new byte[0], 0, 0, default(HashAlgorithmName)));
+        }
+
+        [Theory, MemberData("AllImplementations")]
+        public void SignDataByteArraySpan_EmptyHashAlgorithm_ThrowsArgumentException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentException>("hashAlgorithm",
+                () => ecdsa.SignData(new byte[10], 0, 10, new HashAlgorithmName("")));
+        }
+
+        [Theory, MemberData("AllImplementations")]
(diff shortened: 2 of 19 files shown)


stephentoub:
I assume a Unix / OpenSSL implementation will be coming separately?

src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
```diff
@@ -0,0 +1,473 @@
...
+        public void VerifyDataStream_DefaultHashAlgorithm_ThrowsArgumentException(ECDsa ecdsa)
+        {
+            Assert.Throws<ArgumentException>("hashAlgorithm",
+                () => ecdsa.VerifyData(new MemoryStream(), new byte[0], default(HashAlgorithmName)));
+        }
+
+        //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+        [Fact]
+        public void SignDataByteArray_BaseWithValidArguments_ThrowsNotImplementedBecauseHashDataOverrideIsNeeded()
```
bartonjs: This chunk of tests can be removed; they exist in the desktop tests because the HashData overload with HashAlgorithmName had to be implemented as virtual-throw instead of abstract for compatibility reasons.  It doesn't make as much sense here where we were able to call the methods abstract.

AtsushiKan:
Yes. There will be a ramp-up period for getting a Linux VM and workflow set up.

src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
```diff
@@ -0,0 +1,473 @@
...
+        
+        public static IEnumerable<object[]> RealImplementations()
+        {
+            return new[] { 
+                new ECDsa[] { ECDsaFactory.Create() },
+            };
+        }
+
+        [Theory, MemberData("RealImplementations")]
+        public void SignHash_NullHash_ThrowsArgumentNullException(ECDsa ecdsa)
```
bartonjs: I could be wrong, but I don't really see a situation where we'd have more than one implementation type per library; so it feels to me like the Theory-ness can be squished down to [Fact] and the variable replaced with a factory call.

But it might not be worth the trouble.

stephentoub:
> Yes. There will be a ramp-up period for getting a Linux VM and workflow set up

Sounds good; just wanted to make sure it was coming.

src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
```diff
@@ -0,0 +1,473 @@
+using System;
```
bartonjs: Missing license header

src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaStub.cs
```diff
@@ -0,0 +1,30 @@
+using System;
```
bartonjs: Missing license header

src/Common/tests/Cryptography/AlgorithmImplementations/ECDsa/ECDsaTests.cs
AtsushiKan: I assumed the Stub class was about verifying the behavior of the abstract ECDsa class. Since it’s third-party derivable, it is an api (albeit a paper-thin one) that arguably calls for its own test coverage.

bartonjs: Yeah, I didn't mean up in the AllImplementations land, just down here in RealImplementations.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                _lazyKey.Dispose();
+            }
+
+            _lazyKey = key;
+        }
+
+        public static byte[] HashData(byte[] data, int offset, int count, HashAlgorithmName hashAlgorithm)
+        {
+            // we're sealed and the base should have checked this already.
```
bartonjs: The comment doesn't quite apply now that you've factored it out of an instance method :)

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                hashProvider.AppendHashData(data, offset, count);
+                byte[] hash = hashProvider.FinalizeHashAndReset();
+                return hash;
+            }
+        }
+
+        public static byte[] HashData(Stream data, HashAlgorithmName hashAlgorithm)
+        {
+            // We're sealed and the base should have checked these already.
```
bartonjs: Another instance of the doesn't-quite-fit comment

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+                byte[] hash = hashProvider.FinalizeHashAndReset();
+                return hash;
+            }
+        }
+
+        public static unsafe byte[] SignHash(CngKey key, byte[] hash, AsymmetricPaddingMode paddingMode, void* pPaddingInfo, int estimatedSize)
+        {
+#if DEBUG
+            estimatedSize = 2;  // Make sure the NTE_BUFFER_TOO_SMALL scenario gets exercised.
```
bartonjs: What about the over-large scenario?  And the just right one?

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,144 @@
...
+                while ((bytesRead = data.Read(buffer, 0, buffer.Length)) > 0)
+                {
+                    hashProvider.AppendHashData(buffer, 0, bytesRead);
+                }
+                byte[] hash = hashProvider.FinalizeHashAndReset();
+                return hash;
+            }
+        }
+
+        public static unsafe byte[] SignHash(CngKey key, byte[] hash, AsymmetricPaddingMode paddingMode, void* pPaddingInfo, int estimatedSize)
```
bartonjs: I know that these days we all have big monitors that can hold lots of text; but the github review UI is limited to about 114 characters.  It's nice when the code doesn't require horizontal scrolling.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            byte[] signature = new byte[estimatedSize];
+            int numBytesNeeded;
+            ErrorCode errorCode = Interop.NCrypt.NCryptSignHash(keyHandle, pPaddingInfo, hash, hash.Length, signature, signature.Length, out numBytesNeeded, paddingMode);
+            if (errorCode == ErrorCode.NTE_BUFFER_TOO_SMALL)
+            {
+                signature = new byte[numBytesNeeded];
+                errorCode = Interop.NCrypt.NCryptSignHash(keyHandle, pPaddingInfo, hash, hash.Length, signature, signature.Length, out numBytesNeeded, paddingMode);
+            }
+            if (errorCode != ErrorCode.ERROR_SUCCESS)
```
bartonjs: This if touching the previous if made me read it as an else if (despite the lack of the word else).  Can you separate it out with a blank line? (I'm also a fan of explicit bracing outside of precondition checks, but I'll be happy with the blank line)

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
AtsushiKan: There are only two build flavors so our ability to auto-coverage this way is limited.

As a practical matter, retail builds do the just-right case on every EcDsaCng.SignHash() and the over-large case on (almost) every RsaCng.Decrypt().

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/ECDsaCng.SignVerify.cs
```diff
@@ -0,0 +1,63 @@
...
+            int estimatedSize;
+            switch (KeySize)
+            {
+                case 256: estimatedSize = 64; break;
+                case 384: estimatedSize = 96; break;
+                case 521: estimatedSize = 132; break;
+                default:
+                    // If we got here, the range of legal key sizes for ECDsaCng was expanded and someone didn't update this switch.
+                    // Since it isn't a fatal error to miscalculate the estimatedSize, don't throw an exception. Just truck along.
+                    estimatedSize = 132;
```
bartonjs: The first two are exactly KeySize / 4, the third one comes in a little shallow on that approach.  But maybe KeySize / 4 is a better default than 132?

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
AtsushiKan: I don’t know about this one – the errorCode check is really tightly bound to the Interop call(s). Adding a blank line sends the wrong message to my brain.

src/System.Security.Cryptography.Cng/tests/ECDsaCngTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            ECDsaCng e = new ECDsaCng(key);
+
+            {
+                byte[] sig = ("998791331eb2e1f4259297f5d9cb82fa20dec98e1cb0900e6b8f014a406c3d02cbdbf5238bde471c3155fc25565524301429"
+                            + "d8713dad9a67eb0a5c355e9e23dc").HexToByteArray();
+                bool verified = e.VerifyHash(TestData.s_hashSha512, sig);
+                Assert.True(verified);
+            }
+
+            {
```
bartonjs: Please split the positive and negative halves into separate methods.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngAsymmetricAlgorithmCore.cs
AtsushiKan: I shudder to think how much vertical scrolling we’d add if we formatted code to accommodate GitHub’s (inadequate) PR UI.

I pull down commits and use windiff [insert favorite diffing tool] when I want to do a major CR.

bartonjs:
Part of me wants to hold this until the Unix implementation is written, so that we don't end up risking platform disparity.  But, since it can't really be used (aside from by a very determined person) without the ref assembly being updated; I guess it's a reasonable approach to do it in stages.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/ECDsaCng.Key.cs
```diff
@@ -0,0 +1,54 @@
...
+                int keySize = KeySize;
+                // Map the current key size to a CNG algorithm name
+                CngAlgorithm algorithm = null;
+                switch (keySize)
+                {
+                    case 256: algorithm = CngAlgorithm.ECDsaP256; break;
+                    case 384: algorithm = CngAlgorithm.ECDsaP384; break;
+                    case 521: algorithm = CngAlgorithm.ECDsaP521; break;
+                    default:
+                        Debug.Assert(false, "Illegal key size set");
```
stephentoub: ```Debug.Assert(false, ...)``` can be ```Debug.Fail(...)```.

AtsushiKan: Sure, I can throw this one in when I squash.

PR #2871, opened 2015-08-18
Don't do null check on every loop when filling buffer.

hackcraft:
Buffer has an optimisation of using a null array to represent an empty state, that
also plays well with it being a value type.

When the buffer is filled from an enumerable, this possible null state is checked
for on every loop.

Restructure to avoid this redundancy.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -3442,18 +3442,21 @@ namespace System.Linq
 
             if (items == null)
             {
-                foreach (TElement item in source)
+                using (IEnumerator<TElement> e = source.GetEnumerator())
                 {
-                    if (items == null)
+                    if (e.MoveNext())
                     {
                         items = new TElement[4];
+                        items[0] = e.Current;
+                        count = 1;
+                        
+                        while (e.MoveNext())
+                        {
+                            if (items.Length == count) Array.Resize(ref items, checked(count * 2));
+                            items[count] = e.Current;
+                            ++count;
+                        }
                     }
-                    else if (items.Length == count)
-                    {
-                        Array.Resize(ref items, checked(count * 2));
-                    }
-                    items[count] = item;
-                    count++;
                 }
             }
 


stephentoub:
This duplicates #2668 from @James-Ko.  However, this version addresses the concerns raised there, so I'd be inclined to take this version (with thanks to you both).  There was a question on the other PR about perf impact... do you have measurements with results you can share?

cc: @VSadov

James-Ko:
Great, I'll close my PR. Thanks.

hackcraft:
I compared some methods with just those two fragements, and measurements varied a lot depending on the source enumerable, from around less than 1% worse to 20% better (but only one case was that extreme), with most a couple of percent better.

If the source was something like an array or a list that could iterate quickly by moving an index against an array it actually did quite a bit worse, but that case wouldn't be hit here.

hackcraft:
@James-Ko I didn't see that PR. GMTA. The addition of `return`s in yours has an advantage, that we can also get with reordering, without abandoning the use of locals.

stephentoub:
@dotnet-bot test this please

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3431,32 +3432,34 @@ internal Buffer(IEnumerable<TElement> source, bool queryInterfaces = true)
                     if (collection != null)
                     {
                         count = collection.Count;
-                        if (count > 0)
+                        if (count != 0)
                         {
                             items = new TElement[count];
                             collection.CopyTo(items, 0);
                         }
+                        goto storeArray;
```
stephentoub: I liked the previous version that just touched the below code (optimizing away the null check on each iteration) much more than I like this version with the gotos.  It seems like we're adding complexity and code simply to avoid an additional null check / branch on each ToArray call, but there's already a whole bunch of casts and null checks and branches on these code paths.  Does this really amount to a measurable win?

hackcraft: You're completely correct. The only case where it saves anything but a single compare is in the case of a zero-item collection source, and there not only does it not save much, but that case has a much more straightforward way of dealing with it by just bailing out then, anyway.

hackcraft:
Have reset on the moment of madness in those last two commits.

James-Ko:
@hackcraft Agreed, I didn't think the gotos should be necessary either.

stephentoub:
LGTM.  Thanks.

stephentoub:
@VSadov, you ok with the change?

VSadov:
LGTM

PR #2872, opened 2015-08-18
Add tests for System.Runtime based on mscorlib code coverage data

Sridhar-MS:
System.Array - add 'BinarySearch' overload tests
System.DateTime - add 'GetDateTimeFormats' tests
System.Decimal - add 'GetBits' tests
System.GC - add 'ReRegisterForFinalize', 'GetTotalMemory' and 'GetGeneration' tests
System.Runtime.CompilerServices.ConditionalWeakTable - add new tests
System.TimeSpan - add 'ParseExact' tests
System.Type - add 'GetType' by name tests
System.Uri - fix the C# project to include Uri tests.

### DIFF
--- a/src/System.Runtime/tests/System/Array.cs
+++ b/src/System.Runtime/tests/System/Array.cs
@@ -223,145 +223,85 @@ public static unsafe class ArrayTests
         Assert.Throws<RankException>(() => il4.IndexOf(0));
     }
 
-    [Fact]
-    public static void TestBinarySearch()
+    public static IEnumerable<object[]> BinarySearchTestData
     {
-        //@todo: Must pass in a non-null comparer for now (due to incomplete support in underlying Compare class)
-        //
-        IComparer comparer = new IntegerComparer();
-        IComparer<int> icomparer = new IntegerComparer();
-        int idx;
-        int[] ia = { 1, 3, 6, 6, 8, 10, 12, 16 };
-        idx = Array.BinarySearch((Array)ia, 8, comparer);
-        Assert.Equal(idx, 4);
+        get
+        {
+            int[] intArray = { 1, 3, 6, 6, 8, 10, 12, 16 };
+            IComparer intComparer = new IntegerComparer();
+            IComparer<int> intGenericComparer = new IntegerComparer();
 
-        //Return value semantics: Quoted from MSDN without comment.
-        //
-        //The index of the specified value in the specified array, 
-        //if value is found. If value is not found and value is 
-        //less than one or more elements in array, a negative 
-        //number which is the bitwise complement of the index of 
-        //the first element that is larger than value. 
-        //If value is not found and value is greater than any 
-        //of the elements in array, a negative number which is the bitwise complement of (the index of the last element plus 1).
-        idx = Array.BinarySearch((Array)ia, 99, comparer);
-        Assert.Equal(idx, ~(ia.Length));
+            string[] strArray = { null, "aa", "bb", "bb", "cc", "dd", "ee" };
+            IComparer strComparer = new StringComparer();
+            IComparer<string> strGenericComparer = new StringComparer();
 
-        idx = Array.BinarySearch<int>(ia, 0, 8, 99, icomparer);
-        Assert.Equal(idx, ~(ia.Length));
-
-        idx = Array.BinarySearch((Array)ia, 6, comparer);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one 
-
-        idx = Array.BinarySearch((Array)ia, 0, 8, 6, comparer);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one 
-
-        idx = Array.BinarySearch<int>(ia, 6, icomparer);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one 
-
-        idx = Array.BinarySearch<int>(ia, 0, 8, 6, icomparer);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one
-
-        comparer = new StringComparer();
-        IComparer<String> scomparer = new StringComparer();
-
-        String[] sa = { null, "aa", "bb", "bb", "cc", "dd", "ee" };
-        idx = Array.BinarySearch((Array)sa, "bb", comparer);
-        Assert.Equal(idx, 3);
-
-        idx = Array.BinarySearch<String>(sa, 0, sa.Length, "bb", scomparer);
-        Assert.Equal(idx, 3);
-
-        idx = Array.BinarySearch((Array)sa, 3, 4, "bb", comparer);
-        Assert.Equal(idx, 3);
-
-        idx = Array.BinarySearch<String>(sa, 3, 4, "bb", scomparer);
-        Assert.Equal(idx, 3);
-
-        idx = Array.BinarySearch((Array)sa, 4, 3, "bb", comparer);
-        Assert.Equal(idx, -5);
-
-        idx = Array.BinarySearch<String>(sa, 4, 3, "bb", scomparer);
-        Assert.Equal(idx, -5);
-
-        idx = Array.BinarySearch((Array)sa, 4, 0, "bb", comparer);
-        Assert.Equal(idx, -5);
-
-        idx = Array.BinarySearch<String>(sa, 4, 0, "bb", scomparer);
-        Assert.Equal(idx, -5);
-
-        idx = Array.BinarySearch((Array)sa, 0, 7, null, comparer);
-        Assert.Equal(idx, 0);
-
-        idx = Array.BinarySearch<String>(sa, 0, 7, null, scomparer);
-        Assert.Equal(idx, 0);
+            return new[]
+            {
+               new object[] { intArray, 8, intComparer, intGenericComparer, new Func<int, bool>(i => i == 4) },
+               new object[] { intArray, 99, intComparer, intGenericComparer, new Func<int, bool>(i => i == ~(intArray.Length))  },
+               new object[] { intArray, 6, intComparer, intGenericComparer, new Func<int, bool>(i => i == 2 || i == 3)  },
+               new object[] { strArray, "bb", strComparer, strGenericComparer, new Func<int, bool>(i => i == 2 || i == 3)  },
+               new object[] { strArray, null, strComparer, null, new Func<int, bool>(i => i == 0)  },
+           };
+        }
     }
 
-    [Fact]
-    public static void TestBinarySearch_NoIComparer()
+    [Theory, MemberData("BinarySearchTestData")]
+    public static void TestBinarySearch<T>(T[] array, T value, IComparer comparer, IComparer<T> genericComparer, Func<int, bool> verifier)
     {
-        int idx;
-        int[] ia = { 1, 3, 6, 6, 8, 10, 12, 16 };
-        idx = Array.BinarySearch((Array)ia, 8);
-        Assert.Equal(idx, 4);
+        int idx = Array.BinarySearch((Array)array, value, comparer);
+        Assert.True(verifier(idx));
 
-        //Return value semantics: Quoted from MSDN without comment.
-        //
-        //The index of the specified value in the specified array, 
-        //if value is found. If value is not found and value is 
-        //less than one or more elements in array, a negative 
-        //number which is the bitwise complement of the index of 
-        //the first element that is larger than value. 
-        //If value is not found and value is greater than any 
-        //of the elements in array, a negative number which is the bitwise complement of (the index of the last element plus 1).
-        idx = Array.BinarySearch((Array)ia, 99);
-        Assert.Equal(idx, ~(ia.Length));
+        idx = Array.BinarySearch<T>(array, value, genericComparer);
+        Assert.True(verifier(idx));
 
-        idx = Array.BinarySearch<int>(ia, 0, 8, 99);
-        Assert.Equal(idx, ~(ia.Length));
+        idx = Array.BinarySearch((Array)array, value);
+        Assert.True(verifier(idx));
 
-        idx = Array.BinarySearch((Array)ia, 6);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one 
+        idx = Array.BinarySearch<T>(array, value);
+        Assert.True(verifier(idx));
+    }
 
-        idx = Array.BinarySearch((Array)ia, 0, 8, 6);
-        Assert.True(idx == 2 || idx == 3); // Duplicates are allowed but no promises on which one 
+    public static IEnumerable<object[]> BinarySearchTestDataInRange
+    {
+        get
+        {
+            int[] intArray = { 1, 3, 6, 6, 8, 10, 12, 16 };
+            IComparer intComparer = new IntegerComparer();
+            IComparer<int> intGenericComparer = new IntegerComparer();
 
-        idx = Array.BinarySearch<int>(ia, 6);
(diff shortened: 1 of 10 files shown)


Sridhar-MS:
I am looking into the CI failures.

src/System.Runtime/tests/System/Array.cs
```diff
@@ -297,6 +298,73 @@ public static void TestBinarySearch()
...
+
+        //Return value semantics: Quoted from MSDN without comment.
+        //
+        //The index of the specified value in the specified array, 
+        //if value is found. If value is not found and value is 
+        //less than one or more elements in array, a negative 
+        //number which is the bitwise complement of the index of 
+        //the first element that is larger than value. 
+        //If value is not found and value is greater than any 
+        //of the elements in array, a negative number which is the bitwise complement of (the index of the last element plus 1).
```
stephentoub: What's the purpose of this quote?

src/System.Runtime/tests/System/Array.cs
```diff
@@ -297,6 +298,73 @@ public static void TestBinarySearch()
...
+        Assert.Equal(idx, -5);
+
+        idx = Array.BinarySearch<String>(sa, 4, 0, "bb");
+        Assert.Equal(idx, -5);
+
+        idx = Array.BinarySearch((Array)sa, 0, 7, null);
+        Assert.Equal(idx, 0);
+
+        idx = Array.BinarySearch<String>(sa, 0, 7, null);
+        Assert.Equal(idx, 0);
```
stephentoub: A bunch of these seem like they might be better combined into a theory.

src/System.Runtime/tests/System/Buffer.cs
```diff
@@ -79,6 +79,8 @@ public static void TestByteLength()
 
         i = Buffer.ByteLength(new double[33]);
         Assert.Equal(i, 33 * sizeof(double));
+
```
stephentoub: Should we have a test for each of the primitives to ensure they all succeed?

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -291,6 +294,68 @@ public static void TestTryParseExactA()
         Assert.Equal(s, actual);
     }
 
+    [Fact]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public static void TestGetDateTimeFormats()
+    {
+        // Running in the new thread so that we do not interfere with the Main thread's culture
```
stephentoub: How about grabbing it and then resetting it in a finally instead?

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,85 @@ private class TestObject
...
+
+            Assert.True(TestObject.Finalized);
+        }
+
+        private static void CreateObject()
+        {
+            using (var obj = new TestObject())
+            {
+                GC.SuppressFinalize(obj);
+            }
```
stephentoub: What's the value of the IDisposable implementation? Couldn't this just be:
```C#
var obj = new TestObject();
GC.SuppressFinalize(obj);
GC.ReRegisterForFinalize(obj);
```
?

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,85 @@ private class TestObject
...
+        }
+    }
+
+    [Fact]
+    public static void GetTotalMemoryTest_ForceCollection()
+    {
+        long garbageSize = MakeSomeGarbage(1024 * 1024);
+        long garbageCollected = GC.GetTotalMemory(false) - GC.GetTotalMemory(true);
+        Assert.True(garbageCollected >= garbageSize - 10000); // 10000 bytes threshold to stabilize the test
```
stephentoub: Parallelism is enabled for the System.Runtime tests, right?  Seems like this could be flaky as a result.  If the goal here is to verify that GetTotalMemory forces a collection, how about allocating an object, getting its generation before, and then getting its generation after and verifying that it was promoted?

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,85 @@ private class TestObject
...
+    }
+
+    [Fact]
+    public static void GetTotalMemoryTest_NoForceCollection()
+    {
+        long garbageSize = MakeSomeGarbage(1024 * 1024);
+        long before = GC.GetTotalMemory(false);
+        GC.Collect();
+        long garbageCollected = before - GC.GetTotalMemory(false);
+        Assert.True(garbageCollected >= garbageSize - 10000); // 10000 bytes threshold to stabilize the test
```
stephentoub: Similar concerns to above.

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,85 @@ private class TestObject
...
+            Assert.Equal(i, GC.GetGeneration(obj));
+            GC.Collect();
+        }
+    }
+
+    private static long MakeSomeGarbage(int size)
+    {
+        byte[] bytes = new byte[size];
+        GC.Collect();
+        return bytes.Length;
```
stephentoub: The way you've constructed this, you'r effectively forcing bytes into gen1... is that by design?

src/System.Runtime/tests/System/GC.cs
mellinoe: There should be a way to disable parallelism for this test, right? I think you can put this test into its own "TestCollection" ( [Xunit.TestCollection] ) and that would force xUnit to run it individually. I'm not 100% sure about that, but that's my understanding of the feature.

stephentoub: There's no way to disable parallelism for a single test, unfortunately:
https://github.com/xunit/xunit/issues/318

src/System.Runtime/tests/System/Array.cs
Sridhar-MS: This was copied from an older test. I have fixed this now by rewriting the test as a xunit theory.

src/System.Runtime/tests/System/Array.cs
Sridhar-MS: Fixed.

src/System.Runtime/tests/System/Buffer.cs
Sridhar-MS: Fixed.

src/System.Runtime/tests/System/DateTime.cs
Sridhar-MS: Makes sense. Fixed it.

src/System.Runtime/tests/System/GC.cs
Sridhar-MS: Yes. You are right. I wanted to make sure that these tests are stable by introducing some level of threshold when asserting. But that is not sufficient. I have rewritten them to be much simpler.

src/System.Runtime/tests/System/GC.cs
Sridhar-MS: I am trying the follow the pattern where ReRegisterForFinalize is usually called from within the object (from dispose method or finalizer method itself) to re-register to the finalizer queue. Do you think it makes the test more confusing?

weshaggard:
As for the Uri tests we intentionally split those out into their own test projects cc @davidsh @CIPop, I think we should probably just delete the limited Uri tests we have here in favor of the more complete Uri tests.

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,84 @@ private class TestObject
...
+
+    private class ReRegisterForFinalizeTest
+    {
+        public static void Run()
+        {
+            TestObject.Finalized = false;
+            CreateObject();
+
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
```
weshaggard: Adding @Maoni0 to look at these as I know in the past similar tests have been unreliable as depending on various aspects such as the JIT and such these will not be guaranteed be finialized.

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -291,6 +294,78 @@ public static void TestTryParseExactA()
...
+        Assert.Throws<FormatException>(() => july28.GetDateTimeFormats('x'));
+    }
+
+    [Fact]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public static void TestGetDateTimeFormats_FormatProvider()
+    {
+        DateTime july28 = new DateTime(2009, 7, 28, 5, 23, 15);
+
+        List<string> actualJuly28Formats = july28.GetDateTimeFormats(new CultureInfo("en-US")).ToList();
```
weshaggard: @tarekgh is this going guaranteed to always return the same set of formats for en-us? Or can that change based on the underlying culture data in the OS?

src/System.Runtime/tests/System/GC.cs
mellinoe: Ah I see. So even putting it in its own collection, it would still be run in parallel with other collections (aka the rest of the tests) so it wouldn't actually help.

src/System.Runtime/tests/System/DateTime.cs
tarekgh: We cannot assume the date/time formats are not going to change. this wrong. we shouldn't assume any specific formats in our tests. if we need to go to the details, we should use our specific format in our test (like mm/dd/yyyy) and tests the results.

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -291,6 +294,78 @@ public static void TestTryParseExactA()
...
+        var savedCulture = CultureInfo.CurrentCulture;
+
+        try
+        {
+            CultureInfo.CurrentCulture = new CultureInfo("en-US");
+            DateTime july28 = new DateTime(2009, 7, 28, 5, 23, 15);
+
+            List<string> actualJuly28Formats = july28.GetDateTimeFormats().ToList();
+
+            Assert.Equal(expectedJuly28Formats.OrderBy(t => t), actualJuly28Formats.OrderBy(t => t));
```
tarekgh: please change any test doing this pattern. we cannot assume the date/time format will never change for any culture

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -291,6 +294,78 @@ public static void TestTryParseExactA()
         Assert.Equal(s, actual);
     }
 
+    [Fact]
+    [ActiveIssue(846, PlatformID.AnyUnix)]
+    public static void TestGetDateTimeFormats()
+    {
+        var savedCulture = CultureInfo.CurrentCulture;
```
tarekgh: Also we don't need to set the current culture. we can format the dates with specific culture (like en-US) without changing the current culture.

src/System.Runtime/tests/System/GC.cs
Maoni0: I don't see a description of the problem. Can someone describe what problem you are hitting?

Sridhar-MS: The problem is whether the test for ReRegisterForFinalize has potential to be unreliable. It depends on the Finalizer of an object to be run to test the correctness of ReRegisterForFinalize. The question is whether this can be a source of flakiness in the test. Can you review the test to give feedback on this?

Maoni0: I believe this should be reliable since you are creating the object in CreateObject and upon return of this method I don't see why JIT would ever report it as live (even in dbg build).

Sridhar-MS:
@weshaggard - I had a chat with @davidsh . We decided the keep the Uri tests because it is going to be a while before bringing a bulk of Uri tests into github.

src/System.Runtime/tests/System/DateTime.cs
Sridhar-MS: Thanks @tarekgh for the help. I simplified the test case to to Parse the returned DateTime format string and make sure that works.

src/System.Runtime/tests/System/GC.cs
Sridhar-MS: Thanks @Maoni0 !

tarekgh:
LGTM

CIPop:
@weshaggard 
>As for the Uri tests we intentionally split those out into their own test projects cc @davidsh @CIPop, I think we should probably just delete the limited Uri tests we have here in favor of the more complete Uri tests.

The two tests that I can see here have been renamed in the bigger project but preserved as behavior pinning tests. It should be ok to remove them from here as long as we preserve them in the larger test project.

I can't find the Uri and Iri tests in GitHub at this moment. Is there an open issue tracking it? /cc: @SidharthNabar

src/System.Runtime/tests/System/GC.cs
stephentoub: > Do you think it makes the test more confusing?

Personally, yes, but it's not a big deal, and I understand your motivation for doing so.

src/System.Runtime/tests/System/GC.cs
```diff
@@ -226,4 +226,84 @@ private class TestObject
...
+    [Fact]
+    public static void GetTotalMemoryTest_NoForceCollection()
+    {
+        GC.Collect();
+
+        byte[] bytes = new byte[50000];
+        int genBefore = GC.GetGeneration(bytes);
+
+        Assert.True(GC.GetTotalMemory(false) >= bytes.Length);
+        Assert.True(GC.GetGeneration(bytes) == genBefore);
```
stephentoub: The GetGeneration approach works well for the ForceCollection test.  We do run some risk in this NoForceCollection test: if a GC were to occur between the first and second calls to GetGeneration, the object would be promoted.  But as I don't think it's possible to distinguish between a collection forced by this code and one initiated for some other reason, I'm not sure there's much that can be done about it; I suggest we leave this for now, but be prepared to delete the test if it fails with any frequency.

stephentoub:
LGTM

src/System.Runtime/tests/System/GC.cs
Maoni0: I would change the last line to

`Assert.True(GC.GetGeneration(bytes) >= genBefore);`

If you run this test in a simple way, eg, you run it in a standalone exe with all the other simple tests and in a sequential fashion (in other words, if you were to do a GC.Collect/WaitForPending/Colllect sequence, all the memory from the last test would be gone). Then this test is ok...However, as soon as these conditions are not true (eg, if you run this test with a bunch of other tests at the same time, say if we had an infrastructure that just picked random tests and called their Run method on multiple threads in a process) then this test would be meaningless 'cause you can never predict the generation. It could be the same, larger or smaller than before (there could be a GC that happens between new and the 1st GetGeneration, so genBefore could be >0).

stephentoub: Right.  The problem, though, is that once the condition is changed to >= instead of ==, it's no longer testing what the test was trying to test.  If, for example, there was a bug in GetTotalMemory where it always forced a collection even if the argument was false, the test would still pass, even though the whole point of the test is to verify that the collection isn't forced.  That's why I was suggesting that we either wait to see if it's flaky and how often, or we just delete the test now.

Sridhar-MS: One other option is to move all the System.GC test cases into a separate test project. That way the xunit parallelism will not affect these GC tests.

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -2,7 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
```
tarekgh: could you please add the test case I mentioned to you which is listed in the issue 
https://github.com/dotnet/coreclr/issues/1322
Thx

Sridhar-MS: I forgot about that test. I will create a new pull request for that.

PR #2873, opened 2015-08-18
Unix NamedServerPipeStream disposal no longer deletes the FIFO

### LINKED ISSUE #2764: NamedPipeServerStream constructor allows MaxNumberOfServerInstances > 1
The constructor for NamedPipeServerStream allows MaxNumberOfServerInstances > 1 and we even do the same validity checks (between 1-255) as we do on Windows. However, Unix Named pipes don't support more than one connection and an exception is thrown when more are attempted.

ianhays:
Per the discussion in #2764, the clean up of a Unix FIFO named pipe upon disposal of the NamedPipeServerStream that created it has been removed.

- Resolves #2764

cc: @stephentoub

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
@@ -15,7 +15,6 @@ namespace System.IO.Pipes
     /// </summary>
     public sealed partial class NamedPipeServerStream : PipeStream
     {
-        private bool _createdFifo;
         private string _path;
         private PipeDirection _direction;
         private PipeOptions _options;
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
```diff
@@ -123,12 +122,9 @@ public void Disconnect()
 
         protected override void Dispose(bool disposing)
         {
-            // If we created the FIFO object, be a good citizen and clean it up.
-            // If this doesn't happen, worst case is we leave a temp file around.
-            if (_createdFifo && _path != null)
-            {
-                Interop.libc.unlink(_path); // ignore any errors
-            }
+            // Don't clean up the FIFO object, regardless of whether we created it. If we remove
```
stephentoub: We shouldn't need to override Dispose at all now.  Consider removing it entirely and putting the comment somewhere else that makes sense.

stephentoub:
LGTM. Thanks.

src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs
ianhays: I'll move the comment to where we create the FIFO and change it up a bit to make sense there.

PR #2875, opened 2015-08-18
Removed the Unix restriction on MaxNumberOfServerInstances

### LINKED ISSUE #2764: NamedPipeServerStream constructor allows MaxNumberOfServerInstances > 1

ianhays:
There were previously checks to ensure that MaxNumberOfServerInstances was within the same boundaries as it was on Windows despite never actually being used in the Unix implementation. This commit removes those checks and adds platform specific tests for the new behavior.

- Resolves #2764


ianhays:
cc: @stephentoub

stephentoub:
Thanks, Ian. LGTM.

PR #2878, opened 2015-08-18
Consolidate *CompletionSources logic in System.IO.Pipes

Maxwe11:
Common cancellation and completion logic moved into base class.

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
@@ -21,6 +21,7 @@ namespace System.IO.Pipes
         private readonly ThreadPoolBoundHandle _threadPoolBinding;
 
         private CancellationTokenRegistration _cancellationRegistration;
+        private int _errorCode;
         private NativeOverlapped* _overlapped;
         private int _state;
 
@@ -52,8 +53,6 @@ namespace System.IO.Pipes
             [SecurityCritical]get { return _overlapped; }
         }
 
-        protected int ErrorCode { get; private set; }
-
         internal void RegisterForCancellation()
         {
 #if DEBUG
@@ -83,7 +82,7 @@ namespace System.IO.Pipes
 
                 // If we have the result state of completed IO call CompleteCallback(result).
                 // Otherwise IO not completed.
-                if (state == ResultSuccess || state == ResultError)
+                if ((state & (ResultSuccess | ResultError)) != 0)
                 {
                     CompleteCallback(state);
                 }
@@ -107,9 +106,16 @@ namespace System.IO.Pipes
 
         protected virtual void AsyncCallback(uint errorCode, uint numBytes)
         {
-            ErrorCode = (int)errorCode;
-
-            int resultState = errorCode == 0 ? ResultSuccess : ResultError;
+            int resultState;
+            if (errorCode == 0)
+            {
+                resultState = ResultSuccess;
+            }
+            else
+            {
+                resultState = ResultError;
+                _errorCode = (int)errorCode;
+            }
 
             // Store the result so that other threads can observe it
             // and if no other thread is registering cancellation, continue.
@@ -125,7 +131,7 @@ namespace System.IO.Pipes
             }
         }
 
-        protected abstract void HandleError();
+        protected abstract void HandleError(int errorCode);
 
         private void Cancel()
         {
@@ -133,16 +139,13 @@ namespace System.IO.Pipes
             NativeOverlapped* overlapped = Overlapped;
 
             // If the handle is still valid, attempt to cancel the IO
-            if (!handle.IsInvalid)
+            if (!handle.IsInvalid && !Interop.mincore.CancelIoEx(handle, overlapped))
             {
-                if (!Interop.mincore.CancelIoEx(handle, overlapped))
-                {
-                    // This case should not have any consequences although
-                    // it will be easier to debug if there exists any special case
-                    // we are not aware of.
-                    int errorCode = Marshal.GetLastWin32Error();
-                    Debug.WriteLine("CancelIoEx finished with error code {0}.", errorCode);
-                }
+                // This case should not have any consequences although
+                // it will be easier to debug if there exists any special case
+                // we are not aware of.
+                int errorCode = Marshal.GetLastWin32Error();
+                Debug.WriteLine("CancelIoEx finished with error code {0}.", errorCode);
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/System.IO.Pipes/src/System/IO/Pipes/PipeIOCompletionSource.cs
```diff
@@ -2,126 +2,40 @@
...
-                Debug.Assert(completionSource.Overlapped == pOverlapped);
-
-                completionSource.AsyncCallback(errorCode, numBytes);
-            }, this, buffer);
-        }
-
-        internal NativeOverlapped* Overlapped
-        {
-            [SecurityCritical]get { return _overlapped; }
+            _numBytes = 0;
```
stephentoub: This setting of _numBytes isn't necessary, right?  It'll default to 0.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Windows.cs
```diff
@@ -312,7 +312,7 @@ private Task WriteAsyncCorePrivate(byte[] buffer, int offset, int count, Cancell
             }
             else
             {
-                var completionSource = new PipeStreamCompletionSource(this, buffer, cancellationToken, isWrite: true);
+                var completionSource = new PipeIOCompletionSource(this, buffer, cancellationToken, isWrite: true);
```
stephentoub: Reading through this again, I still think ReadWriteCompletionSource would be a better name.  Then we have ConnectionCompletionSource for connecting, ReadWriteCompletionSource for reading/writing, and PipeCompletionSource for the base.

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+                else if (state != CompletedCallback)
+                {
+                    // IO already completed and we have grabbed result state.
+                    // Set NoResult to prevent invocation of CompleteCallback(result state) from AsyncCallback(...)
+                    state = Interlocked.Exchange(ref _state, NoResult);
+                }
+
+                // If we have the result state of completed IO call CompleteCallback(result).
+                // Otherwise IO not completed.
+                if (state == ResultSuccess || state == ResultError)
```
stephentoub: I believe this could instead be:
```C#
if ((state & (ResultSuccess | ResultError)) != 0)
```
to avoid the extra conditional.

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+            }
+        }
+
+        internal abstract void SetCompletedSynchronously();
+
+        protected virtual void AsyncCallback(uint errorCode, uint numBytes)
+        {
+            ErrorCode = (int)errorCode;
+
+            int resultState = errorCode == 0 ? ResultSuccess : ResultError;
```
stephentoub: These two lines could instead be:
```C#
if (errorCode == 0)
{
    resultState = ResultSuccess;
}
else
{
    ErrorCode = (int)errorCode;
    resultState = ResultError;
}
```
Then we don't have to store the error code if we'd just be overwriting 0 with 0.

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+                // Now try to prevent invocation of CompleteCallback(resultState) from RegisterForCancellation().
+                // Otherwise, thread responsible for registering cancellation stole the result and it will invoke CompleteCallback(resultState).
+                if (Interlocked.Exchange(ref _state, CompletedCallback) != NoResult)
+                {
+                    CompleteCallback(resultState);
+                }
+            }
+        }
+
+        protected abstract void HandleError();
```
stephentoub: Rather than having a protected ErrorCode property, what if the base just passed the error code into HandleError?

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+                // If we have the result state of completed IO call CompleteCallback(result).
+                // Otherwise IO not completed.
+                if (state == ResultSuccess || state == ResultError)
+                {
+                    CompleteCallback(state);
+                }
+            }
+        }
+
+        internal void ReleaseResources()
```
stephentoub: Can this be private?

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+            _overlapped = _threadPoolBinding.AllocateNativeOverlapped((errorCode, numBytes, pOverlapped) =>
+            {
+                var completionSource = (PipeCompletionSource<TResult>)ThreadPoolBoundHandle.GetNativeOverlappedState(pOverlapped);
+                Debug.Assert(completionSource.Overlapped == pOverlapped);
+
+                completionSource.AsyncCallback(errorCode, numBytes);
+            }, this, pinData);
+        }
+
+        internal NativeOverlapped* Overlapped
```
stephentoub: Is this Overlapped property accessed anywhere except for this base type now?  If so, could we just get rid of the property and have all accesses be to the _overlapped field?

stephentoub: Nevermind; I see it's used by the consumers.

src/System.IO.Pipes/src/System/IO/Pipes/PipeCompletionSource.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        private void Cancel()
+        {
+            SafeHandle handle = _threadPoolBinding.Handle;
+            NativeOverlapped* overlapped = Overlapped;
+
+            // If the handle is still valid, attempt to cancel the IO
+            if (!handle.IsInvalid)
+            {
+                if (!Interop.mincore.CancelIoEx(handle, overlapped))
```
stephentoub: This could be simplified slightly to
```C#
if (!handle.IsInvalid && !Interop.mincore.CancelIoEx(handle, overlapped))
{
    ...
}
```
rather than having the nested ifs.  Either ways fine, though.

stephentoub:
Thanks for the consolidation, @Maxwe11.

Maxwe11:
updated

stephentoub:
Looks great.  Thanks, @Maxwe11.

cc: @ianhays

PR #2880, opened 2015-08-19
Improve tasks tests

mellinoe:
Here are some small, incremental improvements to some of the Tasks tests. I cleaned up four files that had some particularly ugly assertions and assertion messages, and which were all crammed into a single "test method" in each class. In general I did the following:

* Split things into individual test methods where it was simple
* Replace old assertion calls with regular, standard ones
* Remove old, unnecessary comments, debug statements, dead code, etc
* Formatting, etc.

This is just my incremental progress from the past day, I'm still planning on doing more cleanup in time, as this only covers 4 of the 30 files in the project (some of which I cleaned up enough previously). We could merge the fixes in incrementally or keep this PR open for more commits.

@stephentoub

### DIFF
--- a/src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
+++ b/src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
@@ -13,212 +13,206 @@ namespace System.Threading.Tasks.Tests
 {
     public class AsyncTaskMethodBuilderTests
     {
-        // building tasks
+        // Test captured sync context with successful completion (SetResult)
         [Fact]
-        public static void RunAsyncMethodBuilderTests()
+        public static void VoidMethodBuilder_TrackedContext()
         {
-            // AsyncVoidMethodBuilder
+            SynchronizationContext previousContext = SynchronizationContext.Current;
+            var trackedContext = new TrackOperationsSynchronizationContext();
+            SynchronizationContext.SetSynchronizationContext(trackedContext);
+
+            // TrackedCount should increase as Create() is called, and decrease as SetResult() is called.
+
+            // Completing in opposite order as created.
+            var avmb1 = AsyncVoidMethodBuilder.Create();
+            Assert.Equal(1, trackedContext.TrackedCount);
+            var avmb2 = AsyncVoidMethodBuilder.Create();
+            Assert.Equal(2, trackedContext.TrackedCount);
+            avmb2.SetResult();
+            Assert.Equal(1, trackedContext.TrackedCount);
+            avmb1.SetResult();
+            Assert.Equal(0, trackedContext.TrackedCount);
+
+            // Completing in same order as created
+            avmb1 = AsyncVoidMethodBuilder.Create();
+            Assert.Equal(1, trackedContext.TrackedCount);
+            avmb2 = AsyncVoidMethodBuilder.Create();
+            Assert.Equal(2, trackedContext.TrackedCount);
+            avmb1.SetResult();
+            Assert.Equal(1, trackedContext.TrackedCount);
+            avmb2.SetResult();
+            Assert.Equal(0, trackedContext.TrackedCount);
+
+            SynchronizationContext.SetSynchronizationContext(previousContext);
+        }
+
+        // Test not having a sync context with successful completion (SetResult)
+        [Fact]
+        public static void VoidMethodBuilder_NoContext()
+        {
+            SynchronizationContext previousContext = SynchronizationContext.Current;
+            try
             {
-                // Test captured sync context with successful completion (SetResult)
-                {
-                    SynchronizationContext previousContext = SynchronizationContext.Current;
-                    var trackedContext = new TrackOperationsSynchronizationContext();
-                    SynchronizationContext.SetSynchronizationContext(trackedContext);
-
-                    // Completing in opposite order as created
-                    var avmb1 = AsyncVoidMethodBuilder.Create();
-                    Assert.True(trackedContext.TrackedCount == 1, "RunAsyncMethodBuilderTests > FAILED: Should have been one active builder (SetResult, opposite order).");
-                    var avmb2 = AsyncVoidMethodBuilder.Create();
-                    Assert.True(trackedContext.TrackedCount == 2, "RunAsyncMethodBuilderTests > FAILED: Should have been two active builders (SetResult, opposite order).");
-                    avmb2.SetResult();
-                    Assert.True(trackedContext.TrackedCount == 1, "RunAsyncMethodBuilderTests > FAILED: Completed builder should have decremented count (SetResult, opposite order).");
-                    avmb1.SetResult();
-                    Assert.True(trackedContext.TrackedCount == 0, "RunAsyncMethodBuilderTests > FAILED: Completed builder should have returned count to 0 (SetResult, opposite order).");
-
-                    // Completing in same order as created
-                    avmb1 = AsyncVoidMethodBuilder.Create();
-                    Assert.True(trackedContext.TrackedCount == 1, "RunAsyncMethodBuilderTests > FAILED: Should have been one active builder (SetResult, same order).");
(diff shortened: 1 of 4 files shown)


src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
```diff
@@ -13,212 +13,206 @@ namespace System.Threading.Tasks.Tests
...
+            Assert.Equal(2, trackedContext.TrackedCount);
+            avmb1.SetResult();
+            Assert.Equal(1, trackedContext.TrackedCount);
+            avmb2.SetResult();
+            Assert.Equal(0, trackedContext.TrackedCount);
 
-                    SynchronizationContext.SetSynchronizationContext(previousContext);
-                }
-            }
+            SynchronizationContext.SetSynchronizationContext(previousContext);
```
stephentoub: Resetting the context should ideally be done in a finally block (here and elsewhere the pattern is employed).

src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
```diff
@@ -13,212 +13,206 @@ namespace System.Threading.Tasks.Tests
...
+            Assert.Equal(0, trackedContext.TrackedCount);
+            atmb.SetResult("async");
+            Assert.Equal(0, trackedContext.TrackedCount);
+
+            SynchronizationContext.SetSynchronizationContext(previousContext);
+        }
+
+        // Incorrect usage for AsyncTaskMethodBuilder
+        [Fact]
+        [ActiveIssue("Hangs")]
```
stephentoub: This test hangs?

src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
```diff
@@ -13,212 +13,206 @@ namespace System.Threading.Tasks.Tests
...
+        public static void TaskMethodBuilder_IncorrectUsage()
+        {
+            var atmb = new AsyncTaskMethodBuilder();
+            Assert.Throws<ArgumentNullException>(() => { atmb.SetException(null); });
+        }
+
+        // Incorrect usage for AsyncVoidMethodBuilder
+        [Fact]
+        [ActiveIssue("Hangs")]
```
stephentoub: This test is invalid as written; it makes sense that it hangs.  The void method builder posts an OperationStarted to the current SynchronizationContext if there is one, the point of which is to allow a host to wait for the corresponding OperationCompleted event.  xunit has a custom SynchronizationContext it uses, and it supports waiting until all of the operations have completed. You'll want to add a line after the assert that completes the atmb successfully so that the appropriate event is raised.  Alternatively you could null out the sync ctx prior to Create'ing the ATMB, and then reset the context in a finally block, as is done in some other tests.

src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/AsyncTaskMethodBuilderTests.cs
```diff
@@ -233,244 +227,232 @@ public static void AsyncMethodBuilderCreate_SetExceptionTest2()
...
-                catch (Exception e) { tcs.SetException(e); }
-                Assert.True(ValidateFaultedTask(tcs.Task), "     > FAILURE. Task's stack trace is incorrect");
-            }
+            var atmb = new AsyncTaskMethodBuilder<int>();
+            var t1 = atmb.Task;
+            var t2 = atmb.Task;
+            Assert.NotNull(t1);
+            Assert.NotNull(t2);
+            Assert.Same(t1, t2);
+        }
```
stephentoub: It'd be nice to add a variation to this test that validates an important optimization these have of using a cached task if the method builder is completed synchronously and there's a cached task available, e.g.

```C#
[Fact]
public static void TaskMethodBuilder_UsesCompletedCache()
{
    var atmb1 = new AsyncTaskMethodBuilder();
    var atmb2 = new AsyncTaskMethodBuilder();
    atmb1.SetResult();
    atmb2.SetResult();
    Assert.Same(atmb1.Task, atmb2.Task);
}

[Theory]
[InlineData(true)]
[InlineData(false)]
public static void TaskMethodBuilderBoolean_UsesCompletedCache(bool result)
{
    TaskMethodBuilderT_UsesCompletedCache(result, shouldBeCached);
}

[Theory]
[InlineData(0, true)]
[InlineData(5, true)]
[InlineData(-5, true)]
[InlineData(42, false)]
public static void TaskMethodBuilderInt32_UsesCompletedCache(int result, bool shouldBeCached)
{
    TaskMethodBuilderT_UsesCompletedCache(result, shouldBeCached);
}

[Theory]
[InlineData((string)null, true)]
[InlineData("test", false)]
public static void TaskMethodBuilderRef_UsesCompletedCache(string result, bool shouldBeCached)
{
    TaskMethodBuilderT_UsesCompletedCache(result, shouldBeCached);
}

private static void TaskMethodBuilderT_UsesCompletedCache<T>(T result, bool shouldBeCached)
{
    var atmb1 = new AsyncTaskMethodBuilder<T>();
    var atmb2 = new AsyncTaskMethodBuilder<T>();

    atmb1.SetResult(result);
    atmb2.SetResult(result);

    Assert.Equal(shouldBeCached, object.ReferenceEquals(atmb1.Task, atmb2.Task));
}
```

src/System.Threading.Tasks/tests/Task/TaskAPMTest.cs
```diff
@@ -18,101 +19,120 @@ namespace System.Threading.Tasks.Tests
...
     /// A class that implements the APM pattern to ensure that TPL can support APM patttern
     /// </summary>
-    public sealed class TaskAPMTest
+    public sealed class TaskAPMTests : IDisposable
     {
         /// <summary>
         /// Used to indicate whether to test TPL's Task or Future functionality for the APM pattern
         /// </summary>
-        private readonly bool _hasReturnType;
+        private bool _hasReturnType;
```
stephentoub: Why do we need this at all?

src/System.Threading.Tasks/tests/Task/TaskAPMTest.cs
```diff
@@ -18,101 +19,120 @@ namespace System.Threading.Tasks.Tests
...
+        private const int IntInput = 1000;
 
         /// <summary>
-        /// Ctor that reads the XML testcase and sets the boolean to indicate whether to test Task or Future
-        /// functionality
+        /// The constant that defines the amount time to spinwait (to simulate work) in the LongTask class
         /// </summary>
-        /// <param name="hasReturnType"></param>
-        public TaskAPMTest(bool hasReturnType)
+        private const int LongTaskMilliseconds = 1000;
```
stephentoub: I think we should decrease this value.  Each test is spinning up a Task that spins for this long.

src/System.Threading.Tasks/tests/TaskFactory/TaskFactory_FromAsyncTests.cs
```diff
@@ -23,36 +23,23 @@ public static void RunAPMFactoryTests()
...
-            if (check.Length != 0)
-            {
-                Assert.True(false, string.Format("RunAPMFactoryTests:    > FAILED on Write3 -- expected empty fac."));
-            }
-            else if (((IAsyncResult)t).AsyncState != stateObject)
-            {
-                Assert.True(false, string.Format("RunAPMFactoryTests:    > FAILED on Write3 -- state object not stored correctly."));
-            }
+            Assert.Equal(0, check.Length);
+            Assert.Equal(stateObject, ((IAsyncResult)t).AsyncState);
```
stephentoub: These checks for the state object should probably be Assert.Same rather than Assert.Equal, though in practice it doesn't matter given that it's an Object.

stephentoub:
Thanks for working to clean these up.

stephentoub:
The revisions LGTM.

PR #2884, opened 2015-08-19
Fix failing TaskWaitAllAny* tests

stephentoub:
Some of these tests have been failing in CI.  While unwieldy in general, the specific issue is that when the tests are shutting down, they cancel any tasks that haven't yet completed, but then simply wait on the tasks to complete.  If the task hadn't yet started, the cancellation request would cause the task to end as Canceled, which would cause the Wait to throw.  The fix is simply to catch and ignore the resulting OperationCanceledException.

cc: @mellinoe


mellinoe:
Thanks, LGTM

PR #2887, opened 2015-08-19
Restore virtual generic usage in LINQ

stephentoub:
During the bring-up of .NET Native, there was a period of time during which virtual generic methods were not supported.  As a result, LINQ's implementation was changed to avoid the usage of virtual generics, which occur in some optimizations around Select usage.  That limitation no longer exists.  This commit restores the original code.

cc: @VSadov, @jaredpar


VSadov:
LGTM

hackcraft:
Oh, hurrah! I was wondering about that, and if virtual generics were blocked everywhere.

stephentoub:
:smile:

PR #2889, opened 2015-08-19
Fix access to user's home directory on Unix

### LINKED ISSUE #2820: Use getpwuid_r to read the user's home directory, not the HOME environment variable

stephentoub:
Today in several places we get a user's home directory by accessing the HOME environment variable.  This is the right place to check first, but in certain cases it may not be set, such as in initialization scripts.

This commit implements a fallback, such that if HOME doesn't have a path, we consult the pw_dir path from a call to getpwuid_r for the current user.

As long as I was adding getpwuid_r to the native shim, I also moved our existing geteuid and getegid imports to the shims.

Fixes #2820 
cc: @nguerrera, @bartonjs, @sokket, @akoeplinger

### DIFF
--- a/src/Common/src/System/IO/PersistedFiles.Unix.cs
+++ b/src/Common/src/System/IO/PersistedFiles.Unix.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 
 namespace System.IO
 {
@@ -81,8 +82,7 @@ namespace System.IO
 
         private static void EnsureUserDirectories()
         {
-            // TODO (2820): Use getpwuid_r(geteuid(), ...) instead of the HOME environment variable
-            string userHomeDirectory = Environment.GetEnvironmentVariable("HOME");
+            string userHomeDirectory = GetHomeDirectory();
 
             if (string.IsNullOrEmpty(userHomeDirectory))
             {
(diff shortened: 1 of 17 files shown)


src/Common/src/System/IO/PersistedFiles.Unix.cs
```diff
@@ -94,5 +94,95 @@ private static void EnsureUserDirectories()
...
+        {
+            // First try to get the user's home directory from the HOME environment variable.
+            // This should work in most cases.
+            string userHomeDirectory = Environment.GetEnvironmentVariable("HOME");
+            if (!string.IsNullOrEmpty(userHomeDirectory))
+                return userHomeDirectory;
+
+            // In initialization conditions, however, the "HOME" enviroment variable may 
+            // not yet be set. For such cases, consult with the password entry.
+            unsafe
```
bartonjs: Should we bring s_userHomeDirectory back, so that we don't have to do this multiple times per process? (Yeah, it'd still be once per library, but that's better than once per call)

bartonjs:
One memoization optimization suggestion, otherwise LGTM.

Thanks for jumping on this :)

src/Common/src/System/IO/PersistedFiles.Unix.cs
sokket: Could there be an issue around memoization if the user changed their home directory? It looks like usermod can be used to change a home directory so there's the potential to return incorrect data if we save it off and then it changes

stephentoub: I went back and forth on this.  It's already cached in the places that use it: EnsureUserDirectories caches it as part of the larger string, and the only other consumer is Console which uses it to find the terminfo database, and it only does that once.  I was also slightly concerned about the home directory changing, either due to the environment variable getting changed or due to the a temporary change of user identity (e.g. seteuid)... though if we were actually concerned about that, we'd want to reevaluate the caching in EnsureUserDirectory as well, since it includes the home directory.  In any event, I think it makes sense to leave it for now without the caching, and then add it later if we determine it's important.  Ok?

bartonjs: Yep, sounds like fine reasoning to me.

sokket:
LGTM

src/Native/System.Native/pal_uid.cpp
```diff
@@ -0,0 +1,74 @@
...
+/**
+* Gets a password structure for the given uid.
+* Implemented as shim to getpwuid_r(3).
+*
+* Returns 0 for success, -1 for failure. Sets errno on failure.
+*/
+extern "C"
+int32_t GetPwUid(
+	int32_t         uid,
+	struct Passwd  *pwd,
```
nguerrera: Nit 1: 'struct' isn't necessary here as we're using C++. 

Nit 2: I prefer Passwd* pwd over Passwd *pwd.

Nit disclaimer: These are just my personal preferences and we haven't really set C++ coding guidelines.

src/Native/System.Native/pal_uid.cpp
```diff
@@ -0,0 +1,74 @@
...
+		// after the call.
+		*pwd = { };
+		*result = nullptr;
+	}
+
+	return rv;
+}
+
+extern "C"
+int32_t GetEUid(void)
```
nguerrera: Another C++-biased style nit: we can declare as GetEUid() rather thant GetEUid(void)

src/Common/src/System/IO/PersistedFiles.Unix.cs
```diff
@@ -94,5 +94,95 @@ private static void EnsureUserDirectories()
...
+                        path = null; // we'll still return true, as false indicates the buffer was too small
+                    }
+                    else
+                    {
+                        Debug.Assert(result == (IntPtr)(&passwd));
+                        Debug.Assert(passwd.HomeDirectory != null);
+                        path = Marshal.PtrToStringAnsi((IntPtr)passwd.HomeDirectory);
+                    }
+                    return true;
+                }
```
nguerrera: I'm not sure, but wondering about replacing result with a return values. e.g positive = not found, zero = success, negative = error. We don't do anything with the result pointer as it can only ever equal the pointer we passed in or null.

nguerrera:
LGTM. Thanks. I left a few nits and one open question, but up to you on all of them.

src/Common/src/System/IO/PersistedFiles.Unix.cs
stephentoub: I'd be fine with that... I was just trying to keep the signature as close to the underlying one as possible.  You want me to change it?

nguerrera: I'm not sure. On the one hand, I like keeping it as 1:1 as possible, but on the other it seems like it would make the code a bit easier to follow. Up to you.

stephentoub: Ok, thanks.  I'll fix the other C++-isms, but I'll leave this one as-is for now.  I'm really hoping we can keep the shims as 1:1 as possible unless we find a need to diverge.

stephentoub:
Updated with the style fixes (and to address a compilation failure for OSX in MemoryMappedFiles... on OSX it needs the PersistedFiles file, which now needs some interop that wasn't being included in the project). Thanks for the reviews, guy.

nguerrera:
:+1:

PR #2892, opened 2015-08-19
Fix order-dependent TypeExtensions tests

### LINKED ISSUE #2619: Three TypeExtensions tests failing on Linux with xUnit 2.1.0
A few of the reflection tests seem to be failing on Linux when run with xUnit 2.1.0. These tests are doing fairly simple stuff, just subscribing some events and then triggering them, and checking the state afterwards.

    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest2
    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest3
    System.Reflection.Compatibility.UnitTests.EventInfoAddEventHandler.PosTest1

mellinoe:
These tests were expected to be run in sequential order, as they all modified a shared static variable on some class. Instead, they now each modify a separate variable and should therefore be order-independent. There was no need for them to be sharing that state; it wasn't important to the test.

Fixes #2619 

@Priya91

### DIFF
--- a/src/System.Reflection.TypeExtensions/tests/EventInfo/EventInfoAddEventHandler.cs
+++ b/src/System.Reflection.TypeExtensions/tests/EventInfo/EventInfoAddEventHandler.cs
@@ -15,7 +15,6 @@ namespace System.Reflection.Compatibility.UnitTests
     {
         // Positive Test 1: add Event handler to the not static event
         [Fact]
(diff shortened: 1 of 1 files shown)


Priya91:
LGTM

stephentoub:
LGTM

Issue #2562, opened 2015-07-29
Process.TestEnableRaiseEvents fails on Linux build


Priya91:
This test failed once in Linux build. EnableRaisingEvents was set to true on Process, and on exiting the specifed callback method was not called. 

```
System.Diagnostics.ProcessTests.ProcessTests.TestEnableRaiseEvents [FAIL]
08:42:31       Assert+WrapperXunitException : File path: d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Diagnostics.Process\tests\System.Diagnostics.Process.Tests\ProcessTests.cs. Line: 88
08:42:31       ---- TestCanRaiseEvents0001: isExited Event not called when EnableRaisingEvent is set to true.
08:42:31       Stack Trace:
08:42:31             at Assert.WrapException(Exception e, String callerFilePath, Int32 callerLineNumber)
08:42:31             at Assert.True(Boolean condition, String userMessage, String path, Int32 line)
08:42:31             at System.Diagnostics.ProcessTests.ProcessTests.TestEnableRaiseEvents()
08:42:31          ----- Inner Stack Trace -----
08:42:31             at Assert.True(Boolean condition, String userMessage, String path, Int32 line)
```

Build [1378](http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/1378/console)

stephentoub:
This just failed again:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst/735/console
```
20:21:32    System.Diagnostics.ProcessTests.ProcessTests.TestEnableRaiseEvents [FAIL]
20:21:32       Assert+WrapperXunitException : File path: d:\j\workspace\dotnet_corefx_windows_debug\src\System.Diagnostics.Process\tests\ProcessTests.cs. Line: 89
20:21:32       ---- TestCanRaiseEvents0001: isExited Event not called when EnableRaisingEvent is set to true.
20:21:32       Expected: True
20:21:32       Actual:   False
20:21:32       Stack Trace:
20:21:33             at Assert.WrapException(Exception e, String callerFilePath, Int32 callerLineNumber)
20:21:33             at Assert.True(Boolean condition, String userMessage, String path, Int32 line)
20:21:33             at System.Diagnostics.ProcessTests.ProcessTests.TestEnableRaiseEvents()
20:21:33          ----- Inner Stack Trace -----
[...]

PR #2900, opened 2015-08-20
Fix Process TestEnableRaiseEvent race condition

stephentoub:
Process.WaitForExit and the callback invoked when a Process exits both call the RaiseOnExited method, which is implemented like this:
```C#
private void RaiseOnExited()
{
    if (!_raisedOnExited)
    {
        lock (this)
        {
            if (!_raisedOnExited)
            {
                _raisedOnExited = true;
                OnExited();
            }
        }
    }
}
```
Because _raisedOnExited is set prior to invoking OnExited, if Process.WaitForExit races with the exited callback to invoke RaiseOnExited, WaitForExit may see _raisedOnExited as true even if OnExited hasn't yet been executed or completed execution.  That means that when WaitForExit returns, there's no guarantee that the Exited delegate has been invoked, yet the TestEnableRaiseEvents method was relying on that.

This commit fixes the test to not rely on that, by spinning a bit waiting for the flag to be set.  I also took the opportunity to rewrite the test as a theory to avoid some duplication.

I contemplated changing the RaiseOnExited implementation itself to do:
```C#
try
{
    OnExited();
}
finally
{
    _raisedOnExited = true;
}
```
instead of:
```C#
_raisedOnExited = true;
OnExited();
```
but that would potentially make certain deadlock conditions more likely, so I left it as is and changed the test instead.

FIxes https://github.com/dotnet/corefx/issues/2562
cc: @Priya91

### DIFF
--- a/src/System.Diagnostics.Process/tests/ProcessTests.cs
+++ b/src/System.Diagnostics.Process/tests/ProcessTests.cs
@@ -74,40 +74,35 @@ namespace System.Diagnostics.ProcessTests
             }
         }
 
-        [Fact]
-        public void TestEnableRaiseEvents()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        [InlineData(null)]
+        public void TestEnableRaiseEvents(bool? enable)
         {
-            {
-                bool isExitedInvoked = false;
+            bool exitedInvoked = false;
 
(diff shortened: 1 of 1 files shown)


Priya91:
LGTM! Good idea :)

PR #2901, opened 2015-08-20
 Use shims for open, close, unlink, shm_open, shm_unlink 

nguerrera:
cc @stephentoub @sokket 

I've decided to put these with stat and renamed pal_stat to pal_io The various IO functions share constant definitions such as O_* open flags and S_I* permissions, etc, so having one IO header makes things more cohesive. The rename is a separate commit (2cd3926).

shm_open is interesting in that we only need it on Linux currently. It pulls in librt there and to my surprise, not passing -lrt resulted in a runtime error, but no linking error. I've added the linker flag everywhere, but suspect it will break on OS X and BSD, which I'll try now.

It is also available on other platforms but doesn't work quite like we'd like. For now, I've just defined it everywhere since it will compile, but only actually call it on Linux as before. We could consider collapsing MemoryMappedFiles into a single UNIX implementation and using runtime detection. 

Also, I took a shortcut with Linux-specific pipe2 by forking the one O_* flag we use with it  and still P/Invoking to libc directly with the raw value for now.

Both pipe2 and shm_open have me thinking about stubbing out functions with E_NOTSUP and having the managed library fallback to alternatives rather than detecting specific OSes or having separate managed builds. This needs more thought and investigation, so I opted to stop here.


src/Native/System.Native/pal_io.h
```diff
@@ -59,6 +59,33 @@ enum
...
+ * up conversion. We can add more as needed.
+ */
+enum
+{
+    // Access modes (mutually exclusive).
+    PAL_O_RDONLY           = 0x0000, // Open for read-only
+    PAL_O_WRONLY           = 0x0001, // Open for write-only
+    PAL_O_RDWR             = 0x0002, // Open for read-write
+
+    // Mask to get just the access mode. Some room is left for mor. POSIX
```
stephentoub: Nit: "mor" => "more"

src/System.Console/src/System.Console.csproj
```diff
@@ -151,21 +154,6 @@
...
-  <ItemGroup Condition=" '$(TargetsLinux)' == 'true' ">
-    <Compile Include="$(CommonPath)\Interop\Linux\libc\Interop.OpenFlags.cs">
-      <Link>Common\Interop\Linux\Interop.OpenFlags.cs</Link>
-    </Compile>
-  </ItemGroup>
-  <ItemGroup Condition=" '$(TargetsOSX)' == 'true' ">
-    <Compile Include="$(CommonPath)\Interop\OSX\libc\Interop.OpenFlags.cs">
-      <Link>Common\Interop\OSX\Interop.OpenFlags.cs</Link>
-    </Compile>
-  </ItemGroup>
```
stephentoub: Yay!

src/System.Net.Http/src/System.Net.Http.csproj
```diff
@@ -149,11 +149,14 @@
...
+    <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.Close.cs">
+      <Link>Common\Interop\Unix\System.Native\Interop.Close.cs</Link>
     </Compile>
-    <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.open.cs">
-      <Link>Common\Interop\Unix\libc\Interop.open.cs</Link>
+    <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.Open.cs">
+      <Link>Common\Interop\Unix\System.Native\Interop.Open.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.OpenFlags.cs">
+      <Link>Common\Interop\Unix\System.Native\Interop.OpenFlags.cs</Link>
```
stephentoub: Now that we don't need separate definitions of the flags per platform, we could recombine OpenFlags.cs back into Open.cs.

nguerrera: I did that at first and broke shm_open which also uses the same flags.

stephentoub: Ah, of course.

stephentoub:
> We could consider collapsing MemoryMappedFiles into a single UNIX implementation and using runtime detection.

I'd be fine with that in this case.  Essentially we're just deciding whether to create a memory or file object, and could make that determination based on platform at runtime rather than at compile time.  Alternatively, we could have the managed code first try ShmOpen, falling back to a file if that fails, and code the shim to return an error if it doesn't delegate to shm_open (essentially making it a compile-time decision in the shim instead of in the managed binary).  I just read the rest of your comment and see that's what you refer to later in it.

> Also, I took a shortcut with Linux-specific pipe2 by forking the one O_* flag we use with it and still P/Invoking to libc directly with the raw value for now.

We use pipe2 on Linux because it supports controlling inheritance.  On OS X we don't have pipe2, so just use pipe, and we end up ignoring the inheritance option.  We're ignoring it in managed code.  We could instead just expose a Pipe function from the shim, using that from managed code always, and have the shim either compile against pipe2 if it's available or pipe otherwise... then the shim rather than the managed code would be the one to ignore the value if it can't use it.

nguerrera:
Thanks. I agree with the proposed direction for both shm_open and pipe. I'll do that as a follow-up.

nguerrera:
(I'll squash that last commit, but keeping it independent until merging so reviews in progress see what's changed.)

nguerrera:
@stephentoub I fixed the typo and I'm working on the changes discussed for shm_open and pipe, but I'll submit them as a separate PR since they're a logically separate improvement. Was there anything else? If not, I'll squash the fixups and let CI ponder that.

stephentoub:
LGTM

nguerrera:
Making note of a conversation that is now hidden behind rebase to link this PR to a follow-up issue.

@mikedn 
> Wouldn't be better to use  intptr_t  instead of  int32_t  and thus solve the pesky issue with  SafeHandle  only working with  IntPtr  (see the  SafeFileHandle.Open  comment)?

@nguerrera 
>> Yes, I was actually going to open an issue on that. My thinking was to get everything over to the shims and then use intptr_t/IntPtr throughout.

@nguerrera 
>> #2904

PR #2906, opened 2015-08-20
Added ZLib decompression support to DeflateStream

### LINKED ISSUE #2024: DeflateStream isn't using zlib for decompression, only compression
Our implementation of DeflateStream supports using either zlib or a purely managed implementation.  However, even when choosing zlib, the managed implementation is always used for inflation, and the inflation wrapper code around zlib is never used.  We should investigate whether there are benefits to be gained from using zlib for both decompression and compression (presumably, yes).

ianhays:
The DeflateStream class currently uses ZLib for compression (if available) but doesn't use ZLib for decompression. This commit adds support to use either ZLib or the Managed implementation for decompression (inflation).

- Resolves #2024; Added ZLib support in InflaterZLib.cs
- Inflater.cs the old managed implementation is renamed to InflaterManaged.cs
- Extracted interface of necessary Inflater methods into the IInflater interface from which InflaterManaged and InflaterZLib inherit.
- Roughly a 3X speedup for decompression when using the ZLib library instead of the Managed implementation!

### DIFF
--- a/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
@@ -15,7 +15,7 @@ namespace System.IO.Compression
         private Stream _stream;
         private CompressionMode _mode;
         private bool _leaveOpen;
-        private Inflater _inflater;
+        private IInflater _inflater;
         private IDeflater _deflater;
         private byte[] _buffer;
 
@@ -38,6 +38,23 @@ namespace System.IO.Compression
         {
         }
 
+        // Since a reader is being taken, CompressionMode.Decompress is implied
+        internal DeflateStream(Stream stream, bool leaveOpen, IFileFormatReader reader)
+        {
+            Debug.Assert(reader != null, "The IFileFormatReader passed to the internal DeflateStream constructor must be non-null");
+            if (stream == null)
+                throw new ArgumentNullException("stream");
+            if (!stream.CanRead)
+                throw new ArgumentException(SR.NotReadableStream, "stream");
+
+            _inflater = CreateInflater(reader);
+            _stream = stream;
+            _mode = CompressionMode.Decompress;
+            _leaveOpen = leaveOpen;
+            _buffer = new byte[DefaultBufferSize];
+        }
+
+
         public DeflateStream(Stream stream, CompressionMode mode, bool leaveOpen)
         {
             if (stream == null)
@@ -50,7 +67,7 @@ namespace System.IO.Compression
                     {
                         throw new ArgumentException(SR.NotReadableStream, "stream");
                     }
-                    _inflater = new Inflater();
+                    _inflater = CreateInflater();
                     break;
 
                 case CompressionMode.Compress:
@@ -126,11 +143,34 @@ namespace System.IO.Compression
             }
         }
 
-        internal void SetFileFormatReader(IFileFormatReader reader)
+        private static IInflater CreateInflater(IFileFormatReader reader = null)
         {
-            if (reader != null)
+            // The deflator type (zlib or managed) is normally determined by s_deflatorType,
+            // which is initialized by the provider based on what's available on the system.
+            // But for testing purposes, we sometimes want to override this, forcing
+            // compression/decompression to use a particular type.
+            WorkerType deflatorType = s_deflaterType;
+#if DEBUG
+            if (s_forcedTestingDeflaterType != WorkerType.Unknown)
+                deflatorType = s_forcedTestingDeflaterType;
+#endif
+
+            if (deflatorType == WorkerType.ZLib)
             {
-                _inflater.SetFileFormatReader(reader);
+                // Rather than reading raw data and using a FormatReader to interpret
+                // headers/footers manually, we instead set the zlib stream to parse
+                // that information for us.
+                if (reader == null)
+                    return new InflaterZlib(ZLibNative.Deflate_DefaultWindowBits);
+                else
+                {
+                    Debug.Assert(reader.ZLibWindowSize == 47, "A GZip reader must be designated with ZLibWindowSize == 47. Other header formats aren't supported by ZLib.");
+                    return new InflaterZlib(reader.ZLibWindowSize);
+                }
+            }
+            else
+            {
+                return new InflaterManaged(reader);
             }
         }
 
@@ -255,8 +295,6 @@ namespace System.IO.Compression
                     Debug.Assert(_inflater.AvailableOutput == 0, "We should have copied all stuff out!");
                     break;
                 }
-
-                Debug.Assert(_inflater.NeedsInput(), "We can only run into this case if we are short of input");
 
                 int bytes = _stream.Read(_buffer, 0, _buffer.Length);
                 if (bytes <= 0)
@@ -421,7 +459,6 @@ namespace System.IO.Compression
             InternalWrite(array, offset, count);
         }
 
-        // isAsync always seems to be false. why do we have it?
         internal void InternalWrite(byte[] array, int offset, int count)
         {
             DoMaintenance(array, offset, count);
@@ -496,7 +533,7 @@ namespace System.IO.Compression
             if (_mode != CompressionMode.Compress)
                 return;
 
-            // Some deflaters (e.g. ZLib write more than zero bytes for zero bytes inpuits.
+            // Some deflaters (e.g. ZLib) write more than zero bytes for zero byte inputs.
             // This round-trips and we should be ok with this, but our legacy managed deflater
             // always wrote zero output for zero input and upstack code (e.g. GZipStream)
             // took dependencies on it. Thus, make sure to only "flush" when we actually had
@@ -564,10 +601,13 @@ namespace System.IO.Compression
                     {
                         if (_deflater != null)
                             _deflater.Dispose();
+                        if (_inflater != null)
+                            _inflater.Dispose();
                     }
                     finally
                     {
                         _deflater = null;
+                        _inflater = null;
                         base.Dispose(disposing);
                     }
                 }  // finally
--- a/src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
@@ -1,0 +1,242 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace System.IO.Compression
+{
+    /// <summary>
+    /// Provides a wrapper around the ZLib decompression API that implements the
+    /// IInflater interface such that this implementation functions may be used interchangeably
+    /// with the Managed implementation found in InflaterManaged.cs
+    /// </summary>
+    internal class InflaterZlib : IInflater
+    {
+        private bool _finished;                             // Whether the end of the stream has been reached
+        private bool _isDisposed;                           // Prevents multiple disposals
+        private ZLibNative.ZLibStreamHandle _zlibStream;    // The handle to the primary underlying zlib stream
+        private GCHandle _inputBufferHandle;               // The handle to the buffer that provides input to _zlibStream
+        private readonly object _syncLock = new object();   // Used to make writing to unmanaged structures atomic 
+
+        #region Exposed Members
+
+        /// <summary>
+        /// Initialized the Inflater with the given windowBits size
+        /// </summary>
+        internal InflaterZlib(int windowBits)
+        {
+            _finished = false;
+            _isDisposed = false;
+            InflateInit(windowBits);
+        }
+
+        public int AvailableOutput
+        {
+            get
+            {
+                return (int)_zlibStream.AvailOut;
+            }
+        }
+
+        /// <summary>
+        /// Returns true if the end of the stream has been reached.
+        /// </summary>
+        public bool Finished()
+        {
+            return _finished && _zlibStream.AvailIn == 0 && _zlibStream.AvailOut == 0;
+        }
+
+        public int Inflate(byte[] bytes, int offset, int length)
+        {
+            Debug.Assert(null != bytes, "Can't pass in a null output buffer!");
+
+            // If Inflate is called on an invalid or unready inflater, return 0 to indicate no bytes
+            // have been read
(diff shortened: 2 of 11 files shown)


src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
```diff
@@ -256,8 +279,6 @@ public override int Read(byte[] array, int offset, int count)
                     break;
                 }
 
-                Debug.Assert(_inflater.NeedsInput(), "We can only run into this case if we are short of input");
```
ianhays: This debug was causing exceptions because NeedsInput() was calling the handle that was at that time uninitialized. I considered adding a check around the handle, but determined the slowdown not worth the sake of this debug.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+            return _finished && _zlibStream.AvailIn == 0 && _zlibStream.AvailOut == 0;
+        }
+
+        public int Inflate(byte[] bytes, int offset, int length)
+        {
+            Debug.Assert(null != bytes, "Can't pass in a null output buffer!");
+
+            // If Inflate is called on an invalid or unready inflater, return 0 to indicate no bytes
+            // have been read
+            if (((IInflater)this).NeedsInput() || _inputBufferHandle == null || _inputBufferHandle.Value == null || length == 0)
```
ianhays: Occurs in the first Inflate call within DeflateStream.Read when no input has yet been given

src/System.IO.Compression/src/System/IO/Compression/ZLibNative.cs
```diff
@@ -403,6 +403,38 @@ public ErrorCode InflateEnd()
                 return errC;
             }
 
+            /// <summary>
+            /// This function is equivalent to inflateEnd followed by inflateInit.
+            /// The stream will keep attributes that may have been set by inflateInit2.
+            /// </summary>
+            [SecurityCritical]
+            public ErrorCode InflateReset(int windowBits)
```
ianhays: I'm not 100% on the try/finally usage within this method - could use another look for sure.

ianhays:
This commit is a work in progress on getting ZLib running as the default method of decompression on Windows and Unix systems that support it. The potential benefits are substantial; my initial perf benchmarks that measure the speed of the ```DeflateStream.ReadAsync``` (awaited) method when reading the test file ```"GZTestData/GZTestDocument.txt.gz"``` give the following measurements:

	• DeflateStream.ReadAsync Windows 1000 iterations:
		○ Windows Zlib: 42 +- 3 ms
		○ Windows Managed: 128 +- 12 ms
	• DeflateStream.ReadAsync Unix 50 iterations:
		○ Unix Zlib: 132 +- 47 ms
		○ Unix Managed: 360 +- 52 ms

To clarify, this means that my test did a ReadAsync 1000/50 times, and I ran the test n times. Across the n times, the average runtime for all iterations is given. I apologize for the convolution, but building my perf tests on top of xunit made things a bit messy :)

Regardless, a 3X DeflateStream.ReadAsync speedup is exciting!

ianhays:
One more thing, the reason I ran the tests on Unix with only 50 iterations vs 1000 on Windows is because my Unix VM is significantly slower and I didn't have the patience to run the tests that many times.

cc: @stephentoub

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+                    DeallocateInputBufferHandle();
+            }
+        }
+
+        bool IInflater.NeedsInput()
+        {
+            return _zlibStream.AvailIn == 0;
+        }
+
+        public void SetFileFormatReader(IFileFormatReader reader)
```
ianhays: Another possible perf improvement would be to add an internal, overloaded DeflateStream constructor that takes an IFileFormatReader as a parameter and then creates the Inflater with the desired window size. This would eliminate the need to call InflateInit2->InflateEnd->InflateInit2 and allow us just to use a single InflateInit2.

We would only need to modify the GZipStream constructors to allow this, since theyre the only places that set the Reader anyways. They would change from:

```
        public GZipStream(Stream stream, CompressionMode mode, bool leaveOpen)
        {
            _deflateStream = new DeflateStream(stream, mode, leaveOpen);
            SetDeflateStreamFileFormatter(mode);  // if mode == decompress, calls _deflateStream.SetFileFormatReader(new GZipDecoder))
        }
```
to:
```
        public GZipStream(Stream stream, CompressionMode mode, bool leaveOpen)
        {
            if (mode == CompressionMode.Decompress)
            {
                _deflateStream = new DeflateStream(stream, leaveOpen, new GZipDecoder());
            }
            else
            {
                _deflateStream = new DeflateStream(stream, mode, leaveOpen);
                SetDeflateStreamFileFormatter(mode);
            }
        }
```

stephentoub:
> Regardless, a 3X DeflateStream.ReadAsync speedup is exciting!

Excellent!

stephentoub:
@dotnet-bot test this please

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// Provides a wrapper around the ZLib decompression API that implements the
+    /// IInflater interface such that this implementation functions may be used interchangeably
+    /// with the Managed implementation found in InflaterManaged.cs
+    /// </summary>
+    internal class InflaterZlib : IInflater
+    {
+        private bool _finished;                             // Whether the end of the stream has been reached
+        private bool _isDisposed;                           // Prevents multiple disposals
+        private ZLibNative.ZLibStreamHandle _zlibStream;    // The handle to the primary underlying zlib stream
+        private GCHandle? _inputBufferHandle;               // The handle to the buffer that provides input to _zlibStream
```
stephentoub: This shouldn't need to be nullable.  You should be able to just use ```GCHandle``` and then test whether ```IsAllocated```.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+            finally
+            {
+                // Before returning, make sure to release input buffer if necesary:
+                if (0 == _zlibStream.AvailIn && _inputBufferHandle.HasValue)
+                    DeallocateInputBufferHandle();
+            }
+        }
+
+        bool IInflater.NeedsInput()
```
stephentoub: Why is this implemented explicitly?  If it were changed to ```public bool NeedsInput()```, you wouldn't need to cast to ```IInflater``` above when trying to call NeedsInput.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
stephentoub: Sounds good.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+
+                _zlibStream.Dispose();
+                _isDisposed = true;
+            }
+        }
+
+        void IDisposable.Dispose()
+        {
+            Dispose(true);
+        }
```
stephentoub: Is there a finalizer somewhere that will deallocate the input buffer handle if the stream is dropped without being disposed?

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Translates the given byte array to a GCHandle so that it can be passed to the ZLib
+        /// Inflate function, then returns the result of that call.
+        /// </summary>
+        private ZLibNative.ErrorCode ReadInflateOutput(byte[] outputBuffer, int offset, int length, ZLibNative.FlushCode flushCode, out int bytesRead)
+        {
+            lock (_syncLock)
+            {
+                GCHandle outputBufferHndl = GCHandle.Alloc(outputBuffer, GCHandleType.Pinned);
```
stephentoub: You should be able to use ```fixed``` instead of a pinning handle here, e.g.
```C#
fixed (byte* bufPtr = outputBuffer)
{
    _zlibStream.NextOut = bufPtr + offset;
    _zlibStream.AvailOut = (uint)length;

    ZLibNative.ErrorCode errC = Inflate(flushCode);
    bytesRead = length - (int)_zlibStream.AvailOut;

    return errC;
}
```

src/System.IO.Compression/src/System/IO/Compression/ZLibNative.cs
stephentoub: This pattern of putting code in a finally block like this is done to deal with thread aborts, in particular because thread aborts are delayed across finally's so you don't have to worry about them interrupting a finally block.  However, thread aborts aren't a concern for .NET Core, and even if this library ends up being used on the desktop, it's very unlikely the library is entirely hardened against them.  I wouldn't worry about using try/finally's in this way.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+                int bytesRead;
+                ZLibNative.ErrorCode errc = ReadInflateOutput(bytes, offset, length, ZLibNative.FlushCode.NoFlush, out bytesRead);
+                if (errc == ZLibNative.ErrorCode.StreamEnd)
+                    _finished = true;
+                return bytesRead;
+            }
+            finally
+            {
+                // Before returning, make sure to release input buffer if necesary:
+                if (0 == _zlibStream.AvailIn && _inputBufferHandle.HasValue)
```
stephentoub: So after ReadInflateOutput has returned, it's copied everything it needs from the input buffer and no longer expects to have it available?

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        public void SetFileFormatReader(IFileFormatReader reader)
+        {
+            // Rather than reading raw data and using a FormatReader to interpret
+            // headers/footers manually, we instead reset the zlib stream to parse
+            // that information for us.
+            ZLibNative.ErrorCode error;
+            try
+            {
+                error = _zlibStream.InflateReset(reader.ZLibWindowSize);
```
stephentoub: What if reader isn't for GZip but for something else?  This implementation only works if it is GZip, right?  If so, that's fine, but we should probably add an assert that the type of the reader is what we expect.

stephentoub:
cc: @macohn, @ericstj

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
stephentoub: Also, from looking at the code coverage information, it looks like Dispose is never being called.  Is that just a problem with the tests, that they're not cleaning up appropriately?  Or is there a disconnect somewhere such that disposal of the stream isn't disposing the inflater?

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
ianhays: A few questions:

- Fixed requires that the block or the method be marked as ```unsafe```; is this fine and if so, which context would be preferred to be marked as such?
- There is currently a lock around the code to make writing to unmanaged structures atomic; does ```fixed``` preclude the necessity of the lock, or do they accomplish unrelated things? 
- Note: The logic for the lock was taken from DeflaterZLib, so I'll be mirroring changes from your feedback to that file as well.

stephentoub: > is this fine and if so, which context would be preferred to be marked as such?

That's fine.  I think it'd be fine just to add unsafe to the method signature, since effectively the whole method needs to be able to use pointers.

> does fixed preclude the necessity of the lock, or do they accomplish unrelated things?

They are orthogonal, can be used together, and are both needed.  The lock ensures that only one thread at a time can access this region.  The fixed ensures that for the duration of the fixed block, the buffer is pinned, and it gives you a pointer to it.

> so I'll be mirroring changes from your feedback to that file as well.

Sounds good. Thanks.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
ianhays: Yes. ReadInflateOutput will inflate all input and place it in the output buffer such that AvailIn == 0 afterward ReadInflateOutput returns.

Subsequent calls to SetInput() will create a new _inputBufferHandle with new input data.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
ianhays: Yes, if Zlib doesn't support reading the header then subsequent inflation calls will fail due to a Z_DATA_ERROR. This method actually sets the WindowBits size which happens to also enable setting of header/footer parsing, so its a bit of a misnomer (although still does what the method promises in the case where reader if for GZip).

I wrote it such that if ZLib supports the header with the given window size as it's designation, then we do as well. Since GZip is the only supported option for this though, I will add a Debug.Assert for it.

src/System.IO.Compression/src/System/IO/Compression/InflaterZlib.cs
ianhays: There is not a finalizer. I will add one and split the code up like so:
```
if (!_isDisposed)
{
    if (disposing)
    {
        _zlibStream.Dispose();
    }
    if (_inputBufferHandle.IsAllocated)
        DeallocateInputBufferHandle();

    _isDisposed = true;
}
```

DeflateStream disposal is not calling the InflateStream's dispose method; it only calls disposal on the _deflater. I'll add some disposals for the inflater in the same place the deflater is being disposed.

ianhays: Commit with changes for feedback pushed. 

This comment/diff about Dispose is now outdated but is not showing so on Github.

stephentoub:
A few more comments, but otherwise LGTM.  Nice job, Ian.

PR #2909, opened 2015-08-21
Fixed a loss of streamed bytes issue.

shmao:
When reading stream content with `ReadContentAsBase64`, if the buffer size (count) is less than 3, XmlBaseReaders may lost bytes. The cause is that `XmlBaseReader.MoveToContent` may skip bytes in trailing bytes buffer, `_trailBytes`, which is used only when buffer size is less than 3.

The fix is to let `XmlBaseReader.MoveToContent` check if there are any bytes in trailing buffer.

### DIFF
--- a/src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseReader.cs
+++ b/src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseReader.cs
@@ -832,6 +832,12 @@ namespace System.Xml
                 {
                     if (_node.NodeType != XmlNodeType.Text && _node.NodeType != XmlNodeType.CDATA)
                         break;
+
+                    if (_trailByteCount > 0)
+                    {
+                        break;
+                    }
+
                     if (_value == null)
                     {
                         if (!_node.Value.IsWhitespace())


shmao:
Fixes dotnet/wcf#267

shmao:
@SGuyGe @zhenlan @iamjasonp @KKhurin

nguerrera:
CI failed due to issue being addressed by #2911

zhenlan:
LGTM

nguerrera:
@dotnet-bot test this please

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseReader.cs
```diff
@@ -832,6 +832,12 @@ public override XmlNodeType MoveToContent()
                 {
                     if (_node.NodeType != XmlNodeType.Text && _node.NodeType != XmlNodeType.CDATA)
                         break;
+
+                    if (_trailByteCount > 0)
+                    {
+                        break;
+                    }
+
```
iamjasonp: This code already exists in the desktop code - any idea how this chunk of code disappeared from the CoreFX implementation? I want to make sure we haven't accidentally backed out a fix for something else, and Git blame isn't useful here as it was part of the initial commit. 

http://referencesource.microsoft.com/#System.Runtime.Serialization/System/Xml/XmlBaseReader.cs,858

iamjasonp:
LGTM.

It'd be good to get some clarity on how this chunk of code went missing between desktop -> CoreFx.

src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseReader.cs
jasonwilliams200OK: Maybe @khdang can give any insights because this is going to be the second change in this file since the initial commit by @khdang. See history: https://github.com/dotnet/corefx/commits/master/src/System.Private.DataContractSerialization/src/System/Xml/XmlBaseReader.cs.

shmao: I had the same concern. I checked the history of the desktop.  It seems we have two code bases for serialization. The CoreFx code was copied from NDP\FX\src\CDF, which doesn't have this fix. The other one is under NDP\CDF\src, and it has the fix. Do you know the relationships of these two code base? @iamjasonp

iamjasonp: The ndp\cdf\src codebase is the correct one. ndp\fx\src\cdf code was a fork to support some other internal projects at the time.

zhenlan: The ndp\fx\src\cdf is the SL code base. It is also what current GH code started with. I suspect the code change discussed here was added to the desktop framework code base after the SL fork.

PR #2911, opened 2015-08-21
Fix build failure due to conflict between concurrent PRs

nguerrera:
New usage of Interop.libc.close was added while it was
simultaneously being removed in another PR, which defeated
catching it via CI PR validation.

cc @stephentoub @CIPop

### DIFF
--- a/src/System.Net.NameResolution/src/System.Net.NameResolution.csproj
+++ b/src/System.Net.NameResolution/src/System.Net.NameResolution.csproj
@@ -155,8 +155,8 @@
     <Compile Include="$(CommonPath)\Interop\Unix\Interop.Libraries.cs">
       <Link>Interop\Unix\Interop.Libraries.cs</Link>
     </Compile>
-    <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.close.cs">
-      <Link>Interop\Unix\libc\Interop.close.cs</Link>
+    <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.Close.cs">
+      <Link>Interop\Unix\System.Native\Interop.Close.cs</Link>
     </Compile>
     <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.freeaddrinfo.cs">
       <Link>Interop\Unix\libc\Interop.freeaddrinfo.cs</Link>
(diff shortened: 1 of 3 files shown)


src/System.Net.NameResolution/tests/PalTests/System.Net.NameResolution.Pal.Tests.msbuild
```diff
@@ -133,9 +133,6 @@
     <Compile Include="$(CommonPath)\Interop\Unix\Interop.Libraries.cs">
       <Link>Interop\Unix\Interop.Libraries.cs</Link>
     </Compile>
-    <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.close.cs">
-      <Link>Interop\Unix\libc\Interop.close.cs</Link>
-    </Compile>
```
stephentoub: This just wasn't necessary?

stephentoub:
LGTM

CIPop:
@pgavlin PTAL.

LGTM

src/System.Net.NameResolution/tests/PalTests/System.Net.NameResolution.Pal.Tests.msbuild
nguerrera: I couldn't find any usage of it in the projects that included it.

nguerrera: And both OSX and Linux build succeed without it.

src/System.Net.NameResolution/src/System.Net.NameResolution.csproj
```diff
@@ -155,8 +155,8 @@
     <Compile Include="$(CommonPath)\Interop\Unix\Interop.Libraries.cs">
       <Link>Interop\Unix\Interop.Libraries.cs</Link>
     </Compile>
-    <Compile Include="$(CommonPath)\Interop\Unix\libc\Interop.close.cs">
-      <Link>Interop\Unix\libc\Interop.close.cs</Link>
+    <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.Close.cs">
+      <Link>Interop\Unix\libc\Interop.Close.cs</Link>
```
nguerrera: I left libc in the link path. I'm not going to bother resetting CI since all PRs are blocked on this right now. 

Most projects don't have the library name in the link path, but some do. Let's decide which way we want to standardize and I'll fix up the whole lot to match. cc @stephentoub

nguerrera: OK, I made another mistake so I fixed this too to at least be consistent with the same file, but I would like to standardize all of the link paths.

src/System.Net.NameResolution/tests/PalTests/System.Net.NameResolution.Pal.Tests.msbuild
nguerrera: ... but the windows build fails and it's needed for the same source line that changed being pulled into other projects. Fixed. :blush:

PR #2915, opened 2015-08-21
HttpClient xplat: Fix bugs in polling and callback logic

vijaykota:
- On OSX, the read end of the pipe sometimes wakes up the poll
  with a POLLOUT. Since we are interested in only reading on it,
  scoping down to POLLIN prevents this from happening. Else we
  attempt a read on it which fails

- CheckForCompletedTransfers was being unnecessarily called from the
  socket callback. We are only interested in MSG_DONE after the socket
  callback indicates removal. This will anyway wake up PollFunction which
  calls CheckForCompletedTransfers. This race is also causing a seg fault
  sometimes because curl_multi_socket_action has been called on an fd
  that was getting cleaned up via curl_easy_cleanup (call sequence is
  CurlSocketCallback --> CheckForCompletedTransfers --> EndRequest)

- Also removed some unnecessary CharSet attributes

### DIFF
--- a/src/Common/src/Interop/Unix/libcurl/Interop.libcurl.cs
+++ b/src/Common/src/Interop/Unix/libcurl/Interop.libcurl.cs
@@ -25,13 +25,13 @@ internal static partial class Interop
         public static extern void curl_multi_cleanup(
             IntPtr handle);
 
-        [DllImport(Interop.Libraries.LibCurl, CharSet = CharSet.Ansi)]
+        [DllImport(Interop.Libraries.LibCurl)]
         public static extern int curl_multi_setopt(
             SafeCurlMultiHandle multi_handle,
             int option,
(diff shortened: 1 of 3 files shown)


src/Common/src/Interop/Unix/libcurl/Interop.libcurl.cs
```diff
@@ -155,5 +155,5 @@ internal static partial class libcurl
 
         [DllImport(Interop.Libraries.LibCurl)]
         public static extern IntPtr curl_version_info(int curlVersionStamp);
-    }
+ }
```
stephentoub: Nit: spacing got changed here oddly

stephentoub:
LGTM.  I had one very minor nit on spacing, but I'm going to merge this to hopefully unblock usage.

PR #2916, opened 2015-08-21
Introducing a bunch of tests for http client handler

kapilash:
This checkin tests the client handler in following scenarios:
* Http and Https calls
* Authentication
* Redirection
* Redirection + Authentication
* custom headers
* cookies

### DIFF
--- a/src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
+++ b/src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
@@ -11,5 +11,6 @@ internal static partial class Interop
         internal const string INODE64SUFFIX = "";
 
         internal const string LibRt = "librt";  // POSIX Realtime Extensions library
+        internal const string LibCurl = "libcurl";             // Curl HTTP client library	
     }
 }
--- a/src/Common/src/Interop/Linux/Interop.Libraries.cs
+++ b/src/Common/src/Interop/Linux/Interop.Libraries.cs
@@ -11,5 +11,6 @@ internal static partial class Interop
         internal const string INODE64SUFFIX = "";
 
         internal const string LibRt = "librt"; // POSIX Realtime Extensions library
+        internal const string LibCurl = "libcurl.so.4";             // Curl HTTP client library
     }
 }
--- a/src/Common/src/Interop/OSX/Interop.Libraries.cs
+++ b/src/Common/src/Interop/OSX/Interop.Libraries.cs
@@ -18,5 +18,6 @@ internal static partial class Interop
         internal const string CoreServicesLibrary   = "/System/Library/Frameworks/CoreServices.framework/CoreServices";
         internal const string libproc = "libproc";
         internal const string LibSystemKernel = "/usr/lib/system/libsystem_kernel";
+        internal const string LibCurl = "libcurl";             // Curl HTTP client library
     }
 }
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
@@ -63,7 +63,7 @@ namespace System.Net.Http
         private bool _preAuthenticate = false;
         private CredentialCache _credentialCache = null;
         private CookieContainer _cookieContainer = null;
-        private bool _useCookie = false;
+        private bool _useCookie = true;
         private bool _automaticRedirection = HttpHandlerDefaults.DefaultAutomaticRedirection;
         private int _maxAutomaticRedirections = HttpHandlerDefaults.DefaultMaxAutomaticRedirections;
 
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
@@ -71,11 +71,6 @@ namespace System.Net.Http
 
         static CurlHandler()
         {
-            int result = Interop.libcurl.curl_global_init(Interop.libcurl.CurlGlobalFlags.CURL_GLOBAL_ALL);
-            if (result != CURLcode.CURLE_OK)
-            {
-                throw new InvalidOperationException("Cannot use libcurl in this process");
-            }
             curlVersionInfoData = Marshal.PtrToStructure<CurlVersionInfoData>(Interop.libcurl.curl_version_info(CurlAge));
             if (curlVersionInfoData.age < MinCurlAge)
             {
@@ -581,13 +576,9 @@ namespace System.Net.Http
 
         private void SetCookieOption(SafeCurlHandle requestHandle, Uri requestUri)
         {
-            if (!_useCookie)
+            if (!_useCookie || _cookieContainer == null)
             {
                 return;
-            }
-            else if (_cookieContainer == null)
-            {
-                throw new InvalidOperationException(SR.net_http_invalid_cookiecontainer);
             }
 
             string cookieValues = _cookieContainer.GetCookieHeader(requestUri);                    
--- a/src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
+++ b/src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
@@ -163,6 +163,7 @@ namespace System.Net.Http.Tests
             }
         }
 
+        /* Commenting out below test till the https://github.com/dotnet/corefx/issues/3015 is fixed.
         [Theory]
         [InlineData(6)]
         public async void GetAsync_MaxAutomaticRedirectionsNServerHopsNPlus1_Throw(int hops)
@@ -174,6 +175,7 @@ namespace System.Net.Http.Tests
                 await Assert.ThrowsAsync<HttpRequestException>(() => client.GetAsync(HttpTestServers.RedirectUriHops(hops + 1)));
             }
         }
+        */
 
         [Fact]
         public async void GetAsync_CredentialIsNetworkCredentialUriRedirect_StatusCodeUnauthorized()
--- a/src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
+++ b/src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
@@ -16,6 +16,27 @@ namespace System.Net.Http.Tests
     public class HttpClientHandlerTest
     {
         readonly ITestOutputHelper _output;
+        private const string Username = "testuser";
+        private const string Password = "password";
+        private const string DecompressedContentPart = "Accept-Encoding";
+        private readonly NetworkCredential _credential = new NetworkCredential(Username, Password);
+
+        private static bool JsonMessageContainsKeyValue(string message, string key, string value)
+        {
+            // TODO: Align with the rest of tests w.r.t response parsing once the test server is finalized.
+            // Currently not adding any new dependencies
+            var pattern = string.Format(@"""{0}"": ""{1}""", key, value);
+            return message.Contains(pattern);
+        }
+
+        private static async Task AssertSuccessfulGetResponse(HttpResponseMessage response, Uri uri, ITestOutputHelper output)
+        {
+            Assert.Equal<HttpStatusCode>(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal<string>("OK", response.ReasonPhrase);
+            string responseContent = await response.Content.ReadAsStringAsync();
+            Assert.True(JsonMessageContainsKeyValue(responseContent, "url", uri.AbsoluteUri));
+            output.WriteLine(responseContent);
+        }
 
         public HttpClientHandlerTest(ITestOutputHelper output)
         {
@@ -23,16 +44,43 @@ namespace System.Net.Http.Tests
         }
         
         [Fact]
-        public void SendAsync_SimpleGet_Success()
+        public async void SendAsync_SimpleGet_Success()
         {
-            var handler = new HttpClientHandler();
-            var client = new HttpClient(handler);
-            
-            // TODO: This is a placeholder until GitHub Issue #2383 gets resolved.
-            var response = client.GetAsync(HttpTestServers.RemoteGetServer).Result;
-            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            var responseContent = response.Content.ReadAsStringAsync().Result;
-            _output.WriteLine(responseContent);
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    // TODO: This is a placeholder until GitHub Issue #2383 gets resolved.
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.RemoteGetServer);
+                    await AssertSuccessfulGetResponse(response, HttpTestServers.RemoteGetServer, _output);
+                }
+            }
+        }
+
+        [Fact]
+        public async void SendAsync_SimpleHttpsGet_Success()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+                    await AssertSuccessfulGetResponse(response, HttpTestServers.SecureRemoteGetServer, _output);
+                }
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_ResponseContentAfterClientAndHandlerDispose_Success()
+        {
+            HttpResponseMessage response = null;
+            using (var handler = new HttpClientHandler())
+            using (var client = new HttpClient(handler))
+            {
+                response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+            }
+            Assert.NotNull(response);
+            await AssertSuccessfulGetResponse(response, HttpTestServers.SecureRemoteGetServer, _output);
         }
 
         [Fact]
@@ -55,6 +103,152 @@ namespace System.Net.Http.Tests
 
                 Assert.True(ex.CancellationToken.IsCancellationRequested,
                     "Expected cancellation requested on token attached to exception.");
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_DefaultAutomaticDecompression_ContentDecompressed()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.RemoteServerGzipUri);
+                    string responseContent = await response.Content.ReadAsStringAsync();
+                    Assert.True(responseContent.Contains(DecompressedContentPart));
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_SetCredential_StatusCodeOK()
+        {
+            Uri uri = HttpTestServers.BasicAuthUriForCreds(Username, Password);
+            using (var handler = new HttpClientHandler())
+            {
+                handler.Credentials = _credential;
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(uri);
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_AllowRedirectFalse_StatusCodeRedirect()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                handler.AllowAutoRedirect = false;
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.RemoteServerRedirectUri);
+                    Assert.Equal(HttpStatusCode.Redirect, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_AllowRedirectDefault_StatusCodeOK()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.RemoteServerRedirectUri);
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData(6)]
+        public async void GetAsync_MaxAutomaticRedirectionsNServerHopsNPlus1_Throw(int hops)
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                handler.MaxAutomaticRedirections = hops;
+                var client = new HttpClient(handler);
+                await Assert.ThrowsAsync<HttpRequestException>(() => client.GetAsync(HttpTestServers.RedirectUriHops(hops + 1)));
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_CredentialIsNetworkCredentialUriRedirect_StatusCodeUnauthorized()
+        {
+            Uri redirectUri = HttpTestServers.RedirectUriForCreds(Username, Password);
+            using (var handler = new HttpClientHandler())
+            {
+                handler.Credentials = _credential;
+                var client = new HttpClient(handler);
+                HttpResponseMessage unAuthResponse = await client.GetAsync(redirectUri);
+                Assert.Equal(HttpStatusCode.Unauthorized, unAuthResponse.StatusCode);
+            }
+       }
+
+        [Fact]
+        public void GetAsync_CredentialIsCredentialCacheUriRedirect_HttpStatusCodeOK()
+        {
+            Uri uri = HttpTestServers.BasicAuthUriForCreds(Username, Password);
+            Uri redirectUri = HttpTestServers.RedirectUriForCreds(Username, Password);
+            var credentialCache = new CredentialCache();
+            credentialCache.Add(uri, "Basic", _credential);
+
+            using (var handler = new HttpClientHandler())
+            {
+                handler.Credentials = credentialCache;
+                var client = new HttpClient(handler);
+                var response = client.GetAsync(uri).Result;
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            }
+        }
+
+        [Theory]
+        [InlineData("cookiename", "cookievalue")]
+        public async void GetAsync_SetCookieContainer_CookieSent(string name, string value)
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                var cookieContainer = new CookieContainer();
+                cookieContainer.Add(HttpTestServers.RemoteServerCookieUri, new Cookie(name, value));
+                handler.CookieContainer = cookieContainer;
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage httpResponse = await client.GetAsync(HttpTestServers.RemoteServerCookieUri);
+                    Assert.Equal(httpResponse.StatusCode, HttpStatusCode.OK);
+                    string responseText = await httpResponse.Content.ReadAsStringAsync();
+                    Assert.True(JsonMessageContainsKeyValue(responseText, name, value));
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData("X-Cust-Header","x-value")]
+        public async void GetAsync_RequestHeadersAddCustomHeaders_HeaderAndValueSent(string name, string value)
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    client.DefaultRequestHeaders.Add(name, value);
+                    HttpResponseMessage httpResponse = await client.GetAsync(HttpTestServers.RemoteServerHeadersUri);
+                    httpResponse.EnsureSuccessStatusCode();
+                    string responseText = await httpResponse.Content.ReadAsStringAsync();
+                    Assert.True(JsonMessageContainsKeyValue(responseText, name, value));
+                }
+            }
+        }
+
+        [Fact]
+        public async void SendAsync_HttpRequestMsgResponseHeadersRead_StatusCodeOK()
+        {
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, HttpTestServers.SecureRemoteGetServer);
+            using (var client = new HttpClient())
+            {
+                HttpResponseMessage response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead);
+                await AssertSuccessfulGetResponse(response, HttpTestServers.SecureRemoteGetServer, _output);
             }
         }
     }
--- a/src/Common/src/Interop/Unix/Interop.Libraries.cs
+++ b/src/Common/src/Interop/Unix/Interop.Libraries.cs
@@ -8,7 +8,6 @@ internal static partial class Interop
         internal const string Libc = "libc";                   // C library
         internal const string LibCoreClr= "libcoreclr";        // CoreCLR runtime
         internal const string LibCrypto = "libcrypto";         // OpenSSL crypto library
-        internal const string LibCurl = "libcurl";             // Curl HTTP client library
         internal const string Zlib = "libz";                   // zlib compression library
 
         // Shims
--- a/src/Common/tests/System/Net/HttpTestServers.cs
+++ b/src/Common/tests/System/Net/HttpTestServers.cs
@@ -17,11 +17,30 @@ namespace System.Net.Tests
         public readonly static Uri RemotePostServer = new Uri("http://" + Host + "/post");
         public readonly static Uri SecureRemoteGetServer = new Uri("https://" + Host + "/get");
         public readonly static Uri SecureRemotePostServer = new Uri("https://" + Host + "/post");
+        public readonly static Uri RemoteServerGzipUri = new Uri("http://" + Host + "/gzip");
+        public readonly static Uri RemoteServerRedirectUri = new Uri("http://" + Host + "/relative-redirect/6");
+        public readonly static Uri RemoteServerCookieUri = new Uri("http://" + Host + "/cookies");
+        public readonly static Uri RemoteServerHeadersUri = new Uri("http://" + Host + "/headers");
 
         public const string RemoteStatusCodeServerFormat = "http://" + Host + "/status/{0}";
         public const string SecureRemoteStatusCodeServerFormat = "https://" + Host + "/status/{0}";
 
         public readonly static object[][] GetServers = { new object[] { RemoteGetServer }, new object[] { SecureRemoteGetServer } };
         public readonly static object[][] PostServers = { new object[] { RemotePostServer }, new object[] { SecureRemotePostServer } };
+
+        public static Uri BasicAuthUriForCreds(string userName, string password)
+        {
+            return new Uri("http://" + Host + "/basic-auth/" + userName + "/" + password);
+        }
+
+        public static Uri RedirectUriForCreds(string userName, string password)
+        {
+            return new Uri("http://" + Host + "/redirect-to?url=http://" + Host + "/basic-auth/" + userName + "/" + password);
+        }
+
+        public static Uri RedirectUriHops(int hops)
+        {
+            return new Uri("http://" + Host + "/relative-redirect/" + hops);
+        }
     }
 }
--- a/src/System.Net.Http/src/System.Net.Http.csproj
+++ b/src/System.Net.Http/src/System.Net.Http.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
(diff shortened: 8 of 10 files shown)


vijaykota:

stephentoub:
@kapilash, @vijaykota, all of the tests are failing in CI because libcurl isn't installed.  What package should @mmitche install into the CI servers? libcurl3? libcurl3-dev?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -36,6 +47,21 @@ public void SendAsync_SimpleGet_Success()
...
+            var handler = new HttpClientHandler();
+            var client = new HttpClient(handler);
+
+            var response = client.GetAsync(HttpTestServers.SecureRemoteGetServer).Result;
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            var responseContent = response.Content.ReadAsStringAsync().Result;
+            Assert.Equal<string>(response.ReasonPhrase, "OK");
+            _output.WriteLine(responseContent);
+            handler.Dispose();
+            client.Dispose();
```
stephentoub: Since the expected use case is to await the async operations rather than block synchronously for them, I'd suggest these tests be written as "async Task" rather than "void", e.g.
```
[Fact]
public async Task SendAsync_SimpleHttpsGet_Success()
{
    {
        HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
        Assert.Equal("OK", response.ReasonPhrase);

        string responseContent = await response.Content.ReadAsStringAsync();
        _output.WriteLine(responseContent);
    }
}
```
A few other things to note there:
- The expected value should come first in Assert.Equal; you did that for the first one, but the second one had its arguments reverse
- I've used a using to cleanup deterministically so that if an assert does fail, we don't rely on the finalizer to clean up and potentially impact other tests more than necessary
- I've avoided the use of var except in cases where the type is easily clear from the right-hand side, e.g. we try to only use var in the cast of constructing an object, a cast, etc.

Separately, is there anything we can verify about responseContent?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +83,142 @@ public void SendAsync_SimpleGet_Success()
                     "Expected cancellation requested on token attached to exception.");
             }
         }
+
+        [Fact]
+        public void GetAsyncAutomaticDecompression()
+        {
+            var handler = new HttpClientHandler();
+            if (!handler.SupportsAutomaticDecompression)
```
stephentoub: In what situations won't the handler support automatic decompression?  Should we instead be asserting that it does support it?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +83,142 @@ public void SendAsync_SimpleGet_Success()
...
+            var handler = new HttpClientHandler();
+            var client = new HttpClient(handler);
+            var response = client.GetAsync(HttpTestServers.RemoteServerRedirectUri).Result;
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            handler.Dispose();
+            client.Dispose();
+        }
+
+        [Theory]
+        [InlineData(6)]
```
stephentoub: Are we planning to add additional data inputs here?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +83,142 @@ public void SendAsync_SimpleGet_Success()
...
+        public void UriMaxRedirectionAllow(int hops)
+        {
+            var handler = new HttpClientHandler();
+            handler.MaxAutomaticRedirections = hops;
+            var client = new HttpClient(handler);
+            Assert.Throws<System.AggregateException>(() =>
+                { var _ = client.GetAsync(HttpTestServers.RedirectUriHops(hops + 1)).Result; }
```
stephentoub: By using await as I suggested earlier, you can change this Throws to catch the actual exception type that's getting thrown (before it gets wrapped in an aggregate).

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +83,142 @@ public void SendAsync_SimpleGet_Success()
...
+            var handler = new HttpClientHandler();
+            var client = new HttpClient(handler);
+            client.DefaultRequestHeaders.Add(name, value);
+            var httpResponse = client.GetAsync(HttpTestServers.RemoteServerHeadersUri).Result;
+            Assert.Equal(httpResponse.StatusCode, HttpStatusCode.OK);
+            var response = httpResponse.Content.ReadAsStringAsync().Result;
+            Assert.True(JsonMessageContainsKeyValue(response, name, value));
+            handler.Dispose();
+            client.Dispose();
+        }
```
stephentoub: Since the primary initial consumer of this component will be NuGet package restore, is there a test we can add that uses HttpClient in the same way it does, similar combinations of features, etc.?  I think it's this: https://github.com/aspnet/dnx/blob/22dbb542f72158f20d0af7eba8abd4bea5a14d22/src/Microsoft.Dnx.Tooling/Restore/NuGet/HttpSource.cs

davidsh:
These tests are all violating our naming convention for tests.  We name tests based on a 3-part naming scheme 

See 'Art of Unit Testing' http://www.amazon.com/Art-Unit-Testing-examples/dp/1617290890/ref=sr_1_1?s=books&ie=UTF8&qid=1440161774&sr=1-1&keywords=art+of+unit+testing

I thought we were going to review design proposals first before getting a GitHub PR.

You should review the existing ToF tests and conform to the naming, style, and design conventions of those.

kapilash:
@stephentoub 
> all of the tests are failing in CI because libcurl isn't installed. What package should @mmitche install into the CI servers? libcurl3? libcurl3-dev?

we need to install the libcurl3 (or a greater version).

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
davidsh: For CoreFx the handler should ALWAYS support automatic decompression even if it has to do it itself rather than rely on the underlying stack to do it.

For example, for Windows native WinHTTP supports it but only on Windows 8 and beyond.  Since CoreFx supports Windows 7, we have "manual" code to do the automatic decompression in the WinHttpHandler.

So, the design principle for CoreFx HttpClientHandler is that it should always return TRUE for supporting this feature. For implementation, if libcurl supports it, then it is easy.  Otherwise additional code needs to be added to the Curlhandler to support it.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +83,142 @@ public void SendAsync_SimpleGet_Success()
...
+            var handler = new HttpClientHandler();
+            handler.Credentials = _credential;
+            var client = new HttpClient(handler);
+            var response = client.GetAsync(uri).Result;
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(new Version(1,1), response.Version);
+            handler.Dispose();
+            client.Dispose();
+        }
+
```
davidsh: As further comments on this test design, there are too many things being tested in this one 'test'.

This test, for example, should be split up into at least 2 tests with names something like this:

GetAsync_UseNoAuthentication_Unauthorized()
GetAsync_UseBasicAuthentication_OK()

etc.

In general, tests should conform to the "Arrange, Act, Assert" principle.

kapilash:
@davidsh 

Thanks for confirmation about naming conventions for tests. I did check the [coding conventions](https://github.com/dotnet/corefx/tree/master/Documentation/coding-guidelines) and did not find any reference for tests and quick check on github gave both types of names for tests so wasn't really sure.


> I thought we were going to review design proposals first before getting a GitHub PR.

What design proposal are you referring to?  Do you want me to close this PR here and send out a separate CR?  I thought we decided exactly otherwise. 

Thanks for info about unit testing and the art there of, although, these qualify for "functional tests".

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
davidsh: +1

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: libcurl supports it. I was only worried about windows version on our CI tests. I thought this should not break a checkin. I will change this to an assert in the next iteration.

davidsh:
@vijaykota @SidharthNabar @CIPop @pgavlin 

It doesn't matter that these are "functional tests" and not "unit tests".  We use the same design principles to guide us especially on design, naming, and not testing too much stuff in a single test method.

The coding conventions link you sent is not the entire set of guidelines.  It is only a base set of guidelines for CoreFx.

Vijay should have discussed with you the procedure we were going to use for enhancing these Http tests.

We don't need a separate CR (CodeFlow).  If you want to iterate on this PR and get it updated to conform that is fine.  Otherwise, we are simply adding technical debt and these tests are going to have to be restructured to conform.  We are in the process of moving 100's of tests to GitHub and these 'new' tests are not aligned with our current collateral.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: Sure. Will fix this and related tests - I was just following the existing test.

kapilash:
> If you want to iterate on this PR and get it updated to conform that is fine. Otherwise, we are simply adding technical debt and these tests are going to have to be restructured to conform

Yes. Whichever is comfortable for you. I will send out the PR with naming conventions.

I was bringing about the distinction  with "functional tests" only w.r.t multiple asserts in a single test. I will definitely rewrite the tests as per the "Arrange, Act, Assert" principle that you described in other comment.

stephentoub:
> w.r.t multiple asserts in a single test

FWIW, I don't think there's anything wrong with multiple asserts in a single test.  I agree a test should focus on a single use case, but it's often the case that you set something up, do something, and then want to verify multiple things about it.  For example, if we were testing a collection's Add method, I think it's perfectly reasonable to have multiple asserts like:
```C#
SomeCollection c = new SomeCollection();
var item = new Item();
c.Add(item);
Assert.Equal(0, c.Count);
Assert.Equal(item, c[0]);
```
I'd much prefer that over one test doing:
```C#
SomeCollection c = new SomeCollection();
var item = new Item();
c.Add(item);
Assert.Equal(0, c.Count);
```
and another doing:
SomeCollection c = new SomeCollection();
var item = new Item();
c.Add(item);
Assert.Equal(item, c[0]);
```

**Comment by davidsh** [comment]
Sometimes it is unavoidable to have multiple asserts in a single test method.  We are not dogmatic about getting it perfect 100% of the time. But it is one of our guiding principles to help structure and manage a lot of tests.

**Comment by stephentoub** [comment]
@dotnet-bot test this please (@mmitche installed libcurl on the CI machines)

File: src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs (hunk 2 above)
Comment (davidsh) [inline src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs]: If you look at our ToF tests for the Http component, you will see that we test (validate) in the HttpClientHandler Constructor test that the property always return TRUE.

**Comment by stephentoub** [comment]
@davidsh, you marked this as "Needs extra testing before merge".  Beyond the changes discussed on this thread, what else is required?  Whatever it is, I'd like to see this get prioritized, so that we have at least some basic coverage in CI asap for the CurlHandler as it's being brought up. Thanks!

**Comment by davidsh** [comment]
I understand the urgency of getting this component stable and validating sufficiently in CI.  I would like to see sufficient changes in the PR as discussed in my comments on this thread.

I marked it "extra testing" as such because these tests are testing *all* platforms and I want to make sure that the assumptions in the tests are matching the ToF tests.  In fact, it would be more productive for @kapilash  to look at those tests and compare assumptions to make sure that we are testing the right assumptions for the design. As you know, just because a test passes doesn't mean it is testing the right design. So, comparing with the ToF tests (i.e. validate default property values, etc.) is an important quality step.

**Comment by stephentoub** [comment]
Thanks.  If that's going to take any non-trivial amount of time, I'd be fine seeing these tests marked as PlatformSpecific for now (or ActiveIssue with PlatformID.Windows due to them not having been sufficiently vetted), or moved into a test project dedicated to Unix and not run on Windows.  I of course want to make sure the right tests are in place longer term, and in fact I expect that the tests you're porting over will cover everything these are testing and much more, so my hope for these would be that they're a stop-gap in the interim that helps give us some minimum level of confidence as the component is enabled.  If we end up throwing them away later, I have no qualms with that.  If they prove useful to keep, I have no qualms with that, either.

**Comment by kapilash** [comment]
Actually I did look at all the tests from ToF. I [migrated them to Xunit](https://github.com/kapilash/corefx/tree/master/src/System.Net.Http/tests/Extra) too.  In this PR, we are focussed on checking things that ensure that curlHandler gets invoked and that it plays well with the rest of .Net code. 
If you prefer, I can move these to out a separate project. 

And I will definitely send out a new iteration incorporating all the comments from @stephentoub  and @davidsh. 
I am only explaining the purpose of these particular set of tests.

**Comment by davidsh** [comment]
Don't create a separate test project.  Just continue to get the tests conforming better so that we don't add too much more technical debt.  

@kapilash Since you indicate that you have looked thru the ToF tests, then I will remove the "extra testing" marker.  Please make sure *all* of your tests run on all platforms and pass.

File: src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs [hunk 7]
```diff
@@ -16,7 +16,18 @@ namespace System.Net.Http.Tests
     public class HttpClientHandlerTest
     {
         readonly ITestOutputHelper _output;
+        private const string Username = "testuser";
```
CIPop: These are not respecting System.Net's test requirements: all constants regarding servers must be separated from the tests.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -16,7 +16,18 @@ namespace System.Net.Http.Tests
...
+        private const string Password = "password";
+        private const string DecompressedContentPart = "Accept-Encoding";
+        private readonly NetworkCredential _credential = new NetworkCredential(Username, Password);
 
+        private static bool JsonMessageContainsKeyValue(string message, string key, string value)
+        {
+            // Poor Man's json parsing.
```
CIPop: Sounds like a TODO. Please add a tracking issue.

CIPop: Actually, I don't really understand why a HTTP test is taking a dependency on Json. (I'm missing context given that the PR is missing the server-side code.)

davidsh: it's probably because the current test server is the httpbin.org server which sends responses back in JSON format.

That's not going to be the test server long term. But it's possible that the test server might use JSON responses (although our current ToF test server don't).

stephentoub:
@dotnet-bot test this please

kapilash:
I understand libcurl was installed on this CI machine.
But somehow the error is 
 
>   System.DllNotFoundException : Unable to load DLL 'libcurl': The specified module could not be found.

stephentoub:
@kapilash, what is the exact package name that @mmitche should have installed? i.e. ```sudo apt-get install whatgoeshere```?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: Ack. I will move these two constants to the server constants file. Only point of confusion  was that this pair is not tied to the current test server. The current test server (httpbin.org) can provide a url for any given pair of username and password. In a sense they seemed like test constants than the server constants.

kapilash:
@stephentoub  The command would be: 
``` sudo apt-get install libcurl3 ```

This is the link to the package in ubuntu package repo: [http://packages.ubuntu.com/trusty/libcurl3](http://packages.ubuntu.com/trusty/libcurl3)

**Comment by mmitche** [comment]
@kapilash It was already installed:

dotnet-bot@dci-ub-fbld-3:~$ sudo apt-get install libcurl3
Reading package lists... Done
Building dependency tree
Reading state information... Done
libcurl3 is already the newest version.
0 upgraded, 0 newly installed, 0 to remove and 148 not upgraded.

**Comment by kapilash** [comment]
Thanks @mmitche I believe libcurl.so would now be present in ```/usr/local/lib``` or whatever location ```dpkg -L libcurl3``` would indicate. Which am sure would be available for the all the programs running on that machine.

Would it be possible that we restrict, via the jenkins job config, to pick libraries only from /usr/lib or such? Or override the ld_library_path?

**Comment by kapilash** [comment]
Please ignore my earlier question.. I was wrong. I verified that environment variables aren't overridden.

File: src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs (hunk 7 above)
Comment (davidsh) [inline src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs]: @kapilash You don't need to move these 2 constants.  They are test constants and not server constants since httpbin.org allows dynamic username/password to be sent/used as part of the uri path itself.

**Comment by akoeplinger** [comment]
@kapilash generally, `libXX.so` is only installed by the development packages, normal packages install `libXX.so.3` or whatever soname the library uses (e.g. compare the files included in the [libcurl3 package](http://packages.ubuntu.com/de/trusty/amd64/libcurl3/filelist) vs. [libcurl4-gnutls-dev](http://packages.ubuntu.com/de/trusty/amd64/libcurl4-gnutls-dev/filelist), only the latter has the `libcurl.so`).

This throws a wrench into the normal DllImport logic, check https://github.com/dotnet/coreclr/issues/444 or https://github.com/dotnet/coreclr/issues/930.

CoreFx has a workaround here since it does conditional compilation of the Interop classes, so you can just change the Linux PInvoke to `libcurl.so.3`.

**Comment by kapilash** [comment]
Thanks, @akoeplinger . 
Is it fair to say that this is happening only because I requested @mmitche  to install ```libcurl3``` where as, ```libcurl4``` is the "standard"  (or a more commonly depended upon) version of the library for ubuntu? 
The files list for each flavor of libcurl4 seems to contain an entry for ```libcurl.so```.  Also, the package repositories of other flavors of linux (for instance opensuse) seem to have [only libcurl4](http://software.opensuse.org/search?utf8=%E2%9C%93&q=libcurl4&search_devel=false&search_unsupported=false&baseproject=openSUSE%3A13.2).

If am right then, we will be unblocked if @mmitche could kindly install libcurl4 (via ```sudo apt-get install libcurl4-openssl-dev```) ? The [files list](http://packages.ubuntu.com/trusty/amd64/libcurl4-openssl-dev/filelist) definitely contains libcurl.so

On a related note, for OSX this is not an issue, this library is a part of xcode/OSX distribution. So we will always have a  libcurl.dylib soft-linked to the latest version.

**Comment by vijaykota** [comment]
> CoreFx has a workaround here since it does conditional compilation of the Interop classes, so you can just change the Linux PInvoke to  libcurl.so.3

@akoeplinger, so this workaround is for Linux only? For other platforms, we should continue to use libcurl.so, right?

**Comment by akoeplinger** [comment]
@kapilash no no, the `libcurl3` package is perfectly fine on Debian/Ubuntu, it already provides `libcurl.so.4` (yes the naming is unfortunate, likely kept for backwards compatibility). As you've found on other distros other packages provide that file. You shouldn't rely on `-dev` packages like `libcurl4-openssl-dev` for normal machines, they are intended for developer boxes.

You should change your PInvoke to `libcurl.so.4`.

> On a related note, for OSX this is not an issue, this library is a part of xcode/OSX distribution. So we will always have a libcurl.dylib soft-linked to the latest version. 

I don't think you want to rely on having xcode installed for normal machines, but if it's provided by OSX proper then yeah, you could continue using that.

@vijaykota hmm, depends on what the other platforms are doing, I don't know about FreeBSD/OSX.

@vijaykota yes,

**Comment by stephentoub** [comment]
cc: @nguerrera

File: src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs [hunk 9]
```diff
@@ -57,5 +88,140 @@ public void SendAsync_SimpleGet_Success()
...
+            using (var handler = new HttpClientHandler())
+            {
+                handler.MaxAutomaticRedirections = hops;
+                var client = new HttpClient(handler);
+                Assert.ThrowsAsync<HttpRequestException>(() => client.GetAsync(HttpTestServers.RedirectUriHops(hops + 1)));
+            }
+        }
+
+        [Fact]
+        public async void GetAsync_CredentialIsNetworkCredentialUriRedirect_HttpStatusCodeUnauthorized()
```
vijaykota: Nit: This should read as "StatusCodeUnauthorized" to be consistent with naming of earlier test cases

vijaykota:
In addition to the test case suggested by @stephentoub  (https://github.com/dotnet/corefx/pull/2916#discussion_r37628092), we need one where the responseContent is accessed after HttpClient has been Dispose'd. That is probably the only place we cannot use "using"

src/Common/src/Interop/FreeBSD/Interop.Libraries.cs
```diff
@@ -11,5 +11,6 @@ private static partial class Libraries
         internal const string INODE64SUFFIX = "";
 
         internal const string LibRt = "librt";  // POSIX Realtime Extensions library
+	internal const string LibCurl = "libcurl";             // Curl HTTP client library
```
stephentoub: Nit: indentation appears to be messed up

src/Common/src/Interop/Linux/Interop.Libraries.cs
```diff
@@ -11,5 +11,6 @@ private static partial class Libraries
         internal const string INODE64SUFFIX = "";
 
         internal const string LibRt = "librt"; // POSIX Realtime Extensions library
+	internal const string LibCurl = "libcurl.so.4";             // Curl HTTP client library
```
stephentoub: Nit: same here.

stephentoub: This will also require new packages to be installed in CI, correct?

src/Common/src/Interop/OSX/Interop.Libraries.cs
```diff
@@ -18,5 +18,6 @@ private static partial class Libraries
         internal const string CoreServicesLibrary   = "/System/Library/Frameworks/CoreServices.framework/CoreServices";
         internal const string libproc = "libproc";
         internal const string LibSystemKernel = "/usr/lib/system/libsystem_kernel";
+	internal const string LibCurl = "libcurl";             // Curl HTTP client library
```
stephentoub: Nit: same indentation

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -23,16 +35,34 @@ public HttpClientHandlerTest(ITestOutputHelper output)
...
-            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            var responseContent = response.Content.ReadAsStringAsync().Result;
-            _output.WriteLine(responseContent);
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+                    string responseContent = await response.Content.ReadAsStringAsync();
+		    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
stephentoub: Nit: indentation

src/Common/src/Interop/Linux/Interop.Libraries.cs
stephentoub: Why not libcurl.so.3?  What non-dev package do we install for libcurl4 on Ubuntu?

akoeplinger: The last libcurl release with soname=3 was 7.15.5 from August 2006 and it's replaced by the newer version practically everywhere. Ubuntu symlinks libcurl.so.3 -> libcurl.so.4, other distros don't even ship libcurl.so.3 anymore :)

stephentoub: Thanks, @akoeplinger.  So the libcurl3 package is the one to install for libcurl4?  Confusing :smile:

akoeplinger: Yep. Welcome to the world of Linux packaging :smile:

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -23,16 +35,53 @@ public HttpClientHandlerTest(ITestOutputHelper output)
...
-            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            var responseContent = response.Content.ReadAsStringAsync().Result;
-            _output.WriteLine(responseContent);
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+                    string responseContent = await response.Content.ReadAsStringAsync();
+		                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
stephentoub: Nit: spacing here is off... there are still tabs in the spacing

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -23,16 +35,53 @@ public HttpClientHandlerTest(ITestOutputHelper output)
...
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    // TODO: This is a placeholder until GitHub Issue #2383 gets resolved.
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.RemoteGetServer);
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                    string responseContent = await response.Content.ReadAsStringAsync();
+                    _output.WriteLine(responseContent);
```
stephentoub: There's nothing about the data in responseContent we want to be verifying / asserting?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -23,16 +35,53 @@ public HttpClientHandlerTest(ITestOutputHelper output)
...
-            var responseContent = response.Content.ReadAsStringAsync().Result;
-            _output.WriteLine(responseContent);
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+                    string responseContent = await response.Content.ReadAsStringAsync();
+		                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                    _output.WriteLine(responseContent);
```
stephentoub: There's nothing about the data in responseContent we want to be verifying / asserting?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -23,16 +35,53 @@ public HttpClientHandlerTest(ITestOutputHelper output)
...
+            HttpResponseMessage response = null;
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+            if (response != null)
```
stephentoub: In what circumstance is it acceptable for response to be null here?

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +106,152 @@ public void SendAsync_SimpleGet_Success()
...
+
+        [Theory]
+        [InlineData(6)]
+        public void GetAsync_MaxAutomaticRedirectionsNServerHopsNPlus1_Throw(int hops)
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                handler.MaxAutomaticRedirections = hops;
+                var client = new HttpClient(handler);
+                Assert.ThrowsAsync<HttpRequestException>(() => client.GetAsync(HttpTestServers.RedirectUriHops(hops + 1)));
```
stephentoub: ThrowsAsync returns a Task that you need to await.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
```diff
@@ -57,5 +106,152 @@ public void SendAsync_SimpleGet_Success()
...
+            Uri uri = HttpTestServers.BasicAuthUriForCreds(Username, Password);
+            Uri redirectUri = HttpTestServers.RedirectUriForCreds(Username, Password);
+            var credentialCache = new CredentialCache();
+            credentialCache.Add(uri, "Basic", _credential);
+
+            using (var handler = new HttpClientHandler())
+            {
+                handler.Credentials = credentialCache;
+                var client = new HttpClient(handler);
+                var response = client.GetAsync(uri).Result;
```
stephentoub: Any reason we're using Result here rather than await'ing?

kapilash: My bad. This got missed while moving to async. Apologies

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: It is a running test case, I thought, since the exception is being raised, the task is getting executed. OK, will add await on it.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: I was worried if adding further asserts would look like I am testing too many things in a single test. 
The validation on content can be added for other tests. Am more than happy to assert on the content (and any other) as well.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
kapilash: Under no circumstance it is acceptable. wrong code on my part.

src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs
stephentoub: Our goal here is to validate that the HTTP request worked.  We set up the client/handler, we make the request, and then we need to validate the response, which might include validating more than one thing.  I don't think there's anything wrong with multiple asserts in a case like this: in fact, I think it's very desired.  Imagine if there were a method on Assert called ResponseIsCorrect, and the test looked like:
```C#
HttpResponseMessage response = await client.GetAsync(HttpTestServers.SecureRemoteGetServer);
string responseContent = await response.Content.ReadAsStringAsync();
Assert.ResponseIsCorrect(response, responseContent);
```
Presumably that verification method would check multiple things, as there are multiple things that could go wrong.  Obviously there isn't such an assert method, but it highlights that the intent of of "arrange-act-assert" is maintained.

kapilash:
CurlHandler failed with Debug.Assert

    -- DEBUG ASSERTION FAILED ----
    07:28:02 ---- Assert Short Message ----
    07:28:02 
    07:28:02 ---- Assert Long Message ----
    07:28:02 
    07:28:02    at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
    07:28:02    at System.Environment.get_StackTrace()
    07:28:02    at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
    07:28:02    at System.Diagnostics.Debug.Assert(Boolean condition)
    07:28:02    at Interop.libcurl.SafeCurlMultiHandle.ReleaseHandle()
    07:28:02    at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
     07:28:02    at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
     07:28:02    at Interop.libcurl.SafeCurlMultiHandle.Dispose(Boolean disposing)
     07:28:02    at System.Runtime.InteropServices.SafeHandle.Finalize()

stephentoub:
@dotnet-bot test this please

vijaykota:
> CurlHandler failed with Debug.Assert

The fix for this is in https://github.com/dotnet/corefx/pull/2944. There is an ArgumentNullException because _multiHandlePtr is freed during Dispose. This has also been commented out in this PR and is being tracked in https://github.com/dotnet/corefx/issue/2555

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

mmitche:
@dotnet-bot test this please (testing private PR functionality)

stephentoub:
@dotnet-bot test this please

stephentoub:
@kapilash, this test is still failing:
```
09:53:48    System.Net.Http.Tests.HttpClientHandlerTest.GetAsync_SetCookieContainer_CookieSent(name: "cookiename", value: "cookievalue") [FAIL]
09:53:48       Assert.True() Failure
09:53:48       Expected: True
09:53:48       Actual:   False
09:53:48       Stack Trace:
09:53:48             at System.Net.Http.Tests.HttpClientHandlerTest.<GetAsync_SetCookieContainer_CookieSent>d__4a.MoveNext()
09:53:48          --- End of stack trace from previous location where exception was thrown ---
09:53:48             at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_0(Object state)
```

stephentoub:
@dotnet-bot test this please

(tests are passing but I want to verify that it wasn't a fluke ;)

stephentoub:
@dotnet-bot test this please

kapilash:
@davidsh,  you are right - It was a bug in httpClientHandler for unix. The default value of  ```handler.UseCookies```, is different from that of winhttphandler. Fixed it now.

kapilash:
@dotnet-bot  test this please (the earlier build seems to have been aborted)

kapilash:
The current error is: Looks like #3002 got merged. Perhaps I need to rebase and push again?

         System.Net.Http.Tests.HttpClientHandlerTest.SendAsync_SimpleGet_Success [FAIL]
          01:05:16       System.TypeInitializationException : The type initializer for 'System.Net.Http.CurlHandler' threw an exception.
           01:05:16       ---- System.TypeInitializationException : The type initializer for 'libcurl' threw an exception.
            01:05:16       -------- System.TypeInitializationException : The type initializer for 'HttpInitializer' threw an exception.
            01:05:16       ------------ System.DllNotFoundException : Unable to load DLL 'System.Net.Http.Native': The specified module could not be found.
             01:05:16        (Exception from HRESULT: 0x8007007E)
             01:05:16       Stack Trace:
             01:05:16 Finished:    System.Net.NameResolution.Linux.Pal.Tests 
01:05:16

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -63,7 +63,7 @@ internal partial class CurlHandler : HttpMessageHandler
         private bool _preAuthenticate = false;
         private CredentialCache _credentialCache = null;
         private CookieContainer _cookieContainer = null;
-        private bool _useCookie = false;
+        private bool _useCookie = true;
```
stephentoub: This should be pulled out into that shared defaults file.  I'm ok if that's done in a follow-up PR.

stephentoub:
@nguerrera, the new System.Net.Http.Native binary isn't getting built in CI; the leg is spitting out the warning:
```
CMake Warning at System.Net.Http.Native/CMakeLists.txt:5 (message):
  Cannot find libcurl, skipping build for System.Net.Http.Native.  .NET
  HttpClient is not expected to function.  Try installing libcurl4-dev (or
  the appropriate package for your platform)
```
Can you take a look?

nguerrera:
@mmitche, this time we really need to install the dev package as we need it to build the native Curl shim.

We will ultimately need to set up test runs on different machines where we don't have the dev packages and tools installed, but so long as we use the same machine config to build and test, we need to install the dev packages. We already do this for openssl and we need to do the same for curl.

Also, I'm going to change that warning in to an error. Maybe we can add --no-crypto and --no-curl options to build.sh, but the default should build all.

mmitche:
So libcurl3-dev?

nguerrera:
Libcurl4-dev



On Aug 28, 2015, at 8:56 AM, Matt Mitchell <notifications@github.com<mailto:notifications@github.com>> wrote:


So libcurl3-dev?

-

mmitche:
The one I was just asked to install last week was libcurl3 (non-dev)...libcurl4-dev is the matching version?

nguerrera:
I’m still confused about 3 vs. 4 TBH. Libcurl3-dev is a virtual package that just redirects to libcurl4-openssl-dev, which is what I actually installed last night. Libcurl4-dev doesn’t exist (incorrect recollection earlier missing the openssl part, sorry). Anyway, either libcurl3-dev or libcurl4-openssl-dev would work as they install the same thing on Ubuntu 14.04 at least.

mmitche:
Okay I did this on the Linux machines.  What about osx?

mmitche:
@dotnet-bot test this please

kapilash:
@dotnet-bot test this please (previous build seems aborted without any other error)

stephentoub:
> previous build seems aborted without any other error

The HttpClient tests hung.

mmitche:
Last one is showing compilation errors.

kapilash:
> Last one is showing compilation errors.

Yes, there was a new checkin which got into master just a little before the last build:

I need to make the same changes in    ```System.Security.Cryptography.X509Certificates/src/System.Security.Cryptography.X509Certificates.csproj``` as those needed in the System.Net.Http.csproj. I will send out the fix as soon as am able to resolve the tests-hanging issue.

kapilash:
Opened two issues - one for hanging curlHandler ( #3015 )   and one for discrepancy vis-a-vis the cookie behavior ( #3016  ). 
Commented out the negative tests till #3015 is fixed.

kapilash:
I just looked at the test results, they seemed to have passed. But I am afraid we cannot  merge these into the master yet.  These tests seem very risky, am afraid.

stephentoub:
> But I am afraid we cannot merge these into the master yet. These tests seem very risky, am afraid.

@kapilash, I'm finishing up a PR to address the threading and resource cleanup issues.  Before submitting that PR, I'll cherry-pick your tests commit into my branch, and then we can close this PR.

kapilash:
Closing this. The tests are part of #3050

PR #2917, opened 2015-08-21
Fully import Zip Linq tests from scripted legacy import

hackcraft:
Converts the Zip tests imported in 2874 into a more directly XUnit-ian style.


src/System.Linq/tests/ZipTests.cs
```diff
@@ -0,0 +1,393 @@
+// Copyright (c) Microsoft. All rights reserved.
```
Maxwe11: This file was extracted from  `LegacyTests` folder by `git rename` ?

hackcraft: Copied to new location, and then edited in place, before deleting the original after that was finished, so I could check I hadn't missed on of the cases in the original.

src/System.Linq/tests/ZipTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return this.GetEnumerator();
+            }
+        }
+
+        [Fact]
+        public void ExceptionThrownFromFirstGetEnumerator()
```
stephentoub: Isn't the exception coming from MoveNext, not from GetEnumerator?

src/System.Linq/tests/ZipTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            MyIEnum<int> first = new MyIEnum<int>(new int[] { 1, 3, 3 });
+            IEnumerable<int> second = new int[] { 2, 4, 6 };
+            Func<int, int, int> func = (x, y) => x + y;
+            IEnumerable<int> expected = new int[] { 3, 7, 9 };
+            
+            Assert.Equal(expected, first.Zip(second, func));
+
+            first = new MyIEnum<int>(new int[] { 1, 2, 3 });
+            
+            Assert.ThrowsAny<Exception>(() => first.Zip(second, func).ToList());
```
stephentoub: Should the first.Zip move to before the ThrowsAny to make it clear that the exception is coming from the enumeration and not from the operator itself?

src/System.Linq/tests/ZipTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            MyIEnum<int> second = new MyIEnum<int>(new int[] { 1, 3, 3 });
+            IEnumerable<int> first = new int[] { 2, 4, 6 };
+            Func<int, int, int> func = (x, y) => x + y;
+            IEnumerable<int> expected = new int[] { 3, 7, 9 };
+
+            Assert.Equal(expected, first.Zip(second, func));
+
+            second = new MyIEnum<int>(new int[] { 1, 2, 3 });
+
+            Assert.ThrowsAny<Exception>(() => first.Zip(second, func).ToList());
```
stephentoub: Same comments as in the First case above.

src/System.Linq/tests/ZipTests.cs
hackcraft: It is indeed, I think I'll rename to `ExceptionThrownFromFirstGetEnumerator`, since the test is fine, just the way I turned the comment on the original into a test name.

src/System.Linq/tests/ZipTests.cs
```diff
@@ -0,0 +1,393 @@
...
+        [Fact]
+        public void FirstEmptySecondSingle()
+        {
+            IEnumerable<int> first = new int[] { };
+            IEnumerable<int> second = new int[] { 2 };
+            Func<int, int, int> func = (x, y) => x + y;
+            IEnumerable<int> expected = new int[] { };
+
+            Assert.Equal(expected, first.Zip(second, func));
+        }
```
stephentoub: It's fine the way you have it, but I wonder if it'd be easier to understand these tests if the data was in-place, e.g.
```C#
[Fact]
public void FirstEmptySecondSingle()
{
    Assert.Equal(
        expected: new int[] { },
        actual: Enumerable.Zip(
            new int[] { }, 
            new int[] { 2 }, 
            (x, y) => x + y));
}
```

src/System.Linq/tests/ZipTests.cs
hackcraft: That'd definitely be a tighter test. Wilco.

src/System.Linq/tests/ZipTests.cs
hackcraft: My personal preference is to just put the expected value in directly, but I think there's a value in fidelity to the original when it doesn't do anything either very unidiomatic for XUnit and isn't doing a lot to reinvent XUnit's wheels.

stephentoub: Ok.

stephentoub:
Thanks for cleaning these up.  Other than my few comments, LGTM.

src/System.Linq/tests/ZipTests.cs
hackcraft: Also, while `ThrowsAny` duplicates the original better, `Throws` is stricter.

hackcraft:
Actioned, squashed, pushed.

stephentoub:
cc: @VSadov

PR #2918, opened 2015-08-21
Fully import ToArray and ToList tests from scripted legacy import.

hackcraft:
Converts the ToArray and ToList tests imported in 2874 into a more directly
XUnit-ian style, and merges them with the existing tests.


src/System.Linq/tests/ToArrayTests.cs
```diff
@@ -199,5 +199,95 @@ public void ToArray_ListWhereSelect(int[] sourceIntegers, string[] convertedStri
...
+            IEnumerable<int?> source = NullSeq(5);
+            int?[] expected = { null, null, null, null, null };
+
+            Assert.Null(source as ICollection<int>);
+    
+            Assert.Equal(expected, source.ToArray());
+        }
     }
-}
+}
```
Maxwe11: bring back new lines

hackcraft:
I'm adding the empty ones into this. While I'm batching these up, those are very short.

src/System.Linq/tests/ToArrayTests.cs
hackcraft: Sure will do

src/System.Linq/tests/ToArrayTests.cs
```diff
@@ -199,5 +199,95 @@ public void ToArray_ListWhereSelect(int[] sourceIntegers, string[] convertedStri
...
         }
 
+        [Fact]
+        public void SameResultsRepeatCallsFromWhereOnIntQuery()
+        {
+            var q = from x in new[] { 9999, 0, 888, -1, 66, -777, 1, 2, -12345 }
+                    where x > Int32.MinValue
+                    select x;
+
+            Assert.Equal(q.ToArray(), q.ToArray());
```
stephentoub: I know it wasn't in the original test, but for completeness, consider adding:
```C#
Assert.NotSame(q.ToArray(), q.ToArray());
```
In other words, we're verifying that we get the same contents but different objects.

src/System.Linq/tests/ToArrayTests.cs
```diff
@@ -199,5 +199,95 @@ public void ToArray_ListWhereSelect(int[] sourceIntegers, string[] convertedStri
...
+            Assert.Equal(expected, collection.ToArray());
+        }
+
+        [Fact]
+        public void SourceIsICollectionTWithFewElements()
+        {
+            int?[] source = { -5, null, 0, 10, 3, -1, null, 4, 9 };
+            int?[] expected = { -5, null, 0, 10, 3, -1, null, 4, 9 };
+
+            ICollection<int?> collection = source as ICollection<int?>;
```
stephentoub: This test is a bit odd.  The only overload of ToArray is strongly-typed to take an ```IEnumerable<T>```, so this doesn't impact what function is called, and the implementation does its own internal casting to check for special interfaces, including ```ICollection<T>```.  In case ToArray ever added specialized code paths for ```T[]``` rather than just for ```ICollection<T>```, I'd think it'd be better if this was some ```ICollection<T>``` that's not a ```T[]```, like a ```List<T>```, but we can leave it this way for now to stay true to the original test, and update it as needed I guess.

stephentoub:
LGTM

stephentoub:
cc: @VSadov

src/System.Linq/tests/ToArrayTests.cs
hackcraft: I'll add a new test for that specifically, and then a further for empty arrays, as returning the same object in that case is a possible optimisation, so it could come up in someone trying to improve the code, but has potential risks, so it would have to be discussed first.

src/System.Linq/tests/ToArrayTests.cs
hackcraft: That it tests on both `source` and `collection` was my addition; the original checked only `source` after confirming it was indeed an `ICollection<T>`, so checking that both the collection-based and non-collection-based tests worked. I figured since there was another variable there, there was no harm checking that some future change didn't do anything weird based on the type.

hackcraft:
Comments actioned. Squashed. Pushed.

stephentoub:
@dotnet-bot test this please

Maxwe11:
@dotnet-bot test this please

PR #2919, opened 2015-08-21
Use List<T>.ToArray instead of CopyTo

Maxwe11:


stephentoub:
LGTM

stephentoub:
cc: @CIPop, FYI, there are a couple of small tweaks here to NameResolution.

CIPop:
In general, we are not accepting any PRs to code within System.Net namespaces just yet as we are still trying to split up System.Private.Networking.
For this special case, I believe the changes are small enough and given that the tests passed this should be fine.

:+1:

PR #2920, opened 2015-08-21
BitVector32.CreateSection - add missing validation

Clockwork-Muse:
`BitVector32.CreateSection` claims, [in documentation](https://msdn.microsoft.com/en-us/library/530d8529(v=vs.110).aspx), to validate a given `maxValue` will fit, but not only does it not do so, the overflow check it does have is occluded in debug.  Attempting to test/verify that expected exceptions were being thrown was failing.

(For whatever reason, `Debug.Assert` is causing something in the stack to just up and die when run from a command prompt, which sent me off in the wrong direction initially.  I would have expected a message....)

This PR fixes the two issues, by removing that part of the debug check, and adding an extra check for the overflow.

This is a breaking change, in that such a `Section` can no longer be created.  However, any such 'invalid' `Section` would be dangerous; it silently truncates any data that wouldn't fit in the leftover space.  You'd get (sometimes **very**) different results out on read!

### DIFF
--- a/src/System.Collections.Specialized/src/System/Collections/Specialized/BitVector32.cs
+++ b/src/System.Collections.Specialized/src/System/Collections/Specialized/BitVector32.cs
@@ -172,16 +172,15 @@ namespace System.Collections.Specialized
                 throw new ArgumentException(SR.Format(SR.Argument_InvalidValue, "maxValue", 0), "maxValue");
             }
 #if DEBUG
-            int maskCheck = CreateMaskFromHighValue(maxValue);
-            int offsetCheck = priorOffset + CountBitsSet(priorMask);
-            Debug.Assert(maskCheck <= short.MaxValue && offsetCheck < 32, "Overflow on BitVector32");
+            Debug.Assert(CreateMaskFromHighValue(maxValue) <= short.MaxValue, "Mask out of range");
 #endif
             short offset = (short)(priorOffset + CountBitsSet(priorMask));
-            if (offset >= 32)
+            short mask = CreateMaskFromHighValue(maxValue);
+            if (offset >= 32 || offset + CountBitsSet(mask) > 32)
             {
                 throw new InvalidOperationException(SR.BitVectorFull);
             }
-            return new Section(CreateMaskFromHighValue(maxValue), offset);
+            return new Section(mask, offset);
         }
 
         public override bool Equals(object o)
--- a/src/System.Collections.Specialized/tests/BitVector32/CreateSectionShortSectionTests.cs
+++ b/src/System.Collections.Specialized/tests/BitVector32/CreateSectionShortSectionTests.cs
@@ -69,8 +69,8 @@ namespace System.Collections.Specialized.Tests
 
             sectionArgument = section;
             expected = 30;                  //expected offset
-            section = BitVector32.CreateSection(maxValue, sectionArgument);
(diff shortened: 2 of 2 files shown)


stephentoub:

Clockwork-Muse:
Argh.  For that matter, it doesn't check during set that the valid value is in range anyways.  That'd be a bit more of a breaking change.... especially since MSDN doesn't list anything.  Squawks in debug, though, which means testing for this behavior is a pain. 
It wouldn't be hard to add the validation there, should I go ahead and do it?  Or that would be too big of a change?

At least with a valid section you could just check against the mask.  With a section hanging off the end, though, you'd have to store and then immediately read to make sure your value was safe.

mikedn:
Wouldn't adding such validation be a breaking change? AFAIR the story of this and other "specialized" and non-generic collection types is that they exist in corefx for compatibility reasons so changing them in incompatible ways wouldn't be of much help. Besides, it has always been a weird type and I've rarely seen it used.

Clockwork-Muse:
@mikedn - hey, I called out that it would be a breaking change.  At minimum, I'd like to remove the `Debug.Assert` that's shadowing an expected error condition, so at least it would test cleanly....  Otherwise there's a few tests that need to be changed to not run in debug mode, which seems... problematic, long term.

I'm actually curious about the use of `Section` here too, as you're storing the equivalent of an int (2 shorts) so you can store a truncated int value...

mikedn:
@Clockwork-Muse Oops, sorry, looks like I didn't wake up completely yet. I was trying to highlight the fact that you're trying to introduce a breaking change in a type that exists mainly for compatibility.

> I'm actually curious about the use of  Section  here too, as you're storing the equivalent of an int (2 shorts) so you can store a truncated int value...

It has nothing to do with `short` and the use of `short` here is completely unwarranted. There's no reason I should not be able to store a 24 bit integer value and a couple of "flags" in the remaining bits but `BitVector32` thinks otherwise.

And in case you haven't discovered already: if you want to see usage example check the WinForms source code, it's the main user of this type.

Clockwork-Muse:
@mikedn - hey, I've made my own share of those sorts of gaffes, or worse.  
In this case, I was wondering about the _why_, which may be unanswerable.

mikedn:
@Clockwork-Muse 
> In this case, I was wondering about the why, which may be unanswerable.

Well, it seems that you just answered yourself, everyone can make gaffes :smile:. Another possibility is that due to the use of `short` sections have only 4 bytes instead of 8 bytes that would have been required if `int` was used. But that's not very helpful as sections are usually used as static fields so the size isn't really relevant. Not to mention that JIT32 can't pass such structs in a register even they would fit.

stephentoub:
@terrajobst, can you take a look at this one?  What should we do here?  You recently closed out another issue related to BitVector32.

terrajobst:
As @mikedn pointed out, we consider these types legacy and provided for compat only. Unfortunately, this type doesn't have a "modern replacement" but we aren't keen on evolving them.

Partially because nobody on the current team was around when this type was written so we'd have to reverse engineer the intent and behavior when making fixes like this. As @Clockwork-Muse just learned: that can be quite hard without breaking *something*. Of course, legacy doesn't mean we wouldn't fix bugs but the bar is fairly high as it's unlikely that anybody could point out a super impactful bug in a type that is around for more than a decade.

So long story short: I'd rather close as won't fix than messing with it...

PR #2922, opened 2015-08-21
Improve NtProcessManager.GetModuleInfos

Maxwe11:
- StringBuilder allocation moved out of loop;
- Win32Exception will be allocated and thrown only if it should be propagated to the end user.

### DIFF
--- a/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
+++ b/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
@@ -410,61 +410,66 @@ namespace System.Diagnostics
                     if (moduleCount <= moduleHandles.Length) break;
                     moduleHandles = new IntPtr[moduleHandles.Length * 2];
                 }
-                List<ModuleInfo> moduleInfos = new List<ModuleInfo>();
 
-                int ret;
+                List<ModuleInfo> moduleInfos = new List<ModuleInfo>(firstModuleOnly ? 1 : moduleCount);
+                StringBuilder baseName = new StringBuilder(1024);
+                StringBuilder fileName = new StringBuilder(1024);
+
                 for (int i = 0; i < moduleCount; i++)
                 {
-                    try
+                    if (i > 0)
                     {
-                        ModuleInfo moduleInfo = new ModuleInfo();
-                        IntPtr moduleHandle = moduleHandles[i];
-                        Interop.mincore.NtModuleInfo ntModuleInfo = new Interop.mincore.NtModuleInfo();
-                        if (!Interop.mincore.GetModuleInformation(processHandle, moduleHandle, ntModuleInfo, Marshal.SizeOf(ntModuleInfo)))
-                            throw new Win32Exception();
-                        moduleInfo._sizeOfImage = ntModuleInfo.SizeOfImage;
-                        moduleInfo._entryPoint = ntModuleInfo.EntryPoint;
-                        moduleInfo._baseOfDll = ntModuleInfo.BaseOfDll;
-
-                        StringBuilder baseName = new StringBuilder(1024);
-                        ret = Interop.mincore.GetModuleBaseName(processHandle, moduleHandle, baseName, baseName.Capacity * 2);
-                        if (ret == 0) throw new Win32Exception();
-                        moduleInfo._baseName = baseName.ToString();
-
-                        StringBuilder fileName = new StringBuilder(1024);
-                        ret = Interop.mincore.GetModuleFileNameEx(processHandle, moduleHandle, fileName, fileName.Capacity * 2);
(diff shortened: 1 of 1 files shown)


src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
```diff
@@ -410,55 +410,61 @@ private static ModuleInfo[] GetModuleInfos(int processId, bool firstModuleOnly)
...
 
-                int ret;
+                List<ModuleInfo> moduleInfos = new List<ModuleInfo>(moduleCount);
+                StringBuilder baseName = new StringBuilder(1024);
+                StringBuilder fileName = new StringBuilder(1024);
+
                 for (int i = 0; i < moduleCount; i++)
                 {
-                    try
+                    while (true)
```
stephentoub: I don't understand the purpose of this while loop.  Is it just to enable those ```break;```s in the body?  If so, wouldn't it be better to use ```continue;``` instead of ```break;``` and get rid of the ```while (true)```?

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
```diff
@@ -410,55 +410,61 @@ private static ModuleInfo[] GetModuleInfos(int processId, bool firstModuleOnly)
...
-                        }
-                        else
-                        {
-                            throw;
-                        }
+                        break;
                     }
 
+                    baseName.Clear();
+                    fileName.Clear();
```
stephentoub: These should be at the top of the for loop so that we clear even if we skip some portion of the loop due to an error, right?

stephentoub:
A couple of issues to be addressed, but otherwise nice improvements.

cc: @Priya91, @pallavit

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
Maxwe11: >I don't understand the purpose of this while loop. Is it just to enable those break;s in the body?

exactly

>If so, wouldn't it be better to use continue; instead of break; and get rid of the while (true)?

That will require move the rest of the loop to beginning and add one more check (`i == 1`)

	baseName.Clear();
	fileName.Clear();

	// If the user is only interested in the main module, break now.
	// This avoid some waste of time. In addition, if the application unloads a DLL                     
	// we will not get an exception. 
	if (firstModuleOnly && i == 1) { break; }

and Clear() foк SB will be invoked at first iteration.
It's okay?

Maxwe11:
updated with `continue` instead of `break`

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
stephentoub: It's fine to move the Clears to the top; the only impact there is we'd be clearing when it was already cleared, which should be very cheap.  I didn't understand why we'd need to move and augment the firstModuleOnly check to the top.  Can you elaborate?

stephentoub: Oh, I see.  It's not that we want to return the first module we're able to construct, it's that we want to return the first module period, even if we couldn't actually load it (in which case we'd be returning an empty set).

src/System.Diagnostics.Process/src/System/Diagnostics/ProcessManager.Windows.cs
```diff
@@ -410,61 +410,66 @@ private static ModuleInfo[] GetModuleInfos(int processId, bool firstModuleOnly)
...
+
+                    int ret = Interop.mincore.GetModuleBaseName(processHandle, moduleHandle, baseName, baseName.Capacity * 2);
+                    if (ret == 0)
+                    {
+                        HandleError();
+                        continue;
+                    }
+
+                    moduleInfo._baseName = baseName.ToString();
+                    ret = Interop.mincore.GetModuleFileNameEx(processHandle, moduleHandle, fileName, fileName.Capacity * 2);
```
stephentoub: Nit: I'd add a blank line after setting moduleInfo._baseName.  Otherwise the grouping makes it look like we're setting it in order to prepare for the GetModuleFileNameEx call.

stephentoub:
LGTM

Maxwe11:
@stephentoub @mikedn has [left a comment](https://github.com/Maxwe11/corefx/commit/68f4b221689f1674e2d183d0cf70d7a103bea548#commitcomment-12832887) at previous commit. Do you know why passed Capacity multiplied by 2 and should it be fixed?

Maxwe11:
@dotnet-bot test this please

Priya91:
LGTM

PR #2924, opened 2015-08-21
Fix deadlocks in outer loop task tests

stephentoub:
The TaskRunSyncTests tests are creating a custom TaskScheduler that spins up a bunch of threads, and then when the scheduler is disposed, it prevents additional tasks from being queued, empties the queue of remaining tasks, and waits for its threads to complete (meaning waiting for any currently running tasks to exit).

A few of the TaskRunSyncTests tests are creating tasks that, when executed, create their own scheduler to run a child task on.  This scheduler needs to be disposed (because it's another of these custom TaskSchedulers that needs to be shut down), so a continuation off of the child task is created to dispose of the child scheduler.

There are a two problems here.
- By the time this continuation is queued, the parent scheduler may have already had shutdown requested, in which case the continuation can't be queued to it, and the scheduler can't be shutdown.  Fix: Schedule the continuation to the thread pool rather than to the parent scheduler.
- The child scheduler is being created to target the parent scheduler.  I don't know if that was by design or if it was an accident, but it's a recipe for deadlock if any users of these tests were to run concurrently, as then multiple child schedulers expecting a specific number of workers might get fewer than that.  More importantly, this contributes to the previous issue: because the child scheduler was created on top of the parent, the parent couldn't shut down its threads until the child shut down its threads, but the child didn't know to shut down its threads because the continuation scheduled to the parent couldn't run.  Fix: Have the schedulers always target the thread pool rather than targeting the parent.

cc: @mellinoe, @mmitche

### DIFF
--- a/src/System.Threading.Tasks/tests/Task/TaskRunSyncTests.cs
+++ b/src/System.Threading.Tasks/tests/Task/TaskRunSyncTests.cs
@@ -71,12 +71,11 @@ namespace System.Threading.Tasks.Tests
             AbleToExecuteInline = ableToExecuteInline;
 
             /*need at least two threads since we might schedule two tasks (parent-child)*/
-            const int minThreads = 2;
-            int numberOfThreads = Environment.ProcessorCount > minThreads ? Environment.ProcessorCount : minThreads;
+            int numberOfThreads = Math.Max(Environment.ProcessorCount, 2);
             _threads = new Task[numberOfThreads];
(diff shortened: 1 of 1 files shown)


nguerrera:
LGTM

mellinoe:
LGTM, thanks for identifying the problem and fixing.

stephentoub:
@dotnet-bot test this please

PR #2874, opened 2015-08-18
System.Linq: Initial port of legacy tests

akoeplinger:
From https://github.com/VSadov/corefx/tree/UnportedLegacyTests/src/System.Linq/LegacyTestsToPort/LINQ

The tests were grouped into classes based on the tested area. The Main() method was wrapped
with an Assert that verifies it returns 0 (which means a successful test).

Further cleanup and embracing of Xunit can be done in follow-up commits. I already ran the CodeFormatter on this.

#2851 

Code coverage increased from 70.4% to 94.2%.  Fixes https://github.com/dotnet/corefx/issues/1143

PR #2927, opened 2015-08-21
Fully import First/Last First/LastOrDefault tests from scripted import

hackcraft:
Convert First, Last, FirstOrDefault & LastOrDefault tests imported in #2874
into a more directly XUnit-ian style.

Added a few more cases (some more types on some tests, and the opposite
predicate on tests on empty sequences.


src/System.Linq/tests/FirstTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            Assert.Equal(expected, source.First());
+        }
+
+        [Fact]
+        public void ManyElementsNotIListT()
+        {
+            IEnumerable<int> source = NumList(3, 10);
+            int expected = 3;
+
+            Assert.Null(source as IList<int>);
```
stephentoub: These asserts about the results of NumList are really asserting things about test code rather than about source code.  If the goal is to point out that the test depends on this not being an IList, should we just move the assert into NumList itself; that would save these asserts at the call sites, and it would highlight to anyone in the future trying to change NumList that there are dependencies on this condition.

hackcraft: True, but an improvement that gets rid of NumList entirely would slip past that.

Indeed, I considered replacing NumList with `Enumerable.Range(-5, 1)` as being the most natural way of doing the same thing, and then I considered that I myself have suggested that that should return something that implements `IList<int>`, so I was considering a change that combined with something I myself suggested would break the purpose of the test, and decided that maybe that check isn't so silly after all.

stephentoub: Ok.

stephentoub:
LGTM.  Thanks for doing this.

src/System.Linq/tests/FirstTests.cs
hackcraft: Oh. Maybe the answer is to be blatant; change `NumList()` to `NumberRangeGuaranteedNotToBeList()`. That would be pretty self-documenting as to the requirement.

I want to consolidate a lot of what's reused between these once all the tests are done (doing it before would create either order dependencies between the PRs or conflicts between them) so I think I'll do that then.

PR #2929, opened 2015-08-21
Improve FileStream WriteAsync buffering and performance

### LINKED ISSUE #1531: Win32FileStream buffer-full flushes are also done in a blocking fashion

stephentoub:
Replaces PR #2802 
Fixes #1531 

This set of changes improves how FileStream WriteAsync handles buffering, with some significant performance benefits.  Two primary issues addressed:
- Buffers were handled relatively poorly with asynchronous writes, often allocating an entirely new internal buffer in order to satisfy a request.
- Flushes were performed synchronously, which meant doing the write asynchronously and blocking for it to complete.

The changes in this PR:
- Avoid ever needing to allocate a secondary buffer
- Enable flushing asynchronously

I tested this out with a variety of small benchmarks that involve writing asynchronously, e.g.
```C#
using (FileStream fs = new FileStream("test.bin", FileMode.Create, FileAccess.Write, FileShare.None, 0x1000, useAsync:true))
{
    for (int i = 0; i < 100000; i++)
    {
        await fs.WriteAsync(arr, 0, arr.Length).ConfigureAwait(false);
    }
}
```
with various ```arr``` sizes.

As an example, on my laptop:
- On 4.6, with an ```arr``` length of 0x100,such a test took ~3.3 seconds and incurred ~27 gen0 GCs and ~2 gen1 GCs.  
- With changes we previously made for .NET Core, prior to this PR the same test took ~2.8 seconds with ~14 gen0 GCs and no gen1s.
- After this PR, this same test took ~0.23 seconds with ~1 gen0 GC.

In other words, for this test, this PR improved perf by ~10x.  The biggest benefit comes when writing out smaller than the internal buffer size.  When writing out >= the internal buffer size, throughput stayed about the same as before the change, but GCs dropped by ~6x.

The change should incur few behavioral differences.  The primary observable difference is that if a WriteAsync occurs and we need to flush the buffer, we're now flushing asynchronously rather than synchronously, but because we need to ensure that Position is correct when WriteAsync returns to its synchronous caller, the subsequent asynchronous write may execute concurrently with the flush.  That means that if the flush fails asynchronously, whereas before we wouldn't have issued the subsequent write and updated the position, now we will have.  I believe this is a worthwhile change for the benefits.

(There are some subsequent improvements we could look at as well.  For example, now that we're never allocating a secondary buffer, it becomes easier to consider using a PreAllocatedOverlapped; we would need to ensure that we only use it when there aren't concurrent writes, but that's not the common use case.)

Thanks to @tymlipari for kicking this off.

cc: @ericstj, @FiveTimesTheFun, @ianhays, @KrzysztofCwalina


stephentoub:
@dotnet-bot test this please

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1259,52 +1284,107 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
-                if (_writePos > 0)
-                    FlushWrite(false);
+                // Determine how much space remains in the buffer
+                int remainingBuffer = _bufferSize - _writePos;
+                Debug.Assert(remainingBuffer >= 0);
 
-                return WriteInternalCoreAsync(array, offset, numBytes, cancellationToken);
-            }
+                // Simple/common case:
+                // - The write is smaller than our buffer, such that it's worth considering buffering it.t
```
ianhays: nit: typo - extra 't' after period

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1259,52 +1284,107 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                // - There's no active flush operation, such that we don't have to worry about the existing buffer being in use.
+                // - And the data we're trying to write fits in the buffer, meaning it wasn't already filled by previous writes.
+                // In that case, just store it in the buffer.
+                if (numBytes < _bufferSize && !HasActiveBufferOperation && numBytes <= remainingBuffer)
```
ianhays: So if ```numBytes == _buffersize``` it skips the buffer? Makes sense; not much point in buffering it if you have to flush it on the next write regardless.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
```diff
@@ -1259,52 +1284,107 @@ private Task WriteInternalAsync(byte[] array, int offset, int numBytes, Cancella
...
+                    Buffer.BlockCopy(array, offset, _buffer, _writePos, numBytes);
+                    _writePos += numBytes;
+                    writeDataStoredInBuffer = true;
+
+                    // There is one special-but-common case (common because devs often use
+                    // byte[] sizes that are powers of 2 and thus fit nicely into our buffer, which is
+                    // also a power of 2), where we fill the buffer to the brim. In that case, we want 
+                    // to issue the asynchronous flush/write now rather than waiting until the next write.
+                    // So we only return a completed task here rather than continuing on to do the write
+                    // if the buffer isn't yet full.
```
ianhays: I'm not certain that I understand the comment here. At first it says that there is a special-but-common case where we fill the buffer to the brim and want to issue a flush/write now. but then it says we return a completed task rather than writing if the buffer isn't full.

It seems like the code is actually handling the <i>opposite</i> case from the comment i.e. the buffer is <i>not</i> filled to the brim. I think I understand now what the comment is saying, but a reword would certainly be worthwhile :)

Something like:

```
// There is one special-but-common case (common because devs often use
// byte[] sizes that are powers of 2 and thus fit nicely into our buffer, which is
// also a power of 2), where we fill the buffer to the brim. In that case, we want 
// to issue the asynchronous flush/write now rather than waiting until the next write.

// However, if the buffer still has remaining space for future writes we instead return a
// Task.CompletedTask and defer the write to a later operation.

ianhays:
A few minor comments about comments, but otherwise LGTM.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: Ok, I'll try to make the comment more clear.

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: Will fix

src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs
stephentoub: Correct.  If you're only ever writing out >= to the buffer size, we can avoid allocating the buffer at all.  And even if the buffer needs to be allocated, by skipping the buffer for the == case, we can kick off the async operation now rather than deferring it until later, which is good since we know we want to do this and we may as well overlap it as much as possible with whatever we're going to do next.  The synchronous Write already skips the buffer for ==; this brings WriteAsync in line with it.

stephentoub:
@ericstj, could you please review as well?

factormystic:
awesome :+1:

PR #2931, opened 2015-08-21
Increase System.Reflection.Emit.ILGeneration test coverage.

Priya91:


src/System.Reflection.Emit.ILGeneration/tests/Label/LabelEquals1.cs
```diff
@@ -77,5 +77,15 @@ public void PosTest6()
...
             Assert.False(label.Equals(ob));
         }
+
+        [Fact]
+        public void TestLabelEqualityOperators()
+        {
+            Label lb1 = new Label();
+            Label lb2 = new Label();
+
+            Assert.True(lb1 == lb2);
```
mellinoe: Might want to add a user message if you use Assert.True/False, as otherwise it gives a pretty useless error message.

mellinoe:
LGTM

src/System.Reflection.Emit.ILGeneration/tests/Label/LabelEquals1.cs
Priya91: The line number is reported as well. Should be easy from there.

PR #2933, opened 2015-08-22
Fully import Concat, Except, Intersect & Union Tests

hackcraft:
Convert Concat, Except, Intersect & Union tests imported in #2874 into a more
directly XUnit-ian style.


src/System.Linq/tests/ExceptTests.cs
```diff
@@ -0,0 +1,196 @@
...
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class ExceptTests
+    {
+        // Class which is passed as an argument for EqualityComparer
+        private class AnagramEqualityComparer : IEqualityComparer<string>
```
akoeplinger: I've seen this class copied and used in a bunch of the unit tests by the legay tests, might be a good future cleanup to extract it and use the same one everywhere (assuming it *is* actually identical).

hackcraft: It is identical. While I've adapted it a bit from the original (equals has some short-circuits, quicker hash code that is likely less good distribution but that doesn't matter on the sizes used here) I think merging it is best left until the end, as otherwise it will just lead to the batches of transfers in each PR either depending on each other in a particular path, or else conflicting with each other.

akoeplinger: Sounds good.

src/System.Linq/tests/ExceptTests.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        [Fact]
+        public void FirstNullCustomComparer()
+        {
+            string[] first = null;
+            string[] second = { "bBo", "shriC" };
+
+            var ane = Assert.Throws<ArgumentNullException>(() => first.Except(second, new AnagramEqualityComparer()));
+            Assert.Equal("first", ane.ParamName);
```
stephentoub: I believe this could be done slightly more concisely as:
```C#
Assert.Throws<ArgumentNullException>("first", () => first.Except(second, new AnagramEqualityComparer()));
```
but it's not a big deal either way.

src/System.Linq/tests/IntersectTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        {
+            int?[] first = { 1, 2, null, 3, 4, 5, 6 };
+            int?[] second = { 6, 7, 7, 7, null, 8, 1 };
+            int?[] expected = { 1, null, 6 };
+
+            Assert.Equal(expected, first.Intersect(second));
+        }
+
+        [Fact]
+        public void EachHasRepeatsBetweenAndAmonthThemselves()
```
stephentoub: Amonth => Amongst ?

stephentoub:
A few minor nits, but generally looks good. Thanks!  This should be rebased due to merge conflicts (I assume with the project file).

src/System.Linq/tests/ExceptTests.cs
hackcraft: TIL.

hackcraft:
Yes, the project file was the only conflict. I should try doing them for these by hand and keeping the changed references on the original line, hopefully that would reduce the conflict rate.

PR #2939, opened 2015-08-23
Fully import ToDictionaryTests, ToLookupTests & ToSequenceTests

hackcraft:
Convert ToDictionary, ToLookup and AsEnumerable tests imported in #2874 into a
more directly XUnit-ian style.

Merge ToDictionary with those already present.

Changed AsEnumerable tests to explicitly check for equality with the source, as
AsEnumerable() should result in that by definition.


src/System.Linq/tests/ToSequenceTests.cs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class ToSequenceTests
```
stephentoub: The file and class should should probably be renamed to AsEnumerableTests (rather than ToSequenceTests), right?

stephentoub:
One nit, otherwise LGTM.  Thanks for continuing to plow through cleaning these up.

cc: @VSadov

hackcraft:
Nit actioned, squashed, pushed.

src/System.Linq/tests/ToSequenceTests.cs
jasonwilliams200OK: @hackcraft, The PR title can also be changed accordingly: ToSequenceTests => AsEnumerableTests. :smile:

stephentoub:
Thanks, @hackcraft.

PR #2941, opened 2015-08-23
Fix IsTransparentProxy check in Microsoft.CSharp

stephentoub:
Fixes https://github.com/dotnet/coreclr/issues/1425

Thanks to @JamesNK for finding the issue and @mikedn for spotting the problem.

@AlekseyTs, @VSadov, there are no tests for Microsoft.CSharp in the repo.  Can you help with whatever verification needs to be done on this?

### DIFF
--- a/src/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/BinderHelper.cs
+++ b/src/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/BinderHelper.cs
@@ -195,7 +195,10 @@ namespace Microsoft.CSharp.RuntimeBinder
 
         private static bool IsTransparentProxy(object obj)
         {
-            return obj != null; 
+            // In the full framework, this checks:
+            //     return obj != null && RemotingServices.IsTransparentProxy(obj);
+            // but transparent proxies don't exist in .NET Core.
+            return false;
         }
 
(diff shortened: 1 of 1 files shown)


mikedn:
@stephentoub  Thanks for making a fix, I was thinking to do it myself but I was sure that discussions about tests will start up :smile:

stephentoub:
:smile:

src/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/BinderHelper.cs
```diff
@@ -195,7 +195,10 @@ internal static bool IsWindowsRuntimeObject(DynamicMetaObject obj)
 
         private static bool IsTransparentProxy(object obj)
         {
-            return obj != null;
```
VSadov: This looks like just a logical bug in trying to keep the part of the check unaffected by absence of RemotingServices.IsTransparentProxy

stephentoub: Yes. You ok with the fix? Any additional validation needed?

VSadov:
System.Dynamic.Runtime has a bunch of tests for C# dynamic. Perhaps not the most intuitive place, but that is where they are now.
I think the change is good and testing seems sufficient.

PR #2943, opened 2015-08-24
Fully import Average and Aggregate Tests

JonHanna:
Convert Average, Aggregate tests imported in #2874 into a more directly
XUnit-ian style.


src/System.Linq/tests/AggregateTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        [Fact]
+        public void TwoElementsSeedResultSelector()
+        {
+            int[] source = { 5, 6 };
+            long seed = 2;
+            long expected = 65;
+
+            Assert.Equal(expected, source.Aggregate(seed, (x, y) => x * y, x => x + 5.0));
```
stephentoub: I notice that the original test had ```x + 5``` and you changed it to ```x + 5.0```... any reason?  Same for the next one.

JonHanna: The original had its selector typed as `Func<long, double>` so it was effectively `x => (double)(x + 5)`. In inlining the seletor, `x => x + 5.0` was a way to keep the selector doing what it did in the original.

stephentoub: Ah, you're right.

stephentoub:
LGTM

PR #2944, opened 2015-08-24
Fix an assert in CurlHandler multi handle release path

vijaykota:
- If a timeout happens, the socket fd is not removed from _fdSet. When
  SafeCurlMultiHandle is finalized, it checks if _fdSet is empty and
  this assertion was failing. So ensuring that all codepaths of
  EndRequest result in cleanup from _fdSet

### DIFF
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
@@ -341,7 +341,6 @@ namespace System.Net.Http
         {
             SafeCurlHandle requestHandle = new SafeCurlHandle();
             GCHandle stateHandle = new GCHandle();
-            Exception exception = null;
 
             try
             {
@@ -351,31 +350,27 @@ namespace System.Net.Http
                 requestHandle = CreateRequestHandle(state, stateHandle);
                 state.RequestHandle = requestHandle;
 
-                if (!state.CancellationToken.IsCancellationRequested && (state.RequestMessage.Content != null))
+                if (state.RequestMessage.Content != null)
                 {
                     Stream requestContentStream =
                         await state.RequestMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
                     state.RequestContentStream = requestContentStream;
-                    if (!state.CancellationToken.IsCancellationRequested)
+                    if (state.CancellationToken.IsCancellationRequested)
                     {
-                        state.RequestContentBuffer = new byte[s_requestBufferSize];
+                        RemoveEasyHandle(_multiHandle, stateHandle, state);
+                        state.TrySetCanceled(state.CancellationToken);
+                        return;
                     }
+                    state.RequestContentBuffer = new byte[s_requestBufferSize];
                 }
 
-                if (!state.CancellationToken.IsCancellationRequested)
-                {
-                    AddEasyHandle(state);
-                    return;
-                }
+                AddEasyHandle(state);
             }
             catch (Exception ex)
             {
-                exception = ex;
+                RemoveEasyHandle(_multiHandle, stateHandle, state);
+                HandleAsyncException(state, ex);
             }
-
-            // Either there was a cancellation or an exception
-            RemoveEasyHandle(_multiHandle, stateHandle, state);
-            HandleAsyncException(state, exception);
         }
 
         private static void EndRequest(SafeCurlMultiHandle multiHandle, IntPtr statePtr, int result)
@@ -956,11 +951,11 @@ namespace System.Net.Http
 
         private static void RemoveEasyHandle(SafeCurlMultiHandle multiHandle, GCHandle stateHandle, RequestCompletionSource state)
         {
-            RemoveEasyHandle(multiHandle, state, false);
             if (stateHandle.IsAllocated)
             {
                 stateHandle.Free();
             }
+            RemoveEasyHandle(multiHandle, state, false);
         }
 
         private static void RemoveEasyHandle(SafeCurlMultiHandle multiHandle, GCHandle stateHandle, bool onMultiStack)
@@ -968,8 +963,8 @@ namespace System.Net.Http
             Debug.Assert(stateHandle.IsAllocated);
 
             RequestCompletionSource state = (RequestCompletionSource)stateHandle.Target;
-            RemoveEasyHandle(multiHandle, state, onMultiStack);
             stateHandle.Free();
+            RemoveEasyHandle(multiHandle, state, onMultiStack);
         }
 
         private static void RemoveEasyHandle(SafeCurlMultiHandle multiHandle, RequestCompletionSource state, bool onMultiStack)
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
@@ -341,7 +341,7 @@ namespace System.Net.Http
         {
             SafeCurlHandle requestHandle = new SafeCurlHandle();
             GCHandle stateHandle = new GCHandle();
-            bool needCleanup = false;
+            Exception exception = null;
 
             try
             {
@@ -350,52 +350,32 @@ namespace System.Net.Http
                 stateHandle = GCHandle.Alloc(state);
                 requestHandle = CreateRequestHandle(state, stateHandle);
                 state.RequestHandle = requestHandle;
-                needCleanup = true;
 
-                if (state.CancellationToken.IsCancellationRequested)
-                {
-                    state.TrySetCanceled(state.CancellationToken);
-                    return;
-                }
-
-                if (state.RequestMessage.Content != null)
+                if (!state.CancellationToken.IsCancellationRequested && (state.RequestMessage.Content != null))
                 {
                     Stream requestContentStream =
                         await state.RequestMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
-                    if (state.CancellationToken.IsCancellationRequested)
-                    {
-                        state.TrySetCanceled(state.CancellationToken);
-                        return;
-                    }
                     state.RequestContentStream = requestContentStream;
-                    state.RequestContentBuffer = new byte[s_requestBufferSize];
+                    if (!state.CancellationToken.IsCancellationRequested)
+                    {
+                        state.RequestContentBuffer = new byte[s_requestBufferSize];
+                    }
                 }
 
-                AddEasyHandle(state);
-                needCleanup = false;
+                if (!state.CancellationToken.IsCancellationRequested)
+                {
+                    AddEasyHandle(state);
+                    return;
+                }
             }
             catch (Exception ex)
             {
-                HandleAsyncException(state, ex);
+                exception = ex;
             }
-            finally
-            {
-                if (needCleanup)
-                {
-                    RemoveEasyHandle(_multiHandle, stateHandle, false);
-                }
-                else if (state.Task.IsCompleted)
-                {
-                    if (stateHandle.IsAllocated)
-                    {
-                        stateHandle.Free();
-                    }
-                    if (!requestHandle.IsInvalid)
-                    {
-                        SafeCurlHandle.DisposeAndClearHandle(ref requestHandle);
-                    }
-                }
-            }
+
+            // Either there was a cancellation or an exception
+            RemoveEasyHandle(_multiHandle, stateHandle, state);
+            HandleAsyncException(state, exception);
         }
 
         private static void EndRequest(SafeCurlMultiHandle multiHandle, IntPtr statePtr, int result)
@@ -403,21 +383,14 @@ namespace System.Net.Http
             GCHandle stateHandle = GCHandle.FromIntPtr(statePtr);
             RequestCompletionSource state = (RequestCompletionSource)stateHandle.Target;
 
-            try
+            if (CURLcode.CURLE_OK == result && state.ResponseMessage.StatusCode != HttpStatusCode.Unauthorized && state.Handler.PreAuthenticate)
             {
-                if (CURLcode.CURLE_OK == result && state.ResponseMessage.StatusCode != HttpStatusCode.Unauthorized && state.Handler.PreAuthenticate)
+                ulong availedAuth;
+                if (Interop.libcurl.curl_easy_getinfo(state.RequestHandle, CURLINFO.CURLINFO_HTTPAUTH_AVAIL, out availedAuth) == CURLcode.CURLE_OK)
                 {
-                    ulong availedAuth;
-                    if (Interop.libcurl.curl_easy_getinfo(state.RequestHandle, CURLINFO.CURLINFO_HTTPAUTH_AVAIL, out availedAuth) == CURLcode.CURLE_OK)
-                    {
-                        state.Handler.AddCredentialToCache(state.RequestMessage.RequestUri, availedAuth, state.NetworkCredential);
-                    }
-                    // ignore errors
+                    state.Handler.AddCredentialToCache(state.RequestMessage.RequestUri, availedAuth, state.NetworkCredential);
                 }
-            }
-            catch
-            {
-                // ignoring the exception in this case.
+                // ignore errors
                 // There is no point in killing the request for the sake of putting the credentials into the cache
             }
 
@@ -484,7 +457,7 @@ namespace System.Net.Http
                 if (state.RequestMessage.Content == null)
                 {
                     SetCurlOption(requestHandle, CURLoption.CURLOPT_POSTFIELDSIZE, 0L);
-                    SetCurlOption(requestHandle, CURLoption.CURLOPT_POSTFIELDS, "");
+                    SetCurlOption(requestHandle, CURLoption.CURLOPT_COPYPOSTFIELDS, String.Empty);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


stephentoub:
Thanks, @vijaykota, but with this change I'm still hitting the other asserts in ReleaseHandle.  I added message strings to the asserts to see which ones were being hit, and I frequently get this when running the System.Net.Requests.Tests:
```
_requestCount isn't 0
---- Assert Long Message ----

   at System.Diagnostics.Debug.Assert(Boolean condition, String message)
   at Interop.libcurl.SafeCurlMultiHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
   at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
   at Interop.libcurl.SafeCurlMultiHandle.Dispose(Boolean disposing)
   at System.Runtime.InteropServices.SafeHandle.Finalize()
_pollCanceled isn't true
---- Assert Long Message ----

   at System.Diagnostics.Debug.Assert(Boolean condition, String message)
   at Interop.libcurl.SafeCurlMultiHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
   at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
   at Interop.libcurl.SafeCurlMultiHandle.Dispose(Boolean disposing)
   at System.Runtime.InteropServices.SafeHandle.Finalize()
```
so while it's not hitting that first assert, it's still hitting both of the next two.

stephentoub:
With the System.Net.Requests tests, I also hit this just now:
```
Unhandled Exception: System.InvalidOperationException: Cannot write data: 9
   at Interop.libcurl.SafeCurlMultiHandle.SignalFdSetChange(Int32 fd, Boolean isRemove)
   at System.Net.Http.CurlHandler.CurlSocketCallback(IntPtr handle, Int32 socketFd, Int32 socketAction, IntPtr context, IntPtr sockPtr)
```
I've also hit a seg fault, but I've not yet been able to capture a core dump to find out from where it's coming from.  It happens maybe 10% of the time I run the Requests tests.

vijaykota:
> I've also hit a seg fault, but I've not yet been able to capture a core dump to find out f

Sorry @stephentoub , I didn't realize that you were running System.Net.Requests tests. I think we should be able to repro this as well as the asserts. I will investigate and update the PR accordingly. Apologies for the premature PR

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -400,6 +402,27 @@ private static void EndRequest(SafeCurlMultiHandle multiHandle, IntPtr statePtr,
...
+                    if (Interop.libcurl.curl_easy_getinfo(state.RequestHandle, CURLINFO.CURLINFO_HTTPAUTH_AVAIL, out availedAuth) == CURLcode.CURLE_OK)
+                    {
+                        state.Handler.AddCredentialToCache(state.RequestMessage.RequestUri, availedAuth, state.NetworkCredential);
+                    }
+                    // ignore errors
+                }
+            }
+            catch
+            {
+                // ignoring the exception in this case.
```
stephentoub: What exception is this ignoring / where would it be coming from, AddCredentialToCache?  Can we be more specific in what kinds of exceptions we're catching rather than catching everything?

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -280,10 +280,12 @@ protected override void Dispose(bool disposing)
...
-                if (_multiHandlePtr.IsAllocated)
-                {
-                    _multiHandlePtr.Free();
-                }
+                // TODO: Check for the correct place to free this since
+                //       its lifetime should match _multiHandle
+                //if (_multiHandlePtr.IsAllocated)
+                //{
+                //    _multiHandlePtr.Free();
+                //}
```
stephentoub: Is there an issue tracking this?

stephentoub:
Thanks, @vijaykota.  This seems to have made the Debug.Assert problems go away, but I'm still hitting memory corruption and seg fault errors.

Here's a stack I see aborting because of a memory corruption problem:
```
Breakpoint 1 at 0x7f764ab75cc9: file ../nptl/sysdeps/unix/sysv/linux/raise.c, line 56.
(gdb) bt
#0  0x00007f764ab75cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f764ab790d8 in __GI_abort () at abort.c:89
#2  0x00007f764abb2394 in __libc_message (do_abort=1, fmt=fmt@entry=0x7f764acc0b28 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007f764abbff36 in malloc_printerr (ptr=0x7f764bb35010, str=0x7f764acbcc9a "malloc(): memory corruption", action=<optimized out>)
    at malloc.c:4996
#4  _int_malloc (av=0x7f75e0000020, bytes=64) at malloc.c:3447
#5  0x00007f764abc17b0 in __GI___libc_malloc (bytes=64) at malloc.c:2891
#6  0x00007f764ac0b32e in gaih_inet (name=name@entry=0x1f6ebc0 "httpbin.org", service=<optimized out>, req=req@entry=0x1f6eb80, 
    pai=pai@entry=0x7f75d3fe6cd0, naddrs=naddrs@entry=0x7f75d3fe6cc0) at ../sysdeps/posix/getaddrinfo.c:1204
#7  0x00007f764ac0ee0d in __GI_getaddrinfo (name=0x1f6ebc0 "httpbin.org", service=0x7f75d3fe6e80 "80", hints=0x1f6eb80, pai=0x7f75d3fe6e38)
    at ../sysdeps/posix/getaddrinfo.c:2406
#8  0x00007f7616dd1164 in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#9  0x00007f7616dde014 in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#10 0x00007f7616ddb8bb in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#11 0x00007f764a02a182 in start_thread (arg=0x7f75d3fe7700) at pthread_create.c:312
#12 0x00007f764ac3947d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```
This stack appears to be from a thread managed by libcurl, but presumably it's failing because of some input we gave it, maybe some pointer that we inappropriately free'd or something like that.  For example, are there any P/Invokes where we're passing in a string under the expectation that the marshaled char* will be copied such that it's fine the runtime frees the char* upon returning from the P/Invoke, but where it may not be copied?  As an example, I took a peek at the curl_easy_setopt docs, which state:
```
Strings passed to libcurl as 'char *' arguments, are copied by the library; thus the string storage associated to the pointer argument may be overwritten after curl_easy_setopt returns. The only exception to this rule is really CURLOPT_POSTFIELDS, but the alternative that copies the string CURLOPT_COPYPOSTFIELDS has some usage characteristics you need to read up on.
```
and I see in the code that under some circumstances we are using CURLOPT_COPYPOSTFIELDS and just passing in a string, which would presumably cause such memory problems.

And here's one I see for the seg fault:
```
(gdb) bt
#0  __GI_fgets_unlocked (buf=0x7fb4897d9620 "", n=n@entry=1032, fp=0x7fb4897d9620) at iofgets_u.c:53
#1  0x00007fb4c05d1afa in get_contents (stream=<optimized out>, len=1032, linebuf=0x7fb4897d9620 "") at nss_files/files-XXX.c:201
#2  internal_getent (result=0x7fb4897d9580, buffer=buffer@entry=0x7fb4897d9600 "", buflen=buflen@entry=1064, errnop=errnop@entry=0x7fb4897d9b70, 
    herrnop=herrnop@entry=0x7fb4897d9bc0, af=af@entry=0, flags=flags@entry=0) at nss_files/files-XXX.c:246
#3  0x00007fb4c05d2ca3 in _nss_files_gethostbyname4_r (name=name@entry=0x7fb4e00243d0 "httpbin.org", pat=pat@entry=0x7fb4897d9ba0, 
    buffer=buffer@entry=0x7fb4897d9600 "", buflen=buflen@entry=1064, errnop=errnop@entry=0x7fb4897d9b70, herrnop=herrnop@entry=0x7fb4897d9bc0, 
    ttlp=ttlp@entry=0x0) at nss_files/files-hosts.c:402
#4  0x00007fb500df4757 in gaih_inet (name=<optimized out>, name@entry=0x7fb4e00243d0 "httpbin.org", service=<optimized out>, 
    req=req@entry=0x7fb4e0024390, pai=pai@entry=0x7fb4897d9cd0, naddrs=naddrs@entry=0x7fb4897d9cc0) at ../sysdeps/posix/getaddrinfo.c:850
#5  0x00007fb500df7e0d in __GI_getaddrinfo (name=0x7fb4e00243d0 "httpbin.org", service=0x7fb4897d9e80 "80", hints=0x7fb4e0024390, 
    pai=0x7fb4897d9e38) at ../sysdeps/posix/getaddrinfo.c:2406
#6  0x00007fb4dd309164 in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#7  0x00007fb4dd316014 in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#8  0x00007fb4dd3138bb in ?? () from /usr/lib/x86_64-linux-gnu/libcurl.so
#9  0x00007fb500213182 in start_thread (arg=0x7fb4897da700) at pthread_create.c:312
#10 0x00007fb500e2247d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```
Maybe it's the same issue?

stephentoub:
> Maybe it's the same issue?

FYI, this may be one issue, but it's not the only issue.  I tried changing the POSTFIELDS to use IntPtr.Zero instead of sending a string, and while that's likely a good thing to do and maybe makes some issues go away, it doesn't make them all go away.  After changing that, I still hit such errors.  I'll let you debug further.

stephentoub:
Another thing to investigate/fix (may or may not be related): doesn't look like we're following all of the required steps for using libcurl in a multithreaded environment: http://curl.haxx.se/libcurl/c/threadsafe.html

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
vijaykota: Saw that this is just a Dictionary.Add. So will remove the try/catch

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
vijaykota: I want to fix this in same PR. Still investigating a fix

vijaykota:
> memory corruption and seg fault errors.

I will work with you offline regarding repro for this. I think I am still missing some steps. I still get an Assert because Finalize is getting called while poll is in progress. But not getting a seg fault

vijaykota:
>  I tried changing the POSTFIELDS to use IntPtr.Zero 

We were thinking that POSTFIELDSIZE is checked for non-zero before libcurl uses POSTFIELDS. Maybe this is not the case. If this turns out to be cause of some seg faults, we will change this to use COPYPOSTFIELDS (instead of IntPtr.Zero which is the default value)

vijaykota:
> required steps for using libcurl in a multithreaded environment: http://curl.haxx.se/libcurl/c/threadsafe.html

We followed this guidance as well as some threads on gmane.org/stackoveflow/curl mailing list etc. For handles, we are ensuring that the handles are not used concurrently. Only part I am unsure about is curl_global_init. I don't know if there is a scenario for an existing thread before the static CurlHandler ctor which can later call libcurl APIs

stephentoub:
> we followed this guidance 

Great.

What about the the comments about NOSIGNAL?  We're using multiple threads, and the docs state:
```
When using multiple threads you should set the CURLOPT_NOSIGNAL option to 1L for all handles.
```
but I don't see us using CURLOPT_NOSIGNAL in the code.

And what about:
```
If you are accessing HTTPS or FTPS URLs in a multi-threaded manner, you are then of course using the underlying SSL library multi-threaded and those libs might have their own requirements on this issue. 
```
Don't we need to ensure that OpenSSL is properly initialized (e.g. its locking mechanism) as we do in our crypto libraries?

And:
```
These functions, provided either by your operating system or your own replacements, must be thread safe. You can use curl_global_init_mem to set your own replacement memory functions.
```
are we sure that the memory routines being used on at least Linux and OS X are thread-safe?  If yes, great.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
vijaykota: Removed the try/catch

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -348,58 +350,52 @@ protected override void Dispose(bool disposing)
...
-                }
 
-                if (state.RequestMessage.Content != null)
+                if (!state.CancellationToken.IsCancellationRequested && (state.RequestMessage.Content != null))
                 {
                     Stream requestContentStream =
                         await state.RequestMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
-                    if (state.CancellationToken.IsCancellationRequested)
+                    state.RequestContentStream = requestContentStream;
+                    if (!state.CancellationToken.IsCancellationRequested)
```
stephentoub: This cancellation check adds value, because it's being done after an await where a cancellation token couldn't be provided to the async operation.  The other two cancellation checks in this method don't seem to add much value, as each of them comes very soon after another check was made.  Doesn't have to be as part of this PR, but I'd suggest removing those extra checks.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -971,31 +954,57 @@ private void AddEasyHandle(RequestCompletionSource state)
...
+            }
+        }
+
         private static void RemoveEasyHandle(SafeCurlMultiHandle multiHandle, GCHandle stateHandle, bool onMultiStack)
         {
+            Debug.Assert(stateHandle.IsAllocated);
+
             RequestCompletionSource state = (RequestCompletionSource)stateHandle.Target;
+            RemoveEasyHandle(multiHandle, state, onMultiStack);
+            stateHandle.Free();
```
stephentoub: I'd suggest moving the stateHandle.Free() up one line, to be as soon as you're done with it.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -971,31 +954,57 @@ private void AddEasyHandle(RequestCompletionSource state)
             }
         }
 
+        private static void RemoveEasyHandle(SafeCurlMultiHandle multiHandle, GCHandle stateHandle, RequestCompletionSource state)
+        {
+            RemoveEasyHandle(multiHandle, state, false);
```
stephentoub: Can this RemoveEasyHandle call be moved down to below the if block?  Ideally we dispose of these resources as soon as we no longer need them, so that the GC doesn't have to track it for longer and for a bit more reliability in call to RemoveEasyHandle were to throw.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
vijaykota: Makes sense. Restructured this code not to use extra checks and variables

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
```diff
@@ -348,58 +349,48 @@ protected override void Dispose(bool disposing)
...
-                }
 
                 if (state.RequestMessage.Content != null)
                 {
                     Stream requestContentStream =
                         await state.RequestMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
+                    state.RequestContentStream = requestContentStream;
                     if (state.CancellationToken.IsCancellationRequested)
                     {
+                        RemoveEasyHandle(_multiHandle, stateHandle, state);
```
stephentoub: We haven't added this handle yet, right?  Is this call here to instead clean up things like stateHandle?

stephentoub:
LGTM

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
vijaykota: Yes, the handle hasn't been added yet. (so onMultiStack==false)

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.cs
stephentoub: Actually, I may have given bad advice here.  Is this easy handle still potentially running and causing events to be invoked?  If so, we don't want to free the state handle until after we're sure no more callbacks could be invokd with the context ptr.

PR #2945, opened 2015-08-24
Fully import ‎Count and LongCount Tests

hackcraft:
Convert ‎Count and LongCount tests imported in #2874 into a more directly
XUnit-ian style.


stephentoub:
LGTM!

PR #2946, opened 2015-08-24
Fully import Cast and OfType Tests

JonHanna:
Convert Cast and OfType tests imported in #2874 into a more directly
XUnit-ian style.

Removed one duplicate. Expanded on test to check it throws at the precise point
within the sequence where it should throw.


stephentoub:
LGTM.  Thanks, @JonHanna.

PR #2948, opened 2015-08-24
Throw correct exception when Enumerable.Cast null to non-nullable

### LINKED ISSUE #2947: Enumerable.Cast() throws wrong exception in casting null to non-nullable value-type.
Enumerable.Cast() is documented as throwing:

`ArgumentNullException`: source enumerable is null.
`InvalidCastException`: (upon enumerating) an element in the sequence cannot be cast to the requested type.

If however a null element is in a sequence being cast to a non-nullable value-type, the exception thrown is `NullReferenceException`, which is both not documented, and an inappropriate exception in any case as the calling code has not directly used a null reference.

Example:

    new int?[]{1, null}.Cast<int>().ToList()

JonHanna:
Fixes #2947

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1122,7 +1122,16 @@ namespace System.Linq
 
         private static IEnumerable<TResult> CastIterator<TResult>(IEnumerable source)
         {
-            foreach (object obj in source) yield return (TResult)obj;
+            if (default(TResult) == null)
+                foreach (object obj in source) yield return (TResult)obj;
+            else
+            {
+                foreach (object obj in source)
+                {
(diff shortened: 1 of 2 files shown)


akoeplinger:
I'd separate out the migration of the legacy tests to xunit in another PR, so it doesn't get mixed up with your exception fix (which I guess needs to be discussed whether it's appropriate to change behavior).

akoeplinger:
Ah sorry, you already did that, I'm dumb :sob:

stephentoub:
cc: @VSadov

stephentoub:
If the decision is this is the behavioral change make, the change LGTM.  @VSadov?

stephentoub:
Closing per discussion on the associated issue.  Thanks.

PR #2950, opened 2015-08-24
Fully import Range, Repeat & Reverse tests

JonHanna:
Convert Range, Repeat & Reverse tests imported in #2874 into a more
directly XUnit-ian style and merge with existing tests.

A few dropped as duplicates of those already there.


stephentoub:
Thanks, @JonHanna.  This one took me a bit longer to verify, due to having to match up old legacy tests against existing tests while ensuring that none were missed, but it all looks good.  Thanks, again.

PR #2951, opened 2015-08-24
Implement ECDsaOpenSsl

AtsushiKan:


AtsushiKan:
cc @bartonjs

src/Common/src/Interop/Unix/libcrypto/Interop.EcKey.cs
```diff
@@ -0,0 +1,45 @@
...
+        [return: MarshalAs(UnmanagedType.Bool)]
+        internal static extern bool EC_KEY_generate_key(SafeEcKeyHandle eckey);
+
+        [DllImport(Libraries.LibCrypto)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        internal static extern bool EC_KEY_up_ref(IntPtr r);
+
+        internal static int EcKeyGetCurveName(SafeEcKeyHandle ecKey)
+        {
+            IntPtr ecGroup = EC_KEY_get0_group(ecKey.DangerousGetHandle());
```
bartonjs: Why pass the DangerousGetHandle instead of making the method take the SafeHandle?

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/ECDsaOpenSsl.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <param name="handle">A pointer to an OpenSSL <c>EC_KEY*</c></param>
+        /// <exception cref="ArgumentException"><paramref name="handle" /> is invalid</exception>
+        public ECDsaOpenSsl(IntPtr handle)
+        {
+            if (handle == IntPtr.Zero)
+                throw new ArgumentException(SR.Cryptography_OpenInvalidHandle, "handle");
+
+            SafeEcKeyHandle ecKeyHandle = SafeEcKeyHandle.DuplicateHandle(handle);
+
+            // Set base.KeySize to avoid throwing an extra Lazy at the GC when
```
bartonjs: I don't see what this comment is pertaining to; there aren't any calls to base.KeySize here...

bartonjs:
Nice.  The p/invoke question is really the only issue I have (the comment suggests that there's a problem pattern that I don't know about, so it could be that I just need to learn what that problem pattern is).  Well, and the comment I didn't follow about base.KeySize.

Other than those two, LGTM.

bartonjs:
Since we don't have OSX test execution as part of the CI flow I went ahead and checked there, and everything passed.  Yay :smile:.

src/Common/src/Interop/Unix/libcrypto/Interop.EcKey.cs
AtsushiKan: Making EC_KEY_get0_group(0) take a SafeHandle would give a false sense of security.
Forcing the caller to call DangerousGetHandle() also reminds him that he needs to keep the SafeHandle alive across
the entire sequence and not just rely on P/Invoke to do the right thing.

stephentoub: Then shouldn't we also be doing the DangerousAddRef/DangerousRelease dance?

AtsushiKan: There’s an GC.KeepAlive() call below the visible part of the diff window that handles the lifetime guarantee.

stephentoub: It doesn't handle the case where code elsewhere calls Dispose on the SafeHandle.

Issue #812, opened 2015-02-18
Implement async I/O support in System.IO.FileSystem/Pipes on Unix


stephentoub:
System.IO.FileSystem has an initial implementation on Unix, but it doesn't yet have special support for async I/O (right now the async operations just rely on the base Stream type's implementation, which will schedule a work item to invoke the synchronous implementations).

stephentoub:
After further investigation, there doesn't seem to be much more we can do here for FileStream.  io_submit would require restrictions on the FileStream (e.g. requiring O_DIRECT) and is only available on Linux.  epoll doesn't work with files.  And aio would essentially provide the same solution we already employ, offloading the synchronous call to another thread.

sokket:
Okay cool, if this isn't actionable anymore then I'll close the issue. Thanks Stephen!

PR #2952, opened 2015-08-24
Make PipeStream.ReadAsync on an async stream cancelable on Unix

stephentoub:
Contributes to #812.
cc: @ianhays, @ericstj, @ellismg, @bpschoch, @Maxwe11 
I suggest reviewing a commit at a time; I had to make some changes along the way to how things were structured, and that'll make it easier to follow along.

Today on Unix PipeStream.Read/WriteAsync just delegate to the base Stream's ReadAsync/WriteAsync methods, which in turn queue work items to call Read/Write.  This is how Window's implementation of PipeStream works, too, but only when the stream has isAsync false.  When the stream is created to be asynchronous, on Windows ReadAsync/WriteAsync use overlapped I/O.  In addition to avoiding blocked threads, this has a big advantage: the operations are cancelable.  With our Unix implementation, once we do the initial check for whether cancellation has been requested, we then call the blocking read/write libc methods, and ignore cancellation from that point on. This PR addresses that, at least for ReadAsync.

First, I refactored the implementation to avoid relying on the base Stream's ReadAsync/WriteAsync for the isAsync==true case.  This doesn't in and of itself enable cancellation, but it does give us a place to plug it in.  It also fixes a difference we had between Windows and Unix: on Windows when isAsync==true, Read/Write calls delegated to ReadAsync/WriteAsync, such that they were all appropriately synchronized.  That wasn't the case on Unix, which meant the same usage patterns could result in non-thread-safe accesses to the type.  That's now fixed.

With that in place, I added a cancellation mechanism to ReadAsync.  Instead of blocking in a call to read, we instead block in a call to poll, passing to it both the target pipe's descriptor along with a special descriptor we create and signal when the cancellation token is signaled.  I did not enable this for WriteAsync because the same approach either requires a very expensive change (writing one byte at a time) or is still potentially blocking (poll doesn't tell us whether there's enough room to write all of the data, just whether we can write something).

(I considered other approaches here, e.g. aio_read/aio_write/aio_cancel, io_submit, etc., but from everything I could find about these approaches, they were either too platform-specific, or didn't work with pipes, or would yield no better efficiency than what this approach is doing.  If folks have other suggestions, I'd be happy to consider alternative approaches.)


stephentoub:
@ianhays, I addressed your feedback, squashed down a bit, and pushed.  Any additional feedback?  Anyone else?

stephentoub:
@dotnet-bot test this please

Issue #2583, opened 2015-07-31
PfxTests.TestPrivateKey intermittent failure


nguerrera:
I can't repro it, but I got the following once:

```
 System.Security.Cryptography.X509Certificates.Tests.PfxTests.TestPrivateKey [FAIL]
               System.Security.Cryptography.CryptographicException : OpenCSP failed with error code 2148073494.
               Stack Trace:                  D:\Src\corefx\src\System.Security.Cryptography.Csp\src\System\Security\Cryptography\CapiHelper.cs(306,0): at Internal.NativeCrypto.CapiHelper.CreateProvHandle(CspParameters parameters, Boolean randomKeyContainer)
                  D:\Src\corefx\src\System.Security.Cryptography.Csp\src\System\Security\Cryptography\CapiHelper.cs(698,0): at Internal.NativeCrypto.CapiHelper.GetKeyPairHelper(CspAlgorithmType keyType, CspParameters parameters, Boolean randomKeyContainer, Int32 keySize, SafeProvHandle& safeProvHandle, SafeKeyHandle& safeKeyHandle)
                  D:\Src\corefx\src\System.Security.Cryptography.Csp\src\System\Security\Cryptography\RSACryptoServiceProvider.cs(89,0): at System.Security.Cryptography.RSACryptoServiceProvider.GetKeyPair()
                  D:\Src\corefx\src\System.Security.Cryptography.Csp\src\System\Security\Cryptography\RSACryptoServiceProvider.cs(69,0): at System.Security.Cryptography.RSACryptoServiceProvider..ctor(Int32 keySize, CspParameters parameters, Boolean useDefaultKeySize)
                  D:\Src\corefx\src\System.Security.Cryptography.Csp\src\System\Security\Cryptography\RSACryptoServiceProvider.cs(50,0): at System.Security.Cryptography.RSACryptoServiceProvider..ctor(CspParameters parameters)
                  D:\Src\corefx\src\System.Security.Cryptography.X509Certificates\src\Internal\Cryptography\Pal.Windows\CertificatePal.PrivateKey.cs(49,0): at Internal.Cryptography.Pal.CertificatePal.get_PrivateKey()
                  D:\Src\corefx\src\System.Security.Cryptography.X509Certificates\src\System\Security\Cryptography\X509Certificates\RSACertificateExtensions.cs(55,0): at System.Security.Cryptography.X509Certificates.RSACertificateExtensions.GetRSAPrivateKey(X509Certificate2 certificate)
[...]

PR #2955, opened 2015-08-24
Disable intermittently failing Windows X509 tests

bartonjs:


bartonjs:
Mitigating issues #2583 and #2885 by disabling those tests.  What they both have in common is an access of a private key which came from a PFX import (most other PFX imports don't actually read the key).

I vaguely feel like the ultimate fix is that the tests should to use a [collection fixture](http://xunit.github.io/docs/shared-context.html#collection-fixture) to isolate them from parallel execution; as it feels like the parallel PFX tests on the same certificate are causing the same private key blob to be utilized, leading to a shared resource deletion problem... but that's just speculation.

stephentoub:
LGTM

PR #2956, opened 2015-08-24
Enable AuthorityInformationAccess downloads for X509Chain on Unix.

### LINKED ISSUE #2273: Enable AuthorityInformationAccess downloads for X509Chain on Unix.
Windows will follow the AuthorityInformationAccess record and attempt to download any intermediate certificates as part of building the X509Chain, we should enable this for the Unix platforms as well, to maintain feature parity.

**Important**: Downloaded self-signed certificates are NOT considered trustworthy, and they need to make the chain report as not valid (unless root trust was bypassed).

Also important: Any testing for this should be [OuterLoop], because it would depend on the network.

bartonjs:
Use AuthorityInformationAccess to download intermediate CA certificates on Unix.

When the issuer certificate cannot be found in the locally understood stores, use the AIA record to extract the URL for the issuer certificate.  Then use libcurl to download it, and apply it into the chain candidates set.  If the chain then is considered successful save those certificates that had to be downloaded to the "CA" (intermediate) store for the current user.

The timeout value can be exceeded, but once it has been exceeded no further downloads will take place.  The same timeout will be used in the future for CRL/OCSP checks.

No unit tests were added for this feature because they would require both the network (downloading the files) and the filesystem (deleting a file to force it to download, then the product code causing it to be written).

As part of this change the libcurl interop files were broken up into chunks (_easy, _multi, and _slist) to cut down on the dependency chain that would have had to be imported by X509Certificates (even though it would have never gone down codepaths that would have used it).

Fixes #2273.


bartonjs:
cc: @AtsushiKan

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: @vijaykota has a PR with some changes to these files.  I believe you're just separating these out without any other changes, but you guys should coordinate to make sure we don't lose any changes in the process.

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_global.cs
```diff
@@ -0,0 +1,21 @@
...
+internal static partial class Interop
+{
+    internal static partial class libcurl
+    {
+        [DllImport(Interop.Libraries.LibCurl)]
+        public static extern int curl_global_init(
+            long flags);
+
+        [DllImport(Interop.Libraries.LibCurl)]
+        public static extern void curl_global_cleanup();
```
stephentoub: How do these differ from the same named functions in Interop.libcurl.cs? https://github.com/dotnet/corefx/pull/2956/files#diff-aea5728308b9f9f8961a847e7a60822dR15

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
```diff
@@ -0,0 +1,127 @@
...
+
+namespace Internal.Cryptography.Pal
+{
+    internal static class CertificateAssetDownloader
+    {
+        private static unsafe Interop.libcurl.curl_unsafe_write_callback s_writeCallback = CurlWriteCallback;
+
+        static CertificateAssetDownloader()
+        {
+            Interop.libcurl.curl_global_init(Interop.libcurl.CurlGlobalFlags.CURL_GLOBAL_ALL);
```
stephentoub: What happens if both this library and HttpClient call curl_global_init?  Eventually all of those will be called through our shims I believe (@nguerrera, correct?), but at least in the meantime do we need some sort of initialization mechanism like we have for OpenSSL?  Or can this be called any number of times and it's fine?

The documention for this is also pretty scary:
```
This function is not thread safe. You must not call it when any other thread in the program (i.e. a thread sharing the same memory) is running. This doesn't just mean no other thread that is using libcurl. Because curl_global_init calls functions of other libraries that are similarly thread unsafe, it could conflict with any other thread that uses these other libraries.
```

stephentoub:
> No unit tests were added for this feature because they would require both the network (downloading the files) and the filesystem (deleting a file to force it to download, then the product code causing it to be written).

How can we validate it then and make sure we don't regress it?

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_global.cs
bartonjs: Oh, this file shouldn't have been added, it was a split-out that I abandoned when there was still the version function remaining in libcurl.cs.  I'll remove it.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
bartonjs: The first initter seems to win, but if we're both calling it with the same flags then it's moot.

bartonjs:
@stephentoub I'm open to suggestion on the ongoing validation.  If [OuterLoop] seems sensible enough for this kind of operation (so it won't be destructive and maybe flaky on every build.cmd/build.sh invocation) then I can add something there.

CRL and OCSP checks are going to face a similar problem (and anything to do with chains is also subject to date-based failures, since certificates expire... though there are mitigations to that, too; like always using the current www.microsoft.com SSL cert by grabbing it from an SslStream)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -285,28 +324,38 @@ private static X509ChainStatusFlags MapVerifyErrorToChainStatus(Interop.libcrypt
...
+            ref TimeSpan remainingDownloadTime)
         {
-            X509Certificate2Collection candidates = new X509Certificate2Collection();
+            List<X509Certificate2> candidates = new List<X509Certificate2>();
             Queue<X509Certificate2> toProcess = new Queue<X509Certificate2>();
             toProcess.Enqueue(leaf);
 
-            using (var rootStore = new X509Store(StoreName.Root, StoreLocation.LocalMachine))
-            using (var intermediateStore = new X509Store(StoreName.CertificateAuthority, StoreLocation.LocalMachine))
+            using (var systemRootStore = new X509Store(StoreName.Root, StoreLocation.LocalMachine))
```
AtsushiKan: I think this has reached the point where it makes more sense to create a List<> of X509Store's and then use a loop to execute the same logic on all of them.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -380,6 +433,85 @@ private static X509ChainStatusFlags MapVerifyErrorToChainStatus(Interop.libcrypt
...
+            }
+
+            DerSequenceReader reader = new DerSequenceReader(authorityInformationAccess);
+
+            while (reader.HasData)
+            {
+                DerSequenceReader innerReader = reader.ReadSequence();
+
+                // If the sequence's first element is a sequence, unwrap it.
+                if (innerReader.PeekTag() == 0x30)
```
AtsushiKan: 0x30 - magic number - can we define a constant?

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
```diff
@@ -0,0 +1,237 @@
...
+                }
+                Debug.Assert((revents & PollFlags.POLLIN) != 0);
+                int pipeReadFd = _specialFds[libc.ReadEndOfPipe];
+                int bytesRead = 0;
+                unsafe
+                {
+                    do
+                    {
+                        // Read available data from the pipe
+                        int bufferLength = 1024;
```
AtsushiKan: Nit: Like to see "bufferLength" declared as "const"

src/Common/src/Interop/Unix/libcurl/Interop.libcurl_slist.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+using size_t = System.UInt64;
```
AtsushiKan: Seeing size_t declared as UInt64 in some files and IntPtr in others is making me nervous. Do we test (or support) both bitnesses on Unix?

stephentoub: Right now we're only targeting 64-bit.  By the time we support 32-bit as well (assuming we do), all of this should be smoothed over by our native shims.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                return null;
+            }
+
+            List<byte[]> dataPieces = new List<byte[]>();
+
+            using (Interop.libcurl.SafeCurlHandle curlHandle = Interop.libcurl.curl_easy_init())
+            {
+                GCHandle gcHandle = GCHandle.Alloc(dataPieces);
+                IntPtr dataHandlePtr = GCHandle.ToIntPtr(gcHandle);
```
AtsushiKan: The "try" should commence right after the GCHandle.Alloc(). Not that I expect ToIntPtr() to throw but people might add more code here in the future.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
```diff
@@ -0,0 +1,127 @@
...
+                Buffer.BlockCopy(piece, 0, data, offset, piece.Length);
+                offset += piece.Length;
+            }
+
+            return data;
+        }
+
+        private static unsafe ulong CurlWriteCallback(byte* buffer, ulong size, ulong nitems, IntPtr context)
+        {
+            ulong totalSize = size * nitems;
```
AtsushiKan: These unsafe casts and unchecked arithmetic combined with unmanaged memory copying is triggering my "Needs a security review" instinct. Even if it is a strictly internal function, could it be written without such red flags?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
```diff
@@ -380,6 +433,85 @@ private static X509ChainStatusFlags MapVerifyErrorToChainStatus(Interop.libcrypt
...
+                    authorityInformationAccess,
+                    ref remainingDownloadTime);
+
+                if (downloaded != null)
+                {
+                    downloadedCerts.Add(downloaded);
+
+                    return new List<X509Certificate2>(1) { downloaded };
+                }
+
```
AtsushiKan: Nit: Extra line.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
stephentoub: > then it's moot

Ok, but... what about the thread safety issues?

bartonjs: I think that their comment is "we're going to try to initialize our dependencies, like OpenSSL, if you're already using OpenSSL we don't really know what will happen; so we'll just slap a big don't be multithreaded warning on here... it'll probably all work fine, but if it doesn't you don't have any room to complain".

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: If it was capable of growing much more than this I'd probably agree; but I like having the explicit usings; and then all of their `Certificates` collections are already packed into an array and handled collectively past this point.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: Done

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
bartonjs: This is part of a non-semantic partitioning of Interop.libcurl.cs into separated files; it's not net new code.  (Though I do agree with you).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
bartonjs: Fair point, done.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
bartonjs: When @nguerrera's "shim everything" project gets to libcurl this will probably have a more managed-focused solution.  Until then, this is the signature that we need to match libcurl's callback.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509ChainProcessor.cs
bartonjs: Fixed

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
stephentoub: While that may be true, at the very least we need to ensure we're not invoking this function concurrently with itself, which we might be doing now if one thread used HttpClient at the same time another is using X509. We also need to ensure that we've properly initialized OpenSSL when using HttpClient. That all implies we need a native initialization shim for libcurl that both libraries can use, just as we do for SSL, and that shim will need to use our SSL initialization routine as well. Right?

nguerrera: Here is what @bartonjs, @ellismg, and I just discussed:

* System.Net.Http.Native.so [NEW]
  * exports thread-safe, idempotent EnsureLibCurlIsInitialized() [NEW]
  * requires caller to have already initialized OpenSsl (see below)

* System.Security.Cryptography.so  [EXISTING]
  * exports thread-safe, idempotent EnsureOpenSslIsInitialized [EXISTING]

Interop.Crypto.cctor -> calls EnsureOpenSslIsInitialized() [EXISTING]
Interop.Http.cctor -> calls EnsureOpenSslIsInitialized(); EnsureCurlIsIntialized() [NEW]

(And Interop.libcrypto and Interop.libcurl would do the same as Interop.Crypto and Interop.Http respective pending their elimination once everything is shimmed.)

@stephentoub does that make sense to you? I can work on this next to seed the curl shim and address this at the same time. 

One thing that got me worried just now is that it's unclear to me if the libcurl we use is guaranteed to be using the same libcrypto as us or if there could end up being two SxS libcryptos to initialize...


Interop.Crytp

stephentoub: Thanks, Nick (and Matt and Jeremy). That makes sense and is along the lines of what I was thinking.

vijaykota: Yes this is needed since multiple calls to curl_global_init are not safe

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
```diff
@@ -0,0 +1,237 @@
...
+                            numFds--;
+                        }
+                    }
+                }
+            }
+
+            internal void SignalFdSetChange(int fd, bool isRemove)
+            {
+                Debug.Assert(Monitor.IsEntered(this));
+                bool changed = isRemove ? _fdSet.Remove(fd) : _fdSet.Add(fd);
```
vijaykota: Please merge the latest code for this method

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
vijaykota: I will fix that in a later PR

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlSlistHandle.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+using size_t = System.IntPtr;
+using libc = Interop.libc;
+using libcurl = Interop.libcurl;
+using PollFlags = Interop.libc.PollFlags;
```
vijaykota: Nit: You should be able to get rid of some of these.

src/Common/src/Interop/Unix/libcurl/Interop.libcurl.cs
```diff
@@ -19,141 +19,6 @@ internal static partial class libcurl
         public static extern void curl_global_cleanup();
```
vijaykota: Nit: size_t and curl_socket_t are not needed

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                GCHandle gcHandle = GCHandle.Alloc(dataPieces);
+
+                try
+                {
+                    IntPtr dataHandlePtr = GCHandle.ToIntPtr(gcHandle);
+                    Interop.libcurl.curl_easy_setopt(curlHandle, Interop.libcurl.CURLoption.CURLOPT_URL, uri);
+                    Interop.libcurl.curl_easy_setopt(curlHandle, Interop.libcurl.CURLoption.CURLOPT_WRITEDATA, dataHandlePtr);
+                    Interop.libcurl.curl_easy_setopt(curlHandle, Interop.libcurl.CURLoption.CURLOPT_WRITEFUNCTION, s_writeCallback);
+
```
vijaykota: Depending on your security requirements, set CURLOPT_FOLLOWLOCATION if you want to automatically redirect.

vijaykota:
LGTM unless EnsureCurlIsIntialized is going to be part of this PR.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
bartonjs: Removed this cctor in favor of the System.Net.Http.Native initializer.

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlMultiHandle.cs
bartonjs: Done.

src/Common/src/Interop/Unix/libcurl/Interop.SafeCurlSlistHandle.cs
bartonjs: I trimmed down a good number of using aliases across the broken-out files.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CertificateAssetDownloader.cs
bartonjs: The files are signed, so they're self-validating, I've gone ahead and applied that; thanks.

bartonjs:
Pulled in changes, rebased and merged.  What was looked at before was https://github.com/dotnet/corefx/compare/master...bartonjs:aia_chain_reviewed?expand=1, I don't really know of a good way of diffing branches without a common reference point.

As far as the open question about a test; I talked with @joshfree about destructive testing, and I'll write one up to put in [OuterLoop] as a followup change.

bartonjs:
@dotnet-bot test this please

bartonjs:
@stephentoub gotcha, thanks.

PR #2958, opened 2015-08-24
Add coverage to DriveInfo tests

ianhays:
FileSystem.DriveInfo.dll coverage raised to at or near testable maximum. Added several scenarios that weren't present before and also split up Volume Label setting/getting to several tests so that errors can be better diagnosed.

### DIFF
--- a/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs
+++ b/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -12,8 +13,7 @@ namespace System.IO
     {
         private static string NormalizeDriveName(string driveName)
         {
-            if (driveName == null)
-                throw new ArgumentNullException("driveName");
+            Debug.Assert(driveName != null);
             Contract.EndContractBlock();
 
             string name;
--- a/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs
+++ b/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs
@@ -14,7 +14,6 @@ namespace System.IO
         private static string NormalizeDriveName(string driveName)
         {
             Debug.Assert(driveName != null);
-            Contract.EndContractBlock();
 
             string name;
 
--- a/src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Unix.Tests.cs
+++ b/src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Unix.Tests.cs
@@ -46,15 +46,15 @@ namespace System.IO.FileSystem.DriveInfoTests
             string invalidDriveName = "NonExistentDriveName";
             var invalidDrive = new DriveInfo(invalidDriveName);
 
-            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.AvailableFreeSpace; });
-            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.DriveFormat; });
+            Assert.Throws<DriveNotFoundException>(() =>invalidDrive.AvailableFreeSpace);
+            Assert.Throws<DriveNotFoundException>(() => invalidDrive.DriveFormat);
             Assert.Equal(DriveType.NoRootDirectory, invalidDrive.DriveType);
             Assert.False(invalidDrive.IsReady);
             Assert.Equal(invalidDriveName, invalidDrive.Name);
             Assert.Equal(invalidDriveName, invalidDrive.ToString());
             Assert.Equal(invalidDriveName, invalidDrive.RootDirectory.Name);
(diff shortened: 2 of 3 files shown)


ianhays:
cc: @stephentoub

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Windows.Tests.cs
```diff
@@ -126,53 +130,66 @@ public void TestValidDiskSpaceProperties()
...
+        }
 
-            // Set Volume Label - valid drive
-            string newLabel = "DriveInfoTests";
-            string originalLabel = validDrive.VolumeLabel;
-
-            if (validDrive.IsReady)
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public void SetVolumeLabel_Roundtrips()
```
ianhays: VolumeLabel setting like this does make me slightly nervous since we don't really know which volume we're messing with when we run the tests. The old test set it to a different name in a ```try``` and then set it back to the original in the corresponding ```finally``` - I switched it to this in the hopes that it will be more disaster-proof regardless of which volume is being modified.

src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs
```diff
@@ -12,8 +13,7 @@ public sealed partial class DriveInfo
     {
         private static string NormalizeDriveName(string driveName)
         {
-            if (driveName == null)
-                throw new ArgumentNullException("driveName");
+            Debug.Assert(driveName != null);
```
stephentoub: You can get rid of the Contract.EndContractBlock() as well.  Even if we were using contracts, EndContractBlock only applies after certain if/throw patterns.

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Unix.Tests.cs
```diff
@@ -41,27 +41,46 @@ public void TestGetDrives()
...
+
+            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.AvailableFreeSpace; });
+            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.DriveFormat; });
             Assert.Equal(DriveType.NoRootDirectory, invalidDrive.DriveType);
+            Assert.False(invalidDrive.IsReady);
+            Assert.Equal(invalidDriveName, invalidDrive.Name);
+            Assert.Equal(invalidDriveName, invalidDrive.ToString());
+            Assert.Equal(invalidDriveName, invalidDrive.RootDirectory.Name);
+            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.TotalFreeSpace; });
+            Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.TotalSize; });
```
stephentoub: We shouldn't need to store the results in locals, e.g. instead of:
```C#
Assert.Throws<DriveNotFoundException>(() => { var invalid = invalidDrive.TotalSize; });
```
you should be able to do:
```C#
Assert.Throws<DriveNotFoundException>(() => invalidDrive.TotalSize);
```

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Windows.Tests.cs
stephentoub: I like this new version better, too, thanks.

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Unix.Tests.cs
```diff
@@ -41,27 +41,46 @@ public void TestGetDrives()
...
+            Assert.True(root.AvailableFreeSpace > 0);
+            var format = root.DriveFormat;
             Assert.Equal(DriveType.Fixed, root.DriveType);
             Assert.True(root.IsReady);
-            Assert.True(root.AvailableFreeSpace > 0);
+            Assert.Equal("/", root.Name);
+            Assert.Equal("/", root.ToString());
+            Assert.Equal("/", root.RootDirectory.FullName);
             Assert.True(root.TotalFreeSpace > 0);
             Assert.True(root.TotalSize > 0);
```
stephentoub: You could use probably use Assert.InRange for these, if you wanted.

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Windows.Tests.cs
```diff
@@ -116,6 +115,11 @@ public void TestValidDiskSpaceProperties()
                 if (fbUser != drive.AvailableFreeSpace)
                     Assert.True(drive.AvailableFreeSpace >= 0);
 
+                // valid property getters shouldn't throw
+                string name = drive.Name;
+                string format = drive.DriveFormat;
```
stephentoub: Anything we can validate about ```format```?

stephentoub:
Some comments, but otherwise LGTM.

src/System.IO.FileSystem.DriveInfo/tests/DriveInfo.Windows.Tests.cs
ianhays: We could make a manual interop call to statfs/GetVolumeInformation and then compare the result to what is returned from drive.DriveFormat, but that would just be comparing the same call to itself. 

Beyond that I don't know of anything consistent that we could test for since we don't have any solid information on the drive we're testing other than the fact that it is valid.

stephentoub:
LGTM

PR #2959, opened 2015-08-24
Consistently Throw NotSupportedException on Enumerator Reset

### LINKED ISSUE #2879: Would changing Where().GetEnumerator().Reset()'s exception break anything?
Calling `Reset()` on an `Enumerable.Itereator` currently throws `NotImplementedException`.

This strikes me as a bug in itself; almost all other cases of `Reset()` which throw, including those that are compiler-generated, throw `NotSupportedException`. The compiler-generated iterators are defined as doing so, and the documentation for `Reset` would suggest that it one should expect `NotSupportedException` or `InvalidOperationException` (for an enumerator that does support `Reset()` but is in a state where it cannot) though that's a suggestion inferred from the remarks rather than an explicit list of Exceptions.

That in itself would seem to be a reason to change it, but I also am now finding it's getting in the way of wanting to experiment with perhaps using the class in other cases; I either need to change it, need to change other methods to result in it (more likely to be breaking, I imagine) or need to have overrides as appropriate.

This last option is not arduous, but just changing the exception thrown seems more sensible.

But all that said, the current code seems very deliberate in its choice, with the use of `NotImplemented.ByDesign` suggesting that while it may have been a mistake (or not, for some reason I don't see) it's now something compatibility is being kept with.

So, is there such a reason, or can we just change it to `NotSupportedException`?

JonHanna:
Fixes #2879

Have Enumerable.Iterator<TSource>'s implementation of IEnumerator.Reset throw
NotSupportedException instead of NotImplementedException as both more
appropriate in itself, and more consistent with other IEnumerator.Reset
implementations.

### DIFF
--- a/src/System.Linq/src/System.Linq.csproj
+++ b/src/System.Linq/src/System.Linq.csproj
@@ -16,9 +16,6 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="$(CommonPath)\System\NotImplemented.cs">
-      <Link>Common\System\NotImplemented.cs</Link>
-    </Compile>
     <Compile Include="System\Linq\Enumerable.cs" />
     <Compile Include="System\Linq\Errors.cs" />
   </ItemGroup>
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -139,7 +139,7 @@ namespace System.Linq
 
             void IEnumerator.Reset()
             {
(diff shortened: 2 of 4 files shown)


stephentoub:
cc: @VSadov

stephentoub:
If this is a change we want to make, it LGTM.  @VSadov, do you want to change the exception employed here?

VSadov:
Compiler-generated Reset throws NotSupportedException, so it makes sense for a hand-written substitutes to throw that as well.

I do not think there is a compat issue here since calling Reset on iterators is rare, dependency on a particular exception thrown even more so, in particular since LINQ operators implemented as C# iterators (most of the operators) throw NSE.

VSadov:
LGTM

PR #2960, opened 2015-08-25
Add unit tests for System.Reflection.Module class.

Sridhar-MS:


src/System.Reflection/tests/Module/ModuleTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            var customAttrs = module.CustomAttributes.GetEnumerator();
+
+            CustomAttributeData attribute = null;
+            while (customAttrs.MoveNext())
+            {
+                if (customAttrs.Current.AttributeType.Equals(attrType))
+                {
+                    attribute = customAttrs.Current;
+                }
+            }
```
stephentoub: Not a big deal, but any reason not to use foreach or LINQ for this kind of thing? e.g.
```C#
CustomAttributeData attribute = module.CustomAttributes.FirstOrDefault(a => a.AttributeType.Equals(attrType));
```
instead of:
```C#
var customAttrs = module.CustomAttributes.GetEnumerator();
CustomAttributeData attribute = null;
while (customAttrs.MoveNext())
{
    if (customAttrs.Current.AttributeType.Equals(attrType))
    {
        attribute = customAttrs.Current;
    }
}
```

src/System.Reflection/tests/ParameterInfo/ParamInfo_PropertyTests.cs
```diff
@@ -259,6 +259,41 @@ public static void TestIsRetval2()
...
+            Assert.True(pi.Attributes.HasFlag(ParameterAttributes.Out));
+            Assert.True(pi.Attributes.HasFlag(ParameterAttributes.In));
+        }
+
+        [Theory]
+        [InlineData(typeof(OptionalAttribute))]
+        [InlineData(typeof(MarshalAsAttribute))]
+        [InlineData(typeof(OutAttribute))]
+        [InlineData(typeof(InAttribute))]
+        public static void CustomeAttributesTest(Type attrType)
```
stephentoub: Nit: Custome => Custom

src/System.Reflection/tests/Assembly/Assembly_EmbeddedFiles.cs
```diff
@@ -16,6 +16,10 @@ public class EmbeddedFilesTest
         [Fact]
         public void TestEmbeddedFiles()
         {
+            var resources = typeof(EmbeddedFilesTest).GetTypeInfo().Assembly.GetManifestResourceNames();
+
+            Assert.Contains("EmbeddedImage.png", resources);
+
```
stephentoub: In the existing code below, should we be verifying that we can read from the stream and verifying something about what we read?  Or is that handled elsewhere, or unnecessary for some reason?

src/System.Reflection/tests/Module/ModuleTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            Assert.NotNull(attribute);
+
+            Assert.True(attribute.ConstructorArguments.Count == 1);
+            Assert.True(attribute.NamedArguments.Count == 1);
```
stephentoub: Assert.Equal instead of Assert.True?

src/System.Reflection/tests/ParameterInfo/ParamInfo_PropertyTests.cs
```diff
@@ -259,6 +259,41 @@ public static void TestIsRetval2()
...
+            CustomAttributeData attribute = null;
+            while (customAttrs.MoveNext())
+            {
+                if (customAttrs.Current.AttributeType.Equals(attrType))
+                {
+                    attribute = customAttrs.Current;
+                }
+            }
+
+            Assert.NotNull(attribute);
```
stephentoub: Same LINQ question as earlier.

stephentoub:
Some comments/questions, but otherwise LGTM.

src/System.Reflection/tests/Module/ModuleTests.cs
Sridhar-MS: These are remnants from code that I reused for these tests. Fixed it now to use linq.

src/System.Reflection/tests/ParameterInfo/ParamInfo_PropertyTests.cs
Sridhar-MS: Fixed

src/System.Reflection/tests/Module/ModuleTests.cs
Sridhar-MS: Fixed.

src/System.Reflection/tests/Assembly/Assembly_EmbeddedFiles.cs
Sridhar-MS: Reading the stream of an embedded resource is getting verified in one of the tests I wrote earlier - https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/tests/ResourceAssemblyLoadContext.cs#L30

PR #2964, opened 2015-08-25
Issue#2930: Make GetRSAPrivateKey prefer CNG

### LINKED ISSUE #2930: Change Windows GetRSAPrivateKey to prefer CNG
When GetRSAPrivateKey was added into the repository the CNG types weren't yet online.  Now that they are, we should have GetRSAPrivateKey and GetRSAPublicKey use CNG when able, and fallback to CAPI for legacy support.

AtsushiKan:
This brings these two extension methods into the age
of CNG:

- GetRSAPublicKey() - now returns an RSACng object
  rather than an RSACryptoServiceProvider object.

- GetRSAPrivateKey() - returns an RSACng object
  if a stored CNG key is available for the certificate's
  public key. Prior behavior was that it blindly
  asked CAPI to open a CNG KSP name and key name, and
  dumped the resulting CryptoException into the caller's
  lap.

Fixes #2930.


AtsushiKan:
cc @bartonjs

stephentoub:
LGTM

PR #2971, opened 2015-08-25
Harden implementation of Vector<T>.Count

mellinoe:
The Vector<T>.Count property, as currently implemented, does not work when called from a Debug version of a user program running against the release build of System.Numerics.Vectors.dll. Because the user code is unoptimized, the IL version of Vector<T>.Count is run, but since System.Numerics.Vectors.dll is optimized, the internal call to Vector.IsHardwareAccelerated returns true, leading to the NotSupportedException, which was intended to only be thrown when called via reflection. This essentially makes it impossible to call the method in Debug-mode user code, when running with RyuJIT.

This change hardens the implementation, fixing the above issue. In addition, it also lets users call the method via reflection (if they wanted to), and also has the nice effect of allowing its use in the
Debugger.

We use a simple helper struct which has two fields, the Vector<T> whose size we want to measure, followed a byte field. We just grab the address of the byte field, and compute its difference from the address of the vector's first byte element. We then divide that total vector size by the size of T. This should be robust to future changes, and should accomodate differing Vector sizes across SIMD instruction sets. Since we are just relying on the layout of the struct, the result should remain correct in the future. New SIMD register sizes will dictate a new size for Vector<T>, and therefore a new layout for our helper structure. The cost for computing this is once per generic type instantiation; the value is cached. I've tested that this performs as expected in a machine with pre-AVX2 support, and with AVX2 support.

I've added a test case that simply verifies the value returned from calling Vector<T>.Count matches the value returned from a regular property access.

I also removed the x86-workaround we had in place; it seems like the latest version of CoreCLR has the JIT fixes in it that were needed to run the tests on the 64-bit runtime.

/cc: @CarolEidt

### DIFF
--- a/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs
+++ b/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs
@@ -53,17 +53,10 @@ namespace System.Numerics
         {
             get
             {
-                if (Vector.IsHardwareAccelerated)
-                {
-                    throw new NotSupportedException(SR.Reflection_MethodNotSupported);
-                }
-                else
-                {
-                    return count;
-                }
+                return s_count;
             }
         }
-        private static int count = InitializeCount();
+        private static int s_count = InitializeCount();
 
         /// <summary>
         /// Returns a vector containing all zeroes.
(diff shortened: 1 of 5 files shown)


CarolEidt:
LGTM

stephentoub:
LGTM2

PR #2976, opened 2015-08-26
Fix several System.Diagnostics.Tracing.Telemetry issues

stephentoub:
Source:
- Fixes bug where disposal of an AllListeners subscription would fail if the subscription wasn't the most recent.  The linked list walk was looking for the wrong object.
- Fixes bug where double disposal of a subscription to AllListeners would result in OnCompleted for that observer being called multiple times, and in some cases resulting in a NullReferenceException.
- Fixes bug where double disposal of a subscription to a TelemetryListener or to AllListeners would trigger a Debug.Assert
- Marks several fields that could be readonly as such.

Tests:
- Fixes a build warning in the tests project caused by a missing reference in the project.json to the correct version of System.Globalization.
- Changes the locked status of the project.lock.json from false to true so that it's not edited on dev machines any time a build is done.
- Brings line code coverage of System.Diagnostics.Tracing.Telemetry up to 100%.

cc: @vancem

### DIFF
--- a/src/System.Diagnostics.Tracing.Telemetry/src/System/Diagnostics/Tracing/TelemetryListener.cs
+++ b/src/System.Diagnostics.Tracing.Telemetry/src/System/Diagnostics/Tracing/TelemetryListener.cs
@@ -229,7 +229,7 @@ namespace System.Diagnostics.Tracing
             {
                 if (subscriptions == null)
                 {
-                    Debug.Assert(false, "Could not find subscription");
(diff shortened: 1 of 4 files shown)


vancem:
Thanks Stephen.   In fact your code is basically the same as code I prepped code for the double-dispose issue.    The bug in Remove() was a good catch.  I was hunting down a bug in the Stress test for AllListeners that is fixed by it.  

Looks good to me.

Issue #2962, opened 2015-08-25
Needless calls to comparer in EnumerableSorter


epsitec:
I was surprised to see that when sorting items using `System.Linq.Enumerable.OrderBy<TSource, TKey>`, while specifying a comparer:

```c#
var sort = new MySort();
var sorted = items.OrderBy(x => x, sort);
...
class MySort : IComparer<Foo>
{
    public int Compare(Foo x, Foo y) { ... }
}
```

the implementation of `EnumerableSorter<T>.QuickSort` happens to call `MySort.Compare` for comparisons which are bound to result to equality, i.e. where arguments `x` and `y` refer to the same object.

I can't see why this does not get caught earlier, either in the Quick Sort implementation, or in the call to `CompareKeys`, where `0` could be returned immediately if `index1` equals `index2`:

![enumerable](https://cloud.githubusercontent.com/assets/3872435/9463960/8e2d3102-4b20-11e5-9990-df0966cfb9c4.png)

https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Enumerable.cs#L3363

Can this design be altered to avoid useless calls to compare objects to themselves?

epsitec:
Sorting `[ "A", "C", "B" ]` results in a total of six (6) comparisons, three (3) of them being useless:

  * C, A
  * _C, C_
  * C, B
  * _A, A_
  * A, B
  * _A, A_

JonHanna:
It's common enough with quicksort generally, since the number of comparisons tends to be proportional to `n log n` (more precisely, if we consider that as meaning `k * n log n` comparisons, then the average `k` depends on partition strategy; around `1.386` for random and probably less good here were the strategy is just to pick the middle, in any case `k` is always greater than `1`)  while the number of comparisons with self tends to be close to `n`.

And as such if the comparison itself is cheap, the `n log n * k` tests for identity of comparands could outweight the `n` times it allows a short-circuit.

Before: `n log n * k * cost-of-comparison`
After: `(n log n * k * cost-of-comparison) - (n * cost-of-identity-test)`

When we add in the fact that comparers very often have themselves a short-circuit on identity, this means that as a general strategy, just allowing the compare-with-self to happen is generally the best approach. (I note though, that `GenericComparer<T>` does not short-circuit on that directly, though sometimes the implementation it calls into does, perhaps considering that such a check would include the `x == null && y == null` possibility that it has to checked for, that would be a worthwhile change in itself).

[...]

PR #2977, opened 2015-08-26
Short-circuit key comparison on equal indexes in EnumerableSorter

epsitec:
Useless calls to `comparer.Compare` should be avoided: when both `index1` and `index2` refer to the same key, there is no need to execute the recursive `CompareAnyKeys` method.

See full discussion in issue #2962.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -3289,7 +3289,7 @@ namespace System.Linq
     {
         internal abstract void ComputeKeys(TElement[] elements, int count);
 
-        internal abstract int CompareKeys(int index1, int index2);
+        internal abstract int CompareAnyKeys(int index1, int index2);
 
         internal int[] Sort(TElement[] elements, int count)
         {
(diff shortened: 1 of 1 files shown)


jasonwilliams200OK:
The same class with signature `internal abstract class EnumerableSorter<TElement>` also exists in [/src/System.Xml.XDocument/src/System/Linq/Enumerable.cs](https://github.com/dotnet/corefx/blob/7906323e4/src/System.Xml.XDocument/src/System/Linq/Enumerable.cs). Should this fixed be applied there as well? Or better should it be moved to a common place (common dependency) so both assemblies can share the definition?

stephentoub:
cc: @VSadov

JonHanna:
During the discussion of that issue, I experimented with this change. I found that with a simple comparison it was a slight improvement, but with long `.ThenBy` chains it quickly becomes a more significant improvement. I couldn't find a case where it made anything worse.

epsitec:
Obviously, the gains of this fix will be much higher if the user-provided `Compare` function  takes time, does not include the usual test for object reference equality, or if the generic `TElement` on which the code operates is a larger `struct`.

VSadov:
LGTM.

VSadov:
entered a separate bug about same change in System.Xml.XDocument - #3008

stephentoub:
LGTM

PR #2978, opened 2015-08-26
Simplify ObjectToIdCache.GetPrime

JonHanna:
Because ObjectToIdCache always attempts a starting size of 1, resulting
in a starting size of 3 (first prime in list) all further possible
values can be pre-computed directly.

This simplifies the GetPrime() code.

This also allows the possibility of the cache working at a capacity greater than
1535867969, which is the point at which it would currently overflow (after trying
to double 1535867962 which overflows on 32-bit signed integers). This is neither
likely to happen, nor likely to be allowed by memory size, but that case now tries
the largest prime that is an allowed array size (viz 2146435069) then the largest
array size (ignoring primality), at which point it is at the theoretical limit
of the approach ObjectToIdCache takes, if not already long past the practical
limits of memory.

It's an unlikely case, but being robust in the face of it is better than not being robust in
the face of it, unless there's a downside I don't see.

### DIFF
--- a/src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/ObjectToIdCache.cs
+++ b/src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/ObjectToIdCache.cs
@@ -111,7 +111,7 @@ namespace System.Runtime.Serialization
 
         private void Rehash()
         {
-            int size = GetPrime(m_objs.Length * 2);
+            int size = GetPrime(m_objs.Length + 1); // The lookup does an inherent doubling
             int[] oldIds = m_ids;
             object[] oldObjs = m_objs;
             m_ids = new int[size];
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @khdang, @ellismg

src/System.Private.DataContractSerialization/src/System/Runtime/Serialization/ObjectToIdCache.cs
```diff
@@ -171,6 +150,9 @@ private static bool IsPrime(int candidate)
         {
             3, 7, 17, 37, 89, 197, 431, 919, 1931, 4049, 8419, 17519, 36353,
             75431, 156437, 324449, 672827, 1395263, 2893249, 5999471,
+            11998949, 23997907, 47995853, 95991737, 191983481, 383966977, 767933981, 1535867969,
+            2146435069, 0X7FEFFFFF
```
khdang: @JonHanna Why these two numbers are after 1535867969? Isn't it only 2147483423, which is the largest prime that is less than Int32.MaxValue?

JonHanna: @khdang  0X7FEFFFFF is the largest allowable array (that isn't of an element that is a single byte in size), and 2146435069 is the largest prime smaller than it, so having followed the path the previous code would have taken, and when it will then overflow, I use 2146435069 as a primal last-ditch, and 0X7FEFFFFF as the ultimate last-ditch (giving up on the advantages of prime modulos). After that there's nowhere left to go.

khdang:
I have one comment. Other than that LGTM.

stephentoub:
LGTM (I also computed the sequence and got the same numbers)

PR #2980, opened 2015-08-26
Fully import OrderBy/Descending & ThenBy/Descending tests

JonHanna:
Convert OrderBy, ThenBy, OrderByDescending & ThenByDescending tests imported
in #2874 into a more directly XUnit-ian style

OrderBy8 and OrderByDescending8 seemed to have been adapted from tests that
checked stability of sort, to tests that imposed stability of sort external
to OrderBy and then tested it. This is apparently for use with PLINQ where
OrderBy etc. are not specified as providing stable sorts. Since in Linq to
objects they ARE specified as providing stable sorts, these have been returned
to relying just on OrderBy/OrderByDescending for that stability.


JonHanna:
@dotnet-bot please test this, if you'd be so kind.

JonHanna:

stephentoub:
@JonHanna, this doesn't look like a spurious failure... these two tests failed:
```
10:56:14    System.Linq.Tests.OrderByDescendingTests.FirstAndLastAreDuplicatesNullPassedAsComparer [FAIL]
10:56:15       Assert.Equal() Failure
10:56:15       Expected: String[] ["Prakash", "Prakash", "DAN", "dan", "Alpha"]
10:56:15       Actual:   OrderedEnumerable<String, String> ["dan", "Prakash", "Prakash", "DAN", "Alpha"]
10:56:15       Stack Trace:
10:56:15             at System.Linq.Tests.OrderByDescendingTests.FirstAndLastAreDuplicatesNullPassedAsComparer()
10:56:15 Starting:    System.Linq.Expressions.Tests
10:56:17    System.Linq.Tests.OrderByTests.FirstAndLastAreDuplicatesNullPassedAsComparer [FAIL]
10:56:17       Assert.Equal() Failure
10:56:17       Expected: String[] ["Alpha", "dan", "DAN", "Prakash", "Prakash"]
10:56:17       Actual:   OrderedEnumerable<String, String> ["Alpha", "DAN", "Prakash", "Prakash", "dan"]
10:56:17       Stack Trace:
10:56:17             at System.Linq.Tests.OrderByTests.FirstAndLastAreDuplicatesNullPassedAsComparer()
```

JonHanna:
Hmm. Was my simplifying to use `StringComparer.OrdinalIgnoreCase` not as portable as I thought, perhaps?

stephentoub:
Globalization is one of the areas we're still making active progress on for Unix (cc: @ellismg).  There are still some gaps, in particular around string collation.

JonHanna:
Nope, not my use of `StringComparer.OrdinalIgnoreCase` as I thought at first, it's the tests that didn't use them that are failing.

JonHanna:
Got it:

1. Test machine that is failing is using code-point order as the default ordering for strings.
2. The test doesn't expect this.
3. The old test code uses a order-agnostic comparison to check the results, so it didn't fail before.

That third point is a bit of a flaw in the tests as written.

I'll replace these with a numeric test.

src/System.Linq/tests/OrderByDescendingTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                new Record{ Name = "Prakash", Score = 99 },
+                new Record{ Name = "Alpha", Score = 90 },
+                new Record{ Name = "Bob", Score = 90 },
+                new Record{ Name = "Robert", Score = 45 },
+                new Record{ Name = "Thomas", Score = 45 },
+                new Record{ Name = "Tim", Score = 45 },
+                new Record{ Name = "Mark", Score = 45 },
+            };
+
+            Assert.Equal(expected, source.OrderByDescending(e => e.Score));
```
stephentoub: Was the ThenBy removed from the original because the test is duplcated somewhere?

stephentoub:
Other than my one question, LGTM.  It does look like maybe we're missing stable ordering tests for ThenBy/ThenByDescending?

src/System.Linq/tests/OrderByDescendingTests.cs
JonHanna: ThenBy was removed to make the test actually check stability.

To judge from the comments, I believe what happened was:

1. The test was equivalent to how I have it now. (The equivalent is included and commented out in the old file), and tested stability.
2. When adopted for PLINQ, the test failed, which is fine as PLINQ's OrderBy doesn't promise stability.
3. Rather than just remove this test, it was changed to coerce the results into a stable order by imposing a total ordering outside of `OrderBy` with a combination of Select and ThenBy. At this point the results were stable, but the test code made them stable rather than the tested code, so linq-to-object's OrderBy's stability (which is specified) was no longer tested for.
4. Due to problems with other PLINQ code not being stable in ordering, `Verification.Allequal` was changed to be order-insensitive, making most of these tests pointless anyway. (Also why the earlier version of what I call `FirstAndLastAreDuplicatesNullPassedAsComparer` wasn't failing; it wasn't really checked).

Removing the `Select` and `ThenBy` turns this back into a test for stability.

JonHanna:
While we already were missing a true stability test on `ThenBy`, as per my note above, adding one in seems like a reasonable idea.

stephentoub:
Thanks for the explanation.  We can add one separately.

PR #2987, opened 2015-08-27
Change file system APIs to implicitly add extended syntax when needed

JeremyKuhne:
Update the file system APIs to use extended syntax when needed.

@stephentoub, @weshaggard, @ellismg

### DIFF
--- a/src/Common/src/Interop/Windows/mincore/Interop.CreateFile.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.CreateFile.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Win32.SafeHandles;
 using System;
+using System.IO;
 using System.Runtime.InteropServices;
 
 internal partial class Interop
@@ -10,7 +11,7 @@ internal partial class Interop
     internal partial class mincore
     {
         [DllImport(Libraries.CoreFile_L1, EntryPoint = "CreateFileW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal static extern SafeFileHandle CreateFile(
+        private static extern SafeFileHandle CreateFilePrivate(
             string lpFileName,
             int dwDesiredAccess,
             System.IO.FileShare dwShareMode,
@@ -18,5 +19,18 @@ internal partial class Interop
             System.IO.FileMode dwCreationDisposition,
             int dwFlagsAndAttributes,
             IntPtr hTemplateFile);
+
+        internal static SafeFileHandle CreateFile(
+            string lpFileName,
+            int dwDesiredAccess,
+            System.IO.FileShare dwShareMode,
+            [In] ref SECURITY_ATTRIBUTES securityAttrs,
+            System.IO.FileMode dwCreationDisposition,
+            int dwFlagsAndAttributes,
+            IntPtr hTemplateFile)
+        {
+            lpFileName = PathInternal.AddExtendedPathPrefixForLongPaths(lpFileName);
+            return CreateFilePrivate(lpFileName, dwDesiredAccess, dwShareMode, ref securityAttrs, dwCreationDisposition, dwFlagsAndAttributes, hTemplateFile);
+        }
     }
 }
--- a/src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Win32.SafeHandles;
+using System.IO;
 using System.Runtime.InteropServices;
 
 internal partial class Interop
@@ -13,6 +14,8 @@ internal partial class Interop
 
         internal static bool MoveFile(string src, string dst)
         {
+            src = PathInternal.AddExtendedPathPrefixForLongPaths(src);
+            dst = PathInternal.AddExtendedPathPrefixForLongPaths(dst);
             return MoveFileEx(src, dst, 2 /* MOVEFILE_COPY_ALLOWED */);
         }
     }
--- a/src/Common/src/System/IO/PathInternal.Windows.cs
+++ b/src/Common/src/System/IO/PathInternal.Windows.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Diagnostics.Contracts;
+using System.Diagnostics;
 using System.Text;
 
 namespace System.IO
@@ -13,6 +13,7 @@ namespace System.IO
         internal const string UncPathPrefix = @"\\";
         internal const string UncExtendedPrefixToInsert = @"?\UNC\";
         internal const string UncExtendedPathPrefix = @"\\?\UNC\";
+        internal const string DevicePathPrefix = @"\\.\";
         internal const int MaxShortPath = 260;
         internal const int MaxShortDirectoryPath = 248;
         internal const int MaxExtendedPath = short.MaxValue;
@@ -70,11 +71,27 @@ namespace System.IO
         }
 
         /// <summary>
-        /// Adds the extended path prefix (\\?\) if not already present.
+        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative,
+        /// AND the path is more than 259 characters. (> MAX_PATH + null)
+        /// </summary>
+        internal static string AddExtendedPathPrefixForLongPaths(string path)
+        {
+            if (path != null && path.Length >= MaxShortPath)
+            {
+                return AddExtendedPathPrefix(path);
+            }
+            else
+            {
+                return path;
+            }
+        }
+
+        /// <summary>
+        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative.
         /// </summary>
         internal static string AddExtendedPathPrefix(string path)
         {
-            if (IsExtended(path))
+            if (IsExtended(path) || IsPathRelative(path) || IsDevicePath(path))
                 return path;
 
             // Given \\server\share in longpath becomes \\?\UNC\server\share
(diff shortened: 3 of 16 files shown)


src/Common/src/Interop/Windows/mincore/Interop.CreateFile.cs
```diff
@@ -18,5 +19,18 @@ internal partial class mincore
             System.IO.FileMode dwCreationDisposition,
             int dwFlagsAndAttributes,
             IntPtr hTemplateFile);
+
+        internal static SafeFileHandle CreateFile(
```
weshaggard: For my information what happens if you get a FileHandle to a long path file? If you query the full path for it will it have the extended path prefix?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -227,6 +239,37 @@ internal static int GetRootLength(string path)
...
+        internal static bool IsPathRelative(string path)
+        {
+            if (path.Length < 2)
+            {
+                // It isn't fixed, it must be relative.  There is no way to specify a fixed
+                // path with one character (or less).
+                return true;
+            }
+
+            if ((path[0] == '\\') || (path[0] == '/'))
```
weshaggard: I assume you want to use some of the existing helpers like IsDirectorySeparator and other constants like Path.VolumeSeparatorChar etc.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -227,6 +239,37 @@ internal static int GetRootLength(string path)
...
         }
 
+        /// <summary>
+        /// Returns true if the path specified is relative to the current drive or working directory.
+        /// Returns false if the path is fixed to a specific drive or UNC path.  This method does no
+        /// validation of the path (URIs will be returned as relative as a result).
+        /// </summary>
+        /// <remarks>
+        /// Handles paths that use the alternate directory separator.  It is a frequent mistake to
+        /// assume that rooted paths (Path.IsPathRooted) are not relative.  This isn't the case.
```
weshaggard: Interesting... I wonder if we should have another public API this IsPathRelative method.

src/Common/tests/Tests/System/IO/PathInternal.Windows.Tests.cs
```diff
@@ -9,11 +9,31 @@ public class PathInternal_Windows_Tests
...
     [PlatformSpecific(PlatformID.Windows)]
     public void AddExtendedPathPrefixTest(string path, string expected)
     {
         Assert.Equal(expected, PathInternal.AddExtendedPathPrefix(path));
     }
+
+    [Theory,
+        InlineData("", true),
+        InlineData("C:", true),
+        InlineData("**", true),
```
weshaggard: Just for completeness it is probably worth adding cases for "\" and "/".

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: What's an example of a non-rooted path that's not relative?

weshaggard: From looking at the code a drive relative path would be an example. Something like "C:test"

stephentoub: It'd be good to be specific about that in the comment.

src/Common/tests/Tests/System/IO/PathInternal.Windows.Tests.cs
JeremyKuhne: Sure

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I'll call the "C:" case out explicitly. @weshaggard I think we should have this API public. The Unix impl would be "starts with \"?

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I was doing this for perf. I assume the compiler won't inline?

src/Common/src/Interop/Windows/mincore/Interop.CreateFile.cs
JeremyKuhne: I'll check what happens for short and long paths. I don't think the handle is connected in any way with the original query.

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: > I'll call the "C:" case out explicitly

Thanks

> The Unix impl would be "starts with \"?

You mean "not starts with /", right?  In other words, it's relative iff it's not rooted.

JeremyKuhne: Yes, relative doesn't start with \ (on Unix)

stephentoub: Relative could start with '\'... it can't start with '/'.  "\" is even a valid file name.

JeremyKuhne: Doh- yeah. Sorry- too much Windows in my head at the moment. Thanks for clarifying.

src/Common/src/Interop/Windows/mincore/Interop.CreateFile.cs
JeremyKuhne: GetFinalPathNameByHandle doesn't seem to care about the original format of the string. I'll spend more time to validate all of the various handle scenarios, but I'm pretty sure that Windows doesn't associate the passed in string with handles.

JeremyKuhne:
cc: @Priya91

JeremyKuhne:
@stephentoub, @weshaggard I've updated all of the ones used in FileSystem. Should be ready to merge. I'll unblock the length restrictions next (and fix/add tests).

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -70,11 +71,27 @@ internal static bool IsDirectoryTooLong(string fullPath)
...
+            }
+        }
+
+        /// <summary>
+        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative.
         /// </summary>
         internal static string AddExtendedPathPrefix(string path)
         {
-            if (IsExtended(path))
+            if (IsExtended(path) || IsPathRelative(path) || IsDevicePath(path))
```
JeremyKuhne: I realized I wasn't allowing for device paths- we should never prepend those

src/Common/src/System/IO/PathInternal.Windows.cs
weshaggard: We should only do this for perf if we know that there is a problem, otherwise we should go for readability and consistency. At a minimum we can use the constants so that intentions are clearer.

JeremyKuhne: I'll change to constants then.

src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
```diff
@@ -13,6 +14,8 @@ internal partial class mincore
 
         internal static bool MoveFile(string src, string dst)
         {
+            src = PathInternal.AddExtendedPathPrefixForLongPaths(src);
+            dst = PathInternal.AddExtendedPathPrefixForLongPaths(dst);
```
stephentoub: Nit: Why is this P/Invoke called SomethingEx and the others are SomethingPrivate?

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -70,11 +71,27 @@ internal static bool IsDirectoryTooLong(string fullPath)
         }
 
         /// <summary>
-        /// Adds the extended path prefix (\\?\) if not already present.
+        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative,
+        /// AND the path is more than 259 characters. (> MAX_PATH + null)
+        /// </summary>
+        internal static string AddExtendedPathPrefixForLongPaths(string path)
```
stephentoub: Nit: Personally, I'd prefer it if the name of the method made it clear that it might not add anything, e.g. AddLongPathExtendedPrefixIfNecessary.

stephentoub: (Reading it again, I think that might be what you were going for with "ForLongPaths".  FWIW, I didn't interpret it to mean "will add the prefix only if it's necessary to support a long path" but rather "always add the prefix that's used with long paths".)

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: Does anyone call this function beyond the above one that does it conditionally?  If not, should this be private?  It's also strange to me that we have an "Add" method that implies it adds something unconditionally but then has all of these conditions at the beginning.  Folding it into the "IfNecessary" version would help with that disconnect.

stephentoub:
Some minor naming/visibility nits, otherwise LGTM.

src/Common/src/Interop/Windows/mincore/Interop.MoveFileEx.cs
JeremyKuhne: I didn't want to hunt down all of the usages to normalize p/Invoke naming in this PR- I'll do that as a separate PR to keep the functional changes isolated.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Wording with these is a bit awkward without getting really long. Technically it's AddExtendedPathPrefixForLongPathsIfNecessary (as it won't add it if it is already there). I'm hesitant to go that far. :)

stephentoub: So how about just AddExtendedPathPrefixIfNecessary?

JeremyKuhne: Because there already is one. Or would be if I added IfNecessary.  (AddExtendedPathPrefix exists already.)

stephentoub: Huh?  I don't understand.  AddExtendedPathPrefix does already exist (I suggested below that you make it private unless it needs to remain internal), but I don't see any function called AddExtendedPathPrefixIfNecessary.  Where is that function?

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Yes, it is called. Or is in the next change anyway.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: The next change actually uses the other method. We prepend under MAX_PATH for directory creation.

PR #2990, opened 2015-08-27
Remove unused gnu_get_libc_version P/Invoke declaration

nguerrera:
This was previously needed for the logic to switch between GNU
strerror_r and POSIX strerror_r, but that is now addressed by the shims.

cc @stephentoub


stephentoub:
Sweet.  LGTM.

PR #2993, opened 2015-08-27
Shim chmod, mkdir, mkfifo to make permission handling consistent

nguerrera:
We were mixing Interop.libc.Permissions with Interop.Sys.Stat. This
is not incorrect because the numeric values are fixed by POSIX, but
it was inconsistent and confusing.

We now have shims for all functions taking mode_t.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Native/Interop.MkDir.cs
+++ b/src/Common/src/Interop/Unix/System.Native/Interop.MkDir.cs
@@ -4,13 +4,11 @@
 using System;
 using System.Runtime.InteropServices;
 
-using mode_t = System.Int32;
-
 internal static partial class Interop
 {
-    internal static partial class libc
+    internal static partial class Sys
     {
-        [DllImport(Libraries.Libc, SetLastError = true)]
-        internal static extern int mkdir(string pathname, mode_t mode);
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal static extern int MkDir(string path, int mode);
     }
 }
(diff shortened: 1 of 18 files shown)


src/Common/src/Interop/Unix/System.Native/Interop.Permissions.cs
```diff
@@ -5,7 +5,7 @@
 
 internal static partial class Interop
 {
-    internal static partial class libc
+    internal static partial class Sys
     {
         [Flags]
         internal enum Permissions
```
stephentoub: This is the only time I can remember where I wished we had octal in C#, so these constants could be defined the same as they are on the native side.

stephentoub:
LGTM

src/Common/src/Interop/Unix/System.Native/Interop.Permissions.cs
nguerrera: I considered making them hex on the native side for the sake of consistency, but octal and permissions go together like peanut butter and jelly. :)

src/Common/src/Interop/Unix/System.Native/Interop.MkDir.cs
```diff
@@ -4,13 +4,11 @@
...
 using System.Runtime.InteropServices;
 
-using mode_t = System.Int32;
-
 internal static partial class Interop
 {
-    internal static partial class libc
+    internal static partial class Sys
     {
         [DllImport(Libraries.Libc, SetLastError = true)]
```
nguerrera: This needs to be SystemNative not libc and CI is failing. Waiting for full report to make sure there's nothing else before resetting CI with that fix.

stephentoub: It'll be nice when we can remove the Libc constant :smile:

PR #2994, opened 2015-08-27
Remove sql/ from gitignore

nguerrera:
There is no clear indication of why this folder is excluded (it's
int the "others" section) and it is causing problems for SqlClient.

@khellang This came in from 46bd33b0. Can you shed more light on why we'd want to exclude sql/?

cc @stephentoub @saurabh500

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,7 +147,6 @@ csx/
 AppPackages/
 
 # Others
-sql/
 *.Cache
 ClientBin/
 [Ss]tyle[Cc]op.*


stephentoub:
LGTM

saurabh500:
Works for me. 
@khellang Can you qualify this path with a more specific path based on where this folder is expected to be generated?

khellang:
@nguerrera It came from GitHub's "official" .gitignore for Visual Studio. It's since been removed, though; https://github.com/github/gitignore/pull/1315

PR #2997, opened 2015-08-27
Fully import ‎Skip and SkipWhile tests

JonHanna:
Convert ‎Skip and SkipWhile  tests imported in #2874 into a more directly
XUnit-ian style

Split existing tests between Skip and SkipWhile.

Of Skip tests, the following were omitted as duplicating
an existing test:

Skip1.Test1() → SkipExcessive()
Skip2.Test2() → SkipNone()
Skip3.Test3() → SkipNegative()
Skip5.Test5() → SkipAllExactly()

Of the SkipWhile tests SkipWhile1a.Test1a() and SkipWhile1b.Test1b() were
omited as duplicating SkipWhileAllTrue()

SkipWhile6.Test6() was crippled in the original, probably because it causes an
OutOfMemoryException. Fixed this, but it was very slow. Improved the
running time, but it's still very slow so marked it as Skipped.


stephentoub:
> Skip1.Test1() → SkipExcessive()

How do these map?  Skip1.Test1 was using an empty source and a positive count, whereas SkipExcessive is using an a source with 2 elements and a larger skip count.  I no longer see any tests for an empty source... am I just missing it?

src/System.Linq/tests/SkipWhileTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            public void Dispose()
+            {
+            }
+            public int Current
+            {
+                get { return 0; }
+            }
+        }
+
+        [Fact(Skip="Long-running")]
```
stephentoub: How long does it take?  Would it make sense to mark it as [OuterLoop] instead?

stephentoub:
A few questions, but otherwise LGTM.

JonHanna:
It was me that was missing something. I saw that they both tested an excessive skip, but not that the legacy test also covered an empty source. I've added a test for an empty source with zero, negative and positive skips (should be the same type of empty enumerable back).

src/System.Linq/tests/SkipWhileTests.cs
JonHanna: I wasn't aware of [OuterLoop] it matches a category only used in some tests? I got the time down to around 50-60 seconds when running on a Core i7. What's considered reasonable for a long-running test in [OuterLoop]?

akoeplinger: Yes, [OuterLoop] is used for tests that take a long time or are otherwise difficult to run (e.g. require admin rights).

stephentoub:
Thanks, @JonHanna.

stephentoub:
@dotnet-bot test this please

src/System.Linq/tests/SkipWhileTests.cs
JonHanna: Okay, I'll make use of that, so.

PR #3001, opened 2015-08-27
Unblock paths > MAX_PATH without extended syntax.

JeremyKuhne:
Also allow directories up to max component length (255) as opening does not
require extended syntax.

@stephentoub, @weshaggard, @Priya91

### DIFF
--- a/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
 
@@ -9,10 +10,30 @@ partial class Interop
 {
     partial class mincore
     {
+        /// <summary>
+        /// WARNING: This overload does not implicitly handle long paths.
+        /// </summary>
         [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
         internal unsafe static extern int GetFullPathNameW(char* path, int numBufferChars, char* buffer, IntPtr mustBeZero);
 
-        [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal static extern int GetFullPathNameW(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        private static extern int GetFullPathNameWPrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+
+        internal static int GetFullPathNameW(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
+        {
+            bool wasExtended = PathInternal.IsExtended(path);
+            if (!wasExtended)
+            {
+                path = PathInternal.AddExtendedPathPrefixForLongPaths(path);
+            }
+            int result = GetFullPathNameWPrivate(path, buffer.Capacity, buffer, mustBeZero);
+
+            if (!wasExtended)
+            {
+                // We don't want to give back \\?\ if we possibly added it ourselves
+                PathInternal.RemoveExtendedPathPrefix(buffer);
+            }
+            return result;
+        }
     }
 }
--- a/src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.IO;
 using System.Text;
 using System.Runtime.InteropServices;
 
@@ -8,10 +9,30 @@ partial class Interop
 {
     partial class mincore
     {
+        /// <summary>
+        /// WARNING: This overload does not implicitly handle long paths.
+        /// </summary>
         [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
         internal unsafe static extern int GetLongPathNameW(char* path, char* longPathBuffer, int bufferLength);
 
-        [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal static extern int GetLongPathNameW(string path, [Out]StringBuilder longPathBuffer, int bufferLength);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetLongPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        private static extern int GetLongPathNameWPrivate(string path, [Out]StringBuilder longPathBuffer, int bufferLength);
+
+        internal static int GetLongPathNameW(string path, [Out]StringBuilder longPathBuffer, int bufferLength)
+        {
+            bool wasExtended = PathInternal.IsExtended(path);
+            if (!wasExtended)
+            {
+                path = PathInternal.AddExtendedPathPrefixForLongPaths(path);
+            }
+            int result = GetLongPathNameWPrivate(path, longPathBuffer, longPathBuffer.Capacity);
+
+            if (!wasExtended)
+            {
+                // We don't want to give back \\?\ if we possibly added it ourselves
+                PathInternal.RemoveExtendedPathPrefix(longPathBuffer);
+            }
+            return result;
+        }
     }
 }
--- a/src/Common/src/System/IO/PathInternal.Windows.cs
+++ b/src/Common/src/System/IO/PathInternal.Windows.cs
@@ -16,7 +16,7 @@ namespace System.IO
         internal const string DevicePathPrefix = @"\\.\";
         internal const int MaxShortPath = 260;
         internal const int MaxShortDirectoryPath = 248;
-        internal const int MaxExtendedPath = short.MaxValue;
+        internal const int MaxLongPath = short.MaxValue;
 
         internal static readonly char[] InvalidPathChars =
         {
@@ -41,16 +41,29 @@ namespace System.IO
         /// </summary>
         internal static bool IsPathTooLong(string fullPath)
         {
+            // We'll never know precisely what will fail as paths get changed internally in Windows and
+            // may grow to exceed MaxExtendedPath. We'll only try to catch ones we know will absolutely
+            // fail.
+
+            if (fullPath.Length < MaxLongPath - UncExtendedPathPrefix.Length)
+            {
+                // We won't push it over MaxLongPath
+                return false;
+            }
+
+            // We need to check if we have a prefix to account for one being implicitly added.
             if (IsExtended(fullPath))
             {
-                return fullPath.Length >= MaxExtendedPath;
+                // We won't prepend, just check
+                return fullPath.Length >= MaxLongPath;
             }
-            else
+
+            if (fullPath.StartsWith(UncPathPrefix, StringComparison.Ordinal))
             {
-                // Will need to be updated with #2581 to allow all paths to MaxExtendedPath
-                // minus legth of extended local or UNC prefix.
-                return fullPath.Length >= MaxShortPath;
+                return fullPath.Length + UncExtendedPrefixToInsert.Length >= MaxLongPath;
             }
+
+            return fullPath.Length + ExtendedPathPrefix.Length >= MaxLongPath;
         }
 
         /// <summary>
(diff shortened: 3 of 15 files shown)


src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
```diff
@@ -9,10 +10,30 @@ partial class Interop
 {
     partial class mincore
     {
+        /// <summary>
+        /// WARNING: This overload does not implicitly handle long paths.
+        /// </summary>
```
stephentoub: Let's rename it then. It's strange to have two overloads so similar and yet with so different behavior.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
```diff
@@ -9,10 +10,30 @@ partial class Interop
...
-        internal static extern int GetFullPathNameW(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        private static extern int GetFullPathNameWPrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+
+        internal static int GetFullPathNameW(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
+        {
+            bool wasExtended = PathInternal.IsExtended(path);
+            if (!wasExtended)
+            {
+                path = PathInternal.AddExtendedPathPrefixForLongPaths(path);
```
stephentoub: Should we increase the size of buffer by a commensurate amount as we increased the length of path?

src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
```diff
@@ -8,10 +9,30 @@ partial class Interop
 {
     partial class mincore
     {
+        /// <summary>
+        /// WARNING: This overload does not implicitly handle long paths.
+        /// </summary>
```
stephentoub: Ditto on using a different name.  Do we expect any code other than the other overload to use these non-extending versions?  Let's make them private if possible.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -41,16 +41,29 @@ internal static partial class PathInternal
         /// </summary>
         internal static bool IsPathTooLong(string fullPath)
         {
+            // We'll never know precisely what will fail as paths get changed internally in Windows and
+            // may grow to exceed MaxExtendedPath. We'll only try to catch ones we know will absolutely
+            // fail.
+
+            if (fullPath.Length < MaxExtendedPath - UncExtendedPathPrefix.Length)
+            {
+                // For sure is ok
```
stephentoub: Why?  The previous comment just stated that paths can grow, and it's not clear why the checked condition makes this ok.  Please augment the comment to explain fully.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -272,17 +276,17 @@ internal static bool IsPathRelative(string path)
                 return true;
             }
 
-            if ((path[0] == '\\') || (path[0] == '/'))
+            if ((path[0] == Path.DirectorySeparatorChar) || (path[0] == Path.AltDirectorySeparatorChar))
             {
                 // There is no valid way to specify a relative path with two initial slashes
-                return !((path[1] == '\\') || (path[1] == '/'));
+                return !((path[1] == Path.DirectorySeparatorChar) || (path[1] == Path.AltDirectorySeparatorChar));
```
stephentoub: We have an internal IsDirectorySeparator function somewhere.  Maybe it's on Path rather than on PathInternal?  Consider adding one to PathInternal (or moving it to PathInternal) to simplify these checks, e.g.
```C#
if (IsDirectorySeparator(path[0]))
{
    return !IsDirectorySeparator(path[1]);
}
```

src/Common/tests/Tests/System/IO/PathInternal.Windows.Tests.cs
```diff
@@ -36,6 +36,7 @@ public void AddExtendedPathPrefixTest(string path, string expected)
         InlineData(@"\\?\C:\Path", false),
         InlineData(@"Path", true),
         InlineData(@"X", true)]
+    [PlatformSpecific(PlatformID.Windows)]
```
stephentoub: Please add an AnyUnix version of this test as well.

src/System.IO.FileSystem/tests/Directory/Move.cs
```diff
@@ -144,14 +144,26 @@ public void Path_Longer_Than_MaxPath_Throws_Exception()
...
         {
             string testDir = GetTestFilePath();
             Directory.CreateDirectory(testDir);
+            Assert.True(Directory.Exists(testDir), "test directory should exist");
             Assert.All((IOInputs.GetPathsLongerThanMaxDirectory()), (path) =>
             {
-                Assert.Throws<PathTooLongException>(() => Move(testDir, path));
-                Assert.Throws<PathTooLongException>(() => Move(path, testDir));
+                string baseDestinationPath = Path.GetDirectoryName(path);
+                if (!Directory.Exists(baseDestinationPath))
```
stephentoub: Why this check?  Doesn't CreateDirectory already check?

src/System.IO.FileSystem/tests/File/Move.cs
```diff
@@ -144,7 +144,42 @@ public void FileNameWithSignificantWhitespace()
         }
 
         [Fact]
-        public void LongPath()
+        [PlatformSpecific(PlatformID.Windows)]
+        public void LongPath_Windows()
```
stephentoub: This test seems to be testing a bunch of different things.  I think it'd be better to split it up.

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
```diff
@@ -21,13 +21,17 @@ internal static class IOInputs
     // Max path length (minus trailing \0). Unix values vary system to system; just using really long values here likely to be more than on the average system.
     public static readonly int MaxPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 259 : 10000;
 
+    // Same as MaxPath on Unix
```
stephentoub: Then shouldn't the else just be MaxPath? e.g.
```C#
public static readonly int MaxLongPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? MaxExtendedPath : MaxPath;
```

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
```diff
@@ -21,13 +21,17 @@ internal static class IOInputs
...
+    public static readonly int MaxLongPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? MaxExtendedPath : 10000;
+
     // Windows specific, this is the maximum length that can be passed to APIs taking directory names, such as Directory.CreateDirectory & Directory.Move.
     // Does not include the trailing \0.
-    public static readonly int MaxDirectory = 247;
+    public static readonly int MaxDirectory = MaxComponent - 1;
```
stephentoub: This is changing the value from 247 to 254.  That's intentional?

src/System.Runtime.Extensions/tests/System/IO/PathTests.cs
```diff
@@ -408,7 +408,9 @@ public static void GetFullPath_Windows_NormalizedLongPathTooLong()
         {
             longPath.Append(Path.DirectorySeparatorChar).Append('a').Append(Path.DirectorySeparatorChar).Append('.');
         }
-        Assert.Throws<PathTooLongException>(() => Path.GetFullPath(longPath.ToString()));
+
+        // Now no longer throws unless over ~32K
```
stephentoub: Shouldn't we have PathTests for these additional cases now?

src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
```diff
@@ -8,10 +9,30 @@ partial class Interop
...
-        internal static extern int GetLongPathNameW(string path, [Out]StringBuilder longPathBuffer, int bufferLength);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetLongPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        private static extern int GetLongPathNameWPrivate(string path, [Out]StringBuilder longPathBuffer, int bufferLength);
+
+        internal static int GetLongPathNameW(string path, [Out]StringBuilder longPathBuffer, int bufferLength)
+        {
+            bool wasExtended = PathInternal.IsExtended(path);
+            if (!wasExtended)
+            {
+                path = PathInternal.AddExtendedPathPrefixForLongPaths(path);
```
stephentoub: In the previous review, you said we couldn't change AddExtendedPathPrefixForLongPaths to AddExtendedPathPrefixIfNecessary because it would conflict with something in this next PR.  I'm not seeing it.  What's the conflict?

JeremyKuhne: CreateDirectory is the one doing it.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: Do you have any suggestions for naming? GetFullPathNameNoLongPath? This one in particular is used for the stack allocating code in PathHelper (all under MAX_PATH).

Like I mentioned, I intend to pass through these P/Invokes and normalize the naming. I want to drop the "W", use "Private" (anything better for that?), add comments on the privates to avoid leaking in other P/Invoke wrappers.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: Yes, I'll change that.  Good catch.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: For our purposes it's ok.  We won't put it over the max by prepending.  I'll clarify.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I'll go ahead and change it- I'll try not to second guess the Jitter. :)

src/Common/tests/Tests/System/IO/PathInternal.Windows.Tests.cs
JeremyKuhne: Need to implement the Unix version of the api first.

src/System.IO.FileSystem/tests/Directory/Move.cs
JeremyKuhne: For troubleshooting when tests fail. If the prereq's aren't there you can rat hole into collateral fallout.

src/System.IO.FileSystem/tests/File/Move.cs
JeremyKuhne: I'll clarify.

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
JeremyKuhne: Yep. Fixing.

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
JeremyKuhne: I really meant MaxPath. Changing.

src/System.Runtime.Extensions/tests/System/IO/PathTests.cs
JeremyKuhne: This was checking the normalizes to over max path. I'll make sure we have the right test.

src/System.IO.FileSystem/tests/PortedCommon/IOInputs.cs
JeremyKuhne: Actually I'm going to change it back and clarify.

src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
JeremyKuhne: Spoke offline. Going to use EnsureExtendedPrefix/OverMaxPath in a follow-up.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -272,17 +276,17 @@ internal static bool IsPathRelative(string path)
                 return true;
             }
 
-            if ((path[0] == '\\') || (path[0] == '/'))
+            if (IsDirectorySeparator(path[0]))
             {
                 // There is no valid way to specify a relative path with two initial slashes
-                return !((path[1] == '\\') || (path[1] == '/'));
+                return !(IsDirectorySeparator(Path.DirectorySeparatorChar));
```
weshaggard: Shouldn't this be `!IsDirectorySeparator(path[1])`

weshaggard:
Other then the one issue and the renaming which you will be doing in a follow-up. The change looks good to me.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Yeah :/ I'll add the test for this one.

PR #3006, opened 2015-08-28
Fix WrappedStream.ThrowIfCantRead

Maxwe11:
Currently `ThrowIfCantRead` checks `CanWrite` property which looks incorrect.
Actually throw line not covered by tests and seems to me it's a potential dead code
because `ZipArchive.Init` ensures that stream is readable.


stephentoub:
cc: @ianhays

stephentoub:
LGTM

Issue #3009, opened 2015-08-28
PathHelper stackalloc version can't grow


JeremyKuhne:
Will always fail with MAX_PATH if it goes beyond when resolving paths (relative, short names, etc.). Need to change to allow switching between stack and heap when needed.

JeremyKuhne:
#3992 addresses this

PR #3010, opened 2015-08-28
Clean naming and disable stack pathhelper

JeremyKuhne:
PathHelper doesn't handle growing past MAX_PATH when using
the stackalloc version. #3009 tracks this.

@weshaggard, @stephentoub, @Priya91

### DIFF
--- a/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
@@ -11,27 +11,30 @@ partial class Interop
     partial class mincore
     {
         /// <summary>
-        /// WARNING: This overload does not implicitly handle long paths.
+        /// WARNING: This method does not implicitly handle long paths. Use GetFullPathName.
         /// </summary>
-        [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal unsafe static extern int GetFullPathNameW(char* path, int numBufferChars, char* buffer, IntPtr mustBeZero);
-
         [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        private static extern int GetFullPathNameWPrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+        internal unsafe static extern int GetFullPathNameUnsafe(char* path, int numBufferChars, char* buffer, IntPtr mustBeZero);
 
-        internal static int GetFullPathNameW(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
+        /// <summary>
+        /// WARNING: This method does not implicitly handle long paths. Use GetFullPathName.
+        /// </summary>
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        private static extern int GetFullPathNamePrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
+
+        internal static int GetFullPathName(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
         {
             bool wasExtended = PathInternal.IsExtended(path);
             if (!wasExtended)
             {
-                path = PathInternal.AddExtendedPathPrefixForLongPaths(path);
+                path = PathInternal.EnsureExtendedPrefixOverMaxPath(path);
             }
-            int result = GetFullPathNameWPrivate(path, buffer.Capacity, buffer, mustBeZero);
+            int result = GetFullPathNamePrivate(path, buffer.Capacity, buffer, mustBeZero);
 
             if (!wasExtended)
             {
                 // We don't want to give back \\?\ if we possibly added it ourselves
-                PathInternal.RemoveExtendedPathPrefix(buffer);
+                PathInternal.RemoveExtendedPrefix(buffer);
             }
             return result;
         }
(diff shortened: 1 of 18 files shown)


src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -176,16 +176,7 @@ private unsafe static string NormalizePath(string path, bool fullCheck, int maxP
...
-            if (path.Length + 1 <= MaxPath)
-            {
-                char* m_arrayPtr = stackalloc char[MaxPath];
-                newBuffer = new PathHelper(m_arrayPtr, MaxPath);
-            }
-            else
-            {
-                newBuffer = new PathHelper(path.Length + MaxPath, maxPathLength);
-            }
+            PathHelper newBuffer = new PathHelper(path.Length + MaxPath, maxPathLength);
```
weshaggard: Do we know the perf impact of this change? Should we still use stackalloc for smaller paths?

weshaggard:
Renames look reasonable to me. I just had the one question but you have filed an issue tracking that so you could use it for thinking about the perf impact as well.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: I can't because the stackalloc cannot grow currently. I've opened an issue on it and will fix so we can dynamically move up to the heap.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
```diff
@@ -11,27 +11,30 @@ partial class Interop
     partial class mincore
     {
         /// <summary>
-        /// WARNING: This overload does not implicitly handle long paths.
+        /// WARNING: This method does not implicitly handle long paths. Use GetFullPathName.
         /// </summary>
-        [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal unsafe static extern int GetFullPathNameW(char* path, int numBufferChars, char* buffer, IntPtr mustBeZero);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        internal unsafe static extern int GetFullPathNameUnsafe(char* path, int numBufferChars, char* buffer, IntPtr mustBeZero);
```
stephentoub: Why does this one have a different suffix?  It can't be an overload of GetFullPathNamePrivate?

src/Common/src/Interop/Windows/mincore/Interop.GetLongPathNameW.cs
```diff
@@ -10,27 +10,30 @@ partial class Interop
     partial class mincore
     {
         /// <summary>
-        /// WARNING: This overload does not implicitly handle long paths.
+        /// WARNING: This method does not implicitly handle long paths. Use GetLongPathName.
         /// </summary>
-        [DllImport(Libraries.CoreFile_L1, SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
-        internal unsafe static extern int GetLongPathNameW(char* path, char* longPathBuffer, int bufferLength);
+        [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetLongPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
+        internal unsafe static extern int GetLongPathNameUnsafe(char* path, char* longPathBuffer, int bufferLength);
```
stephentoub: Ditto.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
stephentoub: Isn't there a known maximum amount by which it will need to grow, measured in just a few characters?  If so, couldn't we just proactively make the stackalloc'd size that much larger?

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: Because it is internal and unsafe. It is used by the PathHelper class. Best I could come up with, but always open to suggestions.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: Technically it shouldn't go beyond 520 characters (start with 260 relative and add to current directory). But there is no guarantee of that- I know of ways you can leak long paths into GetCurrentDirectory, and there might be others. There is also some hope that current directory will support > 260 in the future. I've got a lot of ideas about how to approach this issue- any pointers to perf experts would be appreciated. (One thought is a thread local string builder cache for path normalization.)

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
stephentoub: There are lots of methods that use unsafe code / take ptr arguments that don't have Unsafe in the name.  More importantly, though, that's not the defining characteristic with regards to these various methods: the important thing we want to highlight is that it doesn't handle long paths / doesn't provide the wrapping behavior.  How about using "Core" as the suffix, to highlight that it's the core functionality, and then the other things wrap it?  Ideally we'd use it consistently instead of using "Private" in some places and other suffixes in others.  If we can't do that, I'd prefer something like "Native" here, since it at least that highlights that this is the Win32 P/Invoke function.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
stephentoub: If StringBuilderCache solves the need here, I'm fine with that.  We'd want to make sure, though, that this particular usage didn't end up poisoning the cache, since IIRC StringBuilderCache only puts builders back into the cache if their capacity is below a threshold.

Issue #642, opened 2015-02-05
Clean up System.Xml.Linq.XDocument tests


krwq:
- [x] Remove spam messages (there are few left)
- [x] Some of the tests might be failing but not throwing (look at the messages - https://github.com/dotnet/corefx/issues/641)
- [x] Convert tests to use Xunit (right now they are only partially)
- [ ] Replace if (expr) throw new TestException, etc with proper use of Xunit asserts

sepidehMS:
Failing tests and spam messages resolved by https://github.com/dotnet/corefx/pull/5392

kasper3:
> Convert tests to use Xunit (right now they are only partially)

Is this done?

krwq:
@kasper3 I'd call it done as do not think it is worth to do any more work there but this is not 100% accurate:
- we have converted most of the projects to Xunit
- for remaining tests I have added test discoverer so that Xunit can see Xml custom framework - this has achieved the goal of being able to see better errors and making it easier to investigate failures but the custom framework is still there

last point here is to do some clean ups - I think this might not be worth it unless we add some new tests at the same time - closing this issue.

PR #3019, opened 2015-08-30
Convert System.Xml.XDocument.SDMSample.Tests to use Xunit

Maxwe11:
Progress against #642

I recommend to review without whitespace changes
https://github.com/dotnet/corefx/pull/3019/files?w=1

/cc: @krwq


Maxwe11:
@dotnet-bot test this please

Maxwe11:
@dotnet-bot test this please

src/System.Xml.XDocument/tests/SDMSample/SDMDocument.cs
```diff
@@ -4,189 +4,146 @@
...
-
-                    // First node: declaration
-                    Validate.IsEqual(doc.Declaration.ToString(), doc2.Declaration.ToString());
-
-                    // Next node: comment
-                    Validate.IsEqual(e.MoveNext(), true);
-                    Validate.Type(e.Current, typeof(XComment));
-                    Validate.IsNotReferenceEqual(e.Current, comment);
-                    XComment comment2 = (XComment)e.Current;
-                    Validate.IsEqual(comment2.Value, comment.Value);
```
akoeplinger: `Validate.IsEqual` verified that neither of the both parameters is `null`, however `Assert.Equal` with which it was replaced doesn't.

You might want to audit each of those and make sure to add `Assert.NotNull` where appropriate (e.g. most of the time one of the two arguments is known to be non-null so it's not necessary there).

Maxwe11: Thanks, will audit same issue for `Validate.ReferenceEquals` -> `Assert.Same` replacement.
But in this case `comment.Value` initialized with non null value. If it will null for some reason (e.g. regression) another test in XComment's tests will fail.

stephentoub:
Thanks, @Maxwe11!  It's great seeing all that custom goo go away.  Same comment as on the other one: we can address the namespace issue separately.  LGTM.

Issue #516, opened 2015-01-23
Replace hand-rolled tests for exceptions with xunit's Assert.Throws


jhendrixMSFT:
There is a lot of hand-rolled exception checking in various test code.  This should be refactored to use xunit's Assert.Throws method as it improves readability and can reduce possible bugs in test code.

For an example please see PR https://github.com/dotnet/corefx/pull/512

Clockwork-Muse:
I can take this on, shouldn't be too bad.

Clockwork-Muse:
Hmm.... at least some of the tests mentioned in the linked issue are situations where you'd want to invoke a `fail(...)`, which doesn't exist for xUnit (and the given pattern is the recommendation, apparently).
So some of them are expecting to _not_ throw an exception, essentially, and need to fail the test if certain conditions are met.

Clockwork-Muse:
And some of them are using the pattern to pop messages if a test is failed (`System.Linq.Parallel` is what I'm looking at right now).  Similar messages should be popped by changing things to theories, etc.

stephentoub:
Just FYI, I fixed this kind of thing for TPL Dataflow as part of overhauling those tests.  See PR #513.

Clockwork-Muse:
@stephentoub - Thanks for letting us know.

Clockwork-Muse:
@stephentoub , @jhendrixMSFT - 
Before I go too much farther, I was wondering what the, err, limit was on changes.  Because currently my changes seem to be looking like this:
https://github.com/Clockwork-Muse/corefx/commit/d4d2210126d951bae1cfd78f8668436d6c4aef32
https://github.com/Clockwork-Muse/corefx/commit/a0e4af8d0f2aae212f3a7ddb630aba974bcfbbc5

(My time so far has mostly been spent figuring out how failures were reported differently for various ways to write the tests, so things would be understandable and discoverable if/when that does happen)

YingP99:
I've written a simple tool to transform try-catch to Assert.Throws (for certain patterns). It may not fix every case, but do save you a lot of typing time. I've not figured out how to share it.
Note: it uses Roslyn SynatxRewriter, and works on cs file (no need project or solution).

Clockwork-Muse:
@YingP99 - Thanks!
Potentially, you could share a repository with the source for the tool (ie, place a link here), which could potentially include a download link.
[...]

PR #3020, opened 2015-08-30
Remove custom error checking

Clockwork-Muse:
Remove custom error checking, per #516 .
In general, I ignored deeper issues with the tests, such as maintaining own frameworks (such as logging errors for reporting later).  A number of projects have improper namespaces, as per #2898 ; this will be resolved in a separate PR.

This removes all (current! stuff keeps being added) cases of `Assert.True(false, ...);` and `Assert.False(true, ...);`, except:
 - System.Collections.Specialized - **ALL** tests are of this nature (no `Assert.Equal(...)`, etc), and wonky otherwise besides.  In process of general cleanup (ala PLINQ).
 - System.Threading.Tasks - A number of tests rely heavily on an internal (to the tests) framework; unwinding will involve additional effort.
 - System.Linq.Parallel - a few instances in the Cancellation tests remain.  I'm waiting to submit a PR until I get feedback on some of my submitted issues (will add/change/remove some related tests).


Clockwork-Muse:
Hmm, it seems I've forgotten to notify some people....?
@jhendrixMSFT ,

Clockwork-Muse:
Hmm... looks like the X509 tests timed out, but I didn't modify them?

@dotnet-bot test this please

Clockwork-Muse:
@Maxwe11 - Moved `ShouldNotBeInvokedException` to Common folder, changed references, and rebased to master.  
System.Net.Requests ended up staying in, after finding new instances of the issue.  There was a couple new instances in Systems.Diagnositcs.Process, not sure if I missed them the first time around.
System.Net.Security is new.

src/System.Collections.Concurrent/tests/BlockingCollectionTests.cs
```diff
@@ -282,9 +278,7 @@ private static void Test2_ConcurrentAdd(int numOfThreads, int numOfElementsPerTh
 
                     for (int j = startOfSequence; j < endOfSequence; ++j)
                     {
-                        if (!blockingCollection.TryAdd(j))
-                            Assert.False(true,
-                                String.Format("Test2_ConcurrentAdd(numOfThreads={0}, numOfElementsPerThread={1}): > test failed - TryAdd returned false unexpectedly", numOfThreads, numOfElementsPerThread));
+                        Assert.True(blockingCollection.TryAdd(j));
```
stephentoub: We're losing some information here, e.g. the numOfThreads and numOfElementsPerThreads values.

src/System.Collections/tests/Generic/HashSet/HashSetTestSupport.cs
```diff
@@ -1015,7 +1016,7 @@ public class Item
         {
             public int x;
 
-            private static int s_nextIndex = 1;
+            private static int s_nextIndex = 0;
             private static object s_nextIndexLock = new object();
```
stephentoub: Is this field needed anymore?

src/System.Composition.Convention/tests/ExceptionAssert.cs
```diff
@@ -10,315 +10,31 @@ namespace System.Composition.Convention
...
-        /// <summary>
-        ///     Verifies that the specified action throws an exception of type <typeparam name="T"/>,
-        ///     with the specified inner exception, and indicating whether to retry.
-        /// </summary>
-        public static T Throws<T>(Exception innerException, RetryMode retry, Action action)
-            where T : Exception
-        {
-            return Throws<T>(innerException, RetryMode.Retry, action, (Action<T, int>)null);
-        }
+        // NOTE: To catch state corrupting exceptions, it by design that
```
stephentoub: Nit: "it is by design" (I realize the error was there in the original, just pointing it out)

src/System.Globalization/tests/CultureInfo/CultureInfoAll.cs
```diff
@@ -21,28 +21,24 @@ public void TestAllCultures()
...
-                Validate(ci);
-            }
+            Assert.All(cultures, Validate);
         }
 
         private void Validate(CultureInfo ci)
         {
             string val = GetLocaleInfo(ci, LOCALE_SENGLISHDISPLAYNAME);
-            Assert.True(ci.EnglishName.Equals(val), String.Format("Expected {0} and got {1}", ci.EnglishName, val));
+            Assert.Equal(ci.EnglishName, GetLocaleInfo(ci, LOCALE_SENGLISHDISPLAYNAME));
```
stephentoub: Why duplicate the line above?

src/System.IO.Packaging/tests/Tests.cs
```diff
@@ -2040,72 +1941,6 @@ public void T007_PackageProperties()
         }
 
         [Fact]
-        public void T006_CreateExternalRelationshipInvalidUri()
```
stephentoub: Why is this test being deleted?

src/System.IO.Packaging/tests/Tests.cs
```diff
@@ -2587,31 +2419,29 @@ public void T131_String_OpenOrCreate_ReadWrite_Star()
...
 
-                    using (Stream partStream = mainPart.GetStream(FileMode.Open, FileAccess.Read))
-                    {
-                        XDocument xd = XDocument.Load(partStream);
-                        Assert.True(xd.DescendantNodes().Count() != 0);
-                    }
+                using (Stream partStream = mainPart.GetStream(FileMode.Open, FileAccess.Read))
+                {
+                    XDocument xd = XDocument.Load(partStream);
+                    Assert.True(xd.DescendantNodes().Count() != 0);
```
stephentoub: Assert.NotEqual?

src/System.IO.Packaging/tests/Tests.cs
```diff
@@ -2656,40 +2486,38 @@ public void T128_String_Open_ReadWrite_OpenOrCreate_Read_Write_ReadWrite()
...
                            new Uri("/", UriKind.Relative),
                                  docPackageRelationship.TargetUri);
 
-                    var mainPart = package.GetPart(documentUri);
+                var mainPart = package.GetPart(documentUri);
+
+                using (Stream partStream = mainPart.GetStream(FileMode.OpenOrCreate, FileAccess.Read))
+                {
+                    XDocument xd = XDocument.Load(partStream);
+                    Assert.True(xd.DescendantNodes().Count() != 0);
```
stephentoub: Ditto

src/System.IO.Packaging/tests/Tests.cs
```diff
@@ -2706,37 +2534,35 @@ public void T127_String_Open_ReadWrite_Open_Read_Write_ReadWrite()
...
 
-                    using (Stream partStream = mainPart.GetStream(FileMode.Open, FileAccess.Read))
-                    {
-                        XDocument xd = XDocument.Load(partStream);
-                        Assert.True(xd.DescendantNodes().Count() != 0);
-                    }
+                using (Stream partStream = mainPart.GetStream(FileMode.Open, FileAccess.Read))
+                {
+                    XDocument xd = XDocument.Load(partStream);
+                    Assert.True(xd.DescendantNodes().Count() != 0);
```
stephentoub: Ditto.  (I'll stop commenting on these.)

src/System.Reflection/tests/FieldInfo/FieldInfo_ConstantTests.cs
```diff
@@ -68,11 +66,22 @@ public void TestSetValue_constbyteField()
             TestSetValue_constantField("byteField", (object)byte.MaxValue);
         }
 
-        //Verify SetValue method does not set value for RO Field
+        //Verify SetValue method _set_ set value for RO Field
```
stephentoub: I don't understand this change.

src/System.Runtime/tests/System/Nullable.cs
```diff
@@ -13,88 +13,31 @@ public static void TestBasics()
...
-        v = n.GetValueOrDefault();
-        Assert.Equal(v, default(int));
-
-        v = n.GetValueOrDefault(999);
-        Assert.Equal(v, 999);
+        Assert.False(n.HasValue);
+        Assert.Throws<InvalidOperationException>(() => n.Value);
+        Assert.Throws<InvalidOperationException>(() => (int)n);
+        Assert.Equal(n, null);
+        Assert.NotEqual(n, 7);
```
stephentoub: The arguments to the above two calls should be swapped, right?

stephentoub: The arguments to the above two calls should be swapped, right?

src/System.Xml.ReaderWriter/tests/XmlReader/XmlResolver/XmlSystemPathResolverTests.cs
```diff
@@ -75,6 +68,7 @@ private static void AssertInvalidPath(string path)
             // Due to different shipping behavior in different products, different exceptions may be thrown
             // when an invalid path is specified on different platforms. We try to catch all the types
             // that can be thrown here to verify that more exception types are not introduced inadvertently.
+            // Ideally, expected exceptions should be parceled out by shipped product.
```
stephentoub: I don't understand the addition.

stephentoub:
Thanks, @Clockwork-Muse.  I went through all of the changes, and overall it looks great... thanks for the cleanup.

A few things:
1) I left some comments.
2) Please either revert or separate out into a separate PR the changes for System.Net.*.  There's a lot of work happening in that area right now in other branches from master, and while it's unlikely this would cause a problem, I don't want there to be any friction there.
3) Please update the titles of the commits to be something more specific than the assembly names.  You could also squash this down into a single commit if you'd like.

src/System.IO.Packaging/tests/Tests.cs
Clockwork-Muse: Probably missed because I was just going through looking for explicit `Assert.True(false...` and `Assert.False(true...`, and ends up showing in the diff because I switched to checking for null as a test, instead of as a condition (so indentation changed).  I can fix these, but we probably need to audit most tests for stuff like this.

src/System.IO.Packaging/tests/Tests.cs
Clockwork-Muse: I'll verify when I get a chance, but I believe it turned out the failing line was 2085: `Uri uri = new Uri("a b c d e", UriKind.Absolute);`, which meant we weren't testing anything in packaging behavior.

src/System.Collections/tests/Generic/HashSet/HashSetTestSupport.cs
Clockwork-Muse: No, we do not.  

I'm not sure what lead to me discovering and making that particular change in the first place.  Maybe I thought I was getting deadlock or something?  Still, use of interlocked reads cleaner...
There a second instance of this in the file, so replacing that too.

Clockwork-Muse: No, we do not.  

I'm not sure what lead to me discovering and making that particular change in the first place.  Maybe I thought I was getting deadlock or something?  Still, use of interlocked reads cleaner...
There a second instance of this in the file, so replacing that too.

src/System.Composition.Convention/tests/ExceptionAssert.cs
Clockwork-Muse: Fixed.

src/System.Globalization/tests/CultureInfo/CultureInfoAll.cs
Clockwork-Muse: `val` isn't used anywhere else, so I think I forgot to remove the line.

src/System.Collections.Concurrent/tests/BlockingCollectionTests.cs
Clockwork-Muse: Updating these tests to use InlineData instead of passing via method call.  The entire set probably needs to be gone over for this, though.  Future intern work.

src/System.Reflection/tests/FieldInfo/FieldInfo_ConstantTests.cs
Clockwork-Muse: That first 'set' should have been 'does'.

The way the original delegated test (`TestSetValue_constant field`) was written, it masked that read-only fields weren't throwing the exceptions constant fields did.  (I ended up posting a short-lived issue because I had a bug in my repro code, but I digress)    I changed it to this form for documentation purposes.

src/System.Runtime/tests/System/Nullable.cs
Clockwork-Muse: Yes.

src/System.Xml.ReaderWriter/tests/XmlReader/XmlResolver/XmlSystemPathResolverTests.cs
Clockwork-Muse: Removed the comment (and updated the test to guarantee an error is thrown).

This was an attempt to say we should correlate which platforms throw which exceptions (to at least be aware of such changes), although I couldn't do that work myself.  Should I raise an issue?

src/System.IO.Packaging/tests/Tests.cs
Clockwork-Muse: I'm making them Assert.Equal to whatever it is they currently return (which I'm assuming is both correct and stable).  Should I raise an issue for somebody to check these?

src/System.IO.Packaging/tests/Tests.cs
Clockwork-Muse: Verified, it was the Uri throwing the error.

After a little more research, this test was maybe attempting to test for invalid use of an absolute Uri as an internal relationship.  I've changed it to reflect this possibility.

Clockwork-Muse:
@stephentoub - 
 1) Comments addressed.
 2) Pulled.  Since some of the changes depend on ShouldNotBeInvokedException, I'm going to wait to issue the PR to until this one clears.
 3) Squashed, with additional comments.  Was that sufficient, or did I miss something?

stephentoub:
Thanks for the cleanup!

PR #3021, opened 2015-08-30
Tighten loops on EnumerableHelpers.ToArray()

JonHanna:
One change is analogous to that in 6cf2b4c in that it removes a check for the
initial state.

Also removes assignment in case where it is just overwritten.

Also moves check for new size being smaller than current size to within branch
that can lead to it.

Reduces number of resizes when between maximum non-single-byte-element array
size and maximum single-byte-element array size.

Use EnumerableHelper.ToArray in Buffer constructor:

Greater code re-use, and gains from EnumerableHelper's better support for
near-upper-limit arrays.

### DIFF
--- a/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
+++ b/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
@@ -7,6 +7,8 @@ namespace System.Collections.Generic
     internal static class EnumerableHelpers
     {
         /// <summary>Converts an enumerable to an array using the same logic as does List{T}.</summary>
+        /// <param name="source">The enumerable to convert.</param>
+        /// <returns>The resulting array.</returns>
         internal static T[] ToArray<T>(IEnumerable<T> source)
         {
             int count;
@@ -16,6 +18,7 @@ namespace System.Collections.Generic
         }
 
         /// <summary>Converts an enumerable to an array using the same logic as does List{T}.</summary>
+        /// <param name="source">The enumerable to convert.</param>
         /// <param name="length">The number of items stored in the resulting array, 0-indexed.</param>
         /// <returns>
         /// The resulting array.  The length of the array may be greater than <paramref name="length"/>,
@@ -23,18 +26,11 @@ namespace System.Collections.Generic
         /// </returns>
         internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
         {
-            T[] arr;
-            int count = 0;
-
             ICollection<T> ic = source as ICollection<T>;
             if (ic != null)
             {
-                count = ic.Count;
-                if (count == 0)
-                {
-                    arr = Array.Empty<T>();
-                }
-                else
+                int count = ic.Count;
+                if (count != 0)
                 {
                     // Allocate an array of the desired size, then copy the elements into it. Note that this has the same 
                     // issue regarding concurrency as other existing collections like List<T>. If the collection size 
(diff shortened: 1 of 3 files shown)


src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -15,26 +17,22 @@ internal static T[] ToArray<T>(IEnumerable<T> source)
             return results;
         }
 
-        /// <summary>Converts an enumerable to an array using the same logic as does List{T}.</summary>
+        /// <summary>Converts an enumerable to an array using similar as does List{T}.</summary>
```
Eyas: > `using similar as does List{T}`

typo, I assume.

JonHanna: Yes. Since it's not precisely the same, I went to replace "the same" with "similar", but I obviously must have selected more than I thought. Will fix.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -42,53 +40,69 @@ internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
...
-                            newLength = count + 1;
+                            if (count == arr.Length)
+                            {
+                                // MaxArrayLength is defined in Array.MaxArrayLength and in gchelpers in CoreCLR.
+                                // It represents the maximum number of elements that can be in an array where
+                                // the size of the element is greater than one byte; a separate, slightly larger constant,
+                                // is used when the size of the element is one.
+                                const int MaxArrayLength = 0x7FEFFFFF;
+        
+                                // We attempt to
```
justinvp: Nit: the word wrapping seems like it could be cleaned up here

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -42,53 +40,69 @@ internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
...
+                        return arr;
                     }
-                    arr[count++] = item;
                 }
             }
 
-            length = count;
-            return arr;
+            length = 0;
+            return Array.Empty<T>();
```
stephentoub: Isn't this now returning ```Array.Empty<T>``` from LINQ's ToArray?

JonHanna: No. `EnumerableHelper` is unchanged in this regard (the only observable change is that it might have a larger excess with one-byte-element arrays larger than 0x7FEFFFFF) and while using `EnumerableHelper` to build `Buffer` means it would internally have a reference to `Array.Empty` where before it would have had `null` in actually using the array it still pass through [https://github.com/dotnet/corefx/pull/3021/files#diff-95a14d3b4ae14339b623dd95c7537fd0R3398](the check):

    if (count == 0) return new TElement[0]; // consider replacing with Array.Empty<TElement>()

If the possibility in the comment is ever decided as the way to go, then this now means just removing that line, but in the meantime a fresh zero-length array is still created each time.

JonHanna: Also, there's at least one test that would fail if `ToArray()` did return `Array.Empty<T>`, so the tests confirm what I say above.

stephentoub: Yup, you're right; while looking at this on my phone, I glossed over that line in Buffer.ToArray.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -42,53 +40,68 @@ internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
...
+                                {
+                                    if (count < MaxArrayLength)
+                                    {
+                                        newLength = MaxArrayLength;
+                                    }
+                                    else if ((newLength = unchecked(count + 2383)) < 0)
+                                    {
+                                        newLength = int.MaxValue; // Just let the attempt to create an array this size fail.
+                                    }
+                                }
```
stephentoub: While it's possible this could be an improvement in an obscure corner case, I'd rather see us keep the tried-and-true logic that's existed in ```List<T>``` for a long time: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L400

stephentoub:
cc: @VSadov

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -42,53 +40,68 @@ internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
...
+                                if ((uint)newLength > (uint)MaxArrayLength)
+                                {
+                                    if (count < MaxArrayLength)
+                                    {
+                                        newLength = MaxArrayLength;
+                                    }
+                                    else if ((newLength = unchecked(count + 2383)) < 0)
+                                    {
+                                        newLength = int.MaxValue; // Just let the attempt to create an array this size fail.
+                                    }
```
VSadov: defaulting to       newLength = count + 1;   
 seems slightly better here because we do not care much performance-wise (it is a degenerate case and will be slow regardless), but in case someone relied on 1element increases here, we could keep their code running.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
VSadov: @stephentoub   It seems that extending by 2383 , as long as we eventually go into +1 element mode, is at least as good as going into +1 element mode immediately once *2 mode no longer possible.

I think it might make sense to change List<T> as well, but not very high priority. 
It does feel like a corner case since any code that does a lot of Adds in +1 mode would be hardly runnable right now because of the copying costs.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
JonHanna: This was perhaps gilding the lily on my main idea, of a simple improvement akin to that of 6cf2b4c, while keeping the observable behaviour as-was.
I'll bring this PR back to that, tomorrow.

JonHanna:
Have removed the changes relating to the large byte-size-element arrays, leaving just the reduction in tests done by the code while remaining the same logic as List<T>, and the refactoring Buffer to use it.

src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
```diff
@@ -42,53 +38,64 @@ internal static T[] ToArray<T>(IEnumerable<T> source, out int length)
...
+                                // 2^30, since doubling to 2^31 is 1 larger than Int32.MaxValue.  In that case, we instead 
+                                // constrain the length to be MaxArrayLength (this overflow check works because of of the 
+                                // cast to uint).  Because a slightly larger constant is used when T is one byte in size, we 
+                                // could then end up in a situation where arr.Length is MaxArrayLength or slightly larger, such 
+                                // that we constrain newLength to be MaxArrayLength but the needed number of elements is actually 
+                                // larger than that.  For that case, we then ensure that the newLength is large enough to hold 
+                                // the desired capacity.  This does mean that in the very rare case where we've grown to such a 
+                                // large size, each new element added after MaxArrayLength will end up doing a resize.
+                                int newLength = count << 1;
```
JonHanna: Favouring `<< 1` over `* 2` here means that if this file is included in a project which is compiled with overflow checking, the results are the same.

AdamSpeight2008: @JonHanna  Does `<< 1` preserve sign of the value?

JonHanna: No, so the subsequent check catches the overflow as it would with checked times-two, and all is well.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3390,44 +3387,10 @@ internal Buffer(IEnumerable<TElement> source, bool queryInterfaces = true)
...
-                            items[count] = e.Current;
-                            ++count;
-                        }
-                    }
-                }
-            }
-
-            this.items = items;
-            this.count = count;
+            items = EnumerableHelpers.ToArray(source, out count);
```
stephentoub: The ```queryInterfaces``` parameter no longer has the same meaning it used to.  It previously dictated whether to query for any base/interfaces at all, or whether to always fall back to enumeration, but now regardless of the value ICollection will always be queried for inside of ToArray.  I think that's ok, because the one call site that sets it to false really only cares about whether we check for Iterator, but the name is a misnomer now (I guess it was previously, too, because Iterator is a base class rather than an interface).  Let's rename it to something like "queryIterator" or something better. :smile:

JonHanna: #2401 gets rid of it entirely, with iterators not all claiming to be able to produce arrays, and those that can only do so sometimes returning a null that can be checked for when they fail rather than calling back into Buffer.
That was done because separating the ability to create an array (which not all iterators can do anyway) to another interface let the ordered objects implement the same interface, but it might be worth doing anyway.

PR #3022, opened 2015-08-30
Fix GetFullPathName wrap.

JeremyKuhne:
GetFullPathName supports arbitrarily long paths, but doesn't
support extended syntax correctly. Change the wrap to handle
paths correctly and add a number of tests. Fix the common
test solution to honor build system configs.

Also add more StringBuilder overloads for the helpers.

Note that the incorrect wrapping wasn't exposed anywhere yet.
(we normalize out paths that would have created a problem)

@stephentoub, @weshaggard, @Priya91

### DIFF
--- a/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
+++ b/src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
@@ -6,9 +6,9 @@ using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
 
-partial class Interop
+internal partial class Interop
 {
-    partial class mincore
+    internal partial class mincore
     {
         /// <summary>
         /// WARNING: This method does not implicitly handle long paths. Use GetFullPathName.
@@ -22,19 +22,25 @@ partial class Interop
         [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
         private static extern int GetFullPathNamePrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
 
-        internal static int GetFullPathName(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
+        internal static int GetFullPathName(string path, int numBufferChars, [Out]StringBuilder buffer)
         {
-            bool wasExtended = PathInternal.IsExtended(path);
-            if (!wasExtended)
-            {
-                path = PathInternal.EnsureExtendedPrefixOverMaxPath(path);
-            }
-            int result = GetFullPathNamePrivate(path, buffer.Capacity, buffer, mustBeZero);
+            // GetFullPathName is backwards from most APIs. It handles long "DOS" paths without a problem. While it understands
+            // extended paths (\\?\), it roots them incorrectly, eating into the volume name with relative paths.
 
-            if (!wasExtended)
+            PathInternal.ExtendedType extendedType = PathInternal.RemoveExtendedPrefix(ref path);
+            int result = GetFullPathNamePrivate(path, buffer.Capacity, buffer, IntPtr.Zero);
+
+            if ((extendedType & PathInternal.ExtendedType.Extended) != 0)
             {
-                // We don't want to give back \\?\ if we possibly added it ourselves
-                PathInternal.RemoveExtendedPrefix(buffer);
+                // We want to *always* give back whatever we removed
+                if (extendedType.HasFlag(PathInternal.ExtendedType.ExtendedUnc))
+                {
+                    buffer.Insert(2, PathInternal.UncExtendedPrefixToInsert);
+                }
+                else
+                {
+                    buffer.Insert(0, PathInternal.ExtendedPathPrefix);
+                }
             }
             return result;
         }
--- a/src/Common/src/System/IO/PathInternal.Windows.cs
+++ b/src/Common/src/System/IO/PathInternal.Windows.cs
@@ -36,6 +36,14 @@ namespace System.IO
             (char)31, '*', '?'
         };
 
+        [Flags]
+        internal enum ExtendedType
+        {
+            NotExtended = 0x0,
+            Extended = 0x1,
+            ExtendedUnc = 0x2
+        }
+
         /// <summary>
         /// Returns true if the path is too long
         /// </summary>
@@ -91,7 +99,7 @@ namespace System.IO
         }
 
         /// <summary>
-        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative.
+        /// Adds the extended path prefix (\\?\) if not already present and if the path is not relative or a device (\\.\).
         /// </summary>
         internal static string EnsureExtendedPrefix(string path)
         {
@@ -106,18 +114,30 @@ namespace System.IO
         }
 
         /// <summary>
+        /// Adds the extended path prefix (\\?\) if not already present and if the path is not relative or a device (\\.\).
+        /// </summary>
+        internal static void EnsureExtendedPrefix(StringBuilder path)
+        {
+            if (IsExtended(path) || IsRelative(path) || IsDevice(path))
+                return;
+
+            // Given \\server\share in longpath becomes \\?\UNC\server\share
+            if (path.StartsWithOrdinal(UncPathPrefix))
+            {
+                path.Insert(2, PathInternal.UncExtendedPrefixToInsert);
+                return;
+            }
+
+            path.Insert(0, PathInternal.ExtendedPathPrefix);
+        }
+
+        /// <summary>
         /// Removes the extended path prefix (\\?\) if present.
         /// </summary>
         internal static string RemoveExtendedPrefix(string path)
         {
-            if (!IsExtended(path))
-                return path;
-
-            // Given \\?\UNC\server\share we return \\server\share
-            if (IsExtendedUnc(path))
-                return path.Remove(2, 6);
-
-            return path.Substring(4);
+            RemoveExtendedPrefix(ref path);
+            return path;
         }
 
         /// <summary>
@@ -136,11 +156,38 @@ namespace System.IO
         }
 
         /// <summary>
+        /// Remove the extended prefix from the given path. Return the type of path found.
+        /// </summary>
+        internal static ExtendedType RemoveExtendedPrefix(ref string path)
+        {
+            if (!IsExtended(path))
+                return ExtendedType.NotExtended;
+
+            // Given \\?\UNC\server\share we return \\server\share
+            if (IsExtendedUnc(path))
+            {
+                path = path.Remove(2, 6);
+                return ExtendedType.Extended | ExtendedType.Extended;
+            }
+
+            path = path.Substring(4);
+            return ExtendedType.Extended;
+        }
+
+        /// <summary>
         /// Returns true if the path uses the device syntax (\\.\)
         /// </summary>
         internal static bool IsDevice(string path)
         {
             return path != null && path.StartsWith(DevicePathPrefix, StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Returns true if the path uses the device syntax (\\.\)
+        /// </summary>
+        internal static bool IsDevice(StringBuilder path)
+        {
+            return path != null && path.StartsWithOrdinal(DevicePathPrefix);
         }
 
         /// <summary>
@@ -236,7 +283,7 @@ namespace System.IO
                 // UNC or simple rooted path (e.g. "\foo", NOT "\\?\C:\foo")
 
                 i = 1; //  Drive rooted (\foo) is one character
-                if (extendedUncSyntax || (length > 1 && (IsDirectorySeparator(path[1]))))
+                if (extendedUncSyntax || (length > 1 && IsDirectorySeparator(path[1])))
                 {
                     // UNC (\\?\UNC\ or \\), scan past the next two directory separators at most
                     // (e.g. to \\?\UNC\Server\Share or \\Server\Share\)
@@ -250,7 +297,7 @@ namespace System.IO
                 // Path is at least longer than where we expect a colon, and has a colon (\\?\A:, A:)
                 // If the colon is followed by a directory separator, move past it
                 i = volumeSeparatorLength;
-                if (length >= volumeSeparatorLength + 1 && (IsDirectorySeparator(path[volumeSeparatorLength]))) i++;
+                if (length >= volumeSeparatorLength + 1 && IsDirectorySeparator(path[volumeSeparatorLength])) i++;
             }
             return i;
         }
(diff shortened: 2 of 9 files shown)


JeremyKuhne:
@dotnet-bot test this please

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
```diff
@@ -22,19 +22,25 @@ partial class mincore
         [DllImport(Libraries.CoreFile_L1, EntryPoint = "GetFullPathNameW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
         private static extern int GetFullPathNamePrivate(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero);
 
-        internal static int GetFullPathName(string path, int numBufferChars, [Out]StringBuilder buffer, IntPtr mustBeZero)
+        internal static int GetFullPathName(string path, int numBufferChars, [Out]StringBuilder buffer)
```
stephentoub: As an aside, I'm starting to worry about the amount of logic that's crept into the interop files.  Are there any projects that depend on these interop definitions and that don't also need PathInternal, or that couldn't be made to easily use PathInternal?  I know we previously had a discussion about this, but I'm thinking now we should move this logic all out of the interop files into PathInternal, which the projects that need the logic can use.  That then also helps avoid what feels like a backwards dependency: the interop files using PathInternal.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
```diff
@@ -22,19 +22,25 @@ partial class mincore
...
-            }
-            int result = GetFullPathNamePrivate(path, buffer.Capacity, buffer, mustBeZero);
+            // GetFullPathName is backwards from most APIs. It handles long "DOS" paths without a problem. While it understands
+            // extended paths (\\?\), it roots them incorrectly, eating into the volume name with relative paths.
+
+            PathInternal.ExtendedType extendedType = PathInternal.RemoveExtendedPrefix(ref path);
+            int result = GetFullPathNamePrivate(path, buffer.Capacity, buffer, IntPtr.Zero);
 
-            if (!wasExtended)
+            if (extendedType.HasFlag(PathInternal.ExtendedType.Extended))
```
stephentoub: HasFlag is expensive.  It's be better to do:
```C#
if ((extendedType & ExtendedType.Extended) != 0)
```

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -106,10 +114,37 @@ internal static string EnsureExtendedPrefix(string path)
         }
 
         /// <summary>
+        /// Adds the extended path prefix (\\?\) if not already present, IF the path is not relative or a device (\\.\).
```
stephentoub: nit: ", IF" => "and if"

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -117,22 +152,26 @@ internal static string RemoveExtendedPrefix(string path)
...
-            return path.Substring(4);
+            return path.Remove(0, 4);
         }
 
         /// <summary>
-        /// Removes the extended path prefix (\\?\) if present.
+        /// Remove the extended prefix from the given path. Return the type of path found.
         /// </summary>
-        internal static StringBuilder RemoveExtendedPrefix(StringBuilder path)
+        internal static ExtendedType RemoveExtendedPrefix(ref string path)
```
stephentoub: I understand why you defined the signature this way, but I think it'd be more consistent with the rest of the overloads if you instead defined it as:
```C#
internal static string RemoveExtendedPrefix(string path, out ExtendedType originalType)
```

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -289,6 +336,40 @@ internal static bool IsRelative(string path)
...
+            {
+                // It isn't fixed, it must be relative.  There is no way to specify a fixed
+                // path with one character (or less).
+                return true;
+            }
+
+            if (IsDirectorySeparator(path[0]))
+            {
+                // There is no valid way to specify a relative path with two initial slashes
+                return !(IsDirectorySeparator(path[1]));
```
stephentoub: Nit: why the extra parens?

src/Common/tests/Common.Tests.csproj
```diff
@@ -8,14 +8,17 @@
...
-  </PropertyGroup>
+  <!-- Default configurations to help VS understand the configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'FreeBSD_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Linux_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'OSX_Release|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU' " />
```
stephentoub: I didn't think we were set up to enable one test project to be compilable for different platforms.  Does this actually work, meaning does the Linux CI actually use the assembly compiled for Linux?

src/Common/tests/Tests/Interop/Windows/mincore/GetFullPathName.Tests.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        [PlatformSpecific(PlatformID.Windows)]
+        [Theory]
+        [InlineData(@"C:\", @"C:\")]
+        [InlineData(@"C:\.", @"C:\")]
+        [InlineData(@"C:\..", @"C:\")]
+        [InlineData(@"C:\..\..", @"C:\")]
+        [InlineData(@"C:\A\..", @"C:\")]
+        [InlineData(@"C:\..\..\A\..", @"C:\")]
+        public static void GetFullPathName_Windows_RelativeRoot(string path, string expected)
```
stephentoub: That we have all of these Interop.Windows.mincore tests is I think another indicator that all of this logic should be in PathInternal rather than in Interop.mincore.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: My biggest concern is exposing direct access to the P/Invoke and unwrapped usages slipping through.

Isn't normalizing access to the APIs legitimate? I would think that we would be trying to enforce proper behavior at this level. For example: Why do we allow you to specify a StringBuffer and a size? It would prevent a lot of issues to only expose StringBuffer and get the Capacity in the wrappers.

There is also a concern about discovery.

I think of PathInternal as "interop" as well. There isn't anything in there that doesn't also directly apply to interacting with the Win32 APIs.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: Hmm. I explicitly didn't do that to try and not second guess the compiler/JITter. No worries though- I'll change it.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Thanks, I'll fix that.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: I started that way but it made usage ugly and I went this way. In addition, I haven't played with this, but I assume that when you don't care about the out that local allocation doesn't get optimized away. Do you know?

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Just because I started without the method call. I'll clean it up.

src/Common/src/System/IO/PathInternal.Windows.cs
stephentoub: > I assume that when you don't care about the out that local allocation doesn't get optimized away. Do you know?

What local allocation do you think is getting allocated with the signature one way and not the other way?

src/Common/tests/Common.Tests.csproj
JeremyKuhne: I don't think it has been switched over, but I don't know for sure at this point. I know it will and should.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
stephentoub: > 

FWIW, this doesn't really have much to do with the compiler.  HasFlag as implemented at https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Enum.cs#L868-L877 and https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/vm/reflectioninvocation.cpp#L3707-L3747 is simply much more expensive than a bit-wise-and plus a comparison.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Alternatively I could switch them all to ref. Maybe. Hmm... I need to think about that. :)

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
stephentoub: > My biggest concern is exposing direct access to the P/Invoke and unwrapped usages slipping through.

Aren't there other cases where some of the raw P/Invokes have been left as internal rather than private?

> Isn't normalizing access to the APIs legitimate?

To a limited degree, sure.  But we're trying to keep these interop files as close to the original as possible, with the few exceptions being places where we want to make small tweaks, such as hiding a particular argument and always filling in the same default value.

> I would think that we would be trying to enforce proper behavior at this level.

At this level, the proper behavior is what the underlying OS provides.  Imagine we had the ability to #include a .h file directly into a C# assembly... at that point the need for these Interop.*.cs files would go away and we'd be left with the PathInternal files that would include this behavior.  We obviously can't just #include them, though, which is why the Interop.*.cs files exist, to provide that level of functionality.

> Why do we allow you to specify a StringBuffer and a size? It would prevent a lot of issues to only expose StringBuffer and get the Capacity in the wrappers.

I don't understand the comparison.  DllImports are able to marshal both strings and StringBuilders as char*, and the runtime can't magically assume that a char* and an int next to each other in a DllImport are related, so the DllImport needs to be written to take both the thing to be marshaled as a char* and the thing to be marshaled as the length.  That said, I'd actually be ok with us making such a DllImport private and exposing as internal a wrapper that did the appropriate delegation to the underlying P/Invoke... that's an example of minimal smoothing over to hide an argument, as I noted above.

> There is also a concern about discovery.

I'm fine if there's a comment in the P/Invoke files stating that the PathInternal functions should be used instead.  Beyond that, I'm also not particularly concerned about discovery as these functions are meant to be used by only a small handful of lower-level assemblies in corefx... anything above those assemblies should be using those assemblies, not the P/Invokes.

> I think of PathInternal as "interop" as well.

In which case, why not have a clean separation between the files that define the P/Invokes to the corresponding native functions and the files that define the wrappers that do the propery path normalization?

JeremyKuhne: > Aren't there other cases where some of the raw P/Invokes have been left as internal rather than private?

Yes, but they either don't need normalization or haven't gotten it yet (which was my comment on wrapping to avoid needing to explicitly specify the StringBuilder size outside of interop).

>at that point the need for these  Interop.*.cs  files would go away and we'd be left with the PathInternal files that would include this behavior

I could simply remove these from interop and put them directly in the PathInternal. That does, however, risk reintroduction and missing the important wrapping behavior. I suppose we could keep empty stub files that say "use PathInternal".

I have a hard time seeing the value beyond aesthetics. The way I've got it doesn't look as clean, but it has no chance of accidental misuse. I suppose I can buy that it makes it more difficult to enforce restricting the interop layer to normalizing access to the Windows APIs (because it is harder to audit). As such I'd be fine with stubbing the files and pushing the P/Invokes (in this case) to PathInternal.

stephentoub: > I could simply remove these from interop and put them directly in the PathInternal.

No, we should keep the P/Invokes in the ```Interop.*.cs``` files.  The point of these files is to have them be callable versions of the OS functions, and nothing more, as if we were able to #include .h files.  The exposed functions should have the same behavior as the OS functions, so that someone calling them can look at MSDN and see the documented behavior.  Functionality above these files can then layer on additional/different behaviors.  Once we start letting arbitrary changes in to these ```Interop.*.cs``` files, they lose their meaning.

src/Common/src/System/IO/PathInternal.Windows.cs
JeremyKuhne: Here is the example:

```C#
        static void Test()
        {
            string foo = "foo";
            Sample.DoSomething(ref foo);
        }

        static void TestToo()
        {
            bool dontCare;
            string foo = "foo";
            foo = Sample.DoSomethingToo(foo, out dontCare);
        }

        public static bool DoSomething(ref string value)
        {
            value.ToUpper();
            return value.Length > 5;
        }

        public static string DoSomethingToo(string value, out bool result)
        {
            result = value.Length > 5;
            return value.ToUpper();
        }
```

You end up with the following in IL from the callers:
```
.method private hidebysig static void  Test() cil managed
{
  // Code size       15 (0xf)
  .maxstack  1
  .locals init ([0] string foo)
  IL_0000:  ldstr      "foo"
  IL_0005:  stloc.0
  IL_0006:  ldloca.s   foo
  IL_0008:  call       bool ConsoleApplication2.Sample::DoSomething(string&)
  IL_000d:  pop
  IL_000e:  ret
}

.method private hidebysig static void  TestToo() cil managed
{
  // Code size       14 (0xe)
  .maxstack  2
  .locals init ([0] bool dontCare)
  IL_0000:  ldstr      "foo"
  IL_0005:  ldloca.s   dontCare
  IL_0007:  call       string ConsoleApplication2.Sample::DoSomethingToo(string,
                                                                         bool&)
  IL_000c:  pop
  IL_000d:  ret
}
```
The maxstack+1 is what I was referring to.

src/Common/src/Interop/Windows/mincore/Interop.GetFullPathNameW.cs
JeremyKuhne: > Once we start letting arbitrary changes in to these  Interop.*.cs  files, they lose their meaning.

I agree that we shouldn't let arbitrary changes in- but these are specific changes. Wrapping that ensures correct and/or safe usage of the underlying API is within scope of what the interop layer should have.

As I said, I can't understand what value there in forcing this code into another layer is beyond aesthetics. I can, however, easily fathom how having the correctness method outside of interop would introduce nasty and subtle bugs- so I'm seriously uncomfortable doing so.

stephentoub: It's not aesthetics.  If we switch to something like MCG which generates these interop functions for us, we'll delete these ```Interop.*.cs``` files, and there goes your special handling.  If someone looks at a function named Zoo in these Interop files, they should be able to assume that Zoo behaves identically to Zoo as defined by the platform, as these functions are meant to be EXACTLY what the behavior of the platform, simply making the platform functionality available to C#; these behavior changes nullify that.

JeremyKuhne: What is MCG?

stephentoub: See http://blogs.msdn.com/b/dotnet/archive/2014/06/13/net-native-deep-dive-debugging-into-interop-code.aspx

JeremyKuhne: Thanks. I'll give it a read.

I'm assuming that you're saying that we should have only pure P/Invokes and no other methods in interop. Do you have proposals for discoverable places to put methods that wrap proper invocation per the documentation?

JeremyKuhne:
@dotnet-bot test this please

JeremyKuhne:
@dotnet-bot test this please

JeremyKuhne:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

JeremyKuhne:
Will follow up with wrapping changes to move some of the logic from interop to address @stephentoub's concerns.

PR #3023, opened 2015-08-30
XLinq: Throw ArgumentNullException from Extensions.Nodes immediately instead of lazily

justinvp:
I noticed one of the extension methods in `System.Xml.Linq` was throwing `ArgumentNullException` from an iterator method, which means the exception is thrown from the enumerator the first time `MoveNext` is called, rather than the intended behavior of throwing immediately when the extension method is called.

This change fixes the `Nodes` extension method to throw when the method is called, to behave the way it was intended and to be consistent with the behavior of the other extension methods on this class.

This is a minor behavior change, but I think it's worth making in .NET Core because this is a developer error (should not be caught in a try/catch) and this is making the behavior consistent with the behavior of the other extension methods on this class (as well as all the extension methods provided by Linq itself).

I was going to add a test, but wasn't sure where to add it. Maybe open an up-for-grabs issue to add `Assert.Throws` tests for these extension methods, or add a work item to #642?

### DIFF
--- a/src/System.Xml.XDocument/src/System/Xml/Linq/Extensions.cs
+++ b/src/System.Xml.XDocument/src/System/Xml/Linq/Extensions.cs
@@ -118,6 +118,11 @@ namespace System.Xml.Linq
         public static IEnumerable<XNode> Nodes<T>(this IEnumerable<T> source) where T : XContainer
         {
             if (source == null) throw new ArgumentNullException("source");
+            return NodesIterator(source);
+        }
+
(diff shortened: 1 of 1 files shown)


stephentoub:
LGTM.

@krwq, @weshaggard, you ok with the small behavioral change (the ArgumentNullException being thrown out of Nodes rather than out of the MoveNext from the IEnumerable returned from Nodes)?

weshaggard:
Yes I'm fine with it and it seems more consistent with the other methods.

stephentoub:
Thanks, Wes.

PR #3024, opened 2015-08-30
Using SIMD Acceleration for System.Linq Operations

TPoise:
As a  sample, I've made a modificatioin to the System.Linq.Enumerable.Add<int>() method to show 2-4x speedup when using SIMD for large enumerable sources.  I've done this only for the integer addition, and I'd like to do it for the other primitive types if this acceptable.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -4,12 +4,16 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Numerics;
 using System.Threading;
 
 namespace System.Linq
 {
     public static class Enumerable
     {
+
+        private static readonly int SIMD_THRESHOLD = 64;
+
         public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate)
         {
             if (source == null) throw Error.ArgumentNull("source");
@@ -1610,10 +1614,61 @@ namespace System.Linq
             int sum = 0;
             checked
             {
-                foreach (int v in source) sum += v;
+                if (Vector.IsHardwareAccelerated && source.Count() >= SIMD_THRESHOLD)
+                    sum = VectorSum(source);
+                else
+                    foreach (int v in source) sum += v;
             }
             return sum;
         }
+
+        private static int VectorSum(IEnumerable<int> source) 
+        {
+            int sum = 0;
+
+            int[] sourceArray = source.ToArray();
+            int vectorLength = Vector<int>.Count;
+  
+            // Normalize the arrays to be the same divisor as of the SIMD Length
+            // Example: if the "Vector<int>.Count" is 4, then the array size need to be a multiple of 4
+            // Add zeros to the end of the array if necessary.
+            int correctionSize = (sourceArray.Length % vectorLength);
+            int midpoint = (sourceArray.Length / 2);
+            int correctedMidpoint = midpoint + correctionSize;
+
+            // create two arrays with a multiple of the vector length
+            int[] arrayA = new int[correctedMidpoint];
+            int[] arrayB = new int[correctedMidpoint];
+
+            // Copy over the source data from the original IEnumerable source.
+            Array.Copy(sourceArray, 0, arrayA, 0, midpoint);
+            Array.Copy(sourceArray, midpoint, arrayB, 0, midpoint);
+
+            // if it's not even, then go ahead and "add" the final number to the sum variable
+            // because we're only going to vectorize the "even" part of the arrays.
+
+            if (sourceArray.Length % 2 != 0)
+                sum = sourceArray[sourceArray.Length - 1];
+
+            Vector<int> vectorA;
+            Vector<int> vectorB;
+
+            for (int i = 0; i < midpoint; i += vectorLength)
+            {
+
+                vectorA = new Vector<int>(arrayA, i);
+                vectorB = new Vector<int>(arrayB, i);
+
+                vectorA = Vector.Add(vectorA, vectorB);
+
+                // you can't easily do this with generics without hacks because Operator+ isn't supported
+                for (int j = 0; j < vectorLength; j++) sum += vectorA[j];
+                
+            }
+
+            return sum;
+        }
+        
 
         public static int? Sum(this IEnumerable<int?> source)
         {
--- a/src/System.Linq/tests/EnumerableTests.cs
+++ b/src/System.Linq/tests/EnumerableTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using Xunit;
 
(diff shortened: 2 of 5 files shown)


TPoise:
Ideally this should go to the "future" branch.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -4,12 +4,16 @@
...
 using System.Collections.Generic;
+using System.Numerics;
 using System.Threading;
 
 namespace System.Linq
 {
     public static class Enumerable
     {
+
+        private static readonly int SIMD_THRESHOLD = 64;
```
Maxwe11: >SIMD_THRESHOLD

doesn't match guidelines

TPoise: Ideally how should I handle?  You don't need vectorization for small
workloads.  In my testing, roughly 64 a good size before you apply
vectorization.   Or is it the code styling?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1610,10 +1614,61 @@ public static int Sum(this IEnumerable<int> source)
             int sum = 0;
             checked
             {
-                foreach (int v in source) sum += v;
+                if (Vector.IsHardwareAccelerated && source.Count() >= SIMD_THRESHOLD)
```
michael-damatov: The _Count_ method would probably introduce the additional iteration over the _source_, which is not intended here.

mikedn:
SIMD can't be used to implement LINQ's Sum as doing that breaks the overflow checking behavior. Moreover, such an optimization is in general questionable because it trades memory for performance and it's not easy to measure the exact impact of doing that in GC based system.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1610,10 +1614,61 @@ public static int Sum(this IEnumerable<int> source)
...
+                    foreach (int v in source) sum += v;
+            }
+            return sum;
+        }
+
+        private static int VectorSum(IEnumerable<int> source) 
+        {
+            int sum = 0;
+
+            int[] sourceArray = source.ToArray();
```
JonHanna: This goes via `Buffer` which creates an internal array which might have excess zeros, and the creates yet another array if such excess does exist. Using `Buffer` directly could reduce the number of allocations needed.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Potentially the source may not be replayable, and so it would end up with `Sum` throwing an exception or returning `0` because on the subsequent attempt there were no elements.

TPoise: I'll remove the source.Count() check.  Good observation.

JonHanna:
Even if this was changed to use `Buffer` directly rather than `ToArray()`, it's still going to mean creating an array and copying to it.
With many sources `Sum()` is currently faster than that step is.

TPoise:
That was my thinking behind the "threshold".  For small collections (<64)
this is slower. But as the source grows, the speedup from the simd
acceleration kicks in.

JonHanna:
In most cases ToArray is O(n) in time and space, while Sum is O(n) in time but O(1) in space. Also, the constant costs are such that the time cost of ToArray is either greater than Sum or close to it. In any such case this change will slow things down, even if the bug that `Enumerable.Repeat(int.MaxValue, 100).Sum()` will no longer throw like it should could be fixed.

src/System.Linq/tests/EnumerableTests.cs
```diff
@@ -172,6 +173,57 @@ public void Distinct()
...
+
+            int sum = 0;
+            int simdSum = 0;
+            // Sum those numbers using the tradition LINQ
+            Stopwatch sw = Stopwatch.StartNew();
+
+            for (int i = 0; i < 1000; i++)
+            {
+                // hard-coding in the "traditional" addition method from IEnumerable<int>.Sum()
+                foreach (int num in array)
```
JonHanna: This is not quite the same as the Linq method. This would use `System.Array+SZArrayEnumerator` whereas Linq would be using `System.SZArrayHelper+SZGenericArrayEnumerator<int>`. While the results are likely not very different, you're measuring the cost of something similar to the Linq approach here, not the linq approach.

src/System.Linq/tests/EnumerableTests.cs
```diff
@@ -172,6 +173,57 @@ public void Distinct()
...
+                    sum += num;
```
JonHanna: Likewise, this isn't checked, so it's not measuring the same thing.

src/System.Linq/tests/EnumerableTests.cs
```diff
@@ -172,6 +173,57 @@ public void Distinct()
...
+
+            sw.Stop();
+            
+            double linqMS = sw.Elapsed.TotalMilliseconds;
+            Debug.WriteLine("LINQ Method: " + sw.Elapsed.TotalMilliseconds.ToString() + "ms");
+
+            // Sum those numbers using SIMD (must have x64-release mode turned on for RyuJIT)
+            sw = Stopwatch.StartNew();
+            for (int j = 0; j < 1000; j++)
+                simdSum = array.Sum();
```
JonHanna: What happens with `array.Where(i => true).Sum()`?

TPoise:
To ease some concern, would it be better to have a seperate API method?  

E.g., **VectorizedSum()** or something like it?  The end user would know then the trade-offs between requiring additional memory allocations in exchange for faster performance, and would know (through documentation) that it works better under larger sets of data.

JonHanna:
Since it gives a different answer to calling `Sum()` and is slower than `Sum()` in the majority of cases, it should probably be considered as very different to `Sum()` indeed.

It might also be well to have its signature accepting `int[]` rather than `IEnumerable<int>` as it's only faster when the source is an array (`List<int>` begins to merge so there might be a large size above which it does better even so forcing a call to `List<T>.ToArray()` instead of `Enumerable.ToArray()` would give a slight further gain, most other `IEnumerable<int>` will have returned an answer from `Sum()` before `VectorSum` has got to line 1630 in the current code).

If it belongs anywhere, it's probably closer in System.Numerics.Vectors rather than in System.Linq.

stephentoub:
@TPoise, thank you for the contribution/suggestion.  I agree with others' comments, though, that this does not belong in System.Linq.  Personally, I think a higher-level library of ```T[]```-based and/or ```Vector<T>```-based algorithms built on top of ```System.Numerics.Vectors``` could be quite interesting, but instead of a PR there should be an issue opened to discuss what that would look like.  corefxlabs might also be a more appropriate place for an initial implementation of such a library.  cc: @KrzysztofCwalina, @mellinoe, @CarolEidt

TPoise:
Thanks all for the comments and feedback.  This was my first PR but it won't be my last, I want to contribute back to the dotnet/* projects in some way or another.  I'll grab one of the "community" tagged issues.

PR #3026, opened 2015-08-31
Tests for coreclr/pull/1460

bbowyersmyth:
Covers both IList and IEnumerable paths of String.Join.

This only needs to be merged if https://github.com/dotnet/coreclr/pull/1460 goes ahead


stephentoub:
LGTM

PR #3027, opened 2015-08-31
Fix tm_mon assignment in MakeTimeT

bartonjs:
tm_mon is the zero-indexed month of the year, but the callers use the
one-indexed month of the year.  This makes things like expirations come
up a month sooner than they should.

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -52,7 +52,7 @@ MakeTimeT(
 {
     struct tm currentTm;
     currentTm.tm_year = year - 1900;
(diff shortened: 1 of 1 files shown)


bartonjs:
StartSSL's "StartCom Class 1 Primary Intermediate Server CA" CA has a one month validity period on their current CRL; and my current CRL checking work-in-progress reported that the freshly downloaded CRL was expired already.

Yay for short validity periods.  I opened issue #3028 to track getting good tests for this (it requires some prep to keep it running without needing the network or persistent storage).

bartonjs:
@dotnet-bot test this please

nguerrera:
LGTM

PR #3029, opened 2015-08-31
Make LowLevelDictionary and LowLevelList more low-level

JonHanna:
Only track low-level collection version on debug.

Dictionary and List have, like many other collections a _version field that is
used to prohibit enumerating on the collection if the collection is changed in
the meantime, simplifying the complications of such a state to a simple ban.

This is generally not an exception one meaningfully recovers from; one changes
the code so it doesn't happen.

The low-level versions used internally in some assemblies have this same
functionality, but as it is only used internally, one can ensure that the case
it guards against never happens.

Change the checks to only be present in D‎EBUG builds, where they can still
serve as assertions, and hence reduce the work done in release builds.

Only test if key type is IEquatable<T> once per type, rather than once
per Equal‎s()

Reduce indirections in most common cases: Extra lookup if Current is accessed
repeatedly, but this doesn't happen.

### DIFF
--- a/src/Common/src/System/Collections/Generic/LowLevelDictionary.cs
+++ b/src/Common/src/System/Collections/Generic/LowLevelDictionary.cs
@@ -28,12 +28,12 @@ namespace System.Collections.Generic
         private const int DefaultSize = 17;
 
         public LowLevelDictionary()
-            : this(DefaultSize, new DefaultComparer<TKey>())
+            : this(DefaultSize, DefaultComparer)
         {
         }
 
         public LowLevelDictionary(int capacity)
-            : this(capacity, new DefaultComparer<TKey>())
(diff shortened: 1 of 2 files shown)


nguerrera:
cc @weshaggard 

I believe the purpose of the WithIXxx variants is to hand out interface implementations that are expected to match the behavior of previous implementations on other platforms that used the non-low-level counterparts. Once a reference is given out, we no longer control enumeration nor modification, so debug-only checking would be a behavioral change even if we never do it ourselves.

JonHanna:
I could perhaps debate that point, but I've realised that there is an (albeit wrapped) use more exposed than I thought when I first went through this list. There may be something to my idea here, but if there is, this stab at it certainly isn't as ready as I thought this morning.

PR #3030, opened 2015-08-31
Fully import legacy Take and TakeWhile tests.

JonHanna:
Convert Take and TakeWhile tests imported in #2874 into a more directly
XUnit-ian style

TakeWhile6.Test6() was crippled in the original, probably because it causes an
OutOfMemoryException. Fixed this, but it is very slow, so have marked as
OuterLoop

Added tests on ArgumentNullException being thrown where appropriate.

Also fix accidential omission of assertion from f27e77d.


stephentoub:
LGTM.  Thanks, @JonHanna.

PR #3031, opened 2015-08-31
Fully import legacy Single and SingleOrDefault tests

JonHanna:
Convert Single and SingleOrDefault tests imported in #2874 into a more directly
XUnit-ian style

All the previous Single tests added in f8faadf are duplicated by these bar
FindSingleMatch, so they were removed.

Added SingleOrDefault version of FindSingleMatch.

Added tests on ArgumentNullException being thrown where appropriate.


stephentoub:
LGTM.  Thanks, @JonHanna.

PR #3036, opened 2015-08-31
Remove use of LowLevelList<T> from X509Certificates

justinvp:
`System.Security.Cryptography.X509Certificates` already has a dependency on `System.Collections` and is already making use of `List<T>`. This PR removes all uses of `LowLevelList<T>` in this assembly. In some cases, switching to an array was sufficient.

cc @bartonjs


bartonjs:
LGTM, thanks @justinvp,

stephentoub:
LGTM2

justinvp:
@dotnet-bot test this please

PR #3038, opened 2015-08-31
Moving the Unix calls to getrlimit and setrlimit over to the PAL layer

sokket:
Moving the Unix calls to getrlimit and setrlimit over to the PAL layer. Also fixing a UT failure in the OSX Process tests where OS X throw Win32Exception and other platforms threw InvalidOperationException.

/cc: @nguerrera


src/Common/src/Interop/Unix/System.Native/Interop.ResourceLimits.cs
```diff
@@ -0,0 +1,37 @@
...
+            RLIMIT_MEMLOCK  = 7,        // Locked-in-memory address space
+            RLIMIT_NPROC    = 8,        // Number of processes
+            RLIMIT_NOFILE   = 9,        // Number of open files
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct RLimit
+        {
+            internal ulong CurrentLimit;
+            internal ulong MaximumLimit;
```
stephentoub: How do we want to handle RLIM_INFINITY?

src/Native/System.Native/pal_process.cpp
```diff
@@ -0,0 +1,69 @@
...
+        case RlimitResources::PAL_RLIMIT_RSS:
+            return RLIMIT_RSS;
+        case RlimitResources::PAL_RLIMIT_MEMLOCK:
+            return RLIMIT_MEMLOCK;
+        case RlimitResources::PAL_RLIMIT_NPROC:
+            return RLIMIT_NPROC;
+        case RlimitResources::PAL_RLIMIT_NOFILE:
+            return RLIMIT_NOFILE;
+        default:
+            assert(!"Unknown RLIMIT value");
```
stephentoub: If the assert is only compiled into debug builds, what happens here for release?

src/Native/System.Native/pal_process.cpp
```diff
@@ -0,0 +1,69 @@
...
+    assert(limits != nullptr);
+
+    int32_t platformLimit = ConvertRlimitResourcesToPlatformValue(resourceType);
+    rlimit internalLimit;
+    int result = getrlimit(platformLimit, &internalLimit);
+    if (result == 0)
+    {
+        limits->CurrentLimit = internalLimit.rlim_cur;
+        limits->MaximumLimit = internalLimit.rlim_max;
+    }
```
stephentoub: else do we want to zero-out limits, and then we can use out rather than ref in the C# DllImport?

src/Native/System.Native/pal_process.cpp
sokket: Ah good catch; I'll do what errno does and return -1

src/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
```diff
@@ -90,8 +92,16 @@ private void GetWorkingSetLimits(out IntPtr minWorkingSet, out IntPtr maxWorking
             minWorkingSet = IntPtr.Zero;
 
             // Get the max working set size
-            Interop.libc.rlimit info = Interop.libc.getrlimit(Interop.libc.RLIMIT_Resources.RLIMIT_RSS);
-            maxWorkingSet = new IntPtr(Convert.ToInt64(info.rlim_cur));
+            Interop.Sys.RLimit limit = new Interop.Sys.RLimit();
+            if (Interop.Sys.GetRLimit(Interop.Sys.RlimitResources.RLIMIT_RSS, ref limit) == 0)
```
stephentoub: Per previous comment, I think ideally the signature here would use ```out```, and limit wouldn't need to be initialized.

src/Native/System.Native/pal_process.cpp
```diff
@@ -0,0 +1,69 @@
...
+            return RLIMIT_NPROC;
+        case RlimitResources::PAL_RLIMIT_NOFILE:
+            return RLIMIT_NOFILE;
+        default:
+            assert(!"Unknown RLIMIT value");
+    }
+}
+
+extern "C"
+int GetRLimit(
```
nguerrera: return type should be int32_t

src/Common/src/Interop/Unix/System.Native/Interop.ResourceLimits.cs
sokket: Shouldn't it just be ulong.MaxValue?

src/Native/System.Native/pal_process.cpp
sokket: Oops thanks!

src/Native/System.Native/pal_process.cpp
nguerrera: A trick is to put this assert after the switch instead of default, then clang will warn if not all the RlimitResources enum values are handled. See errno shim for example.

src/Common/src/Interop/Unix/System.Native/Interop.ResourceLimits.cs
stephentoub: Is it going to be defined that way on all platforms, or do we need to do something special to make it so for consistency?

src/Native/System.Native/pal_process.cpp
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "../config.h"
+#include "pal_process.h"
+#include <sys/resource.h>
+#include <assert.h>
+
+// Each platform type has it's own RLIMIT values but the same name, so we need
+// to convert our standard types into the platform specific ones.
+int32_t ConvertRlimitResourcesToPlatformValue(RlimitResources value)
```
nguerrera: This should be static since we don't export it.

Also, I suggest using the same naming convention as in the errno shim ConvertRlimitResourcesPalToPlatform.

src/Native/System.Native/pal_process.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
```
nguerrera: We should probably put this together with ForkAndExecProcess. I'm ok with the name pal_process, but maybe then rename pal_exec to pal_process.

src/Native/System.Native/pal_process.h
```diff
@@ -0,0 +1,54 @@
...
+    RLimit*             limits);
+
+/**
+ * Set the soft and hard limits for the specified resource. 
+ * Only a super-user can increase hard limits for the current process.
+ * Returns 0 on success; returns -1 on failure and errno is set to the error reason.
+ */
+extern "C"
+int SetRLimit(
+    RlimitResources     resourceType,
```
nguerrera: You have mixed casing of RLimit and Rlimit. Pick one. I prefer RLimit visually, but don't care which you pick so long as it's consistent.

src/Native/System.Native/pal_process.h
```diff
@@ -0,0 +1,54 @@
...
+    uint64_t            MaximumLimit;
+};
+
+
+/**
+ * Get the current limit for the specified resource of the current process.
+ * Returns 0 on success; returns -1 on failure and errno is set to the error reason.
+ */
+extern "C"
+int GetRLimit(
```
nguerrera: int32_t instead of int again. (I'm not going to point out all of them, just make sure to use int32_t not int in all exported signature in header and impl file.)

sokket: Fixed

src/Native/System.Native/pal_process.h
sokket: Fixed

src/Native/System.Native/pal_process.cpp
sokket: Fixed

src/Native/System.Native/pal_process.cpp
sokket: Ah thanks for the example!

src/Common/src/Interop/Unix/System.Native/Interop.ResourceLimits.cs
sokket: Looks like it differs slightly from platform to platform...I'll do a check in the C++ layer if it equals the C# ulong MaxValue (which should be (ulong)-1) and if so, convert the value to RLIM_INFINITY. This will be okay because on Ubuntu, that is the correct value; on OS X, RLIM_INFINITY is lower than that value so we should reduce it to conform with the expected value

src/Native/System.Native/pal_process.cpp
sokket: Sure

src/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
sokket: Sure, fixed

src/Native/System.Native/pal_process.cpp
sokket: Sure, will consolidate everything into one file

sokket:
@nguerrera New version is up, please take a look when you have a minute :)

src/Native/System.Native/pal_process.cpp
```diff
@@ -142,3 +142,92 @@ int32_t ForkAndExecProcess(
...
+    assert(!"Unknown RLIMIT value");
+    return -1;
+}
+
+// Because RLIM_INFINITY is different per-platform, use the max value of a uint64 (which is RLIM_INFINITY on Ubuntu)
+// to signify RLIM_INIFINITY; on OS X, where RLIM_INFINITY is slightly lower, we'll translate it to the correct value here.
+static
+uint64_t ConvertFromManagedRLimitInfinityToPalIfNecessary(uint64_t value)
+{
+    if (value == ((uint64_t)-1))
```
nguerrera: Nit: I think UINT64_MAX would be nicer than this cast.

src/Native/System.Native/pal_process.cpp
```diff
@@ -142,3 +142,92 @@ int32_t ForkAndExecProcess(
...
+        return value;
+}
+
+// Because RLIM_INFINITY is different per-platform, use the max value of a uint64 (which is RLIM_INFINITY on Ubuntu)
+// to signify RLIM_INIFINITY; on OS X, where RLIM_INFINITY is slightly lower, we'll translate it to the correct value here.
+static
+uint64_t ConvertFromNativeRLimitInfinityToManagedIfNecessary(uint64_t value)
+{
+    if (value == RLIM_INFINITY)
+        return ((uint64_t)-1);
```
nguerrera: Ditto nit: UINT64_MAX vs cast.

src/Native/System.Native/pal_process.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Set the soft and hard limits for the specified resource. 
+ * Only a super-user can increase hard limits for the current process.
+ * Returns 0 on success; returns -1 on failure and errno is set to the error reason.
+ */
+extern "C"
+int32_t SetRLimit(
+    RLimitResources     resourceType,
+    const RLimit* const limits);
```
nguerrera: Nit: const RLimit* limits is sufficient. The second level is sort of like passing const int32_t, which is strange -- why do we care if impl moves its copy of pointer to point elsewhere?

src/Native/System.Native/pal_process.cpp
sokket: Good idea

src/Native/System.Native/pal_process.cpp
sokket: Good idea

src/Native/System.Native/pal_process.h
sokket: I prefer the const Type const * syntax because it makes it perfectly clear that this function is a purely read-only function; nothing about the pointer or the pointers data is changed in any way. We could leave off either consts and be fine, but by keeping both consts, we stick to strict C++ const correctness and make it very clear to any readers that the implementation of SetRLimit is strictly read-only. Does that make sense?

src/Native/System.Native/pal_process.h
```diff
@@ -0,0 +1,79 @@
...
+    PAL_RLIMIT_NOFILE       = 9,        // Number of open files
+};
+
+/**
+ * The current and maximum resource values for the current process.
+ * These values are depict the resource according to the above enum.
+ */
+struct RLimit
+{
+    uint64_t            CurrentLimit;
```
nguerrera: Nit: style choice,  but I wouldn't create so much white space to line up things that aren't even in the same struct/function.

sokket: Fixed

src/Native/System.Native/pal_process.cpp
```diff
@@ -142,3 +142,92 @@ int32_t ForkAndExecProcess(
...
+
+extern "C"
+int32_t SetRLimit(
+    RLimitResources     resourceType,
+    const RLimit* const limits)
+{
+    assert(limits != nullptr);
+
+    int32_t platformLimit = ConvertRLimitResourcesPalToPlatform(resourceType);
+    rlimit internalLimit = { ConvertFromManagedRLimitInfinityToPalIfNecessary(limits->CurrentLimit), ConvertFromManagedRLimitInfinityToPalIfNecessary(limits->MaximumLimit) };
```
nguerrera: Is current before maximum guaranteed by POSIX? Safer to set these by name. I think it would be more readable to have rlimit -> RLimit and RLimit -> rlimit conversion helpers, which could take care of that.

nguerrera:
@socket Thanks. I left couple of nits and one new comment with a bit more substance...

src/Native/System.Native/pal_process.h
nguerrera: The callee gets a copy of the pointer so there is no impact on caller if it moves it to another const RLimit. IMHO, that's why it's weird to declare functions as such. Like I said, we don't pass const ints. It's putting implementation constraints that are of no consequence to the caller in the signature.

Is the argument that this make it less necessary for reader to remember difference between const T* and T* const?

src/Native/System.Native/pal_process.cpp
sokket: Good idea

src/Native/System.Native/pal_process.h
sokket: True that the pointer is copied and only the copy can be changed, which has no effect on the caller, but the extra const means that attempting to change the pointer (which we agreed would be useless since it's a copy) would result in a compilation error. I was trying to follow the ISO const-correctness guide here https://isocpp.org/wiki/faq/const-correctness#const-ptr-vs-ptr-const where it says that adding a couple extra keystrokes for extra consts in return for getting more metadata about the function is, in their opinion, worth it. I can remove the const if you would like :)

nguerrera: I see what you mean. We should probably look at making consistent guidelines and using throughout. In the meantime, it's OK by me either way.

sokket: I opened issue #3051 to track the work of creating a standard and making sure the PAL adheres to said standard; assigned the issue to myself.

sokket:
Merging, thanks!

PR #3041, opened 2015-09-01
Allow a SwitchExpression to have no Cases.

### LINKED ISSUE #1877: [System.Linq.Expressions] Allow Switch with empty cases list

JonHanna:
Fixes #1877

If type of is not specified, it will be that of default body, or
null if there is no default body.

If there is no default body, and the type is specified, it must be void.

On emitting, the result is the same as if the default body was emitted.


JonHanna:
As well as adding the `Reduce` implementation making sense in itself, it's possible that something out there that would chock on a zero-case select might very well survive if it's in the habit of reducing expressions before visiting them.

stephentoub:
cc: @VSadov

JonHanna:
@mikedn not the right thing, or certainly not with all paths, so very good point. I shall rectify that.

JonHanna:
@mikedn that case is now covered, with relevant tests also.

mikedn:
@JonHanna Thanks, it looks correct now.

However, I took a closer look at all this and I'm not entirely sure that this is the right approach. It appears to me that `Reduce` is intended to transform the expression in such way that it can be compiled while your use of `Reduce` looks more like an optimization. You could argue that you actually use `Reduce` the way it was intended to but you also changed `EmitSwitchExpression` to make it support switches without cases and thus made reduction unnecessary, at least in theory It turns out that the compiler has special knowledge about reducible nodes and only reduces those before compiling - see `StackSpiller.RewriteReducibleExpression` and its only caller.

There's also the pesky interpreter which isn't used by CoreCLR but it is used by .NET Native. That one might need to be updated as well.

JonHanna:
@mikedn EmitSwitchExpression is the main change here, and what compilation would use. 
Reduce was changed primarily for the possibility that of the set of Expression processing code out there that might choke on zero cases, some of them might try reducing reducible nodes before progressing. In such a case, they will have turned it into an expression they can handle.
That some code might choke on zero cases was a concern raised in the issue, and while those cases wouldn't have worked previously (so nothing has been lost) it seemed prudent to consider a possibility that might increase the number of situations where it does work.

mikedn:
@JonHanna Well, the idea is that you could get rid of the EmitSwitchExpression changes and keep everything in SwitchExpression.Reduce. Granted, that means that you need to fix StackSpiller instead to not make assumptions about which node types are reducible.

JonHanna:
@mikedn Oh, you mean `Reduce` should be used by both?

mikedn:
@JonHanna Yes, the compiler already relies on Reduce for a few binary and unary ops - op=, ++ and --.

src/System.Linq.Expressions/src/System/Linq/Expressions/Compiler/LambdaCompiler.Statements.cs
```diff
@@ -152,6 +152,28 @@ private void EmitLoopExpression(Expression expr)
         private void EmitSwitchExpression(Expression expr, CompilationFlags flags)
         {
             SwitchExpression node = (SwitchExpression)expr;
+            
+            if (node.Cases.Count == 0)
+            {
```
VSadov: Just wondering - is this code reachable, considering the node is reducible?

JonHanna: You're pretty much hitting on the point @mikedn made in the opposite direction.

I added the reducibility later, with the hope that of consuming code that choked on case-less switches, maybe some of it attempts reducing, and as such doesn't actually get the expression that would cause it trouble.

I'd already changed up EmitSwitchExpression, and since EmitExpression doesn't attempt reducing on all nodes, it's still needed.

@mikedn  though points out that since reducing is now in there, changing stackspiller to remove its assumptions about what nodes can be reduced will mean this code becomes unreachable, and as such could be left unchanged.

And I think he's correct. I'm going to change this PR to match that approach.

VSadov: I looked through both compiler and interpreter and it appears that they consider only particular ExpressionTypes as reducible, so they will not attempt to reduce the switch.

Also in both cases, once the node is one of the reducible ExpressionTypes, it is treated as unconditionally reducible - for example interpreter throws if a node of reducible ExpressionType returns false from CanReduce and stack spiller in the compiler will force rewriting of all the parent nodes regardless of what CanReduce reports.

It would not be easy to treat SwitchExpression the same as other reducible expressions since it is "sometimes reducible".

VSadov: It appears that there is implicit assumption that all core ExpressionTypes are not reducible except for special 18 ExpressionTypes like ExpressionType.PostIncrementAssign and the like.

The part that all the core nodes are treated either as reducible or not based on ExpressionType makes me think we should not make switch reducible.

VSadov: We are not helping compiler or interpreter since they will not reduce switches, but because they both have trivial for loops going over all cases, they should be able to handle 0-cases situation just fine.
As for the other consumers - we do not know if we are helping them or making things worse by making switch expression reducible, but only sometimes.

Considering that being reducible is not helping known consumers and effect on unknown consumers could go either way, let's not make switch reducible.

JonHanna: I would have thought on first blush that this was a matter of changing LightCompiler.cs at line 2937 and StackSpiller.Generated.cs (generated by what?) at line 189 to branch on `.CanReduce`. And that seems to be the case with StackSpiller anyway, but I'm not a all clear on how to test LightCompiler.cs, which makes me obviously much less confident with this than once I was.

PR #3044, opened 2015-09-01
Allow block expressions to have zero sub-expressions.

### LINKED ISSUE #3043: Allow empty BlockExpression expressions
Issue #1877 discusses having `SwitchExpression` objects with no cases.

One might similarly have a case where `BlockExpression` elements were being created dynamically and one hit the case of their being no expressions within the block.

Such a block would be essentially the same as `Expression.Empty`, but arrived at through happening to hit zero item in a dynamic construction.

It would seem that the arguments [here](https://github.com/dotnet/corefx/pull/2803#issuecomment-136514021) about zero-case switches apply equally to zero-item blocks.

JonHanna:
Fixes #3043

Type of the block is always void. It remains an error to explicitly set it to
anything else.


stephentoub:
cc: @VSadov

PR #3046, opened 2015-09-01
Use BlockExpression subclasses more widely.

JonHanna:
BlockExpression has several subclasses that cover various numbers of sub-
expressions when the type is the same as the last expression, and there are
no variables. These are only used in one or two of the possible paths under
which they make sense. Use them more widely.


stephentoub:
cc: @VSadov

JonHanna:
Closing as covered by #3686

PR #3047, opened 2015-09-01
Add support for X509 revocation checks on Unix

### LINKED ISSUE #2203: Add support for X509 revocation checks on Unix
Due to a lack of a stable networking stack, the X509Chain for Unix does not support CRL downloading or OCSP.

There are a couple of different parts to be done:
* OCSP (Online checks)
* CRL fetching (Online checks)
* CRL caching (Online/Offline checks)
  * CRL cache refresh
* Using the CRL during chain building

bartonjs:
Add support for CRL checking in X509Chain on Unix.
- Look up the CRL address via the CRL Distribution Point extension.
  - Only "http" protocol addresses will be attempted (mostly this skips ldap addresses)
- Download the CRL using libcurl
- Save the result to ~/.dotnet/corefx/cryptography/crls/
  *\* The filename is the hash value from libcrypto, plus ".crl"
- Later chains will use this same CRL for the remainder of its validity period (crl.NextUpdate)
  *\* This matches the Windows CAPI behavior of caching it for the full lifetime.  It's definitely the easiest to write.

OCSP is not supported with this change.

Fixes #2203.


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
```diff
@@ -0,0 +1,253 @@
...
+
+                    if (nextUpdate < verificationTime)
+                    {
+                        return false;
+                    }
+
+                    if (!Interop.libcrypto.X509_STORE_add_crl(store, crl))
+                    {
+                        //error:0B07D065:x509 certificate routines:X509_STORE_add_crl:cert already in hash table
+                        //throw Interop.libcrypto.CreateOpenSslCryptographicException();
```
bartonjs: Oh, right, this.

For full chains, when the root provides its own CDP extension it will be the same as the one from the intermediate.  So there's one legit code that shouldn't throw.  Other things will likely just cause the chain to fail, so not throwing here isn't the end of the world.

Or we could plumb the OpenSSL error code into the HResult property of the exception, and check it here.  Exposing the code probably has positive utility, it just isn't really an "HResult", and I didn't want to be the first Unix thing to introduce that paradigm without a discussion.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
```diff
@@ -0,0 +1,253 @@
...
+
+                    // If crl.LastUpdate is in the past, downloading a new version isn't really going
+                    // to help, since we can't rewind the Internet. So this is just going to fail, but
+                    // at least it can fail without using the network.
+                    //
+                    // If crl.NextUpdate is in the past, try downloading a newer version.
+                    DateTime nextUpdate = OpenSslX509CertificateReader.ExtractValidityDateTime(
+                        Interop.Crypto.GetX509CrlNextUpdate(crl));
+
+                    if (nextUpdate < verificationTime)
```
stephentoub: Are there any timezone or daylight savings issues to worry about here?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
stephentoub: > Or we could plumb the OpenSSL error code into the HResult property of the exception, and check it here

I'm confused... what code is creating the exception?  You say "check it here", but it looks like the commented out code here is the one creating the exception.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
```diff
@@ -0,0 +1,253 @@
...
+                // null is a valid return (e.g. no remainingDownloadTime)
+                if (crl != null && !crl.IsInvalid)
+                {
+                    if (!Interop.libcrypto.X509_STORE_add_crl(store, crl))
+                    {
+                        //error:0B07D065:x509 certificate routines:X509_STORE_add_crl:cert already in hash table
+                        //throw Interop.libcrypto.CreateOpenSslCryptographicException();
+                    }
+
+                    // It doesn't matter if saving it fails
```
stephentoub: Why?  Maybe clarify in the comment.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
```diff
@@ -0,0 +1,253 @@
...
+                    }
+                }
+            }
+        }
+        
+        private static string GetCachedCrlPath(X509Certificate2 cert, bool mkDir=false)
+        {
+            OpenSslX509CertificateReader pal = (OpenSslX509CertificateReader)cert.Pal;
+
+            string crlDir = PersistedFiles.GetUserFeatureDirectory("cryptography", "crls");
```
stephentoub: Is there somewhere we could centralize the "cryptography" constant so that we don't accidentally introduce spelling mistakes in the various call sites?

stephentoub:
A few comments/questions, but LGTM.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: There are three options that I see:

* Remove the if, ignore the function result (effectively what I currently have)
* Try to determine the current OpenSSL error code without creating an exception
    * The way that OpenSSL's error queue works that feels slightly problematic
* Attach the error code to the exception object, and here do something like `CryptographicException e = CreateException(); if (e.HResult != X509_R_CERT_ALREADY_IN_HASH_TABLE) { throw e; }`

The first has a certain distaste to it, the second might be non-destructively-achievable with some native interop, and the third puts values into the HResult field that weren't HResult values.

And I'm sort of torn on which one I like the most...

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: Will do

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: I'll find/make somewhere, sure.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: I was about to say "no, we take care of that", but that's when sending the date into the chain for full verification.

if verificationTime is DateTimeKind.Utc there might be an offset conversion that doesn't happen... I'll play around with this and determine the answer; and fix/document it when so determined.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: Oh, bah.  There's no CryptographicException ctor that takes both a code and a message; so option 3 might not exist.

bartonjs: Sometimes it doesn't occur to me that I have the power to change things CryptographicException :smile:.

Currently we don't FormatMessage the HR-based CryptographicExceptions on Windows\CoreFX; but if we built Primitives in a platform-specific manner then we could use FormatMessage on Windows and the OpenSSL formatted message on Unix; then we'd just make `CreateOpenSslCryptographicException` always use the "HResult" constructor.

It's a bit heavy-handed for this change; so it might be a "make an issue for the messaging change, and a followup one for making this method utilize it" approach.

That really just comes down to "do we want to call OpenSSL's error codes HResults because they happen to both be 32-bit error values?".  I know we have errno vs GetLastWin32Error; but how have we been addressing this in other areas?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
bartonjs: Good catch.  We normalized the DateTime when sending it down to OpenSSL (it wants a local value, which it's going to turn into a UTC value...), and we normalized numbers coming back from OpenSSL to Local to match the Windows half; but this managed to float outside the normalization zone.

So I moved the normalization to the ChainPal entrypoint and made here and SetX509ChainVerifyTime assert that they got a non-UTC DateTime.

While this would have been logistically fine in the western hemisphere (we'd start downloading at your offset sooner than the expiration), positive offset timezones would have encountered "CRL is expired" for a few hours while we steadfastly asserted that our on disk copy was good enough.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CrlCache.cs
stephentoub: I think it's reasonable to carry the error codes in the otherwise defunct HResult field.  We do the same thing in a bunch of places with IOException.

jimmyca15:
@bartonjs is the CRL cache location documented/stable? I'm working on sharing this between containers and to do so I need to mount share the CRL cache directory from the container host to all the containers.

bartonjs:
> is the CRL cache location documented

No, it's considered an internal implementation detail.

>  is the CRL cache location stable

It hasn't changed yet :smile:

PR #3048, opened 2015-09-01
Merge branch 'master' of https://github.com/dotnet/corefx into dev/metadata

tmat:


tmat:
@nguerrera

nguerrera:
I think we need to (temporarily) leave AssemblyFileVersion as v1.1.0.0 in the csproj because the dev/metadata builds are using a different mechanism than official builds out of master and it's resulting in v1.0.0.0 file version without that line, which breaks when putting the dev build in an msi that upgrades v1.0.x. Actually, we may need to bump it passed 4.0.x because those are the file versions that were used for all the last wave of stable nuget packages. cc @jasonmalinowski @ericstj

nguerrera:
LGTM otherwise.

stephentoub:
@dotnet-bot test this please (failed with connection errors in the native OS X builds)

PR #3052, opened 2015-09-02
XLinq: Snapshot to an array instead of List<T>

justinvp:
Now that `EnumerableHelpers` is used in this class, make use of it in the `Remove` methods to snapshot to an array instead of a `List<T>`. This avoids the extra `List<T>` instance allocation with slightly faster enumeration as we can loop over the array directly (vs. indirectly through `List<T>`'s enumerator).


stephentoub:
Nice. LGTM.

PR #3053, opened 2015-09-02
Separate ToArray specialisation from Enumerable.Iterator

JonHanna:
Enumerable.Iterator has a virtual ToArray that is overridden to provide
optimised array-creation & buffer-filling when possible.

Most Enumerable.Iterator subclasses don't override the default, which
results in it passing through Buffer's constructor a second time. This can
also happen if those that do override it have a predicate.

Separate out this part of Iterator's interface into a separate interface, for
only those classes that override it, and signal a failure to create an array
by returning null.

Implement this IArrayProvider in OrderedEnumerable, Lookup, GroupedEnumerable,
RangeIterator & RepeatIterator.

Bypass Buffer entirely on ToArray when this allows.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -106,14 +106,9 @@ namespace System.Linq
 
             public IEnumerator<TSource> GetEnumerator()
             {
-                if (state == 0 && _threadId == Environment.CurrentManagedThreadId)
-                {
-                    state = 1;
-                    return this;
-                }
-                Iterator<TSource> duplicate = Clone();
-                duplicate.state = 1;
-                return duplicate;
+                Iterator<TSource> enumerator = state == 0 && _threadId == Environment.CurrentManagedThreadId ? this : Clone();
+                enumerator.state = 1;
+                return enumerator;
             }
 
             public abstract bool MoveNext();
@@ -413,7 +408,7 @@ namespace System.Linq
 
             public TResult[] ToArray()
             {
-                if (_predicate != null) return null;
+                if (_predicate != null && _source.Length != 0) return null;
 
                 var results = new TResult[_source.Length];
                 for (int i = 0; i < results.Length; i++)
@@ -479,7 +474,7 @@ namespace System.Linq
 
             public TResult[] ToArray()
             {
-                if (_predicate != null) return null;
+                if (_predicate != null && _source.Count != 0) return null;
 
                 var results = new TResult[_source.Count];
                 for (int i = 0; i < results.Length; i++)
@@ -1004,6 +999,12 @@ namespace System.Linq
         public static TSource[] ToArray<TSource>(this IEnumerable<TSource> source)
         {
             if (source == null) throw Error.ArgumentNull("source");
+            IArrayProvider<TSource> arrayProvider = source as IArrayProvider<TSource>;
+            if (arrayProvider != null)
+            {
+                TSource[] array = arrayProvider.ToArray();
+                if (array != null) return array;
+            }
             return new Buffer<TSource>(source).ToArray();
         }
 
@@ -1428,23 +1429,134 @@ namespace System.Linq
         {
             long max = ((long)start) + count - 1;
             if (count < 0 || max > Int32.MaxValue) throw Error.ArgumentOutOfRange("count");
-            return RangeIterator(start, count);
+            return new RangeIterator(start, count);
         }
 
-        private static IEnumerable<int> RangeIterator(int start, int count)
+        private sealed class RangeIterator : Iterator<int>, IArrayProvider<int>
         {
-            for (int end = start + count; start != end; start++) yield return start;
+            private readonly int _start;
+            private readonly int _end;
+            
+            public RangeIterator(int start, int count)
+            {
+                _start = start;
+                _end = start + count;
+            }
+            
+            public override Iterator<int> Clone()
+            {
+                return new RangeIterator(_start, _end - _start);
+            }
+            
+            public override bool MoveNext()
+            {
+                switch(state)
+                {
+                    case 1:
+                        if (_start == _end) break;
+                        current = _start;
+                        state = 2;
+                        return true;
+                    case 2:
+                        if (++current == _end)
+                        {
+                            --current; // Not crucial but maintains compatibility with previous versions. 
+                            break;
+                        }
+                        return true;
+                }
+                state = -1;
+                return false;
+            }
+            
+            public override void Dispose()
+            {
+                state = -1; // Don't reset current
+            }
+
+            public override IEnumerable<TResult> Select<TResult>(Func<int, TResult> selector)
+            {
+                return new WhereSelectEnumerableIterator<int, TResult>(this, null, selector);
+            }
+
+            public override IEnumerable<int> Where(Func<int, bool> predicate)
+            {
+                return new WhereEnumerableIterator<int>(this, predicate);
+            }
+
+            public int[] ToArray()
+            {
+                int[] array = new int[_end - _start];
+                int cur = _start;
+                for (int i = 0; i != array.Length; ++i)
+                {
+                    array[i] = cur;
+                    ++cur;
+                }
+
+                return array;
+            }
         }
 
         public static IEnumerable<TResult> Repeat<TResult>(TResult element, int count)
         {
             if (count < 0) throw Error.ArgumentOutOfRange("count");
-            return RepeatIterator<TResult>(element, count);
+            return new RepeatIterator<TResult>(element, count);
         }
 
-        private static IEnumerable<TResult> RepeatIterator<TResult>(TResult element, int count)
+        private sealed class RepeatIterator<TResult> : Iterator<TResult>, IArrayProvider<TResult>
         {
-            for (int i = 0; i < count; i++) yield return element;
+            private readonly int _count;
+            private int _sent;
+
+            public RepeatIterator(TResult element, int count)
+            {
+                current = element;
+                _count = count;
+            }
+            
+            public override Iterator<TResult> Clone()
+            {
+                return new RepeatIterator<TResult>(current, _count);
+            }
+            
+            public override void Dispose()
+            {
+                // Don't let base Dispose wipe current.
+                state = -1;
+            }
+            
+            public override bool MoveNext()
+            {
+                if (state == 1 & _sent != _count)
+                {
+                    ++_sent;
+                    return true;
+                }
+                state = -1;
+                return false;
+            }
+
+            public override IEnumerable<TSelected> Select<TSelected>(Func<TResult, TSelected> selector)
+            {
+                return new WhereSelectEnumerableIterator<TResult, TSelected>(this, null, selector);
+            }
+
+            public override IEnumerable<TResult> Where(Func<TResult, bool> predicate)
+            {
+                return new WhereEnumerableIterator<TResult>(this, predicate);
+            }
+            
+            public TResult[] ToArray()
+            {
+                TResult[] array = new TResult[_count];
+                if (current != null)
+                {
+                    for (int i = 0; i != array.Length; ++i) array[i] = current;
+                }
+
+                return array;
+            }
         }
 
         public static IEnumerable<TResult> Empty<TResult>()
@@ -2759,8 +2871,15 @@ namespace System.Linq
         }
     }
 
+    /// <summary>
+    /// An iterator that can (or sometimes can) produce an array through an optimized path.
+    /// </summary>
     internal interface IArrayProvider<TElement>
     {
+        /// <summary>
+        /// Produce an array of the sequence through an optimized path, if possible.
+        /// </summary>
+        /// <returns>The array, or null if an optimized path isn't possible, and default behaviour should be used.</returns>
         TElement[] ToArray();
     }
 
@@ -2789,7 +2908,7 @@ namespace System.Linq
         bool Contains(TKey key);
     }
 
-    public class Lookup<TKey, TElement> : IEnumerable<IGrouping<TKey, TElement>>, ILookup<TKey, TElement>
+    public class Lookup<TKey, TElement> : IEnumerable<IGrouping<TKey, TElement>>, ILookup<TKey, TElement>, IArrayProvider<IGrouping<TKey, TElement>>
     {
         private IEqualityComparer<TKey> _comparer;
         private Grouping<TKey, TElement>[] _groupings;
@@ -2858,6 +2977,23 @@ namespace System.Linq
                     yield return g;
                 } while (g != _lastGrouping);
             }
+        }
+        
+        public IGrouping<TKey, TElement>[] ToArray()
+        {
+            IGrouping<TKey, TElement>[] array = new IGrouping<TKey, TElement>[_count];
+            int index = 0;
+            Grouping<TKey, TElement> g = _lastGrouping;
+            if (g != null)
+            {
+                do
+                {
+                    g = g.next;
+                    array[index] = g;
+                    ++index;
+                } while (g != _lastGrouping);
+            }
+            return array;
         }
 
         public IEnumerable<TResult> ApplyResultSelector<TResult>(Func<TKey, IEnumerable<TElement>, TResult> resultSelector)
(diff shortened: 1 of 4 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1439,23 +1429,125 @@ public static IEnumerable<int> Range(int start, int count)
...
+            
+            public override void Dispose()
+            {
+                // Don't let base Dispose wipe current.
+                state = -1;
+            }
+            
+            public override Boolean MoveNext()
+            {
+                if(state == 1 & _sent != _count)
```
justinvp: nit: space after "if"

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2865,6 +2962,18 @@ public bool Contains(TKey key)
                 } while (g != _lastGrouping);
             }
         }
+        
+        public IGrouping<TKey, TElement>[] ToArray()
+        {
+            IGrouping<TKey, TElement>[] array = new IGrouping<TKey, TElement>[_count];
+            int index = 0;
+            foreach(IGrouping<TKey, TElement> group in this)
```
justinvp: nit: should be a space after "foreach"

justinvp: I wonder if it'd be better to manually loop through the groupings here, instead of `foreach`, to avoid the enumerator allocation.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2770,6 +2862,11 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
         }
     }
 
+    internal interface IArrayProvider<TElement>
+    {
+        TElement[] ToArray();
```
justinvp: Doc comment here with an explanation about returning null?

JonHanna: Yeah, that should definitely be added.

src/System.Linq/tests/LegacyTests/GroupByTests.cs
```diff
@@ -1151,5 +1151,23 @@ public void Test()
...
+                new Record{ Name = "Chris", Score = 49 },
+                new Record{ Name = "Robert", Score = -100 },
+                new Record{ Name = "Chris", Score = 24 },
+                new Record{ Name = "Prakash", Score = 9 },
+                new Record{ Name = "Tim", Score = 25 }
+            };
+            
+            IGrouping<string, Record>[] groupedArray = source.GroupBy(r => r.Name).ToArray();
+            Assert.Equal(4, groupedArray.Length);
+            Assert.Equal(source.GroupBy(r => r.Name), groupedArray);
```
Maxwe11: What we test here? `source.GroupBy` against `source.GroupBy`

Maxwe11: o sorry, misread ToArray

JonHanna: NP. Though actually a lot of the legacy linq tests do indeed test things against themselves, which can seem rather strange.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1439,23 +1429,125 @@ public static IEnumerable<int> Range(int start, int count)
...
+            public override Iterator<TResult> Clone()
+            {
+                return new RepeatIterator<TResult>(current, _count);
+            }
+            
+            public override void Dispose()
+            {
+                // Don't let base Dispose wipe current.
+                state = -1;
+            }
```
stephentoub: Why isn't this also necessary on RangeIterator?

JonHanna: Repeat has a constant value, so we just set current to that value and leave it. Doesn't apply elsewhere.

stephentoub: I realize RepeatIterator is a bit more special in that you're using current to store the supplied value.  But the iterator generated by the compiler today for Enumerable.Range doesn't wipe current as part of Dispose, such that calling Current after Dispose will return the last value rather than 0.  I'm not sure why the base Iterator is clearing current at all (@VSadov?), but seems like we should probably retain the same behavior Range has today.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1439,23 +1429,125 @@ public static IEnumerable<int> Range(int start, int count)
...
-            return RepeatIterator<TResult>(element, count);
+            return new RepeatIterator<TResult>(element, count);
         }
 
-        private static IEnumerable<TResult> RepeatIterator<TResult>(TResult element, int count)
+        private sealed class RepeatIterator<TResult> : Iterator<TResult>, IArrayProvider<TResult>
         {
-            for (int i = 0; i < count; i++) yield return element;
+            private readonly int _count;
+            private int _sent;
```
stephentoub: Rather than storing both _send and _count, could you just store a single value that you then decrement until it gets to zero?

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: I'm cool with that change, which if nothing else is a Microsoft in itself.
Clearing in the base Dispose does seem pointless anyway, though it could perhaps have a collection impact on some cases.

stephentoub: For now I suggest you override Dispose in RangeIterator to maintain the same behavior for it that exists today (and if you could add a test, that'd be great).  We can subsequently look at whether the base iterator type should have the current clearing removed, at which point we can remove these Dispose overrides at the same time.

> which if nothing else is a Microsoft in itself?

I didn't follow that...?

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: We have to store _count anyway for replays. Whether we then also store the number sent or to send is of little odds. (We could force state into dual service, but that seems a bridge too far).

stephentoub: Ah, of course, I always forget about Reset :)

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Ha. My phone's autocorrect got a bit pointed there. I'd typed "micro-opt in itself".

Agreed on the Dispose override.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2770,6 +2862,18 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
...
 
+    /// <summary>
+    /// An iterator that can (or sometimes can) produce an array through an optimised path.
+    /// </summary>
+    internal interface IArrayProvider<TElement>
+    {
+        /// <summary>
+        /// Produce an array of the sequence through an optimised path, if possible.
+        /// </summary>
+        /// <returns>The array, or null if an optimised path isn't possible, and default behaviour should be used.</returns>
```
stephentoub: I realize this is very US of me, but would you mind changing "optimised" to "optimized"?

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Reset just throws, but a subsequent call to GetEnumerator would be in trouble.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Sure, if I can tolerate Websterisms like "color" I can certainly accept where the US has the more traditional spelling :)
More seriously, I agree comments should be consistent as to dialect, that was just a slip.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: Yup.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3243,6 +3369,20 @@ public IEnumerator<TElement> GetEnumerator()
...
         }
+        
+        public TElement[] ToArray()
+        {
+            Buffer<TElement> buffer = new Buffer<TElement>(source);
+            int count = buffer.count;
+            TElement[] array = new TElement[count];
+            if (count > 0)
+            {
+                int[] map = GetEnumerableSorter(null).Sort(buffer.items, count);
```
stephentoub: It'd be nice if this paralleled the same construct as in GetEnumerator above.  Since this version is more concise, could you change the three corresponding lines in GetEnumerator to match this, e.g.
```C#
int[] map = GetEnumerableSorter(null).Sort(buffer.items, buffer.count);
```

JonHanna: I'll have them both go through a `private int[] SortedMap(Buffer<TElement> buffer)` method, that should be small enough to be inlined in most cases.

Otherwise I'd worry that maybe some compiler wouldn't have the sorter only as a local within `MoveNext()` of the produced iterator? While the current coding comes close to guaranteeing it becomes a field, it firmly guarantees such a field gets cleared again in the same `MoveNext()` call that sets it.
If some compilers create a field for the unnamed sorter, then that could have a GC impact, since the sorter can be quite large (one array full of number-of-source-elements keys for each level of sorting) that wouldn't be great.
MSBuild seems fine with it, but I'd want to be sure.
Going through a separate method removes that risk.

stephentoub:
Thanks, @JonHanna.  The updates LGTM.   Could you squash things down a bit?

cc: @VSadov

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: It turns out, one has to back `current` up one again to completely duplicate the previous behaviour, which is mildly irritating, but hardly a burden.

VSadov: RE: clearing of Current in Dispose.

I think the behavior of Current after Disposing is undefined. Some implementations throw ObjectDisposed, some return last value, some return default value ...

Maintaining same behavior of "Current after Dispose" does not seem very important, just because behavior is undefined in spec in inconsistent among implementations.

The reason for clearing Current in dispose is typically an opportunistic optimization to not keep it rooted longer than necessary. So IMO clearing is a mildly good thing (unless we know T is a primitive value), but not extremely important. 

It is a bit more important for iterators that are also enumerables. - While it is not common for disposed iterators to be retained, the enumerable itself can be kept alive for quite a long time and if that keeps an instance of Current alive after the first iterator (same instance as enumerable) is disposed, it may become a cause of unexpected "leaks". 

It may make sense to review all the Dispose implementations here and make sure we clear Current where convenient, but again, I do not see this as highly important.

JonHanna: In that regard, whether the override is here or not is minor (might as well leave it) but the line at https://github.com/dotnet/corefx/pull/3053/files#diff-95a14d3b4ae14339b623dd95c7537fd0R1463 is just pointless busywork that should be removed?
I'd be happier to remove that and the test for it if nobody objected.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2770,6 +2871,18 @@ public static decimal Average<TSource>(this IEnumerable<TSource> source, Func<TS
...
 
+    /// <summary>
+    /// An iterator that can (or sometimes can) produce an array through an optimized path.
+    /// </summary>
+    internal interface IArrayProvider<TElement>
+    {
+        /// <summary>
+        /// Produce an array of the sequence through an optimized path, if possible.
+        /// </summary>
+        /// <returns>The array, or null if an optimized path isn't possible, and default behaviour should be used.</returns>
```
justinvp: nit: also "behaviour" -> "behavior"

src/System.Linq/tests/RangeTests.cs
```diff
@@ -133,5 +133,18 @@ public void ArbitraryStart()
 
             Assert.Equal(expected, Enumerable.Range(start, count));
         }
+        
+        [Fact]
+        public void RangeAsEnumeratorStartsZeroEndsAtLast()
+        {
+            // The behaviour here is not specified, but any change that breaks this is
```
justinvp: nit: also "behaviour" -> "behavior"

VSadov:
LGTM

JonHanna:
What to getting rid of line 1463 as discussed above?

VSadov:
I think the line 1463  is to enforce another behavior that is in the spec - it is ok to call MoveNext repeatedly after reaching the end of the iteration and you should keep getting "false".

JonHanna:
No, it's just to maintain consistency of Current after MoveNext has returned false with the previous version, as I added it for that purpose due to earlier discussion here. If keeping that consistent between builds isn't important it should go, but I can just do another PR when I'm back at a computer.

VSadov:
Ah, I see. We will not get there again because of the state change.
Ok, to not guess whether it is a break or not, can we do something like

```C
int c = current;
if (c == _end)  break;
current = c + 1;
return true;
```

it is likely to translate in mostly the same code as 
```C
if (++current == _end)  break;
return true;
```
but there will not be any differences in behavior.

VSadov:
That is not very important though. The version with current-- is probably good enough.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -2865,6 +2978,23 @@ public bool Contains(TKey key)
                 } while (g != _lastGrouping);
             }
         }
+        
+        public IGrouping<TKey, TElement>[] ToArray()
```
justinvp: I missed this in the initial review... `Lookup<TKey, TElement` is a public class, so this is adding a new public `ToArray` method. Instead, this new method should be implemented explicitly (private).

justinvp: (If we want to make it public, which seems reasonable, it would need to go through the new API process).

JonHanna: Well, that was very foolish of me. I'll mop up my mess right now.

justinvp: While you're at it, I know this is nitpicky, but would you mind changing "behaviour" -> "behavior" (similar to the "optimize" change)?

Also, one other nit I noticed that was hard to see in the diff: this PR introduces a bunch of trailing whitespace that could be removed (tab spaces on "blank lines").

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -1439,23 +1429,134 @@ public static IEnumerable<int> Range(int start, int count)
...
-            for (int end = start + count; start != end; start++) yield return start;
+            private readonly int _start;
+            private readonly int _end;
+            
+            public RangeIterator(int start, int count)
+            {
+                _start = start;
+                _end = start + count;
+            }
+
```
justinvp: nit: trailing whitespace (there's some of these throughout)

JonHanna: #3064 I removed all the trailing whitespace throughout because why not, and that nit with the MoveNext too.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: Good catch, @justinvp.

JonHanna: The embarrassing bit is I had this as it should be at the start of my coding it and for some mysterious reason changed it to be public.

> (If we want to make it public, which seems reasonable, it would need to go through the new API process).

It would, but it wouldn't be reasonable, IMO: Any time this class is available, `Enumerable.ToArray()` is available, and now just calls into this anyway, so the upside would be tiny. (A discoverable `Count` and `CopyTo` I would argue for, though I understand the counter-arguments, but not fully public there either).

VSadov: Good catch indeed. I noticed public in couple places, but those were in internal classes, so I figured it was ok...

PR #3057, opened 2015-09-02
Add back WindowsIdentity::.ctor(string)

AtsushiKan:
cc @bartonjs 

Adding back ctor(string) for WindowsIdentity.

Tested on local machine but the api (like its desktop counterpart) requires a domain-joined machine to work at all, and possibly spams the audit logs too, so the public unit tests are not a suitable place to validate this.

Desktop version also contains a bunch of scary code to add the TCB privilege to the token in case an  operating system that happens to be passing through wants to use this api to exercise its awesome LocalSystem powers. I don't think we need that functionality so I only ported the fallback case where we return a identification token only.

### DIFF
--- a/src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
+++ b/src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
@@ -1,15 +1,23 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Win32;
 using Microsoft.Win32.SafeHandles;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
-using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Security.Claims;
+using System.Text;
 using System.Threading;
+
+using KERB_LOGON_SUBMIT_TYPE = Interop.SspiCli.KERB_LOGON_SUBMIT_TYPE;
+using KERB_S4U_LOGON = Interop.SspiCli.KERB_S4U_LOGON;
+using KerbS4uLogonFlags = Interop.SspiCli.KerbS4uLogonFlags;
+using LUID = Interop.LUID;
+using LSA_STRING = Interop.SspiCli.LSA_STRING;
+using QUOTA_LIMITS = Interop.SspiCli.QUOTA_LIMITS;
+using SECURITY_LOGON_TYPE = Interop.SspiCli.SECURITY_LOGON_TYPE;
+using TOKEN_SOURCE = Interop.SspiCli.TOKEN_SOURCE;
 
 namespace System.Security.Principal
 {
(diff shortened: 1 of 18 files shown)


bartonjs:
@pallavit I don't see the Lsa functions on the API interop list; is this going to cause us any problems?

src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
```diff
@@ -41,6 +49,145 @@ private WindowsIdentity()
...
         { }
 
+        /// <summary>
+        /// Initializes a new instance of the WindowsIdentity class for the user represented by the specified User Principal Name (UPN).
+        /// </summary>
+        /// <remarks>
+        /// Unlike the desktop version, we connect to Lsa only as an untrusted caller. We do not attempt to explot Tcb privilege or adjust the current
+        /// thread privilege to include Tcb.
+        /// </remarks>
+        public WindowsIdentity(string sUserPrincipalName)
```
bartonjs: You'll need to add this to the copy in `ref` as well (the ref directory is used to build the contract assembly now)

bartonjs:
Aside from the missing ref edit it looks good to me; but let Pallavi chime in on the interop permissibility before merging.

src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
```diff
@@ -41,6 +49,145 @@ private WindowsIdentity()
...
+
+                // 8 byte or less name that indicates the source of the access token. This choice of name is visible to callers through the native GetTokenInformation() api
+                // so we'll use the same name the CLR used even though we're not actually the "CLR."
+                byte[] sourceName = { (byte)'C', (byte)'L', (byte)'R', (byte)0 };
+
+                TOKEN_SOURCE sourceContext;
+                if (!Interop.SecurityBase.AllocateLocallyUniqueId(out sourceContext.SourceIdentifier))
+                    throw new SecurityException(Interop.mincore.GetMessage(Marshal.GetLastWin32Error()));
+                sourceContext.SourceName = new byte[TOKEN_SOURCE.TOKEN_SOURCE_LENGTH];
+                Array.Copy(sourceName, sourceContext.SourceName, sourceName.Length);
```
stephentoub: Nit: it'd be slightly cheaper to do:
```C#
Array.Copy(sourceName, 0, sourceContext.SourceName, 0, sourceName.Length);
```
so that Array.Copy doesn't have to look up the lower bound.

src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
```diff
@@ -41,6 +49,145 @@ private WindowsIdentity()
...
+
+                // Desktop compat: Desktop never null-checks sUserPrincipalName. Actual behavior is that the null makes it down to Encoding.Unicode.GetBytes() which then throws
+                // the ArgumentNullException (provided that the prior LSA calls didn't fail first.) To make this compat decision explicit, we'll null check ourselves 
+                // and simulate the exception from Encoding.Unicode.GetBytes().
+                if (sUserPrincipalName == null)
+                    throw new ArgumentNullException("s");
```
stephentoub: I realize the "s" is there for desktop compat, but this seems like something we could break compat for in order to do the "right thing", at least with regards to the argument name.

AtsushiKan:
They’re there in OneCoreApis.xml. I ran the PInvokeChecker to be sure before opening the PR. Not in OneCoreUWPApis.xml, but this assembly isn’t abled for .NetNative and it was massively non-compliant there already.

stephentoub:
A few comments, but otherwise LGTM

src/System.Security.Principal.Windows/src/System/Security/Principal/WindowsIdentity.cs
pallavit: >  I don't see the Lsa functions on the API interop list; is this going to cause us any problems?

which interop list are you talking about?

pallavit:
> They’re there in OneCoreApis.xml. I ran the PInvokeChecker to be sure before opening the PR. Not in OneCoreUWPApis.xml, but this assembly isn’t abled for .NetNative and it was massively non-compliant there already.

Thanks for running PinvokeChecker on the library :). The LSA* APIs are not supported on store, but since this does not target .NetNative , we should be good.

bartonjs:
@AtsushiKan Okay.  I didn't see them on the list that I thought was the input into OneCoreApis.xml, but if you ran the tool manually that was my only concern.  Kudos for the due diligence :)

LGTM.

PR #3058, opened 2015-09-02
Verify the environment of the child process itself

### LINKED ISSUE #804: Add test to verify expected set of environment variables
It looks like we have a test to verify that ProcessStartInfo gets the right environment variables, but we don't have a test to verify the environment of the child process itself.

Maxwe11:
Closes #804

### DIFF
--- a/src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
+++ b/src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
@@ -1,15 +1,15 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections;
 using System.Collections.Generic;
-using System.Security;
-using System.Security.Principal;
-using Microsoft.Win32;
 using System.IO;
-using Xunit;
-using Microsoft.Win32.SafeHandles;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Security.Principal;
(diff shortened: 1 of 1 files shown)


src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
```diff
@@ -182,6 +182,63 @@ public void TestEnvironmentProperty()
...
+            // Profilers / code coverage tools can add own environment variables but we start
+            // child process without them. Thus the set of variables from child process will
+            // compose subset of variables from current process.
+            // But in case if tests running directly through the Xunit runner, sets will be equal
+            // and Assert.ProperSubset will throw. We add null to avoid this.
+            currentProcEnv.Add(null);
+            Assert.ProperSubset(currentProcEnv, actualEnv);
+        }
+
+        private static int PrintEnvironment()
```
stephentoub: This should be doable as a lambda so that the test is all self-contained in the test method.  If it isn't for some reason, this is fine.

stephentoub:
LGTM

src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
Maxwe11: will fix

Maxwe11: fixed

stephentoub:
The update looks good. Thanks for the additional test.

PR #3059, opened 2015-09-02
Converting simple almost-one-line functions in I/O and Process over to the PAL layer

sokket:
Converting simple almost-one-line functions in I/O and Process over from platform-specific P/Invoke to the PAL layer

@nguerrera


src/Common/src/Interop/Unix/System.Native/Interop.Access.cs
```diff
@@ -0,0 +1,22 @@
...
+            F_OK = 0,   /* Check for existence */
+            X_OK = 1,   /* Check for execute */
+            W_OK = 2,   /* Check for write */
+            R_OK = 4,   /* Check for read */
+        }
+
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal static extern int Access(
+            [MarshalAs(UnmanagedType.LPStr)]
+            string path,
```
stephentoub: I'd prefer in cases like this for the attribute to be on the same line as the parameter. (That said, in this case the attribute shouldn't even be necessary, as it should be the default.  Another alternative if you did want to be explicit would be to use CharSet in the DllImport.)

src/Common/src/Interop/Unix/System.Native/Interop.FnMatch.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal const int FNM_NOMATCH = 1; // constant from fnmatch.h
```
stephentoub: Where is this used?

src/Common/src/Interop/Unix/System.Native/Interop.GetPid.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
```
stephentoub: This doesn't need SetLastError=true.

src/Common/src/Interop/Unix/System.Native/Interop.Kill.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal enum Signals : int
+        {
+            None = 0,
```
stephentoub: Do we need None?

src/Native/System.Native/pal_io.h
```diff
@@ -309,3 +349,61 @@ extern "C"
...
+/**
+ * Checks the access permissions of the current calling user on the specified path for the specified mode.
+ * Returns -1 if the path cannot be found or the if desired access is not granted and errno is set; otherwise, returns 0.
+ */
+extern "C"
+int32_t Access(const char* path, AccessMode mode);
+
+/**
+ * Tests whether a pathname matches a specified pattern.
+ * Returns 0 if the string matches; returns FNM_NOMATCH; otherwise, returns a non-zero error code.
```
stephentoub: The "returns FNM_NOMATCH" doesn't appear to have a matching condition.

src/Common/src/Interop/Unix/System.Native/Interop.Kill.cs
stephentoub: Nevermind, I wrote the code that does :smile:

src/Native/System.Native/pal_io.h
```diff
@@ -113,6 +113,46 @@ enum class NodeType : int16_t
 };
 
 /**
+ * Constants from sys/file.h for lock types
+ */
+enum class LockOperations : int32_t
```
stephentoub: We should have static asserts that verify these constants all match the platform values, right?

src/Native/System.Native/pal_io.h
```diff
@@ -113,6 +113,46 @@ enum class NodeType : int16_t
...
+ * Constants passed to lseek telling the OS where to seek from
+ */
+enum class SeekWhence : int32_t
+{
+    PAL_SEEK_SET = 0,   /* seek from the beginning of the stream */
+    PAL_SEEK_CUR = 1,   /* seek from the current position */
+    PAL_SEEK_END = 2,   /* seek from the end of the stream, wrapping if necessary */
+};
+
+/**
```
stephentoub: Do we need a corresponding enum for PAL signals?

src/Common/src/Interop/Unix/System.Native/Interop.Access.cs
```diff
@@ -0,0 +1,22 @@
+using System;
```
stephentoub: A bunch of these files are missing the license header.

stephentoub:
Some comments, but otherwise LGTM.

src/Common/src/Interop/Unix/System.Native/Interop.Access.cs
sokket: Ah right, removed the attribute entirely :)

src/Common/src/Interop/Unix/System.Native/Interop.FnMatch.cs
sokket: It is not currently used, but since the documentation calls it out as a very specific return value, I thought it would be useful to declare in case any future code change needed it. I can remove it if you would like

src/Common/src/Interop/Unix/System.Native/Interop.GetPid.cs
sokket: Fixed

src/Common/src/Interop/Unix/System.Native/Interop.Kill.cs
sokket: :)

src/Native/System.Native/pal_io.h
sokket: Good catch, added the return condition for FNM_NOMATCH

src/Native/System.Native/pal_io.h
sokket: Ah right right, will add the static asserts for the new enums

src/Native/System.Native/pal_io.h
sokket: Good idea

src/Common/src/Interop/Unix/System.Native/Interop.Access.cs
sokket: Yikes, good catch...copy+paste problem

nguerrera:
LGTM. Thanks.

sokket:
Thanks guys, merging!

PR #3060, opened 2015-09-02
Address various issues in FileSystemWatcher

stephentoub:
- On Linux, FileSystemWatcher is built on top of inotify, which has various limits configured in the system for how many instances and watches can be created.  When these limits are hit, exceptions are generated, but the Linux-provided error messages are cryptic about what the problem is.  This change adds better error messages for those cases.  It also ensures that errors regarding watch limits always come out of OnError rather than sometimes being thrown synchronously from EnableRaisingEvents and other times coming out asynchronously.
- As part of fixing that, I noticed that we had some reliability issues around marking the instance as enabled.  I fixed this up to ensure that we only mark the instance as enabled when we've created everything we need to successfully launch it.
- While doing that, I noticed that on OS X, if we restart an FSW, we were unnecessarily re-creating a delegate.  I changed it to avoid the unnecessary allocation.
- And while doing that, I also noticed that we we had a bad exception message when constructing an FSW with something other than a directory.  I fixed the error message.
- I added tests for all these things.

cc: @sokket, @dougbu

### DIFF
--- a/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
+++ b/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
@@ -31,8 +31,24 @@ namespace System.IO
             SafeFileHandle handle;
             try { } finally
             {
-                int fd;
-                Interop.CheckIo(fd = Interop.libc.inotify_init());
+                int fd = Interop.libc.inotify_init();
+                if (fd == -1)
+                {
+                    Interop.ErrorInfo error = Interop.Sys.GetLastErrorInfo();
+                    switch (error.Error)
+                    {
+                        case Interop.Error.EMFILE:
+                            string maxValue = ReadMaxUserLimit(MaxUserInstancesPath);
+                            string message = !string.IsNullOrEmpty(maxValue) ?
+                                SR.Format(SR.IOException_INotifyInstanceUserLimitExceeded_Value, maxValue) :
+                                SR.IOException_INotifyInstanceUserLimitExceeded;
+                            throw new IOException(message, error.RawErrno);
+                        case Interop.Error.ENFILE:
+                            throw new IOException(SR.IOException_INotifyInstanceSystemLimitExceeded, error.RawErrno);
+                        default:
+                            throw Interop.GetExceptionForIoErrno(error);
+                    }
+                }
                 handle = new SafeFileHandle((IntPtr)fd, ownsHandle: true);
             }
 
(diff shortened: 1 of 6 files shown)


src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
```diff
@@ -92,12 +108,27 @@ private void FinalizeDispose()
...
         /// </summary>
         private CancellationTokenSource _cancellation;
 
+        /// <summary>Reads the value of a max user limit path from procfs.</summary>
+        /// <param name="path">The path to read.</param>
+        /// <returns>The value read, or "0" if a failure occurred.</returns>
+        private static string ReadMaxUserLimit(string path)
+        {
+            try { return File.ReadAllText(path).Trim(); } catch { }
+            return "0";
```
akoeplinger: Since this returns a string anyway, why not return something like "Couldn't read max user limit" or "N/A" to avoid confusion around the value not really being configured to 0 on the system?

stephentoub: I could do "N/A" I guess.  It's going into a parenthetical in the error message, so a phrase would look strange.  Presumably "N/A" would itself need to be a resource.  This should be a very rare case, and so it didn't seem worth the complication for an error message (I contemplated not having the value at all, but thought it'd be useful), hence the 0.  I could also make it something further non-sensical, like -1.

akoeplinger: Hmm, I don't think it's worth needing to localize N/A, so maybe just use an empty string (if it's really rare then an empty parenthesis shouldn't be too much of a problem). If you prefer a number, then -1 is a better option since you can actually set max_user_watches to 0 while -1 throws an error.

stephentoub: Ok, thanks.  I'll change it to -1.

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
```diff
@@ -64,6 +64,8 @@ private void StartRaisingEvents()
         /// <summary>Stop monitoring the current directory.</summary>
         private void StopRaisingEvents()
         {
+            _enabled = false;
```
sokket: Not really feedback, more just curiosity around what this helps: what does moving the _enabled=false call up here give us? :)

sokket:
LGTM

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs
stephentoub: Mainly consistency across implementations so that we always set it to false first thing in StopRaisingEvents.  But in this case, we also want to make sure it's false even if the the directory handle somehow got invalidated.

sokket: Cool, thanks! Great idea to keep it consistent

src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs
stephentoub: I decided to just have two error messages for each: one with the value and one without.  This matches what various other assemblies do for file-related exceptions with and without known paths.

stephentoub:
Thanks, guys.

PR #3068, opened 2015-09-03
Implement ReadAsync on CurlHandler response stream

### LINKED ISSUE #3067: Implement ReadAsync on CurlHandler's response stream

stephentoub:
Today, CurlHandler provides a response content stream that doesn't override ReadAsync.  As such, any consumers reading data using ReadAsync will actually end up blocking a thread pool thread doing a synchronous Read.

This commit fixes that by implementing ReadAsync to be truly asynchronous with no additional threads blocked.  If data is already available, a task is returned immediately.  Otherwise, the reader asynchronously waits for the writer to publish data and then consumes that data.

The overall design of the response stream remains the same as it was.  However, there are some notable changes:
- A SemaphoreSlim is now used instead of a ManualResetEventSlim for the _writerPublishedDataEvent.  This was done because SemaphoreSlim supports both synchronous and asynchronous waiting.
- The _readerRequestingDataEvent was removed, as it was no longer necessary.
- We no longer allocate an UnmanagedMemoryStream per write operation, instead simply storing the IntPtr buffer and long length on the response stream.
- Additional reorganization was done, primarily to share logic between Read and ReadAsync.
- A bunch of comments and debug asserts were added.

A second commit enables the asynchronous wait to be canceled by either the internal shutdown CancellationToken or by the CancellationToken provided by the caller to ReadAsync.

Fixes #3067 
cc: @vijaykota, @kapilash, @davidsh, @pgavlin, @CIPop

### DIFF
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlResponseMessage.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlResponseMessage.cs
@@ -281,12 +281,29 @@ namespace System.Net.Http
 
         private async Task<int> ReadAsyncCore(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            // There wasn't any data, so wait for the writer to publish more.
+            // Allow cancellation of the wait to be triggered by either the stream shutting down
+            // or by the caller's cancellation token having cancellation requested.
+            CancellationTokenSource cts = cancellationToken.CanBeCanceled ?
+                CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _completionSignal.Token) :
+                null;
             try
             {
-                await _writerPublishedDataEvent.WaitAsync(_completionSignal.Token).ConfigureAwait(false);
+                // There wasn't any data, so wait for the writer to publish more.
+                await _writerPublishedDataEvent.WaitAsync(cts != null ? cts.Token : _completionSignal.Token).ConfigureAwait(false);
             }
-            catch (OperationCanceledException) { }
+            catch (OperationCanceledException)
+            {
+                // If the user's cancellation token has had cancellation requested, propagate that OCE.
+                // Otherwise, if our internal completion signal token had cancellation requested, keep
+                // going, as we'll either need to return a successful 0 (no more data available)
+                // or throw a failure exception.
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+            finally
+            {
+                if (cts != null)
+                    cts.Dispose();
+            }
 
             // Now either cancellation should have been requested or
             // we should have data to read and return.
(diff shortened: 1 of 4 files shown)


src/System.Net.Http/src/System/Net/Http/Unix/CurlResponseMessage.cs
```diff
@@ -53,31 +91,6 @@ internal void SignalFailure(ExceptionDispatchInfo error)
...
 
-        internal unsafe void WaitAndSignalReaders(IntPtr pointer, long length)
-        {
-            try
-            {
-                CheckDisposed();
-                _readerRequestingDataEvent.Wait(Timeout.InfiniteTimeSpan, _cancellationTokenSource.Token);
-                lock (LockObject)
-                {
-                    _innerStream = new UnmanagedMemoryStream((byte*)pointer, length);
```
vijaykota: Nit: You can remove reference to System.IO.UnmanagedMemoryStream

stephentoub: Thanks, done.

src/System.Net.Http/src/System/Net/Http/Unix/CurlResponseMessage.cs
```diff
@@ -126,164 +139,270 @@ public override long Position
...
+        private int ConsumePublishedData(byte[] buffer, int offset, int count)
         {
-            Debug.Assert(Monitor.IsEntered(LockObject), "lock wasn't held");
-            _writerHasDataEvent.Reset();
-            _readerRequestingDataEvent.Set();
+            Debug.Assert(Monitor.IsEntered(LockObject), "Published data must only be accessed while holding the lock");
+
+            // Copy the data into the read buffer
+            int bytesToRead = (int)Math.Min(_availableDataLength, count);
+            Marshal.Copy(_availableData, buffer, offset, bytesToRead);
```
kapilash: Do we need to check the cancellation status here again? Is ```Marshal.Copy``` safe  if there was an error  and ```SignalComplete``` was called (via easyRequest.CleanUp()) ?

stephentoub: It's safe: we're inside of the lock guarding access to the available data. I don't think checking for cancellation again will achieve much... most of the call sites checked for cancellation just before calling the function, anyway.

kapilash: I was worried about the scenario where libcurl may have consumed the native pointer, that it gave us in the curl-callback. I now realize that wont happen. I was wrong.

kapilash:
thanks @stephentoub . LGTM

stephentoub:
Thanks for reviewing, @kapilash!

src/System.Net.Http/src/System/Net/Http/Unix/CurlResponseMessage.cs
stephentoub: I'll add another debug assert to the beginning of the function to reconfirm that there's data available.

vijaykota:
LGTM

PR #3073, opened 2015-09-03
Fix race condition in TestEventListener

### LINKED ISSUE #3070: System.Collections.Concurrent.Tests.EtwTests.TestEtw failed in CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/3588/consoleFull
```
System.Collections.Concurrent.Tests.EtwTests.TestEtw [FAIL]
        System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
        Stack Trace:
              at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
              at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
           d:\j\workspace\dotnet_corefx_windows_release_prtest\src\Common\tests\System\Diagnostics\Tracing\TestEventListener.cs(47,0): at System.Diagnostics.Tracing.TestEventListener.LoadSourceList()
           d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Collections.Concurrent\tests\EtwTests.cs(15,0): at System.Collections.Concurrent.Tests.EtwTests.TestEtw()
```

stephentoub:
One of our ETW tests randomly failed due to a collection being modified while it was concurrently being enumerated.  The problem arose because our TestEventListener we use in the ETW tests has a collection that's populated by calls to OnEventSourceCreated, and it was only expecting the creation of itself to trigger the invocation of this method by its base, but if another thread in the system is creating another EventSource at the same time, that EventSource will also get added to the collection, which could happen as we're enumerating it.  I've added some synchronization to prevent the race condition.

Fixes #3070

### DIFF
--- a/src/Common/tests/System/Diagnostics/Tracing/TestEventListener.cs
+++ b/src/Common/tests/System/Diagnostics/Tracing/TestEventListener.cs
@@ -44,23 +44,34 @@ namespace System.Diagnostics.Tracing
             // and until we finish construction... in that window, OnEventSourceCreated
             // will store the sources into the list rather than try to enable them directly,
             // and then here we can enumerate that list, then clear it out.
-            foreach (var source in _tmpEventSourceList)
+            List<EventSource> sources;
+            lock (_tmpEventSourceList)
+            {
+                sources = _tmpEventSourceList;
+                _tmpEventSourceList = null;
+            }
+            foreach (EventSource source in sources)
             {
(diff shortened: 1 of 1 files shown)


ellismg:
LGTM

stephentoub:
Thanks, Matt.

PR #3079, opened 2015-09-03
Fix multiple MSB3277 build warnings

stephentoub:
Fix several warnings of the form:
```
warning MSB3277: Found conflicts between different versions of the s
ame dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.
```
Brings the number of warnings during a full build back down to 0.

A few tweaks to some test project project.json files, and then the resulting automatic updates to the project.lock.json files.

cc: @mmitche, @ericstj


mellinoe:
LGTM, these are always annoying.

stephentoub:
Yes they are :wink: Thanks, Eric.

PR #3082, opened 2015-09-03
Better CryptoStream async.

AtsushiKan:
Streams that wrap other streams should make sure their
own async methods use the underlying stream's async methods.


AtsushiKan:
cc @stephentoub

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -147,9 +151,36 @@ public override void SetLength(long value)
             throw new NotSupportedException(SR.NotSupported_UnseekableStream);
         }
 
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            CheckReadArguments(buffer, offset, count);
```
Maxwe11: Exceptions thrown from `CheckReadArguments` will be propagated on Task instead of throwing to user immediately.

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -147,9 +151,36 @@ public override void SetLength(long value)
...
+            // conditions with internal buffer indexes in our own streams that 
+            // don't natively support async IO operations when there are multiple 
+            // async requests outstanding, we will block the application's main
+            // thread if it does a second IO request until the first one completes.
+
+            SemaphoreSlim semaphore = AsyncActiveSemaphore;
+            await semaphore.WaitAsync().ForceAsync();
+            try
+            {
+                return await ReadAsyncCore(buffer, offset, count, cancellationToken, useAsync: true);
```
Maxwe11: Could be `.ConfigureAwait(false);` ?

AtsushiKan:
Mac and Linux builds on CI are failing with dozens of errors about

CSC : error CS0656: Missing compiler required member 'System.Diagnostics.DebuggerStepThroughAttribute..ctor' [d:\j\workspace\dotnet_corefx_mac_release_prtest\src\System.Security.Cryptography.Primitives\src\System.Security.Cryptography.Primitives.csproj]

I can't repro this locally. Are we just unable to use "async" and "await" keywords here?

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
AtsushiKan: Probably not. I'll move the async to a helper.

stephentoub:
> I can't repro this locally. Are we just unable to use "async" and "await" keywords here?

I'll take a look in a bit... async/await should definitely be usable.  It's probably some issue regarding what assemblies are/aren't being referenced in the .json files.

AtsushiKan:
Yeah, it turned out I just needed System.Diagnostics.Debug.

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
stephentoub: Ideally it would be, if for no other reason than it'll avoid questions about why it's not.  However, it's likely not necessary, as the ForceAsync Ati's using on the previous line will guarantee that this code is running asynchronously from the caller on a thread pool thread, where there should be a default context/scheduler.

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -290,8 +328,48 @@ public override int Read(byte[] buffer, int offset, int count)
             }
         }
 
+        private static async Task<int> ReadConditionallyAsync(Stream s, byte[] buffer, int offset, int count, CancellationToken cancellationToken, bool useAsync)
+        {
+            if (useAsync)
+                return await s.ReadAsync(buffer, offset, count, cancellationToken);
+            else
+                return s.Read(buffer, offset, count);
+        }
```
stephentoub: I can see why this helper is tempting, and functionally it's fine, but it's going to add more overhead than you'd expect.  If you really want it as a helper, it'd be better to write it as a non-async method, e.g.
```C#
private static Task<int> ReadConditionallyAsync(Stream s, byte[] buffer, int offset, int count, CancellationToken cancellationToken, bool useAsync)
{
    return useAsync ?
        s.ReadAsync(buffer, offset, count, cancellationToken) :
        Task.FromResult(s.Read(buffer, offset, count);
}
```
as that'll have fewer allocations (just the extra one for the Task.FromResult in the synchronous case) than the previous version (which will likely have at least a Task allocation, plus several additional allocations if the async operation doesn't complete synchronously).

But... you can avoid all of those if you just don't have the helper, and the two places you need it just writing it out, e.g.
```
amountRead += useAsync ?
    await _stream.ReadAsync(tempInputBuffer, _inputBufferIndex, numWholeBlocksInBytes - _inputBufferIndex, cancellationToken) :
    _stream.Read(tempInputBuffer, _inputBufferIndex, numWholeBlocksInBytes - _inputBufferIndex);
```

src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptoStream.cs
```diff
@@ -382,6 +463,14 @@ public override void Write(byte[] buffer, int offset, int count)
             return;
         }
 
+        private static async Task WriteConditionallyAsync(Stream s, byte[] buffer, int offset, int count, CancellationToken cancellationToken, bool useAsync)
+        {
+            if (useAsync)
+                await s.WriteAsync(buffer, offset, count, cancellationToken);
+            else
+                s.Write(buffer, offset, count);
+        }
```
stephentoub: Same comment as on the ReadConditionallyAsync method.

src/System.Security.Cryptography.Primitives/src/project.json
```diff
@@ -1,11 +1,13 @@
 {
   "dependencies": {
+    "System.Diagnostics.Debug": "4.0.0",
     "System.Diagnostics.Contracts": "4.0.0",
     "System.IO": "4.0.10",
     "System.Resources.ResourceManager": "4.0.0",
-    "System.Runtime": "4.0.20"
+    "System.Runtime": "4.0.20",
+    "System.Threading": "4.0.0"
```
stephentoub: I'm surprised you don't need System.Threading.Tasks...?

stephentoub:
Other than my perf-related comments, LGTM.

PR #3088, opened 2015-09-03
System.IO.Compression: use consistent namespace name for tests

akoeplinger:
As per https://github.com/dotnet/corefx/issues/2898 the convention is `$(Namespace of Type under Test).Tests`
Some tests were in `.Test`, this change renames them to `.Tests`

### DIFF
--- a/src/Common/tests/Compression/Utilities/ZipTestHelper.cs
+++ b/src/Common/tests/Compression/Utilities/ZipTestHelper.cs
@@ -7,7 +7,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
 
-namespace System.IO.Compression.Test
+namespace System.IO.Compression.Tests
 {
     public partial class ZipTest
     {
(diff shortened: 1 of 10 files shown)


stephentoub:
cc: @ianhays 

LGTM

I notice that there are a few test files here that don't have namespace at all, namely the ones in the Utilities folder.  Did you want to add namespaces there, too?

akoeplinger:
> I notice that there are a few test files here that don't have namespace at all

See https://github.com/dotnet/corefx/issues/2898#issuecomment-137598997

stephentoub:
> See #2898 (comment)

Ah, ok.  So you're thinking we effectively split this into two buckets: existing namespaces that just need to be tweaked (not influencing any other code, other than potentially needing some additional usings elsewhere, and the like) and code without namespaces at all, which would be much more impactful.  That makes sense.

akoeplinger:
Yep, exactly :)

stephentoub:
@dotnet-bot test this please (failed due to https://github.com/dotnet/corefx/issues/3075)

ianhays:
LGTM

stephentoub:
@dotnet-bot test this please

PR #3090, opened 2015-09-04
Fix #3075

### LINKED ISSUE #3075: ProcessStartInfoTests.TestEnvironmentOfChildProcess failed on Windows in CI

Maxwe11:
fixes #3075

### DIFF
--- a/src/Scenarios/tests/InterProcessCommunication/EnvironmentOfChildProcess.cs
+++ b/src/Scenarios/tests/InterProcessCommunication/EnvironmentOfChildProcess.cs
@@ -1,0 +1,80 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Pipes;
+using System.Linq;
+using Xunit;
+
+namespace InterProcessCommunication.Tests
+{
+    public class EnvironmentOfChildProcess : RemoteExecutorTestBase
+    {
+        [Fact]
+        public void Test()
+        {
+            var expectedEnv = new HashSet<string>();
+            var actualEnv = new HashSet<string>();
+
+            using (var inbound = new AnonymousPipeServerStream(PipeDirection.In, HandleInheritability.Inheritable))
+            using (var reader = new StreamReader(inbound))
+            using (var remote = RemoteInvoke(GetEnvironmetOfChildProcess, inbound.GetClientHandleAsString()))
+            {
+                inbound.DisposeLocalCopyOfClientHandle();
+
+                foreach (KeyValuePair<string, string> envVar in remote.Process.StartInfo.Environment)
+                {
+                    actualEnv.Add(envVar.Key + "=" + envVar.Value);
+                }
+
+                string line;
+                while ((line = reader.ReadLine()) != null)
+                {
+                    expectedEnv.Add(line);
+                }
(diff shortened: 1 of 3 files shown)


src/Scenarios/tests/InterProcessCommunication/EnvironmetOfChildProcess.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Pipes;
+using System.Linq;
+using Xunit;
+
+namespace InterProcessCommunication.Tests
+{
+    public class EnvironmetOfChildProcess : RemoteExecutorTestBase
```
akoeplinger: typo: `Environment` (also in the filename)

stephentoub:
@Maxwe11, what was wrong with the previous test, and how does this one fix it?  Trying to understand why we were seeing an intermittent failure.

Maxwe11:
@stephentoub the problem was with that default `Console` encoding doesn't handle non ASCII chars (e.g. K**ö**plinger). I see the same behavior on Desktop.

akoeplinger:
@Maxwe11 hmm, but why didn't https://github.com/dotnet/corefx/pull/3086 fail with the same error then?

stephentoub:
I'm not convinced the encoding issue is the only issue (and if it was, I'd prefer to see a fix that just set up the ProcessStartInfo and Console encodings correctly rather than rewriting the test entirely).  I just tried adding ```Environment.SetEnvironmentVariable(TempVariableName, "\x1234\x5678");``` to the start of the test... with that, sometimes the test fails, sometimes it passes.  I'd like to better understand why there's an inconsistency before any attempted fix is made.

stephentoub:
I think I know the reason... I'll experiment and post back here.

Maxwe11:
replaced by #3102

stephentoub:
Thanks, @Maxwe11.

PR #3097, opened 2015-09-04
Short-circuit on SequenceEqual

JonHanna:
Short-circuit to false on both sequences are ICollection<T> and different Count

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -979,6 +979,14 @@ namespace System.Linq
             if (comparer == null) comparer = EqualityComparer<TSource>.Default;
             if (first == null) throw Error.ArgumentNull("first");
             if (second == null) throw Error.ArgumentNull("second");
+
+            ICollection<TSource> firstCol = first as ICollection<TSource>;
+            if (firstCol != null)
+            {
+                ICollection<TSource> secondCol = second as ICollection<TSource>;
+                if (secondCol != null && firstCol.Count != secondCol.Count) return false;
+            }
+
             using (IEnumerator<TSource> e1 = first.GetEnumerator())
             using (IEnumerator<TSource> e2 = second.GetEnumerator())
             {
@@ -986,9 +994,8 @@ namespace System.Linq
                 {
                     if (!(e2.MoveNext() && comparer.Equals(e1.Current, e2.Current))) return false;
                 }
-                if (e2.MoveNext()) return false;
+                return !e2.MoveNext();
             }
-            return true;
         }
 
         public static IEnumerable<TSource> AsEnumerable<TSource>(this IEnumerable<TSource> source)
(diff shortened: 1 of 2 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -979,6 +979,16 @@ public static bool SequenceEqual<TSource>(this IEnumerable<TSource> first, IEnum
             if (comparer == null) comparer = EqualityComparer<TSource>.Default;
             if (first == null) throw Error.ArgumentNull("first");
             if (second == null) throw Error.ArgumentNull("second");
+
+            if (first == second) return true;
```
JonHanna: This is a standard enough short-circuit on any O(n) equality check, but I suppose something **could** return different results when enumerated twice. To have that and to test for it being equal seems bizarre, but is it bizarre enough to consider this safe?

stephentoub: It's pretty rare to call SequenceEqual with the same object.  And if you're doing so on purpose, then it's almost certainly because you expect it's possible that each enumeration could return different results.  So I think this checks buys us very little and could potentially be problematic, and shouldn't be done.

JonHanna: Well, one doesn't generally do it on purpose as by-the-by as part of a series of comparisons, but I think that case too is probably unlikely here. And if it is done, it's probably as part of some more general equality check that should probably include the identity-check itself. I'm going to take this bit out and squash it to a commit with just the collection test.

VSadov:
LGTM

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -979,6 +979,14 @@ public static bool SequenceEqual<TSource>(this IEnumerable<TSource> first, IEnum
             if (comparer == null) comparer = EqualityComparer<TSource>.Default;
             if (first == null) throw Error.ArgumentNull("first");
             if (second == null) throw Error.ArgumentNull("second");
+
+            ICollection<TSource> firstCol = first as ICollection<TSource>;
+            if (firstCol != null)
+            {
+                ICollection<TSource> secondCol = second as ICollection<TSource>;
+                if (secondCol != null && firstCol.Count != secondCol.Count) return false;
+            }
```
stephentoub: Do we have tests that exercise all of these new paths?

JonHanna: We don't for secondCol == null. I'll add some soon.

JonHanna: There, all the permutations.

src/System.Linq/tests/SequenceEqualTests.cs
```diff
@@ -37,6 +37,16 @@ public int GetHashCode(string obj)
             }
         }
         
+        private static IEnumerable<T> ForceNotCollection<T>(IEnumerable<T> source)
+        {
+            foreach(T item in source) yield return item;
```
justinvp: nit: space after "foreach"

stephentoub:
@dotnet-bot test this please (failed due to a memory issue somewhere in what appears to be the crypto libs)

stephentoub:
Thanks.  LGTM.

Issue #2768, opened 2015-08-12
CoreCLR: TryGetRawMetadata


tmat:
Scenario
====

Roslyn Scripting API provides applications with the capability to evaluate C# and VB code snippets. 

For example,

```C#
using Microsoft.CodeAnalysis.Scripting.CSharp;

public class Globals { public int x; public int y; }

class Program
{
  public static void Main()
  {
    var globals = new Globals() { x = 1, y = 2 };
    var options = ScriptOptions.Default.WithReferences(typeof(MyMath).Assembly);

    var script = CSharpScript.Create("MyMath.CalculateStuff(x, y)", options);

    Console.WriteLine(script.RunAsync(globals).Result);
  }
}
```

In the example above ```MyMath``` is a type defined in an assembly referenced by and deployed with the program. The script is compiled with the following metadata references: mscorlib, the assembly containing the ```Globals``` type and an assembly containing ```MyMath``` class. All these assemblies are implementation (runtime) assemblies, not contracts. Since we are compiling scripts at runtime contracts are not relevant. The application might also run on a machine that doesn’t have any SDK. It is natural for scripting to be able to execute virtually any (public) runtime method available in any loaded assembly. Just like Reflection invocation.

On Desktop CLR we use Assembly.Location to find the metadata image for given assembly. 
On Core CLR this property is not available and there is no other way to get the metadata of a given runtime assembly.

Proposed API
====

We propose that the metadata of a runtime assembly are exposed thru new APIs in a form that is efficiently readable using MetadataReader of System.Reflection.Metadata library. Since metadata might not be available for all assemblies (such as AssemblyBuilder, .NET Native images, etc.) the API is allowed to return false for such assemblies. Roslyn scripting API would then report an error that the script can’t reference specified assembly.

```C#
namespace System.Reflection.Metadata
{
  public static class AssemblyExtensions 
  {
    public unsafe static bool TryGetRawMetadata(this Assembly assembly, out byte* blob, out int length);
  }
}
```

**Lifetime considerations**

The pointer would stay valid as long as the ```AssemblyLoadContext``` with which is the assembly associated is alive. It is the responsibility of the caller to keep the Assembly object alive while accessing the metadata blob.

Considered alternatives
====

It was suggested that Roslyn uses Reflection API to read the metadata. There are multiple issues with such approach. 

1)	Expected bad performance (both working set and time)

[...]

PR #3100, opened 2015-09-04
Add AssemblyExtensions.TryGetRawMetadata to the System.Reflection.Met…

kouvel:
…adata namespace.

This patch adds the new API to the System.Runtime.Loader reference assembly. Tests will be added separately, once the new API is published and can be consumed.

Part of dotnet/corefx#2768

### DIFF
--- a/src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
+++ b/src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
     <OutputType>Library</OutputType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="System.Runtime.Loader.cs" />
--- a/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs
+++ b/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs
@@ -5,6 +5,14 @@
 // ------------------------------------------------------------------------------
 
 
+namespace System.Reflection.Metadata
+{
+    public static partial class AssemblyExtensions
+    {
+        [CLSCompliant(false)] // out byte* blob
+        public unsafe static bool TryGetRawMetadata(this System.Reflection.Assembly assembly, out byte* blob, out int length) { blob = default(byte*); length = default(int); return default(bool); }
+    }
+}
 namespace System.Runtime.Loader
 {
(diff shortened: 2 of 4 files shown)


kouvel:
CC @jkotas @tmat @nguerrera

jkotas:
LGTM

nguerrera:
LGTM

weshaggard:
@dotnet-bot test this please

src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
```diff
@@ -4,8 +4,10 @@
   <PropertyGroup>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
     <OutputType>Library</OutputType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="System.Reflection.Metadata.cs" />
```
weshaggard: Generally we keep all the APIs in the single cs file that matches the name of the contract. @nguerrera were you thinking we should change this given you signed off?

kouvel: I thought the name of the file corresponds to the namespace. I guess this is a rare case where the namespace name is different from the library name, I didn't find another case of this.

kouvel:
@dotnet-bot test this please

src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
jkotas: It is actually very common that the library contains namespaces that are different from its name...

jkotas: I would move the new API into existing single .cs file as @weshaggard suggested.

stephentoub:
LGTM

src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
kouvel: Done

Maxwe11:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please (same crypto failure)

src/System.Runtime.Loader/ref/System.Runtime.Loader.csproj
weshaggard: Thanks. I should I guess call out why we do this which is we have tools that we use to generate these and they generate the entire reference assembly in one cs file.

weshaggard:
LGTM

nguerrera:
Please rebase away that merge commit before merging this in.

PR #3102, opened 2015-09-04
Fix TestEnvironmentOfChildProcess to set output encoding

### LINKED ISSUE #3075: ProcessStartInfoTests.TestEnvironmentOfChildProcess failed on Windows in CI

stephentoub:
Three issues all conspiring together:
- On many machines, none of the environment variables would have non-ASCII characters.
- Some of the Process tests mutate the ambient console encoding on Windows
- The default standard output encoding on Windows results in Unicode characters getting lost in translation

The net effect of this is that the test would fail randomly.

The fix:
- Adds an environment variable that includes non-ASCII characters to force the issue
- Adds an xunit attribute to prevent parallelization of the Process tests
- Forces the test to use UTF8 for the standard output encoding

cc: @akoeplinger, @Maxwe11 
I believe this fixes #3075 and can replace #3090.
The change looks larger than it actually is due to adding a try/catch and adding lots of indentation as a result.

### DIFF
--- a/src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
+++ b/src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs
@@ -10,6 +10,7 @@ using System.Security.Principal;
 using Microsoft.Win32;
 using Microsoft.Win32.SafeHandles;
 using Xunit;
+using System.Text;
 
 namespace System.Diagnostics.ProcessTests
 {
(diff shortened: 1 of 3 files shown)


akoeplinger:
:+1: 

> The change looks larger than it actually is due to adding a try/catch and adding lots of indentation as a result.

Yep, looks nicer with ?w=1: https://github.com/dotnet/corefx/pull/3102/files?w=1

Maxwe11:
:+1:

stephentoub:
Thanks.

PR #3104, opened 2015-09-05
Add SelectXXXIterator separate to WhereSelectXXXIterator

JonHanna:
`WhereSelectXXXIterator` classes can no longer have a null predicate, so that case need no longer be checked for. (Add readonly and assertions to ensure this assumption is true.)

Add IList<T>-based Select specialised class, on same basis; increases cases covered by optimised `ToArray()`, also reduces number of branches non-ilist sources (likely most common case) go
through though at cost of one more for array or list.

IArrayProvider<TElement>.ToArray() will now never return null, so don't test for the case that it may have.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading;
 
 namespace System.Linq
@@ -46,11 +47,16 @@ namespace System.Linq
             if (selector == null) throw Error.ArgumentNull("selector");
             Iterator<TSource> iterator = source as Iterator<TSource>;
             if (iterator != null) return iterator.Select(selector);
-            TSource[] array = source as TSource[];
-            if (array != null) return new WhereSelectArrayIterator<TSource, TResult>(array, null, selector);
-            List<TSource> list = source as List<TSource>;
-            if (list != null) return new WhereSelectListIterator<TSource, TResult>(list, null, selector);
-            return new WhereSelectEnumerableIterator<TSource, TResult>(source, null, selector);
+            IList<TSource> ilist = source as IList<TSource>;
+            if (ilist != null)
+            {
+                TSource[] array = source as TSource[];
+                if (array != null) return new SelectArrayIterator<TSource, TResult>(array, selector);
+                List<TSource> list = source as List<TSource>;
+                if (list != null) return new SelectListIterator<TSource, TResult>(list, selector);
+                return new SelectIListIterator<TSource, TResult>(ilist, selector);
+            }
+            return new SelectEnumerableIterator<TSource, TResult>(source, selector);
         }
 
         public static IEnumerable<TResult> Select<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, int, TResult> selector)
@@ -113,9 +119,15 @@ namespace System.Linq
 
             public abstract bool MoveNext();
 
-            public abstract IEnumerable<TResult> Select<TResult>(Func<TSource, TResult> selector);
+            public virtual IEnumerable<TResult> Select<TResult>(Func<TSource, TResult> selector)
+            {
+                return new SelectEnumerableIterator<TSource, TResult>(this, selector);
+            }
 
-            public abstract IEnumerable<TSource> Where(Func<TSource, bool> predicate);
+            public virtual IEnumerable<TSource> Where(Func<TSource, bool> predicate)
+            {
+                return new WhereEnumerableIterator<TSource>(this, predicate);
+            }
 
             object IEnumerator.Current
             {
@@ -135,12 +147,14 @@ namespace System.Linq
 
         internal class WhereEnumerableIterator<TSource> : Iterator<TSource>
         {
-            private IEnumerable<TSource> _source;
-            private Func<TSource, bool> _predicate;
+            private readonly IEnumerable<TSource> _source;
+            private readonly Func<TSource, bool> _predicate;
             private IEnumerator<TSource> _enumerator;
 
             public WhereEnumerableIterator(IEnumerable<TSource> source, Func<TSource, bool> predicate)
             {
+                Debug.Assert(source != null);
+                Debug.Assert(predicate != null);
                 _source = source;
                 _predicate = predicate;
             }
@@ -197,12 +211,14 @@ namespace System.Linq
 
         internal class WhereArrayIterator<TSource> : Iterator<TSource>
         {
-            private TSource[] _source;
-            private Func<TSource, bool> _predicate;
+            private readonly TSource[] _source;
+            private readonly Func<TSource, bool> _predicate;
             private int _index;
 
             public WhereArrayIterator(TSource[] source, Func<TSource, bool> predicate)
             {
+                Debug.Assert(source != null);
+                Debug.Assert(predicate != null);
                 _source = source;
                 _predicate = predicate;
             }
@@ -244,12 +260,14 @@ namespace System.Linq
 
         internal class WhereListIterator<TSource> : Iterator<TSource>
         {
-            private List<TSource> _source;
-            private Func<TSource, bool> _predicate;
+            private readonly List<TSource> _source;
+            private readonly Func<TSource, bool> _predicate;
             private List<TSource>.Enumerator _enumerator;
 
             public WhereListIterator(List<TSource> source, Func<TSource, bool> predicate)
             {
+                Debug.Assert(source != null);
+                Debug.Assert(predicate != null);
                 _source = source;
                 _predicate = predicate;
             }
@@ -296,13 +314,16 @@ namespace System.Linq
 
         internal class WhereSelectEnumerableIterator<TSource, TResult> : Iterator<TResult>
         {
-            private IEnumerable<TSource> _source;
-            private Func<TSource, bool> _predicate;
-            private Func<TSource, TResult> _selector;
+            private readonly IEnumerable<TSource> _source;
+            private readonly Func<TSource, bool> _predicate;
+            private readonly Func<TSource, TResult> _selector;
             private IEnumerator<TSource> _enumerator;
 
             public WhereSelectEnumerableIterator(IEnumerable<TSource> source, Func<TSource, bool> predicate, Func<TSource, TResult> selector)
             {
+                Debug.Assert(source != null);
+                Debug.Assert(predicate != null);
+                Debug.Assert(selector != null);
                 _source = source;
                 _predicate = predicate;
                 _selector = selector;
@@ -335,7 +356,7 @@ namespace System.Linq
                         while (_enumerator.MoveNext())
                         {
                             TSource item = _enumerator.Current;
-                            if (_predicate == null || _predicate(item))
+                            if (_predicate(item))
                             {
                                 current = _selector(item);
                                 return true;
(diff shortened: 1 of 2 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -46,11 +47,16 @@ private static IEnumerable<TSource> WhereIterator<TSource>(IEnumerable<TSource>
...
-            List<TSource> list = source as List<TSource>;
-            if (list != null) return new WhereSelectListIterator<TSource, TResult>(list, null, selector);
-            return new WhereSelectEnumerableIterator<TSource, TResult>(source, null, selector);
+            IList<TSource> ilist = source as IList<TSource>;
+            if (ilist != null)
+            {
+                TSource[] array = source as TSource[];
+                if (array != null) return new SelectArrayIterator<TSource, TResult>(array, selector);
+                List<TSource> list = source as List<TSource>;
+                if (list != null) return new SelectListIterator<TSource, TResult>(list, selector);
```
stephentoub: Why the specialization for ```List<T>```?  I assume this was done purely for performance, e.g. avoiding the interface calls?  If so, what kinds of gains do you see?

stephentoub: I see, we have the ```List<T>``` special case for Where, but not the ```IList<T>`` case.  Why was it important to add that one for Select?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -400,16 +419,153 @@ public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> s
...
+                return false;
+            }
+
+            public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> selector)
+            {
+                return new WhereSelectListIterator<TSource, TResult2>(_source, _predicate, CombineSelectors(_selector, selector));
+            }
+        }
+
+        internal sealed class SelectEnumerableIterator<TSource, TResult> : Iterator<TResult>
```
stephentoub: What's the value of writing SelectEnumerableIterator out by hand rather than just implementing it as:
```foreach (var item in source) yield return selector(source);```?  It doesn't implement ToArray, and no other code tries to use its selector, so I'm not clear what value this provides.  Is it purely for the Select(...).Select(...) case in order to be able to use CombineSelectors?

File: src/System.Linq/src/System/Linq/Enumerable.cs [hunk 3]
```diff
@@ -447,14 +603,10 @@ public override bool MoveNext()
                         state = 2;
                         goto case 2;
                     case 2:
-                        while (_enumerator.MoveNext())
+                        if (_enumerator.MoveNext())
```
stephentoub: I was expecting to see here code that used ```List<T>``` indexer rather than getting an enumerator and using MoveNext/Current.  Is that not good for some reason?

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -351,22 +372,20 @@ public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> s
             {
```
stephentoub: We used to have test coverage for this method.  With these changes, this method is now uncovered by tests.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -400,16 +419,153 @@ public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> s
...
```
stephentoub: Ditto.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -464,18 +616,76 @@ public override bool MoveNext()
 
             public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> selector)
             {
-                return new WhereSelectListIterator<TSource, TResult2>(_source, _predicate, CombineSelectors(_selector, selector));
```
stephentoub: Ditto.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -464,18 +616,76 @@ public override bool MoveNext()
...
+                }
+                base.Dispose();
+            }
+
+            public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> selector)
+            {
+                return new SelectIListIterator<TSource, TResult2>(_source, CombineSelectors(_selector, selector));
+            }
+
+            public TResult[] ToArray()
```
stephentoub: Please ensure we have tests for this; it doesn't appear to be exercised by any of the tests.

stephentoub:
cc: @VSadov 

Thanks, @JonHanna. Overall it looks good to me.  I'd like to see the test gaps addressed.  It'd also be helpful to understand what the performance improvements look like.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: It's the IList case that's truly new here, with the others being derived from the previous code. I would probably have just done the IList case if coming at a tabla rasa, or maybe just the IList and the array, but I'm loath to remove a previous optimisation unless I'm convinced it's actually making things slower (or if it hits a buggy case, obviously).
The IList case can benefit in terms of ToArray (and there's scope for further work on the cases where an iterator can know its size in advance) with select, so I added that, but it can't with Where, so I left that as is.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: CombineSelectors is its only real benefit, though there's a tiny benefit when it's subsequently used in a Where.
It was though pretty quick to write by hand given `WhereSelectEnumerableIterator` as a starting point, and there's greater consistency between the different cases.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: That would change the behaviour if the source list is altered during enumeration. (I've never been convinced the dangers of that are such as to require an outright ban, but it's way past the point of debating that). The test `Select_SourceListGetsModifiedDuringIteration_ExceptionIsPropagated` checks on this.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: > It was though pretty quick to write by hand given WhereSelectEnumerableIterator as a starting point, and there's greater consistency between the different cases.

in general I'm less concerned in these cases about how hard it was to write, and more concerned with an increase in the amount of code that needs to be maintained.

src/System.Linq/src/System/Linq/Enumerable.cs
stephentoub: That makes sense.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Easy to write doesn't always correspond with easy to maintain, but it can. I see the main advantage being the consistency, which could help maintenance if a future change can be consistently applied.
Or not, since that's a bit crystal-ball.
In the balance, there is also CombineSelectors.

stephentoub: And for CombineSelectors, this was an optimization that already existed for this case, and by splitting Select out of the WhereSelect, it kind of forces the hand. Ok.

src/System.Linq/src/System/Linq/Enumerable.cs
justinvp: Should `SelectListIterator.ToArray` use `List<T>.Enumerator` as well, for the same reason?

JonHanna: No need. List<T>.Enumerator throws to prohibit code like:

```C#
foreach(var item in list)
{
  // alter list here and don't jump out of the loop
}

And `Select_SourceListGetsModifiedDuringIteration_ExceptionIsPropagated` insists that we throw on:

```C#
foreach(var item in list.Select(λ))
{
  // alter list here and don't jump out of the loop
}

`ToArray` doesn't allow for that possibility, so iterating is a valid optimisation.

(There's the potential for problems if the source list to be changed during the call by another thread, but that involves getting non-threadsafe behaviour from methods that don't claim to be threadsafe, and could fail either way).

**Comment by JonHanna** [comment]
I won't get to those tests for a couple of days, but agree they should certainly be in there before this is merged.

**Comment by stephentoub** [comment]
> I won't get to those tests for a couple of days, but agree they should certainly be in there before this is merged.

@JonHanna, just checking in... you still working on this?  No rush, just verifying.

**Comment by JonHanna** [comment]
@stephentoub I will be, but apart from looking in on notifications, I won't be doing anything for at least another few days, as I'm busy with several matters.

**Comment by stephentoub** [comment]
Ok, thanks for the update. Sounds good.

**Comment by JonHanna** [comment]
@stephentoub I've added tests to cover those cases, copying those for the sequence `.Where(…).Select(…)` to deal with `.Where(…).Select(…).Select(…)`.

I wrote something that compares the before and after of these changes at https://github.com/JonHanna/EnumerablePerformanceComparison. It has an `Enumerable` class in the namespace `System.LinqTest.New` and another in `System.LinqTest.Old` and switching which namespace is imported into the test class between runs changes which is tested. I ran it through Windows DNX on a core i7 with the CPU set to have a minimum and maximum of 90% to reduce the impact of Turbo Boost on the results.

Combined results can be seen at https://github.com/JonHanna/EnumerablePerformanceComparison/raw/master/compare.ods

The tests are run for both `ToArray` and a `Consume` that simply spins through the enumerable as quickly as possible. The times include the `Select` and `Where` operations. The Where predicates are pretty artificial (one that is always false, one that is always true, and one that toggles a captured boolean and returns that). Different element types are tried, for arrays, lists, non-list enumerables, readonly collections and a cut down copy of `List<T>` (these last two giving two different `IList` implementations that don't have their own specific paths). The sizes of the collections are a few multiples of 10 and of 2 in case binary-round sizes gave particularly good or particularly bad results.

The biggest change was, as expected, for `ToArray()` on a `Select` without `Where` where the source is an `IList` that is not a `List` or an array. This case is improved by around 60%, which I have to say is even better than I'd thought it would be.

The other cases where much less pronounced, often being within each others margin of error (showing an improvement in one run, and a deterioration in another). However, they average out over runs at around 5% improvement. Cases where the source was an array were the least improved (quite a few runs showing deterioration, and averaging only slightly improved) but non-list enumerables (probably the most common case, especially if there are other linq operations prior to this) were the most improved, around 6-7% for `ToArray()` and approaching 10% for consuming the enumerable.

File: src/System.Linq/src/System/Linq/Enumerable.cs [hunk 8]
```diff
@@ -135,12 +147,14 @@ void IEnumerator.Reset()
 
         internal class WhereEnumerableIterator<TSource> : Iterator<TSource>
```
Maxwe11: `private` ?

JonHanna:
Oh, one case I new would be affected negatively was that of an array or list source to a `Where` where the source has zero elements. It's affected because there's no longer the optimisation for that situation in 780ef34 which I'd introduced because at the time it was cheap, but with these changes it no longer is. Since that optimisation had only been introduced recently, and since the case is a cheap one anyway (being only hit when there are zero results after all) I don't think that case going back to previous timings (perhaps even slightly better) is a big concern.

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: It probably should be. This class has been `internal` since it was brought into the repository, and I made the comparable classes introduced here, like `SelectEnumerableIterator`, `internal` to match that. If there's no good reason for these to be `internal` then I think they should all be made `private` to give a better ability to eyeball that they're correct. While I made a few fields `readonly` for similar ability to assure oneself of correctness at a glance, that was only because I'd changed the degree that invariant was depended on (if `_predicate` changed, part of my logic was flawed).
In other words, I think you're right, but I think I shouldn't change it right here were it's not relevant.

stephentoub:
Thanks, @JonHanna.

@VSadov, thoughts?

VSadov:
LGTM

stephentoub:
@JonHanna, can you rebase this, at least to get rid of the merge commit in there?

JonHanna:
@stephentoub squashed.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -400,16 +419,153 @@ public override IEnumerable<TResult2> Select<TResult2>(Func<TResult, TResult2> s
...
+            {
+                return new SelectEnumerableIterator<TSource, TResult>(_source, _selector);
+            }
 
+            public override void Dispose()
+            {
+                if (_enumerator != null)
+                {
+                    _enumerator.Dispose();
+                    _enumerator = null;
```
stephentoub: Hmm. I just realized, this is going to cause a NullReferenceException in MoveNext where one didn't previously exist, right?

JonHanna: No, the base Dispose is called next, which guarantees the state checked by MoveNext will ignore this field and return false.

stephentoub: Ah, you're right, thanks.

stephentoub:
LGTM

VSadov:
LGTM

PR #3105, opened 2015-09-06
Avoid blocking CurlHandler write callback

### LINKED ISSUE #3089: CurlHandler blocking write callback can lead to deadlock
With the current implementation of CurlHandler in HttpClient on Unix, when downloaded data is available, libcurl invokes a callback to provide that data to the application.  Until a reader is available to consume all of that data, the callback blocks.  I'd previously noted this as a performance problem to be fixed, but it's actually a much worse problem in that it can easily lead to deadlock.

Here's a new test to demonstrate:
```C#
[Fact]
public async Task GetAsync_ResponseHeadersRead_ReadFromEachIterativelyDoesntDeadlock()
{
    {
        const int NumGets = 5;
        Task<HttpResponseMessage>[] responseTasks = 
            Enumerable.Range(0, NumGets)
            .Select(_ => client.GetAsync(HttpTestServers.RemoteGetServer, HttpCompletionOption.ResponseHeadersRead))
            .ToArray();
        for (int i = responseTasks.Length - 1; i >= 0; i--)
        {
            using (HttpResponseMessage response = await responseTasks[i])
            {
                await AssertSuccessfulGetResponse(response, HttpTestServers.RemoteGetServer, _output);
            }
        }
    }
}
```
Here we kick off N downloads and then iteratively process each to completion before moving on to process the next.  This means that we only try to read from a single download at a time, and if this doesn't happen to be the same download that has data available next, our reader will block waiting for data, which will never come because the libcurl callback is blocked waiting for a reader for a different download.

We should fix this by employing libcurl's ability to pause connections.  When the write callback is invoked, if there isn't currently a read pending on that download, the write callback should return to libcurl to inform it that the connection needs to be paused.  When the reader later comes along, it needs to unpause that connection.  Obviously there will be some synchronization required here, but it shouldn't be too bad.

cc: @vijaykota, @kapilash

stephentoub:
When receiving the body of an HTTP response, libcurl provides the data received via a callback that's invoked synchronously from the thread pumping libcurl's processing.  This callback can't return until it's handed off all of the available data, and thus the current implementation of CurlHandler blocks that thread until a reader has consumed the data.  This is both expensive and dangerous.  It's expensive because we end up blocking a thread until some reader comes along, and it's dangerous because this can easily lead to deadlock.  A libcurl multi handle may be used by only one thread at a time, which means that if the handle is being used by a thread while blocked on the write callback, no other thread can access the multi handle, which means no other callbacks for other concurrent requests can be invoked.  If multiple operations were queued, and then the data for them arrived in an order that didn't exactly match the order of readers consuming the data, deadlock would ensue.

This change makes it so that we no longer block the callback, taking advantage of a pausing mechanism libcurl provides that allows one to pause a connection until it's needed.

I rewrote the CurlResponseStream to manage this.  Now when the callback has data to write, it attempts to transfer it to the stream.  If there's a reader already waiting for the data (represented by a task stored on the stream), the callback simply transfers its data to the reader.  If the callback has too much data than would fit in the reader, the remaining data is buffered on the stream.  If instead the callback finds that either there isn't a waiting task or that there's already some remaining data buffered from a previous callback, it instead pauses the connection, such that no more callbacks will be invoked until we unpause.  From the reader side, a reader who tries to read data from the stream can pull data from the remaining buffer if there is any, or can register a task with the stream that will be filled with data at some point when callback has data available.

As part of debugging some issues with this change, I added a verbose tracing mechanism that's enabled by compiling the library with the CURLHANDLER_VERBOSE constant (the same that we use for enabling libcurl's verbose diagnostic output).  I left the tracing in as I found it valuable in diagnosing issues with the library, but we can remove it later if desired.

I also added a test that deadlocks before the change and that completes successfully after.

cc: @vijaykota, @kapilash
Fixes #3089


src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
```diff
@@ -318,9 +329,52 @@ private void Process()
...
+
+                    case IncomingRequestType.Cancel:
+                        Debug.Assert(easy._associatedMultiAgent == this, "Should only cancel associated easy requests");
+                        Debug.Assert(easy.CancellationToken.IsCancellationRequested, "Cancellation should have been requested");
+                        FindAndFailActiveRequest(multiHandle, easy, new OperationCanceledException(easy.CancellationToken));
+                        break;
+
+                    case IncomingRequestType.Unpause:
+                        Debug.Assert(easy._associatedMultiAgent == this, "Should only unpause associated easy requests");
+                        if (!easy.EasyHandle.IsClosed)
```
vijaykota: Nit: Would be good to consistently use only members or only Properties of EasyRequest

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
```diff
@@ -513,18 +588,10 @@ private static size_t CurlReceiveHeadersCallback(IntPtr buffer, size_t size, siz
...
-                            //       potentially return CURL_WRITEFUNC_PAUSE to pause the connection
-                            //       until a reader is ready to read, at which point we could unpause.
-                            if (size != 0)
-                            {
-                                easy.ResponseMessage.ContentStream.WriteDataAndWaitForConsumption(buffer, (long)size);
-                            }
-
-                            return size;
+                            // Try to transfer the data to a reader.  This will return either the
+                            // amount the amount of data transferred (equal to the amount requested
```
vijaykota: Nit: "the amount" got repeated

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
```diff
@@ -154,11 +160,31 @@ private void RequestWakeup()
...
+                // It's possible but unlikely that there will be tons of extra data in the pipe, 
+                // more than we end up reading out here (it's unlikely because we only write a byte to the pipe when 
+                // transitioning from 0 to 1 incoming request).  In that unlikely event, the worst
+                // case will be that the next one or more waits will wake up immediately, with each one
+                // subsequently clearing out more of the pipe.
+                const int ClearBufferSize = 64; // sufficiently large to clear the pipe in any normal case
+                byte* clearBuf = stackalloc byte[ClearBufferSize];
+                long bytesRead;
+                while (Interop.CheckIo(bytesRead = (long)Interop.libc.read(_wakeupRequestedPipeFd, clearBuf, (IntPtr)ClearBufferSize))) ;
+                VerboseTraceIf(bytesRead > 1, "Read more than one byte from wakeup pipe: " + bytesRead);
```
vijaykota: Nit: Did you mean to put this inside the loop?

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+            /// A sentinel object used in the <see cref="_completed"/> field to indicate that
+            /// the stream completed successfully.
+            /// </summary>
+            private static readonly Exception s_completionSentinel = new Exception("s_completionSentinel");
+
+            /// <summary>A object used to synchronize all access to state on this response stream.</summary>
+            private readonly object _lockObject = new object();
+
+            /// <summary>The response message that owns this response stream.</summary>
+            private readonly CurlResponseMessage _parentResponse;
```
vijaykota: More of a doubt. We are only using the EasyRequest inside this. Fine with this for future extensibility

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+                        // lock, which is why we schedule the operation to run asynchronously: if it ran
+                        // synchronously, it could deadlock due to code on another thread holding the lock
+                        // and calling Dispose on the registration concurrently with the call to Cancel
+                        // the cancellation token.  Dispose on the registration won't return until the action
+                        // associated with the registration has completed, but if that action is currently
+                        // executing and is blocked on the lock that's held while calling Dispose... deadlock.
+                        var crs = new CancelableReadState(buffer, offset, count, this, cancellationToken);
+                        crs._registration = cancellationToken.Register(s1 => Task.Factory.StartNew(s2 =>
+                        {
+                            var crsRef = (CancelableReadState)s2;
```
vijaykota: Nit: Maybe good to have a VerboseTrace here

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+
+                        ClearPendingReadRequest();
+                    }
+                }
+            }
+
+            /// <summary>Clears a pending read request, making sure any cancellation registration is unregistered.</summary>
+            private void ClearPendingReadRequest()
+            {
+                Debug.Assert(_pendingReadRequest != null, "Should only be clearing the pending read request if there is one");
```
vijaykota: Nit: Assert that lock is held

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+            private void VerifyInvariants()
+            {
+                Debug.Assert(Monitor.IsEntered(_lockObject), "Can only verify invariants while holding the lock");
+
+                Debug.Assert(_remainingDataCount >= 0, "Remaining data count should never be negative.");
+                Debug.Assert(_remainingDataCount == 0 || _remainingData != null, "If there's remaining data, there must be a buffer to store it.");
+                Debug.Assert(_remainingData == null || _remainingDataCount <= _remainingData.Length, "The buffer must be large enough for the data length.");
+                Debug.Assert(_remainingData == null || _remainingDataOffset <= _remainingData.Length, "The offset must not walk off the buffer.");
+
+                Debug.Assert(!((_remainingDataCount > 0) & (_pendingReadRequest != null)), "We can't have both remaining data and a pending request.");
```
vijaykota: These should be "&&"

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+                        var crs = new CancelableReadState(buffer, offset, count, this, cancellationToken);
+                        crs._registration = cancellationToken.Register(s1 => Task.Factory.StartNew(s2 =>
+                        {
+                            var crsRef = (CancelableReadState)s2;
+                            lock (crsRef._stream._lockObject)
+                            {
+                                if (crsRef._stream._pendingReadRequest == crsRef)
+                                {
+                                    crsRef.TrySetCanceled(crsRef._token);
+                                    crsRef._stream._pendingReadRequest = null;
```
vijaykota: Isn't it better to call ClearPendingRequest here? I didn't quite understand how crs._registration will get cleaned up otherwise.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+            private Exception _completed;
+
+            /// <summary>
+            /// The state associated with a pending read request. When a reader requests data, it puts
+            /// its state here for the writer to fill in when data is available.
+            /// </summary>
+            private ReadState _pendingReadRequest;
+
+            /// <summary>How much data has been read.</summary>
+            private long _position = 0;
```
vijaykota: This is not needed since it is only being incremented

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
```diff
@@ -0,0 +1,500 @@
...
+                        _remainingDataOffset = 0;
+
+                        // Make sure our remaining data buffer exists and is big enough to hold the data
+                        if (_remainingData == null)
+                        {
+                            _remainingData = new byte[_remainingDataCount];
+                        }
+                        else if (_remainingData.Length < _remainingDataCount)
+                        {
+                            _remainingData = new byte[Math.Max(_remainingData.Length * 2, _remainingDataCount)];
```
vijaykota: Nit: Would be good to verify this optimization by tracing both the news

vijaykota:
LGTM. Thanks for fixing this, @stephentoub

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
stephentoub: Changed to use fields.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
stephentoub: Fixed comment.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs
stephentoub: No, the loop is just there to handle the error case of EINTR, in which case we retry the read.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: Changed to store EasyRequest instead of CurlResponseMessage.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: Added.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: Added assert.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: Changed to be &&.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: It doesn't actually matter in this case.  ClearPendingReadRequest does two things: Disposes of the CTR and nulls out the field.  Since we're in the callback for cancellation being requested, we know that cancellation was requested, and thus Dispose'ing of the CTR will be a nop.  But for consistency and to avoid confusion, I changed it to use ClearPendingReadRequest, as while Dispose'ing the CTR won't help, it also won't hurt.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: I had it hear because it was useful in debugging (note that it was both being incremented and traced out in multiple VerboseTrace calls).  I was also hoping we could return this from the stream's get_Position.  But since we don't do that on Windows, and since it was only for debugging purposes, I removed the field.  We could choose to add it back later, potentialy under a #if CURLHANDLER_VERBOSE.

src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
stephentoub: Added a VerboseTrace call.

stephentoub:
Addressed all feedback.  Thanks for the review.

PR #3110, opened 2015-09-07
Remove problematic ctrl-C System.Console test

stephentoub:
Our test for ctrl-C handling on Unix generally works fine, but it causes problems when run in xunit, which has its own ctrl-C handler that stops running tests. Since the test doesn't make sense for now but we might want something like it later, I'm commenting it out (added a comment to the code as to why it's commented out) rather than marking it as [ActiveIssue] or deleting it.

cc: @ellismg


ellismg:
LGTM.

PR #3112, opened 2015-09-07
Improve performance of BigInteger.GCD

axelheer:
- In order to consolidate the big integer calculation code, Lehmer's algorithm gets some refactoring and a new home within the recently added BigIntegerCalculator.
- The remaining code of BigIntegerBuilder to convert a big integer value to a double value moves to BigInteger, where the other conversion code lives too.
- BigIntegerBuilder is finally dead now, it gets dropped.
- NumericsHelpers contains dead code now, it gets some cleanup.

The performance of *GCD* improves a bit after these changes:

**GCD**

| # of bits | # of vals | before ms | after ms |
|----------:|----------:|----------:|---------:|
|        16 | 1,000,000 |       126 |       81 |
|        64 | 1,000,000 |       369 |      328 |
|       256 |   100,000 |       238 |      236 |
|     1,024 |   100,000 |      1113 |     1077 |
|     4,096 |    10,000 |       675 |      581 |
|    16,384 |     1,000 |       629 |      465 |
|    65,536 |       100 |       816 |      557 |

#1307


axelheer:
cc: @stephentoub

(I'm in no hurry, just want to make sure this doesn't get overlooked...)

stephentoub:
Thanks, @axelheer.  Sorry for the delay.  I'll review soon.

stephentoub:
cc: @KrzysztofCwalina

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
```diff
@@ -321,7 +321,7 @@ private static uint[] CreateCopy(uint[] value)
             Debug.Assert(value != null);
             Debug.Assert(value.Length != 0);
 
-            var bits = new uint[value.Length];
```
KrzysztofCwalina: BTW, we are totally fine with var when it's clear what the resulting type is. We just don't like var x = SomeFunction();

KrzysztofCwalina:
This looks good to me. Thanks again for doing great work to improve this component.

src/System.Runtime.Numerics/src/System/Numerics/BigIntegerCalculator.DivRem.cs
axelheer: Okay, good to know. I just stumbled over that line while debugging this stuff.

axelheer:
@KrzysztofCwalina thanks! :)

stephentoub:
I left a few comments on the change about places it looks like we're missing test coverage.  Other than that, LGTM.

axelheer:
@stephentoub thanks, I'll check those return statements. The break statements are the same as the current implementation, I'm not sure how to find values hitting these edge cases, but I'll think about it.

axelheer:
@stephentoub I found some values to cover all those statements and finally deleted two branches, which don't pay off anyway.

stephentoub:
Thanks!

PR #3115, opened 2015-09-07
Remove redundant checks in RegistryKey

jamesqo:
Summary of changes:

- Checked addition is unnecessary as we'll get an OOME before `blob.Length + 1` overflows
- So we can remove the try/catch clauses
- Memory is already 0-initialized, so we don't need to set it to 0

### DIFF
--- a/src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
+++ b/src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
@@ -1202,9 +1202,7 @@ namespace Microsoft.Win32
                         // make sure the string is null terminated before processing the data
                         if (blob.Length > 0 && blob[blob.Length - 1] != (char)0)
                         {
-                            char[] newBlob = new char[blob.Length + 1];
-                            Array.Copy(blob, 0, newBlob, 0, blob.Length);
-                            blob = newBlob;
+                            Array.Resize(ref blob, blob.Length + 1);
                         }
 
                         var strings = new List<String>();
--- a/src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
+++ b/src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
@@ -1202,18 +1202,9 @@ namespace Microsoft.Win32
                         // make sure the string is null terminated before processing the data
                         if (blob.Length > 0 && blob[blob.Length - 1] != (char)0)
                         {
-                            try
-                            {
-                                char[] newBlob = new char[checked(blob.Length + 1)];
-                                Array.Copy(blob, 0, newBlob, 0, blob.Length);
-                                newBlob[newBlob.Length - 1] = (char)0;
-                                blob = newBlob;
-                            }
-                            catch (OverflowException e)
-                            {
-                                throw new IOException(SR.Arg_RegGetOverflowBug, e);
-                            }
-                            blob[blob.Length - 1] = (char)0;
+                            char[] newBlob = new char[blob.Length + 1];
+                            Array.Copy(blob, 0, newBlob, 0, blob.Length);
+                            blob = newBlob;
                         }
 
                         var strings = new List<String>();


stephentoub:
cc: @pallavit

pallavit:
@jamesqo Thanks for the clean up. LGTM. @tarekgh Could you also take a look :)

src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
```diff
@@ -1202,18 +1202,9 @@ internal Object InternalGetValue(String name, Object defaultValue, bool doNotExp
...
-                                blob = newBlob;
-                            }
-                            catch (OverflowException e)
-                            {
-                                throw new IOException(SR.Arg_RegGetOverflowBug, e);
-                            }
-                            blob[blob.Length - 1] = (char)0;
+                            char[] newBlob = new char[blob.Length + 1];
+                            Array.Copy(blob, 0, newBlob, 0, blob.Length);
+                            blob = newBlob;
```
stephentoub: According to the latest code coverage report, this if block is not exercised by any tests.  Can you add a test that exercises it?  While the change looks safe, it'd be good to have some test coverage to verify that.

Also, could this be simplified further to just be:
```C#
Array.Resize(ref blob, blob.Length + 1);
```
?

tarekgh: I second @stephentoub for adding the test. 
Array.Resize is good idea too but it'll add a couple of checks. I still prefer using Array.Resize as it is explicit and simpler. I don't think the extra checks will make much difference.

tarekgh:
LGTM

jamesqo:
Sorry, recently made a PR to Roslyn and signed a CLA there (despite already having signed it here). Suddenly, it seems that @dnfclas is commenting on all my .NET Core pulls. Sorry for the confusion.

stephentoub:
@jamesqo, no problem.  Are you able to add a test to go along with this? Thanks.

jamesqo:
@stephentoub Sure, I can write one within the next day or so. Sorry for the delay, been a bit busy this week so I haven't had much time for contributing.

jamesqo:
@stephentoub Thanks for being patient. Correct me if I'm wrong, but isn't the block only run when corrupt data is fed into the registry? So the only way to run it would be manually calling `RegQuerySetValueEx` from the tests:

```csharp
char[] corrupt = "str1\0str2\0str3\0".ToCharArray(); // missing \0 at end
Interop.mincore.RegSetValueEx(..., corrupt, ...);
```

And then pull it back out. Is it OK to add `RegQuerySetValueEx.cs` to the tests?

Maxwe11:
>Is it OK to add RegQuerySetValueEx.cs to the tests?

Add interop here
https://github.com/dotnet/corefx/blob/master/src/Microsoft.Win32.Registry/tests/Helpers.cs

jamesqo:
@Maxwe11 Awesome, thanks.

tarekgh:
if you are going  to set any corrupted data to the registry please make sure you put under a path/key that is specific to the test (and not under any path used by other parts of Windows or Windows apps). 
Also when running the test try to look if such corrupted data exist and clean it up to avoid failures or unpredictable results. thanks.

src/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs
tarekgh: just checking, are you going to change this to Array.Resize?

jamesqo: @tarekgh Sorry, I thought I had pushed the commit already. That's strange...

jamesqo:
@tarekgh The thing is, I don't even think it's possible to write the corrupt data. The docs for [RegSetValueEx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724923(v=vs.85).aspx) say that the `char[]` **must** end with two null terminators before being passed in, so it may not be possible to make a corrupt entry without provoking UB.

tarekgh:
@jamesqo doing PInvoke is ok. I was trying to tell this corrupted data has to be stored in registry place that will not affect the OS or any running app. 

Also checking for the existence of the data before you add it is to avoid the PInvoke to fail and to make sure the stored corrupted data is matching the data we are testing on. for any reason this data pre-exist before the test run potentially can make the test fail and it'll be hard to diagnose because you'll need access to the machine run the test. I am just trying to make sure the test is reliable enough to avoid any future problem.

jamesqo:
@tarekgh It's not the corrupt registry entries I'm worried about, but the fact that *attempting to do so could cause undefined behavior.* The docs explicitly state that the string must end in 2 `\0`s, so passing in one that doesn't end in one may either seriously screw things up or return a nonzero code.

tarekgh:
@jamesqo you can have a corrupted data by just running the following native C code

    char charArray1[] = { 'a', 'b', 'c' };
    ret = RegSetValueEx(key, L"CorruptValue1", 0, REG_MULTI_SZ, (BYTE *)charArray1, 3);
    if (ERROR_SUCCESS == ret)
    {
        ret = RegQueryValueEx(key, L"CorruptValue1", 0, &type, (LPBYTE)retValue, &cbData);
        if (ERROR_SUCCESS != ret)
        {
             // we read back 3 bytes
        }
     }

the code will just store some 3 bytes as multi strings value. note that the byte count is odd number and the bytes are not ended with double nulls.

our managed code you are modifying is trying to protect against such cases which is reading odd number of bytes and the data is not null terminated. and we'll return whatever we get from registry even if it seems corrupted.

the question is, do we really need to fail instead of trying to read such data? I think what we are currently doing is correct behavior because the native Windows API doesn't fail either which means it can expect to store some malformed data and still can read it.

tarekgh:
@jamesqo I am going to merge the current change and we can add the test in other PR.

jamesqo:
@tarekgh Awesome, thanks! :smile:

PR #3117, opened 2015-09-08
Use Array.Copy in NumericsHelpers

jamesqo:
A few formatting fixes have been made as well (`System.Math` to `Math`, `resize` to `Resize`).

### DIFF
--- a/src/System.Runtime.Numerics/src/System/Numerics/NumericsHelpers.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/NumericsHelpers.cs
@@ -138,21 +138,18 @@ namespace System.Numerics
             else
             {
                 //??? this is weird
-                d = resize(d, d.Length + 1);
+                d = Resize(d, d.Length + 1);
                 d[d.Length - 1] = 1;
             }
             return d;
         }
 
-        public static uint[] resize(uint[] v, int len)
+        public static uint[] Resize(uint[] v, int len)
         {
             if (v.Length == len) return v;
             uint[] ret = new uint[len];
-            int n = System.Math.Min(v.Length, len);
-            for (int i = 0; i < n; i++)
-            {
-                ret[i] = v[i];
-            }
+            int n = Math.Min(v.Length, len);
+            Array.Copy(v, 0, ret, 0, n);
             return ret;
         }
 


src/System.Runtime.Numerics/src/System/Numerics/NumericsHelpers.cs
```diff
@@ -138,21 +138,18 @@ public static uint[] DangerousMakeTwosComplement(uint[] d)
...
         {
             if (v.Length == len) return v;
             uint[] ret = new uint[len];
-            int n = System.Math.Min(v.Length, len);
-            for (int i = 0; i < n; i++)
-            {
-                ret[i] = v[i];
-            }
+            int n = Math.Min(v.Length, len);
+            Array.Copy(v, ret, n);
```
stephentoub: This should be:
```C#
Array.Copy(v, 0, ret, 0, n);
```

stephentoub:
cc: @axelheer

stephentoub:
I'm a fan of simpler code, but I imagine it's pretty common to invoke this method with a small ```n```.  How have you confirmed this doesn't negatively impact perf?

jamesqo:
https://dotnetfiddle.net/ofsyNm. The `Array.Copy` version seems to be consistently running faster, even for small sizes.

stephentoub:
Thanks, but that's not really a valid microbenchmark from which we can make decisions.  It only executes each test once, doesn't account for start-up costs like JIT, doesn't factor in that these operations are so quick that a single iteration is going to be in the noise compared to the overhead of measuring its execution, doesn't factor in GC overheads for the allocations/collections, etc.

That said, I just tried a few out myself, and it doesn't appear to make a material difference on execution time one way or the other.  So I'm ok with it.

cc: @mellinoe

mikedn:
@jamesqo @stephentoub The benchmark is actually incorrect as both the old and new size are 5. No copying ever occurs in that code. On my machine the threshold seems to be ~30.

stephentoub:
> The benchmark is actually incorrect

Thanks, @mikedn.  I'd fixed that locally, but actually introduced another issue.  Fixing that second issue, I now see similar results to yours (maybe even a slightly higher threshold).  Which again makes me nervous about such a change.

zabulus:
Data from benchmark. Size=5, Resize=6, Count=1000000000
Old: 00:01:50.4395829
New:00:02:33.9997508
CPU: Core i3-1.8Ghz
If it helps

jamesqo:
@mikedn I'd fixed that, had been toying around with making them the same size because for some reason it wasn't affecting the tests. Probably due to the overhead of the Stopwatch itself as was just mentioned.

mikedn:
This is rather unfortunate, using `Array.Copy` should be easy but it has a rather high overhead in the case of small arrays. Still, for larger arrays it has really good performance, I get a 4x improvement for an array of 200 uint. And 200 isn't that large for a BigInteger...

Also, the existing code is dubious to begin with. Looks like Resize is only used in one place and in that place the new size is old size + 1. That make the length check and the Math.Min thing useless.

axelheer:
I might add, that `Array.Copy` is already in use within `BigInteger.cs` (7x), may it be good or evil. And it might really be a good idea to just remove `resize` entirely and update `DangerousMakeTwosComplement` accordingly.

axelheer:
Wait, that's weird (as mentioned in the code comment above): the result of `resize` is a new array, which gets returned by `DangerousMakeTwosComplement`, but the result of this function is not used by an reference. Every reference seems to assume the call of `DangerousMakeTwosComplement` just mutates the already existing array. (?)

I think, I should go to bed now. :zzz:

stephentoub:
@dotnet-bot test this please (https://github.com/dotnet/coreclr/pull/1516)

axelheer:
If we change `DangerousMakeTwosComplement` to something like that, we could get rid of this copy stuff:

    public static void DangerousMakeTwosComplement(uint[] d)
    {
        Debug.Assert(d != null);
        Debug.Assert(d.Length >= 1);

        d[0] = ~d[0] + 1;

        // first do complement and +1 as long as carry is needed
        int i = 1;
        for (; d[i - 1] == 0; i++)
        {
            d[i] = ~d[i] + 1;
        }

        // now ones complement is sufficient
        for (; i < d.Length; i++)
        {
            d[i] = ~d[i];
        }
    }

The only case, which would lead to an overflow, would be an array consistent entirely of zeros. Since `BigInteger` doesn't represent zero as array, that cannot be the case. But I'd check the callers of this function as a precaution (which don't handle the "extra" result anyway, as I've already mentioned).

stephentoub:
@axelheer, if there's a change we can make instead that obviates the need for this, avoids doing unnecessary work, and simplifies the code int he process, I'd of course be in favor of that instead.

axelheer:
I ran the tests using the suggested code: seems to be fine. But someone should double check, if an all-zero array is really not possible. (I don't want to "rob" this PR from you, @jamesqo, since you've found this "issue"...)

stephentoub:
> since you've found this "issue"...)

There's a lot to be said for minimizing churn and not changing something unless there's a good reason to do so.

@axelheer, does your proposed version make a measurable improvement on the consuming library?  I worry if we're not 100% sure it's safe.

@jamesqo, does your change make a measurable improvement on the consuming library (I mean code using BigInteger, not a microbenchmark just around Resize)?  Or in the context in which it's used, does it at least not harm anything?  I like that it simplifies the code, and I could believe that it would help with larger BigIntegers as @mikedn pointed out.  If we could demonstrate that it doesn't measurably harm smaller BigIntegers, I'm fine with the change as-is, and then we could subsequently look at eliminating this altogether with a change from @axelheer.

stephentoub:
@dotnet-bot test this please

stephentoub:
So I finally found some time this morning to look at this in a bit more depth.  As alluded to by @axelheer, this resize functionality is never hit, at least not by any of our tests (which are fairly extensive), and appears to be dead code; there's no perf impact either way because it's never run.  I don't want us to muck around in such code, and would prefer to see us either leave this as-is or use a separate PR to do the right fix of cleaning up these code paths.  So I'm going to close out this PR.  Regardless, thanks, @jamesqo.

PR #3118, opened 2015-09-08
Two perf tweaks to UnixFileSystem's enumeration iterator

stephentoub:
1. When enumerating directories recursively, we were computing the Path.Combine for the user path twice for each directory: once as part of pushing the string onto the exploration stack, and once as part of yielding it.  This fixes it to only ever compute it once per directory.

2. Due to https://github.com/dotnet/roslyn/issues/5062, a field was being added to the iterator object for isDir.  I changed the switch to an if/elseif/else to shrink the size of the object.  I also rearranged the order in which we push the directory onto the evaluation stack and yield the string in order to allow userPath not to be lifted, either.

cc: @ianhays, @justinvp


justinvp:
LGTM.

ianhays:
LGTM

stephentoub:
Thanks, guys

PR #3119, opened 2015-09-08
Increase System.Linq line code coverage to 99.5%

stephentoub:
This commit increases LINQ line code coverage as much as is possible given the current code.

New tests:
- There's an existing legacy test that exercises the non-generic IEnumerable.GetEnumerator for IGrouping`3.  I've added the same test but for IGrouping`4.
- IGrouping inherits IList/ICollection, and a bunch of those interface members weren't being exercised.  Added tests for all of them.
- Last and LastOrDefault with predicates weren't being tested with non-IList sources.  Added tests now that do.
- ThenByDescending didn't have a test for a custom comparer.  Added one.
- System.Linq.dll contains some debug view types used by Visual Studio when showing enumerables in the watch window.  I've added tests for those types.

The only lines remaining uncovered are:
- A function from EnumerableHelpers that's not used by this assembly
- An if branch in EnumerableHelpers.ToArray for when trying to allocate an array larger than the maximum array size
- Code in ```Set<T>``` that depends on Add being used after Remove, which never happens in LINQ's code currently

cc: @VSadov, @JonHanna

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -3188,8 +3188,6 @@ namespace System.Linq
         private int _freeList;
         private IEqualityComparer<TElement> _comparer;
 
-        public Set() : this(null) { }
-
         public Set(IEqualityComparer<TElement> comparer)
         {
             if (comparer == null) comparer = EqualityComparer<TElement>.Default;
@@ -3203,12 +3201,6 @@ namespace System.Linq
         public bool Add(TElement value)
         {
             return !Find(value, true);
-        }
-
-        // Check whether value is in set
-        public bool Contains(TElement value)
-        {
-            return Find(value, false);
         }
 
         // If value is in set, remove it and return true; otherwise return false
@@ -3567,6 +3559,11 @@ namespace System.Linq
         }
     }
 
+    // NOTE: DO NOT DELETE THE FOLLOWING DEBUG VIEW TYPES.
+    // Although it might be tempting due to them not be referenced anywhere in this library,
+    // Visual Studio currently depends on their existence to enable the "Results" view in 
+    // watch windows.
+
     /// <summary>
     /// This class provides the items view for the Enumerable
     /// </summary>
@@ -3624,7 +3621,7 @@ namespace System.Linq
         {
             get
             {
-                return Strings.EmptyEnumerable;
+                return SR.EmptyEnumerable;
             }
         }
     }
--- a/src/System.Linq/src/Resources/Strings.resx
+++ b/src/System.Linq/src/Resources/Strings.resx
@@ -120,9 +120,6 @@
   <data name="EmptyEnumerable" xml:space="preserve">
     <value>Enumeration yielded no results</value>
   </data>
-  <data name="InvalidOperation_EnumOpCantHappen" xml:space="preserve">
-    <value>Enumeration has either not started or has already finished.</value>
-  </data>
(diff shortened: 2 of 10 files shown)


JonHanna:
I think the large array case gets hit by an outerloop test. I remember getting bored waiting on that branch when running the tests in the Visual Studio debugger, which doesn't ignore outerloop.

stephentoub:
> I think the large array case gets hit by an outerloop test.

That's possible.  We only do our code coverage runs against inner loop tests.

> Visual Studio debugger, which doesn't ignore outerloop

You can change that if you like.  Go into the project properties for the test project and add "-notrait category=outerloop" to the command-line arguments in the debug settings.

JonHanna:
I ended up doing just that. What I haven't figured out is if I can get it to visualise code coverage in a way akin to SharpDevelop with NUnit.

JonHanna:
An inner loop test on that large array case could be in danger of triggering out of memory when it should pass.

stephentoub:
> An inner loop test on that large array case could be in danger of triggering out of memory when it should pass.

Which is why I didn't add anything for it.

JonHanna:
And probably why I didn't get stuck on the outer one until I got "lucky".

stephentoub:
> What I haven't figured out is if I can get it to visualise code coverage in a way akin to SharpDevelop with NUnit.

You mean in Visual Studio?  We haven't connected all of the dots yet to make that happen.  But you can view the HTML report generated when ```/p:Coverage=true```, and it provides similar visual information (albeit without the interactivity you'd get in VS).

JonHanna:
Now that I'm not on a phone, I can confirm that `ToArray_FailOnExtremelyLargeCollection` does indeed hit that particular branch.

VSadov:
LGTM

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -3567,6 +3559,11 @@ internal TElement[] ToArray()
         }
     }
 
+    // NOTE: DO NOT DELETE THE FOLLOWING DEBUG VIEW TYPES.
+    // Although it might be tempting due to them not be referenced anywhere in this library,
+    // Visual Studio currently depends on their existence to enable the "Results" view in 
+    // watch windows.
+
```
JonHanna: Is the lack of disposal of enumerators in `Items` for a reason? It looks wrong, but it also looks deliberate, like the sort of wrong-seeming thing that is actually right.

stephentoub: Not sure.  I don't see a good reason not to do it.  @VSadov?

VSadov: I do not see why it simply does not do _enumerable.ToArray() 
Also _cachedCollection seems to be there to cache the values while it is not used. 
The ```_cachedCollection = null;``` in SystemCore_EnumerableDebugView ctor seems bogus.

The code clearly could benefit from some clean up. My guess - the scary comment makes people look the other way.

JonHanna: It was all of the rest of it that scared me :smile:

VSadov: I searched through VS/EE sources and nothing seem to be accessing _cachedCollection, or _count.
It seems they could as well be locals.

It kind of looks like _cachedCollection was intended to capture the results, but since it does not, I'd be careful to introduce caching now. If there is any caching in debug visualizers, it must be happening externally and adding caching here might be confusing.

PR #3121, opened 2015-09-08
Delete EnumerableUtils from System.IO.FileSystem.Tests

stephentoub:
Dead code. It also contained the only types in the test assembly that were in a namespace other than System.IO.FileSystem.Tests (there are still types that have no namespace).

cc: @ianhays


ianhays:
LGTM!

stephentoub:
Thanks.

PR #3129, opened 2015-09-08
Fix build.sh build

ellismg:
A while back an update to build tools broke building via build.sh on
Linux and OSX. Move to a newer version of build tools which addresses
this issue and fix up a couple of build breaks that have made their way
in since then.

### DIFF
--- a/src/System.Security.Cryptography.Cng/tests/System.Security.Cryptography.Cng.Tests.csproj
+++ b/src/System.Security.Cryptography.Cng/tests/System.Security.Cryptography.Cng.Tests.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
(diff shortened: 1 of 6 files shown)


src/System.Security.Cryptography.Cng/tests/System.Security.Cryptography.Cng.Tests.csproj
```diff
@@ -30,7 +30,7 @@
     <Compile Include="OpenTests.cs" />
     <Compile Include="ImportExportTests.cs" />
     <Compile Include="PropertyTests.cs" />
-    <Compile Include="RSACngTests.cs" />
```
mellinoe: Are these changes accidental?

ellismg: No.  This is the casing of the file on disc.  Without these changes, MSBuild complains about being unable to find files.

mellinoe: Oh okay, gotcha. So these files were added sometime after the build was originally broken, I guess.

mellinoe:
LGTM

src/System.Security.Cryptography.Cng/tests/System.Security.Cryptography.Cng.Tests.csproj
ellismg: It's also possible that we just never noticed, since we don't run build.sh in CI yet.

nguerrera:
LGTM. FWIW, the casing in the .csproj is actually the casing I would expect on disk for those crypto files (matches the casing of the API).

PR #3130, opened 2015-09-09
Avoid calling Array.Clear on the entire _array when Queue is already empty

DmitryKakurin:
This was a perf bug. Calling Queue<T>.Clear repeatedly on an empty Queue was always Array.Clear'ing the entire _array which caused significant perf impact when capacity is big.

### DIFF
--- a/src/System.Collections/src/System/Collections/Generic/Queue.cs
+++ b/src/System.Collections/src/System/Collections/Generic/Queue.cs
@@ -101,17 +101,21 @@ namespace System.Collections.Generic
         /// <include file='doc\Queue.uex' path='docs/doc[@for="Queue.Clear"]/*' />
         public void Clear()
         {
-            if (_head < _tail)
-                Array.Clear(_array, _head, _size);
-            else
+            if (_size != 0)
             {
-                Array.Clear(_array, _head, _array.Length - _head);
-                Array.Clear(_array, 0, _tail);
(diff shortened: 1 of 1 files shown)


stephentoub:
Thanks, @DmitryKakurin.  Can you share more details about the perf before and after the change?

DmitryKakurin:
I've used MeasureIt to compare the perf when I have found the issue.
Here are the MeasureIt function and results produced: https://gist.github.com/DmitryKakurin/496ba1c068c216158e55

As expected, with higher queue capacity "Empty Queue.Clear" incurs more and more overhead. And real difference is even slightly higher as "Single element Queue.Clear" includes cost of enqueuing a single element (probably negligible).

stephentoub:
Thanks, @DmitryKakurin.  I experimented with this as well.  There's certainly a very measurable improvement from this change for the Count == 0 case, and that improvement grows the larger the capacity of the queue, so I was mainly interested if there were any measurable negative side effects on the Count > 0 case, where we're adding a check and branch that didn't previously exist.  But the other costs involved appear to dwarf any such additional costs, such that I couldn't measure a real difference in those cases.  Given that, the change LGTM.  Thanks!

DmitryKakurin:
Thanks! I've just realized that non-generic Queue has the same issue. I'll submit another fix for it.

PR #3132, opened 2015-09-09
System.Linq.Parallel - more tests

Clockwork-Muse:
Added more tests and refined some existing ones.  There's a few things I'm still experimenting with, or waiting to hear the status of, but I'll issue a different PR for that stuff.
Including the test namespace change, per #2898 .

@stephentoub

### DIFF
--- a/src/System.Linq.Parallel/tests/Helpers/Comparers.cs
+++ b/src/System.Linq.Parallel/tests/Helpers/Comparers.cs
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 
-namespace Test
+namespace System.Linq.Parallel.Tests
 {
     internal class FailingEqualityComparer<T> : IEqualityComparer<T>
     {
--- a/src/System.Linq.Parallel/tests/Helpers/Comparers.cs
+++ b/src/System.Linq.Parallel/tests/Helpers/Comparers.cs
@@ -59,7 +59,7 @@ namespace System.Linq.Parallel.Tests
 
         public int Compare(int x, int y)
         {
-            return -x.CompareTo(y);
+            return y.CompareTo(x);
         }
     }
 
@@ -119,6 +119,24 @@ namespace System.Linq.Parallel.Tests
         public int CompareTo(DelegatedComparable<T> other)
         {
             return _comparer.Compare(Value, other.Value);
+        }
+    }
+
+    internal struct NotComparable
+    {
+        private readonly int _value;
+
+        public int Value
+        {
+            get
+            {
+                return _value;
+            }
+        }
+
+        public NotComparable(int x)
+        {
+            _value = x;
         }
     }
 }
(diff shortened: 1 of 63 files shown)


src/System.Linq.Parallel/tests/Helpers/Comparers.cs
```diff
@@ -122,4 +121,15 @@ public int CompareTo(DelegatedComparable<T> other)
             return _comparer.Compare(Value, other.Value);
         }
     }
+
+    internal struct NotComparable
+    {
+
+        public int Value { get; }
```
stephentoub: get-only auto props were added in C# 6.  We're not building corefx yet using C# 6, so this fails to compile.

Clockwork-Muse: silly me.

stephentoub:
Thanks, @Clockwork-Muse.

I left a few comments on some of the commits (e.g. please remove custom use of [Trait]), but overall it looks fine.  It'd be good to squash this down a bit as well, e.g. combining the commits that deal with multiple settings, combining the commits that simply extract things to a separate file, etc.  

It'd also be helpful to understand what you believe the remaining gaps are, and try to limit future churn to that.  The tests appear to be fairly complete at this time, and if anything it seems like we may be testing more than we need to (e.g. verifying that every combination of parallel merge options fail when used together).

Clockwork-Muse:
@stephentoub - Updated.  Was that more what you were looking for?

I agree the tests are mostly complete.  There's a few areas where, if issues are accepted as bugs, I need to add tests to cover.  Most of these are cases where there's ambiguity over expected behavior relating to ordering and thrown exceptions, or related to such cases.

For instance, I have an expansion of tests relating to cancellation, but have an issue relating to how exceptions are handled when passing a pre-canceled token.  Like so:
```
// Throws OCE:
throwOnFirstMoveNextSource.AsParallel().WithCancellation(precanceledToken).ToArray();
// Throws AggregateException:
throwOnFirstMoveNextSource.AsParallel().WithCancellation(precanceledToken).Zip(otherEnumerable, (y, y) => x).ToArray();
```
(the latter doesn't have to be `Zip` - all binary operators seem to do this, but not unary...)

Clockwork-Muse:
Hmm... failures aren't me, something in System.Process.Tests (and for other people too).

Clockwork-Muse:
@stephentoub - hmm, it looks like `OuterLoop` has been updated in build tools, but it's not showing up in metadata in target projects.  I'm guessing the relevant project somehow didn't get built/published?

stephentoub:
@dotnet-bot test this please

stephentoub:
@Clockwork-Muse, I just tried replacing the three [Trait]s with [OuterLoop] on the classes, and it both compiled fine and treated them as outer loop.  Could you try again?  Maybe rebase to pick up the latest changes?  It's possible I've picked up newer build tools than you had.

Clockwork-Muse:
@stephentoub - it ended up being the `project.lock` file.  Rebasing to a version since this morning fixed that.
Now I just have to figure out why debug/corerun is complaining that an xunit runner is missing...

Clockwork-Muse:
@stephentoub - and rebased for the corrected buildtools package.

stephentoub:
Thanks for the updates.  LGTM.

PR #3139, opened 2015-09-09
Hmackey

AtsushiKan:
This is split into two commits for easier reviewing. The first commit renames an internal field which had a slightly unspecific/misleading name - renaming because the next commit adds a new field which elevates this from being "annoying" to "dangerous."

The second commit contains the meat of the change and adds code to replace incoming lengthy keys with a hash as specified by Rfc2014 section 2.

### DIFF
--- a/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HMACCommon.cs
+++ b/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HMACCommon.cs
@@ -19,9 +19,10 @@ namespace Internal.Cryptography
     //
     internal sealed class HMACCommon
     {
-        public HMACCommon(String hashAlgorithmId, byte[] key)
+        public HMACCommon(String hashAlgorithmId, byte[] key, int BlockSize)
         {
             _hashAlgorithmId = hashAlgorithmId;
+            _blockSize = BlockSize;
             ChangeKey(key);
         }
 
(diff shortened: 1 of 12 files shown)


src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HMACCommon.cs
```diff
@@ -29,42 +30,60 @@ public int HashSizeInBits
...
+            {
+                // Perform RFC 2104, section 2 key adjustment.
+                if (_lazyHashProvider == null)
+                    _lazyHashProvider = HashProviderDispenser.CreateHashProvider(_hashAlgorithmId);
+                _lazyHashProvider.AppendHashData(key, 0, key.Length);
+                key = _lazyHashProvider.FinalizeHashAndReset();
+            }
+
+            HashProvider oldHashProvider = _hMacProvider;
+            _hMacProvider = null;
```
stephentoub: Nit: why set this to null?  We set it to something else in the statement after next.

stephentoub:
LGTM

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HMACCommon.cs
AtsushiKan: While the null-assignment wasn’t actually introduced by this change, my mind just groups it as part of a “paranoid dispose pattern” that’s textually followed by a new “initialization” pattern. That’s probably how it got there in the first place.

AtsushiKan:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please (infrastructure failure)

PR #3142, opened 2015-09-09
Moving more Unix\libc function calls over to the PAL layer

sokket:
Moving more Unix\libc function calls over to the PAL layer along with fixing up the Interop calls for the switched functions.

/cc: @nguerrera, @piotrpMSFT


src/Native/System.Native/pal_time.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include "../config.h"
+#include "pal_time.h"
+
+#include <assert.h>
+#include <utime.h>
+
+
+static
+void ConvertPalUTimBufToNative(UTimBuf& pal, utimbuf& native)
```
Maxwe11: nit: `const UTimBuf& pal`

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
 {
     return mkstemps(pathTemplate, suffixLength);
 }
+
+static
+void ConvertPalPollFdToNative(PollFd& pal, pollfd& native)
```
Maxwe11: Ditto. Assuming `pal` not modified it should be passed as `const` reference. I'll stop commenting on this.

src/Common/src/Interop/Unix/System.Native/Interop.Poll.cs
```diff
@@ -6,38 +6,36 @@
...
+            POLLIN   = 0x0001,  /* any readable data available */
+            POLLOUT  = 0x0004,  /* data can be written without blocked */
+            POLLERR  = 0x0008,  /* an error occurred */
+            POLLHUP  = 0x0010,  /* the file descriptor hung up */
+            POLLNVAL = 0x0020,  /* the requested events were invalid */
         }
 
-        [StructLayout(LayoutKind.Sequential)]
-        internal struct pollfd
+        internal struct PollFd
```
nguerrera: Nit: PollFD because FD is an acronym not an abbreviation. (Most confusing guideline of all time, but that's what we have in Framework Design Guidelines.)

src/Common/src/Interop/Unix/System.Native/Interop.Posix_FAdvise.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal enum AdviseAdvice : int
```
nguerrera: I'd call this FileAdvice or PosixFile even PosixFileAdvice. My PR has a MemoryAdvice to go with MAdvise.

src/Common/src/Interop/Unix/System.Native/Interop.Posix_FAdvise.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="fd">The file descriptor of the file</param>
+        /// <param name="offset">The start of the region to advise about</param>
+        /// <param name="length">The number of bytes of the region (until the end of the file if 0)</param>
+        /// <param name="advice">The type of advice to give the kernel about the specified region</param>
+        /// <returns>
+        /// Returns 0 on success; otherwise, the error code is returned
+        /// </returns>
+        [DllImport(Libraries.SystemNative, SetLastError = false /* this is explicitly called out in the man page */)]
+        internal static extern int Posix_FAdvise(int fd, long offset, long length, AdviseAdvice advice);
```
nguerrera: I'd drop the _. I didn't keep it for shm_open.

src/Common/src/Interop/Unix/System.Native/Interop.SnPrintF.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// PInvokes declared for each one we need. The PInvoke layer will take care of putting the
+        /// arguments in correctly.
+        /// </remarks>
+        /// <returns>
+        /// Returns the number of characters (excluding null terminator) written to the buffer on 
+        /// success; if the return value is equal to the size then the result may have been truncated. 
+        /// On failure, returns a negative value.
+        /// </returns>
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal static unsafe extern int SnPrintF(byte* str, long size, string format, string arg1);
```
nguerrera: nit: SNPrintF by same logic as PollFD.

src/Common/src/Interop/Unix/System.Native/Interop.SysLog.cs
```diff
@@ -0,0 +1,57 @@
...
+            LOG_AUTH        = (4<<3),   /* authorization messages */
+            LOG_SYSLOG      = (5<<3),   /* messages generated internally by syslogd */
+            LOG_LPR         = (6<<3),   /* line printer subsystem */
+            LOG_NEWS        = (7<<3),   /* network news subsystem */
+            LOG_UUCP        = (8<<3),   /* UUCP subsystem */
+            LOG_CRON        = (9<<3),   /* clock daemon */
+            LOG_AUTHPRIV    = (10<<3),  /* authorization messages (private) */
+            LOG_FTP         = (11<<3),  /* ftp daemon */
+            // Between FTP and Local is reserved for system use
+            LOG_LOCAL0      = (16<<3),  /* reserved for local use */
```
nguerrera: Do we need to bother declaring the reserved ones?

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
...
+    pal.FileDescriptor = native.fd;
+    pal.EventsToPollFor = native.events;
+    pal.ReceivedEvents = native.revents;
+}
+
+extern "C"
+int32_t Poll(PollFd* pollData, uint32_t numberOfPollFds, int32_t timeout)
+{
+    // Convert our standardized pollfd to the native one
+    pollfd* fds = new pollfd[numberOfPollFds];
```
nguerrera: If I greped correctly, we only call this with up to 2 fds. Seems like a waste to heap alloc/delete. Could we just put a stackalloc limit and fail otherwise?

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
...
+extern "C"
+int32_t Posix_FAdvise(int32_t fd, int64_t offset, int64_t length, AdviseAdvice advice)
+{
+#if HAVE_POSIX_ADVISE
+    return posix_fadvise(fd, offset, length, (int)advice);
+#else
+    // Not supported on this platform; however, we don't want to error out since
+    // currently, this isn't used on platforms where it isn't supported.
+    return ENOSYS;
+#endif
```
nguerrera: What is meant by "we don't want to error out"? Isn't that exactly what we're doing here?

src/Native/System.Native/pal_io.h
```diff
@@ -404,7 +448,74 @@ int32_t Link(const char* source, const char* linkTarget);
...
+int64_t Read(int32_t fd, void* buffer, uint64_t count);
+
+/**
+ * Takes a path to a symbolic link and attempts to place the link target path into the buffer. If the buffer is too
+ * small, the path will be truncated. No matter what, the buffer will not be null terminated. 
+ *
+ * Returns the number of bytes placed into the buffer on success; otherwise, -1 is returned and errno is set.
+ */
+extern "C"
+int64_t ReadLink(const char* path, char* buffer, int64_t bufferSize);
```
nguerrera: bufferSize should be unsigned as it is in underlying API (size_t).

src/Native/System.Native/pal_io.h
```diff
@@ -404,7 +448,74 @@ int32_t Link(const char* source, const char* linkTarget);
...
+extern "C"
+void Sync();
+
+/**
+ * Writes the specified buffer to the provided open file descriptor
+ *
+ * Returns the number of bytes written on success; otherwise, returns -1 and sets errno
+ */
+extern "C"
+int64_t Write(int32_t fd, void* buffer, int64_t bufferSize);
```
nguerrera: buffer should be const void*
count should be uint64_t

src/Native/System.Native/pal_time.cpp
sokket: Good catch!

src/Native/System.Native/pal_io.cpp
sokket: Will fix this one as well :)

src/Common/src/Interop/Unix/System.Native/Interop.Poll.cs
sokket: Ah gotcha, thanks for pointing this out!

src/Native/System.Native/pal_process.cpp
```diff
@@ -3,16 +3,55 @@
...
-static_assert((int)Signals::PAL_SIGKILL == SIGKILL, "");
+static_assert(static_cast<int>(Signals::PAL_None) == 0, "");
+static_assert(static_cast<int>(Signals::PAL_SIGKILL) == SIGKILL, "");
+
+// Validate that our WaitPidOptions enum values are correct for the platform
+static_assert(static_cast<int>(WaitPidOptions::PAL_None)        == 0, "");
+static_assert(static_cast<int>(WaitPidOptions::PAL_WNOHANG)     == WNOHANG, "");
+static_assert(static_cast<int>(WaitPidOptions::PAL_WUNTRACED)   == WUNTRACED, "");
+
+// Validate that our SysLogPriority values are correct for the platform
```
nguerrera: Can you add comments about whether we're just lucky so far or the numeric values are specified by POSIX? I've been making an effort to do that with these blocks.

src/Common/src/Interop/Unix/System.Native/Interop.Posix_FAdvise.cs
sokket: FileAdvice is definitely more readable

src/Native/System.Native/pal_utilities.cpp
```diff
@@ -16,3 +18,13 @@ void SafeStringCopy(char* destination, int32_t destinationSize, char* source)
     snprintf(destination, destinationSize, "%s", source);
 #endif
 }
+
+extern "C"
+int32_t SnPrintF(char* string, int64_t size, char* format, ...)
```
nguerrera: size should be unsigned as it's size_t underneath. Please check all of the int64_t's off_t is signed, but size_t is not.

We should add -Wconversion to our complilation. Apparently, it isn't included in -Wall. :( I'll do it in a separate change.

src/Common/src/Interop/Unix/System.Native/Interop.Posix_FAdvise.cs
sokket: Cool, I'll rename the function and the file

src/Common/src/Interop/Unix/System.Native/Interop.SnPrintF.cs
sokket: Fixed the function name and file name

src/Common/src/Interop/Unix/System.Native/Interop.SysLog.cs
sokket: Nope, the comment is supposed to say why we skipped (12<<3) - (15<<3). If the comment isn't clear I can change it :)

src/Native/System.Native/pal_io.cpp
sokket: I'd prefer to keep it this way; making it stackalloc'd with a limit seems really brittle and any future callers will need to be aware of that fact. By keeping it like the underlying function, all future callers are protected and can call it with as many pollfd's as they would like without having to go look / potentially change the PAL layer going forward. If you don't think we'll have many more callers in the future and are okay with that tradeoff, I can go ahead and make it stackalloc'd instead :)

src/Native/System.Native/pal_io.cpp
sokket: Sorry for not being clear, I forgot the # before error (making it #error). Since we are looking for certain functionality, I wanted a comment to say that we explicitly aren't causing a compilation error here on purpose when this doesn't exist. I'll add the missing # on error to make that clear

src/Native/System.Native/pal_io.h
sokket: Ah good catch; here and on Write

src/Native/System.Native/pal_io.h
sokket: Fixed

src/Common/src/Interop/Unix/System.Native/Interop.SysLog.cs
nguerrera: Oh, I read that wrong the first time seeing reserved in the description of LOCAL_LOCAL0 and thinking it was part of what the comment above was saying about a range being reserved. So my question was should we remove LOG_LOCALN, but nevermind.

src/Native/System.Native/pal_io.cpp
nguerrera: cc @stephentoub for his take.

stephentoub: I'm not currently aware of any additional planned uses of poll (sockets is using epoll), and if additional needs do come in, I expect we'd either have a small number of fds (1 or 2) or we'd try to use epoll.  I think it'd make sense to avoid the allocation and have a small cap we trip over if we do need more dynamism, at which point we could add an implementation like Jon currently has.  My $.02.

src/Common/src/Interop/Unix/System.Native/Interop.WaitPid.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal enum WaitPidOptions : int
```
nguerrera: Should this be [Flags]?

src/Native/System.Native/pal_io.cpp
sokket: Cool, if we can get a perf improvement from going stackalloc instead and we don't think that will be a limitation, I'll convert to stackalloc and put an assert to make sure we don't pass in more than our stackalloc'd size.

nguerrera:
A general comment on raw integers vs. enum's on C++ side for flags. I've been doing the latter based on my recollection that storing a value that is not any one of the enum members is unspecified or undefined in standard C++. However, I'm no longer sure after looking at recent edits on cppreference.com:

http://en.cppreference.com/mwiki/index.php?title=cpp/language/static_cast&diff=78871&oldid=74481

My understanding was on the LHS of that diff. I don't know yet if the edits reflect errata on the site matching my own misunderstanding or just updates tracking the next version of the standard.

Even if it is well defined, it requires a lot of ugly casting, which kind of eliminates the benefit of strong typing. It's not apparent in this change because your flags come in from C# with the right values, but once you have to twiddle the bits, e.g. ConvertOpenFlags, it gets messy. 

Assuming it is in fact well defined, we could have a macro to define bitwise operators on what we deem Flags enums. This would serve double-duty as documentation to C# caller to add flags. I'm fine leaving this for a later change, but this is something we should standardize one way or the other.

src/Native/System.Native/pal_utilities.cpp
sokket: Good catch, fixed all the size_t's to be unsigned. Thanks for catching this!

src/Common/src/Interop/Unix/System.Native/Interop.WaitPid.cs
sokket: Ah true it should

src/Native/System.Native/pal_process.cpp
sokket: Sure, after speaking offline it seems like more values than not will be called out by name and not value in the POSIX standard (anecdotally) so I added a blanket statement in each file calling this out.

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
...
+{
+    pal.FileDescriptor = native.fd;
+    pal.EventsToPollFor = native.events;
+    pal.ReceivedEvents = native.revents;
+}
+
+extern "C"
+int32_t Poll(PollFD* pollData, uint32_t numberOfPollFds, int32_t timeout)
+{
+    assert(numberOfPollFds <= 2);
```
nguerrera: Make this a real check, otherwise it will buffer overflow in release if somebody passes 3.

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
...
+
+extern "C"
+int32_t PosixFAdvise(int32_t fd, int64_t offset, int64_t length, FileAdvice advice)
+{
+#if HAVE_POSIX_ADVISE
+    return posix_fadvise(fd, offset, length, (int)advice);
+#else
+    // Not supported on this platform; however, we don't want to #error here since
+    // currently, this isn't used on platforms where it isn't supported.
+    return ENOSYS;
```
nguerrera: I've been using ENOTSUP for this. Would be good to stay consistent. Is ENOSYS better?

src/Native/System.Native/pal_io.h
```diff
@@ -22,6 +22,16 @@ struct FileStatus
...
 
+
+/************
+ * The enums below are, unless otherwise specified, common values on the platforms
+ * we currently support. These values are not guaranteed by any standard, unless
+ * called out specifically, and could differ on other platforms. When adding new
+ * enums and values, be sure to add a static_assert in the implementation file
+ * to ensure your constant is as expected on all platforms.
+ */
+
```
nguerrera: This isn't quite what I'd want the header to say. The values below in the header are fixed and correct for managed callers to use forever. We must never change them. The implementation must either: static_assert that they are equal to the native equivalent OR convert them appropriately.

src/Native/System.Native/pal_io.cpp
```diff
@@ -443,3 +464,93 @@ int32_t MksTemps(char* pathTemplate, int32_t suffixLength)
 {
     return mkstemps(pathTemplate, suffixLength);
 }
+
+static
+void ConvertPalPollFDToNative(const PollFD& pal, pollfd& native)
+{
+    native.fd = pal.FileDescriptor;
```
nguerrera: For stat, we kept the names of the underlying struct members and just changed the casing e.g. FD, Events, REvents here. I think we should keep doing that to make it easier to correlate managed declarations with man pages.

Ditto for UTimBuf -> AcTime, ModTime.

Also, the convention for Convert functions so far has been (only one direction needed -> ConvertXxx, two directions -> ConvertXxxPalToPlatform, ConvertXxxPlatformToPal.

src/Native/System.Native/pal_process.h
```diff
@@ -107,3 +169,22 @@ int32_t GetPid();
...
+ * Waits for child process(s) or gathers resource utilization information about child processes
+ *
+ * The return value from WaitPid can very greatly. 
+ * 1) returns the process id of a terminating or stopped child process
+ * 2) if no children are waiting, -1 is returned and errno is set to ECHILD
+ * 3) if WNOHANG is specified and there are no stopped or exited children, 0 is returned
+ * 4) on error, -1 is returned and errno is set
+ */
+extern "C"
+int32_t WaitPid(int32_t pid, int* status, WaitPidOptions options);
```
nguerrera: int32_t* status

src/Native/System.Native/pal_utilities.h
```diff
@@ -10,3 +10,14 @@
...
+/**
+ * snprintf is difficult to represent in C# due to the argument list, so the C# PInvoke
+ * layer will have multiple overloads pointing to this function 
+ *
+ * Returns the number of characters (excluding null terminator) written to the buffer on 
+ * success; if the return value is equal to the size then the result may have been truncated. 
+ * On failure, returns a negative value.
+ */
+extern "C"
+int32_t SNPrintF(char* string, uint64_t size, char* format, ...);
```
nguerrera: const char* format

src/Native/System.Native/pal_utilities.h
```diff
@@ -10,3 +10,14 @@
  * or a different safe copy method, depending on the current platform.
  */
 void SafeStringCopy(char* destination, int32_t destinationSize, char* source);
```
nguerrera: const char* source

src/Common/src/Interop/Unix/System.Native/Interop.Poll.cs
```diff
@@ -50,14 +48,17 @@ internal struct pollfd
...
-        internal unsafe static int poll(int fd, PollFlags flags, int timeout, out PollFlags resultFlags)
+        internal static int Poll(int fd, PollFlags flags, int timeout, out PollFlags resultFlags)
         {
-            pollfd pfd = default(pollfd);
-            pfd.fd = fd;
-            pfd.events = flags;
-            int result = poll(&pfd, 1, timeout);
-            resultFlags = pfd.revents;
-            return result;
+            unsafe
```
stephentoub: If we're going to move the ```unsafe``` out of the signature, we should probably make it as tight as possible around what actually needs to be unsafe, which is just the call to Poll, right?

src/Native/System.Native/pal_io.cpp
sokket: Doesn't matter to me, I picked the first one that made sense to me; I'll change it to use the one you've established that we should use

src/Common/src/Interop/Unix/System.Native/Interop.WaitPid.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// The return value from WaitPid can very greatly. 
+        /// 1) returns the process id of a terminating or stopped child process
+        /// 2) if no children are waiting, -1 is returned and errno is set to ECHILD
+        /// 3) if WNOHANG is specified and there are no stopped or exited children, 0 is returned
+        /// 4) on error, -1 is returned
+        /// </returns>
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal static extern int WaitPid(int pid, out int status, WaitPidOptions options);
+
+        internal static int WEXITSTATUS(int status)
```
stephentoub: I'm a bit concerned about these functions, which I'd basically pulled from .h macros, and I think it likely they're not going to be consistent across platforms.  Can you investigate to either a) confirm that these will always work as written based on POSIX documentation, or b) turn these macros into functions exposed from the shim?

src/Native/System.Native/pal_io.cpp
stephentoub: > I think we should keep doing that

Yeah, I agree.  I see the value in the more descriptive names, but that's a slippery slope, and there's definitely value in having them match as closely as possible to what's in the native functions.

src/Native/System.Native/pal_io.cpp
sokket: Added a check and the assert

src/Native/System.Native/pal_io.h
sokket: Fixed

stephentoub:
Thanks, @sokket.  Left a few more comments, but overall LGTM.  The only really important comment I left is the one about those existing macros which may need to be moved down into the shim as exported functions.

src/Native/System.Native/pal_io.cpp
sokket: Cool, fixed :)

src/Native/System.Native/pal_process.h
sokket: Good catch

src/Native/System.Native/pal_utilities.h
sokket: Fixed

src/Native/System.Native/pal_utilities.h
sokket: Fixed

src/Common/src/Interop/Unix/System.Native/Interop.Poll.cs
sokket: I'll just put it back in the signature :)

src/Common/src/Interop/Unix/System.Native/Interop.WaitPid.cs
sokket: I'll turn them into functions that are exported to be sure that each platform behaves as it should

nguerrera:
LGTM. Thanks! 

You'll have to merge with my recent commits. Sorry. :wink:

nguerrera:
Oh, and I moved pal_utilities to ../Common assuming we'd only put implementation things to share between shim libraries. So SNPrintF should go elsehere in System.Native. Maybe pal_string?

sokket:
Thanks guys! Merging now

PR #3146, opened 2015-09-09
Fix passing byref of enum variables initialized to default values.

### LINKED ISSUE #3007: Calls with out enum arguments fail when expression tree is interpreted.
```C#
public enum MyEnum
{
    Value
}
 
public class Foo
{
    public static void Bar(out MyEnum o)
    {
        o = MyEnum.Value;
    }
 
    public void Test()
    {
        var x = Expression.Variable(typeof (MyEnum), "x");
 
        var expression = Expression.Lambda<Action>(
            Expression.Block(
                new [] { x },
                Expression.Call(null, typeof(Foo).GetMethod("Bar"), x)));
 
        expression.Compile()();
    }
}
```

Interpreter throws -

$exception  {"Object of type 'System.Int32' cannot be converted to type 'App1.MyEnum&'."}      System.ArgumentException

VSadov:
It came to the type of the default value which was based on the result of GetTypeCode and as such would be underlying value for enums.
Underlying values would be ok when passing byval since underlying types would convert to enum so the problem was not noticed, but when we pass these byref, the type of the value must be the actual enum.

Fixes #3007

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Utilities.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Utilities.cs
@@ -98,26 +98,23 @@ namespace System.Linq.Expressions.Interpreter
                     return Int32_1;
                 case 2:
                     return Int32_2;
-                case 3:
-                    return Int32_3;
             }
 
             return i;
         }
         
-        internal static readonly object Int32_m = -1;
-        internal static readonly object Int32_0 = 0;
-        internal static readonly object Int32_1 = 1;
-        internal static readonly object Int32_2 = 2;
-        internal static readonly object Int32_3 = 3;
+        private static readonly object Int32_m = -1;
+        private static readonly object Int32_0 = 0;
+        private static readonly object Int32_1 = 1;
+        private static readonly object Int32_2 = 2;
 
         public static object BooleanToObject(bool b)
         {
             return b ? True : False;
         }
 
-        internal static readonly object True = true;
-        internal static readonly object False = false;
+        private static readonly object True = true;
+        private static readonly object False = false;
 
         internal static object GetPrimitiveDefaultValue(Type type)
         {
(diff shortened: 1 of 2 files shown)


VSadov:
@AlekseyTs  - please take a look

VSadov:
CC: @weshaggard @stephentoub @jaredpar

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Utilities.cs
```diff
@@ -88,37 +88,94 @@ internal class ScriptingRuntimeHelpers
...
+            }
+
             return i;
         }
+        
+        internal static readonly object Int32_m = -1;
+        internal static readonly object Int32_0 = 0;
+        internal static readonly object Int32_1 = 1;
+        internal static readonly object Int32_2 = 2;
+        internal static readonly object Int32_3 = 3;
```
stephentoub: I assume we're doing this to avoid the repeated boxing costs?  If so, how did we arrive at doing this for the range [-1,3] and not a larger range?

stephentoub: Also, can these be private instead of internal?

stephentoub:
LGTM

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Utilities.cs
VSadov: Yes, this is to avoid boxing. That would seem to be the point of Int32ToObject.
Yes the constants can be private.

First I needed only 0 here, but then realized that there are other common cases like 1 and -1 so I have added them and then couple more for 2, 3. 
I think the bigger the number, the smaller is the statistical chance it would be used. It is a complete guess at how many are enough. 
IL constants have range [-1 : 8] , but that at best just an analogy.

VSadov: I will be conservative and make these constant go -1 through 2. Those are definitely common. More can be added if there is a more understood need.

AlekseyTs:
LGTM

PR #3148, opened 2015-09-10
Adding tests for Checked binary expressions and fixing support in int…

### LINKED ISSUE #3141: Expression interpreter - AddOvf does not throw OverflowException
In AddInstruction.cs, the following code `(Int16)checked((Int16)l + (Int16)r)` seems to be wrong and should be `checked((Int16)((Int16)l + (Int16)r))` instead.

Repro:

`Expression.Lambda(
    Expression.AddChecked(
        Expression.Constant(Int16.MinValue),
        Expression.Constant(Int16.MinValue)
    )
).Compile().DynamicInvoke()`

throws TargetInvocationException with an OverflowException when using the compiler. When using the interpreter, it succeeds.

VSadov:
…erpreter

Fixes #3141

### DIFF
--- a/src/System.Linq.Expressions/tests/BinaryOperators/Arithmetic/BinaryAddTests.cs
+++ b/src/System.Linq.Expressions/tests/BinaryOperators/Arithmetic/BinaryAddTests.cs
@@ -47,6 +47,7 @@ namespace Tests.ExpressionCompiler.Binary
                 for (int j = 0; j < array.Length; j++)
                 {
                     VerifyUShortAdd(array[i], array[j]);
+                    VerifyUShortAddOvf(array[i], array[j]);
                 }
             }
         }
@@ -60,6 +61,7 @@ namespace Tests.ExpressionCompiler.Binary
                 for (int j = 0; j < array.Length; j++)
                 {
                     VerifyShortAdd(array[i], array[j]);
+                    VerifyShortAddOvf(array[i], array[j]);
                 }
             }
         }
@@ -73,6 +75,7 @@ namespace Tests.ExpressionCompiler.Binary
                 for (int j = 0; j < array.Length; j++)
                 {
                     VerifyUIntAdd(array[i], array[j]);
+                    VerifyUIntAddOvf(array[i], array[j]);
                 }
             }
         }
@@ -86,6 +89,7 @@ namespace Tests.ExpressionCompiler.Binary
                 for (int j = 0; j < array.Length; j++)
                 {
                     VerifyIntAdd(array[i], array[j]);
+                    VerifyIntAddOvf(array[i], array[j]);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


src/System.Linq.Expressions/tests/BinaryOperators/Arithmetic/BinaryAddTests.cs
```diff
@@ -282,6 +336,54 @@ private static void VerifyShortAdd(short a, short b)
             }
         }
 
+        private static void VerifyShortAddOvf(short a, short b)
```
stephentoub: There's a lot of repetition across these tests, which all follow the same pattern and seem like they could be done simply by making a and b generic, and by having the method accept an add function as a delegate.

VSadov:
@dotnet-bot test his please

VSadov:
@dotnet-bot test this please

src/System.Linq.Expressions/tests/BinaryOperators/Arithmetic/BinaryAddTests.cs
VSadov: The ET part might work with generic a, b and typeof(T), but I am not sure the C# part of validation like "csResult = checked((short)(a + b));"   or   "" would work the same way with generics. 
I guess, if I thought about this before, it could save me some time typing, but now, considering that these are tests and are unlikely to change, I think duplicate code is acceptable here. 
If I have to change these significantly, I will consider a generic version.

VSadov:
@dotnet-bot test this please

src/System.Linq.Expressions/tests/BinaryOperators/Arithmetic/BinaryAddTests.cs
stephentoub: > I am not sure the C# part of validation like "csResult = checked((short)(a + b));"

That's why I was suggesting it also take a delegate, e.g. (a, b) => checked((short)(a + b)).

VSadov:
@stephentoub - it seems the VerifyLongMultiplyOvf hits some kind of a bug on Linux.

I wonder what compilers (C#, JIT) are used on Linux. There might be some issues with checked long multiplication.

Should I disable this test for Linux somehow?

VSadov:
Since CI does not run interpreter, both C# part and ET part are IL-compiled. 
I think, there is a very good chance this is a JIT issue.

stephentoub:
> Should I disable this test for Linux somehow?

Yeah, go ahead and disable it by adding ```[ActiveIssue(123, PlatformID.AnyUnix)]``` to the test, where 123 is the number of the issue you open to track this.  Does seem likely that it's a JIT issue (@LLITCHEV).  If you can create an isolated repro, that'd be useful, too.

PR #3152, opened 2015-09-10
Add /usr/local/include to include path for native shims.

janhenke:
CMake does not add the /usr/local/include to the compiler include search
paths by default. This breaks the build on FreeBSD, where all libraries
are installed in /usr/local instead of /usr. So add /usr/local/include
as include path for all shims.

### DIFF
--- a/src/Native/configure.cmake
+++ b/src/Native/configure.cmake
@@ -2,6 +2,10 @@ include(CheckFunctionExists)
 include(CheckStructHasMember)
 include(CheckCXXSourceCompiles)
 
+#CMake does not include /usr/local/include into the include search path
+#thus add it manually. This is required on FreeBSD.
+include_directories(/usr/local/include)
+
 check_function_exists(
     stat64
     HAVE_STAT64)


src/Native/configure.cmake
```diff
@@ -2,6 +2,10 @@ include(CheckFunctionExists)
 include(CheckStructHasMember)
 include(CheckCXXSourceCompiles)
 
+#CMake does not include /usr/local/include into the include search path
+#thus add it manually. This is requiered on FreeBSD.
```
nguerrera: s/requiered/required/

nguerrera:
Thanks.

Can you comment on why this was done slight differently in the coreclr repo: https://github.com/dotnet/coreclr/commit/7c6e325899bcdc2bb3f0feb1179f1ad3bef10d91

nguerrera:
cc @sokket

janhenke:
I tried the solution from the CoreCLR first, but for some reason that did not seem to work here. CoreCLR does a lot more configuration with CMake, which seems to make it work slightly differently.
Also the symptoms were differently. The commit above was committed, because CMake did not find the libunwind library at all. Here the ```find_library()``` call worked, just the compiler throw an error, because ```/usr/local/include``` was not passed to the compiler.

That is in essence the rationale for the difference.

sokket:
LGTM, thanks for the explanation!

nguerrera:
LGTM too. Thanks!

stephentoub:
LGTM

PR #3159, opened 2015-09-10
Added performance MicroBenchmarks

ianhays:
Summary of changes:
- Added "Performance" folders containing perf xunit tests to a number of highly-used libraries (determined by Usage data)
- Two sets of usage data from store apps; I added a test for every function used in more than 5% of the covered assemblies in both data sets.
- This is a first-round pass of the most highly used libraries and does not include: abstract classes, interfaces, or Attributes
- Created a shared parent abstract class for commonly used functions across perf tests, stored in src\Common\tests\Performance. I only added a reference to this parent if i actually ended up using it, so in the future it can be added as necessary.
- Most functions only have basic microbenchmarks with the exception of a few BigO tests that are made using InlineData or MemberData that scale

Next Steps:
- Modify applicable microbenchmarks to also test BigO by scaling the size of the input data through InlineData/MemberData. I've already done this in a few classes (System.Collections) to demonstrate the plan.
- <b>Add Warmup code before the iteration loops begin.</b>
- I did not include tests for abstract functions/classes or any interface methods. Later I can look into implementing these perf tests for those classes that implement them that are highly utilized. This would be easily reusable since they would share a common ancestor, so we could potentially test several children within the same code. For example, usage data has Object.Equals highly utilized but it's actually a child (or children) of Object that is contributing that high utilization.

Disclaimer: 
These tests won't pass the CI build or even a local build until CoreFX is using xunit beta4. I do not plan on merging this PR to CoreFX. However, before I move forward with adding more tests I'd like to get some feedback on the overall design of the perf tests (their location, structure, and style) that I've added thus far. Any questions, comments, or nitpicks are greatly appreciated :)

Ian

### DIFF
--- a/src/Common/tests/Performance/PerfTestBase.cs
+++ b/src/Common/tests/Performance/PerfTestBase.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.IO;
+
+namespace System
+{
+    /// <summary>Base class for performance test classes</summary>
+    public abstract class PerfTestBase
+    {
+        /// <summary>
+        /// Helper method to create a string containing a number of random
+        /// characters equal to the specified length
+        /// </summary>
+        public static string CreateString(int length)
+        {
+            StringBuilder builder = new StringBuilder();
+            StringBuilder empty = new StringBuilder();
+            while (builder.Length < length)
+            {
+                string toAppend = Guid.NewGuid().ToString();
+                int toAppendLength = Math.Min(toAppend.Length, length - builder.Length);
+                builder.Append(toAppend, 0, toAppendLength);
+            }
+            return builder.ToString();
+        }
+
+        /// <summary>Gets a test file full path that is associated with the call site.</summary>
+        /// <param name="index">An optional index value to use as a suffix on the file name.  Typically a loop index.</param>
+        /// <param name="memberName">The member name of the function calling this method.</param>
+        /// <param name="lineNumber">The line number of the function calling this method.</param>
+        protected string GetTestFilePath(int? index = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
+        {
+            return Path.Combine(Path.GetTempPath(), string.Format(
+                index.HasValue ? "{0}_{1}_{2}_{3}_{4}" : "{0}_{1}_{2}_{3}",
+                memberName ?? "TestBase", lineNumber, GetType().Name, Path.GetRandomFileName(), 
+                index.GetValueOrDefault()));
+        }
+    }
+}
--- a/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
+++ b/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+using Microsoft.Xunit.Performance;
+using System.Collections.Generic;
+
+namespace System.Collections.Tests
+{
+    public class Perf_HashTable : PerfTestBase
+    {
+        private IEnumerable<Hashtable> TestHashtables()
+        {
+            yield return CreateHashtable(100);
+            yield return CreateHashtable(1000);
+        }
+
+        private Hashtable CreateHashtable(int size)
+        {
+            Hashtable ht = new Hashtable();
+            for (int i = 0; i < size; i++)
+                ht.Add(CreateString(50), CreateString(50));
+            return ht;
+        }
+
+        [Benchmark]
+        public void ctor()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    new Hashtable();
+        }
+
+        [Benchmark]
+        [MemberData("TestHashtables")]
+        public void GetItem(Hashtable table)
+        {
+            object result;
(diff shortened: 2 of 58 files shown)


ianhays:
@mellinoe @ericeil @stephentoub @karelz @brianrob

src/Common/tests/Performance/PerfTestBase.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Helper method to create a string containing a number of random
+        /// characters equal to the specified length
+        /// </summary>
+        public static string CreateString(int length)
+        {
+            StringBuilder builder = new StringBuilder();
+            StringBuilder empty = new StringBuilder();
+            while (builder.Length < length)
+            {
+                string toAppend = Guid.NewGuid().ToString();
```
mellinoe: We could probably just use System.Random.NextInt() and cast it to a char or something here instead of creating GUIDs and calling ToString(). We'd have more control over the random element if we did that. It may be slightly efficient, as well.

src/Common/tests/Performance/PerfTestBase.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.IO;
+
+namespace System
+{
+    /// <summary>Base class for performance test classes</summary>
+    public abstract class PerfTestBase
```
mellinoe: I wonder if we should just separate these out into a "PerfUtils" class or something. It's possible that we might want to have a single base class for some library's Unit tests and Perf tests, and this wouldn't work well in that case. That's just speculation though. It would make the second method here more cumbersome to use than it is currently.

ianhays: The second method is <i>very</i> cumbersome, I agree. That's a good idea though, I think I may switch to to a static util class because I've ran into the issue where there is already a parent class in some test libraries and I've had to choose between the two.

The other option would be just to repeat the utilized methods per class since there isn't many of them, but I'd prefer not to do that if possible :)

src/Common/tests/Performance/PerfTestBase.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// Helper method to create a string containing a number of random
+        /// characters equal to the specified length
+        /// </summary>
+        public static string CreateString(int length)
+        {
+            StringBuilder builder = new StringBuilder();
+            StringBuilder empty = new StringBuilder();
```
stephentoub: What's the purpose of this ```empty``` StringBuilder?

src/Common/tests/Performance/PerfTestBase.cs
```diff
@@ -0,0 +1,42 @@
...
+                builder.Append(toAppend, 0, toAppendLength);
+            }
+            return builder.ToString();
+        }
+
+        /// <summary>Gets a test file full path that is associated with the call site.</summary>
+        /// <param name="index">An optional index value to use as a suffix on the file name.  Typically a loop index.</param>
+        /// <param name="memberName">The member name of the function calling this method.</param>
+        /// <param name="lineNumber">The line number of the function calling this method.</param>
+        protected string GetTestFilePath(int? index = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
```
stephentoub: Should PerfTestBase just derive from FileCleanupTestBase?  Or if not, how are temporary files created with this getting cleaned up?

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
```diff
@@ -0,0 +1,60 @@
...
+                ht.Add(CreateString(50), CreateString(50));
+            return ht;
+        }
+
+        [Benchmark]
+        public void ctor()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    new Hashtable();
```
stephentoub: Is the overhead of StartMeasurement and its IDisposable really low enough that we can measure the small cost of Hashtable's ctor appropriately and without lots of noise?  I'd have expected we'd have needed a loop inside even a single iteration in order to get something meaningful.

src/Common/tests/Performance/PerfTestBase.cs
ianhays: It has none; holdover from old code.

src/Common/tests/Performance/PerfTestBase.cs
ianhays: They cleanup of the files currently is left to the test using GetTestFilePath and is not handled by PerfTestBase.

I'd prefer not to add the extra overhead in creating a testdirectory for perf tests that didnt use it (of which there are many), Also, I think Eric's suggestion above of modifying the class to be a static PerfUtils would be preferable anyways, as it would allow classes to inherit from other test parents (such as FileCleanupTestBase).

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
ianhays: This is the case for nearly every microbenchmark that I've written and is a currently a topic of discussion over at https://github.com/Microsoft/xunit-performance/issues/62.

you make a good point about the disposer that I hadn't considered, though. I don't really want to measure disposal of the Hashtable within ```iteration.StartMeasurement()``` (since I'm testing the constructor here), so I should retain a local reference until outside of that using to avoid it being disposed too early.

stephentoub: > you make a good point about the disposer that I hadn't considered

Sorry, I didn't actually mean Hashtable's Dispose (it doesn't have one).  I meant the IDisposable that's returned from StartMeasurement, just showing that there's not only the overhead of starting the measurement but also of stopping it.

stephentoub:
Replaced by #3304

PR #3171, opened 2015-09-11
Fix System.Console test failing on OS X

stephentoub:
The strings used in the EmuTermInfoDoesntBreakParser test are missing semicolons at the end, so they fail when compared against the strings in the emu file, which has semicolons at the end.

cc: @sokket


sokket:
LGTM, thanks for finding this!

PR #3174, opened 2015-09-11
Fix System.Diagnostics.Process.Tests

Maxwe11:
System.Diagnostics.Process.Tests leaves behind temp directories in %TMP%

/cc: @stephentoub


stephentoub:
Thanks! LGTM.

PR #3198, opened 2015-09-12
Fix cancelable ReadAsync hang on OS X

stephentoub:
One of our System.IO.Pipes tests is frequently hanging on OS X. It appears that OS X has a race condition bug in its poll implementation when a FIFO is involved and there's a write to the FIFO of larger than the internal buffer size.  In such cases, even though there's an outstanding write on the FIFO, a poll for data available on the FIFO doesn't notice.

To address this, instead of passing an infinite timeout to poll, I've added a retry loop with a backoff timeout, to ensure we'll at least eventually see the write while not adding too much overhead for platforms that handle this correctly.

cc: @ellismg, @bartonjs

### DIFF
--- a/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
+++ b/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
@@ -358,22 +358,44 @@ namespace System.IO.Pipes
                         const int CancellationSentinel = -42;
                         int rv = (int)SysCall(_handle, (fd, ptr, len, cancellationFd) =>
                         {
-                            // Wait for data to be available on either the pipe we want to read from
+                            // We want to wait for data to be available on either the pipe we want to read from
                             // or on the cancellation pipe, which would signal a cancellation request.
                             Interop.Sys.PollFD* fds = stackalloc Interop.Sys.PollFD[2];
                             fds[0] = new Interop.Sys.PollFD { FD = fd, Events = Interop.Sys.PollFlags.POLLIN, REvents = 0 };
                             fds[1] = new Interop.Sys.PollFD { FD = (int)cancellationFd, Events = Interop.Sys.PollFlags.POLLIN, REvents = 0 };
-                            while (Interop.CheckIo(Interop.Sys.Poll(fds, 2, -1))) ;
 
-                            // If we woke up because of a cancellation request, bail.
-                            if ((fds[1].REvents & Interop.Sys.PollFlags.POLLIN) != 0)
+                            // Some systems (at least OS X) appear to have a race condition in poll with FIFOs where the poll can 
+                            // end up not noticing writes of greater than the internal buffer size.  Restarting the poll causes it 
+                            // to notice. To deal with that, we loop around poll, first starting with a small timeout and backing off
+                            // to a larger one.  This ensures we'll at least eventually notice such changes in these corner
+                            // cases, while not adding too much overhead on systems that don't suffer from the problem.
+                            const int InitialMsTimeout = 30, MaxMsTimeout = 2000;
+                            for (int timeout = InitialMsTimeout; ; timeout = Math.Min(timeout * 2, MaxMsTimeout))
                             {
-                                return CancellationSentinel;
+                                // Do the poll.
+                                int signaledFdCount;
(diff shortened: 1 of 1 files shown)


ellismg:
We should figure out if there's a way for us to engage with the Apple folks on this.  In the Silverlight days, some members of the team were able to file radars with Apple.  Maybe @richlander can help us figure out out what to do?

ellismg:
Do you think this is better suited in the PAL for Poll itself (ifdef'ed to only compile in on Darwin?  I could go either way...).  If you want to keep it here, it LGTM.

stephentoub:
> Do you think this is better suited in the PAL for Poll itself 

I considered that, but as this seems to only affect FIFOs, it felt too specific for the shim. I think we should leave it here for now. We can move it later if desired, or even remove it if we find a better fix or workaround.

stephentoub:
It turns out I can repro this on Linux as well, it just takes a significantly larger write.

Here's a C repro:
```C#
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <pthread.h>
#include <poll.h>

const char* fifoName = "tmpfifo";

void* doRead(void *arg)
{
    int fdRead = open(fifoName, O_RDONLY);
    if (fdRead == -1)
    {
        perror("open for read: ");
        return NULL;
    }
    
    printf("Polling...\n");
    
    struct pollfd fds[1];
    fds[0].fd = fdRead;
    fds[0].events = POLLIN;
    if (poll(fds, 1, -1) == -1)
    {
        perror("poll: ");
        return NULL;
    }
    
    printf("Finished poll");
    return NULL;
}

int main(int argc, const char * argv[]) {
    unlink(fifoName);
    if (mkfifo("tmpfifo", 0644) != 0)
    {
        perror("mkfifo: ");
        return -1;
    }
    
    pthread_t reader;
    if (pthread_create(&reader, NULL, doRead, NULL) != 0)
    {
        perror("pthread_create: ");
        return -1;
    }
    
    int fdWrite = open(fifoName, O_WRONLY);
    if (fdWrite == -1)
    {
        perror("open for write: ");
        return -1;
    }
    
    int dataLen = 12345;
    char* data = (char*)malloc(dataLen);
    if (data == NULL)
    {
        printf("malloc failed\n");
        return -1;
    }
    
    printf("Press enter to write\n");
    getc(stdin);
    
    printf("Writing...\n");
    ssize_t writeResult = write(fdWrite, data, dataLen);
    if (writeResult == -1)
    {
        perror("write: ");
        return -1;
    }
    printf("Successfully wrote %ld bytes\n", writeResult);
    
    pthread_join(reader, NULL);
    return 0;
}
```
Wait until you see "Polling..." printed out, then press enter.  On OSX, this hangs after printing "Writing..." whereas on Ubuntu the write and poll both complete successfully.  If I drop the size from 12345 down to 8192 or less (presumably the size of the buffer for the FIFO), on OS X it then passes as well.  If I increase the size of dataLen to be > 64K, then it also hangs on Ubuntu.

src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs
```diff
@@ -358,22 +358,44 @@ private unsafe int ReadCoreWithCancellation(byte[] buffer, int offset, int count
                         const int CancellationSentinel = -42;
                         int rv = (int)SysCall(_handle, (fd, ptr, len, cancellationFd) =>
                         {
-                            // Wait for data to be available on either the pipe we want to read from
+                            // We want to wait for data to be available on either the pipe we want to read from
                             // or on the cancellation pipe, which would signal a cancellation request.
                             Interop.Sys.PollFD* fds = stackalloc Interop.Sys.PollFD[2];
                             fds[0] = new Interop.Sys.PollFD { FD = fd, Events = Interop.Sys.PollFlags.POLLIN, REvents = 0 };
                             fds[1] = new Interop.Sys.PollFD { FD = (int)cancellationFd, Events = Interop.Sys.PollFlags.POLLIN, REvents = 0 };
```
bartonjs: Do you need the cancellationFd anymore? You don't seem to use it now.

stephentoub: It's still used, allowing waking up without needing to wait for the full timeout. If we wanted to remove it, we could, but we'd want to shorten the max timeout.

PR #3199, opened 2015-09-12
Reduce allocations when parsing response headers in WinHttpHandler

justinvp:
There are currently many unnecessary allocations when creating the response message in `WinHttpHandler`.

The native `WinHttpQueryHeaders` function is called a few times to retrieve response information, such as the version string, reason phrase, and raw headers.

Each call to `WinHttpQueryHeaders` currently results in 3 allocations (9 total for the 3 calls):
 1. The `StringBuilder` buffer passed to the native function to be filled in.
 2. The `char[]` inside the `StringBuilder`.
 3. The `string` allocated when `StringBuilder.ToString()` is called.

On platforms that need to do manual decompression (lower than Win 8.1), there's another 2-3 allocations to get the Content-Encoding header value.

We can reduce the allocations by 1) modifying the p/invoke signature for `WinHttpQueryHeaders` to take a `char[]` instead of `StringBuilder` and 2) allocate a single `char[]` large enough to be reused across multiple calls to `WinHttpQueryHeaders`.

With those changes, 9 allocations are down to 2: an allocation for the raw headers `char[]` and another for the reason phrase `string`.

There are many common reason phrases, such as "OK" for a 200 status code, or "Not Found" for a 404 status code. There's already internal code in `HttpStatusDescription` for looking up a known reason phrase string based on the status code, and we can make use of this existing code to avoid the reason phrase string allocation if the reason phrase `char[]` array segment is equal to a known reason phrase. 9 allocations down to 1, for common reason phrases.

When parsing the raw headers, the current code is using `string.Split` to split the raw headers by lines separated by `"\r\n"`. There are a bunch of unnecessary allocations associated with the use of `string.split`: internally `string.Split` allocates two arrays to keep track of the indexes and separator sizes, there's the overall array returned from `Split`, plus each string in the array representing each line. Each line then needs to be split further to get the header name and value for each line. The header value was also being trimmed with `string.Trim`, another potential allocation. Also, the first line of the raw headers is the status line, which is being allocated as a `string` even though this line is currently being skipped.

All of these intermediate allocations can be avoided by parsing the `char[]` directly, and only allocating `string`s for the individual header names and values. And we can avoid `string` allocations for known header names and values, similar to the known reason phrase optimization.

Notes:

 - `HttpKnownHeaderNames.TryGetHeaderName` is looking up all known header name constants, which include both request and response headers, while we only really care about response headers. My opinion is that this is OK. I think it's easier to maintain the lookup of known headers if it is just looking up all known headers, instead of trying to segregate which ones are only response headers.

 - `HttpKnownHeaderNames.TryGetHeaderName` is currently doing an ordinal lookup. Header names are case insensitive, so I'd be open to changing this to do an case insensitive lookup. I kept it ordinal so the string remains the exact same as the response string from WinHTTP. However, it looks like WinHTTP normalizes the casing of headers it knows about (e.g. WinHTTP returns `"Content-Length"` when the actual server response is `"content-length"`), so maybe switching to case insensitive lookup isn't a big deal.

 - The PR is broken up into multiple commits to make it easier to review. The earlier commits are mostly just adding helpers and tests, and moving some of the internal files to `Common` (e.g. `HttpStatusDescription`). I can squash some of the commits together, if desired.

 - There are still a bunch of allocations associated with how headers are enumerated and stored internally on each request/response in `HttpClient` that I'm looking into improving in a separate PR.


davidsh:

I'm in the process of some serious code re-arranging as I am working on #3000.  So, I'm not sure if this PR will be handled now or later.

davidsh:
I do want to comment on this aspect of the PR:

>This isn't currently a difference between CurlHandler and WinHttpHandler, but we may want to make HttpResponseHeaderReader more lenient about potential white space characters between the http version and status code and between the status code and reason phrase. Right now, both handlers are only looking for a single space (per the RFC), but WinHTTP actually allows additional white space characters between these, which it normalizes to a single space character when WinHttpQueryHeaders is called with WINHTTP_QUERY_RAW_HEADERS_CRLF (on Windows 10; not sure about other versions). I haven't investigated what CURL allows and whether it normalizes the white space between these to a single character like WinHTTP.

WinHttpHandler does not parse out the http version, status code and reason phrase itself using WINHTTP_QUERY_RAW_HEADERS.  We let WinHTTP do that for us and ask for those things separately via other WinHTTP header query APIs.  WinHTTP already does the work of parsing that so there is no point in doing it ourselves.  We only use WINhTTP_QUERY_RAW_HEADERS to get the headers themselves in order to parse that into our strongly typed header collections.

So, I'm not sure of the value of putting that particular parsing logic into Common since WinHttpHandler won't be using it.

justinvp:
> I'm not sure of the value of putting that particular parsing logic into Common since WinHttpHandler won't be using it.

I meant to call this out specifically. This is to avoid the allocations associated with retrieving the version and reason phrase `char[]`s. If we're already parsing those, it's easy enough to also parse the status code as part of that. We could still grab the status code directly from WinHTTP as there are no allocations associated with retrieving a number, or could add a `Debug.Assert` to ensure the status code we parsed is the same as WinHTTP.

justinvp:
Thinking about it more, if we want to avoid parsing the status line ourselves (even though it looks like we have to do it for CURL) and get the version and reason phrase from WinHTTP, we could do it in a way that avoids the two intermediate `char[]` allocations for the version and reason phrase:
 1. Get the size of the `char[]` buffer needed for `WINHTTP_QUERY_RAW_HEADERS_CRLF`, which would be big enough for either the version or reason phrase, because the raw headers includes the status line that has those.
 2. Create a single `char[]` of that size, and use that same single `char[]` buffer for each of the interop calls to retrieve the version, reason phrase, and raw headers from WinHTTP.

justinvp:
I removed the status line parsing from `HttpResponseHeaderReader`; PR updated. `WinHttpHandler` is back to getting the version, status code, and reason phrase by calling `WinHttpQueryHeaders`, but I've reduced the allocations by creating a single shared `char[]` buffer that's large enough to be reused across multiple calls to `WinHttpQueryHeaders`.

justinvp:
@dotnet-bot test this please (unrelated CI test failure)

justinvp:
@dotnet-bot test this please (looks like there was an unrelated Linux CI build abort)

CIPop:
@davidsh @justinvp While the description makes sense, I think the ordering should be:
1. Add the Async perf-enhancement to WinHttpHandler 
2. Add PerformanceTests and analysis to compare CPU, memory and network utilization to existing stacks (HttpWebRequest, NodeJS, etc)
3. Add this change if the comparison with existing results shows improvements.

> The PR is broken up into multiple commits to make it easier to review. The earlier commits are mostly just adding helpers and tests, and moving some of the internal files to  Common  (e.g.  HttpVersion  and  HttpStatusDescription ). I can squash some of the commits together, if desired.

+1 on this PR method :)

davidsh:
@justinvp Do you think you can refresh this PR so that we can consider it?  Thx.

justinvp:
@dotnet-bot test this please

justinvp:
@davidsh I've updated the PR, but for some reason CI doesn't run when I force push the changes after rebasing. I'm going to close and reopen the PR to see if that'll kick off a CI build.

davidsh:
@stephentoub PTAL

davidsh:
@stephentoub Do you have any further comment on this revision?  Thx.

stephentoub:
Sorry, David, I lost track of this... will review later today.

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
```diff
@@ -0,0 +1,218 @@
...
+                    return TryGet(SecWebSocketVersion, key, startIndex, length, out name); // Sec-WebSocket-Version
+
+                case 22:
+                    return TryGet(SecWebSocketProtocol, key, startIndex, length, out name); // Sec-WebSocket-Protocol
+
+                case 24:
+                    return TryGet(SecWebSocketExtensions, key, startIndex, length, out name); // Sec-WebSocket-Extensions
+            }
+
+            // When adding a new constant, add it to HttpKnownHeaderNames.cs as well.
```
stephentoub: Nit: this comment is repeated from above

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            name = null;
+            return false;
+        }
+
+        /// <summary>
+        /// Returns true if <paramref name="known"/> matches the <paramref name="key"/> char[] array segment,
+        /// using an ordinal comparison.
+        /// </summary>
+        private static bool TryGet(string known, char[] key, int startIndex, int length, out string name)
```
stephentoub: Nit: TryMatch might be a better name than TryGet

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// <summary>
+        /// Gets a known header name string from a matching char[] array segment, using an ordinal comparison.
+        /// Used to avoid allocating new strings for known header names.
+        /// </summary>
+        public static bool TryGetHeaderName(char[] key, int startIndex, int length, out string name)
+        {
+            CharArrayHelpers.DebugAssertArrayInputs(key, startIndex, length);
+
+            // When adding a new constant, add it to HttpKnownHeaderNames.cs as well.
+
```
stephentoub: Consider adding a short comment explaining the looking scheme being used here.

src/Common/src/System/Net/HttpKnownHeaderNames.cs
```diff
@@ -68,5 +70,7 @@ internal static class HttpKnownHeaderNames
         public const string Warning = "Warning";
         public const string XAspNetVersion = "X-AspNet-Version";
         public const string XPoweredBy = "X-Powered-By";
+
+        // When adding a new constant, add it to HttpKnownHeaderNames.TryGetHeaderName.cs as well.
```
stephentoub: I see the comment repeated here as well.. I guess the repetition is on purpose?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
```diff
@@ -0,0 +1,168 @@
...
+                // Skip malformed header lines that are missing the colon character.
+                if (colonIndex == -1)
+                {
+                    continue;
+                }
+
+                int nameLength = colonIndex - startIndex;
+
+                // If it's a known header name, use the known name instead of allocating a new string.
+                if (!HttpKnownHeaderNames.TryGetHeaderName(_buffer, startIndex, nameLength, out name))
```
stephentoub: Can we use the same HttpKnownHeaderNames as an optimization in CurlHandler in a subsequent PR? (Same question for any other optimizations.)

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
```diff
@@ -130,104 +156,133 @@ public static uint GetResponseHeaderNumberInfo(SafeWinHttpHandle requestHandle,
...
+        }
+
+        private static int GetResponseHeaderInfo(SafeWinHttpHandle requestHandle, uint infoLevel, char[] buffer)
+        {
+            Debug.Assert(buffer != null, "buffer must not be null.");
+            Debug.Assert(buffer.Length > 0, "buffer must not be empty.");
+
+            uint bufferLengthInBytes = (uint)buffer.Length * 2;
+
+            GCHandle pinnedHandle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
```
stephentoub: It'd be nice to use fixed instead of a pinned GCHandle here, as it's faster.  Are you using a pinned handle because buffer might be empty and you're trying to avoid special casing it?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
```diff
@@ -130,104 +156,133 @@ public static uint GetResponseHeaderNumberInfo(SafeWinHttpHandle requestHandle,
...
-                Interop.WinHttp.WINHTTP_HEADER_NAME_BY_INDEX,
-                buffer,
-                ref bytesNeeded,
-                IntPtr.Zero);
-            if (!results)
+        private static bool TryGetKnownReasonPhrase(HttpStatusCode statusCode, char[] array, int startIndex, int length, out string reasonPhrase)
+        {
+            CharArrayHelpers.DebugAssertArrayInputs(array, startIndex, length);
+
+            reasonPhrase = HttpStatusDescription.Get(statusCode);
```
stephentoub: Is this always correct? i.e. is it not possible for the server to send a different reason phrase than the canonical one associated with the status code?

stephentoub: Oh, nevermind, I see how this is being used.

src/System.Net.WebSockets.Client/src/System/Net/WebSockets/WinHttpWebSocket.cs
```diff
@@ -688,21 +688,29 @@ private string GetResponseHeaderStringInfo(string headerName)
...
-            results = Interop.WinHttp.WinHttpQueryHeaders(
-                _operation.RequestHandle,
-                Interop.WinHttp.WINHTTP_QUERY_CUSTOM,
-                headerName,
-                buffer,
-                ref bytesNeeded,
-                IntPtr.Zero);
-            if (!results)
+            char[] buffer = new char[charsNeeded];
+            GCHandle pinnedHandle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
```
stephentoub: Same GCHandle question.

stephentoub:
A few minor questions/comments, otherwise LGTM.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
davidsh: >Is this always correct? i.e. is it not possible for the server to send a different reason phrase than the canonical one associated with the status code?

This is NOT always correct.  The server can send any random string for the ReasonPhrase.  It is configurable on many servers and even the ASP.NET APIs allow for any string to be passed overriding the default one.

So, we always need to parse out the real reason phrase from what was received on the wire.

davidsh: Adding this code makes me think we need some more interesting tests where a server will send a status code, say 404, and instead of the default "Not Found" reason phrase, will send out something like "Sorry I couldn't find that page".

I will modify the networking test server so that the StatusCode.ashx endpoint can accept a string parameter and I can write tests for that.

stephentoub: Yeah. But I realized after I posted that Justin is already accounting for that. This function is trying to match assist the exact typically expected string. If it matches, it uses the singleton instead of allocating a new identical copy. If it can't, it falls back to using the real response.

stephentoub: > I will modify the networking test server 

Cool.

src/Common/tests/Tests/System/Net/HttpKnownHeaderNamesTest.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Justin Van Patten. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
```
davidsh: nit: Add blank line to separate top level 'using' groups (as per CoreFx coding guidlines).

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Justin Van Patten. All rights reserved.
```
davidsh: This is the first time I've seen extra Copyright headers added to the source files.  

@stephentoub Is that allowed per our licensing guidelines for CoreFx?

src/System.Net.Http.WinHttpHandler/tests/UnitTests/WinHttpResponseHeaderReaderTest.cs
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) Justin Van Patten. All rights reserved.
```
davidsh: @stephentoub This differs from the usual copyright headers.  What is the policy on this?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
justinvp: See the discussion here: https://github.com/dotnet/coreclr/pull/1329#discussion_r36218185

This has been done elsewhere in coreclr and corefx repos.

src/System.Net.Http.WinHttpHandler/tests/UnitTests/WinHttpResponseHeaderReaderTest.cs
justinvp: See the discussion here: https://github.com/dotnet/coreclr/pull/1329#discussion_r36218185

This has been done elsewhere in the coreclr and corefx repos.

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
justinvp: Yeah, I intentionally added the same comment at both the top and bottom of the file. Based on your feedback, I went ahead and removed the comment at the bottom. One comment at the top should be sufficient.

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
justinvp: Agreed. Fixed.

src/Common/src/System/Net/HttpKnownHeaderNames.TryGetHeaderName.cs
justinvp: Done.

src/Common/src/System/Net/HttpKnownHeaderNames.cs
justinvp: Yeah, I intentionally added the same comment at both the top and bottom of the file. Based on your feedback, I went ahead and removed the comment at the bottom. One comment at the top should be sufficient.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
justinvp: I'd like to. Though, here [`CurlHandler` is operating on `string` segments for headers](https://github.com/dotnet/corefx/blob/ac37343ac90fdb002041394e93f107b94ec2a5b2/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.MultiAgent.cs#L546), whereas `WinHttpHandler` is operating on `char[]` array segments, so the `KnownHeaderNames` lookup code would need to be tweaked in order to work with both `string` and `char[]` segments. (If only there was a type, say `Slice<char>`, that made this easier... :wink:)

I was planning to look into this as a separate PR (for this, and the other optimizations).

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
justinvp: Switched to using `fixed`. The reason I used `GCHandle` here was because this is how it was being done elsewhere in WinHttpHandler. (When I originally submitted this PR, the code in WinHttpHandler looked like [this](https://github.com/dotnet/corefx/blob/a75ec027daa53114c788464878e2947ae7d5a65e/src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseStream.cs#L100), for example. It has since changed a bit: it's still using `GCHandle` but is doing some caching of the handle now).

src/System.Net.WebSockets.Client/src/System/Net/WebSockets/WinHttpWebSocket.cs
justinvp: I switched this to `stackalloc` if `charsNeeded <= 128` stack limit, otherwise allocate a `char[]` on the heap and use `fixed`. The 128 limit is arbitrary, but based on other 256 byte stack limit I saw elsewhere in corefx and the fact that it's very likely the header value will be under 128 chars for the single Sec-WebSocket-Protocol header being queried here (based on the lengths of registered identifiers in the [WebSocket Subprotocol Name Registry](https://www.iana.org/assignments/websocket/websocket.xml); the longest being 55 chars in length).

It might be nice to look into the `stackalloc` with heap-allocated-fallback for `WinHttpHandler`'s  buffer, for cases when the response has a status line + headers length under some stack limit, but I don't know how common that is.

justinvp:
Thanks for the feedback, Steve and David. I added some more commits to the PR. I plan to squash these into the existing commits before merging (or I can squash everything into a single commit, if you prefer). Let me know.

davidsh:
@justinvp Thank you for your work on this PR and your patience as well. At this point, I would suggest preparing for merge and thus please squash down to 1 commit.

We still have a pending issue/clarification regarding the copyright headers, but we hope to hear back soon on that.

justinvp:
Squashed

justinvp:
@davidsh Any news on the copyright question? More merge conflicts have cropped up and it looks like #5001 will introduce more conflicts, which I'll need to address before this PR can be considered/merged.

davidsh:
@justinvp The status of a few days ago was that it is getting finalized and hopefully we'll hear back soon. We'll post guidance to all of CoreFx probably after the holidays.

stephentoub:
> Any news on the copyright question?

Thanks for your patience, @justinvp.  Details are now here https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/contributing.md#copyright

stephentoub:
@justinvp, are you able to move ahead with this?

justinvp:
> @justinvp, are you able to move ahead with this?

Yes, I should have the PR updated in the next few days. Sorry for the delay.

stephentoub:
Not a problem.  I just wanted to check if it was still in progress.  If it wasn't, I'd take it over, as I think it's a valuable thing to be done.  Are you planning to incorporate similar changes into CurlHandler?

justinvp:
Test Innerloop OSX Debug Build and Test please

justinvp:
I've fixed the merge conflicts and license headers.

I made some additional changes to the following files since the last review:

 - **WinHttpCookieContainerAdapter.cs**: This file was new since the previous review. I went ahead and made some optimizations as part of this PR: Instead of getting back a `List<string>` headers, iterate through each header directly using a stack-allocated buffer (if possible), otherwise falling back to a heap-allocated buffer.

 - **WinHttpResponseParser.cs**: Replaced the `List<string>` -returning method (as mentioned above) and made some other changes to how `WinHttpQueryHeaders` is called throughout this file.

 - **FakeInterop.cs**: Fixed up how `ERROR_INSUFFICIENT_BUFFER` is returned to match new usage and reduced duplicated code.

 - **WinHttpWebSocket.cs**: Made calls to `WinHttpQueryHeaders` more consistent with how it's now called from `WinHttpResponseParser.cs` using a stack-allocated buffer (if possible), with fallback to heap-allocated buffer.

@stephentoub, I'll look into similar changes to CurlHandler in a separate PR after this has been reviewed and merged.

stephentoub:
Thanks, @justinvp.  I'll re-review.  Would it be possible for you to share a before/after look at what the allocations are involved in an HttpClient request on Windows with your PR?

justinvp:
@dotnet-bot test this please

src/Common/src/System/CharArrayHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+            for (int i = 0; i < left.Length; i++)
+            {
+                uint charA = left[i];
+                uint charB = right[rightStartIndex + i];
+
+                // We're only interested in ASCII characters here.
+                if ((charA - 'a') <= ('z' - 'a')) charA -= 0x20;
+                if ((charB - 'a') <= ('z' - 'a')) charB -= 0x20;
+
+                if (charA != charB)
```
stephentoub: I must be missing it... how is this ignoring case?

stephentoub: Oh, nevermind, I see.  For letter's a to z we're shifting them down to be A to Z.  Got it.

src/Common/src/System/CharArrayHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            int end = length - 1;
+            while (end >= offset && char.IsWhiteSpace(array[startIndex + end]))
+            {
+                end--;
+            }
+
+            startIndex += offset;
+            length = end - offset + 1;
+        }
```
stephentoub: Doesn't need to be now, be we should look at moving this into Common along with code very close to it: https://github.com/dotnet/corefx/blob/master/src/Common/src/System/StringExtensions.cs

src/Common/src/System/CharArrayHelpers.cs
justinvp: Sure, I can fix. BTW, I just lifted this from CompareInfo: https://github.com/dotnet/coreclr/blob/43b39a73cbf832ec13ec29ed356cb75834e7a8d7/src/mscorlib/corefx/System/Globalization/CompareInfo.cs#L319-L320

src/Common/src/System/CharArrayHelpers.cs
stephentoub: I just realized, you already did this :smile:

src/Common/src/System/CharArrayHelpers.cs
stephentoub: Thanks.  I find it just as magical there ;)

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
stephentoub: Sounds good.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
```diff
@@ -0,0 +1,168 @@
...
+                _position = i;
+                return true;
+            }
+
+            startIndex = 0;
+            length = 0;
+            return false;
+        }
+
+        private static bool TryGetKnownHeaderValue(string name, char[] array, int startIndex, int length, out string value)
```
stephentoub: Is the plan to flesh this method out more, or do you expect it'll remain just for gzip/deflate?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
```diff
@@ -33,27 +28,37 @@ internal static class WinHttpResponseParser
...
+            {
+                string reasonPhrase;
+
+                // If it's a known reason phrase, use the known reason phrase instead of allocating a new string.
+                if (!TryGetKnownReasonPhrase(response.StatusCode, buffer, 0, reasonPhraseLength, out reasonPhrase))
+                {
+                    reasonPhrase = new string(buffer, 0, reasonPhraseLength);
+                }
+
+                response.ReasonPhrase = reasonPhrase; // It's OK if this is null.
```
stephentoub: > It's OK if this is null

In what case would it be null?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
justinvp: It could certainly be fleshed out more if there are other headers with common known values. I didn't spend much time investigating whether there are any others like that.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+                else if (CharArrayHelpers.EqualsOrdinalAsciiIgnoreCase(Deflate, array, startIndex, length))
+                {
+                    value = Deflate;
+                    return true;
+                }
+            }
+
+            value = null;
+            return false;
```
stephentoub: Does this method need to be a Try* method, or could it just return the string, either a cached one or an allocated one?

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
```diff
@@ -169,77 +229,104 @@ public static List<string> GetResponseHeaders(SafeWinHttpHandle requestHandle, u
...
+            }
+
+            if (CharArrayHelpers.EqualsOrdinal(reasonPhrase, array, startIndex, length))
+            {
+                return true;
             }
 
-            return buffer.ToString();
+            reasonPhrase = null;
+            return false;
```
stephentoub: Same question here as on the other Try*s... won't the caller always immediately want to reify the value into a string if we couldn't get a cached one?  Why make this a Try then, rather than just returning the allocated string?

stephentoub:
Overall this looks good, but some comments/questions.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
justinvp: I removed the comment (it was a holdover from an earlier iteration that I should have been deleted).

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseHeaderReader.cs
justinvp: No it doesn't need to be a Try method. Fixed.

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
justinvp: Fixed.

justinvp:
I added a new commit that addresses the review feedback as well as another commit that tweaks one of the tests slightly. (I will squash after the review).

@stephentoub, I converted the Try methods to non-Try except for `HttpKnownHeaderNames.TryGetHeaderName` which I left as a Try* because testing found/not-found seems more straightforward with the Try* method. Though, I don't feel strongly about it as it certainly could return a string (allocating a new string if not a known header); I'd just have to update the tests to call the method more than once and check to see if the same object is returned for the found case and different objects for the not-found case.

> Would it be possible for you to share a before/after look at what the allocations are involved in an HttpClient request on Windows with your PR?

I temporarily instrumented `WinHttpResponseParser.CreateResponseMessage` with `Console.WriteLine(GC.GetTotalMemory(forceFullCollection: true))` at the beginning of the method and `Console.WriteLine(GC.GetTotalMemory(forceFullCollection: false))` at the end of the method and ran the following before & after the changes in this PR:

```csharp
{
    var response = client.GetAsync("http://www.bing.com").GetAwaiter().GetResult();
}
```

Here are the results:

|        | GC.GetTotalMemory Beginning | GC.GetTotalMemory End | Difference |
|--------|:----------------------:|:----------------:|:----------:|
| Before |                146,920 |          179,688 |     32,768 |
| After  |                146,920 |          171,496 |     24,576 |

justinvp:
Test Innerloop Ubuntu Debug Build and Test please

davidsh:
@stephentoub Do you have an final comments on this PR?

stephentoub:
LGTM.  Thanks, @justinvp.

Note that while the GC total memory before/after is interesting, it doesn't really reflect the savings from the PR, as you'd expect the GC to be running during the test and cleaning up rather than just letting things leak.  I was more interested in understanding the frequency and amount of allocations happening during the test.  I just ran a similar test:
```C#
using (HttpClient client = new HttpClient())
{
    for (int i = 0; i < 500; i++) 
    {
        client.GetAsync("http://httpbin.org/ip").GetAwaiter().GetResult().Dispose();
    }
}
```
Before:
image
After:
image

davidsh:
LGTM

justinvp:
Thanks guys.

@stephentoub, Ah, the profiler data is much better (thanks). Note: the allocation reductions aren't limited to `String` alone. Here are some other relevant types with reductions (running your same test):

Before:
before

After:
after

src/System.Net.Http.WinHttpHandler/src/System/Net/Http/WinHttpResponseParser.cs
```diff
@@ -33,47 +28,56 @@ internal static class WinHttpResponseParser
             var response = new HttpResponseMessage();
             bool stripEncodingHeaders = false;
 
+            // Create a single buffer to use for all subsequent WinHttpQueryHeaders string interop calls.
+            // This buffer is the length needed for WINHTTP_QUERY_RAW_HEADERS_CRLF, which includes the status line
+            // and all headers separated by CRLF, so it should be large enough for any individual status line or header queries.
+            int bufferLength = GetResponseHeaderCharBufferLength(requestHandle, Interop.WinHttp.WINHTTP_QUERY_RAW_HEADERS_CRLF);
+            char[] buffer = new char[bufferLength];
```
CIPop: @justinvp @davidsh @stephentoub I believe this is now allocating twice (or even 4x) the required buffer, for _each_ request:
  * 2x from the following code path: `GetResponseHeaderCharBufferLength` is calling into `WinHttpQueryHeaders` which returns the buffer size in bytes not C# `char` which is 16bit wide.
  * another 2x from double buffering instead of lazily extracting headers as they are requested by the application.

The last point is debatable and would need to be profiled (which, if I have time, I will do part of my perf. work). Lacking the necessary telemetry it's hard to understand if most apps really care about the headers or not.

justinvp: If you look at the code before this PR you'll see that it was worse before... Previously it queried `WINHTTP_QUERY_RAW_HEADERS_CRLF` along with a few other headers. Each time it did so a `StringBuilder` would be allocated (along with underlying `char[]` buffer), then there would be copying going on back and forth as part of the interop marshaling, and then a call to `ToString` would allocate a string to return, so previously there was a lot of allocations and copying going on. With this PR, all those `StringBuilder`/`char[]`/`string` allocations (and copying) are reduced to a only a single `char[]` buffer allocation.

I agree it'd be nice if the headers were lazily extracted as requested by callers -- avoiding the buffer allocation and parsing, if unnecessary.

CIPop: @justinvp Thanks for explaining: I didn't get the time to closely look at the prior implementation.
I'll open 2 separate issues to track the problems since the first one is much simpler to solve.

CIPop: /cc @davidsh @himadrisarkar 
Opened #11978 and #11979.

PR #3200, opened 2015-09-12
Add missing IsInvalid check to ECDsaOpenSsl's GenerateKey

stephentoub:
After calling EC_KEY_new_by_curve_name, we're currently checking whether the returned SafeHandle is null, but not whether it's Invalid, and if EC_KEY_new_by_curve_name fails, it'll return a null pointer, which will be stored into a non-null SafeHandle instance.  This means that if EC_KEY_new_by_curve_name fails, we then get a "passed a null parameter" error from the subsequent call to EC_KEY_generate_key, rather than an exception for the original error from EC_KEY_new_by_curve_name.

This commit just adds the appropriate check.

cc: @bartonjs, @AtsushiKan


bartonjs:
LGTM

PR #3201, opened 2015-09-12
Fix Path.GetTempPath() test on OS X

stephentoub:
A Path test is assuming that the original temp path will match the temp path created by removing the TMPDIR environment variable.  That's typically the case on Linux, where TMPDIR is usually "/tmp", but on OS X TMPDIR is often initialized to something like "/var/folders/51/kchrq4qs1cd5...", such that after setting TMPDIR to null, the resulting "/tmp/" path no longer matches.

This commit just fixes the test to reset the TMPDIR environment variable back to what it was at the beginning of the test, rather than to null, and then adds a specific test case to validate that deleting TMPDIR results in "/tmp" as the new temp path.

cc: @Sridhar-MS


nguerrera:
LGTM

PR #3204, opened 2015-09-12
Fix several Process shim function to return int32_t instead of bool

stephentoub:
Several of the process shim functions are returning the C++ bool type instead of int32_t.  This is causing values to be marshaled incorrectly to the consuming managed code and causing some tests to fail on OS X.

cc: @sokket, @nguerrera


nguerrera:
LGTM

PR #3208, opened 2015-09-12
Fix incorrect error handling when calling getpwuid_r shim

nguerrera:
getpwuid_r as well as its shim return errors rather than set them via
errno. Fix the single managed caller to interpret the value correctly.

Also, converge on the pattern established by readdir_r shim to map the
extra out **param (which can only be null or equal to input *param) to
a negative return value instead.

Finally, adopt a coding pattern in both of these shims that should
make it easier to see the 3 cases (success, not-found/end-of-stream,
and failure) and more obvious that the return value is an error code
and not the typical syscall 0/-1 + errno.

cc @stephentoub


stephentoub:
LGTM

Issue #3207, opened 2015-09-12
PFX certificates are not well-ordered across implementations


bartonjs:
Similar to issues #2743 and #2745, the fix would be to not use the `PKCS12_parse` and `PKCS12_create` helper functions, but to get into the nitty-gritty with PKCS12 primitives.

This manifests as both writing order and reading order, to be consistent across Windows, Linux, and OSX.

PR #3209, opened 2015-09-12
Make PFX import tests be order-independent for now

### LINKED ISSUE #2893: Two X509 Pfx test failures on OS X
I get these when running locally on OS X:
```
Discovering: System.Security.Cryptography.X509Certificates.Tests
Discovered:  System.Security.Cryptography.X509Certificates.Tests
Starting:    System.Security.Cryptography.X509Certificates.Tests
   System.Security.Cryptography.X509Certificates.Tests.CollectionTests.ExportUnrelatedPfx [FAIL]
      Assert.Equal() Failure
      Expected: String[] ["Microsoft Corporation", "cn.subject.example.org", "Foo"]
      Actual:   String[] ["Foo", "cn.subject.example.org", "Microsoft Corporation"]
      Stack Trace:
            at System.Security.Cryptography.X509Certificates.Tests.CollectionTests.ExportUnrelatedPfx()
   System.Security.Cryptography.X509Certificates.Tests.CollectionTests.ImportFullChainPfx [FAIL]
      Assert.Equal() Failure
      Expected: String[] ["MS Passport Test Sub CA", "MS Passport Test Root CA", "test.local"]
      Actual:   String[] ["MS Passport Test Root CA", "MS Passport Test Sub CA", "test.local"]
      Stack Trace:
            at System.Security.Cryptography.X509Certificates.Tests.CollectionTests.ImportFullChainPfx()
Finished:    System.Security.Cryptography.X509Certificates.Tests
```
Looks like some data is ending up in an unexpected order.

bartonjs:
There seems to be a stack vs queue problem for the PKCS12 helper functions on OSX's version (0.9.8 family) and the Ubuntu version (1.0.1 family), so some of the ordering is not the same between them.

Issue #3207 tracks the ordering problem, mainly that we need to take over the heavy lifting from the OpenSSL PKCS12 helper functions (which was already known, as it is limited to a single private key).

Fixes #2893.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/tests/CollectionImportTests.cs
+++ b/src/System.Security.Cryptography.X509Certificates/tests/CollectionImportTests.cs
@@ -235,7 +235,6 @@ namespace System.Security.Cryptography.X509Certificates.Tests
             Assert.Equal(3, count);
         }
 
-        [ActiveIssue(2893, PlatformID.OSX)]
         [Fact]
         public static void ImportPkcs12File_Chain_VerifyContents()
         {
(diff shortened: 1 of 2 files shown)


bartonjs:
cc @stephentoub

stephentoub:
LGTM.  Thanks for addressing it.

PR #3221, opened 2015-09-13
Make CryptographicException(int) load the correct message for the platform

### LINKED ISSUE #3062: CryptographicException(int hr) should assign Message based on the code
Using Interop.mincore.GetMessage (or whatever the correct API is) on Windows, and the OpenSSL ERR_ APIs on Unix.

bartonjs:
Makes use of FormatMessage on Windows, and libcrypto::ERR_error_string_n on Unix.

Fixes #3062.


bartonjs:
Odd, the failure in the linux test looks like we didn't load error strings properly...

bartonjs:
Okay, what makes sense is that I didn't include the Initialization file.  What doesn't make sense is that it worked when I ran it locally.

src/System.Security.Cryptography.Primitives/tests/CryptographicException.cs
```diff
@@ -20,5 +21,44 @@ public static void Ctor()
             Assert.Same(inner, new CryptographicException(message, inner).InnerException);
             Assert.Equal(message, new CryptographicException(message, inner).Message);
         }
+
+        [Theory]
+        [InlineData(0x00000000, "The operation completed successfully")]
+        [InlineData(0x80090013, "Invalid provider specified")]
+        [InlineData(0x80090016, "Keyset does not exist")]
```
nguerrera: These could end up being different for different OS versions, right? Would we just wait for failure and tolerate different varieties?

nguerrera:
I'm a little disappointed that Primitives is no longer purely portable. 

One can at least imagine using different technologies on different OSes. E.g. OpenSSL on Windows.

Would it be better to have Common throw helpers that all of the non-primitive contracts use that always pass the right message?

nguerrera:
Regarding the lack of initialization file in first iteration working locally, it could be test ordering, i.e. as long as someone initialized before test runs it would pass. I wonder if there's a pattern we can use to make sure the initialization is included.

bartonjs:
@nguerrera The problem is that there's no way to get both the correct code and message currently, because CryptographicException has no (int, string) ctor; and neither Message nor HResult can be set externally.

My motivation is there's one function in the X509 CRL path returns one error that is ignorable; right now it skips error checking altogether because there's no good way to filter it out (because in Unix we set message, not HR; in Windows we set HR, not message).

This change brings it back in line with how it works on Desktop.  Or, if we're okay with the value-of-message behavioral difference there are a couple of options:
* Figure out how to produce 4.0.0 and 4.0.10 concurrently, 4.0.10 containing an (int, string) ctor (which would no longer be a facade on Desktop, and might lead to some wonky type identity problems)
* Leave things alone for now; leave the one place where I wanted to suppress a single error code as ignoring errors altogether, come back post RTM and make a 4.0.10; upgrade all Windows code to make use of that via a helper chokepoint.
* Do the change now, and post RTM introduce the new ctor, revert Primitives to the fixed message for (int) and upgrade the Windows calling code
* Hold off until we can get the new ctor into Desktop, then let the CoreFX code be upgraded (and leave Primitives as always pure portable)

src/System.Security.Cryptography.Primitives/tests/CryptographicException.cs
bartonjs: Yeah, I'm hopeful that these messages haven't been updated since the dawn of time, and will continue to be in the "we already have a message for this" camp.

nguerrera:
Hmm, I wrote that assuming `CryptographicException(string, int)` existed. :(

One workaround would be for the throw helper to throw a private nested class that extends CrytpographicException. It could call base with message and set HResult which is protected. Would that work?

bartonjs:
@nguerrera That's... a clever workaround.  There are a few things that I can think of that we'd want to say we're okay with:

* We'd need to do it for both Windows and Unix, assuming that we want both of them to have good strings and HR values.
* Anyone who was checking `e.GetType() == typeof(CryptographicException)` as opposed to `e is CryptographicException` would be sad
* There'd still be a difference in Message for anyone else calling `new CryptographicException(int)`

But I'm definitely interested in this approach provided we're okay with the "problems".

nguerrera:
I personally OK with all of that, but probably worth a second (third?) opinion.

nguerrera:
No matter what we do, we should have a tracking issue to add the (string, int) ctor and either undo the workaround or collapse Primitive back to portable. This could wait until after we've shipped v1 though.

bartonjs:
@stephentoub How do you feel about a private derived type for CryptographicException to keep Primitives from doing interop (until we can get a change into Desktop)?

stephentoub:
> @stephentoub How do you feel

I like it.

bartonjs:
Based on the consensus, I'm closing this one and will follow up with the new plan.

Thanks, guys.

PR #3222, opened 2015-09-13
Fix inverted return value for user entry not found

nguerrera:
This was an error in a recent refactoring. The comment above
explains exactly why this should be return true, but the wrong
value was used in the actual code below.

Oops.

cc @stephentoub


nguerrera:
I'm not really sure how to hit this case where the current user would not have a passwd entry.

stephentoub:
LGTM.  (I'm not sure about how to test this either.)

bartonjs:
Re: Testing

I think the only case this would hit is to "delete" the user after the test started; perhaps by using vipasswd to change their UID after it launched.  I'd recommend doing this on something other than your user account (and, of course, root).

Issue #3196, opened 2015-09-12
Expression interpreter - IsTrue, IsFalse, UnaryPlus missing cases


bartdesmet:
Evaluating the following expressions results in NotSupportedException, where it doesn't for the compiler. It seems there's the wrong assumption those node types are always used in conjunction with a method, but that's not the case for primitive types:

Test<bool>(Expression.IsTrue(Expression.Constant(true)));
Test<bool>(Expression.IsFalse(Expression.Constant(true)));
Test<int>(Expression.UnaryPlus(Expression.Constant(42)));

where Test<T> is a simple wrapper around Expression.Lambda<Func<T>>(e).Compile()()

VSadov:
closed in https://github.com/dotnet/corefx/pull/3227

PR #3227, opened 2015-09-14
Expression interpretation support for UnaryPlus, IsTrue, and IsFalse

bartdesmet:
Those node types were not supported by the expression interpreter for primitive types, while the factory methods allow them. The expression compiler works fine for those.

This fixes issue #3196.

### DIFF
--- a/src/System.Linq.Expressions/tests/Unary/UnaryIsFalseNullableTests.cs
+++ b/src/System.Linq.Expressions/tests/Unary/UnaryIsFalseNullableTests.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Xunit;
+
+namespace Tests.ExpressionCompiler.Unary
+{
+    public static class UnaryIsFalseNullableTests
+    {
+        #region Test methods
+
+        [Fact] //[WorkItem(3196, "https://github.com/dotnet/corefx/issues/3196")]
+        public static void CheckUnaryIsFalseNullableBoolTest()
+        {
+            bool?[] values = new bool?[] { null, false, true };
+            for (int i = 0; i < values.Length; i++)
+            {
+                VerifyIsFalseNullableBool(values[i]);
+            }
+        }
+
+        #endregion
+
+        #region Test verifiers
+
+        private static void VerifyIsFalseNullableBool(bool? value)
+        {
+            Expression<Func<bool?>> e =
+                Expression.Lambda<Func<bool?>>(
+                    Expression.IsFalse(Expression.Constant(value, typeof(bool?))),
+                    Enumerable.Empty<ParameterExpression>());
+            Func<bool?> f = e.Compile();
+            Assert.Equal((bool?)(value == default(bool?) ? default(bool?) : value == false), f());
+        }
+
+
+        #endregion
+    }
+}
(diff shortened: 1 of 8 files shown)


bartdesmet:
This fixes issue #3196.

bartdesmet:

stephentoub:
cc: @VSadov

bartdesmet:
This is missing the lifted case. I'll revisit the PR to address this but please chime in on the desired approach for adding tests (and running them reliably).

VSadov:
@bartdesmet  I typically set environment variable   IsInterpreting=true  in the shell session I am in. 
Not the most user-friendly approach, but can be enough to at least minimally test interpreter.

src/System.Linq.Expressions/tests/Unary/UnaryIsFalseNullableTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Reflection;
+using Xunit;
+
+namespace Tests.ExpressionCompiler.Unary
+{
+    public static class UnaryIsFalseNullableTests
+    {
+        #region Test methods
+
+        [Fact] //[WorkItem(3196, "https://github.com/dotnet/corefx/issues/3196")]
```
bartonjs: All of these commented out WorkItem attributes are a bit strange (at least non-standard).

Unless the System.Linq.Expressions library has a test decoration style that the rest of the framework doesn't have, please remove the commented-out attributes throughout the change.

bartdesmet: Annotations per recommendation of @VSadov. I can use any existing annotation approach if there's a preferred way.

bartonjs: I'm pretty sure the existing way is undecorated :smile:.

@stephentoub thoughts/opinions?  (Personally, I don't like adornment in tests except for things like `/* Ensure things work when the input length is a prime number, see {reference} for context */`)

bartdesmet: Pieces of prior art with regards to referencing issues:
- Introduce a WorkItemAttribute a la Roslyn
- WinHttpHandlerTest.cs uses a comment with the URI
- Environment.GetEnvironmentVariable.cs uses ActiveIssueAttribute, not sure what will happen with the annotation after it gets fixed

bartonjs: [ActiveIssue] disables the test (due to the referenced issue), and the attribute's supposed to be deleted when the test would pass again :)

VSadov: Roslyn uses WorkItemAttribute as a standardized way for regression tests to refer to bugs. It has been found very useful in the past, since bug descriptions and discussions can contain a lot of relevant information, that would be hard to make a part of a test. 
Being an attribute makes it easy to find in an automated way if so needed (bugs migrated to alternative store or something)

Turns out Linq.Expression does not have any convention for regression tests. I suggested commented out attribute as a temporary measure to not lose the information about corresponding bug while not blocking the change on an infrastructure decision.

The plan is to get an agreement on these kind of adornments (most likely some kind of attribute) and search/replace the comments as a separate change. 

It seems better than to rely on natural text comments or not preserving this data at all.

VSadov: If CoreFX repo conventions are against such decorations, they will be removed. 
It is just easier to remove information, than add it after the fact.

stephentoub:
@dotnet-bot test this please

src/System.Linq.Expressions/tests/Unary/UnaryIsFalseNullableTests.cs
VSadov: Entered an issue to track this: https://github.com/dotnet/corefx/issues/3311

PR #3228, opened 2015-09-14
OverflowException for interpretation of NewArrayBounds

### LINKED ISSUE #3147: Expression interpreter - NewArrayBounds does not throw OverflowException
Consider the following fragment:

`Expression.Lambda(Expression.NewArrayBounds(typeof(int), Expression.Constant(0), Expression.Constant(-1))).Compile().DynamicInvoke()`

With the expression compiler, this throws OverflowException. With the expression interpreter, the call to Array.CreateInstance results in an ArgumentOutOfRangeException.

bartdesmet:
Unlike the vector array type creation logic, the multi-dimensional case didn't throw OverflowException in the interpreter while the expression compiler does.

This fixes #3147.


bartdesmet:

stephentoub:
cc: @VSadov

VSadov:
the changes look good to me, but we need some kind of tests.

stephentoub:
@dotnet-bot test this please

Issue #3197, opened 2015-09-12
Expression interpreter - Missing instruction for unary lifted operators


bartdesmet:
Consider the following expression:

Expression.Negate(Expression.Negate(Expression.Constant(TimeSpan.FromHours(1), typeof(TimeSpan?))))

This fails with IndexOutOfRangeException due to a missing EmitLoadLocal instruction prior to the EmitCall instruction in EmitUnaryMethodCall.

bartdesmet:
There's also something strange about the InstructionList's approach to keeping track of the current stack depth. Consider as an example EmitUnaryMethodCall, shown slightly simplified below:

EmitStoreLocal(temp.Index); // 0
EmitLoadLocal(temp.Index); // 1
EmitLoad(null, typeof(object)); // 2
EmitEqual(typeof(object)); // 1
EmitBranchFalse(notNull); // 0

EmitLoad(null, typeof(object)); // 1
EmitBranch(computed); // 1

MarkLabel(notNull); // 1
EmitLoadLocal(temp.Index); // 2 (suggested addition)
EmitCall(node.Method); // 2

MarkLabel(computed); // 2

The added comments reflect the stack depth as calculated by the InstructionList class. However, the line with MarkLabel(notNull) should have a current stack depth of 1 which it gets from the BranchFalse instruction's stack depth earlier (or generally all branches to the label should have the same stack depth). It seems it just keeps numbering. In the absence of the "suggested addition" line, the calculated stack depth will be 1, so everything seems right. In particular, the assert in CompileNoLabelPush doesn't fire:

Debug.Assert(_instructions.CurrentStackDepth == startingStackDepth + (expr.Type == typeof(void) ? 0 : 1), ...

With the suggested change, it fires noting that the stack depth is 2 where it is expected to be 1, but it seems that's just a result of miscalculating the stack depth in the presence of unconditional branches; it seems to me it should build a control flow graph and check that all incoming edges to a label have the same stack depth.

Finally, note that a single negation expression node (see repro above) is not sufficient to repro the problem. Due to the missing instruction, it pops the slot holding the local variable from the Data in the InterpretedFrame. It may be worth asserting that no pops occur beyond the Interpreter.LocalCount threshold in the Data array of the frame (or put a canary in between the stack and the locals for debug builds).

PR #3229, opened 2015-09-14
Expression interpretation of unary operators with lifted methods

bartdesmet:
This fixes issue #3197 but stumbles upon the problem described in #3197 with regards to calculating the stack depth in the InstructionList class, causing an assert to fail. If that is indeed the case (@VSadov, please chime in), we likely have to open another issue to track the stack depth calculation logic and get that in prior to committing this. Either way, I'm staging the suggested fix for this issue which gets post the reported repro (modulo the debug assert check).

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
@@ -1215,6 +1215,7 @@ namespace System.Linq.Expressions.Interpreter
                 _instructions.EmitBranch(computed);
 
                 _instructions.MarkLabel(notNull);
+                _instructions.EmitLoadLocal(temp.Index);
                 _instructions.EmitCall(node.Method);
 
                 _instructions.MarkLabel(computed);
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @VSadov

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/LightCompiler.cs
```diff
@@ -1215,6 +1215,7 @@ private void EmitUnaryMethodCall(UnaryExpression node)
                 _instructions.EmitBranch(computed);
```
VSadov: I think the code needs  _instructions.EmitPop(); after _instructions.EmitBranchFalse(notNull);

It may also be useful to use EmitCoalescingBranchhere similarly to how it is used in CompileCoalesceBinaryExpression. It is essentially a coalescing branch.

VSadov: Actually, I think this code must use the  EmitCoalescingBranch, just because of the stack behavior.

stephentoub:
@dotnet-bot test this please

VSadov:
Fixes #3197

Issue #3217, opened 2015-09-13
Expression interpreter - LoadField and StoreField don't generate NullReferenceException


bartdesmet:
Consider the following expression:

Expression.Field(Expression.Constant(null, typeof(Bar)), "Foo")

The expression interpreter will throw a TargetException rather than a NullReferenceException due to the use of GetValue (or SetValue for the corresponding assignment instruction).

Issue #2756, opened 2015-08-12
Need to find a way to run tests involving expression trees in interpreter mode


VSadov:
This involves:

Linq.Expressions
System.Dynamic.Runtime
Microsoft.CSharp

What complicates matters is that Linq.Expressions must be rebuilt with "isInterpreting=true" to switch from compiling to interpreting.

VSadov:
sInterpreting=true is sufficient for exploratory testing - to find bugs.

Also we have a public entry point for the interpreter. That should be enough for regression testing as well.

PR #3230, opened 2015-09-14
Throw NullReferenceException for field dereferences in expression interpreter

bartdesmet:
This fixes issue #3217.


stephentoub:
cc: @VSadov

VSadov:
@bartdesmet  code changes look good
in other projects (like Roslyn), we just add regular tests and add an attribute like 

```
        [Fact, WorkItem(4471, "https://github.com/dotnet/roslyn/issues/4471")]
```

I think we should start similar practice here.

VSadov:
Note that Jenkins will not run interpreter right now. I do not know yet what would be the best way to enable that. 
It is tracked by https://github.com/dotnet/corefx/issues/2756 
Need to do changes like this, raise the importance of that workitem.

stephentoub:
@dotnet-bot test this please

VSadov:
Fixes #3217

PR #3231, opened 2015-09-14
Fix for ConditionalExpression.Update undesired cloning behavior

### LINKED ISSUE #3223: ConditionalExpression - Undesirable cloning behavior in Update method
The Update method has the following logic:

if (test == Test && ifTrue == IfTrue && ifFalse == IfFalse)

The IfFalse property is implemented as follows for "IfThen" expressions (through a virtual GetFalse method):

return Expression.Empty();

This results in the creation of a new object each time. As a result, the Update method's logic will never cause the current instance to be returned, causing cloning behavior of the entire (sub)tree, e.g. during a visit. This turns out to be very expensive for large trees.

bartdesmet:
This fixes issue #3223 at the expense of a more compact object layout for the "IfThen" case without an Else branch. Note that this change does not attempt to reduce the number of copies of Expression.Empty created for IfFalse properties (this is done by the factory method). It solely avoids the cloning behavior of the entire ConditionalExpression upon the visitor's invocation of the Update method.

An alternative would be to virtualize the Update method and do a type check for the ifFalse expression that's provided, checking whether it's a DefaultExpression of type void. That'd enable us to keep the compact layout for IfThen cases but it'd suppress cloning of the tree even if another "copy" of the Empty expression is provided, effectively treating those copies as singletons. That's likely to be considered a breaking change (although a similar type check was present in the Make method to begin with, but IfFalse would return a different instance of Empty each time it's called).

Note: Should we add a RegressionTests.cs file with the name of test methods referring to the issues here in GitHub, or should we add the tests to the proper file (most of which seem machine-generated, so that may cause issues).

Fixes #3223

### DIFF
--- a/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
+++ b/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
@@ -115,6 +115,7 @@
     <Compile Include="System\Linq\Expressions\TryExpression.cs" />
     <Compile Include="System\Linq\Expressions\TypeBinaryExpression.cs" />
     <Compile Include="System\Linq\Expressions\UnaryExpression.cs" />
+    <Compile Include="System\Linq\Expressions\Utils.cs" />
     <Compile Include="System\Linq\IQueryable.cs" />
     <Compile Include="System\Runtime\CompilerServices\IRuntimeVariables.cs" />
     <None Include="project.json" />
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/ConditionalExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/ConditionalExpression.cs
@@ -5,6 +5,8 @@ using System;
 using System.Dynamic.Utils;
 using System.Diagnostics;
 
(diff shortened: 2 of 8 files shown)


stephentoub:
cc: @VSadov

VSadov:
Could it be better to just make default implementation of GetFalse to return a singleton Empty. We are changing behavior here in some way anyways. 
There might be such static singleton already somwhere in the codebase.

I'd argue that while factories do have to create new instances, the omitted children could default to singletons. - I.E. if the instance itself was cached in the factory, the caller would be in trouble if he wanted a new node instance for whatever reason. 
But if he wants children to be new instances he can simply pass them explicitly to the factory.

The current implementation tries to save a field. I have to assume it was based on analysis showing that such nodes are very common and saving a field has measurable impact on "bytes allocated" metric.

bartdesmet:
@VSadov, fair points. I'll go for a private singleton used for the default parameter of IfThen's else branch, therefore retaining the original field-saving approach.

stephentoub:
@dotnet-bot test this please

VSadov:
@bartdesmet - would it be possible to squash the commits?

PR #3234, opened 2015-09-14
Make OpenSsl-originated CryptographicExceptions set HResult and Message

### LINKED ISSUE #3232: Unix OpenSsl-based CryptographicExceptions should set both Message and HResult
Currently only Message is set, making it hard to reason about the error (in cases where it makes sense to do so).

This is the Unix half of #3062.

bartonjs:
Since CryptographicException has no constructor that takes both an HResult and a Message this is being achieved by a private derived type which can set both properties. (Fixes #3232.)

### DIFF
--- a/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
+++ b/src/Common/src/Interop/Unix/libcrypto/Interop.ERR.cs
@@ -71,7 +71,7 @@ internal static partial class Interop
 
             // If there was an error code, and it wasn't something handled specially,
             // use the OpenSSL error string as the message to a CryptographicException.
-            return new CryptographicException(ERR_error_string_n(error));
+            return new OpenSslCryptographicException(unchecked((int)error), ERR_error_string_n(error));
         }
 
         internal static void CheckValidOpenSslHandle(SafeHandle handle)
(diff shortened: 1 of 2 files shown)


bartonjs:
cc: @stephentoub, @nguerrera, @AtsushiKan

nguerrera:
LGTM.

nguerrera:
I didn't think about Assert.Throws being one of the "sad" checkers of exact type equality. I think ThrowsAny is fine, but I'm surprised there was only one instance to fix.

bartonjs:
@dotnet-bot test this please (seems to have gone into limbo due to a CI restart)

bartonjs:
@dotnet-bot test this please (reformatted PR description to not have newlines (in hopes it mitigates #3237))

AtsushiKan:
Do we want to get into testing exception message contents (especially in a way that doesn’t take in account localization)?

Otherwise, LGTM.

bartonjs:
@AtsushiKan OpenSSL doesn't appear to localize their error messages, so this should be stable.  A slightly more forgiving approach would be to ensure that it just isn't equal to the default error message for CryptographicException; but I'd like to be specific until it proves to be problematic.

Issue #3093, opened 2015-09-04
Add static method to create an IEnumerable with one element


benjamin-hodgson:
I find that I need the following static method quite often:

```csharp
public static class MoreEnumerableMethods
{
    // ...
    public static IEnumerable<T> Singleton<T>(T value)
    {
        return Enumerable.Repeat(value, 1);
    }
}
```

This seems like a really good fit for `Enumerable` itself, where it would live alongside `Repeat`, `Empty`, and `Range`. I want to be able to say `Enumerable.Singleton(x)`, not `MoreEnumerableMethods.Singleton(x)`.

Even Eric Lippert has [complained](http://ericlippert.com/2013/03/18/monads-part-eight/) about the lack of this method:
> I wish there was a static method on Enumerable specifically for making a one-element sequence.

I have the code (and tests) ready on [a branch in my fork](https://github.com/benjamin-hodgson/corefx/tree/singleton). I can open a pull request right away, subject to any required API reviews.

JonHanna:
I call the version I often use, `Emit`, because I think of calling it as an extension as "emitting" it into an enumerable. *Emit* has some other meanings of course, but perhaps not quite as strongly hit as *Singleton*.

I tend to have it as an extension thus:

    public static IEnumerable<T> Emit<T>(T element)
    {
      return Repeat(element, 1);
    }

JonHanna:
That said, having it as an extension already solves the irritation as to how it is called: Change yours to be an extension and then it's `x.Singleton()`.

benjamin-hodgson:
@JonHanna I actually prefer a static method to an extension in this case. For one, it's more cohesive with similar methods on `Enumerable` such as `Range` and `Repeat`. For another, putting a single element in a list doesn't feel to me like something you _do to the element_. It's a class-level operation of `Enumerable`, not an operation with which we want to extend an arbitrary object - the active party is the `Enumerable` class, not the object. I think it would look a bit strange with the extension-method syntax.

JonHanna:
You could say the same for every `AsXXX` method.

benjamin-hodgson:
Well, yes. I actually would have preferred `List.FromEnumerable` to `Enumerable.ToList`. But it's too late (and not important enough) to complain about that, those methods have already been written.

It seems pretty clear to me that for consistency with closely related methods like `Repeat`, it has to be a static method on `Enumerable`.

JonHanna:
[...]

PR #3235, opened 2015-09-14
#3093 Add Enumerable.Singleton method to produce a sequence with one element

benjamin-hodgson:
The issue over at #3093 doesn't seem to have attracted much attention, so I thought I'd open a PR with some code in it so we have something concrete to discuss :wink: 

For the tests, I simply copied and adjusted some of the tests for `Repeat()`.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -1494,6 +1494,11 @@ namespace System.Linq
             }
         }
 
+        public static IEnumerable<TResult> Singleton<TResult>(TResult element)
+        {
+            return Repeat(element, 1);
+        }
+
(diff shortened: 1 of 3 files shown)


benjamin-hodgson:
The failing test looks to be unrelated to my change. It's printing out the description of this pull request in the error message, which seems a little odd.

akoeplinger:
@benjamin-hodgson yeah, see #3237. Looks like the fix in #3102 wasn't enough, newlines seem to be a problem still.

stephentoub:
Thanks, @benjamin-bader.  Now that there's code available and linked to from the issue, I'm going to go ahead and close this PR.  We would need to take that issue through the full API review process, at which point even if we did decide to take the API, it would need to be merged into the future rather than master branch.

PR #3236, opened 2015-09-14
Add date time parse test

tarekgh:


tarekgh:
@stephentoub I did rebase but still seeing a couple of commits showing in my PR. I don't really know why this is happening. but the file changes is just my change only.

src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -327,6 +327,54 @@ public static void TestGetDateTimeFormats()
...
     [Fact]
+    public static void TestDateTimeParsingWithSpecialCultures()
+    {
+        // Test DateTime parsing with cultures which has the date separator and time separator are same
+        string[] cultureNames = new string[]
+        {
+            "fi-FI",
+            "nb-NO",
+            "nb-SJ"
+        };
```
Maxwe11: Why not to use `[Theory]`?

tarekgh: yes this can be a good idea too.

tarekgh:
@ellismg could you please have a quick look?

ellismg:
Code changes look fine, but I think we need to address the history issue.  Can you try the following?

```
> git checkout AddDateTimeParseTest
> git fetch upstream
> git reset --hard upstream/master
> git cherry-pick cd48895
> git cherry-pick 46d1d74
> git push origin +AddDateTimeParseTest
```

This will take your local copy off the AddDateTimeParseTest branch and reset it back so it matches what's in corefx/master.  Then the cherry-pick command will take your two changes and replay them on top of this new branch.

Finally, we push the new history to GitHub.

tarekgh:
@ellismg thanks Matt, looks your steps took care with the history issue.

PR #3239, opened 2015-09-14
Fixed OSX file creation time format

ianhays:
FileInfo.CreationTime on OSX was returning a correctly timed but incorrectly formatted DateTime. This commit fixes this issue and adds coverage to a existing test to fill the test gap that let the issue through.


ianhays:
cc: @Stephentoub

stephentoub:
LGTM

Issue #3233, opened 2015-09-14
Windows CNG/CAPI-based CryptographicExceptions should set both Message and HResult


bartonjs:
In CoreFx the library that CryptographicException is in is platform agnostic, so it doesn't load a code for the int-only constructor.

In debug builds the message should be of the form (`"(0x{0:X8}) {1}", hr, Interop.mincore.GetMessage(hr))`; later that might change to all builds if we really like it.

This is the Windows half of #3062.

bartonjs:
Windows doesn't have as nice a chokepoint as the Unix implementation; but something like what PR #3234 did would be good to do on Windows, as well.

PR #3240, opened 2015-09-14
Issue #3233 - More useful messages for Crypto exceptions.

AtsushiKan:
Windows version of #3062.

Note that SSC.X509Certificate is another candidate recipient but doing that separately as it requires changing dozens of callsites and I want to get synced up to the K changes first to avoid a merge mess.

### DIFF
--- a/src/Common/src/Internal/Cryptography/Windows/CryptoThrowHelper.cs
+++ b/src/Common/src/Internal/Cryptography/Windows/CryptoThrowHelper.cs
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Security.Cryptography;
+
+namespace Internal.Cryptography
+{
+    internal static class CryptoThrowHelper
+    {
+        public static CryptographicException ToCryptographicException(this int hr)
+        {
+            string message = Interop.mincore.GetMessage(hr);
+            return new WindowsCryptographicException(hr, message);
+        }
+
+        private sealed class WindowsCryptographicException : CryptographicException
+        {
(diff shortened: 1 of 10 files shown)


AtsushiKan:
cc @bartonjs

src/Common/src/Internal/Cryptography/CryptoThrowHelper.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Security.Cryptography;
+
+namespace Internal.Cryptography
+{
+    internal static class CryptoThrowHelper
```
bartonjs: Could you put this under Interop\Windows somewhere?  It's definitely Windows-specific, and I think it would be good to have that included in the path.

bartonjs:
Aside from the one one pathing comment, LGTM.  Thanks for jumping on it so quickly!

src/Common/src/Internal/Cryptography/CryptoThrowHelper.cs
AtsushiKan: Interop/[OS] is subdivided into (Win32) dll-based directories so I don’t see a place for it there as this helper isn’t tied to any Win32 dll.

Maybe under Internal\Cryptography\Windows, though this will no longer match the namespace then.

bartonjs: Internal\Cryptography\Windows is OK with me, as is the namespace mis-match (in X509 the PAL is Pal.Unix and Pal.Windows; but the namespace is just "Pal").

Or, if you wanted to put it in Interop\Windows\BCrypt that would make some sense (the CSP library forward-reaching for the errors helper seems fine to me)

PR #3251, opened 2015-09-15
Fix issue #3233. Better CryptoException messages.

AtsushiKan:


AtsushiKan:
cc @bartonjs

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Interop.crypt32.cs
```diff
@@ -11,276 +11,277 @@
...
 using X509KeyUsageFlags = System.Security.Cryptography.X509Certificates.X509KeyUsageFlags;
 
-namespace Internal.Cryptography.Pal.Native
+using Internal.Cryptography;
+using Internal.Cryptography.Pal.Native;
+
+internal static partial class Interop
 {
-    internal static partial class Interop
+    public static class crypt32
```
bartonjs: Should be declared partial, just in case :).

bartonjs:
On the one hand, I'm not sure why some reformatting of the interop type was in the same change; on the other hand putting ?w=1 (ignore whitespace) on the URL made all the noise drop out.

So, aside from the interop type's lack of being declared partial, LGTM.

AtsushiKan:
It wasn’t reformatting, it was hoisting the Interop type to the root namespace to avoid having two types named “Interop”.

bartonjs:
Fair enough, LGTM :)

PR #3255, opened 2015-09-15
Implement GetECDsaPublicKey/GetECDsaPrivateKey.

AtsushiKan:
- Unix implementations will come later.

- Can't update the ref assembly yet - ECDsa not yet in packages.

cc @bartonjs


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/CertificateExtensionsCommon.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Internal.Cryptography.Pal
+{
+    internal static class CertificateExtensionsCommon
+    {
+        public static A GetPublicKey<A>(this X509Certificate2 certificate) where A : AsymmetricAlgorithm
```
bartonjs: `A` as a generic type parameter is pretty weird.  Why not `T`?

AtsushiKan: “A” for “AsymmetricAlgorithm”

stephentoub: I suggest if we want to use something besides ```T```, we should do so because we want a descrive name, e.g. ```TAlgorithm```.

bartonjs: @stephentoub took the words right out from under my fingers.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/X509Pal.PublicKey.cs
```diff
@@ -51,6 +56,50 @@ public AsymmetricAlgorithm DecodePublicKey(Oid oid, byte[] encodedKeyValue, byte
...
+        private static ECDsa DecodeECDsaPublicKey(byte[] encodedKeyValue)
+        {
+            if (encodedKeyValue.Length < 1 || (AsnTag)(encodedKeyValue[0]) != AsnTag.OCTET_STRING)
+                throw new CryptographicException(SR.InvalidPublicKeyInX509);
+
+            // After the OCTET_STRING tag, the encryptedKeyValue are X, followed by Y, in big-endian format.
+            BCRYPT_ECCKEY_BLOB_MAGIC magic;
+            switch (encodedKeyValue.Length - 1)
+            {
+                case 2 * ((256 + 7) / 8):
```
stephentoub: What is this formula?  A comment and/or using a named constant would be helpful.

AtsushiKan: (<bits> + 7)/8 == bit to byte length conversion.
2* since there are two octets (X and Y).

A named constant would obscure the key bit-length which needs to match up with the key bit-length named in the blob magic constant.

stephentoub: Then at least a comment highlighting what you just wrote

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/X509Pal.PublicKey.cs
```diff
@@ -51,6 +56,50 @@ public AsymmetricAlgorithm DecodePublicKey(Oid oid, byte[] encodedKeyValue, byte
             }
         }
 
+        private static ECDsa DecodeECDsaPublicKey(byte[] encodedKeyValue)
+        {
+            if (encodedKeyValue.Length < 1 || (AsnTag)(encodedKeyValue[0]) != AsnTag.OCTET_STRING)
+                throw new CryptographicException(SR.InvalidPublicKeyInX509);
+
+            // After the OCTET_STRING tag, the encryptedKeyValue are X, followed by Y, in big-endian format.
```
bartonjs: Sort of... not really... it depends.
```
   SubjectPublicKeyInfo  ::=  SEQUENCE  {
        algorithm            AlgorithmIdentifier,
        subjectPublicKey     BIT STRING  }

   AlgorithmIdentifier  ::=  SEQUENCE  {
        algorithm               OBJECT IDENTIFIER,
        parameters              ANY DEFINED BY algorithm OPTIONAL  }
```

EC keys are stored very complicatedly/polymorphicly.  For Windows 8.1 we only support the parameter being the OID for P256, P384, or P521, and then the blob should be 0x04 (uncompressed key, unless it isn't...) then Qx then Qy.

But, really, why are you reading this at all?  Just import the whole public key subject info block into BCrypt and return a ECDsaCng.

We don't really want to have to write decoders for named uncompressed keys, named compressed keys, specified uncompressed keys, and specified compressed keys.

But this code is just assuming it's a named (and supported) uncompressed key, which is bad and dangerous.

AtsushiKan: That’s the answer I was looking for with that mail thread I sent out Friday about decoding the ASN1.

Putting out a PR is truly the most effective ping mechanism…

src/System.Security.Cryptography.X509Certificates/tests/TestData.cs
```diff
@@ -1073,6 +1073,66 @@ internal static class TestData
...
+            + "ff38e85881863b7128ddbb856d8ae158365550bb6330b93d4ef0300906072a8648ce3d040103818b0030818702420090bdf5"
+            + "dfb328501910da4b02ba3ccd41f2bb073608c55f0f2b2e1198496c59b44db9e516a6a63ba7841d22cf590e39d3f09636d0eb"
+            + "cd59a92c105f499e1329615602414285111634719b9bbd10eb7d08655b2fa7d7eb5e225bfdafef15562ae2f9f0c6a943a7bd"
+            + "f0e39223d807b5e2e617a8e424294d90869567326531bcad0f893a0f3a").HexToByteArray();
+    
+        internal struct ECDsaCngPublicKey
+        {
+            public byte[] QX;
+            public byte[] QY;
+            public byte[] D;
```
bartonjs: `D` is the private key; so it either shouldn't be included, or the name of the type should be something like `ECDsaCngKeyBlob`

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/CertificateExtensionsCommon.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Internal.Cryptography.Pal
+{
+    internal static class CertificateExtensionsCommon
```
bartonjs: I'm not seeing what value this set of generic methods provides; particularly since it fans back in to a switch on the generic type parameter...

AtsushiKan: It still captures common code (parameter validation, algorithm-independent behavior.)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/CertificateExtensionsCommon.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (!certificate.HasPrivateKey || oidValue != certificate.PublicKey.Oid.Value)
+                return null;
+
+            if (typeof(T) == typeof(RSA))
+                return (T)(object)certificate.Pal.GetRSAPrivateKey();
+
+            if (typeof(T) == typeof(ECDsa))
+                return (T)(object)certificate.Pal.GetECDsaPrivateKey();
+
+            throw new NotSupportedException(SR.NotSupported_KeyAlgorithm);
```
bartonjs: This probably could use a Debug.Fail, since the only way to reach this code is to update `GetExpectedOidValue` without updating this method, as well.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/CertificatePal.CspParametersStub.cs
```diff
@@ -17,7 +17,7 @@ private enum CspProviderFlags
         {
             // Note: UWP doesn't use any flags other than these, so they have been omitted.
             NoFlags = 0x0000,
-            UseMachineKeyStore = 0x0001,
+            UseMachineKeyStore = 0x0001,
```
bartonjs: We probably don't need that space at the end of the line :)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/CertificatePal.PrivateKey.cs
```diff
@@ -43,13 +79,13 @@ public RSA GetRSAPrivateKey()
 #if NETNATIVE
                 // In .NET Native (UWP) we don't have access to CAPI, so it's CNG-or-nothing.
                 // But we don't expect to get here, so it shouldn't be a problem.
-
+
```
bartonjs: Nit: bad whitespace

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/Native/Interop.crypt32.cs
```diff
@@ -8,9 +8,11 @@
 
 using CryptographicException = System.Security.Cryptography.CryptographicException;
 
+using SafeBCryptKeyHandle = Microsoft.Win32.SafeHandles.SafeBCryptKeyHandle;
 using SafeX509ChainHandle = Microsoft.Win32.SafeHandles.SafeX509ChainHandle;
 using X509KeyUsageFlags = System.Security.Cryptography.X509Certificates.X509KeyUsageFlags;
 
+
```
bartonjs: Nit: bad whitespace

src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs
```diff
@@ -125,5 +126,49 @@ public static void TestKey_RSA()
...
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public static void TestKey_ECDsaCng521()
+        {
+            TestKey_ECDsaCng(TestData.ECDsa521Certificate, TestData.ECDsaCng521PublicKey);
+        }
+
+        private static void TestKey_ECDsaCng(byte[] certBytes, TestData.ECDsaCngKeyValues expected)
+        {
+#if !NETNATIVE
```
bartonjs: I actually don't know if this will make the test system happy or not.  If not, we'll find out eventually :).

bartonjs:
Some nitpicky things, otherwise LGTM.

The ref/contract changes will have to come from the TFS side still, because they need to be updated in both the old place and the new for transitional reasons...

PR #3257, opened 2015-09-15
Finishing up PAL'ing Unix\libc (non-networking and crypto) PInvoke calls

sokket:
Finishing up PAL'ing Unix\libc (non-networking and crypto) PInvoke calls

/c @nguerrera


sokket:
@dotnet-bot test this please

Looks like a Windows test failure around a string comparison...will try to test again

src/Common/src/Interop/Unix/System.Native/Interop.PathConf.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal static int DEFAULT_PC_NAME_MAX = 255;
+
+        internal enum PathConfNames : int
```
nguerrera: This should be PathConfName singular since these aren't flags.

src/Native/System.Native/pal_process.cpp
```diff
@@ -346,3 +356,75 @@ int32_t WTermSig(int32_t status)
 {
     return WTERMSIG(status);
 }
+
+static
+int32_t ConvertPathConfNameToPlatform(PathConfNames& name)
+{
+    switch (name)
```
nguerrera: I'd inline this into PathConf as I did with SysConf, we don't need to convert these for any other reason.

src/Native/System.Native/pal_process.cpp
```diff
@@ -346,3 +356,75 @@ int32_t WTermSig(int32_t status)
...
+{
+    // The POSIX.1-2001 standard allows passing a nullptr buffer and having the 
+    // OS allocate enough space for the path; use this to get the working directory
+    // without having to allocate and reallocate to get the right size buffer. We 
+    // can then pass the C# code a pointer to the start of the buffer and they can
+    // copy the string to managed code before we free it.
+    char* bufferStart = getcwd(nullptr, 0);
+    wdCallback(bufferStart);
+    free(bufferStart);
+}
```
nguerrera: The old managed code had a fast path for stackallocable size. It's probably worth replicating that here.

src/Native/System.Native/pal_process.cpp
```diff
@@ -346,3 +356,75 @@ int32_t WTermSig(int32_t status)
 {
     return WTERMSIG(status);
 }
+
+static
+int32_t ConvertPathConfNameToPlatform(PathConfNames& name)
```
nguerrera: const PathConfName&

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        private unsafe delegate void FoundWorkingDirctory(byte* bufferStart);
```
justinvp: Typo: "Dirctory" -> "Directory"

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        private static extern void GetCwd(FoundWorkingDirctory foundWorkingDirectory);
+
+        internal unsafe static string GetCwd()
+        {
+            string cwd = string.Empty;
+            GetCwd((byte* bufferStart) => cwd = Marshal.PtrToStringAnsi((IntPtr)bufferStart));
+            return cwd;
```
justinvp: This has two managed heap allocations: one for the delegate and another for the closure object. It'd be nice if there was a fast path that avoided unnecessary managed allocations, like the previous managed stackalloc approach.

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
```diff
@@ -107,9 +107,6 @@
     <Compile Include="$(CommonPath)\Interop\Linux\Interop.Libraries.cs">
       <Link>Common\Interop\Linux\Interop.Libraries.cs</Link>
```
nguerrera: I think we can remove the OS-specific Interop.Libraries now since I don't see anything else that would P/Invoke to them. 

(That leaves only PathInternal.CaseSensitive/Insensitive preventing this from having a single unix implementation, which we can consider deciding at runtime. Fine to leave that to a follow-up change, though.)

src/Native/System.Native/pal_process.cpp
nguerrera: Are you sure about this being part of the POSIX standard? The Linux man page as I read it says that it's a Linux extension to POSIX.1-200 and I don't see it documented here: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getcwd.html#tag_16_190

nguerrera: OS X man page does document this too, though. I'm just not sure if it's guaranteed to work everywhere.

nguerrera: At the bottom of the POSIX link above, it calls out this behavior as being "on some implementations" and something that was only considered for standardization. It may be the case that every OS we care about has this behavior (I'm guessing OS X inherits this from BSD lineage). Ideally, we'd test with CMAKE. We could fall back to erroring out if the test passes on FreeBSD, Linux, and OS X.

src/Native/System.Native/pal_process.cpp
```diff
@@ -6,6 +6,11 @@
 
 #include <assert.h>
 #include <errno.h>
+#if HAVE_LINUX_PATH_MAX
+#include <linux/limits.h>
+#else
+#include <iostream>
```
nguerrera: mentioned offline, but this should be limits.h h if not linux/limits.h.

nguerrera: Hmm, plain limits.h works on Linux too (at least on my machine) and it is supposed to be there according to http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

src/Native/System.Native/pal_process.cpp
stephentoub: If we keep this new version, then as @justinvp pointed out on the managed code, we're incurring now non-trivial costs to allocate the delegate/closure on each invocation.  Since this is documented to return malloc'd memory, why not just free it on the managed side (or let the runtime do it as part of marshaling)?  That said, that this isn't standardized makes me nervous, and we should consider going back to the original code.

src/Common/src/Interop/Unix/System.Native/Interop.GetSetPriority.cs
```diff
@@ -28,26 +28,13 @@ internal static partial class libc
         /// getpriority call to act more naturally where the return value is the actual error
         /// value (or 0 if success) instead of forcing the caller to retrieve the last error.
         /// </summary>
-        /// <param name="which"></param>
-        /// <param name="who"></param>
-        /// <param name="priority"></param>
-        /// <returns></returns>
-        internal static int getpriority(PriorityWhich which, int who, out int priority)
+        /// <returns>Returns 0 on success; otherwise, returns -1</returns>
```
stephentoub: The function is returning Marshal.GetLastWin32Error()

src/Common/src/Interop/Unix/System.Native/Interop.PathConf.cs
```diff
@@ -0,0 +1,63 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal static int DEFAULT_PC_NAME_MAX = 255;
```
stephentoub: DEFAULT_PC_PATH_MAX was removed and put behind GetMaximumPath.  The same thing doesn't need to be done for DEFAULT_PC_NAME_MAX?

stephentoub:
Other than the issues called out, LGTM.  Good to see us almost through the non-networking/crypto P/Invokes.

src/Native/System.Native/pal_process.cpp
nguerrera: If you decide not to inline (I don't mind either way), it would be prudent not to pass that -1 on to pathconf in the error case. I'd put a check and error out as we do for the memory map options for example.

src/Common/src/Interop/Unix/System.Native/Interop.PathConf.cs
sokket: Fixed

src/Native/System.Native/pal_process.cpp
sokket: Cool, will do

src/Native/System.Native/pal_process.cpp
```diff
@@ -344,3 +354,75 @@ int32_t WTermSig(int32_t status)
 {
     return WTERMSIG(status);
 }
+
+static
+int32_t ConvertPathConfNameToPlatform(PathConfName& name)
```
Maxwe11: there is no sense to pass `name` as a reference

src/Native/System.Native/pal_process.cpp
sokket: Moved back to the previous way

src/Native/System.Native/pal_process.cpp
sokket: Inlined it :)

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
sokket: Removed for the other method

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
sokket: Moved back to the old approach

src/System.IO.FileSystem.DriveInfo/src/System.IO.FileSystem.DriveInfo.csproj
sokket: Good idea

src/Native/System.Native/pal_process.cpp
sokket: Changed to limits.h

src/Common/src/Interop/Unix/System.Native/Interop.GetSetPriority.cs
sokket: Fixed the comment to accurately describe the return value

src/Common/src/Interop/Unix/System.Native/Interop.PathConf.cs
sokket: The only reason the Max Path default value needed to be pushed into the PAL was because the PATH_MAX value changes between OS X, Linux, and other *nix flavors that we support. The PC Name Max does not look to have different default values (from our original implementation) so it does not need to be pushed down into the PAL and we can use the platform-agnostic 255 value

src/Native/System.Native/pal_process.cpp
sokket: Agreed :) I deleted the whole function anyway

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
```diff
@@ -4,21 +4,22 @@
...
+        internal static unsafe string GetCwd()
+        {      
+            const int StackLimit = 256;
+       
             // First try to get the path into a buffer on the stack
             byte* stackBuf = stackalloc byte[StackLimit];
             string result = getcwd(stackBuf, StackLimit);
             if (result != null)
-            {
+            {
```
justinvp: nit: trailing whitespace

sokket:
@dotnet-bot test this please

sokket:
Any further comments folks? @nguerrera @stephentoub

src/Common/src/Interop/Unix/System.Native/Interop.GetCwd.cs
```diff
@@ -4,27 +4,28 @@
...
+        {      
+            const int StackLimit = 256;
+       
             // First try to get the path into a buffer on the stack
             byte* stackBuf = stackalloc byte[StackLimit];
-            string result = getcwd(stackBuf, StackLimit);
+            string result = GetCwdHelper(stackBuf, StackLimit);
             if (result != null)
-            {
+            {
```
nguerrera: nit: trailing whitespace

sokket: Fixed

src/Native/System.Native/pal_process.cpp
```diff
@@ -323,3 +329,74 @@ extern "C" int32_t WTermSig(int32_t status)
...
+            break;
+        case PathConfName::PAL_PC_VDISABLE:
+            confValue = _PC_VDISABLE;
+            break;
+    }
+
+    if (confValue == -1)
+    {
+        assert(false && "Unknown PathConfName");
+        return EINVAL;
```
nguerrera: This should return -1 and set errno to EINVAL to mimic the semantics of underlying API.

sokket: Good catch, fixed

jasonwilliams200OK:
@sokket, by any chance, does this happen to fix https://github.com/dotnet/corefx/issues/2988?

src/Native/System.Native/pal_process.cpp
```diff
@@ -53,6 +54,11 @@ static_assert(PAL_LOG_LOCAL5 == LOG_LOCAL5, "");
 static_assert(PAL_LOG_LOCAL6 == LOG_LOCAL6, "");
 static_assert(PAL_LOG_LOCAL7 == LOG_LOCAL7, "");
 
+// Validate that out PriorityWhich values are correct for the platform
+static_assert(static_cast<int>(PriorityWhich::PAL_PRIO_PROCESS) == PRIO_PROCESS, "");
```
nguerrera: I made our enums not have the 'class' designation so that we can get rid of a bunch of casts like this: i.e. enum Foo : T instead of enum class Foo : T. Please fix the new ones up to match.

sokket:
@jasonwilliams200OK I don't think so since this doesn't shim anything in System.Net yet; this could be fixed with my next check-in though :). I'm moving on to shim the System.Net calls that are currently in Master next, so I'll assign that bug to me and make sure that the shim code for System.Net compiles on my FreeBSD box as well

src/Native/System.Native/pal_process.cpp
```diff
@@ -323,3 +329,74 @@ extern "C" int32_t WTermSig(int32_t status)
...
+    }
+
+    return result;
+}
+
+extern "C" int32_t GetPriority(PriorityWhich which, int32_t who)
+{
+    // GetPriority uses errno 0 to show succes to make sure we don't have a stale value
+    errno = 0;
+    return getpriority(static_cast<int>(which), static_cast<id_t>(who));
```
nguerrera: Note that you also won't need to cast `which` here after changing from enum class to enum. 

As for id_t, are there any range concerns here?

src/Native/System.Native/pal_process.h
```diff
@@ -180,3 +210,41 @@ extern "C" int32_t WIfExited(int32_t status);
...
+ * This is called out explicitly, rather than using PathConf, since the default value changes depending on the platform.
+ */
+extern "C" int64_t GetMaximumPath();
+
+/**
+ * Gets the priority (nice value) of a certain execution group.
+ *
+ * Returns the nice value (from -20 to 20) of the group on success; otherwise, returns -1. Unfortunately, -1 is also a
+ * valid
+ * nice value, meaning we can't use that value to determine valid output or not. Errno is set on failure so we need to
```
nguerrera: Some weird newlines here. Did clang_fomat do that?

src/Native/System.Native/pal_process.cpp
sokket: Fixed

src/Native/System.Native/pal_process.h
sokket: Yup, looks like it. Fixed

src/Native/System.Native/pal_process.cpp
```diff
@@ -323,3 +329,74 @@ extern "C" int32_t WTermSig(int32_t status)
...
+}
+
+extern "C" int32_t SetPriority(PriorityWhich which, int32_t who, int32_t nice)
+{
+    return setpriority(static_cast<int>(which), static_cast<id_t>(who), nice);
+}
+
+extern "C" char* GetCwd(char* buffer, uint64_t bufferSize)
+{
+    return getcwd(buffer, bufferSize);
```
nguerrera: This is going to overflow and warn as error on 32-bit platforms. Make bufferSize int32_t, assert (and real error for paranoid defensiveness) if < 0, then pass UnsigedCast(bufferSize). See examples/rationale in 648c01d69fba8d23750122d9823f53257785c8ca

src/Native/System.Native/pal_process.cpp
sokket: Nope; it's more because id_t is declared as unsigned but the c# base class declares it as signed, so we need a conversion when sending back to the OS here

nguerrera:
LGTM for the most part, but I left a few more comments.

src/Native/System.Native/pal_process.cpp
sokket: Good catch. Fixed

sokket:
Thanks folks! Double checking that the build passes and then will merge

PR #3259, opened 2015-09-15
Update to BuildTools version 87

weshaggard:
This update fixes the test-runtime\project.json so that it correctly
restores System.Console and also switches GenFacades to be running
on .NET Core isntead of the full framework.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00085</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00087</BuildToolsVersion>
     <DnxVersion>1.0.0-beta5-12101</DnxVersion>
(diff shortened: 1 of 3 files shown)


weshaggard:
@dotnet-bot test this please

stephentoub:
LGTM

weshaggard:
This failed with a known failure that @stephentoub  is fixing with PR https://github.com/dotnet/corefx/pull/3261.

stephentoub:
> This failed with a known failure that @stephentoub is fixing with PR #3261.

We shouldn't have merged this until it passed successfully.  I'm concerned it actually broke the Linux tests, which it hadn't even gotten to yet since it doesn't run those until it passes the Windows ones.

weshaggard:
@stephentoub Sorry about the break it should be addressed now by #3267. I didn't realize we gated the builds in this manor, I expected that all errors would have been shown. At any rate while fixing the linux test runs I learned more about how they currently work which will be useful for some upcoming work that I'm looking into.

stephentoub:
Thanks, Wes!

PR #3263, opened 2015-09-16
Update UnixFileStream to read more from OS after using buffer

stephentoub:
Stream.Read is not required to hand back as many bytes as the caller requested; a conforming implementation can hand back any number > 0 and <= count when there's still data left in the stream, and many streams do frequently return < count.  UnixFileStream currently frequently does this, returning what's left in the internal buffer when there's any data left in the buffer.

Win32FileStream, however, in this case does a subsequent read from the underlying file handle to try to give back as much data as the user requested.  This request still may not return the requested count, either because we reached the end of the file or because Windows decided not to return as much data as was requested, but it frequently will.  As such, some code has been written to expect this behavior, and while that code is flawed and should be fixed even on Windows, it frequently works.

To maintain the consistency with Windows, this change updates UnixFileStream to do such a subsequent read if we tried to satisfy a Read request purely from the buffer but didn't have as much data as the user requested.  I've not added a test for the behavior, because a) it's not guaranteed by Stream.Read, but more importantly b) the test could fail because the OS may not give us back what we requested.

cc: @ianhays, @sokket

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
@@ -618,6 +618,7 @@ namespace System.IO
             // we can just go directly into the user's buffer, if they asked
             // for more data than we'd otherwise buffer.
             int numBytesAvailable = _readLength - _readPos;
+            bool readFromOS = false;
             if (numBytesAvailable == 0)
             {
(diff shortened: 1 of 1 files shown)


stephentoub:
@dotnet-bot test this please

nguerrera:
LGTM. (I'm a little sad that we have to bother with this, but I do see the value in being closer to the Windows implementation.)

stephentoub:
Yeah.

PR #3264, opened 2015-09-16
DoNotMerge: Revert "Update to BuildTools version 87"

bartonjs:
This reverts commit f2d4efc72a78f4d0a4a35d4129b16d790069e5a5. (seeing if the Linux builds get better)

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00087</BuildToolsVersion>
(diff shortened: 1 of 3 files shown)


bartonjs:
FYI @stephentoub and @weshaggard, I seem to be the only one passing, because I reverted that change...

If someone wants to merge this in after sleeping and having some coffee, just strip the DoNotMerge from the title :).

weshaggard:
I believe I've addressed the issue with the buildtools update with PR #3267, so this can be closed.

stephentoub:
Thanks, Jeremy.

PR #3265, opened 2015-09-16
Increase native compilation warnings and fix the fallout

### LINKED ISSUE #3165: Increase warning level for native code, add range checks
We currently build native code with -Wall -Werror, but that doesn't include things like -Wconversion which would catch truncating conversions.

nguerrera:
We were using -Wall but now use -Weverything and opt out of only a handful of issues such as C++-98 compat, which we don't need.

The most notable impact is that the compiler now catches narrowing implicit conversions, which was also the primary motivation for doing this now as there was known debt in our unsupervised usage of 64-bit types for sizes.

We now assert on bad buffer sizes *and* retain full checks in release builds to be completely defensive against overflow.

We now use int32_t for scratch buffer sizes. int64_t remains for file offsets and uint64_t remains for memory mapped files, but int32_t for scratch eliminates a bunch of managed casts and therefore makes it easier to reason about overflow prevention.

INT32_MAX is also the most portable limit on these buffer sizes because the behavior of functions like read is only implementation defined when the input size is greater than SSIZE_MAX and ssize_t is int32_t on the lowest common denominator of modern platforms.

Enums corresponding to platform values are now enum X : T instead of enum class X : T. The PAL_* names are descriptive enough without scoping and the implicit conversion to raw integer is actually helpful to avoid a bunch of casts, which are otherwise needed to interop with the corresponding native API.

Fix #3165

cc @stephentoub @sokket


jasonwilliams200OK:

stephentoub:
@dotnet-bot test this please

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -339,17 +340,24 @@ GetAsn1StringBytes(
     unsigned char* pBuf,
     int cBuf)
 {
-    if (!asn1)
+    if (!asn1 || cBuf < 0)
```
stephentoub: Should we be asserting in these cases?

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -992,6 +1013,8 @@ static
 void
 LockingCallback(int mode, int n, const char* file, int line)
 {
+    (void)file, (void)line; // deliberately unused parameters
```
stephentoub: We could use these in the assert if you wanted.

stephentoub:
LGTM

src/Native/System.Security.Cryptography.Native/openssl.c
nguerrera: This would still be needed since they'd be unused in release. Also, standard assert doesn't give us a way to format messages.

src/Native/System.Security.Cryptography.Native/openssl.c
nguerrera: I think so, but invalid args are handled without asserting throughout this file. I'm looking at crypto shims now and I'll clean this up as part of that.

nguerrera:
(NOTE: The only change in rebase was to wording of commit.)

nguerrera:
(NOTE: The only change in rebase was to wording of commit.)

PR #3267, opened 2015-09-16
Add exclusion for ToolRuntime directory in the linux test script

weshaggard:
This should take care of the current linux tests running failing. The reason they were failing is because the run-test script was finding the windows versions of some of the libraries in the new ToolRuntime directory and it was overlaying those over top the linux overlay.


weshaggard:
CI is green so merging this PR.

PR #3276, opened 2015-09-16
Update EventSource public interface for EventCounter hooks

cshung:
This change reacts to the EventSource public interface changes in CoreCLR for EventCounter hooks.


stephentoub:
LGTM

PR #3279, opened 2015-09-16
Adding a clang-format file and a script to format every cpp and h file

### LINKED ISSUE #3051: Create a C++ coding standard for the PAL layer
Create a C++ code style document describing coding standards and naming standards to use the PAL layer. Also, fix up the current PAL code to adhere to the coding standard

sokket:
Adding a clang-format file and a script to format every cpp and h file. This style is based on the LLVM style with some minor tweaks. This commit also ran the script against the current code to bring us into compliance. This fixes #3051

/cc @nguerrera @stephentoub

### DIFF
--- a/src/Native/.clang-format
+++ b/src/Native/.clang-format
@@ -1,0 +1,18 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AlignEscapedNewlinesLeft: false
+AlignAfterOpenBracket: true
+AllowShortFunctionsOnASingleLine: false
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBraces: Allman
+ColumnLimit:     120
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+IndentCaseLabels: true
+IndentWidth: 4
+PointerAlignment: Left
+TabWidth:    4
+...
+
--- a/src/Native/System.Native/pal_errno.h
+++ b/src/Native/System.Native/pal_errno.h
@@ -26,94 +26,94 @@
  */
 enum Error : int32_t
 {
-    PAL_SUCCESS          = 0,
+    PAL_SUCCESS = 0,
 
-    PAL_E2BIG            = 0x10001,           // Argument list too long.
-    PAL_EACCES           = 0x10002,           // Permission denied.
-    PAL_EADDRINUSE       = 0x10003,           // Address in use.
-    PAL_EADDRNOTAVAIL    = 0x10004,           // Address not available.
-    PAL_EAFNOSUPPORT     = 0x10005,           // Address family not supported.
-    PAL_EAGAIN           = 0x10006,           // Resource unavailable, try again (same value as EWOULDBLOCK),
-    PAL_EALREADY         = 0x10007,           // Connection already in progress.
-    PAL_EBADF            = 0x10008,           // Bad file descriptor.
-    PAL_EBADMSG          = 0x10009,           // Bad message.
-    PAL_EBUSY            = 0x1000A,           // Device or resource busy.
-    PAL_ECANCELED        = 0x1000B,           // Operation canceled.
-    PAL_ECHILD           = 0x1000C,           // No child processes.
-    PAL_ECONNABORTED     = 0x1000D,           // Connection aborted.
-    PAL_ECONNREFUSED     = 0x1000E,           // Connection refused.
-    PAL_ECONNRESET       = 0x1000F,           // Connection reset.
-    PAL_EDEADLK          = 0x10010,           // Resource deadlock would occur.
-    PAL_EDESTADDRREQ     = 0x10011,           // Destination address required.
-    PAL_EDOM             = 0x10012,           // Mathematics argument out of domain of function.
-    PAL_EDQUOT           = 0x10013,           // Reserved.
-    PAL_EEXIST           = 0x10014,           // File exists.
-    PAL_EFAULT           = 0x10015,           // Bad address.
-    PAL_EFBIG            = 0x10016,           // File too large.
-    PAL_EHOSTUNREACH     = 0x10017,           // Host is unreachable.
-    PAL_EIDRM            = 0x10018,           // Identifier removed.
-    PAL_EILSEQ           = 0x10019,           // Illegal byte sequence.
-    PAL_EINPROGRESS      = 0x1001A,           // Operation in progress.
-    PAL_EINTR            = 0x1001B,           // Interrupted function.
-    PAL_EINVAL           = 0x1001C,           // Invalid argument.
-    PAL_EIO              = 0x1001D,           // I/O error.
-    PAL_EISCONN          = 0x1001E,           // Socket is connected.
-    PAL_EISDIR           = 0x1001F,           // Is a directory.
-    PAL_ELOOP            = 0x10020,           // Too many levels of symbolic links.
-    PAL_EMFILE           = 0x10021,           // File descriptor value too large.
-    PAL_EMLINK           = 0x10022,           // Too many links.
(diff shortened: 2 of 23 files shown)


src/Native/.clang-format
```diff
@@ -0,0 +1,16 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AlignEscapedNewlinesLeft: false
+AllowShortFunctionsOnASingleLine: false
+AlwaysBreakTemplateDeclarations: true
+BinPackParameters: false
+BreakBeforeBraces: Allman
+ColumnLimit:     150
```
sokket: I picked 150 chars since that was big enough to fit on the most common screen resolution (1366x768) but okay changing it if there is pushback against this

src/Native/System.Native/pal_io.cpp
```diff
@@ -19,15 +19,14 @@
 #include <syslog.h>
 #include <unistd.h>
 
-
 #if HAVE_STAT64
-#    define stat_ stat64
-#    define fstat_ fstat64
-#    define lstat_ lstat64
```
nguerrera: Is there an option for this sort of preprocessor indenting. I find it more readable, but no biggie if there's no option.

src/Native/System.Native/pal_io.cpp
```diff
@@ -337,43 +316,39 @@ int32_t ReadDirR(DIR* dir, void* buffer, int32_t bufferSize, DirectoryEntry* out
...
+            assert(false && "Unknown flag.");
+            errno = EINVAL;
+            return -1;
     }
 
 #if HAVE_PIPE2
     return pipe2(pipeFds, flags);
 #else
-    return pipe(pipeFds); // CLOEXEC intentionally ignored on platforms without pipe2.
+    return pipe(pipeFds);         // CLOEXEC intentionally ignored on platforms without pipe2.
```
nguerrera: I don't understand why it did this.

src/Native/System.Native/pal_mount.h
```diff
@@ -25,31 +25,22 @@ typedef void (*MountPointFound)(const char* name);
...
  * Since C# is much better at enum and string handling, pass either the char buffer or the long type
  * back, depending on what the platform gives us, and let C# reason on it in an easy way.
  */
-extern "C"
-int32_t GetFormatInfoForMountPoint(
-    const char* name,
-    char*       formatNameBuffer,
-    int32_t     bufferLength,
-    int64_t*    formatType);
+extern "C" int32_t GetFormatInfoForMountPoint(const char* name, char* formatNameBuffer, int32_t bufferLength, int64_t* formatType);
```
nguerrera: GitHub is forcing me to scroll at about the f in formatType. I suggest we use that as the max line length --> 120

src/Native/System.Native/pal_io.cpp
sokket: LLVM style sets the SpacesBeforeTrailingComments option at a value so all trailing comments will be preceded by X number of spaces

src/Native/System.Native/pal_io.cpp
sokket: I couldn't find one :( sorry

src/Native/System.Native/pal_process.h
```diff
@@ -16,49 +16,45 @@
...
-    int32_t redirectStdin,    // whether to redirect standard input from the parent
-    int32_t redirectStdout,   // whether to redirect standard output to the parent
-    int32_t redirectStderr,   // whether to redirect standard error to the parent
-    int32_t* childPid,        // [out] the child process' id
-    int32_t* stdinFd,         // [out] if redirectStdin, the parent's fd for the child's stdin
-    int32_t* stdoutFd,        // [out] if redirectStdout, the parent's fd for the child's stdout
-    int32_t* stderrFd);       // [out] if redirectStderr, the parent's fd for the child's stderr
-
+extern "C" int32_t ForkAndExecProcess(const char* filename,   // filename argument to execve
+                                      char* const argv[],     // argv argument to execve
```
nguerrera: Is there an option to put parameters as they were above when on more than one line. It saves a lot of horizontal real estate and causes less churn if function is renamed?

src/Native/System.Native/pal_mount.h
sokket: Cool, works for me

nguerrera:
The tool is very cool and I'm really glad we'll have this rather than discussing individual preferences on individual PRs. :+1: 

A few minor notes, but LGTM otherwise. If there aren't options for the things I asked about, then I'm fine going with what the tool does by default..

src/Native/System.Native/pal_io.cpp
nguerrera: No worries. We should format config.h.in as well to be consistent.

src/Native/System.Native/pal_io.cpp
nguerrera: Is this the only line that would change if we set it to 1?

jasonwilliams200OK:
@sokket, regarding the code format preferences, would be nice to have [.editorconfig](https://github.com/dotnet/corefx/blob/master/.editorconfig) file updated accordingly for C/C++ source/headers (a rule for `{.hpp,.h,.cpp,.c}`).

src/Native/System.Native/pal_io.cpp
sokket: Added another loop to find *.in files in case we add more

src/Native/System.Native/pal_io.cpp
sokket: Not entirely sure. Do you want this value to be 1?

src/Native/System.Native/pal_process.h
sokket: I couldn't find one...sorry

sokket:
@jasonwilliams200OK Added the brace style and indent brace style in the editorconfig. It doesn't support most of what we need but I added what I could :)

src/Native/System.Native/pal_errno.h
```diff
@@ -26,94 +26,94 @@
...
-    PAL_EROFS            = 0x10048,           // Read-only file system.
-    PAL_ESPIPE           = 0x10049,           // Invalid seek.
-    PAL_ESRCH            = 0x1004A,           // No such process.
-    PAL_ESTALE           = 0x1004B,           // Reserved.
-    PAL_ETIMEDOUT        = 0x1004D,           // Connection timed out.
-    PAL_ETXTBSY          = 0x1004E,           // Text file busy.
-    PAL_EXDEV            = 0x1004F,           // Cross-device link.
+    PAL_E2BIG = 0x10001,           // Argument list too long.
+    PAL_EACCES = 0x10002,          // Permission denied.
+    PAL_EADDRINUSE = 0x10003,      // Address in use.
```
akoeplinger: Can you try adding `AlignConsecutiveAssignments: true`? It is a new option in clang-format 3.7 that should restore the previous (IMHO more readable) style where everything is aligned at the =.

sokket: I went back and forth about this one...I decided against it because this flag is only available on 3.7+ but clang-format-3.6 is the default on Ubuntu 14 and for the version of llvm that we currently support. Once 3.7 and 3.8 are common place we can upgrade and add that flag, but for now I think we should keep it out so we can stay with the most-easily-accessible version

akoeplinger: Ah good point, makes sense :)

src/Native/System.Native/pal_io.cpp
nguerrera: I do, but my care factor is not high. :)

sokket:
@dotnet-bot  test this please

sokket:
@dotnet-bot test this please

Looks like an HTTP timeout

sokket:
Thanks folks! Merging now

PR #3286, opened 2015-09-17
Update to build tools version 89

weshaggard:
Moves us to xunit 2.1 rc1
Refreshs the lock files to include latest versions so we are now
compatible with DNU beta7 format.


stephentoub:
LGTM

sokket:
LGTM

Issue #3266, opened 2015-09-16
Use optimized N-ary MethodCallExpression nodes from all Call factories


bartdesmet:
Unlike Expression.Block, the overloads of Expression.Call that take in an Expression[] or IEnumerable<Expression> do not attempt to check the number of supplied arguments in order to instantiate more compact expression representations than the standard [Instance]MethodCallExpressionN nodes.

The language binds to the Expression[] overload. Various tools on our end, such as expression deserializers, also dispatch through the IEnumerable<Expression> overload.

It'd be good for those factories to analyze the number of supplied arguments and use compact representations if available. Also, [Instance]MethodCallExpression0 nodes are missing, even though it's quite common to have such expressions (e.g. ToString). Those could be added as part of this.

Note that - in addition to this - it may be possible for the language compilers to check for the N-ary factory method overloads and bind to more specialized overloads when it finds them in the corefx libraries it's compiling against (much like other places where the compiler uses "feature detection"). This would avoid the Expression[] allocation when it's not needed, therefore reducing some allocation costs.

bartdesmet:
I can prepare a PR with proposed improvements.

VSadov:
As I remember Expression.Block was new in ET2.0 and factory was implemented this way. The Call (and Invoke) were not new, so it could be that they were intentionally not switched to use this pattern. It could also be just an oversight. 

Is it practically likely that the change will be breaking to some existing code?

bartdesmet:
I’ll have a thorough look from a compatibility angle and report any findings.

One point of debate is whether public factory method additions are warranted, e.g. for Invoke(lambda, arg0, …, argN). If not, the optimization could always be done from the Expression[] and IEnumerable<Expression> overloads by inspecting the number of expressions supplied (similar to the logic in Block), at the cost of an intermediary allocation for the collection prior to the optimization behind the factory method.

I’ll start small and we can iterate / discuss.

VSadov:
Fixed in https://github.com/dotnet/corefx/pull/3288

Issue #3245, opened 2015-09-15
Add N-ary internal specializations for InvocationExpression


bartdesmet:
In Bing, we use an expression normal form whereby method call expressions etc. get erased for invocation expressions of an unbound parameter, e.g.:

Expression.Call(where, xs, filter)

becomes

Expression.Invoke(Expression.Parameter(typeof(...), "where"), xs, filter)

This is done to decouple client libraries and service libraries through erasure of MemberInfo objects. The service uses a binder to bind the unbound parameter expression to a service-side definition and beta-reduces the invocation of the (lambda-based) definition.

A lot of those invocation expressions have 0, 1, 2, 3 arguments. However, unlike MethodCallExpression, InvocationExpression has no N-ary specializations that help to reduce memory. The proposal is to add such specializations to conserve memory.

bartdesmet:
I can prepare a PR with proposed improvements.

VSadov:
Fixed in #3298

PR #3288, opened 2015-09-17
Use optimized N-ary MethodCallExpression nodes from all Call factories

bartdesmet:
This PR contains a proposed change for issue #3266 to optimize N-ary MethodCallExpression nodes, even when constructed through the overloads that take Expression[] or IEnumerable<Expression> (e.g. as emitted by the compilers). This is similar to the approach taken by BlockExpression.

It's worth noting that those Call factory methods are invoked from a few other places than the public API, e.g. from the Update and Rewrite methods used by visitors, and from ArrayIndex factories. As such, those benefit as well.

The change also includes the addition of a 0-ary and 1-ary optimized nodes that were missing, despite there being a lot of methods that match this signature. To reduce code size a bit, I've factored the common pieces of instance method call expression nodes into an internal base class.

Note that this change does not introduce public API changes (see comments in the PR). Those would cause compile-time ambiguities or overload selection changes for calls with null:

Expression.Call(null) would change from (MethodInfo, params Expression[]) to (MethodInfo)
Expression.Call(e, m, null) would change frrom (Expression, MethodInfo, Expression[]) to (Expression, MethodInfo, Expression)

I'm not sure whether compile-time compat for method invocations with null literals is worth pursuing if it prevents the addition of useful overloads. In this case, we can get away without adding those public overloads. Note that while I was doing similar work for InvocationExpression optimizations (see #3245), I ran into a test where a null literal was passed and the addition of a public overload caused a compile-time failure, so it's not far-fetched. The use-site fix would be easy using default(T), but it's still a break.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
@@ -212,6 +212,40 @@ namespace System.Linq.Expressions
         }
     }
 
+    internal class MethodCallExpression0 : MethodCallExpression, IArgumentProvider
+    {
+        public MethodCallExpression0(MethodInfo method)
+            : base(method)
+        {
+        }
+
+        public override Expression GetArgument(int index)
+        {
+            throw new InvalidOperationException();
+        }
+
+        public override int ArgumentCount
+        {
+            get
+            {
+                return 0;
+            }
+        }
+
+        internal override ReadOnlyCollection<Expression> GetOrMakeArguments()
+        {
+            return EmptyReadOnlyCollection<Expression>.Instance;
+        }
+
+        internal override MethodCallExpression Rewrite(Expression instance, IList<Expression> args)
+        {
+            Debug.Assert(instance == null);
+            Debug.Assert(args == null || args.Count == 0);
+
+            return Expression.Call(Method);
+        }
+    }
+
     internal class MethodCallExpression1 : MethodCallExpression, IArgumentProvider
     {
         private object _arg0;       // storage for the 1st argument or a readonly collection.  See IArgumentProvider
@@ -463,6 +497,104 @@ namespace System.Linq.Expressions
         }
     }
 
+    internal class InstanceMethodCallExpression0 : MethodCallExpression, IArgumentProvider
+    {
+        private readonly Expression _instance;
+
+        public InstanceMethodCallExpression0(MethodInfo method, Expression instance)
+            : base(method)
+        {
+            Debug.Assert(instance != null);
+
+            _instance = instance;
+        }
+
+        public override Expression GetArgument(int index)
+        {
+            throw new InvalidOperationException();
+        }
+
+        public override int ArgumentCount
+        {
+            get
+            {
+                return 0;
+            }
+        }
+
+        internal override Expression GetInstance()
+        {
+            return _instance;
+        }
+
+        internal override ReadOnlyCollection<Expression> GetOrMakeArguments()
+        {
+            return EmptyReadOnlyCollection<Expression>.Instance;
+        }
+
+        internal override MethodCallExpression Rewrite(Expression instance, IList<Expression> args)
+        {
+            Debug.Assert(instance != null);
+            Debug.Assert(args == null || args.Count == 0);
+
+            return Expression.Call(instance, Method);
+        }
+    }
+
+    internal class InstanceMethodCallExpression1 : MethodCallExpression, IArgumentProvider
+    {
+        private readonly Expression _instance;
+        private object _arg0;                // storage for the 1st argument or a readonly collection.  See IArgumentProvider
+
+        public InstanceMethodCallExpression1(MethodInfo method, Expression instance, Expression arg0)
+            : base(method)
+        {
+            Debug.Assert(instance != null);
+
+            _instance = instance;
+            _arg0 = arg0;
+        }
+
+        public override Expression GetArgument(int index)
+        {
+            switch (index)
+            {
+                case 0: return ReturnObject<Expression>(_arg0);
+                default: throw new InvalidOperationException();
+            }
+        }
+
+        public override int ArgumentCount
+        {
+            get
+            {
+                return 1;
+            }
+        }
+
+        internal override Expression GetInstance()
+        {
+            return _instance;
+        }
+
+        internal override ReadOnlyCollection<Expression> GetOrMakeArguments()
+        {
+            return ReturnReadOnly(this, ref _arg0);
+        }
+
+        internal override MethodCallExpression Rewrite(Expression instance, IList<Expression> args)
+        {
+            Debug.Assert(instance != null);
+            Debug.Assert(args == null || args.Count == 1);
+
+            if (args != null)
+            {
+                return Expression.Call(instance, Method, args[0]);
+            }
+            return Expression.Call(instance, Method, ReturnObject<Expression>(_arg0));
+        }
+    }
+
     internal class InstanceMethodCallExpression2 : MethodCallExpression, IArgumentProvider
     {
         private readonly Expression _instance;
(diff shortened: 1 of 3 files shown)


src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
```diff
@@ -879,6 +1042,34 @@ public static MethodCallExpression Call(Type type, string methodName, Type[] typ
         ///<paramref name="instance" />.Type is not assignable to the declaring type of the method represented by <paramref name="method" />.-or-The number of elements in <paramref name="arguments" /> does not equal the number of parameters for the method represented by <paramref name="method" />.-or-One or more of the elements of <paramref name="arguments" /> is not assignable to the corresponding parameter for the method represented by <paramref name="method" />.</exception>
         public static MethodCallExpression Call(Expression instance, MethodInfo method, IEnumerable<Expression> arguments)
         {
+            IReadOnlyList<Expression> argumentList = arguments as IReadOnlyList<Expression>;
```
bartdesmet: Note this differs a tiny bit from the approach taken in BlockExpression, where the optimization is only in place for the Expression[] overload. That has limited mileage, because various other paths (such as Rewrite and Update) call into the IEnumerable<Expression> overload. Also note I'm using IReadOnlyList<T> which is covariant in T, so even passing in a compatible collection of a more derived type will pass the type test.

The one little change in behavior here is the virtual calls to the Count property and indexer if the test passes, therefore relying on the list to be well-behaved in terms of the interface contract. To be fair, existing code behind ToReadOnly does a similar check based on ICollection<T> (less applicable given it's not covariant in T, maybe worth looking into separately) so such assumptions of well-behaved interface implementations exist elsewhere.

stephentoub:
cc: @VSadov

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
VSadov: I think the only reason for BlockExpression to not use IReadOnlyList is that it did not exist at that time. We may at some point review these kind of interface checks to see if they can be improved.

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
```diff
@@ -879,6 +1042,34 @@ public static MethodCallExpression Call(Type type, string methodName, Type[] typ
...
+                        return Call(instance, method);
+                    case 1:
+                        return Call(instance, method, argumentList[0]);
+                    case 2:
+                        return Call(instance, method, argumentList[0], argumentList[1]);
+                    case 3:
+                        return Call(instance, method, argumentList[0], argumentList[1], argumentList[2]);
+                }
+
+                if (instance == null)
```
bartdesmet: Note there's always the option to go beyond optimized instance call nodes with 3 parameters. I've run a few LINQ queries over all public types in mscorlib, System, System.Core on Windows 10 and here are the stats:

Static methods:

{ Arity = 0, Count = 553 }
{ Arity = 1, Count = 1724 }
{ Arity = 2, Count = 1450 }
{ Arity = 3, Count = 504 }
{ Arity = 4, Count = 231 }
{ Arity = 5, Count = 118 }
{ Arity = 6, Count = 38 }
{ Arity = 7, Count = 18 }
{ Arity = 8, Count = 6 }
{ Arity = 9, Count = 7 }
{ Arity = 10, Count = 3 }
{ Arity = 11, Count = 1 }
{ Arity = 12, Count = 1 }

Instance methods:

{ Arity = 0, Count = 39542 }
{ Arity = 1, Count = 20498 }
{ Arity = 2, Count = 5331 }
{ Arity = 3, Count = 2032 }
{ Arity = 4, Count = 780 }
{ Arity = 5, Count = 625 }
{ Arity = 6, Count = 171 }
{ Arity = 7, Count = 81 }
{ Arity = 8, Count = 131 }
{ Arity = 9, Count = 27 }
{ Arity = 10, Count = 30 }
{ Arity = 11, Count = 4 }
{ Arity = 12, Count = 4 }
{ Arity = 13, Count = 5 }
{ Arity = 14, Count = 4 }
{ Arity = 15, Count = 4 }
{ Arity = 16, Count = 4 }
{ Arity = 17, Count = 2 }
{ Arity = 18, Count = 2 }

The addition for low argument counts on instance methods seems well worth it. An argument count of 4 and 5 is still quite frequent for those as well.

VSadov:
LGTM. 
The change does not affect public APIs so is very uncontroversial.
@stephentoub - any comments?

bartdesmet:
Base on the data I've gathered (see line notes above), it looks like it may be good to add the optimized nodes for instance calls with 4 and maybe 5 parameters. 99th percentile for static is at 6 parameters, for instance it's at 5 parameters. We can do it without public API additions and still get the benefit.

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
VSadov: Expanding specializations to 4 and 5 is definitely an option. It does not look like the savings will be as dramatic as for 0,1,2,3 args.

Note that the overhead of keeping arguments in a collection is constant, so in addition to the frequency of the scenarios being lower, the relative benefits of the optimization are lower as well.

VSadov: Also I think there is a scenario where the collection of arguments gets realized regardless and then stuffed into the arg0 or something like that. 
I am not sure how common that is, but it would be a scenario where specializations for greater number of arguments would have more redundancy. It is possible these scenarios are not common though.

bartdesmet: Agreed; we can keep it as-is for the time being. Not sure what the history behind the static methods going up to 5 arguments was; maybe C# or VB compilers emit static method calls (a la CompareString) for some constructs that take that many parameters.

bartdesmet: Correct, stepping on the public Arguments property causes the magic morphing into a collection. Some external visitors (a la a ExpressionVisitor<T> we built here) step on that frequently, but the recent change where IArgumentProvider was made public (not sure exactly when that happened) makes it possible to avoid the materialization and pick out the arguments without causing the mutation to a collection to happen.

VSadov: Could also be something in DLR/dynamic that frequently goes to 5 arguments...

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
```diff
@@ -212,6 +223,40 @@ internal override MethodCallExpression Rewrite(Expression instance, IList<Expres
         }
     }
 
+    internal class MethodCallExpression0 : MethodCallExpression, IArgumentProvider
```
stephentoub: Can we seal types like this?  Or do we not do that in this library? (Or maybe something somewhere derives from this?)

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
stephentoub: I forget the details, but isn't there some kind of perf impact with regards to doing casts to interfaces with covariant type arguments?

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
```diff
@@ -879,6 +1042,34 @@ public static MethodCallExpression Call(Type type, string methodName, Type[] typ
...
+                        return Call(instance, method, argumentList[0]);
+                    case 2:
+                        return Call(instance, method, argumentList[0], argumentList[1]);
+                    case 3:
+                        return Call(instance, method, argumentList[0], argumentList[1], argumentList[2]);
+                }
+
+                if (instance == null)
+                {
+                    switch (argumentList.Count)
```
stephentoub: Minor, but you could save the count into a local earlier such that in this case we don't need to make the extra interface call.

stephentoub:
A few additional comments, but otherwise LGTM.  @bartdesmet, do you have any kind of performance test results showing the resulting (hopefully) improvements?

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
bartdesmet: We definitely can. There are no derived types. Haven't seen many other internal types that are sealed in this project though, but I wouldn't mind.

Sent from my Windows Phone
________________________________

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
bartdesmet: Fair point; I'll tweak this.

Sent from my Windows Phone
________________________________

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
bartdesmet: Maybe someone from the JIT team could chime in here? I can run a small test too.

Sent from my Windows Phone
________________________________

bartdesmet:
I do have a small local test that mimics a scenario where many such expressions are rooted (akin to our infrastructure in Bing). The memory savings are good. I can get the numbers when I'm back in the office. Not sure if we have a good means in the repo to share such perf suites?

As an aside, I miss some "playground" console projects in the corefx solutions to do quick F5 experiments. Maybe that's worth considering throughout the repo?

Sent from my Windows Phone
________________________________

bartdesmet:
@stephentoub - I've rerun my perf test shown below:

```
    {
        var N = int.Parse(args[0]);

        var trees = new List<Expression>(N);

        for (var i = 0; i < N; i++)
        {
            Expression<Func<string, string>> f = s => string.Concat(s.Substring(1, 2).Substring(1).ToLower(), string.Concat("Bar", "Foo", "Qux"));
                trees.Add(f);
        }

        GC.Collect();
        Console.WriteLine("Total memory = " + GC.GetTotalMemory(true));

        GC.KeepAlive(trees);
    }
```

The results are as follows:

```
Before
------

>corerun System.Linq.Expressions.Playground.exe 100
Total memory = 177936

>corerun System.Linq.Expressions.Playground.exe 1000
Total memory = 933944

>corerun System.Linq.Expressions.Playground.exe 10000
Total memory = 8493944

>corerun System.Linq.Expressions.Playground.exe 100000
Total memory = 84093944

>corerun System.Linq.Expressions.Playground.exe 1000000
Total memory = 840093944


After
-----

>corerun System.Linq.Expressions.Playground.exe 100
Total memory = 151600

>corerun System.Linq.Expressions.Playground.exe 1000
Total memory = 670008

>corerun System.Linq.Expressions.Playground.exe 10000
Total memory = 5854008

>corerun System.Linq.Expressions.Playground.exe 100000
Total memory = 57694008

>corerun System.Linq.Expressions.Playground.exe 1000000
Fails with EngineExecutionException
```

Savings are significant but it seems I've stepping on what looks like a runtime issue; running the test with the changes with one million iterations causes an EngineExecutionException. I've simplified the code and provided a minimal repro in #3308. It seems unrelated to the expression API or covariant interface conversions and it occurs in the presence of GC pressure.

src/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs
VSadov: I have entered an issue to track this.
https://github.com/dotnet/coreclr/issues/1592

We should have better understanding what particular scenarios to avoid (if perf is bad for a reason) or perhaps fix it.

VSadov:
I think all concerns here are addressed and the part about covariant casts is a broader issue and should be followed up separately

PR #3280, opened 2015-09-16
Restore disabled MMF inter-process test

stephentoub:
Also removing some stale files that got left behind in a previous refactoring.

This test was previously causing fairly consistent "file in use" failures in the lab.  I've been unable to reproduce it locally or in private runs submitted through the CI system.  I'm reenabling it in hopes that whatever issues were previously there are gone, but I'll disable it again at the first sign of problems.

Fixes #2498

PR #3296, opened 2015-09-17
Remove and fix incorrect reflection helpers

### LINKED ISSUE #3247: GetGenericArguments for open generic type returns empty array
The following test fails in CoreCLR:

[Fact]
public static void GenDefArgs()
{
    Assert.Equal(1, typeof(Func<>).GetGenericArguments().Length);
}

On desktop CLR, the following prints 1:

Console.WriteLine(typeof(Func<>).GetGenericArguments().Length);

For closed generic types, things seem fine.

nguerrera:
There were two instances in the code of trying to mimic the old Type.GetGenericArguments() with TypeInfo API that were incorrect. Both attempts ignored that TypeInfo separates generic type definition parameters from generic instantiation arguments.

Case 1: System.Linq.Expressions.Tests
--------------------------------------
These already had a reference to System.Reflection.TypeExtensions, which provides all of the compatibility API it needed, so the custom helpers are simply deleted.

Case 2: System.Linq.Queryable
-----------------------------
This library doesn't depend on System.Reflection.TypeExtensions and so the trivial helper is left there but corrected to match the expected behavior of the classic API.

Fix #3247

cc @VSadov @bartdesmet


stephentoub:
@dotnet-bot test this please (failed because of the test I reenabled in #3280 :frowning: I'll investigate)

stephentoub:
LGTM

VSadov:
LGTM

PR #3298, opened 2015-09-17
Adding optimized N-ary InvocationExpression nodes

bartdesmet:
This is a proposed improvement as described in issue #3245. It follows the strategy used in PR #3288.

Note that no public API additions were made even though some of the factory overloads would make sense to be exposed publicly. This conservative stance could be revisited separately if a decision is made to incorporate public API additions for the expression API.


stephentoub:
cc: @VSadov

PR #3300, opened 2015-09-17
Fixing openssl.c to not #pragma ignore warnings for asserts

sokket:
Fixing openssl.c to not #pragma ignore warnings for asserts

/cc @nguerrera @stephentoub @bartonjs

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -384,19 +384,29 @@ GetX509NameRawBytes(
         return 0;
     }
 
-// length is size_t on some platforms and int on others, so the comparisons
-// are not tautological everywhere. We can let the compiler optimize away
-// any part of the check that is.
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wtautological-compare"
-    if (x509Name->bytes->length < 0 || x509Name->bytes->length > INT_MAX)
+    /* 
+     * length is size_t on some platforms and int on others, so the comparisons
+     * are not tautological everywhere. We can let the compiler optimize away
+     * any part of the check that is. We split the size checks into two checks
+     * so we can get around the warnings on Linux where the Length is unsigned
+     * whereas Length is signed on OS X. The first check makes sure the variable 
+     * value is less than INT_MAX in it's native format; once we know it is not
+     * too large, we can safely cast to an int to make sure it is not negative
+     */
+    if (x509Name->bytes->length > INT_MAX)
     {
-        assert(!"Huge/negative-length X509_NAME");
+        assert(0 && "Huge length X509_NAME");
         return 0;
     }
-#pragma clang diagnostic pop
 
     int length = (int)(x509Name->bytes->length);
+
+    if (length < 0)
+    {
+        assert(0 && "Negative length X509_NAME");
+        return 0;
+    }
+
     if (!pBuf || cBuf < length)
     {
         return -length;
@@ -1027,7 +1037,7 @@ LockingCallback(int mode, int n, const char* file, int line)
 
     if (result != 0)
     {
-        assert(!"LockingCallback failed.");
+        assert(0 && "LockingCallback failed.");
     }
 }
 
@@ -1060,7 +1070,7 @@ EnsureOpenSslInitialized()
     numLocks = CRYPTO_num_locks();
     if (numLocks <= 0)
     {
-        assert(!"CRYPTO_num_locks returned invalid value.");
+        assert(0 && "CRYPTO_num_locks returned invalid value.");
         ret = 1;
         goto done;
     }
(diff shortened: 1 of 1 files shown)


src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -384,19 +384,29 @@ GetX509NameRawBytes(
...
+     * any part of the check that is. We split the size checks into two checks
+     * so we can get around the warnings on Linux where the Length is unsigned
+     * whereas Length is signed on OS X. The first check makes sure the variable 
+     * value is less than INT_MAX in it's native format; once we know it is not
+     * too large, we can safely cast to an int to make sure it is not negative
+     */
+    if (x509Name->bytes->length > INT_MAX)
     {
-        assert(!"Huge/negative-length X509_NAME");
+        assert(0 && "Huge length X509_NAME");
```
stephentoub: I believe I saw Nick using ```false``` instead of ```0``` in similar situations.  It'd be good to be consistent.

bartonjs:
LGTM.

I don't know why my machine didn't like the pragmas; but hopefully this makes it less sad :)

src/Native/System.Security.Cryptography.Native/openssl.c
nguerrera: This file is still .c so `false` won't work.

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -384,19 +384,29 @@ GetX509NameRawBytes(
...
         return 0;
     }
-#pragma clang diagnostic pop
 
     int length = (int)(x509Name->bytes->length);
+
+    if (length < 0)
+    {
+        assert(0 && "Negative-length X509_NAME");
```
stephentoub: Nit: no need for the dash

src/Native/System.Security.Cryptography.Native/openssl.c
stephentoub: Ah, didn't notice that.  Shouldn't we change that, too?

nguerrera:
That #pragma is not about the assert, but the comparison in the `if (...)`. It was failing to compile on OS X for me without disabling the warning. The comment explains the pragma.

I'd like to get the bottom of why those pragmas aren't working, but if we're blocked, you'll need to add -Wno-tautological-compare to the CMakeLists.txt. Unless it's just me that gets the warning on OS X, but that would indicate we're building with different versions of clang and/or openssl.

src/Native/System.Security.Cryptography.Native/openssl.c
nguerrera: Yes. It will cause a bunch more warnings (c-style cast mostly) so was putting it off for now. But we should definitely do it.

bartonjs:
@nguerrera The warnings I got were for the asserts themselves:

```
/home/jbarton/git/bartonjs/corefx/src/Native/System.Security.Cryptography.Native/openssl.c:1082:21: 
error: implicit conversion turns string literal into bool: 'char [27]' to '_Bool' [-Werror,-Wstring-conversion]
            assert(!"pthread_mutex_init failed.");
```

nguerrera:
@bartonjs assert(0 && "") is fine for those.

I missed that the if condition was actually changed to fix the existing pragma length is a signed int on my OS X machine so < 0 is not tautological everywhere. I guess we'll never have negative length, but I was being meticulous about proving everything that ends up becoming a buffer size is safe.

src/Native/System.Security.Cryptography.Native/openssl.c
sokket: This is a C file, so there is no false :) FALSE is also not defined here, so I have to use 0

nguerrera:
Nevermind, re-reading the diff I see all bases are still covered. It was obscured by the comments in between. LGTM. ;)

src/Native/System.Security.Cryptography.Native/openssl.c
sokket: Fixed

nguerrera:
FWIW, the commit description is a bit off. This change 1) addresses string to bool conversion warnings 2) fixes tautological compare warning without needing assert. That was another source of confusion above for me.

src/Native/System.Security.Cryptography.Native/openssl.c
jasonwilliams200OK: > This file is still .c so false won't work.

Will work whenever minimum requirement of building CoreFX would be set to VS2015:

* `bool` is ISO C99 type.
* VS2015 supports it.
* Moving forward, many C89 inabilities workaround can be removed in favor of VS2015's C99 support (or guarded with `#if defined(_MSC_VER) && _MSC_VER < 1900` for now, until VS2013 support is dropped).

nguerrera: Fair point about C99, we could have included <stdbool.h> to use true/false here. But since we'll be converting this code to  C++  soon and making other cleanup such as removing C-style casts, it doesn't really matter IMHO if we wait until then to switch 1/0 to true/false here. 

Note FWIW that we don't compile any of the unix shim code with MSVC.

jasonwilliams200OK: > Note FWIW that we don't compile any of the unix shim code with MSVC.

Which is rather reassuring, as GCC and Clang support C99 since the inception of time.. It was MSVC which came (like 16 years) late to the party. :smile:

PR #3303, opened 2015-09-17
Fix two time zone tests on OS X.

### LINKED ISSUE #3211: Two TimeZoneInfo tests are failing on OS X
```
09:23:54    TimeZoneInfoTests.NearMinMaxDateTimeOffsetConvertTest [FAIL]
09:23:54       Error: Expected value '12/31/9999 23:59:59' but got '12/31/9999 22:59:59', input value is '12/31/9999 04:59:59', Source TimeZone: America/Los_Angeles, Dest. Time Zone: Australia/Sydney
09:23:54       Expected: True
09:23:54       Actual:   False
09:23:54    System.Runtime.Extensions.Tests  Total: 495, Errors: 0, Failed: 1, Skipped: 0, Time: 1.096s
09:23:54       Stack Trace:
09:23:54             at TimeZoneInfoTests.VerifyConvert(DateTime inputTime, String sourceTimeZoneId, String destinationTimeZoneId, DateTime expectedTime)
09:23:54             at TimeZoneInfoTests.NearMinMaxDateTimeOffsetConvertTest()
09:23:54    TimeZoneInfoTests.NearMinMaxDateTimeConvertTest [FAIL]
09:23:54       Error: Expected value '12/31/9999 23:59:59' but got '12/31/9999 22:59:59', input value is '12/31/9999 12:59:59', TimeZone: Australia/Sydney
09:23:54       Expected: True
09:23:54       Actual:   False
09:23:54       Stack Trace:
09:23:54             at TimeZoneInfoTests.VerifyConvert(DateTime inputTime, String destinationTimeZoneId, DateTime expectedTime)
09:23:54             at TimeZoneInfoTests.NearMinMaxDateTimeConvertTest()
```
cc: @eerhardt, @ellismg, @tarekgh

eerhardt:
Fixing two time zone tests on OS X.  OS X appears to have v1 tzfiles, which lack some of the time zone information that v2 tzfiles have.  Adjusting the tests accordingly.

Fix #3211.

@tarekgh, @stephentoub

### DIFF
--- a/src/System.Runtime/tests/System/TimeZoneInfo.cs
+++ b/src/System.Runtime/tests/System/TimeZoneInfo.cs
@@ -10,6 +10,8 @@ using Xunit;
 public static class TimeZoneInfoTests
 {
     private static readonly bool s_isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+    private static readonly bool s_isOSX = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
+
     private static String s_strPacific = s_isWindows ? "Pacific Standard Time" : "America/Los_Angeles";
     private static String s_strSydney = s_isWindows ? "AUS Eastern Standard Time" : "Australia/Sydney";
     private static String s_strGMT = s_isWindows ? "GMT Standard Time" : "Europe/London";
@@ -146,7 +148,6 @@ public static class TimeZoneInfoTests
         VerifyConvertException<Exception>(time1, new String('a', 256)); // long string
     }
 
-    [ActiveIssue(3211, PlatformID.OSX)]
     [Fact]
     public static void NearMinMaxDateTimeOffsetConvertTest()
     {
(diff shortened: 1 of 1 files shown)


src/System.Runtime/tests/System/TimeZoneInfo.cs
```diff
@@ -10,6 +10,8 @@
 public static class TimeZoneInfoTests
 {
     private static readonly bool s_isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+    private static readonly bool s_isOSX = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
```
tarekgh: does it make sense to move the Linux test code into its own methods and mark such methods with the Platform specific attribute? I think this will make the code cleaner and removing many of the runtime checks. of course we can do the same with Windows specific tests. what you think?

stephentoub:
Other than Tarek's question, LGTM.

src/System.Runtime/tests/System/TimeZoneInfo.cs
eerhardt: I don't like breaking these into separate test methods because I think that breaks the cohesiveness of these checks.  In the case of NearMinMaxDateTimeOffsetConvertTest, all 3 branches in the if-else if-else statement are trying to test the same thing, only they all have different tz data to work with on the separate systems.

So instead, I broke the "data" out into a separate method that returns the correct offset for the timezone given the current platform.  Please have a look and tell me what you think.

tarekgh: I am ok with your change but here is another idea you may consider it.

[Theory]
[InlineData(... whatever Linux Specific data...)]
[PlatformSpecific(PlatformID.Linux)]
public static void LinuxNearMinMaxDateTimeOffsetConvertTest(..Linux data...)
{
    NearMinMaxDateTimeOffsetConvertTest(data);
}

[Theory]
[InlineData(... whatever Windows Specific data...)]
public static void WindowsNearMinMaxDateTimeOffsetConvertTest(..Windows data.)
{
    NearMinMaxDateTimeOffsetConvertTest(data);
}

public static void NearMinMaxDateTimeOffsetConvertTest(data)
{
 ...
}

eerhardt: I think that way is more complicated than the runtime checks.

tarekgh: I'll leave to you to decide.

PR #3304, opened 2015-09-18
Added performance microbenchmarks

ianhays:
This is a replacement to #3159 that enables performance tests to run on Windows using [Benchmark] attributes. These tests should also run on Linux whenever the perf runner from https://github.com/Microsoft/xunit-performance/ is moved x-plat.

- Added "Performance" folders containing perf xunit tests to a number of highly-used libraries (determined by Usage data)
- These tests WILL run on Windows when using ```msbuild /t:BuildAndTest /p:performance=true``` from ```corefx\src\library_to_test\tests```
- Two sets of usage data from store apps; I added a test for every function used in more than 5% of the covered assemblies in both data sets.
- This is a first-round pass of the most highly used libraries and does not include: abstract classes, interfaces, or Attributes
- Created a static utility class for commonly used functions across perf tests, stored in src\Common\tests\Performance.

Note: Depends on https://github.com/dotnet/buildtools/pull/272. Until that package is published, it will fail CI.

I do plan on getting test-by-test usage data to determine the ideal unrolled loop size for each function on Linux/Windows and will publish those changes (as well as the data) at a later date. This PR should get some other people unblocked and be a good baseline, though.

cc: @mellinoe @jhendrixMSFT @bryanar @ericeil @vancem @stephentoub @joshfree

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00094</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00095</BuildToolsVersion>
     <DnxVersion>1.0.0-beta7</DnxVersion>
     <DnxPackageName Condition="'$(DnxPackageName)' == '' and '$(OsEnvironment)'!='Unix'">dnx-coreclr-win-x86.$(DnxVersion)</DnxPackageName>
     <DnxPackageName Condition="'$(DnxPackageName)' == '' and '$(OsEnvironment)'=='Unix'">dnx-mono.$(DnxVersion)</DnxPackageName>
@@ -260,6 +260,11 @@
     <SignAppxPackageExeFullPath>$(_WindowsKitBinPath)\signtool.exe</SignAppxPackageExeFullPath>
     <MakePriExtensionPath>$(_WindowsPhoneKitBinPath)\x86\MrmEnvironmentExtDl.dll</MakePriExtensionPath>
     <MakePriExtensionPath_x64>$(_WindowsPhoneKitBinPath)\x64\MrmEnvironmentExtDl.dll</MakePriExtensionPath_x64>
+  </PropertyGroup>
+  
+  <!-- Perf test setup - Required until v5.0 is mandatory across CoreFX-->
+  <PropertyGroup>
+    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(OS)' == 'Windows_NT' and Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
   </PropertyGroup>
 
   <Import Project="$(RoslynPropsFile)"
--- a/src/Common/tests/Performance/PerfUtils.cs
+++ b/src/Common/tests/Performance/PerfUtils.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.IO;
+
+namespace System
+{
+    /// <summary>Static helper class for performance tests</summary>
+    public static class PerfUtils
+    {
+        private static Random rand = new Random(1234132);
+
+        /// <summary>
+        /// Helper method to create a string containing a number of random
+        /// characters equal to the specified length
+        /// </summary>
+        public static string CreateString(int length)
+        {
+            // Random is not thread safe and will cause intermittent thread deadlock if used without being locked
+            lock (rand)
+            {
+                byte[] bytes = new byte[length];
+                rand.NextBytes(bytes);
+                return System.Convert.ToBase64String(bytes);
+            }
+        }
+
+        /// <summary>Gets a test file full path that is associated with the call site.</summary>
+        /// <param name="index">An optional index value to use as a suffix on the file name.  Typically a loop index.</param>
+        /// <param name="memberName">The member name of the function calling this method.</param>
+        /// <param name="lineNumber">The line number of the function calling this method.</param>
+        public static string GetTestFilePath(int? index = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
+        {
+            return Path.Combine(Path.GetTempPath(), string.Format(
+                index.HasValue ? "{0}_{1}_{2}_{3}" : "{0}_{1}_{2}",
+                memberName ?? "TestBase", lineNumber, Path.GetRandomFileName(), 
+                index.GetValueOrDefault()));
+        }
+    }
+}
--- a/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
+++ b/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+using Microsoft.Xunit.Performance;
+using System.Collections.Generic;
+
+namespace System.Collections.Tests
+{
+    public class Perf_HashTable
+    {
+        private static List<object[]> _testData;
+
+        public static IEnumerable<object[]> TestData()
+        {
+            if (_testData == null)
+            {
+                _testData = new List<object[]>();
+                _testData.Add(new object[] { CreateHashtable(100) });
+                _testData.Add(new object[] { CreateHashtable(1000) });
+            }
+            return _testData;
+        }
+
+        public static Hashtable CreateHashtable(int size)
+        {
+            Hashtable ht = new Hashtable();
+            for (int i = 0; i < size; i++)
+                ht.Add(PerfUtils.CreateString(50), PerfUtils.CreateString(50));
+            return ht;
+        }
+
+        [Benchmark]
+        public void ctor()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable();
+                    new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable();
+                    new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable();
+                    new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable(); new Hashtable();
+                }
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void GetItem(Hashtable table)
+        {
+            object result;
+            string key = PerfUtils.CreateString(50);
+            table.Add(key, "value");
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    result = table[key]; result = table[key]; result = table[key]; result = table[key];
+                    result = table[key]; result = table[key]; result = table[key]; result = table[key];
+                    result = table[key]; result = table[key]; result = table[key]; result = table[key];
+                    result = table[key]; result = table[key]; result = table[key]; result = table[key];
+                }
+            }
+            table.Remove(key);
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void Add(Hashtable table)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    table.Add("key1", "value"); table.Add("key2", "value"); table.Add("key3", "value");
+                    table.Add("key4", "value"); table.Add("key5", "value"); table.Add("key6", "value");
+                    table.Add("key7", "value"); table.Add("key8", "value"); table.Add("key9", "value");
+                }
+                for (int i = 1; i <= 9; i++)
+                    table.Remove("key" + i);
+            }
+        }
+    }
+}
--- a/src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
+++ b/src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
@@ -8,6 +8,11 @@
     <RootNamespace>System.Collections.NonGeneric.Tests</RootNamespace>
     <AssemblyName>System.Collections.NonGeneric.Tests</AssemblyName>
     <ProjectGuid>{EE95AE39-845A-42D3-86D0-8065DBE56612}</ProjectGuid>
+  </PropertyGroup>
+  <!-- Performance tests require v5.0 portable tools -->
+  <PropertyGroup Condition="'$(RunPerfTestsForProject)' == 'true'">
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <!-- Default configurations to help VS understand the configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
@@ -145,6 +150,13 @@
     <Compile Include="XunitAssemblyAttributes.cs" />
     <Compile Include="$(CommonTestPath)\System\Diagnostics\DebuggerAttributes.cs">
       <Link>Common\System\Diagnostics\DebuggerAttributes.cs</Link>
+    </Compile>
+  </ItemGroup>
+  <!-- Performance Tests -->
+  <ItemGroup Condition="'$(RunPerfTestsForProject)' == 'true'">
+    <Compile Include="Performance\Perf.HashTable.cs" />
+    <Compile Include="$(CommonTestPath)\Performance\PerfUtils.cs">
+      <Link>Common\Performance\PerfUtils.cs</Link>
     </Compile>
   </ItemGroup>
   <ItemGroup>
--- a/src/System.Collections/tests/Performance/Perf.Dictionary.cs
+++ b/src/System.Collections/tests/Performance/Perf.Dictionary.cs
@@ -1,0 +1,203 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Xunit;
+using Microsoft.Xunit.Performance;
+
+namespace System.Collections.Tests
+{
+    public class Perf_Dictionary
+    {
+        private static List<object[]> _testData;
+
+        /// <summary>
+        /// Yields several Dictionaries containing increasing amounts of string-string
+        /// pairs can be used as MemberData input to performance tests for Dictionary
+        /// </summary>
+        /// <remarks>Any changes made to the returned collections MUST be undone. Collections
+        /// used as MemberData are cached and reused in other perf tests.
+        /// </remarks>
+        public static List<object[]> TestData()
+        {
+            if (_testData == null)
+            {
+                _testData = new List<object[]>();
+                _testData.Add(new object[] { CreateDictionary(100) });
+                _testData.Add(new object[] { CreateDictionary(1000) });
+            }
+            return _testData;
+        }
+
+        /// <summary>
+        /// Creates a Dictionary of string-string with the specified number of pairs
+        /// </summary>
+        public static Dictionary<string, string> CreateDictionary(int size)
+        {
+            Dictionary<string, string> dict = new Dictionary<string, string>();
+            while (dict.Count < size)
+            {
+                string key = PerfUtils.CreateString(50);
+                while (dict.ContainsKey(key))
+                    key = PerfUtils.CreateString(50);
+                dict.Add(key, PerfUtils.CreateString(50));
+            }
+            return dict;
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void Add(Dictionary<string, string> dict)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    dict.Add("key1", "string"); dict.Add("key2", "string"); dict.Add("key3", "string");
+                    dict.Add("key4", "string"); dict.Add("key5", "string"); dict.Add("key6", "string");
+                    dict.Add("key7", "string"); dict.Add("key8", "string"); dict.Add("key9", "string");
+                }
+                for (int i = 1; i <= 9; i++)
+                    Assert.True(dict.Remove("key" + i));
+            }
+        }
+
+        [Benchmark]
+        public void ctor()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    new Dictionary<int, string>(); new Dictionary<int, string>(); new Dictionary<int, string>();
+                    new Dictionary<int, string>(); new Dictionary<int, string>(); new Dictionary<int, string>();
+                    new Dictionary<int, string>(); new Dictionary<int, string>(); new Dictionary<int, string>();
+                }
+        }
+
+        [Benchmark]
+        [InlineData(0)]
+        [InlineData(1024)]
+        [InlineData(4096)]
+        [InlineData(16384)]
+        public void ctor_int(int size)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    new Dictionary<int, string>(size); new Dictionary<int, string>(size); new Dictionary<int, string>(size);
+                    new Dictionary<int, string>(size); new Dictionary<int, string>(size); new Dictionary<int, string>(size);
+                    new Dictionary<int, string>(size); new Dictionary<int, string>(size); new Dictionary<int, string>(size);
+                }
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void GetItem(Dictionary<string, string> dict)
+        {
+            // Setup
+            string retrieved;
+            for (int i = 1; i <= 9; i++)
+                dict.Add("key" + i, "value");
+
+            // Actual perf testing
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    retrieved = dict["key1"]; retrieved = dict["key2"]; retrieved = dict["key3"];
+                    retrieved = dict["key4"]; retrieved = dict["key5"]; retrieved = dict["key6"];
+                    retrieved = dict["key7"]; retrieved = dict["key8"]; retrieved = dict["key9"];
+                }
+            }
+
+            // Teardown
+            for (int i = 1; i <= 9; i++)
+                dict.Remove("key" + i);
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void SetItem(Dictionary<string, string> dict)
+        {
+            // Setup
+            for (int i = 1; i <= 9; i++)
+                dict.Add("key" + i, "value");
+
+            // Actual perf testing
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    dict["key1"] = "string"; dict["key2"] = "string"; dict["key3"] = "string";
+                    dict["key4"] = "string"; dict["key5"] = "string"; dict["key6"] = "string";
+                    dict["key7"] = "string"; dict["key8"] = "string"; dict["key9"] = "string";
+                }
+            }
+
+            // Teardown
+            for (int i = 1; i <= 9; i++)
+                dict.Remove("key" + i);
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void GetKeys(Dictionary<string, string> dict)
+        {
+            IEnumerable<string> result;
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    result = dict.Keys; result = dict.Keys; result = dict.Keys;
+                    result = dict.Keys; result = dict.Keys; result = dict.Keys;
+                    result = dict.Keys; result = dict.Keys; result = dict.Keys;
+                }
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void TryGetValue(Dictionary<string, string> dict)
+        {
+            // Setup
+            string retrieved;
+            string key = PerfUtils.CreateString(50);
+            dict.Add(key, "value");
+
+            // Actual perf testing
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    dict.TryGetValue(key, out retrieved); dict.TryGetValue(key, out retrieved);
+                    dict.TryGetValue(key, out retrieved); dict.TryGetValue(key, out retrieved);
+                    dict.TryGetValue(key, out retrieved); dict.TryGetValue(key, out retrieved);
+                    dict.TryGetValue(key, out retrieved); dict.TryGetValue(key, out retrieved);
+                }
+            }
+
+            // Teardown
+            dict.Remove(key);
+        }
+
+        [Benchmark]
+        [MemberData("TestData")]
+        public void ContainsKey(Dictionary<string, string> dict)
+        {
+            // Setup
+            string key = PerfUtils.CreateString(50);
+            dict.Add(key, "value");
+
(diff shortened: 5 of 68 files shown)


src/Common/tests/Performance/PerfUtils.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            StringBuilder builder = new StringBuilder();
+            while (builder.Length < length)
+            {
+                string toAppend = Guid.NewGuid().ToString();
+                int toAppendLength = Math.Min(toAppend.Length, length - builder.Length);
+                builder.Append(toAppend, 0, toAppendLength);
+            }
+            return builder.ToString();
+        }
```
stephentoub: For better or worse, this choice of string generation is going to potentially have an impact on the tests.  Using Guid.ToString means that the resulting strings will only ever have the characters 'a'-'f' and '0'-'9', which can play a role in the performance of things like compression.  Something to keep in mind.

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        [Benchmark]
+        [MemberData("TestHashtables")]
+        public void GetItem(Hashtable table)
+        {
+            object result;
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                table.Add("key", "value");
```
stephentoub: You're adding the key to the same table for every iteration... is that what you intended?

src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
```diff
@@ -8,6 +8,8 @@
     <RootNamespace>System.Collections.NonGeneric.Tests</RootNamespace>
     <AssemblyName>System.Collections.NonGeneric.Tests</AssemblyName>
     <ProjectGuid>{EE95AE39-845A-42D3-86D0-8065DBE56612}</ProjectGuid>
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
+    <TargetFrameworkProfile />
```
stephentoub: Is this correct?  I didn't think we were setting TargetFrameworkVersion or TargetFrameworkProfile in the .csproj files, letting the buildtools supply whatever common values were needed?

src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
```diff
@@ -89,7 +91,8 @@
     <Compile Include="Hashtable\PropertyValuesTests.cs" />
     <Compile Include="Hashtable\RemoveTests.cs" />
     <Compile Include="Hashtable\SynchronizedTests.cs" />
-    <Compile Include="Queue\Queue_Clear.cs" />
+    <Compile Include="Performance\Perf.HashTable.cs" />
+	<Compile Include="Queue\Queue_Clear.cs" />
```
stephentoub: Nit: spacing

src/System.Collections/tests/Performance/Perf.Dictionary.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                using (iteration.StartMeasurement())
+                {
+                    retrieved = dict["key1"]; retrieved = dict["key2"]; retrieved = dict["key3"];
+                    retrieved = dict["key4"]; retrieved = dict["key5"]; retrieved = dict["key6"];
+                    retrieved = dict["key7"]; retrieved = dict["key8"]; retrieved = dict["key9"];
+                }
+            }
+
+            // Teardown
```
stephentoub: Is this teardown necessary?  It looks like we're creating new dictionary instances for every invocation of the test.

src/System.Collections/tests/Performance/Perf.Dictionary.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            // Actual perf testing
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    dict.TryGetValue("key", out retrieved); dict.TryGetValue("key", out retrieved);
+                    dict.TryGetValue("key", out retrieved); dict.TryGetValue("key", out retrieved);
+                    dict.TryGetValue("key", out retrieved); dict.TryGetValue("key", out retrieved);
+                    dict.TryGetValue("key", out retrieved); dict.TryGetValue("key", out retrieved);
```
stephentoub: Something to consider with these dictionary tests is how GetHashCode/Equals impacts the perf of the type.  Using a short constant string like this is likely to minimize the time involved in executing those operations, which Dictionary does a lot.  It could be useful to have a test that measure dictionary perf on one or more common operations using some type for which computing these is more expensive, so as to help catch any regressions where a change in dictionary's implementation causes them to be evaluated more often than necessary.

src/System.Collections/tests/Performance/Perf.Dictionary.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                using (iteration.StartMeasurement())
+                {
+                    dict.ContainsKey("key"); dict.ContainsKey("key"); dict.ContainsKey("key");
+                    dict.ContainsKey("key"); dict.ContainsKey("key"); dict.ContainsKey("key");
+                    dict.ContainsKey("key"); dict.ContainsKey("key"); dict.ContainsKey("key");
+                    dict.ContainsKey("key"); dict.ContainsKey("key"); dict.ContainsKey("key");
+                }
+            }
+            dict.Remove("key");
```
stephentoub: Same question about teardown.  I'll stop commenting on this.

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    list.Add("TestString1"); list.Add("TestString2"); list.Add("TestString3"); list.Add("TestString4");
+                    list.Add("TestString5"); list.Add("TestString6"); list.Add("TestString7"); list.Add("TestString8");
+                }
+                for (int i = 1; i <= 8; i++)
+                    list.Remove("TestString" + i);
```
stephentoub: You could use RemoveRange instead.

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        [Benchmark]
+        [MemberData("TestLists")]
+        public void Clear(List<object> list)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                // Create a local hard copy so that future iterations arent affected
```
stephentoub: Nit: aren't

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        [Benchmark]
+        [MemberData("TestLists")]
+        public void GetEnumerator(List<object> list)
+        {
+            IEnumerator temp;
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    temp = list.GetEnumerator(); temp = list.GetEnumerator(); temp = list.GetEnumerator();
```
stephentoub: List<T>.GetEnumerator returns a struct Enumerator.  By storing it into temp typed as IEnumerator like this, you're boxing each of those structs that comes back.  Is that intentional?

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        [Benchmark]
+        [MemberData("TestLists")]
+        public void Enumerator_Getcurrent(List<object> list)
+        {
+            object temp;
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                IEnumerator enumerator = list.GetEnumerator();
```
stephentoub: Similarly, unless the JIT is able to devirtualize this, by storing into ```IEnumerator``` rather than into ```List<object>.Enumerator```, all of these Current calls are interface calls.

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -0,0 +1,198 @@
...
+        [Benchmark]
+        [MemberData("TestLists")]
+        public void Enumerator_MoveNext(List<object> list)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                IEnumerator enumerator = list.GetEnumerator();
+                using (iteration.StartMeasurement())
+                {
+                    enumerator.MoveNext(); enumerator.MoveNext(); enumerator.MoveNext(); enumerator.MoveNext();
```
stephentoub: Same here for MoveNExt.

src/System.Diagnostics.Process/tests/Performance/Perf.ProcessStartInfo.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        [Benchmark]
+        public void SetUseShellExecute()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                ProcessStartInfo info = new ProcessStartInfo();
+                using (iteration.StartMeasurement())
+                    info.UseShellExecute = false;
+            }
```
stephentoub: I appreciate the thoroughness, but I'm not sure we need microbenchmarks on all of these things, e.g. on the perf of setting UseShellExecute.

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -0,0 +1,66 @@
...
+        [MemberData("GZTestFiles")]
+        public async void Decompress(string testFilePath)
+        {
+            int _bufferSize = 1024;
+            var bytes = new byte[_bufferSize];
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                int retCount = -1;
+                using (var gzStream = await LocalMemoryStream.readAppFileAsync(testFilePath))
+                using (var strippedMs = StripHeaderAndFooter.Strip(gzStream))
```
stephentoub: Could you move this initialization to outside of the iterations, and then just reset the position of the memory stream to 0 on every iteration?

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -0,0 +1,66 @@
...
+                }
+            }
+        }
+
+        [Benchmark]
+        [InlineData(100)]
+        [InlineData(1000)]
+        public void Compress(int numberOfWrites)
+        {
+            var bytes = Text.Encoding.UTF8.GetBytes(PerfUtils.CreateString(100));
```
stephentoub: Writing the same repeated short sequence of data could make the data quite compressible and thus impact the performance of the Compress operation.  You might want a few tests here, one for very compressible data (e.g. a few characters repeated over and over), one for very non-compressible data (e.g. cryptographically random generated content), and one somewhere in the middle.

src/System.Runtime.Extensions/tests/Performance/Perf.Environment.cs
```diff
@@ -0,0 +1,82 @@
...
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    line = Environment.NewLine; line = Environment.NewLine; line = Environment.NewLine;
+                    line = Environment.NewLine; line = Environment.NewLine; line = Environment.NewLine;
+                    line = Environment.NewLine; line = Environment.NewLine; line = Environment.NewLine;
+                }
+            }
+        }
```
stephentoub: Another example of where I appreciate the goal of completeness, but it's probably not worth having a perf test for this.  NewLine just returns a constant string, so when this gets JIT'd, this is effectively going to be equal to (on Windows) ```line = "\r\n"; line = "\r\n"; line = "\r\n"``` etc.

src/System.Runtime.Extensions/tests/Performance/Perf.Environment.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                // setup the environment variable so we can read it
+                Environment.SetEnvironmentVariable(env, "value");
+
+                // read the valid environment variable
+                foreach (var iteration in Benchmark.Iterations)
+                {
+                    using (iteration.StartMeasurement())
+                    {
+                        Environment.ExpandEnvironmentVariables("%" + env + "%"); Environment.ExpandEnvironmentVariables("%" + env + "%");
```
stephentoub: The ```"%" + env + "%"``` should be separated out into a string created at the beginning of the method , rather than creating that string on each call to ExpandEnvironmentVariables.

src/System.Runtime.Extensions/tests/Performance/Perf.Random.cs
```diff
@@ -0,0 +1,53 @@
...
+                    new Random(); new Random(); new Random();
+                    new Random(); new Random(); new Random();
+                    new Random(); new Random(); new Random();
+                }
+        }
+
+        [Benchmark]
+        public void Next_int()
+        {
+            Random rand = new Random();
```
stephentoub: For a bit more consistency on these Next tests, I'd suggest constructing the Random with a fixed seed.

src/Common/tests/Performance/PerfUtils.cs
akoeplinger: In addition @stephentoub's point, I'm wary of allowing randomness in a benchmark. Sounds like the thing you don't want there (or at least be able to control the seed so it's reproducible).

src/System.Runtime/tests/Performance/Perf.GC.cs
```diff
@@ -0,0 +1,28 @@
...
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                // obj has no managed resources to free, so suppressing the finalizer for it is fine.
+                object obj = new object();
+
+                using (iteration.StartMeasurement())
+                {
+                    GC.SuppressFinalize(obj); GC.SuppressFinalize(obj); GC.SuppressFinalize(obj);
+                    GC.SuppressFinalize(obj); GC.SuppressFinalize(obj); GC.SuppressFinalize(obj);
+                    GC.SuppressFinalize(obj); GC.SuppressFinalize(obj); GC.SuppressFinalize(obj);
```
stephentoub: Is the cost of SuppressFinalize on an object that's already had finalization suppressed the same as on one that hasn't?  I wouldn't be surprised if the answer is no, in which case 8 of these 9 calls would behave differently from what I think you're trying to test.

Additionally, System.Object doesn't actually have a finalizer (even though it exposes the virtual Finalize method), so there's nothing to suppress here.  I imagine you'd instead want to define your own helper class, e.g.
```C#
private class ObjectWithFinalizer
{
    ~ObjectWithFinalizer() { }
}
```
and use that instead of Object.

src/System.Runtime/tests/Performance/Perf.IntPtr.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Benchmark]
+        public void GetZero()
+        {
+            IntPtr ptr;
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    ptr = IntPtr.Zero; ptr = IntPtr.Zero; ptr = IntPtr.Zero;
+                    ptr = IntPtr.Zero; ptr = IntPtr.Zero; ptr = IntPtr.Zero;
+                    ptr = IntPtr.Zero; ptr = IntPtr.Zero; ptr = IntPtr.Zero;
```
stephentoub: IntPtr.Zero is a field.  There's no code being tested here :smile:

src/System.Runtime/tests/Performance/Perf.String.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        [Benchmark]
+        [MemberData("TestStringSizes")]
+        public void Split(int size)
+        {
+            string testString = PerfUtils.CreateString(size);
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    testString.Split("-");
```
stephentoub: Presumably if you end up changing the string generation to not use Guid.NewGuid, some of these string tests will need to change as well.

src/System.Runtime/tests/Performance/Perf.String.cs
```diff
@@ -0,0 +1,243 @@
...
+        public void Empty(int size)
+        {
+            string result;
+            string testString = PerfUtils.CreateString(size);
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                {
+                    result = string.Empty; result = string.Empty; result = string.Empty;
+                    result = string.Empty; result = string.Empty; result = string.Empty;
+                    result = string.Empty; result = string.Empty; result = string.Empty;
```
stephentoub: Another field :wink:

src/System.Runtime/tests/Performance/Perf.String.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        [Benchmark]
+        [MemberData("TestStringSizes")]
+        public void ToLower(int size)
+        {
+            string testString = PerfUtils.CreateString(size);
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    testString.ToLower();
```
stephentoub: Guid.NewGuid().ToString() produces a string with numbers and all lowercase letters... this may not be testing what you want to be testing.

src/System.Runtime/tests/Performance/Perf.String.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        [Benchmark]
+        [MemberData("TestStringSizes")]
+        public void Trim(int size)
+        {
+            string testString = "   " + PerfUtils.CreateString(size) + "   ";
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    testString.Trim();
```
stephentoub: In general, we should consider having some perf tests that exist to "lock in" optimizations we have in place.  For example, Trim is optimized to return the same string instance if nothing needs to be trimmed.

src/System.Text.Encoding/tests/Performance/Perf.Encoding.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Encoding enc = Encoding.UTF8;
+            string toEncode = PerfUtils.CreateString(size);
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    enc.GetBytes(toEncode); enc.GetBytes(toEncode); enc.GetBytes(toEncode);
+                    enc.GetBytes(toEncode); enc.GetBytes(toEncode); enc.GetBytes(toEncode);
+                    enc.GetBytes(toEncode); enc.GetBytes(toEncode); enc.GetBytes(toEncode);
```
stephentoub: Another example where the choice of strings could impact perf, e.g. 'a'-'f' and '0'-'9' are all ASCII.

src/System.Threading/tests/Performance/Perf.EventWaitHandle.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void Set()
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (EventWaitHandle are = new EventWaitHandle(false, EventResetMode.AutoReset))
+                using (iteration.StartMeasurement())
+                {
+                    are.Set(); are.Set(); are.Set();
+                    are.Set(); are.Set(); are.Set();
+                    are.Set(); are.Set(); are.Set();
```
stephentoub: Consider changing this to be a SetReset test, such that all of the repeated operations do Set(); Reset(); rather than having the first one transition it from not-set to set and then all of the rest be nops in terms of changing the signaled state.

src/System.Xml.XmlDocument/tests/Performance/Perf.XmlNodeList.cs
```diff
@@ -0,0 +1,53 @@
...
+                var doc = new XmlDocument();
+                doc.LoadXml("<a><sub1/><sub2/></a>");
+                XmlNodeList list = doc.DocumentElement.ChildNodes;
+
+                // Actual perf testing
+                using (iteration.StartMeasurement())
+                {
+                    list.GetEnumerator(); list.GetEnumerator(); list.GetEnumerator();
+                    list.GetEnumerator(); list.GetEnumerator(); list.GetEnumerator();
+                    list.GetEnumerator(); list.GetEnumerator(); list.GetEnumerator();
```
stephentoub: In these tests where you're doing GetEnumerator (here and elsewhere), is GetEnumerator really the thing we want to be testing, or is it the cost of iterating via that enumerator, e.g. by using a foreach loop to go through all of the contents?

src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
mellinoe: This is probably because the Microsoft.xunit.performance assemblies are built as v5.0 portable assemblies, which forces any referencing assemblies to do the same.

dir.props
```diff
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00089</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00091</BuildToolsVersion>
```
weshaggard: FYI - I'm going to need to publish a new buildtools package today and update it so it may conflict with this change depending one when you get yours merged.

src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
mellinoe: @jhendrixMSFT  brought up a good point in that this will force us to support VS 2015 when building this project. I think this is a step that we are going to have to take eventually, but we might not be ready to do it know. Thoughts? @weshaggard .

I think for the "xunit.performance.core", which is the assembly that test projects actually reference, we may be able to get away with building against an older portable profile. The xunit.performance.execution assembly, which we don't reference directly here, probably needs to still be built as a v5.0 portable project. But that can probably be separated into a different Nuget package which we don't reference from individual test projects. Anyways, that would be quite a bit of fiddling just so that we don't have to use v5.0 portable projects.

weshaggard: I don't think we should need to set these. If that isn't the case please let me know why because I would like to understand it better.

mellinoe: ResolveAssemblyReferences gives this warning otherwise (and no reference to the assembly is allowed):
````
warning MSB3274: The primary reference "xunit.performance.core" could not be resolved because it was built against the ".NETPortable,Version=v5.0" framework. This is a higher version than the currently targeted framework ".NETPortable,Version=v4.5,Profile=Profile7".
````
The performance assemblies are built like so:
https://github.com/Microsoft/xunit-performance/blob/master/src/xunit.performance.core/xunit.performance.core.csproj#L18

Comment (jhendrixMSFT) [inline src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj]: It also appears to require a VS option that's off by default (it was for me in VS 2015 Enterprise).  I had to enable Universal Windows App Development Tools in order to build.

File: src/Common/tests/Performance/PerfUtils.cs (hunk 1 above)
Comment (ianhays) [inline src/Common/tests/Performance/PerfUtils.cs]: Agreed - Eric Mellino made a similar comment on my last PR but in my haste to get this PR up I forgot to change it. I will modify the method to use a Random with static seed and go through the tests that reference this method to ensure they make sense with a wider range of possible characters.

File: src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs (hunk 2 above)
Comment (ianhays) [inline src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs]: There isn't any reason for the add/remove to be per-iteration. I have a feeling that I probably do this elsewhere as well since the per-iteration syntax changed after I wrote some of the tests, so I'll go through and make sure to maintain the distinction between per-iteration setup and per-test setup.

Comment (ianhays) [inline src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs]: >and then get rid of the Remove?

This I'm a bit torn on. In their current state the tests will not modify the collections passed to them as MemberData. I wrote them this way so that we could potentially cache the collections used as MemberData to greatly speed up per-test initialization. However, I'm a bit concerned about future-proofing and possible errors: if a test is added that modifies a MemberData collection that is later used by a different test, it could skew those results.

What do you think? Is the speed-up worth the risk for error?

File: src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj (hunk 3 above)
Comment (ianhays) [inline src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj]: Any other thoughts on this? It isn't immediately obvious how to get the v5.0 libraries necessary to allow opening/running these tests (VS will open a link to Microsoft.com when you tell it to download the v5.0 targeting pack), but I'm uncertain of a workaround.

File: src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs (hunk 2 above)
Comment (stephentoub) [inline src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs]: > What do you think? Is the speed-up worth the risk for error?

My suggestion was mainly to remove unnecessary code.  If you believe it will be necessary, feel free to keep it.

File: src/System.Xml.XmlDocument/tests/Performance/Perf.XmlNodeList.cs (hunk 27 above)
Comment (ianhays) [inline src/System.Xml.XmlDocument/tests/Performance/Perf.XmlNodeList.cs]: For completeness sake I added tests for all items from the usage data that were highly utilized. I tried to filter out the obviously unnecessary ones, but as you noticed I missed some :) I took a very broad approach to my first-pass to get some tests out into the open for other people to start playing around with, but it's likely that ~20% of these tests will be removed for one reason or another (e.g. the field tests). 

The GetEnumerator tests are instances of this as well. GetEnumerator <i>is</i> what I intended to test, but the question of whether or not I <i>should</i> test it slipped by me. Now that these tests are in the open, I need to go back through them and for each test ask: <i>is this really what I care about performance for?</i> In the case of GetEnumerator I think your reasoning is correct that if GetEnumerator is highly utilized then we should perf test it, but GetEnumerator is frequently enough accompanied by iteration that we can instead perf test that.

File: src/System.Collections/tests/Performance/Perf.Dictionary.cs (hunk 6 above)
Comment (ianhays) [inline src/System.Collections/tests/Performance/Perf.Dictionary.cs]: I like that idea. I could add a custom string class that uses the same GetHashCode/Equals but adds a fixed delay to them so that we have a good count of the number of times the functions are called.

A simpler approach would be to change the keys the be randomly generated using PerfUtils.CreateString(50) so that they take longer to Equals/GetHashCode, but that may not take long enough.

File: src/System.Collections/tests/Performance/Perf.List.cs (hunk 10 above)
Comment (ianhays) [inline src/System.Collections/tests/Performance/Perf.List.cs]: Good catch; not intentional. Also, rather than have separate tests for GetEnumerator, GetCurrent, and MoveNext I think a single test with a foreach through the elements would be adequate.

File: src/System.Diagnostics.Process/tests/Performance/Perf.ProcessStartInfo.cs (hunk 13 above)
Comment (ianhays) [inline src/System.Diagnostics.Process/tests/Performance/Perf.ProcessStartInfo.cs]: Agreed. These setters are a good example of where utilization data can be less helpful: the setters may be used by a lot of the assemblies but they're probably not used very often within those assemblies. I'm going to remove them.

File: src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj (hunk 3 above)
Comment (jhendrixMSFT) [inline src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj]: Can this be conditioned so that if the requirements aren't met (i.e. you aren't building with VS 2015 + Universal Windows App Dev Tools) the perf tests are omitted from the build?  Make it pay-for-play in the short term, and document how to set up a build environment to enable building perf tests.  We should also ensure that this will work with VS 2015 Community Edition.

CC @joshfree

File: src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs (hunk 14 above)
Comment (ianhays) [inline src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs]: DeflateStream disposes the stream from which it was created, so we cannot reuse the "strippedMs" stream.

The Strip function disposes the stream passed to it as input.

File: src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj (hunk 3 above)
Comment (mellinoe) [inline src/System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj]: The problem is that you can't even build the test project at all, let alone run the tests, if you don't have VS 2015 installed. Like @ianhays was saying, though, we could probably only include the performance source files if you have Performance=true in your build settings. You would still get the warning about referencing the v5.0 portable project (because it's pulled in by the project.json file), but it would still be able to build the rest of the source files.

File: src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs (hunk 14 above)
Comment (stephentoub) [inline src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs]: > DeflateStream disposes the stream from which it was created

Not if you pass leaveOpen: true to the ctor.

Comment (ianhays) [inline src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs]: Thought that seemed like odd behavior :) WillFix.

File: src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs (hunk 15 above)
Comment (ianhays) [inline src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs]: Good idea. Implementing it now.

File: src/System.Runtime/tests/Performance/Perf.GC.cs (hunk 19 above)
Comment (ianhays) [inline src/System.Runtime/tests/Performance/Perf.GC.cs]: Since this essentially a setter, I'm going to remove the perf tests for it at least for now.

File: src/System.Runtime/tests/Performance/Perf.String.cs (hunk 24 above)
Comment (ianhays) [inline src/System.Runtime/tests/Performance/Perf.String.cs]: Added for Trim. Added to "To-Do" list for other methods as well :)

**Comment by ianhays** [comment]
Per the discussion above and through email, I've removed the fixed requirement for TargetFramework=v5.0. Now the TargetFramework will only be set if Performance=true and the v5.0 libraries are available on the user's machine. If they aren't available, I throw a detailed message with a link to a new wiki page.

This goes along with a buildtools PR: https://github.com/dotnet/buildtools/pull/284.

File: dir.props [hunk 29]
```diff
@@ -261,6 +261,11 @@
     <MakePriExtensionPath>$(_WindowsPhoneKitBinPath)\x86\MrmEnvironmentExtDl.dll</MakePriExtensionPath>
     <MakePriExtensionPath_x64>$(_WindowsPhoneKitBinPath)\x64\MrmEnvironmentExtDl.dll</MakePriExtensionPath_x64>
   </PropertyGroup>
+  
+  <!-- Perf test setup - Required until v5.0 is mandatory across CoreFX-->
+  <PropertyGroup>
+    <RunPerfTestsForProject Condition="'$(Performance)' == 'true' and '$(OS)' == 'Windows_NT' and Exists('C:\Program Files (x86)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets')" >true</RunPerfTestsForProject>
```
ianhays: Will modify Exists to instead check:

```Exists('$(MSBuildProgramFiles32)\MSBuild\Microsoft\Portable\v5.0\Microsoft.Portable.CSharp.targets'```

**Comment by ianhays** [comment]
@dotnet-bot test this please

**Comment by ianhays** [comment]
Getting a test failure on Windows for "System.Threading.Tasks.Tests.WaitAllAny.TaskWaitAllAny.TaskWaitAllAny41"

**Comment by stephentoub** [comment]
@dotnet-bot test this please (#3388)

File: src/Common/tests/Performance/PerfUtils.cs [hunk 30]
```diff
@@ -0,0 +1,42 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.IO;
+
+namespace System
+{
+    /// <summary>Static helper class for performance tests</summary>
+    public static class PerfUtils
+    {
+        private static Random rand = new Random(1234132);
```
ellismg: Can multiple benchmarks run concurrently?  If so, couldn't interleaving mean that different runs a suite test different scenarios?  Is that good or bad (given that we have a fixed seed here it feels like you want things to be deterministic across runs...)

ianhays: They can not run concurrently, no. I'll check up to make sure this is still the case, but last time I checked the Benchmark class made use of Windows ETW tracing which concurrent testing would interfere with.

ianhays:
@dotnet-bot test this please

src/Common/tests/Performance/PerfUtils.cs
ericeil: Benchmarks are not run concurrently.  But there's no guarantee about the *order* in which they run, so if this is shared among several tests they may get different results depending on ordering.

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

src/Common/tests/Performance/PerfUtils.cs
stephentoub: @ianhays, it will be shared amongst multiple tests,right?  That seems problematic, given that the performance of some of the tests will be dependent on the data that gets generated.  I suggest we merge this PR, but then we go back and change this support to be something that each perf test instantiates, creating its own instance with a fixed seed, so that test ordering won't impact performance in this way.

ianhays: Yes, it currently is shared between tests. 

> I suggest we merge this PR, but then we go back and change this support
My thoughts exactly.

ianhays:
LGTM?

stephentoub:
With the planned follow-ups, LGTM.

PR #3309, opened 2015-09-18
Test for exception on cast of null to non-nullable type.

JonHanna:
Since it was decided in #2947 that the current behaviour is desired, and should
be documented as the correct behaviour, there should be a test that checks this
behaviour continues, to avoid regression changes to it.


stephentoub:
@dotnet-bot test this please (lots of FindPackagesById errors)

stephentoub:
LGTM. Thanks.

PR #3313, opened 2015-09-18
[WIP] Avoid race in GetEnumerator with Interlocked rather than ThreadID check

JonHanna:
`Iterator` stores the initial `CurrentManagedThreadId` and then checks it is the same in `GetEnumerator`. This allows it to avoid a race in seeing if it can be used as the enumerator returned. (It can only do this if it has not been used, and it resolves the race of two threads calling it at the same time by only allowing this optimised path to the initial thread).

This approach is copied from the iterators the compiler produces for `yield return`.

This means a total of 2 calls to `CurrentManagedThreadId` in the most common case and another if either another thread had made the call, or if this is not the first call.

Another possible approach is to set the state to other than "unused" with an interlocked operation.

This appears to cost about twice as much as `CurrentManagedThreadId` (from tests on Windows 64-bit), so on balance is about equivalent to making those two `CurrentManagedThreadId` calls, and so wouldn't gain much in most uses of `yield`-produced code (one `GetEnumerator` per use, same thread).

With the `Interator` subclasses however, often the point of these classes is that they might not have `GetEnumerator` called on them at all, but serve to help create yet another `Iterator` subclass that takes over its enumeration rôle entirely. In such cases, the call to `CurrentManagedThreadId` is a waste.

Hence using `Interlocked.CompareExchange` to avoid the race is a win in such cases.

A further benefit is that it also allows the first thread to attempt to call `GetEnumerator()` to win the race, regardless of whether it is the thread that created the enumerable or not (a situation that is increasingly common if a query is created before an `await` and used after it), so avoiding an extra allocation.


JonHanna:
Trying a modified version of the tests for #3104, this seems to have a negligible effect on uses that do enumerate the query (a very slight improvement on average, but probably less than the margin of error) but improves uses that don't enumerate the query by about 8-9%.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -103,10 +97,11 @@ public virtual void Dispose()
                 current = default(TSource);
                 state = -1;
             }
-
+            
             public IEnumerator<TSource> GetEnumerator()
             {
-                Iterator<TSource> enumerator = state == 0 && _threadId == Environment.CurrentManagedThreadId ? this : Clone();
+                if (Interlocked.CompareExchange(ref state, 1, 0) == 0) return this;
```
sharwell: :bulb: The `return this;` should be on the next line:

```csharp
if (Interlocked.CompareExchange(ref state, 1, 0) == 0)
    return this;
```

sharwell:
I like this change :+1:

src/System.Linq/src/System/Linq/Enumerable.cs
JonHanna: Not if I'm being consistent with the rest of the file.

sharwell: I was just going by rule 1 in the [Coding Styles](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) document.

JonHanna: Yep, and it's a good rule, IMO, but I was going by rule 9.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -103,10 +97,11 @@ public virtual void Dispose()
                 current = default(TSource);
                 state = -1;
             }
-
+
```
nguerrera: nit: whitespace added (inadvertently, I presume).

nguerrera:
cc @VSadov

stephentoub:
This generally LGTM.  Another advantage not measured here is that it removes a field.

I do have a vague recollection that something else was used here previously instead of thread ID (maybe an interlocked?).  @VSadov, do you remember?

One additional concern is any potential perf impact on other platforms often more sensitive to synchronization, e.g. ARM.

JonHanna:
Removing the field was what first occurred to me, but harder to measure the practical impact (limited tests would suggest just removing the allocation avoidance and allocating every time is the way to go, but more intense long term multi-threaded use would presumably suffer in scalability).

src/System.Linq/src/System/Linq/Enumerable.cs
VSadov: If we go with this, it might need to be
 if (state == 0 &&  Interlocked.CompareExchange(ref state, 1, 0) == 0)

test-and-CAS is often cheaper than just CAS.

sharwell: @VSadov 

* If state is 0, we have to do the CAS anyway.
* If state is not 0, we are going to allocate a new reference type, which is more expensive that a compare exchange.

I don't think test-and-CAS can save anything here, and actually introduces additional work for the most common path.

stephentoub:
@VSadov, what do you want to do here?

VSadov:
I'd like to do some research here. This is actually something that I wanted to for some time. It applies both to the linq enumerables and to the codegen of iterators in C#.

We have the following options:

1) do not do this optimization - just new up a new iterator.

The cost of allocations on today's systems is roughly proportional to the number of bytes allocated (number of objects or references generally have second degree impact).
Combined enumerator/iterator saves an object header since we have n-1 objects and n is typically 1.
However, we are also carrying fields for the initial data and that is a waste if n > 1

Simple math shows that this optimization could be on the edge of profitability and may no longer be important.
I'd like to revisit this.

There is also a subtle life time problem here - see https://github.com/dotnet/roslyn/issues/376

2) if we do optimization, do we use ThreadId or Interlocked?

Interlocked saves us an int field, but introduces extra bus traffic if a lot of iterators are created. The bus traffic would have impact only on massively concurrent scenarios, though, and can be mitigated by testing the state before doing interlocked.

ThreadId has an advantage of not requiring any synchronization (is that actually correct?) and will scale better, but the penalty is an extra int field.

Both solutions introduce dependency on additional APIs. At least in C# iterator case that is an inconvenience - one would need to add reference to the Threading contract, just to use iterators, or compiler needs to fallback to alternative codegen strategy. 

That is similar to event handling - we would like to use interlocked, but as we discovered on CoreFx, Interlocked is not longer a part of the core lib, so it may be not available, ind it turns out that user can be very confused to get "missing helper" compile errors about interlocked when he deal with events. So dependency on additional APIs is a bit of a problem.

VSadov:
Basically, I'd like to spend some time with this before we do fixes. Perhaps we do not want this optimization at all...

With all the bugs I need to deal with and such, it might take several days for me to get to this though.
Would it be ok, if we keep this open for a while?

VSadov:
Any additional perf data on these options would also be very helpful.

stephentoub:
> it might take several days for me to get to this though.

Sounds good

svick:
@VSadov Would it be possible to move `Interlocked` or any other necessary types/members into the core library?

VSadov:
@svick  -  Fx is generally moving in an opposite direction. Away from big monolitic cor library and towards more modularized environments. 

In particular about threading - it is not hard to imagine a platform that does not support or expose any notion of concurrency. I believe .Net Micro FX is one of such examples (or was at some point).

It is definitely possible to make some kind of System.Threading that would simulate a one-and-only thread environment, but it seems nicer if that comes as an optional library, not as a part that is "always there".

JonHanna:
> do not do this optimization - just new up a new iterator.

I did experiment with that, and it seems to win clearly whether single-threaded or maxing out my cores, but I was very hesitant to up allocations.

However, what scenarios have both the enumerable and the enumerator living long? A lot of the time the enumerable is just to set up the enumerator to be looped through and can become collected by the time the enumeration is happening, whereas on the other hand it's almost only the enumerable that will be held somewhere like a cache that keeps it alive a long time. This would suggest that **maybe** the allocation avoidance is more trouble than it's worth.

> There is also a subtle life time problem here - see dotnet/roslyn#376

I don't think that's as big a concern here, because the more potentially heavy bits of state are all cleared explicitly. Still, allowing things to be cleaned up quickly remains an argument for just creating fresh objects, even when they aren't likely to be as large.

> Combined enumerator/iterator saves an object header since we have n-1 objects and n is typically 1.
However, we are also carrying fields for the initial data and that is a waste if n > 1

And often one or more of those fields is only needed in the enumerable or only needed in the enumeration, and they could be separate types, though that has other costs.

> Interlocked saves us an int field, but introduces extra bus traffic if a lot of iterators are created. The bus traffic would have impact only on massively concurrent scenarios, though, and can be mitigated by testing the state before doing interlocked.

CurrentManagedThreadId is also relatively expensive, as I said I found it to be about half the cost of a CAS. I didn't try heavy mutlicore use though, so I don't have any measures of comparable costs with heavy contention. I'm also not well-equipped to measure anything but x86_64.

As well as saving us an int field, it also saves us work in the constructor that might not be made use of, which I came across first when experimenting with an idea that upped the constructor calls as a way to perhaps get some of the idea in #2844 without leaking the interfaces to calling code (and potentially therefore spend an allocation to lower the total number of allocations). I started some rather convoluted attempts to pass the threadid around rather than obtain it again before trying to just remove it entirely with this PR. Whether that experiment comes to anything is another matter, but if it was hitting CurrentManagedThreadId for no good purpose anyway, and so does other code, with a lot of the current code avoiding `GetEnumerator()` being called on some of the objects, and as such avoiding the point of knowing what the threadid was on object creation.

Anyway, it's saving that call that made the most obvious measurable improvement.

> it is not hard to imagine a platform that does not support or expose any notion of concurrency

True, but if such a platform was strictly single-threaded it could implement its own cheap cas as:

    public static int CompareExchange(ref int location1, int value, int comparand)
    {
      // Single-threaded cheating version.
      int old = location1;
      if (old == comparand)
        location1 = value;
      return old;
    }

And then code using it for the potential of other threads affecting it gets the behaviour it expects as observed from its sole thread, without the CAS costs. This goes against the spirit of pluggable threading support, but so many cases have to consider the threading issues they might have to deal with that they might end up dragging in such dependencies anyway.

Anyway, all that said and back to topic, (and I'm regretting opening a PR rather than an issue now), does anyone know what the worse cases are for just removing the allocation-avoidance entirely? I'd like to see how doing that compares in the case it serves worse.

JonHanna:
> We have the following options:

I'm going to set up some branches that take different approaches, marking this WIP in the meantime.

JonHanna:
Closing this for now. Whichever of the branches mentioned in #3481 seems to be the best will become a new PR.

PR #3325, opened 2015-09-19
Improve performance of BigInteger.Log

axelheer:
Since only the most significant bits of a big number are relevant for computing its logarithm (the current implementation multiplies with 0.5 after processing a bit, which naturally doesn't have any effect after 54 iterations), using the build-in LOG function just for the highest 64-bit block and adjusting the result considering the actual bit length should be sufficient.

#1307

### DIFF
--- a/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
+++ b/src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
@@ -990,26 +990,20 @@ namespace System.Numerics
             if (value._bits == null)
                 return Math.Log((double)value._sign, baseValue);
 
-            Double c = 0, d = 0.5D;
-            const Double log2 = 0.69314718055994529D;
+            ulong h = value._bits[value._bits.Length - 1];
+            ulong m = value._bits.Length > 1 ? value._bits[value._bits.Length - 2] : 0;
+            ulong l = value._bits.Length > 2 ? value._bits[value._bits.Length - 3] : 0;
 
-            int uintLength = Length(value._bits);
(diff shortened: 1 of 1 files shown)


axelheer:
cc: @stephentoub @KrzysztofCwalina

src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
```diff
@@ -990,26 +990,20 @@ public static Double Log(BigInteger value, Double baseValue)
...
-                }
-                indbit = 0x80000000;
-            }
-            return (Math.Log(c) + log2 * bitlen) / Math.Log(baseValue);
+            // extract most significant bits
+            ulong x = (h << 32 + c) | (m << c) | (l >> 32 - c);
+
+            // let v = value, b = bit count, x = v/2^b-64
+            // log ( v/2^b-64 * 2^b-64 ) = log ( x ) + log ( 2^b-64 )
+            return Math.Log(x, baseValue) + Math.Log(Math.Pow(2, b - 64), baseValue);
```
axelheer: Looking at it again, we could replace `Math.Log(Math.Pow(2, b - 64), baseValue)` with `(b - 64) / Math.Log(baseValue, 2)`. Which is the nicer solution?

stephentoub:
Thanks, @axelheer. "should be sufficient"?  Otherwise, LGTM.

axelheer:
Is sufficient, of course. Forgive my English, at my place people like using subjunctive forms too much. ;)

stephentoub:
Ok, thanks, just verifying :smile:

PR #3330, opened 2015-09-20
Restore Tool Runtime as part of root src traversal

weshaggard:
There are cases where multiple projects building in parallel try
to restore the tools runtime at the same time which cause some
copying over and potential failed writes. This add the restoring
of the tool runtime as part of the root traversal project so
the other projects will not have any work to do.


stephentoub:
LGTM

PR #3333, opened 2015-09-20
Remove faulty assert from CurlHandler.CurlResponseStream

### LINKED ISSUE #3319: CurlHandler asserting in WCF tests
```
If disposed, the stream must also be completed.
---- Assert Long Message ----

   at System.Diagnostics.Debug.Assert(Boolean condition, String message)
   at System.Net.Http.CurlHandler.CurlResponseStream.VerifyInvariants()
   at System.Net.Http.CurlHandler.CurlResponseStream.SignalComplete(Exception error)
   at System.Net.Http.CurlHandler.EasyRequest.Cleanup()
   at System.Net.Http.CurlHandler.MultiAgent.FinishRequest(EasyRequest completedOperation, Int32 messageResult)
   at System.Net.Http.CurlHandler.MultiAgent.WorkerLoop()
   at System.Net.Http.CurlHandler.MultiAgent.<>c.<EnsureWorkerIsRunning>b__13_0(Object s)
   at System.Threading.Tasks.Task.Execute()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
   at System.Threading.Tasks.Task.ExecuteEntry(Boolean bPreventDoubleExecution)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
```

stephentoub:
CurlResponseStream.VerifyInvariants has an assert that assumes that if the stream has been disposed, it must have also completed.  But since this stream is exposed publically, it's possible for consuming code to Dispose of the stream before the request has finished, violating the assert.  In such a situation, a subsequent attempt to transfer data to the response stream will fail and cause the request to be aborted as desired, cleaning up resources (there are probably some opportunities here for some optimizations to handle this case, but that can be explored later.)

This commit just removes the bad assert.

Fixes #3319 
cc: @kapilash, @roncain

### DIFF
--- a/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
+++ b/src/System.Net.Http/src/System/Net/Http/Unix/CurlHandler.CurlResponseMessage.cs
@@ -446,8 +446,6 @@ namespace System.Net.Http
                 Debug.Assert(!((_remainingDataCount > 0) && (_pendingReadRequest != null)), "We can't have both remaining data and a pending request.");
                 Debug.Assert(!((_completed != null) && (_pendingReadRequest != null)), "We can't both be completed and have a pending request.");
                 Debug.Assert(_pendingReadRequest == null || !_pendingReadRequest.Task.IsCompleted, "A pending read request must not have been completed yet.");
-
-                Debug.Assert(!_disposed || _completed != null, "If disposed, the stream must also be completed.");
             }
 
             /// <summary>State associated with a pending read request.</summary>


stephentoub:
@kapilash or @vijaykota, can you review?

kapilash:
> But since this stream is exposed publically, it's possible for consuming code to Dispose of the stream before the request has finished, violating the assert. In such a situation, a subsequent attempt to transfer data to the response stream

In order to repro the problem, we need something more than the following.. no?

      	using (var handler = new HttpClientHandler()) {
            var client = new HttpClient(handler);
            HttpResponseMessage response = await client.GetAsync(redirectUri);
            response.Content.Dispose();
       }

Perhaps the response stream should be large enough?

kapilash:
In general, I agree with the logic about removing the assert. LGTM please. Thank you.

stephentoub:
> n order to repro the problem

GetAsync(url) uses HttpCompletionOption.ResponseContentRead by default, so the content is going to be read and buffered before it's returned.  I believe you'd need to use ResponseHeadersRead, and of course then there are timing issues with regards to the race condition between disposing of the stream and the content being downloaded.

kapilash:
OK, thanks. (and about the code fragment, it was copy paste error.. I tried  with HttpCompletionOption.ResponseHeadersRead)

vijaykota:
LGTM

Issue #3224, opened 2015-09-13
Expression compiler - Y combinator issue


bartdesmet:
The following fails to compile correctly:

Expression<X> f = Expression.Lambda<X>(Expression.Empty(), Expression.Parameter(typeof(X)));
var a = Expression.Lambda(Expression.Invoke(f, f));
var m = typ.DefineMethod("A", MethodAttributes.Public | MethodAttributes.Static);
a.CompileToMethod(m);

where X is defined as:

public delegate void X(X a);

The object identity of 'f' used in the invocation expression seems the culprit due to some caching in the lambda compiler.

PR #3341, opened 2015-09-21
Fix incorrect codegen for self-application of recursive expressions.

VSadov:
ET binder associates scope information with corresponding lambda nodes, which is valid in most cases since variables that belong to lambda depend only on the lambda contents.
Except that when lambda is inlined and parameters become locals.

This change makes scopes for inlined lambdas to be associated with parent invocation nodes, which disambiguates them from scopes for lambdas in their regular non-inlined form.

Fixes: #3224


stephentoub:
@dotnet-bot test this please (known crypto test issue)

VSadov:
@AlekseyTs @DiniV - could you take a look?

VSadov:
@AlekseyTs @DinoV - could you take a look?

AlekseyTs:
LGTM

PR #3342, opened 2015-09-21
Let shim handle platform support of posix_fadvise

nguerrera:
Call the shim always and ignore ENOTSUP. This will let new platforms
with posix_fadvise do the right thing without code changes.

Also, remove unused declarations from */Interop.Libraries.cs.

cc @stephentoub @sokket


stephentoub:
LGTM. Thanks, Nick.

PR #3345, opened 2015-09-21
Net Secure Xplat Implementation

rajansingh10:
1. Abstracted the code as per design shared earlier.
2. Made SecureChannel & SSPIWrapper agnostic of os.
3. Created 1 class 1 file for each new classes & move them to windows specific folder.
4. Added new interfaces & modules related to Cert.
5. SSPINegoInterface & SSPIAuthType, currently are not part of csproj as suggested by cipop.

Have included unix Interop & related code as well for completeness. You can choose to review Windows first as well.


rajansingh10:
CC: @stephentoub, @davidsh, @CIPop, @SidharthNabar, @pgavlin, @bartonjs

rajansingh10:
CC:@vijaykota

src/Common/src/Interop/Unix/Interop.Libraries.cs
```diff
@@ -6,13 +6,15 @@ internal static partial class Interop
     private static partial class Libraries
     {
         internal const string Libc = "libc";                   // C library
-        internal const string LibCoreClr= "libcoreclr";        // CoreCLR runtime
+        internal const string LibCoreClr = "libcoreclr";        // CoreCLR runtime
```
stephentoub: Nit: spacing of the comment has it now misaligned

src/Common/src/Interop/Unix/Interop.Libraries.cs
```diff
@@ -6,13 +6,15 @@ internal static partial class Interop
...
         internal const string LibCrypto = "libcrypto";         // OpenSSL crypto library
+        internal const string LibSsl = "libssl";               // OpenSSL library
         internal const string Zlib = "libz";                   // zlib compression library
 
         // Shims
         internal const string SystemNative = "System.Native";
-        internal const string HttpNative   = "System.Net.Http.Native";
+        internal const string HttpNative = "System.Net.Http.Native";
+        internal const string SslNative = "System.Net.Security.Native";
```
stephentoub: Does this need to be a separate native library from System.Security.Cryptography.Native?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            sslContext.isServer = isServer;
+
+            try
+            {
+                IntPtr method = GetSslMethod(isServer, options);
+
+                var contextPtr = libssl.SSL_CTX_new(method);
+                if (IntPtr.Zero == contextPtr)
+                {
+                    throw CreateSslException("Failed to allocate SSL/TLS context");
```
stephentoub: Strings like this need to be in the resources.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                }
+
+                libssl.SSL_CTX_ctrl(contextPtr, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
```
stephentoub: SSL_CTX_ctrl returns a long.  Is that an error value?  Do we need to do something with it here?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        //TODO Set remote certificate options
+        internal static IntPtr AllocateSslContext(long options, SafeX509Handle certHandle, bool isServer, bool remoteCertRequired)
+        {
+            IntPtr sslContextPtr = Marshal.AllocHGlobal(Marshal.SizeOf<SslContext>());
+            SslContext sslContext = new SslContext
+            {              
+                sslPtr = IntPtr.Zero,
+                readBioPtr = IntPtr.Zero,
+                writeBioPtr = IntPtr.Zero,
+                isServer = isServer,
```
stephentoub: The first three fields are already initialized to zero and don'tneed to be set to IntPtr.Zero again.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            internal bool isServer;          
+        }
+        #endregion
+
+        #region internal methods
+
+        //TODO Set remote certificate options
+        internal static IntPtr AllocateSslContext(long options, SafeX509Handle certHandle, bool isServer, bool remoteCertRequired)
+        {
+            IntPtr sslContextPtr = Marshal.AllocHGlobal(Marshal.SizeOf<SslContext>());
```
bartonjs: There are lots of things I don't like about this pattern.
* The error path is prone to crashing
* The error path does not FreeHGlobal this variable
* Richly typed data is reduced to IntPtr.

In the future (not this change) we should seek to eliminate this pattern.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                isServer = isServer,
+            };
+
+            sslContext.isServer = isServer;
+
+            try
+            {
+                IntPtr method = GetSslMethod(isServer, options);
+
+                var contextPtr = libssl.SSL_CTX_new(method);
```
bartonjs: Style: `var` is only allowed to be used with `new`, a short line using `as`, or a hard cast. (The type name has to be on the line somewhere).

bartonjs: This should be a SafeHandle, not just an IntPtr.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+
+                libssl.SSL_CTX_free(contextPtr);     
+
+                if (IntPtr.Zero == sslContext.sslPtr)
+                {
+                    throw CreateSslException("Failed to create SSSL object from SSL context");
+                }      
+
+                IntPtr memMethodRead = libcrypto.BIO_s_mem();
+                IntPtr memMethodWrite = libcrypto.BIO_s_mem();
```
stephentoub: Both memMethodRead and memMethodWrite are being initialized to the same thing? Or does BIO_s_mem actually allocate something and return a different result on each call?  If they'd be the same, then we should only call BIO_s_mem once.  If they'd be different, then presumably in the below error check we'd need to free memMethodRead if it was non-null but memMethodWrite was null?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
bartonjs: In the case of `SSL_CTRL_OPTIONS` it returns the merged flags of `options` and whatever was already there; making the return value... unusual :).

Checking that it's equal to options may be reasonable, but I don't know if it's accurate.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                if ((IntPtr.Zero == memMethodWrite) || (IntPtr.Zero == memMethodRead))
+                {
+                    throw CreateSslException("Failed to return memory BIO method function");
```
bartonjs: While I like the paranoia that went into these checks, it's not anything that any of the crypto code bothers checking.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                }
+
+                sslContext.readBioPtr = libssl.BIO_new(memMethodRead);
+                sslContext.writeBioPtr = libssl.BIO_new(memMethodWrite);
+                if ((IntPtr.Zero == sslContext.readBioPtr) || (IntPtr.Zero == sslContext.writeBioPtr))
+                {
+                    throw CreateSslException("Failed to retun new BIO for a given method type");
```
stephentoub: Seems like readBioPtr could be non-null but writeBioPtr could be null, in which case we'd need to free readBioPtr, right?  Or is that handled in the catch block inside FreeSslContext (didn't seem like it, but maybe I missed it)?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                }      
+
+                IntPtr memMethodRead = libcrypto.BIO_s_mem();
+                IntPtr memMethodWrite = libcrypto.BIO_s_mem();
+                if ((IntPtr.Zero == memMethodWrite) || (IntPtr.Zero == memMethodRead))
+                {
+                    throw CreateSslException("Failed to return memory BIO method function");
+                }
+
+                sslContext.readBioPtr = libssl.BIO_new(memMethodRead);
```
bartonjs: This should use the existing p/invoke for `BIO_new`, which returns a `SafeBioHandle`, not `IntPtr`.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                else
+                {
+                    libssl.SSL_set_connect_state(sslContext.sslPtr);
+                }
+
+                libssl.SSL_set_bio(sslContext.sslPtr, sslContext.readBioPtr, sslContext.writeBioPtr);
+            }
+            catch
+            {
+                Marshal.StructureToPtr(sslContext, sslContextPtr, false);
```
stephentoub: I'm confused... if there's an error, why are we still copying state to the native ptr?

Also, we're throwing exceptions in the try block, but then catching them here and returning IntPtr.Zero?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            sendCount = 0;
+            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
+            bool isServer = context.isServer;
+            if ((IntPtr.Zero != recvPtr) && (recvCount > 0))
+            {
+                BioWrite(context.readBioPtr, recvPtr, recvCount);
+            }
+
+            int retVal = libssl.SSL_do_handshake(context.sslPtr);
+            if ((retVal == 1) && !isServer)
```
bartonjs: Why `!isServer`?  Would a client auth cert connection need to do more work here?  Or is it really just a client side v server side distinction? Comments would help.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            }
+            Marshal.StructureToPtr(sslContext, sslContextPtr, false);
+            return sslContextPtr;
+        }
+
+        internal static bool DoSslHandshake(IntPtr sslContextPtr, IntPtr recvPtr, int recvCount, out IntPtr sendPtr, out int sendCount)
+        {
+            sendPtr = IntPtr.Zero;
+            sendCount = 0;
+            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
```
stephentoub: So is the idea that we're carrying around a pointer to a native structure, and every time we need the state from it, we marshal that state back to a managed equivalent?

If so, I'd prefer we just have a managed object that carries the state around, and if we need to keep a pointer to it in native to be able to carry around context, we use a GCHandle.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                    error = sendCount;
+                    Marshal.FreeHGlobal(sendPtr);
+                    sendPtr = IntPtr.Zero;
+                    sendCount = 0;
+                    error = GetSslError(context.sslPtr, error, "Read Bio failed: ");                 
+                    throw CreateSslException("SSL Handshake failed", error);                   
+                }
+            }
+        
+            return ((libssl.SSL_state(context.sslPtr) == libssl.SslState.SSL_ST_OK)) ? true : false;
```
bartonjs: This line can be reduced to just `return libssl.SSL_state(context.sslPtr) == libssl.SslState.SSL_ST_OK;`, no?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+
+        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int count, int bufferCapacity)
+        {
+            SslContext context = Marshal.PtrToStructure<SslContext>(handlePtr);  
+
+            var retVal = libssl.SSL_write(context.sslPtr, buffer, count);
+
+            if (retVal != count)
+            {
+                throw CreateSslException("OpenSsl::Encrypt failed");
```
bartonjs: Many of these messages feel like they're for debugging.  Are they supposed to be `Debug.Fail` calls? Or is the intent that these messages would make it back to the developer/user consuming the library?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            var cipher = new libssl.SSL_CIPHER();
+            if (IntPtr.Zero != cipherPtr)
+            {
+                cipher = Marshal.PtrToStructure<libssl.SSL_CIPHER>(cipherPtr);
+            }
+
+            return cipher;
+        }
+
+        internal static byte[] GetCertificateRawData(IntPtr certPtr)
```
bartonjs: This already exists in the current crypto interop.  It's also a bit unclear why you'd need it... an X509Certificate2 can be made directly from `certPtr`.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                if (len <= 0)
+                {
+                    throw CreateSslException("Failed to get certificate raw data");
+                }
+            }
+
+            return buffer;
+        }
+
+        internal static void FreeCertificate(IntPtr certPtr)
```
bartonjs: This isn't called, and can be deleted.  (Which is good, because I was going to have lots of questions otherwise)

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            Disconnect(context.sslPtr);
+            Marshal.FreeHGlobal(sslContextPtr);
+            sslContextPtr = IntPtr.Zero;
+        }
+
+        #endregion
+
+        #region private methods
+
+        private static IntPtr GetSslMethod(bool isServer, long options)
```
bartonjs: Is `options` a `long`, or a `libssl.Options`?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            var noSsl2 = (options & libssl.Options.SSL_OP_NO_SSLv2) != 0;
+            var noSsl3 = (options & libssl.Options.SSL_OP_NO_SSLv3) != 0;
+            var noTls10 = (options & libssl.Options.SSL_OP_NO_TLSv1) != 0;
+            var noTls11 = (options & libssl.Options.SSL_OP_NO_TLSv1_1) != 0;
+            var noTls12 = (options & libssl.Options.SSL_OP_NO_TLSv1_2) != 0;
+
+            var method = IntPtr.Zero;
+            if (noSsl2 && noSsl3 && noTls11 && noTls12)
+            {
+                method = isServer ? libssl.TLSv1_server_method() : libssl.TLSv1_client_method();
```
bartonjs: There's also `TLSv1_method`, which leaves the option open for client or server; the call to `SSL_set_connect_state` vs `SSL_set_accept_state` later.  That would save on a bunch of conditional expressions in this method.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Just to clarify, these are all inner exception strings

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: I think this was an oversight.. one BIO_s_mem should be enough

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        {
+            int bytes = libssl.BIO_write(BioPtr, buffer, count);
+            if (bytes != count)
+            {
+                throw CreateSslException("Failed in Write BIO");
+            }
+            return bytes;
+        }
+
+        private static int GetSslError(IntPtr sslPtr, int result, string message)
```
stephentoub: message isn't being used?

stephentoub: And this function is only ever used just before calling CreateSslException... should it just be a part of CreateSslException?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        private sealed class SslException : Exception
+        {
+            public SslException(string message)
+                : base(message + ": " + Marshal.PtrToStringAnsi(libssl.ERR_reason_error_string(libssl.ERR_get_error())))
+            {
+                HResult = (int)libssl.ERR_get_error();
+            }
+
+            public SslException(string message, int error)
+                : base(message + ": " + error)
```
stephentoub: Shouldn't the message end up including te ERR_reason_error_string version of this error, just as in the previous ctor?  As previously mentioned, such string concatenation with messages really needs to be localizable as well... if we need to pass in both a message from the caller and from libssl, then the combining of those two messages should be done via an SR.Format.

src/System.Net.Security/src/System.Net.Security.csproj
```diff
@@ -23,10 +23,13 @@
     <Compile Include="System\Net\_SecurityBuffer.cs" />
     <Compile Include="System\Net\_SecurityBufferType.cs" />
     <Compile Include="System\Net\_SecureChannel.cs" />
+	<Compile Include="System\Net\SSPIInterface.cs" />
     <Compile Include="System\Net\_SSPIWrapper.cs" />
-    <Compile Include="System\Net\_NativeSSPI.cs" />
+	<Compile Include="System\Net\GlobalSSPI.cs" />
+	<Compile Include="System\Net\SecurityStatus.cs" />
+	<Compile Include="System\Net\CertInterface.cs" />
+	<Compile Include="System\Net\CertWrapper.cs" />
```
stephentoub: Nit: spacing is off on all these additional lines

src/System.Net.Security/src/System/Net/SecureProtocols/_SecuritySafeHandles.Windows.cs
```diff
@@ -219,6 +219,7 @@ public new IntPtr DangerousGetHandle()
         }
 #endif
 
+        #if false //TODO if not used in Nego Stream as well, please remove it.
```
stephentoub: I don't understand the TODO here.  What's the next step?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
```diff
@@ -117,25 +117,17 @@ internal void ValidateCreateContext(bool isServer, string targetHost, SslProtoco
...
-                enabledSslProtocols &= (SslProtocols)Interop.SChannel.ServerProtocolMask;
-                if (serverCertificate == null)
-                {
-                    throw new ArgumentNullException("serverCertificate");
-                }
-            }
-            else
-            {
-                enabledSslProtocols &= (SslProtocols)Interop.SChannel.ClientProtocolMask;
-            }
```
stephentoub: Why is it ok that these updates to enabledSslProtocols were removed?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+
+            if (IntPtr.Zero == method)
+            {
+                throw CreateSslException("Failed to get SSL method");
+            }
+
+            return method;
+        }
+
+        //TODO See if SSL_CTX_Set_quite_shutdown can be used
```
bartonjs: If this TODO is really a TODO, make an issue, and reference it here.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                int errorCode = SSPIWrapper.QueryContextRemoteCertificate(GlobalSSPI.SSPISecureChannel, securityContext, out remoteContext);
+                if (remoteContext != null && !remoteContext.IsInvalid)
+                {
+                    result = new X509Certificate2(remoteContext.DangerousGetHandle());
+                }
+            }
+            finally
+            {
+				//TODO: Fetch remoteCertificateStore, if applicable for unix
```
stephentoub: How are we tracking these TODOs?  Are issues being opened for each?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
bartonjs: Typo: SSL_CTX_set_quiet_shutdown.  (Bad casing, and quite vs quiet)

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        {
+            if (IntPtr.Zero != sslPtr)
+            {
+                int retVal = libssl.SSL_shutdown(sslPtr);
+                if (0 == retVal)
+                {
+                    retVal = libssl.SSL_shutdown(sslPtr);
+                    if (retVal < 0)
+                    {
+                        libssl.SSL_get_error(sslPtr, retVal);
```
bartonjs: Why get the error if you're not going to use it?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
stephentoub: If we expose them, even as the messages of inner exceptions, they still need to be localized.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            GlobalLog.Enter("SecureChannel#" + Logging.HashString(this) + "::RemoteCertificate{get;}");
+            X509Certificate2 result = null;
+            SafeFreeCertContext remoteContext = null;
+            try
+            {
+                int errorCode = SSPIWrapper.QueryContextRemoteCertificate(GlobalSSPI.SSPISecureChannel, securityContext, out remoteContext);
+                if (remoteContext != null && !remoteContext.IsInvalid)
+                {
+                    result = new X509Certificate2(remoteContext.DangerousGetHandle());
```
stephentoub: There doesn't appear to be anything keeping remoteContext alive after the call to get its DangerousGetHandle, which means it's theoretically possible it could get collected and its handle released boefre X509Certificate2's ctor gets a chance to use and dup it.  This should really be surroundd by DangerousAddRef/DangerousRelease calls, but if not that GC.KeepAlive should at least be used.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+namespace System.Net
+{
+    // For SSL connections:
+    internal class SSPISecureChannelType : SSPIInterface
+    {
+        public Exception GetException(SecurityStatus status)
+        {
+            // TODO: To be implemented
+            throw new Exception("status = " + status);
```
stephentoub: I realize from the comment that it's temporary, but this should be ```return new Exception``` rather than ```throw new Exception```, right?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        public SafeFreeCredentials AcquireCredentialsHandle(X509Certificate certificate,
+            SslProtocols protocols, EncryptionPolicy policy, bool isServer)
+        {
+            SafeFreeCredentials retVal = new SafeFreeCredentials(certificate, protocols, policy);
+            if ((null != retVal) && !retVal.IsInvalid)
+            {
+                // Caller does a ref count decrement
+                bool ignore = false;
+                retVal.DangerousAddRef(ref ignore);
```
stephentoub: Why are we AddRef'ing it?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+            {
+                count = resultSize;
+            }
+            return retVal;
+        }
+
+        public int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute,
+            out SafeFreeContextBufferChannelBinding refHandle)
+        {
+            throw NotImplemented.ByDesignWithMessage(SR.net_MethodNotImplementedException);
```
stephentoub: Other places have comments with "TODO"... presumably this is a TODO as well?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            return new SslException(message);
+        }
+
+        private static SslException CreateSslException(string message, int error)
+        {
+            if (error == libssl.SslErrorCode.SSL_ERROR_SYSCALL)
+            {
+                return new SslException(message);
+            }
+            else
```
bartonjs:
```C#
else if (error == libssl.SslErrorCode.SSL_ERROR_SSL)
{
    throw Interop.libcrypto.CreateOpenSslCryptographicException();
}
```

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        {
+            return CreateSslException(message, libssl.SSL_get_error(sslPtr, error));
+        }
+
+        private sealed class SslException : Exception
+        {
+            public SslException(string message)
+                : base(message + ": " + Marshal.PtrToStringAnsi(libssl.ERR_reason_error_string(libssl.ERR_get_error())))
+            {
+                HResult = (int)libssl.ERR_get_error();
```
bartonjs: Calling ERR_get_error changes the state of the error queue, but you've called it twice in this ctor.

It also would only apply if the SslErrorCode value was SSL_ERROR_SSL.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        public int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
+        {
+            bool gotReference = false;
+            connectionInfo = null;
+            try
+            {
+                securityContext.DangerousAddRef(ref gotReference);
+                Interop.libssl.SSL_CIPHER cipher = Interop.OpenSsl.GetConnectionInfo(securityContext.DangerousGetHandle());
```
stephentoub: Seems like GetConnectionInfo should take the SafeHandle rather than passing the dangerous handle into it. Same for similar methods below.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                if (gotReference)
+                {
+                    securityContext.DangerousRelease();
+                }
+            }
+        }
+
+        public int QueryContextIssuerList(SafeDeleteContext securityContext, out Object issuerList)
+        {
+            throw NotImplemented.ByDesignWithMessage(SR.net_MethodNotImplementedException);
```
stephentoub: Same "TODO" question.

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -0,0 +1,135 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class libssl
+    {
+
+        [DllImport(Interop.Libraries.LibSsl, EntryPoint = "ERR_load_BIO_strings")]
```
bartonjs: Why specify all of the EntryPoint values?

bartonjs: I'm surprised this doesn't get an EntryPointNotFoundException, since it's a function in libcrypto, not libssl.

It's also not necessary to call; since CryptoInitializer would have already loaded it.  Oh, and it isn't called... that explains a lot (other than why it's here).

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                else
+                {
+                    outputBuffer.token = null;
+                }      
+
+                return done ? SecurityStatus.OK : SecurityStatus.ContinueNeeded;
+            }
+            catch
+            {
+                return SecurityStatus.InternalError;
```
stephentoub: Should we Debug.Fail or something as well?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                bool done;
+
+                if (null == inputBuffer)
+                {
+                    done = Interop.OpenSsl.DoSslHandshake(context.DangerousGetHandle(), inputPtr, 0, out outputPtr, out outputSize);
+                }
+                else
+                {
+                    inputHandle = GCHandle.Alloc(inputBuffer.token, GCHandleType.Pinned);
+                    inputPtr = Marshal.UnsafeAddrOfPinnedArrayElement(inputBuffer.token, inputBuffer.offset);
```
stephentoub: Could we use ```fixed``` instead of a GCHandle?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                    done = Interop.OpenSsl.DoSslHandshake(context.DangerousGetHandle(), inputPtr, inputBuffer.size, out outputPtr, out outputSize);
+                }           
+
+                outputBuffer.size = outputSize;
+                outputBuffer.offset = 0;
+                if (outputSize > 0)
+                {
+                    outputBuffer.token = new byte[outputBuffer.size];
+                    Marshal.Copy(outputPtr, outputBuffer.token, 0, outputBuffer.size);
+                    Marshal.FreeHGlobal(outputPtr);
```
stephentoub: We don't want to free it also if outputSize <= 0?

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        internal const int SSL_CTRL_OPTIONS = 32;
+        internal static bool OPENSSL_NO_COMP = true; //no compression true by default
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct SSL_CIPHER
+        {
+            internal int valid;
+            internal string name; /* text name */
+            internal long id; /* id, 4 bytes, first is version */
```
bartonjs: All of these `long`s should really be `ulongs`; except that they really should be `UIntPtr` to get the marshal size correct.  But since we're going to move these to shims anyways it probably doesn't matter.

(FYI, @nguerrera and @sokket: a native `unsigned long` struct is being added)

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                }
+                else
+                {
+                    resultSize = Interop.OpenSsl.Decrypt(securityContext.DangerousGetHandle(), inputPtr, size);
+                }
+                return SecurityStatus.OK;
+            }
+            catch
+            {
+                return SecurityStatus.InternalError;
```
stephentoub: Debug.Fail as well?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        private SecurityStatus EncryptDecryptHelper(SafeDeleteContext securityContext, byte[] buffer, int offset, int size, int headerSize, int trailerSize, bool encrypt, out int resultSize)
+        {
+            bool gotReference = false;
+            GCHandle inputHandle = new GCHandle();
+            resultSize = 0;
+            try
+            {
+                securityContext.DangerousAddRef(ref gotReference);
+                inputHandle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
```
stephentoub: fixed instead of GCHandle?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            get
+            {
+                return IntPtr.Zero == handle;
+            }
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            if (IntPtr.Zero != handle)
```
stephentoub: This shouldn't be necessary.  ReleaseHandle shouldn't be called if IsInvalid.

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -0,0 +1,72 @@
...
+            internal long algorithm_mkey; /* key exchange algorithm */
```
bartonjs: I'm pretty sure that we didn't write these comments, so we should delete them.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+    //
+    // Implementation of handles dependable on FreeCredentialsHandle
+    //
+#if DEBUG
+    internal sealed class SafeFreeCredentials : DebugSafeHandle
+    {
+#else
+    internal sealed class SafeFreeCredentials : SafeHandle
+    {
+#endif
```
stephentoub: I don't understand what's being done with SafeHandle here... SetHandle is never being called, so we're never actually storing a handle into this.  Seems like it's really just being used as a finalizable object?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+    //
+    // This is a class holding a Credential handle reference, used for static handles cache
+    //
+#if DEBUG
+    internal sealed class SafeCredentialReference : DebugCriticalHandleMinusOneIsInvalid
+    {
+#else
+    internal sealed class SafeCredentialReference : CriticalHandleMinusOneIsInvalid
+    {
+#endif
```
stephentoub: Same question here about why we're using a CriticalHandle when we're not actually storing a handle.  I'll stop asking about this if it comes up again.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+        protected override bool ReleaseHandle()
+        {
+            if (IntPtr.Zero != handle)
+            {
+                Interop.OpenSsl.FreeSslContext(handle);
+            }
+            return true;
+        }
+
+        public override string ToString()
```
stephentoub: Is this just for debugging purposes?

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -0,0 +1,72 @@
...
+            internal long algorithm_enc; /* symmetric encryption */
+            internal long algorithm_mac; /* symmetric authentication */
+            internal long algorithm_ssl; /* (major) protocol version */
+            internal long algo_strength; /* strength and export flags */
+            internal long algorithm2; /* Extra flags */
+            internal int strength_bits; /* Number of bits really used */
+            internal int alg_bits; /* Number of bits for algorithm */
+        }
+
+        internal class Options
```
bartonjs: Why isn't this just a [Flags] enum?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            {
+                method = isServer ? libssl.TLSv1_1_server_method() : libssl.TLSv1_1_client_method();
+            }
+            else if (noSsl2 && noSsl3 && noTls10 && noTls11)
+            {
+                method = isServer ? libssl.TLSv1_2_server_method() : libssl.TLSv1_2_client_method();
+            }
+            else if (noSsl2 && noTls10 && noTls11 && noTls12)
+            {
+                method = isServer ? libssl.SSLv3_server_method() : libssl.SSLv3_client_method();
```
bartonjs: What's the minimum SSL version we're supporting these days?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+            }
+            else
+            {
+                method = isServer ? libssl.SSLv23_server_method() : libssl.SSLv23_client_method();
```
bartonjs: In particular, 2 seems like a no-no.

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -0,0 +1,72 @@
...
+            internal const int TRAILER_SIZE = 16;
+        }
+
+        internal static class SslErrorCode
+        {
+            internal const int SSL_ERROR_WANT_READ = 2;
+            internal const int SSL_ERROR_SYSCALL = 5;
+        }
+
+        internal static class SslState
```
bartonjs: Enum?

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -0,0 +1,72 @@
...
+                ? SSL3_RT_MAX_PLAIN_LENGTH
+                : SSL3_RT_MAX_PLAIN_LENGTH + SSL3_RT_MAX_COMPRESSED_OVERHEAD;
+
+            internal static int SSL3_RT_MAX_ENCRYPTED_LENGTH = (SSL3_RT_MAX_ENCRYPTED_OVERHEAD +
+                                                                SSL3_RT_MAX_COMPRESSED_LENGTH);
+            internal const int HEADER_SIZE = 5;
+            internal const int TRAILER_SIZE = 16;
+        }
+
+        internal static class SslErrorCode
```
bartonjs: Enum?

src/System.Net.Security/src/System/Net/CertWrapper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using Microsoft.Win32.SafeHandles;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Net
+{
+    internal static class CertWrapper
+    {
+        static CertInterface certModule = new CertModule();
```
bartonjs: Style: Access modifiers are required.

src/System.Net.Security/src/System/Net/SecurityStatus.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    internal enum SecurityStatus
+    {
+        // Success / Informational
+        OK,
```
bartonjs: I always find it worrying when a status enumeration has "success" as the zero value.

"NotSet", or some other otherwise unused value, would be a good thing to put here to ensure that "OK" is always intentional.

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: Can I assume that any code moved into the /Windows/ folder is exactly what the original code was, has only been moved and not changed at all, etc.?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
stephentoub: Oh, after reading further, I think I understand... these types are part of the PAL layer?

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Net
+{   
+    internal partial class CertModule : CertInterface
+    {
+        #region internal Methods
+        internal override SslPolicyErrors VerifyRemoteCertName(X509Chain chain, bool isServer, string hostName)
+        {
+            SslPolicyErrors sslPolicyErrors = SslPolicyErrors.None;
```
bartonjs: // TODO (issue): write this.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            return result;
+        }      
+
+        //
+        // Used only by client SSL code, never returns null.
+        //
+        internal override string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext)
+        {
+			//TODO populate issuers
```
bartonjs: (Note to self, write this, too)

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+            try
+            {
+                securityContext.DangerousAddRef(ref gotReference);
+                Interop.libssl.SSL_CIPHER cipher = Interop.OpenSsl.GetConnectionInfo(securityContext.DangerousGetHandle());
+                connectionInfo =  new SslConnectionInfo(cipher);
+                return 0;
+            }
+            catch
+            {
+                return -1;
```
bartonjs: Why are we eating all exceptions?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                _certHandle = Interop.libcrypto.X509_dup(certificate.Handle);
+            }          
+            _protocols = protocols;
+            _policy = policy;
+        }
+
+        public override bool IsInvalid
+        {
+            get { return SslProtocols.None == _protocols; }
```
bartonjs: If _protocols is none then we won't dispose _certHandle.  Since it's a SafeHandle the finalizer will eventually figure things out; but it would be better if we got this in the Dispose path.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+#else
+    internal sealed class SafeCredentialReference : CriticalHandleMinusOneIsInvalid
+    {
+#endif
+
+        //
+        // Static cache will return the target handle if found the reference in the table.
+        //
+        internal SafeFreeCredentials _Target;
```
bartonjs: Naming: _target

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -0,0 +1,101 @@
...
+        public readonly int DataCipherAlg;
+        public readonly int DataKeySize = 0;
+        public readonly int DataHashAlg;
+        public readonly int DataHashKeySize = 0;
+        public readonly int KeyExchangeAlg;
+        public readonly int KeyExchKeySize = 0;
+
+        internal SslConnectionInfo(Interop.libssl.SSL_CIPHER cipher)
+        {
+            Protocol = (int) MapProtocol(cipher.algorithm_ssl);
```
bartonjs: Why is Protocol an `int` if it's only assigned an `SslProtocols` value?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+        {
+            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
+
+            var retVal = BioWrite(context.readBioPtr, outBufferPtr, count);
+
+            if (retVal == count)
+            {
+                retVal = libssl.SSL_read(context.sslPtr, outBufferPtr, retVal);
+
+                if (retVal > 0) count = retVal;
```
bartonjs: Style: Unbraced ifs require the body to be on a newline.
Style: Unbraced ifs are not allowed within a braced if.

bartonjs:
While there's a lot here that should be resolved before going to master; I don't think there's anything here that would stop us from merging into the feature branch to let other people get started on changes herein and/or downstream.

@CIPop I'll wait for one of you guys to give at least the same level of signoff for the refactoring portions.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: options here is a long. libssl.Options is a class defined to hold option constants only.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: Removed unused message parameter. CreateSslException is called based on output from GetSslError at one place. Since CreateSslException is a generic function, do not want to add libssl.SslErrorCode.SSL_ERROR_WANT_READ check in CreateSslException method. So keeping GetSslError as a seperate method.

src/System.Net.Security/src/System/Net/SecureProtocols/_SecuritySafeHandles.Windows.cs
rajansingh10: Removed. Not needed in current implementation.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
rajansingh10: The idea was _sslState, secureChannel etc uber level classes will be independent of interop level functionality. Here code was using Interop.SChannel, which is windows SChannel specific. Instead, now we are directly validating enabledSslProtocols directly & storing the same in secureChannel in place of a masked int. Other logic have been modified as well to support this change.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
rajansingh10: Created issue #3362 , will add all the TODOs to the issue.

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
vijaykota: Yes.. unused ones should be removed. As an aside, calling from libssl works (holds true even for some *x509* and *BIO* methods)

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
rajansingh10: yes, corrected.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
rajansingh10: Yeah, .. inline with windows SafeHandles.

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
rajansingh10: The values are used for matching against output of SSL_get_error which is an Int. If converted to Int, this will have to be casted again to int for matching.

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
rajansingh10: Windows code have not touched upon mostly except where fitting was required to make SSPIWrapper & above layer agnostic of underlying os types.

src/Common/src/Interop/Unix/Interop.Libraries.cs
rajansingh10: Done

src/Common/src/Interop/Unix/Interop.Libraries.cs
rajansingh10: removed.Unused.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: fixed var related comment.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: Restructured the code, it will throw now & marshaling is done only when everything goes fine.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: removed.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/System.Net.Security/src/System.Net.Security.csproj
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
rajansingh10: added TODO

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
rajansingh10: removed.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/CertWrapper.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/SecurityStatus.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
rajansingh10: added TODO with issue

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
rajansingh10: done

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
rajansingh10: This is inline with windows version. Windows one gets an int from lower level Apis.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: 1. Modifying error path to free properly & avoid any crashing.
2. Allocating GCHandle only at the end now.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: We were not freeing in this case earlier. Added it now.

src/Common/src/Interop/Unix/System.Net.Security.Native/Interop.Initialization.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
```
CIPop: Nit: filename and folder name doesn't seem to follow corefx interop guidance.

CIPop: This code is probably better placed in the PAL layer?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,424 @@
...
+                    throw CreateSslException("Failed to allocate SSL/TLS context");
+                }
+
+                libssl.SSL_CTX_ctrl(contextPtr, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
+
+                sslContext.sslPtr = libssl.SSL_new(contextPtr);
+
+                libssl.SSL_CTX_free(contextPtr);     
+
+                if (IntPtr.Zero == sslContext.sslPtr)
```
CIPop: Nit: consistency - no Yoda expressions :) (Please reverse the order of operands.)

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
CIPop: Note that the final code needs to have all strings extracted in SR.

src/System.Net.Security/src/System/Net/CertInterface.cs
```diff
@@ -0,0 +1,19 @@
...
+    internal abstract class CertInterface
+    {
+        internal abstract  X509Certificate2 GetRemoteCertificate(SafeDeleteContext securityContext, out X509Certificate2Collection remoteCertificateStore);
+
+        internal abstract SslPolicyErrors VerifyRemoteCertName(X509Chain chain, bool isServer, string hostName);
+
+        internal abstract string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext);
+
+    }
+}
```
CIPop: Nit: missing new line at EOF.

src/System.Net.Security/src/System/Net/CertInterface.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace System.Net
+{
+    internal abstract class CertInterface
+    {
+        internal abstract  X509Certificate2 GetRemoteCertificate(SafeDeleteContext securityContext, out X509Certificate2Collection remoteCertificateStore);
+
+        internal abstract SslPolicyErrors VerifyRemoteCertName(X509Chain chain, bool isServer, string hostName);
+
+        internal abstract string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext);
+
```
CIPop: Nit: please remove extra new line.

src/System.Net.Security/src/System/Net/CertWrapper.cs
```diff
@@ -0,0 +1,30 @@
...
+        internal static X509Certificate2 GetRemoteCertificate(SafeDeleteContext securityContext, out X509Certificate2Collection remoteCertificateStore)
+        {
+            return certModule.GetRemoteCertificate(securityContext, out remoteCertificateStore);
+        }
+
+        internal static string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext)
+        {
+            return certModule.GetRequestCertificateAuthorities(securityContext);
+        }
+
```
CIPop: Nit: please remove extra new line.

src/System.Net.Security/src/System/Net/GlobalSSPI.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    // Need a global so we can pass the interfaces as variables.
+    internal static class GlobalSSPI
+    {
+        internal static SSPIInterface SSPISecureChannel = new SSPISecureChannelType();      
+    }
+
```
CIPop: Nit: extra new line.
Missing new line at EOF.

src/System.Net.Security/src/System/Net/GlobalSSPI.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    // Need a global so we can pass the interfaces as variables.
+    internal static class GlobalSSPI
+    {
+        internal static SSPIInterface SSPISecureChannel = new SSPISecureChannelType();
```
CIPop: `SSPIAuth` got removed. Please add a TODO to add it back.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+        SecurityStatus InitializeSecurityContext(SafeFreeCredentials credential, ref SafeDeleteContext context, string targetName, SecurityBuffer[] inputBuffers, SecurityBuffer outputBuffer);
+        SecurityStatus EncryptMessage(SafeDeleteContext securityContext, byte[] buffer, int size, int headerSize, int trailerSize, out int resultSize);
+        SecurityStatus DecryptMessage(SafeDeleteContext securityContext, byte[] buffer, ref int offset, ref int count);     
+        int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute, out SafeFreeContextBufferChannelBinding refHandle);
+        int QueryContextStreamSizes(SafeDeleteContext securityContext, out StreamSizes streamSizes);
+        int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo conectionInfo);
+        int QueryContextRemoteCertificate(SafeDeleteContext securityContext, out SafeFreeCertContext remoteCert);
+        int QueryContextIssuerList(SafeDeleteContext securityContext, out Object issuerList);
+    }   
+}
```
CIPop: Nit: no new line at EOF.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Security.Authentication;
+
+namespace System.Net
+{
+    // Used to define the interface for security to use.
+    internal interface SSPIInterface
+    {
+        void VerifyPackageInfo();
+        Exception GetException(SecurityStatus status);
+        SafeFreeCredentials AcquireCredentialsHandle(X509Certificate certificate, SslProtocols protocols, EncryptionPolicy policy, bool isServer);
```
CIPop: This is very tightly coupled with TLS which uses certs, etc. I fail to see how this will actually work in sharing the interface with Nego.
Sounds like it may be better to just invent something else, other than SSPIInterface which is Windows specific and use that for PAL purposes.

src/System.Net.Security/src/System/Net/SSPIInterfaceNego.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Net.Security;
+using System.Runtime.InteropServices;
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Authentication;
+
+namespace System.Net
+{
+    // Used to define the interface for security to use.
+    internal interface SSPIInterfaceNego
```
CIPop: The name is incorrect: this is actually the Schannel SSPI interface for both Nego and TLS. It is Windows specific and so shouldn't be used by common code.

src/System.Net.Security/src/System/Net/SecureProtocols/_SecuritySafeHandles.Windows.cs
CIPop: Please _do not_ remove code for NegotiateStream. It's on our TODO list and I need to understand where it landed.
Instead, move the code in a Windows-specific implementation.

src/System.Net.Security/src/System/Net/SecurityStatus.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    internal enum SecurityStatus
```
CIPop: Is this really needed? Since it's an internal type (it's actually the Windows SDK `SECURITY_STATUS`) I would expect it'll disappear and manifest itself as Exceptions thrown by the PAL layer.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
CIPop: Where did this move for the Windows code?

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    cbSize = (uint)Marshal.SizeOf<Interop.Crypt32.CERT_CHAIN_POLICY_PARA>(),
+                    dwFlags = 0,
+                    pvExtraPolicyPara = &eppStruct
+                };
+
+                fixed (char* namePtr = hostName)
+                {
+                    eppStruct.pwszServerName = namePtr;
+                    cppStruct.dwFlags |= (Interop.Crypt32.CertChainPolicyIgnoreFlags.CERT_CHAIN_POLICY_IGNORE_ALL &
```
CIPop: Nit: alignment was lost during copy/paste. Please try to keep line length under 120 chars.

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -0,0 +1,180 @@
...
+        //
+        // Used only by client SSL code, never returns null.
+        //
+        internal override string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext)
+        {
+            string[] issuers = Array.Empty<string>();
+
+            object outObj;         
+
+            int errorCode = SSPIWrapper.QueryContextIssuerList(GlobalSSPI.SSPISecureChannel, securityContext, out outObj);
```
CIPop: Shouldn't errorCode be strongly typed? If not, why do we need `SecurityStatus`?

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            GlobalLog.Assert(errorCode == 0, "QueryContextIssuerList returned errorCode:"+errorCode);
```
CIPop: Please follow the GlobalLog style within this function. (`SecureChannel::FunctionName():` ) - note that I have forgot to update the GlobalLog myself in the original PR.
Nit: spacing around `+errorCode`.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Thanks for confirming. @rajansingh10 is reevaluating these. As mentioned elsewhere, we can convert many of these to Debug.Fail. So wherever we need an exception, they will be added to Strings.resx

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: > In the future (not this change) we should seek to eliminate this pattern.

Completely agree. We thought of using SafeHandles throughout but then postponed this to post-Beta8. We will soon send a PR to use SafeHandles everywhere.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: We didn't use SafeBioHandle because the BIOs are "owned" by sslPtr after SSL_set_bio and will get freed in SSL_free(). I will take care of this in a later PR to use SafeHandles everywhere

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Agreed. I am planning to rewrite this using a SafeSslContext that derives from SafeHandle

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Yes.. these were meant for dev debugging. Debug.Fail is a much better idea and @rajansingh10 is revisiting these

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
vijaykota: Rewrote this in https://github.com/dotnet/corefx/pull/3396

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Good catch, @bartonjs. Thanks!

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
vijaykota: libssl does not require the pointer to be freed. We could still wrap it inside a SafeHandle but just want to check if the usage is appropriate

vijaykota: Agreed. We are planning to change all OpenSsl APIs to use SafeHandles

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: Yes, these Safe* classes are part of the PAL layer. They require the ref-counting semantics even and some cleanup though there is no actual handle

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
vijaykota: In Windows case, there is no exception and the caller code (platform-agnostic) expects a status code. We will be adding Debug.Fail statements though for debugging

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: Did you mean that constructor is called with SslProtocols.None? That won't happen but probably an Assert is in order.

I didn't understand the Dispose requirement. We don't want to dispose the cert (and private key) as long as references are alive. In fact the caller (GenerateToken) calls Dispose on this handle assuming refcount has been bumped up.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
vijaykota: That's why we initially started with a SecureChannelPal approach. Anyway, we will have to revisit this once we start the xplat investigation for NegotiateStream

src/System.Net.Security/src/System/Net/SSPIInterfaceNego.cs
vijaykota: @rajansingh10  can clarify. But I think this was for incorporating your design comment about unimplemented methods of SSPISecureChannelType (but which are implemented by SSPIAuthType)

PR #3346, opened 2015-09-21
Add base types for legacy crypto algorithms.

AtsushiKan:
cc @bartonjs

### DIFF
--- a/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/Helpers.cs
+++ b/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/Helpers.cs
@@ -14,6 +14,11 @@ namespace Internal.Cryptography
             return (byte[])(src.Clone());
         }
 
+        public static KeySizes[] CloneKeySizesArray(this KeySizes[] src)
+        {
+            return (KeySizes[])(src.Clone());
+        }
+
         public static bool UsesIv(this CipherMode cipherMode)
         {
             return cipherMode != CipherMode.ECB;
@@ -64,6 +69,27 @@ namespace Internal.Cryptography
                 arr[offset + 2] = (byte)(i >> 8);
                 arr[offset + 3] = (byte)i;
             }
+        }
+
+        public static byte[] FixupKeyParity(this byte[] key)
+        {
+            byte[] oddParityKey = new byte[key.Length];
+            for (int index = 0; index < key.Length; index++)
+            {
+                // Get the bits we are interested in
+                oddParityKey[index] = (byte)(key[index] & 0xfe);
+
+                // Get the parity of the sum of the previous bits
+                byte tmp1 = (byte)((oddParityKey[index] & 0xF) ^ (oddParityKey[index] >> 4));
+                byte tmp2 = (byte)((tmp1 & 0x3) ^ (tmp1 >> 2));
+                byte sumBitsMod2 = (byte)((tmp2 & 0x1) ^ (tmp2 >> 1));
+                
+                // We need to set the last bit in oddParityKey[index] to the negation
+                // of the last bit in sumBitsMod2
+                if (sumBitsMod2 == 0)
+                    oddParityKey[index] |= 1;
+            }
+            return oddParityKey;
         }
     }
 }
--- a/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
+++ b/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
@@ -1,0 +1,139 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+using Internal.Cryptography;
+
+namespace System.Security.Cryptography
+{
+    public abstract class DES : SymmetricAlgorithm
+    {
+        protected DES()
+        {
+            KeySize = 64;
+            BlockSize = 64;
+        }
+
+        public static DES Create()
+        {
+            throw new NotImplementedException(SR.WorkInProgress);
+        }
+
+        public override KeySizes[] LegalKeySizes
+        {
+            get
+            {
+                return s_legalKeySizes.CloneKeySizesArray();
+            }
+        }
+
+        public override KeySizes[] LegalBlockSizes
+        {
+            get
+            {
+                return s_legalBlockSizes.CloneKeySizesArray();
+            }
+        }
+
+        public override byte[] Key
+        {
+            get
+            {
+                byte[] key = base.Key;
+                while (IsWeakKey(key) || IsSemiWeakKey(key))
+                {
+                    GenerateKey();
+                    key = base.Key;
+                }
+                return key;
+            }
+
+            set
+            {
+                if (value == null)
+                    throw new ArgumentNullException("value");
+
+                if (!(value.Length*8).IsLegalSize(s_legalKeySizes))
+                    throw new ArgumentException(SR.Cryptography_InvalidKeySize);
+
+                if (IsWeakKey(value))
+                    throw new CryptographicException(SR.Format(SR.Cryptography_InvalidKey_Weak, "DES"));
+
+                if (IsSemiWeakKey(value))
+                    throw new CryptographicException(SR.Format(SR.Cryptography_InvalidKey_SemiWeak, "DES"));
+
+                base.Key = value;
+            }
+        }
+
+        public static bool IsSemiWeakKey(byte[] rgbKey)
+        {
+            if (rgbKey == null)
+                throw new CryptographicException(SR.Cryptography_InvalidKeySize);  // Desktop compat: Strange exception for a null value, but this is what we threw in classic CLR. 
+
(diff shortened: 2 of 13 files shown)


src/System.Security.Cryptography.Algorithms/tests/RC2Tests.cs
```diff
@@ -0,0 +1,72 @@
...
+        public static void RC2EffectiveKeySize()
+        {
+            using (RC2 rc2 = new RC2Minimal())
+            {
+                for (int keySize = 40; keySize <= 1024; keySize += 8)
+                {
+                    rc2.KeySize = keySize;
+                    Assert.Equal(keySize, rc2.KeySize);
+
+
```
bartonjs: Nit: too many blank lines

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DSA.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+using Internal.Cryptography;
+
+namespace System.Security.Cryptography
+{
+    public abstract class DSA : AsymmetricAlgorithm
```
bartonjs: We'd probably want to add the following public members:
```C#
byte[] SignData(Stream, HashAlgorithmName);
byte[] SignData(byte[] buffer, HashAlgorithmName);
byte[] SignData(byte[] buffer, int offset, int count, HashAlgorithmName);
bool VerifyData(byte[] data, byte[] signature, HashAlgorithmName);
bool VerifyData(Stream, byte[] signature, HashAlgorithmName);
```
and the protected drivers thereof:
```C#
protected abstract byte[] HashData(byte[] data, int offset, int count, HashAlgorithmName hashAlgorithm);
protected abstract byte[] HashData(Stream data, HashAlgorithmName hashAlgorithm);
```

But we may already be in a bind for adding those, since Desktop doesn't have them yet.

@nguerrera: Thoughts?

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/RC2.cs
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+using Internal.Cryptography;
+
+namespace System.Security.Cryptography
+{
+    public abstract class RC2 : SymmetricAlgorithm
```
bartonjs: RC2CryptoServiceProvider has `public bool UseSalt { get; set; }`.  Do we want/need that on the base class?  (I'm not familiar enough with RC2 to know.)

bartonjs:
We should get Nick's opinion on the DSA base class enhancements and contract version/management before merging; just to know what the plan forward is.

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/Helpers.cs
```diff
@@ -14,6 +14,11 @@ public static byte[] CloneByteArray(this byte[] src)
             return (byte[])(src.Clone());
         }
 
+        public static KeySizes[] CloneKeySizesArray(this KeySizes[] src)
+        {
+            return (KeySizes[])(src.Clone());
+        }
+
```
stephentoub: Should we just have a helper:
```C#
public static T[] CloneArray<T>(this T[] src)
{
    return (T[])src.Clone();
}
```
rather than one method per T?

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        protected DES()
+        {
+            KeySize = 64;
+            BlockSize = 64;
+        }
+
+        public static DES Create()
+        {
+            throw new NotImplementedException(SR.WorkInProgress);
```
stephentoub: These NIEs are going to be handled in a subsequent PR?

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/Helpers.cs
AtsushiKan: A valid point but also brings in more ancillary code changes (and inconsistencies with other crypto assemblies) that I don’t want to distract from this change.

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/TripleDES.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+                return key;
+            }
+
+            set
+            {
+                if (value == null)
+                    throw new ArgumentNullException("value");
+
+                if (!(value.Length*8).IsLegalSize(s_legalKeySizes))
```
stephentoub: I personally don't like us having domain-specific extension methods on integers like this, but I realize you're not introducing this as part of your change, so ok.

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
AtsushiKan: Yeah, we have to figure out the right type-forwarding strategy for these Create() methods. The priority here is to get these base classes into the Algorithms assembly and packages
so we can add the Cng subclasses w/out friction once we switch away from P2P references.

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return true;
+            }
+            return false;
+        }
+
+        public static bool IsWeakKey(byte[] rgbKey)
+        {
+            if (rgbKey == null)
+                throw new CryptographicException(SR.Cryptography_InvalidKeySize);  // Desktop compat: Strange exception for a null value, but this is what we threw in classic CLR.
```
stephentoub: These types are being added purely for legacy compat, right?  If so, makes sense to keep the behavior as consistent as possible.  If, however, we believe there's non-legacy value in these types, then we should consider breaking compat here in the name of throwing the right exception.

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/RC2.cs
```diff
@@ -0,0 +1,100 @@
...
+                if (value > KeySize)
+                    throw new CryptographicException(SR.Cryptography_RC2_EKSKS);
+
+                if (value == 0)
+                {
+                    _effectiveKeySize = value;
+                }
+                else if (value < 40)
+                {
+                    throw new CryptographicException(SR.Cryptography_RC2_EKS40);
```
stephentoub: This case won't be caught by the subsequent IsLegalSize check?

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/DES.cs
AtsushiKan: Legacy compat is indeed the only reason we want these ancient algorithms around ☺

src/System.Security.Cryptography.Algorithms/tests/DesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                Assert.Equal(64, des.BlockSize);
+                Assert.Equal(CipherMode.CBC, des.Mode);
+                Assert.Equal(PaddingMode.PKCS7, des.Padding);
+            }
+        }
+
+        [Fact]
+        public static void DesIsWeakPositive()
+        {
+            foreach (byte[] key in _knownWeakKeys)
```
stephentoub: Consider doing this with a [Theory] and [MemberData] instead.  Then if the asserts fail, the test output will include the failed inputs.

stephentoub:
LGTM

src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/RC2.cs
AtsushiKan: It does seem redundant but the original desktop code did set aside this specific error string for the  “< 40” case so I figured it was safer not to tamper.

AtsushiKan:
Removed the extranous whitespace.

Removed DSA.cs from PR - will handle this separately since we have some contract issues to haggle out (and depending on the answer, we'll probably want to refactor existing code to avoid code duplication.)

bartonjs:
It feels to me like we'd hold `DSAParameters` until adding DSA.  Thoughts?

AtsushiKan:
Well the question isn’t whether we’ll add DSA, just what methods will be on it. So I’d rather not create unnecessary new debt.

bartonjs:
That's fair.  It's not like we'd really debate anything on that type, and we can just leave it out of the contract until DSA makes it in (even if that's post RTM)

PR #3350, opened 2015-09-21
Ensure getting random bytes/numbers produces an unpredicatable sequence.

eerhardt:
To ensure we are getting truly random values, we now call RAND_poll()
during OpenSSL initialization.  Also, calling RAND_bytes instead of the
deprecated RAND_psuedo_bytes.

While I was fixing the code, I also added a shim for the native call
instead of invoking libcrypto directly from managed code.

Also, fixed a small bug in the tests to assert that the byte elements are different
in the arrays, and not the arrays themselves.

@bartonjs

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -13,6 +13,7 @@
 #include <openssl/asn1.h>
 #include <openssl/bio.h>
 #include <openssl/evp.h>
+#include <openssl/rand.h>
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
@@ -1007,6 +1008,27 @@ GetPkcs7Certificates(
     return 0;
 }
 
+/*
+Function:
+GetRandomBytes
+
+Puts num cryptographically strong pseudo-random bytes into buf.
+
+Return values:
+Returns a bool to managed code.
+1 for success
+0 for failure
+*/
+int
+GetRandomBytes(
+    unsigned char* buf, 
+    int num)
+{
+    int ret = RAND_bytes(buf, num);
+
+    return ret == 1;
+}
+
 // Lock used to make sure EnsureopenSslInitialized itself is thread safe
 static pthread_mutex_t g_initLock = PTHREAD_MUTEX_INITIALIZER;
 
@@ -1057,6 +1079,7 @@ EnsureOpenSslInitialized()
     int ret = 0;
     int numLocks = 0;
     int locksInitialized = 0;
+    int randPollResult = 0;
 
     pthread_mutex_lock(&g_initLock);
 
@@ -1097,6 +1120,15 @@ EnsureOpenSslInitialized()
 
     // Initialize the callback
     CRYPTO_set_locking_callback(LockingCallback);
+
+    // Initialize the random number generator seed
+    randPollResult = RAND_poll();
+    if (randPollResult < 1)
+    {
+        assert(0 && "RAND_poll() failed.");
+        ret = 4;
+        goto done;
+    }
 
 done:
     if (ret != 0)
--- a/src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
+++ b/src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
@@ -183,15 +183,12 @@ namespace System.IO
             return path.Length > 0 && path[0] == DirectorySeparatorChar;
         }
 
-        private static unsafe byte[] CreateCryptoRandomByteArray(int byteLength)
+        private static byte[] CreateCryptoRandomByteArray(int byteLength)
         {
             var arr = new byte[byteLength];
-            fixed (byte* buf = arr)
+            if (!Interop.Crypto.GetRandomBytes(arr, arr.Length))
             {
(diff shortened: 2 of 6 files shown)


src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -1098,6 +1121,15 @@ EnsureOpenSslInitialized()
     // Initialize the callback
     CRYPTO_set_locking_callback(LockingCallback);
 
+    // Initialize the random number generator seed
+    randPollResult = RAND_poll();
+    if (randPollResult < 1)
+    {
+        assert(!"RAND_poll() failed.");
```
bartonjs: This causes a type coercion warning-as-error on some machines (so far, only mine...); so this should be
`assert(0 && "RAND_poll() failed.");` 

The rest were converted on Friday in https://github.com/dotnet/corefx/commit/d01a2f4d2fe33e358a83d64a42fb830012037c5a#diff-eeff6f0c4ba65023e3af722532f18575.

src/System.Security.Cryptography.Algorithms/tests/RandomNumberGeneratorTests.cs
```diff
@@ -151,7 +151,7 @@ private static void DifferentSequential(int arraySize)
             // = 1/256 * 1/256 * ... * 1/256
             // = 1/(256^10)
             // = 1/1,208,925,819,614,629,174,706,176
-            Assert.NotEqual(first, second);
+            Assert.NotEqual<byte>(first, second);
```
bartonjs: Does this have a measurable change for you?  I just verified that Equal, at least, seems to do what I want:

```C#
byte[] data = ByteUtils.RepeatByte(0x71, 11);
byte[] data2 = (byte[])data.Clone();
data2[0] = 3;

Assert.Equal(data, data2);
```


```
 System.Security.Cryptography.EcDsaOpenSsl.Tests.EcDsaOpenSslTests.VerifyDuplicateKey_ValidHandle [FAIL]
      Assert.Equal() Failure
      Expected: Byte[] [113, 113, 113, 113, 113, ...]
      Actual:   Byte[] [3, 113, 113, 113, 113, ...]
      Stack Trace:
            at System.Security.Cryptography.EcDsaOpenSsl.Tests.EcDsaOpenSslTests.VerifyDuplicateKey_ValidHandle()
```

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -183,15 +183,12 @@ public static bool IsPathRooted(string path)
             return path.Length > 0 && path[0] == DirectorySeparatorChar;
         }
 
-        private static unsafe byte[] CreateCryptoRandomByteArray(int byteLength)
+        private static byte[] CreateCryptoRandomByteArray(int byteLength)
         {
             var arr = new byte[byteLength];
-            fixed (byte* buf = arr)
+            if (!Interop.Crypto.GetRandomBytes(arr, arr.Length))
```
stephentoub: Are you sure we want to do this?  I'd explicitly decided to use RAND_pseudo_bytes because RAND_bytes will fail if there's not enough entropy available, which is definitely possible, and I don't believe any of these consuming APIs are supposed to fail in such a case.

bartonjs: With the default provider, at least, `RAND_bytes()` only fails if the entropy pool never cleared the minimum threshold.  Once it has returned an answer you're good for the rest of the process (at least with 0.9.8).

`RAND_poll` is "keep doing things with randomness sources until the entropy pool exceeds the minimum threshold"; if it returns `0` that means OpenSSL couldn't figure out how randomness works for the system (compile-time behavior).

And since MSDN says that System.Security.Cryptography.RandomNumberGenerator is the cryptographically-secure RNG, we can't have it returning pseudo-bytes (like Random :smile:).

stephentoub: As long as we're sure that RAND_bytes won't fail due to lack of entropy (and won't block for arbitrary periods of time waiting for, say, mouse movement, like /dev/random often does), I'm happy with the change.

eerhardt: Another point of interest is that RAND_pseudo_bytes is deprecated according to http://openssl.org/docs/manmaster/crypto/RAND_bytes.html
> RAND_pseudo_bytes() has been deprecated. Users should use RAND_bytes() instead.

src/System.Security.Cryptography.Algorithms/tests/RandomNumberGeneratorTests.cs
eerhardt: You are right, this change doesn't have any measurable difference.  Assert.[Not]Equal checks for IEnumerable underneath the covers and does the enumeration.  This is different behavior than MSTest, which is why I was confused.
I'll revert these changes.

bartonjs: @eerhardt Phew!  I was afraid we'd have to update a LOT more places :)

eerhardt:
Updated for PR feedback.

bartonjs:
LGTM, thanks @eerhardt .

PR #3352, opened 2015-09-21
Fix opening of stdin/stdout/stderr to work with sockets

stephentoub:
In Console (and a couple of other low-level libs where we write directly to stderr), we currently open stdin/stdout/stderr via the paths "/dev/stdin", "/dev/stdout", and "/dev/stderr".  This generally works fine.  However, if the underlying descriptors have been redirected to a socket, attempting to open these is explicitly coded in Linux to fail.  And some applications that launch child processes use socketpair instead of pipe to communicate with the child process, in which case the child process' Console will fail to open /dev/*, resulting in an ENXIO error ("no such device or address").  One such application is node.js.

This fix changes how we open stdin/stdout/stderr in corefx.  Rather than opening e.g. "/dev/stdin", which effectively ends up dup'ing the corresponding file descriptor, we just dup the corresponding file descriptor directly, e.g. dup(STDIN_FILENO).

(While touching code in SafeFileHandle.Open, I also removed a try { } finally that I'd added months ago to help with thread aborts, before I comprehended we didn't need to worry about them.)

Fixes https://github.com/aspnet/dnx/issues/2763#issuecomment-141896625
cc: @nguerrera, @sokket, @MartinJohns


src/Native/System.Native/pal_io.h
```diff
@@ -293,6 +293,12 @@ extern "C" int32_t Open(const char* path, int32_t flags, int32_t mode);
 extern "C" int32_t Close(int32_t fd);
 
 /**
+* Duplicates a file descriptor.
+*
+* Returns the duplication descriptor for success, -1 for failure. Sets errno on failure.
+*/
+extern "C" int32_t Dup(int32_t oldfd);
+/**
```
nguerrera: Nit: align *'s as with other functions, leave one newline between functions.

nguerrera:
LGTM

src/Native/System.Native/pal_io.h
stephentoub: Fixed.

andschwa:
Thanks for fixing this :)

PR #3355, opened 2015-09-21
Expose DuplicateKeyHandle on OpenSSL-specific asymmetric algorithms.

bartonjs:
A lot of the lines in the change are from changing the casing on SafeEvpPKeyHandle while making it public, and/or updating usings to account for it having a new namespace.

X509Certificates no longer needs to include the file because it sees it publicly from the OpenSsl library.


bartonjs:
cc: @AtsushiKan @nguerrera

nguerrera:
LGTM

PR #3356, opened 2015-09-21
Restrict the SSL handshake test to 3 seconds.

bartonjs:
When things don't go so well this test takes significantly longer than an acceptable waiting period.  On success it's fractions of a second.

### DIFF
--- a/src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
+++ b/src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
@@ -29,7 +29,8 @@ namespace System.Net.Security.Tests
                 auth[0] = client.AuthenticateAsClientAsync(certificate.Subject);
                 auth[1] = server.AuthenticateAsServerAsync(certificate);
 
-                Task.WaitAll(auth);
+                bool finished = Task.WaitAll(auth, TimeSpan.FromSeconds(3));
+                Assert.True(finished, "Handshake completed in the allotted time");
             }
         }
 


bartonjs:
cc: @stephentoub @CIPop

bartonjs:
On the one hand, I meant dev-sslstream.  On the other; master won't hurt; we'll just need to merge it in to the dev branch.

bartonjs:
I sometimes have to think about if the assert message is in the right direction, but here's the Unix PR running with this test:

```
jbarton@jsb-ubuntu2:~/coreclr$ ./runtests.sh System.Net.Security.Tests.dll -method System.Net.Security.Tests.SslStreamStreamToStreamTest.SslStream_StreamToStream_Authentication_Success
xUnit.net console test runner (64-bit .NET Core)

Discovering: System.Net.Security.Tests
Discovered:  System.Net.Security.Tests
Starting:    System.Net.Security.Tests
   System.Net.Security.Tests.SslStreamStreamToStreamTest.SslStream_StreamToStream_Authentication_Success [FAIL]
      Handshake completed in the allotted time
      Expected: True
      Actual:   False
      Stack Trace:
            at System.Net.Security.Tests.SslStreamStreamToStreamTest.SslStream_StreamToStream_Authentication_Success()
Finished:    System.Net.Security.Tests

   System.Net.Security.Tests  Total: 1, Errors: 0, Failed: 1, Skipped: 0, Time: 3.475s
```

stephentoub:
> here's the Unix PR running with this test:

Just to clarify, we expect the Unix implementation to fail this test currently, right?

stephentoub:
LGTM (though if there's any chance that a successful run could take 3seconds due to load in the lab, etc., consider making the timeout longer)

bartonjs:
@stephentoub Yep, I was being ambitious and seeing if anything would magically pass.  Most things failed due to wanting System.Net.Sockets 4.1.0.0 (I have a preview copy which is labelled 4.0.x), and this one used a mock network (yay) but "hung" (boo).

CIPop:
LGTM for this branch and dotnet/master.

>Just to clarify, we expect the Unix implementation to fail this test currently, right?

@bartonjs You could create a copy of most (if not all) of the tests that use the mock network until Sockets is done. The SslStream tests are using Sockets just from inertia: they were ported from the old Desktop test suite.

Issue #2498, opened 2015-07-24
MemoryMappedFilesTests.DataShared failed in CI


stephentoub:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/1332/console
```
InterProcessCommunication.Tests.MemoryMappedFilesTests.DataShared [FAIL]
12:44:49       System.IO.IOException : The process cannot access the file '/tmp/MemoryMappedFilesTests_8729349d9699eb9cc0a6adf22b1b7afa/DataShared_16' because it is being used by another process.
12:44:50       Stack Trace:
12:44:50 Starting:    System.Collections.Immutable.Tests
12:44:50             at System.IO.UnixFileStream.SysCall[TArg1,TArg2](Func`4 sysCall, TArg1 arg1, TArg2 arg2, Boolean throwOnError)
12:44:50             at System.IO.UnixFileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
12:44:50             at System.IO.UnixFileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
12:44:50             at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
12:44:50             at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
12:44:50             at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync)
12:44:50             at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
12:44:50             at InterProcessCommunication.Tests.MemoryMappedFilesTests.DataShared()
12:44:50 Finished:    InterProcessCommunication.Tests
```

stephentoub:
Failed again:
InterProcessCommunication.Tests.MemoryMappedFilesTests.DataShared [FAIL]
08:51:47       System.IO.IOException : The process cannot access the file '/tmp/MemoryMappedFilesTests_339a79be42f1b33c1db9597642dcdd9c/DataShared_16' because it is being used by another process.
08:51:47       Stack Trace:
08:51:47 Starting:    System.Collections.Immutable.Tests
[...]

PR #3359, opened 2015-09-22
Shim HMAC functions

nguerrera:
cc @bartonjs @stephentoub @sokket


stephentoub:
LGTM.  Nice use of result_of and enable_if to clean things up.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Hmac.cs
```diff
@@ -0,0 +1,28 @@
...
+
+internal static partial class Interop
+{
+    internal static partial class Crypto
+    {
+        [DllImport(Libraries.CryptoNative)]
+        internal extern static unsafe SafeHmacCtxHandle HmacCreate(byte* key, int keyLen, IntPtr md);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal extern static unsafe void HmacDestroy(IntPtr ctx);
```
bartonjs: This one doesn't seem to need to be unsafe...

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HashProviderDispenser.Unix.cs
```diff
@@ -130,7 +130,11 @@ public unsafe HmacHashProvider(IntPtr algorithmEvp, byte[] key)
 
                 fixed (byte* keyPtr = key)
                 {
-                    Check(Interop.libcrypto.HMAC_Init(out _hmacCtx, keyPtr, key.Length, algorithmEvp));
+                    _hmacCtx = Interop.Crypto.HmacCreate(keyPtr, key.Length, algorithmEvp);
+                    if (_hmacCtx == null)
```
bartonjs: Wouldn't this never be null, but might be IsInvalid?

stephentoub: Yup.

src/Native/System.Security.Cryptography.Native/pal_hmac.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (ctx == nullptr)
+    {
+        assert(false && "Allocation failed.");
+        return nullptr;
+    }
+
+    // !REVIEW! Zeroing out HMAC_CTX prior to init mimics prior managed initialization
+    //          and our tests fail without it, but I can't find a reference stating it's
+    //          necessary.
+    *ctx = {};
```
bartonjs: You need to do both this -and- HMAC_CTX_init?  Because all HMAC_CTX_init does is zero out some fields; and HMAC_Init_ex resets the rest of them (I don't see it ever reading from the current structure to decide initialization already happened)

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HashProviderDispenser.Unix.cs
bartonjs: This if/throw could be replaced with a call to Interop.libcrypto.CheckValidOpenSslHandle.

bartonjs:
Aside from the one logic bug (checking null vs IsInvalid) LFTM.  (C++ templates exceed my native-fu, so I can't quite give it a "G" :))

src/Native/System.Security.Cryptography.Native/pal_hmac.cpp
nguerrera: Something questionable is happening here. 

HMAC_CTX_Init is the documented thing to do, so we should always due that and not assume anything about it's implementation being equivalent/subset of zeroing.

I need to debug through what exactly in context is left uninitialized without the zeroing before HMAC_CTX_init, and where it's getting read to impact tests. I won't check anything in until it's fully understood and I'll make sure the solution is well commented and not of the "I don't get this variety" here. I found this late and wanted to get the rest of the code reviewed while I work it out.

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HashProviderDispenser.Unix.cs
nguerrera: Great catch. Oops. Easy mistake to make when you just typed return nullptr on the other side. This would be a good Roslyn analyzer, safehandle returned from P/Invoke compared to null. ;)

src/Native/System.Security.Cryptography.Native/pal_hmac.cpp
nguerrera: OK, determined the issue is only in the case where key is null and keyLen is 0. OpenSSL treats null key arg specially and doesn't overwrite "existing" key. Added commit (that I'll ultimately squash) with more deliberate handling.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Hmac.cs
nguerrera: Fixed

src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/HashProviderDispenser.Unix.cs
nguerrera: Fixed.

nguerrera:
@bartonjs I think I've addressed everything Mind taking another look? (Don't merge if it looks good as I'll squash that fixup first.)

bartonjs:
LGTM; squash-merge at will!

nguerrera:
@dotnet-bot test this please (reopened #2498)

nguerrera:
This last failure looks like a race between the ICU merge hitting coreclr, but not corefx yet. Waiting for #3381 before attempting another test.

ellismg:
Yeah, sorry Nick, working as fast as I can.

nguerrera:
@ellismg No worries. Exciting to see x-plat glob work coming in. Nice job!

stephentoub:

PR #3361, opened 2015-09-22
Move max path directory check up front.

JeremyKuhne:
No need to create a bunch of intermediate directories when we'll fail at the end.
This was having significant impact on test times with #3192.

@nguerrera, @weshaggard, @stephentoub 

This takes times back down to something reasonable when testing the exception on CreateDirectory.

Note that this doesn't address the device name issue that Wes is seeing. Still looking into what we can do with that. I get the same sort of crazy increased times when I'm connected to corpnet.


stephentoub:
LGTM

PR #3363, opened 2015-09-22
Update to build tools version 94

weshaggard:
This refreshs the tool-runtime so it supports win7-x86 and
ubuntu-x64. We need the win7-x86 support to get tools working in
our internal builds and the ubuntu support is just prep work for
some future changes.


stephentoub:
LGTM

PR #3365, opened 2015-09-22
Dummy PR with xplat SslStream fixes

vijaykota:
Sharing some fixes found during unit testing. We will be sending these changes as part of official PRs. The idea is to unblock devs who may be hitting issues while testing.

Note that the first 2 commits are from https://github.com/dotnet/corefx/pull/3345


vijaykota:
@bartonjs  has already submitted this for review via https://github.com/dotnet/corefx/pull/3386. Hence closing this

PR #3367, opened 2015-09-22
Add tests for the null checks ToLookup should do.

JonHanna:


JonHanna:
Also delete LegacyTests/SingleTests.cs which was a zero-byte file left behind when its tests were moved.

stephentoub:
LGTM. Thanks.

Issue #3156, opened 2015-09-10
Can we determine if a FileSystem is case sensitive or not through an API?


troydai:
I'm exploring the possibility of making our cross-platform file globbing to determine if it acts as case sensitive or not based on current file system. It can't be determined just by the current operating system since case insensitive file system can still be mounted on Linux/Unix/Darwin. Is there an API can help to determine if a file system at given path is case sensitive.

stephentoub:
Unfortunately, no: https://github.com/dotnet/corefx/issues/1086

stephentoub:
(Hmm, upon rereading the issue I linked to, it's about getting the actual casing of a path.  I was sure we had an issue about an API for getting the case sensitivity of the current file system... not sure what happened to that one, but I guess we can use this issue in its place.) cc: @sokket, @ellismg

sokket:
@stephentoub I thought we had an issue too but I couldn't find it either. I know we discussed this when dealing with the FileSystemWatcher but pushed it out later due to complexities around mount points with different underlying File Systems mounted anywhere. This seems like it might be a fit for an API discussion to figure out if we want to provide an API to determine if the given path is case sensitive.

nguerrera:
cc @JeremyKuhne

yishaigalatzer:
This will be very useful for us as well. I'd like to see an API that take a folder, because the cases where have heard about are:

1. Windows machine with a samba share
2, Linux machine with a folder mounted to a window share
3. Mac + mount points to either case sensitive or case insensitive drive.

I would expect such an API to be reasonably fast with some sort of caching if needed so it can be called multiple times without a major penalty, and if not properly document it on the API so it is clear if the consumer is required to cache.

karelz:
Related to #1086.

JeremyKuhne:
#1086 should handle most people's needs.

In this case we need something similar to `DriveInfo` but available for any path. Perhaps `VolumeInfo`.  In Win32 you need to use Volume Management APIs such as [GetVolumeInformation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364993) to find the relevant file system info (which `DriveInfo` already utilizes for some of it's functionality, but doesn't give you the ability to use anything other than a root path- you would have to figure out where mounted folders actually point to).

If someone can pull together the necessary APIs on the Unix/MacOS side then we can move forward with this. For now I'll close in favor of #1086. Please feel free to reopen if you can drive this issue forward.

PR #3372, opened 2015-09-22
Add internal platform-agnostic case sensitivity detection

stephentoub:
After the recent refactoring based on introducing native shims, the only reason there are special builds of System.IO.FileSystem and System.IO.FileSystem.DriveInfo for each of Linux and OS X is because of PathInternal.CaseSensitive/CaseInsensitive.cs, which just provides a hard-coded value as to whether the target platform is case-sensitive or case-insensitive-but-preserving.  Similarly, these files are the only things forcing System.IO.Compression.ZipFile to not be completely platform-agnostic and to have separate Windows/Unix builds.

This commit recombines those assemblies by adding a platform-agnostic helper that simply creates a temporary file with upper-case letters and then checks whether that file can be found via the lower-cased version.  This isn't perfect, in that it's possible to mount multiple file systems each with different case sensitivities, but a) it's better than the hardcoded version we currently have, and b) I don't know of any better portable solution.

(Issue #3156 tracks adding an API to provide information on case sensitivity.  This change can be replaced by using such an API if/when it's available.)

cc: @ericstj, @nguerrera, @sokket, @chcosta


nguerrera:
LGTM. Love how this cleans things up. 

cc @JeremyKuhne who was thinking about the longer term solution for this for a PathsAreEquivalent proposal IIRC.

sokket:
This cleans things up nicely! :) LGTM 2

stephentoub:
Thanks, guys.

JeremyKuhne:
It looks like we'll need a PathStartsWith as well as PathsAreEquivalent.

ericstj:
LGTM

PR #3376, opened 2015-09-22
Fix System.Net.Security.csproj to account for the SafeEvpPKey change

bartonjs:


bartonjs:
Okay, so we can't build without this change.  But apparently we can't build with  this change, either.

For some reason the S.S.C.OpenSsl library (via S.Net.Security) is trying to find the ECDsa base class via the package, but it has a P2P reference over to Algorithms.  It's building fine on my Windows machine (using /p:Configuration=Linux_Debug), so I can't replicate the error to try to fix it.

I don't think we have a published package that contains the ECDsa base class in the contract assembly for S.S.C.Algorithms, since that was just checked in last Thursday.

Thoughts, anyone?

stephentoub:
> It's building fine on my Windows machine (using /p:Configuration=Linux_Debug),

What happens if you instead do /p:OSGroup=Linux ?

bartonjs:
> What happens if you instead do /p:OSGroup=Linux ?

```
Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:29.93
```

It looks like it's not using EvpPKey in this library yet, just pre-included (or over-included) interop files.  So I'm going to try dropping the files that talk about it, and at least get the branch unblocked.

bartonjs:
https://github.com/bartonjs/corefx/commit/19077421366ef6f4015bfbd25556a973c0cab6bb was the original commit, in case anyone wants to play with reproducing the error.

stephentoub:
LGTM

PR #3377, opened 2015-09-22
Implement GetECDsaPrivateKey() on Unix.

AtsushiKan:
cc @bartonjs

### DIFF
--- a/src/Common/src/Interop/Unix/libcrypto/Interop.EvpPkey.EcKey.cs
+++ b/src/Common/src/Interop/Unix/libcrypto/Interop.EvpPkey.EcKey.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.InteropServices;
+
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
+{
+    internal static partial class libcrypto
+    {
+        [DllImport(Libraries.LibCrypto)]
+        internal static extern SafeEcKeyHandle EVP_PKEY_get1_EC_KEY(SafeEvpPkeyHandle pkey);
+    }
+}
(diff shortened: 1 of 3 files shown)


bartonjs:
The ECDsa certificate tests can probably have their PlatformSpecific attribute removed now, no?

https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs#L145 (and others)

AtsushiKan:
The PublicKey extension method is still RSA-only on Linux. Given that we punted on Import/Export for ECDsa, I’m not sure there is a feasible way to get that implemented now.

bartonjs:
Oh, hm. those tests actually use CNG knowledge.

What about something like

```C#
byte[] expectedSignature = "steal this from the implementation";
byte[] data = ByteUtils.RepeatByte(0x37, 161);
byte[] signature;

using (X509Certificate2 cert = new X509Certificate2(TestData.ECDsa384Certificate))
using (ECDsa ecdsa = cert.GetECDsaPrivateKey())
{
    signature = ecdsa.SignData(words, go here);
}

Assert.Equal(expectedSignature, signature);
```

AtsushiKan:
We’d have to build a PFX or store a private key on the disk (not nice behavior for a unit test) to do private key testing (which is why the state of private key testing in X509.Tests is pretty sad right now in general.)

What’s our current state regarding PFX support in Unix?

AtsushiKan:
Also, we couldn’t compare the sigs byte for byte anyway. Intentional non-determinism and all…

bartonjs:
If the PFX contains exactly one cert and it has a private key, things are great.  The PFX bytes could be stored in TestData.

Except, now that I've said PFX, the test would likely want to be disabled on Windows, because we run into key concurrency problems there frequently (tests which use PFX-private keys need to all be put in one class to prevent concurrent execution).

So, sigh; I guess there's not really a good way to test this right now.

bartonjs:
Oh yeah, silly non-deterministic signatures :)

bartonjs:
As for GetECDsaPublicKey, I think that you can get there with `EVP_PKEY* X509_PUBKEY_get(X509_PUBKEY*)`.  Probably want to just add it to the native interop library as `EVP_PKEY* GetX509EvpPublicKey(X509* x509)`.  Then you can just unpack the EC_KEY the same way you did for the private key here.

If we don't have an open issue to remind us that we don't have GetECDsaPublicKey on Unix, please open one.

bartonjs:
Code LGTM.  Oh, please also open an issue saying we have no GetECDsaPrivateKey tests on Linux.

PR #3380, opened 2015-09-22
Unix X509Chains treats root certs in ExtraStore differently than Windows

### LINKED ISSUE #3226: Investigate if Unix and Windows X509Chains treat root certs in ExtraStore the same
Right now I think Unix says "you told me about it, clearly it's trusted" and Windows says "it wasn't in one of the trust stores, it isn't trusted".  So the Unix version needs to be less trusting.

But this behavior is currently unconfirmed.

eerhardt:
When building an X509Chain on Unix, if the root cert isn't in the system store, the building still succeeds with no errors for UntrustedRoot.  This is different than how it works on Windows.

I changed OpenSslX509ChainProcessor.BuildChain to check if the root cert is contained in the LocalMachine root or CurrentUser root store.  If it isn't, add an UntrustedRoot error.

Fix #3226.

@bartonjs


bartonjs:
LGTM. Continued thanks, @eerhardt  :)

PR #3395, opened 2015-09-23
adding the usage of SSL_CTX_set_quiet_shutdown

shrutigarg:


shrutigarg:
CC: @stephentoub, @davidsh, @CIPop, @SidharthNabar, @pgavlin, @bartonjs , @vijaykota

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -310,19 +312,14 @@ private static IntPtr GetSslMethod(bool isServer, long options)
...
                 int retVal = libssl.SSL_shutdown(sslPtr);
-                if (0 == retVal)
+                if (retVal < 0)
                 {
-                    retVal = libssl.SSL_shutdown(sslPtr);
-                    if (retVal < 0)
-                    {
-                        libssl.SSL_get_error(sslPtr, retVal);
-                    }
+                    libssl.SSL_get_error(sslPtr, retVal);
```
vijaykota: Since error value is not used, the if can be converted to a Debug.Assert(retVal == 1, "SSL shutdown failed with " + libssl.SSL_get_error(sslPtr, retVal))

LGTM otherwise

stephentoub:

stephentoub:
Can you provide more details about why we want to set this?  The docs for this option state "This behaviour violates the TLS standard."  Is SslStream doing this on Windows and thus we're matching its behavior?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
stephentoub: Does this work with an error queue like the rest of OpenSSL does?  If so, we wouldn't want to convert it to a Debug/Assert like that, as the SSL_get_error call would be removed in release and we'd leave the error in the queue.  We'd instead want to do something like:
```
if (retVal < 0)
{
    int err = libssl.SSL_get_error(ssPtr, retVal);
    Debug.Fail(...);
}
```

vijaykota:
Currently SslStream on Windows is not doing the handshake specified in https://msdn.microsoft.com/en-us/library/windows/desktop/aa380138(v=vs.85).aspx So in this PR, we are trying to cleanup the local resources

ellismg:
@dotnet-bot test this please (restarting after @bartonjs has merged master into dev-sslstream to pick up globalization fixes).

stephentoub:
@shrutigarg, I just noticed that the CLA still needs to be signed.  Can you please do that? Thanks.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: I think there is an error queue but it is specific to sslPtr. Since that is getting freed as the next step, this should be ok.
But even otherwise, if the error is left in the queue, the queue is supposed to be consulted only if some SSL API fails in which case the new error would be at the front of the queue, right?

stephentoub: It's not always documented which operations do or do not push errors into the queue, so we check for any operation that might.  That means that if an error is left in the queue, we may end up picking it up when querying for a completely unrelated operation.  At least that's how things were a while ago... @bartonjs can correct me if things have changed.

vijaykota: Oh I see. Then I agree it is safer to clear the queue.

bartonjs: Well, ERR_get_error is a queue.
SSL_get_error mostly reports back information on the internal state of `s` (or `sslPtr` in our calling context); it doesn't do anything with side effects that I see.

PR #3396, opened 2015-09-23
Fix handle leak in xplat SslStream.SafeFreeCredentials

vijaykota:
If a cert is used in SSL handshake, the certificate information needs
to persist for the SSL session. This is achieved by storing the cert
in a SafeFreeCredentials and bumping up its refcount. But we were
incorrectly doing this in AcquireCredentials instead of the first
invocation of InitializeSecurityContext/AcceptSecurityContext

cc: @stephentoub, @bartonjs, @rajansingh10, @shrutigarg, @pgavlin , @CIPop , @davidsh, @SidharthNabar

### DIFF
--- a/src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
+++ b/src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
@@ -27,15 +27,7 @@ namespace System.Net
         public SafeFreeCredentials AcquireCredentialsHandle(X509Certificate certificate,
             SslProtocols protocols, EncryptionPolicy policy, bool isServer)
         {
-            SafeFreeCredentials retVal = new SafeFreeCredentials(certificate, protocols, policy);
-            if ((null != retVal) && !retVal.IsInvalid)
-            {
-                // Caller does a ref count decrement
-                bool ignore = false;
-                retVal.DangerousAddRef(ref ignore);
-                //TODO retVal is not getting released now, need to be fixed.
-            }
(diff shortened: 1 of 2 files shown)


stephentoub:

ellismg:
@dotnet-bot test this please (restarting after @bartonjs has merged master into dev-sslstream to pick up globalization fixes).

bartonjs:
On the one hand, this feels a little weird.  On the other hand, the Windows version has a similar approach; and I don't see anything necessarily functionally wrong with it.

But I'm definitely having an unsure feeling, so I'd like someone else to decide it's the right change.  (Or, heck, even another opinion like mine)

stephentoub:
Am I understanding correctly that this is effectively a parent/child relationship between safe handles, that SafeDeleteContext uses a SafeFreeCredentials and thus the former AddRefs the latter when the former is constructed and ReleaseRefs the latter when the former is ReleaseHandle'd?  If so, that seems reasonable.

vijaykota:
> Am I understanding correctly that this is effectively a parent/child relationship 

This understanding is accurate. The code flow using SslStream can be paraphrased thusly:

```
SafeFreeCredentials cred = AcquireCredentials(..);
while (true)
{
    cred.AddRef();
    using (SafeDeleteContext session = new SafeDeleteContext(cred._handle, ...)
    {
        // Do SSL/TLS handshake
        // loop to process session read/write
    }
    cred.ReleaseRef();
    // break if no new session needed
}
cred.Dispose();
```

PR #3399, opened 2015-09-23
Fully import Min and Max tests

JonHanna:
Separate existing Min and Max tests.

Convert Min and Max tests imported in #2874 into a more directly XUnit-ian style, merging them with those already present.

Add some further tests for null checks, bringing branch coverage for the methods up to 100%.

The only duplicates in the legacy tests where a few that were both covering all values passed being `NaN` or all being `null`

Where a legacy test duplicated an existing test, I cut the corresponding portion from the existing test, as those test methods are doing too much each anyway.


VSadov:
LGTM
Thanks!

PR #3400, opened 2015-09-23
Aestest

AtsushiKan:
Move the Aes tests to the common directory and switch over to the AesFactory.Create() pattern so we can efficiently leverage these tests for AesCng whenever that class comes online.

cc @bartonjs


bartonjs:
LGTM

stephentoub:
LGTM2

PR #3401, opened 2015-09-23
Fully import legacy Select tests.

JonHanna:
Convert Select tests imported in #2874 into a more directly XUnit-ian style,
merging them with those already present.

Select1a.Test1a() left out as covered by
Select_SourceIsEmptyEnumerable_ReturnedCollectionHasNoElements()

Select2f.Test2f re-enabled, but made outerloop (takes over a minute to run).

Strengthen null-check tests.
Add parameter name check to existing tests. Add tests for overloads not
covered.


VSadov:
LGTM
Thanks!

PR #3402, opened 2015-09-23
Tweak and run native code formatting script

nguerrera:
* Rename script to format-code.sh to match naming convention of other shell scripts in the repo.

* Make script work when executed from any working directory.

* Run script to mop up some minor formatting issues that crept in.

### DIFF
--- a/src/Native/format-code.sh
+++ b/src/Native/format-code.sh
@@ -11,6 +11,8 @@ else
    exit 1
 fi
 
+cd $(dirname "$0")
+
 for D in */; do
     for file in "${D}"*.cpp; do
         if [ -e $file ] ; then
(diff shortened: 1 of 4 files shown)


nguerrera:
cc @sokket

src/Native/format-code.sh
```diff
@@ -11,6 +11,8 @@ else
    exit 1
 fi
 
+cd $(dirname "$0")
```
stephentoub: Should we pushd/pop?

stephentoub:
LGTM

src/Native/format-code.sh
nguerrera: cd doesn't propagate back to parent process, so I don't think it matters.

PR #3404, opened 2015-09-23
Fix several issues in UnixFileStream

stephentoub:
I'm trying to track down the cause of a sporadic failure in CI of the MemoryMappedFilesTests.DataShared test.  In doing so, I've run across a few issues to be fixed (though I'm not sure any of these is the cause):

- Fix the checking/setting of several FileOptions... we weren't correctly treating the options as a [Flags], which meant we'd incorrectly ignore some options if multiple options were specified.

- When opening a FileStream, if the PosixFAdvise or SeekCore failed, we weren't deterministically disposing of the file handle, instead leaving it up to finalization to close.

- If Close on the file descriptor fails (e.g. it's interrupted), we don't attempt a retry, by design.  However, closing of the handle is what releases the advisory lock we placed on the file descriptor when the UnixFileStream was constructed.  Added an explicit call to remove the lock, separate from closing the descriptor. (I thought this one would have been the cause of the sporadic test failure, but if it were the problem, the assert in SafeFileHandle.ReleaseHandle should have fired, and it didn't in the one occurence I saw since that assert got added.)

- Removed several stale TODOs and replaced them with explanatory comments.

- Added several more Debug.Asserts

Separately, I also tweaked TempFile to simplify its ctor to use File.WriteAllBytes.

cc: @nguerrera, @sokket, @ianhays

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
@@ -89,40 +89,42 @@ namespace System.IO
 
             // Open the file and store the safe handle. Subsequent code in this method expects the safe handle to be initialized.
             _fileHandle = SafeFileHandle.Open(path, openFlags, (int)openPermissions);
-            _fileHandle.IsAsync = _useAsyncIO;
-
-            // Lock the file if requested via FileShare.  This is only advisory locking. FileShare.None implies an exclusive 
-            // lock on the file and all other modes use a shared lock.  While this is not as granular as Windows, not mandatory, 
-            // and not atomic with file opening, it's better than nothing.
             try
             {
+                _fileHandle.IsAsync = _useAsyncIO;
+
+                // Lock the file if requested via FileShare.  This is only advisory locking. FileShare.None implies an exclusive 
+                // lock on the file and all other modes use a shared lock.  While this is not as granular as Windows, not mandatory, 
+                // and not atomic with file opening, it's better than nothing.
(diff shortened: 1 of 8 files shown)


src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
```diff
@@ -403,19 +402,24 @@ private void VerifyOSHandlePosition()
...
                     {
                         // Since we still have the file open, this will end up deleting
-                        // it (assuming we're the only link to it) once it's closed.
-                        Interop.Sys.Unlink(_path); // ignore any error
+                        // it (assuming we're the only link to it) once it's closed, but the
+                        // name will be removed immediatly.
+                        int result = Interop.Sys.Unlink(_path);
+                        Debug.Assert(result == 0,
+                            string.Format("unlink on {0} failed with result {1} and errno {2}",
+                                _path, result, Marshal.GetLastWin32Error()));
```
nguerrera: Not sure, but Interop.Sys.GetLastError() wold give easy to read enum in error message, but could drop unknown error down to NONSTANDARD. Maybe add a ToString() to ErrorInfo and use that to print both the friendly converted enum name and the raw errno?

nguerrera:
LGTM.

src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
stephentoub: Done.

nguerrera:
ToString LGTM as well. Thanks.

sokket:
LGTM

PR #3407, opened 2015-09-23
improve FileOptions tests for FileStream

stephentoub:
We didn't have any tests for the DeleteOnClose behavior with FileStream.  We also weren't testing combinations of FileOptions, nor validating that we could read/write files created with non-default options.

cc: @ianhays

### DIFF
--- a/src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs_buffer_fo.cs
+++ b/src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs_buffer_fo.cs
@@ -25,19 +25,58 @@ namespace System.IO.FileSystem.Tests
             Assert.Throws<ArgumentOutOfRangeException>("options", () => CreateFileStream(GetTestFilePath(), FileMode.Create, FileAccess.ReadWrite, FileShare.Read, c_DefaultBufferSize, ~FileOptions.None));
         }
 
-        [Fact]
-        public void ValidFileOptions()
+        [Theory]
+        [InlineData(FileOptions.None)]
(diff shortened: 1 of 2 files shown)


src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs_buffer_fo.cs
```diff
@@ -25,19 +25,58 @@ public void InvalidFileOptionsThrow()
...
-        [Fact]
-        public void ValidFileOptions()
+        [Theory]
+        [InlineData(FileOptions.None)]
+        [InlineData(FileOptions.DeleteOnClose)]
+        [InlineData(FileOptions.Encrypted)]
+        [InlineData(FileOptions.RandomAccess)]
+        [InlineData(FileOptions.SequentialScan)]
+        [InlineData(FileOptions.WriteThrough)]
+        [InlineData((FileOptions)0x20000000)] // FILE_FLAG_NO_BUFFERING on Windows
```
ianhays: But still allowed on Unix?

ianhays:
LGTM

src/System.IO.FileSystem/tests/FileStream/ctor_str_fm_fa_fs_buffer_fo.cs
stephentoub: Yeah, we just ignore it.  I considered mapping it to O_DIRECT, but that's not portable and it comes with additional requirements, plus this option isn't actually part of the surface area.

stephentoub:
Thanks, Ian.

PR #3411, opened 2015-09-23
Implement IDNA2008 on top of ICU

### LINKED ISSUE #810: Implement System.Globalization.Extensions on Unix
The library has been refactored to have the Unix implementation stubbed out.  It needs to be properly implemented.

### LINKED ISSUE #2779: System.Globalization.Extensions: IDNA
We need to support [IdnMapping.cs](https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Globalization/IdnMapping.Unix.cs) for IDNA2008 on Unix.

This is straightforward on-top of ICU, but many of our existing ICU tests assume either ICU2003 or ICU2008 as implemented by Win32.  There are some differences between 2008 and 2003 in ICU for compatibility, especially around how windows works when UseStd3Rules is false.

In speaking with NLS, they think it may make sense for us to just follow the 2008 standard.

ellismg:
This change uses the ToUnicode and ToAscii methods from
System.Globalization.Native which are implemented on top of ICU's UTS46
implementation.

There are some test issues that need to be addressed due to differences
between Windows implementation of the standard and ICUs (which follows
the standard more closely). These will be addressed at a later date. I
have confirmed manually that things are working as expected.

Fixes #2779
Fixes #810

### DIFF
--- a/src/Common/src/Interop/Unix/System.Globalization.Native/Interop.Idna.cs
+++ b/src/Common/src/Interop/Unix/System.Globalization.Native/Interop.Idna.cs
@@ -1,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class GlobalizationNative
+    {
+        internal const int AllowUnassigned = 0x1;
+        internal const int UseStd3AsciiRules = 0x2;
+
+        [DllImport(Libraries.GlobalizationNative, CharSet = CharSet.Unicode)]
+        internal static extern int ToAscii(uint flags, string src, int srcLen, char[] dstBuffer, int dstBufferCapacity);
+
+        [DllImport(Libraries.GlobalizationNative, CharSet = CharSet.Unicode)]
+        internal static extern int ToUnicode(uint flags, string src, int srcLen, char[] dstBuffer, int dstBufferCapacity);
+    }
+}
--- a/src/Common/src/System/Globalization/IdnMapping.Unix.cs
+++ b/src/Common/src/System/Globalization/IdnMapping.Unix.cs
@@ -5,35 +5,67 @@ namespace System.Globalization
 {
     sealed partial class IdnMapping
     {
-        // TODO: Rather than having a Unix-specific implementation, we could also consider having a purely
-        // managed implementation as is available in mscorlib for pre-Windows 8.  This would require including 
-        // the relevant data tables.
-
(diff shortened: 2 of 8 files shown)


src/Common/src/Interop/Unix/System.Globalization.Native/Interop.Idna.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal static partial class GlobalizationNative
+    {
+        internal const int AllowUnassigned = 0x1;
+        internal const int UseStd3AsciiRules = 0x2;
+
+        [DllImport(Libraries.GlobalizationNative, CharSet = CharSet.Unicode)]
+        internal static extern int ToAscii(uint flags, string src, int srcLen, [Out] char[] dstBuffer, int dstBufferCapacity);
+
+        [DllImport(Libraries.GlobalizationNative, CharSet = CharSet.Unicode)]
+        internal static extern int ToUnicode(uint flags, string src, int srcLen, [Out] char[] dstBuffer, int dstBufferCapacity);
```
stephentoub: Does the [Out] actually impact anything?  If not, that's fine, just verifying.

stephentoub:
LGTM

src/Common/src/Interop/Unix/System.Globalization.Native/Interop.Idna.cs
ellismg: I don't believe so and I will remove this.  Thinking about it more, the default should be in/out and the marshaler should just pin the array.

justinvp: With MCG, there may be a slight perf benefit to specifying `[Out]`. Not sure if this applies to the CoreCLR marshaller.

From http://blogs.msdn.com/b/dotnet/archive/2014/06/13/net-native-deep-dive-debugging-into-interop-code.aspx:

> However, if you look very closely, there is actually a step that is a bit unnecessary: MCG copies the contents of the StringBuilder into the buffer, while the API doesn't expect you to pass in any string value at all - it simply fill in the buffer. The copying step is actually wasted. Fortunately, .NET allows you to customize the marshaling behavior by specifying [In]/[Out] attribute on the parameter (for more information, you can refer to this MSDN article). StringBuilder by default gets [In, Out], and in this case, we only want the [Out] direction because we only need to receive data from the LCIDToLocaleName API

ellismg: Yeah, I'm not sure if that optimization matters when you are trying to marshal a char array to native code and CharSet is Unicode.  Ideally, you would just pin the thing and be done with it.

justinvp: Yeah, you're right; with `char[]` it just pins the array without the need for any copying.

PR #3412, opened 2015-09-23
OpenSSL multi-threading support on OSX

### LINKED ISSUE #2210: Consider multi-threading implications on the OpenSSL ERR_get_error queue.
OpenSSL compiled with OPENSSL_FIPS seems to have a single static ERR_STATE object for the process.  The default appears to be set up for providing the capabilities for stack-per-thread, but OPENSSL_thread_id() appears to return the process ID, not the thread ID.

The mainline seems to be served by `OPENSSL_set_id_callback` to something returning a thread ID; and lightup for FIPS could try invoking `int_ERR_lib_init`, which resets the library to the non-FIPS way of error management (though that theoretically has FIPS implications...).

eerhardt:
In order to support multi-threading in OpenSSL on OSX, we need to set the CRYPTO_set_id_callback to a method that will return the current thread id.

This is necessary because OSX uses a version of OpenSSL less than 1.0.  Before 1.0, setting this callback was necessary because the default uses getpid(), which will be the same value for multiple threads in the same process.  Any version 1.0 or greater of OpenSSL will use the address of errno to differentiate between threads.

Fix #2210

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -1063,6 +1063,26 @@ LockingCallback(int mode, int n, const char* file, int line)
     }
 }
 
+#ifdef __APPLE__
+/*
+Function:
+GetCurrentThreadId
+
+Called back by OpenSSL to get the current thread id.
+
+This is necessary because OSX uses an earlier version of
+OpenSSL, which requires setting the CRYPTO_set_id_callback.
+*/
+static
+unsigned long
+GetCurrentThreadId()
+{
+    uint64_t tid;
+    pthread_threadid_np(pthread_self(), &tid);
+    return tid;
+}
+#endif // __APPLE__
+
 /*
 Function:
 EnsureOpenSslInitialized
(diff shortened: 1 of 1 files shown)


eerhardt:
@bartonjs, @stephentoub

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -1063,6 +1063,25 @@ LockingCallback(int mode, int n, const char* file, int line)
...
+/*
+Function:
+GetCurrentThreadId
+
+Called back by OpenSSL to get the current thread id.
+
+This is necessary because OSX uses an earlier version of
+OpenSSL, which requires setting the CRYPTO_set_id_callback.
+*/
+unsigned long
```
bartonjs: I'm not a C guru, but since we don't really need to target this function outside of this file, can/should it be declared static?

bartonjs:
Aside from the one goodness comment, LGTM.

src/Native/System.Security.Cryptography.Native/openssl.c
eerhardt: Fixed.

bartonjs:
:shipit:

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -1063,6 +1063,26 @@ LockingCallback(int mode, int n, const char* file, int line)
     }
 }
 
+#if __APPLE__
```
bartonjs: We write too much C# :)

```C
#ifdef __APPLE__
...
```

eerhardt: :( it built on my mac...  Fixed.

stephentoub:
@dotnet-bot test this please (Jenkins error "Hung processs on target machine")

PR #3413, opened 2015-09-23
Resurrect NuGet.config for a more pleasant VS experience

nguerrera:
The deletion of nuget.config made VS unaware of our feeds and was causing builds in VS 2015 to "fail" package restore, but still build. So even successful builds had "errors" in the  error list, which was most annoying.

Putting the file back gets clean builds without any junk in the error list. :)

cc @ericstj @weshaggard @jhendrixMSFT

### DIFF
--- a/src/NuGet.config
+++ b/src/NuGet.config
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <!-- NOTE: Leave this file here and keep it in sync with list in dir.props. -->
+  <!-- The command-line doesn't need it, but the IDE does.                    -->
+  <packageSources>
+    <clear/>
+    <add key="myget.org dotnet-core" value="https://www.myget.org/F/dotnet-core/" />
+    <add key="myget.org dotnet-coreclr" value="https://www.myget.org/F/dotnet-coreclr/" />
+    <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
+    <add key="myget.org dotnet-corefxtestdata" value="https://www.myget.org/F/dotnet-corefxtestdata/" />
+    <add key="myget.org dotnet-buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
+    <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
+  </packageSources>
+  <config>
+    <add key="repositoryPath" value="..\packages" />
+  </config>
+</configuration>
--- a/src/NuGet.config
+++ b/src/NuGet.config
@@ -6,7 +6,6 @@
     <clear/>
     <add key="myget.org dotnet-core" value="https://www.myget.org/F/dotnet-core/" />
     <add key="myget.org dotnet-coreclr" value="https://www.myget.org/F/dotnet-coreclr/" />
-    <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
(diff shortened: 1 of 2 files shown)


nguerrera:
Hmm, it's not working as expected now, but was for a minute. Digging and will reopen when I get to the bottom of it.

weshaggard:
@nguerrera if I remember correctly DNU didn't respect this which is why we switch to using the source command line.

ericstj:
DNU does honor it, but at different precedence.  We shouldn't be relying on it for dnu/nuget but should keep it around and keep it in sync for the IDE.

nguerrera:
@weshaggard +1 to what @ericstj said. I've reopened with a version that exactly matches the --source arguments we pass and a comment as to why we should keep this file. Things are much better in VS 2015 with this. 

There are other issues such as VS having a hard time with P2P refererences that we have in places going outside the sln, which is why I closed it yesterday before understanding that it's a completely separate issue. This is strictly better than status quo.

src/NuGet.config
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <!-- NOTE: Leave this file here and keep it in sync with list in dir.props. -->
+  <!-- The command-line doesn't need it, but the IDE does.                    -->
+  <packageSources>
+    <clear/>
+    <add key="myget.org dotnet-core" value="https://www.myget.org/F/dotnet-core/" />
+    <add key="myget.org dotnet-coreclr" value="https://www.myget.org/F/dotnet-coreclr/" />
+    <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
```
weshaggard: dotnet-corefx and dotnet-coreclr shouldn't actually be needed. I want to clean them out of the dirs.props as well.

src/NuGet.config
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <!-- NOTE: Leave this file here and keep it in sync with list in dir.props. -->
+  <!-- The command-line doesn't need it, but the IDE does.                    -->
```
weshaggard: I think I would actually make this a little more specific and say the command line tool doesn't correctly support this yet but we want to use the config file for both in the future.

src/NuGet.config
nguerrera: I'll do it.

weshaggard:
Other then the couple comments I'm cool with this. I want to make our VS experience work well. We should also give thought about our VS2013 experience and decide if we want to carry that forward now since VS2015 has RTM'ed.

src/NuGet.config
nguerrera: What is "incorrect" about the command-line tool today. Is that something being tracked as a bug. This comment accurately reflects the status quo. If we want to merge these lists, we can open a bug on that and link to dnu/nuget bugs as appropriate. I'd rather not put the TODO here.

src/NuGet.config
nguerrera: This happens when I remove dotnet-coreclr:

D:\Src\corefx\build.proj(24,5): error : Unable to locate Microsoft.DotNet.CoreCLR 1.0.0-prerelease

weshaggard: Another thing I want to remove from our builds. We should try to start using Microsoft.NETCore.Runtime.CoreCLR instead of this package. Can you please file an issue on removing that dependency?

nguerrera: #3439

PR #3415, opened 2015-09-24
ET Interpreter: More null checks on instance member accesses.

VSadov:
Accessing an instance member on null should throw NRE.
We were already handling field accesses and method calls, but properties and indexers were getting around it and could result in InvalidOperationException from reflection, which is unexpected.

Since all instance accesses eventually result in either a field access or a call, performing these checks on the spot seems more robust and also cheaper than to represent them as a special instruction. Therefore NullCheckInstruction is gone.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Instruction.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Instruction.cs
@@ -55,6 +55,15 @@ namespace System.Linq.Expressions.Interpreter
         {
             return null;
         }
+
+        // throws NRE when o is null
+        protected static void NullCheck(object o)
+        {
+            if (o == null)
+            {
+                o.GetType();
+            }
+        }
     }
 
     internal abstract class NotInstruction : Instruction
--- a/src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
+++ b/src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
@@ -229,17 +229,7 @@ namespace Tests.ExpressionCompiler.MemberAccess
                     Enumerable.Empty<ParameterExpression>());
             Func<int> f = e.Compile();
 
-            var err = default(Exception);
-            try
-            {
-                f();
-            }
-            catch (NullReferenceException ex)
-            {
-                err = ex;
-            }
-
-            Assert.NotNull(err);
+            Assert.Throws<NullReferenceException>(()=>f());
         }
 
         [Fact] // [Issue(3217, "https://github.com/dotnet/corefx/issues/3217")]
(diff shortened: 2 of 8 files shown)


VSadov:
@AlekseyTs @stephentoub - please review

VSadov:
related to #3217

src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
```diff
@@ -243,6 +243,32 @@ public static void CheckMemberAccessClassInstanceFieldNullReferenceTest()
...
+            try
+            {
+                f();
+            }
+            catch (NullReferenceException ex)
+            {
+                err = ex;
+            }
+
+            Assert.NotNull(err);
```
stephentoub: Couldn't this second half of the function just be:
```C#
Assert.Throws<NullReferenceException>(() => f());
```
?

src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
```diff
@@ -265,5 +291,57 @@ public static void CheckMemberAccessClassInstancePropertyNullReferenceTest()
...
```
stephentoub: Ditto

src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
```diff
@@ -265,5 +291,57 @@ public static void CheckMemberAccessClassInstancePropertyNullReferenceTest()
...
```
stephentoub: Ditto

stephentoub:
Other than the comments on the tests, LGTM.

src/System.Linq.Expressions/tests/Member/MemberAccessTests.cs
VSadov: i was lazy, I just copy-pasted existing test. 
Makes sense. Will change.

src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Instruction.cs
```diff
@@ -55,6 +55,15 @@ public virtual object GetDebugCookie(LightCompiler compiler)
         {
             return null;
         }
+
+        // throws NRE when o is null
+        protected static void NullCheck(object o)
+        {
+            if (o == null)
+            {
+                o.GetType();
```
VSadov: unused   "o.getClass" is an idiomatic null check in java (since unused getClass is otherwise not considered sideeffecting). 
Unfortunately our JITs seem to emit the actual call, so I need the condition above.

I wonder if this is worth entering a bug...

VSadov:
@bartdesmet - in case you are looking at something related

bartdesmet:
Cool. NullCheckInstruction looked at bit out of place to me as well. On vacation now, so wasn't looking at the code :-).

PR #3424, opened 2015-09-24
Tidy up enumerable tests after import of legacy completed

JonHanna:
Move tests from EnumerableTests into specific classes (average tests dropped
as all duplicates).

Change some Fact tests calling helpers into Theory tests fed by properties.

Make EnumerableTests base of test classes.

Move repeated methods and classes into EnumerableTests and inherit. In some
cases just replacing such classes with anonymous classes.

Check parameter name on all argument exception tests.

Add tests for check of null key‎Selector on OrderBy, ThenBy, etc.


src/System.Linq/tests/CountTests.cs
```diff
@@ -128,5 +118,34 @@ public void MultipleElementsAllMatchPredicate()
...
+        {
+            yield return new object[] { default(T), count, enumerable };
+            yield return new object[] { default(T), count, enumerable.ToArray() };
+            yield return new object[] { default(T), count, enumerable.ToList() };
+            yield return new object[] { default(T), count, new Stack<T>(enumerable) };
+        }
+
+        public static IEnumerable<object[]> CountsAndTallies()
+        {
+            int count = 100;
```
stephentoub: Minor, but we could use a smaller number here.

src/System.Linq/tests/DistinctTests.cs
```diff
@@ -164,5 +136,57 @@ public void CustomEqualityComparer()
...
+            // Convert to list to avoid repeated enumerations of the enumerables.
+            var originalList = original.ToList();
+            var distinctList = originalList.Distinct().ToList();
+
+            // Ensure the result doesn't contain duplicates.
+            var hashSet = new HashSet<T>();
+            foreach (var i in distinctList)
+            {
+                Assert.False(hashSet.Contains(i), "Found a duplicate in the result of Distinct(): " + i);
+                hashSet.Add(i);
```
stephentoub: Rather than assertin on Contains, you could assert that Add returns true, right?

src/System.Linq/tests/DistinctTests.cs
```diff
@@ -164,5 +136,57 @@ public void CustomEqualityComparer()
...
+            }
+
+            // Ensure that every element in 'originalList' exists in 'distinctList'.
+            Assert.DoesNotContain(originalList, x => !distinctList.Contains(x));
+
+            // Ensure that every element in 'distinctList' exists in 'originalList'.
+            Assert.DoesNotContain(distinctList, x => !originalList.Contains(x));
```
stephentoub:
```C#
var originalSet = new HashSet<T>(originalList);
Assert.Superset(originalList, hashSet);
Assert.Superset(hashSet, originalList);
```
?

src/System.Linq/tests/CountTests.cs
JonHanna: Yes, it anything it would make it cheaper to add further variants of there was ever a reason to do so.

src/System.Linq/tests/EmptyEnumerable.cs
```diff
@@ -61,5 +61,14 @@ public void CastToArray()
...
         }
+
+        [Fact]
+        public void CachedEnumerator()
+        {
+            var emptyEnumerable1 = Enumerable.Empty<object>();
+            var emptyEnumerable2 = (IList<object>)emptyEnumerable1;
+
+            Assert.Same(emptyEnumerable1.GetEnumerator(), emptyEnumerable2.GetEnumerator());
+        }
```
stephentoub: Instead of moving this test here, how about deleting it and instead adding the line:
```
Assert.Same(enumerable1.GetEnumerator(), enumerable2.GetEnumerator());
```
to the end of ```TestEmptyCached<T>()```.  That way we verify this for all of the tests that use that helper with various Ts.

src/System.Linq/tests/DistinctTests.cs
JonHanna: Good idea. On a related note, there's a lot of places that should be Assert.Empty but aren't.

src/System.Linq/tests/EnumerableTests.cs
```diff
@@ -1,178 +1,291 @@
...
 
+using System.Collections;
 using System.Collections.Generic;
 using Xunit;
+using Xunit.Extensions;
 
 namespace System.Linq.Tests
 {
-    public partial class EnumerableTests
+    public class EnumerableTests
```
stephentoub: Consider making this abstract to highlight that it's meant to be used only as a base class.

src/System.Linq/tests/OrderByTests.cs
```diff
@@ -165,28 +157,33 @@ public void SameKeysVerifySortStable()
...
 
             Assert.Equal(expected, source.OrderBy(e => e.Score).ToArray());
         }
 
+        [Fact]
+        public void EmptyOrderedToArray()
+        {
+            Assert.Empty(Enumerable.Empty<int>().OrderBy(e => e).ToArray());
+        }
         //FIXME: This will hang with a larger source. Do we want to deal with that case?
```
stephentoub: What's this about?

stephentoub: Nit: please add a blank line above this comment

stephentoub:
Left a few comments, but overall looks good.  Thanks!

src/System.Linq/tests/OrderByTests.cs
JonHanna: This originates from a test in the legacy tests, but as said, it doesn't actually work with larger inputs. I'm not sure the test should be there at all (we should just hang on such bad comparers and consider a complaint about it PEBKAC) but I remained faithful to the legacy tests in leaving these two in.

JonHanna:
Not to be unaware of the fallacy of considering branch coverage as a pure indicator of test quality in and of itself, but **W00t! `System.Linq.Enumerable` branch coverage at 100%!**

(Normally it's pointless to hit the first branch of a compiler-produced iterator, that can only be hit with `((IEnumerator<T>)enumerable).MoveNext()` which should always return false, but since we've a lot of hand-coded classes that take the same approach, and could hence potentially add a buggy one in the future, it's worth testing for).

src/System.Linq/tests/AverageTests.cs
```diff
@@ -44,6 +38,25 @@ public void EmptyNullableFloatSource()
         }
 
         [Fact]
+        public void NullNFloatSource()
+        {
+            Assert.Throws<ArgumentNullException>("source", () => default(IEnumerable<float?>).Average());
```
stephentoub: @jaredpar, in our CI system, this (and a bunch of these other tests) is producing the warning as error:
```
AverageTests.cs(43,66): error CS1720: Warning as Error: Expression will always cause a System.NullReferenceException because the default value of 'System.Collections.Generic.IEnumerable<float?>' is null [d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Linq\tests\System.Linq.Tests.csproj]
```
Is that just a bug in the old C# compiler?  These are extension methods, so this warning is wrong.

jaredpar: @stephentoub it's a bug in the native compiler.  I tried out this minimal repro and verified it warns in the native compiler but not in Roslyn.

``` csharp
    interface ITest { }
    static class Program
    {
        static void M(this ITest t) { }

        {
            default(ITest).M();
        }
    }
```

JonHanna: It's easily worked around, as far as this PR requires.

stephentoub: Thanks, guys.

stephentoub:
@dotnet-bot test this please (Jenkins failure to download some files on Linux)

stephentoub:
@dotnet-bot test this please (lots of FindPackagesById failures... grr)

JonHanna:
And a push to master broke it anyway. Nevermind, I spotted a couple of typos in the meantime, so now I get to fix them too.

stephentoub:
LGTM.  Thanks for the new test cases, and woohoo on 100% branch coverage of Enumerable! Nice job.

PR #3429, opened 2015-09-24
Enumerable.ToList() Optimization

justinvp:
Via a new internal `IListProvider<TElement>` interface similar to `IArrayProvider<TElement>` for `ToArray()`.

### DIFF
--- a/src/System.Linq/src/System/Linq/Enumerable.cs
+++ b/src/System.Linq/src/System/Linq/Enumerable.cs
@@ -529,7 +529,7 @@ namespace System.Linq
         }
 
 
-        internal sealed class SelectArrayIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>
+        internal sealed class SelectArrayIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>, IListProvider<TResult>
         {
             private readonly TSource[] _source;
             private readonly Func<TSource, TResult> _selector;
@@ -573,9 +573,20 @@ namespace System.Linq
                 }
                 return results;
             }
+
+            public List<TResult> ToList()
+            {
+                TSource[] source = _source;
+                var results = new List<TResult>(source.Length);
+                for (int i = 0; i < source.Length; i++)
+                {
+                    results.Add(_selector(source[i]));
+                }
+                return results;
+            }
         }
 
-        internal sealed class SelectListIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>
+        internal sealed class SelectListIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>, IListProvider<TResult>
         {
             private readonly List<TSource> _source;
             private readonly Func<TSource, TResult> _selector;
@@ -628,9 +639,20 @@ namespace System.Linq
                 }
                 return results;
             }
+
+            public List<TResult> ToList()
+            {
+                int count = _source.Count;
+                var results = new List<TResult>(count);
+                for (int i = 0; i < count; i++)
+                {
+                    results.Add(_selector(_source[i]));
+                }
+                return results;
+            }
         }
 
-        internal sealed class SelectIListIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>
+        internal sealed class SelectIListIterator<TSource, TResult> : Iterator<TResult>, IArrayProvider<TResult>, IListProvider<TResult>
         {
             private readonly IList<TSource> _source;
             private readonly Func<TSource, TResult> _selector;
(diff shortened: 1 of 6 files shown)


src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -693,6 +713,16 @@ public TResult[] ToArray()
                 }
                 return results;
             }
+
+            public List<TResult> ToList()
+            {
+                var results = new List<TResult>(_source.Count);
+                for (int i = 0; i < _source.Count; i++)
```
stephentoub: Count here is an interface call.  I'd suggest storing the result into a local and then using the local.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -573,9 +573,19 @@ public TResult[] ToArray()
                 }
                 return results;
             }
+
+            public List<TResult> ToList()
+            {
+                var results = new List<TResult>(_source.Length);
```
stephentoub: There's something interesting with List to consider here, and in all of the other cases below where we presize the list.  Previously, we'd just use the normal ```List<T>``` ctor, which would check for ```ICollection<T>``` and use its count if it could or else just use the normal list growth algorithm, starting at 0, growing to 4 for the first element, and then doubling after that.   Since all of these cases didn't implement ```ICollection<T>```, we would get the standard 0/4/doubling behavior.  And that means that, for example, if you did ```new int[2] { ... }.Select(...).ToList()```, you'd get back a list with a capacity of 4 rather than 2, which means that if the caller added an additional item or two into the list, it wouldn't result in the list needing to grow.

It's probably relatively rare to add items to the returned list, so what you have here is probably the right thing to do, but I wanted to mention it.

JonHanna: I've an idea for passing `ICollection<T>` instances to the `List<T>` ctor, among other places, which would give the benefit of this and some others, but the best way to do it depends on what we do about #3313.

src/System.Linq/tests/ToListTests.cs
```diff
@@ -135,7 +136,79 @@ public void ToList_WorkWhenCountChangedAsynchronously()
...
+        [InlineData(new int[] { 1, 2, 3 }, new string[] { "1", "2", "3" })]
+        public void ToList_IListWhereSelect(int[] sourceIntegers, string[] convertedStrings)
+        {
+            IList<int> sourceList = new ReadOnlyCollection<int>(sourceIntegers);
+            IList<string> convertedList = new ReadOnlyCollection<string>(convertedStrings);
+
+            IList<int> emptyIntegersList = new ReadOnlyCollection<int>(Array.Empty<int>());
+            IList<string> emptyStringsList = new ReadOnlyCollection<string>(Array.Empty<string>());
+
+            Assert.Equal(convertedList, sourceList.Select(i => i.ToString()).ToList());
```
stephentoub: This assertion is failing for me locally, e.g.
```
System.Linq.Tests.ToListTests.ToList_IListWhereSelect(sourceIntegers: [1], convertedStrings: ["1"]) [FAIL]
        Assert.Equal() Failure
        Expected: ReadOnlyCollection<String> ["1"]
        Actual:   List<String> ["1"]
```

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -573,9 +573,19 @@ public TResult[] ToArray()
...
+                for (int i = 0; i < _source.Length; i++)
+                {
```
VSadov: You need to capture _source into a local here to let JIT optimize the range check. Otherwise JIT must assume that _source may change on every iteration and will not optimize.

src/System.Linq/src/System/Linq/Enumerable.cs
```diff
@@ -628,9 +638,19 @@ public TResult[] ToArray()
                 }
                 return results;
             }
+
+            public List<TResult> ToList()
+            {
+                var results = new List<TResult>(_source.Count);
+                for (int i = 0; i < _source.Count; i++)
+                {
```
VSadov: for (int i = 0, cnt = _source.Count; i < cnt; i++)

stephentoub: (You probably want to move the setting of cnt to before constructing the list so you can use the cnt in the ctor call as well.)

src/System.Linq/tests/ToListTests.cs
justinvp: Ah, I changed the declarations from `var` to `IList<T>` at the last minute before submitting, and forgot to re-run the tests locally.

stephentoub:
LGTM now.  Thanks, Justin.

PR #3432, opened 2015-09-24
Adjust the PAL separation for certificate verification

bartonjs:
The Windows version requires that there be a valid handle to the native chain representation; the Unix version never has a valid native handle.

So move the PAL separation to a slightly higher abstraction level, and let each side party on it in their own way.

### DIFF
--- a/src/System.Net.Security/src/System/Net/Windows/CertModule.cs
+++ b/src/System.Net.Security/src/System/Net/Windows/CertModule.cs
@@ -19,44 +19,63 @@ namespace System.Net
         private static volatile X509Store s_myMachineCertStoreEx;
 
         #region internal Methods
-        internal override SslPolicyErrors VerifyRemoteCertName(X509Chain chain, bool isServer, string hostName)
+        internal override SslPolicyErrors VerifyCertificateProperties(X509Chain chain, X509Certificate2 certificate, bool checkCertName, bool isServer, string hostName)
         {
             SslPolicyErrors sslPolicyErrors = SslPolicyErrors.None;
 
-            unsafe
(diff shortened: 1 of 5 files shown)


src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -19,46 +19,65 @@ internal partial class CertModule : CertInterface
...
 
         #region internal Methods
-        internal override SslPolicyErrors VerifyRemoteCertName(X509Chain chain, bool isServer, string hostName)
+        internal override SslPolicyErrors VerifyCertificateProperties(X509Chain chain, X509Certificate2 certificate, bool checkCertName, bool isServer, string hostName)
         {
             SslPolicyErrors sslPolicyErrors = SslPolicyErrors.None;
 
-            unsafe
+            if (!chain.Build(certificate)       // Build failed on handle or on policy.
+                && chain.SafeHandle.DangerousGetHandle() == IntPtr.Zero)   // Build failed to generate a valid handle.
```
stephentoub: Not chain.SafeHandle.IsInvalid?

stephentoub: (After getting to the end I see this is just the same logic moved from elsewhere.)

stephentoub:
LGTM

bartonjs:
cc @CIPop, @DavidSh, @JosGuil @SidharthNabar, @PGavlin, @rajansingh10, @vijaykota

stephentoub:
@dotnet-bot test this please

Issue #3391, opened 2015-09-23
TestDateTimeParsingWithSpecialCultures failing on ICU for sr-Latn-XX


ellismg:
As part of b1995378c63df05117c57505c50f46e256c32b3a Tarek added some additional cultures to TestDateTimeParsingWithSpecialCultures in System.Runtime.  These additional tests don't pass when run with ICU.

```
MESSAGE:
Assert.True() Failure Expected: True Actual: False
+++++++++++++++++++
STACK TRACE:
at DateTimeTests.TestDateTimeParsingWithSpecialCultures(String cultureName)
```

nguerrera:
@tarekgh @ellismg These tests are also failing locally for me on Windows 10 RTM.

Should I file a different bug? Disable on Windows as well?

```
  DateTimeTests.TestDateTimeParsingWithSpecialCultures(cultureName: "sr-Cyrl-XK") [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           D:\Src\corefx\src\System.Runtime\tests\System\DateTime.cs(365,0): at DateTimeTests.TestDateTimeParsingWithSpecialCultures(String cultureName)
     DateTimeTests.TestDateTimeParsingWithSpecialCultures(cultureName: "sr-Latn-ME") [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           D:\Src\corefx\src\System.Runtime\tests\System\DateTime.cs(365,0): at DateTimeTests.TestDateTimeParsingWithSpecialCultures(String cultureName)
     DateTimeTests.TestDateTimeParsingWithSpecialCultures(cultureName: "sr-Latn-RS") [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           D:\Src\corefx\src\System.Runtime\tests\System\DateTime.cs(365,0): at DateTimeTests.TestDateTimeParsingWithSpecialCultures(String cultureName)
  Finished:    System.Runtime.Tests
     DateTimeTests.TestDateTimeParsingWithSpecialCultures(cultureName: "sr-Latn-XK") [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           D:\Src\corefx\src\System.Runtime\tests\System\DateTime.cs(365,0): at DateTimeTests.TestDateTimeParsingWithSpecialCultures(String cultureName)
  ```

[...]

PR #3433, opened 2015-09-24
Disable test against #3391 on Windows too.

nguerrera:
The test is failing on Windows 10 as well.

cc @tarekgh @ellismg


stephentoub:
LGTM

stephentoub:
This failed due to the Jenkins issue we've been seeing today. But since the change doesn't affect Linux and that's the leg that failed, I'm going to merge.

Issue #3421, opened 2015-09-24
DateTimeFormatInfoGetInstance.PosTest2 failed on Linux in CI


stephentoub:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2662/console
```
03:20:10    System.Globalization.Tests.DateTimeFormatInfoGetInstance.PosTest2 [FAIL]
03:20:10       Assert.Equal() Failure
03:20:10       Expected: DateTimeFormatInfo { AMDesignator = "AM", AbbreviatedDayNames = ["Sun", "Mon", "Tue", "Wed", "Thu", ...], AbbreviatedMonthGenitiveNames = ["Jan", "Feb", "Mar", "Apr", "May", ...], AbbreviatedMonthNames = ["Jan", "Feb", "Mar", "Apr", "May", ...], Calendar = GregorianCalendar { CalendarType = Localized, Eras = [...], IsReadOnly = True, MaxSupportedDateTime = 9999-12-31T23:59:59.9999999, MinSupportedDateTime = 0001-01-01T00:00:00.0000000, ... }, ... }
03:20:10       Actual:   DateTimeFormatInfo { AMDesignator = "de.", AbbreviatedDayNames = ["V", "H", "K", "Sze", "Cs", ...], AbbreviatedMonthGenitiveNames = ["jan.", "febr.", "m��rc.", "��pr.", "m��j.", ...], AbbreviatedMonthNames = ["jan.", "febr.", "m��rc.", "��pr.", "m��j.", ...], Calendar = GregorianCalendar { CalendarType = Localized, Eras = [...], IsReadOnly = False, MaxSupportedDateTime = 9999-12-31T23:59:59.9999999, MinSupportedDateTime = 0001-01-01T00:00:00.0000000, ... }, ... }
03:20:10       Stack Trace:
03:20:11             at System.Globalization.Tests.DateTimeFormatInfoGetInstance.PosTest2()
03:20:11 Finished:    System.Globalization.Tests
```

stephentoub:
PosTest5 just faild as well:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2735/console
```
14:14:24    System.Globalization.Tests.DateTimeFormatInfoGetInstance.PosTest5 [FAIL]
14:14:24       Assert.Equal() Failure
[...]

PR #3443, opened 2015-09-25
Remove failing globalization test on Linux

steveharter:
Root cause TBD


nguerrera:
Should the active be #3421 and not #846?

LGTM

stephentoub:
LGTM

PR #3448, opened 2015-09-25
Remove overaggressive assert in UnixFileStream

stephentoub:
I'd added an assert that fires if a DeleteOnClose file's unlink during disposal fails.  But it's valid for it to fail, e.g. if something elsewhere in this process or otherwise deletes or moves the file while it's open.  This happens, for example, in a bunch of our FileSystemWatcher tests.

Removing the assert

cc: @nguerrera, @sokket

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
@@ -418,10 +418,7 @@ namespace System.IO
                         // Since we still have the file open, this will end up deleting
                         // it (assuming we're the only link to it) once it's closed, but the
                         // name will be removed immediatly.
-                        int result = Interop.Sys.Unlink(_path);
-                        Debug.Assert(result == 0,
-                            string.Format("unlink on {0} failed with result {1} and error {2}",
-                                _path, result, Interop.Sys.GetLastErrorInfo()));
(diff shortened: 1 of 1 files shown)


src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
```diff
@@ -418,10 +418,7 @@ protected override void Dispose(bool disposing)
                         // Since we still have the file open, this will end up deleting
                         // it (assuming we're the only link to it) once it's closed, but the
                         // name will be removed immediatly.
```
sokket: NIT: immediately

sokket:
Typo from a different commit right near yours...feel free to ignore if you'd like. Otherwise, LGTM

src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
jasonwilliams200OK: :bulb: if spell checker extension is installed, https://visualstudiogallery.msdn.microsoft.com/7c8341f1-ebac-40c8-92c2-476db8d523ce, it would probably highlight all the typos. Then the committer can fix typos for only the touched files in the commit / change-set. Soon enough, there won't be any typo left in the entire source tree.

PR #3449, opened 2015-09-25
Validate hostname when using SslStream as a client

bartonjs:
Added a method in the crypto shim (CheckX509Hostname) which does the heavy lifting. It may end up making sense to push this up into managed for better flexibility when trying to add support for IDNA; but that would be greatly assisted by having good managed decoders for the subject name pieces and the SubjectAltNames extension.

Verified that it works against an explicit SAN entry and a wildcard SAN entry.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
@@ -111,6 +111,9 @@ internal static partial class Interop
             int second,
             [MarshalAs(UnmanagedType.Bool)] bool isDst);
 
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern int CheckX509Hostname(SafeX509Handle x509, string hostname, int cchHostname);
+
         internal static byte[] GetAsn1StringBytes(IntPtr asn1)
         {
             return GetDynamicBuffer((ptr, buf, i) => GetAsn1StringBytes(ptr, buf, i), asn1);
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -719,6 +719,225 @@ GetX509NameInfo(
 
 /*
 Function:
+CheckX509HostnameMatch
+
+Checks if a particular ASN1_STRING represents the entry in a certificate which would match against
+the requested hostname.
+
+Prameter sanRules: 0 for match rules against the subject CN, 1 for match rules against a SAN entry
+
+Return values:
+1 if the hostname is a match
+0 if the hostname is not a match
+Any negative number indicates an error in the arguments.
+*/
+static
+int
+CheckX509HostnameMatch(ASN1_STRING* candidate, const char* hostname, int cchHostname, char sanRules)
+{
+    assert(candidate);
+    assert(hostname);
+
+    if (!candidate->data || !candidate->length)
+    {
+        return 0;
+    }
+
+    // If the candidate is *.example.org then the smallest we would match is a.example.org, which is the same
+    // length. So anything longer than what we're matching against isn't valid.
+
+    // This assumption might not hold under IDNA, it might need to be more sophisticated
+    // "UTF character string length"
+    if (candidate->length > cchHostname)
+    {
+        return 0;
+    }
+
+    if (sanRules)
+    {
+        // RFC2818 says to use RFC2595 matching rules, but then gives an example that f*.com would match foo.com
+        // RFC2595 says that '*' may be used as the left name component, in which case it is a wildcard that does
+        // not match a '.'.
+        //
+        // In the interest of time, and the idea that it's better to err on the side of more restrictive,
+        // this implementation does not support mid-string wildcards.
+        //
+        // TODO (3444): Determine if we're too restrictive here.
+
+        char* candidateStr;
+        int i;
+        int hostnameFirstDot = -1;
+
+        // A GEN_DNS name is supposed to be a V_ASN1_IA5STRING.  If it isn't, we don't know how to read it.
+        if (candidate->type != V_ASN1_IA5STRING)
+        {
+            return 0;
+        }
+
+        // Great, candidateStr is just candidate->data!
+        candidateStr = (char*)candidate->data;
+
+        // First, verify that the string is alphanumeric, plus hyphens or periods and maybe starting with an asterisk.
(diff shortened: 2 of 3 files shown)


bartonjs:
cc @stephentoub @CIPop, @DavidSh, @JosGuil @SidharthNabar, @PGavlin, @rajansingh10, @vijaykota

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -20,7 +26,23 @@ internal override SslPolicyErrors VerifyCertificateProperties(X509Chain chain, X
                 sslPolicyErrors |= SslPolicyErrors.RemoteCertificateChainErrors;
```
vijaykota: This is actually related to https://github.com/dotnet/corefx/pull/3432. Windows code throws a CryptoException if Build fails. Unix code should behave the same way

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -719,6 +719,224 @@ GetX509NameInfo(
...
+        // First, verify that the string is alphanumeric, plus hyphens or periods and maybe starting with an asterisk.
+        for (i = 0; i < candidate->length; ++i)
+        {
+            char c = candidateStr[i];
+
+            if ((c < 'a' || c > 'z') &&
+                (c < '0' || c > '9') &&
+                (c != '.') &&
+                (c != '-') &&
+                (c != '*' || i != 0))
```
nil4: Shouldn't this be `(c != '*' && i != 0)`?

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
```diff
@@ -111,6 +111,12 @@ internal static partial class Crypto
             int second,
             [MarshalAs(UnmanagedType.Bool)] bool isDst);
 
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern int CheckX509Hostname(SafeX509Handle x509, string hostname, int cchHostname);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern int CheckX509Hostname(IntPtr x509, string hostname, int cchHostname);
```
stephentoub: Why do we need both the SafeHandle and IntPtr versions?

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -20,7 +26,23 @@ internal override SslPolicyErrors VerifyCertificateProperties(X509Chain chain, X
...
-            // TODO (3431): Ensure that the certificate is valid for the hostName context.
+            if (checkCertName)
+            {
+                if (string.IsNullOrEmpty(hostName))
+                {
+                    sslPolicyErrors |= SslPolicyErrors.RemoteCertificateNameMismatch;
+                }
+                else
+                {
+                    int hostnameMatch = Interop.Crypto.CheckX509Hostname(certificate.Handle, hostName, hostName.Length);
```
stephentoub: Do we need to do some kind of validation that Handle isn't invalid?

src/Native/System.Security.Cryptography.Native/openssl.c
stephentoub: I believe it's correct as is.  After checking the rest of the characters and finding it's thus far invalid, we check to see if it's a '*'.  If it's not, then the condition ```c != '*'``` is satisfied and we return 0 correctly as an error.  If it is a '*', then ```c != '*'``` is false and we proceed to check ```i != 0```... if it's the first character, then ```i != 0``` is false and we don't error out; if it's not the first character, then ```i != 0``` is true and we return 0 correctly.

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -719,6 +719,224 @@ GetX509NameInfo(
...
+        {
+            // It's possible that this should be considered a match if the entire SAN entry is '*',
+            // aka candidate->length == 1; but nothing talks about this case.
+            return 0;
+        }
+
+        {
+            // If hostname is 10 bytes long, and the '.' was at index 3, then we eliminate the first 3
+            // characters (www) from the match constraint.  This forces the wildcard to be the last
+            // character before the . in its match group.
```
stephentoub: I don't see us checking or verifying these numbers (3 and 10)... do we not need to?

stephentoub:
A couple of minor comments, but LGTM.

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: That comment was mainly me explaining to myself that there wasn't an off-by-one error in the math.  If it's adding confusion I can delete it (or some rewording suggestions would be welcome).

stephentoub: Maybe just prefice it with "Determine how many characters there are after the wildcard.  For example, "

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
bartonjs: The Windows code returns if the chain fails AND the handle is invalid.  I feel like that if is "if Build returns true there'll be a handle; otherwise I guess we should check...".

Unix has no theoretical-equivalent state.  The handle is always representing null; and Build true means happy, Build false means validation errors.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
bartonjs: I've now added the call to X509_dup to promote the IntPtr to a SafeHandle, and added the using; so deleted the IntPtr version here.

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: Done :)

src/Native/System.Security.Cryptography.Native/openssl.c
nil4: You are right, thank you @stephentoub.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
vijaykota: I see... so Windows has an "intermediate" state of Build failing with a valid handle.

PR #3451, opened 2015-09-25
Initialize OpenSSL RAND with an unpredictable seed.

eerhardt:
When we initialize OpenSSL, we are currently calling RAND_poll(), which isn't guaranteed to always seed the PRNG with purely random numbers.

To fix this, OpenSSL will also be initialized with a random 32-byte seed from PAL_Random.  PAL_Random will be modified to block until it can ensure a purely random number, and if it can't it returns false.  If PAL_Random returns false while initializing OpenSSL, an exception is thrown.

@bartonjs, @morganbr, @ellismg, @stephentoub

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Initialization.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Initialization.cs
@@ -31,7 +31,15 @@ internal static partial class Interop
     {
         static CryptoInitializer()
         {
-            if (EnsureOpenSslInitialized() != 0)
+            byte[] randomSeed = new byte[32];
+            if (!Random(true, randomSeed, randomSeed.Length))
+            {
+                // unable to get a reliable random seed, so we cannot
+                // safely assume crypto is going to work securely.
+                throw new InvalidOperationException();
+            }
+
+            if (EnsureOpenSslInitialized(randomSeed, randomSeed.Length) != 0)
             {
                 // Ideally this would be a CryptographicException, but we use
                 // OpenSSL in libraries lower than System.Security.Cryptography.
@@ -56,7 +64,13 @@ internal static partial class Interop
         }
 
         [DllImport(Libraries.CryptoNative)]
-        private static extern int EnsureOpenSslInitialized();
+        private static extern int EnsureOpenSslInitialized(byte[] randomSeed, int randomSeedLength);
+
+        [DllImport(Libraries.LibCoreClr, EntryPoint = "PAL_Random")]
+        private extern static bool Random(
+            bool bStrong,
+            [Out, MarshalAs(UnmanagedType.LPArray)] byte[] buffer, 
+            int length);
 
         [DllImport(Libraries.LibCrypto)]
         private static extern void ERR_load_crypto_strings();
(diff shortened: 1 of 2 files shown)


stephentoub:
Why do we believe that RAND_poll is getting its entropy from somewhere other than the same place /dev/random is getting it?

Why are we not concerned with the /dev/random blocking behavior? Or worse that if it can't get such entropy the entire system will fail?

And why are we ok introducing another dependency to libcoreclr? We've been actively trying to get rid of them all.

I'm not a cryptography, so please take this with a grain of salt. But to me this appears to be extra work that's not really improving security and could actually harm the behavior of the system.

eerhardt:
>Why do we believe that RAND_poll is getting its entropy from somewhere other than the same place /dev/random is getting it?

If /dev/urandom is available, RAND_poll doesn't ever read from /dev/random.  See the section "OpenSSL Concerns" in https://www.blackhat.com/docs/us-15/materials/us-15-Potter-Understanding-And-Managing-Entropy-Usage-wp.pdf.  Basically RAND_poll doesn't block or fail if there isn't enough entropy in the system, which is what we are trying to address.

>Why are we not concerned with the /dev/random blocking behavior? Or worse that if it can't get such entropy the entire system will fail?

We are concerned, which is why we only read /dev/random once, and use /dev/urandom for all other reads.  But blocking and/or failing seems to be the lesser of two evils when compared to potentially having a security threat.

>And why are we ok introducing another dependency to libcoreclr?

I wasn't aware that we are trying to remove dependencies on libcoreclr. The reason I made it is to have a single place that does this "entropy initialization".  @ellismg needs a cryptographically secure random number for his work on string hashing, which is in coreclr. So I felt PAL_Random was the best place to put these rules. I can change it, if we are OK with duplicating it.

A thing I forgot to mention here and in the [PAL_Random PR](https://github.com/dotnet/coreclr/pull/1623) is that we have further plans to enhance PAL_Random to check for  [getrandom()](http://man7.org/linux/man-pages/man2/getrandom.2.html), which is only available in kernel 3.17 and later. If getrandom() is available, we will use that instead. But if getrandom() isn't available, then we will use the behavior that is being enhanced here.

stephentoub:
> If /dev/urandom is available, RAND_poll doesn't ever read from /dev/random

I understand it doesn't read from /dev/random, I was questioning where it gets its source of entropy.... both /dev/random and /dev/urandom pull from the same place.

> Basically RAND_poll doesn't block or fail if there isn't enough entropy in the system, which is what we are trying to address.

What does BCryptGenRandom on Windows do?  Does it ever block?

> which is why we only read /dev/random once

As I noted in the thread on coreclr, I don't believe that's sufficient to avoid problems.

> if we are OK with duplicating it

Increasing the number of places we try to read from /dev/random will only make the blocking potential worse.

> But blocking and/or failing seems to be the lesser of two evils 

I'm not convinced of that.  The paper you cited suggests that it could take several seconds for enough entropy to enter the system for /dev/random to return even a single byte.  That means potentially increasing startup time for even a simple app by several seconds.

eerhardt:
After some internal discussions, we've decided not to take these changes for Beta8.  I've opened #3474 to track this for the future.

PR #3452, opened 2015-09-25
Correct test namespaces for System.Collections.*

Clockwork-Muse:
Correct test namespaces for some System.Collections.* projects, per #2898 .
Only those assemblies listed here were checked.

Namespaces were changed, and `using`s were cleaned up in updated files.
Some classes were renamed to avoid conflicts.

### DIFF
--- a/src/System.Collections.Concurrent/tests/BlockingCollectionCancellationTests.cs
+++ b/src/System.Collections.Concurrent/tests/BlockingCollectionCancellationTests.cs
@@ -1,14 +1,12 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
-namespace Test
+namespace System.Collections.Concurrent.Tests
 {
     public class BlockingCollectionCancellationTests
     {
--- a/src/System.Collections.Immutable/tests/BadHasher.cs
+++ b/src/System.Collections.Immutable/tests/BadHasher.cs
@@ -3,7 +3,7 @@
 
 using System.Collections.Generic;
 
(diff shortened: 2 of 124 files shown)


stephentoub:
Per the discussion on that issue, I thought e we're going to hold off on adding namespaces where we didn't previously have them to avoid all of these extraneous whitespace changes?

Clockwork-Muse:
@stephentoub - argh, I keep forgetting about that.  I think I ended up doing most of mine because "well, it's got to be done at some point", to potentially contribute to that final PR.

What about removal, where it was previously nested?  In the case where an assembly has both changes and additions, should I revert the additions, but keep the changes?

Clockwork-Muse:
@stephentoub - hmm, I guess, given System.Threading, just leave additions, but do the updates.  What about removals?

I'll audit the PRs I've submitted - I think most have at least some updates.  `System.Collections` might not respond well, though.

stephentoub:
Re: "just leave additions, but do the updates"

Yes, thanks. Otherwise this is going to end up impacting _a ton_ of code, even if it's just trivially via indentation. We should do the trivial changes now, and then at some point when we're less concerned about churn, do the native namespace additions in one fell swoop.

Re: "What about removals?"

Let's err on the side of not removing them, for the same reason. But if there are isolated cases that won't impact a lot of code, it's ok.

Clockwork-Muse:
@stephentoub - Done (and for all other PRs, too).

I had to pull System.Collections completely - there's just way too many things going on with conflicts and namespaces in that one to only do the updates.  I did pull a copy of all changes, so it can be used for the later PR too.

I left some instances of moving a (helper) type inside an existing namespace in System.Collections.NonGeneric.  The namespace was already inside the file, and was being updated.  Most of them are small (~5 lines-ish), although a couple are larger.  Was that okay, or should I pull it?

stephentoub:
Thanks, @Clockwork-Muse.  I think what you did is a reasonable compromise.  Thanks for cleaning up all of these namespaces.

LGTM.

PR #3457, opened 2015-09-25
Correct test namespaces for System.ObjectModel

Clockwork-Muse:
Correct test namespaces for System.ObjectModel, per #2898 .

Only namespaces were adjusted, usings cleaned up, no other work was performed.


stephentoub:
LGTM

PR #3369, opened 2015-09-22
Incorporated some comments - Net Secure Xplat Implementation

rajansingh10:
Incorporated some of the comments.
Sending it as a new PR as original PR will be closed, once we merge into dev_sslstream to unblock partners.
Tracking all issues via #3362 . (I am yet to updated the issue with complete list).
Updated all TODO's in the code to refer this issue.
This also includes fixes from Vijay's dummy PR https://github.com/dotnet/corefx/pull/3365.

PR #3463, opened 2015-09-25
Enable SslStream for Unix

bartonjs:
Merging the current status of dev-sslstream into master.

* The Windows version should continue to work just fine.
* The Unix version has received sanity testing, and has limited functionality, but is believed to be in a "basic scenarios, for integration purposes" state.
    * `stream.AuthenticateAsClient(string)` :white_check_mark: 
    * `stream.AuthenticateAsClientAsync(string)` :white_check_mark:
    * `stream.AuthenticateAsServer(X509Certificate)` :ok: (requires the provided instance be X509Certificate2, and using RSA)
    * `stream.Write(byte[])` :white_check_mark:
    * `stream.Read(byte[], int, int)` :white_check_mark:
    * {Read|Write}Async :grey_question: (it's expected to work, but wasn't explicitly verified)
    * Off the golden path may be a bit rough; unexpected disconnects and the like may have different exception flows than their Windows counterparts (if encountered, please report!)
    * Mutual authentication / client authentication certificates :x:


bartonjs:
cc @stephentoub @CIPop, @DavidSh, @JosGuil @SidharthNabar, @PGavlin, @rajansingh10, @vijaykota

stephentoub:
How does this relate to https://github.com/dotnet/corefx/pull/3369? Are those changes invalidated by this?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
```
CIPop: Nit: Doesn't seem to belong to Interop : https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/interop-guidelines.md

> With few exceptions, the only methods that should be defined in these interop types are DllImports.
> * Exceptions are limited to times when most or every consumer of a particular DllImport will need to wrap its invocation in a helper, e.g. to provide additional marshaling support, to hide thread-safety issues in the underlying OS implementation, to do any required manipulation of safe handles, etc. In such cases, the DllImport should be private whenever possible rather than internal, with the helper code exposed to consumers rather than having the DllImport exposed directly.

For other System.Net implementations we've placed this kind of code in the PAL layer.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+            internal IntPtr readBioPtr;
+            internal IntPtr writeBioPtr;
+            internal bool isServer;          
+        }
+        #endregion
+
+        #region internal methods
+
+        //TODO Set remote certificate options
+        internal static IntPtr AllocateSslContext(long options, SafeX509Handle certHandle, SafeEvpPKeyHandle certKeyHandle, bool isServer, bool remoteCertRequired)
```
CIPop: Should probably use a kind of `SafeHandle` instead of `IntPtr`

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -0,0 +1,140 @@
...
+using System;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
+{
+    internal static partial class libssl
+    {
+
```
CIPop: Nit: extra new lines

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -0,0 +1,140 @@
...
+        [DllImport(Interop.Libraries.LibSsl, EntryPoint = "SSL_CTX_check_private_key")]
+        internal static extern int SSL_CTX_check_private_key(IntPtr ssl);
+
+        [DllImport(Interop.Libraries.LibSsl, EntryPoint = "BIO_ctrl_pending")]
+        internal static extern int BIO_ctrl_pending(IntPtr bio);
+
+        [DllImport(Interop.Libraries.LibSsl, EntryPoint = "SSL_CTX_set_quiet_shutdown")]
+        internal static extern void SSL_CTX_set_quiet_shutdown(IntPtr ssl, int mode);
+    }
+}
```
CIPop: Nit: no new line at EOF

src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -719,6 +719,225 @@ GetX509NameInfo(
...
+
+        // Great, candidateStr is just candidate->data!
+        candidateStr = (char*)candidate->data;
+
+        // First, verify that the string is alphanumeric, plus hyphens or periods and maybe starting with an asterisk.
+        for (i = 0; i < candidate->length; ++i)
+        {
+            char c = candidateStr[i];
+
+            if ((c < 'a' || c > 'z') &&
```
CIPop: `isalnum()` instead?

CIPop: Ah, I see the * check now.

src/System.Net.Security/src/System.Net.Security.csproj
```diff
@@ -23,10 +23,13 @@
     <Compile Include="System\Net\_SecurityBuffer.cs" />
     <Compile Include="System\Net\_SecurityBufferType.cs" />
     <Compile Include="System\Net\_SecureChannel.cs" />
+	<Compile Include="System\Net\SSPIInterface.cs" />
```
CIPop: Nit: spacing seems off.

src/System.Net.Security/src/System.Net.Security.csproj
```diff
@@ -122,8 +122,22 @@
...
 
-    <!-- Interop -->
-    <Compile Include="$(CommonPath)\Interop\Windows\Interop.Libraries.cs">
+   
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetsWindows)' == 'true' ">  
+	<Compile Include="System\Net\SecurityContextTokenHandle.cs" />
+	<Compile Include="System\Net\Windows\CertModule.cs" />
+	<Compile Include="System\Net\Windows\NegotiationInfo.cs" />
```
CIPop: Nit: Feel free to comment NegotiateStream parts out for now.

src/System.Net.Security/src/System/Net/CertInterface.cs
```diff
@@ -0,0 +1,21 @@
...
+    internal abstract class CertInterface
+    {
+        internal abstract  X509Certificate2 GetRemoteCertificate(SafeDeleteContext securityContext, out X509Certificate2Collection remoteCertificateStore);
+
+        internal abstract SslPolicyErrors VerifyCertificateProperties(X509Chain chain, X509Certificate2 certificate, bool checkCertName, bool isServer, string hostName);
+
+        internal abstract string[] GetRequestCertificateAuthorities(SafeDeleteContext securityContext);
+
+        internal abstract X509Store EnsureStoreOpened(bool isMachineStore);
+
```
CIPop: Nit: extra new line, missing new line at EOF.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+            sslContext.isServer = isServer;
+
+            try
+            {
+                IntPtr method = GetSslMethod(isServer, options);
+
+                var contextPtr = libssl.SSL_CTX_new(method);
+                if (IntPtr.Zero == contextPtr)
+                {
+                    throw CreateSslException("Failed to allocate SSL/TLS context");
```
CIPop: /cc: @stephentoub 
Please add an issue to convert all these strings to localizable resources. (see `SR.` usage throughout the Windows code).

src/System.Net.Security/src/System/Net/CertInterface.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Win32.SafeHandles;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Net
+{
+    internal abstract class CertInterface
```
CIPop: NIT: Is this part of the PAL? Please rename to get the code uniform - see Primitives, NameResolution and Sockets for a model.
I would recommend `CertInterfacePAL` instead.

src/System.Net.Security/src/System/Net/CertWrapper.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Win32.SafeHandles;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Net
+{
+    internal static class CertWrapper
```
CIPop: NIT: same as above - please rename to `CertWrapperPAL`.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Net.Security;
+using System.Runtime.InteropServices;
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Authentication;
+
+namespace System.Net
+{
+    // Used to define the interface for security to use.
+    internal interface SSPIInterface
```
CIPop: Nit: this is really a TLS interface only - has hard dependencies on X509, etc. Also needs to be marked as PAL.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
```diff
@@ -117,25 +117,17 @@ internal void ValidateCreateContext(bool isServer, string targetHost, SslProtoco
...
-                enabledSslProtocols &= (SslProtocols)Interop.SChannel.ClientProtocolMask;
-            }
+            }            
 
             if ((int)enabledSslProtocols == 0)
             {
                 throw new ArgumentException(SR.Format(SR.net_invalid_enum, "SslProtocolType"), "sslProtocolType");
             }
+         
+            if (isServer && serverCertificate == null)
```
CIPop: Behavior breaking: please move back up to preserve order of checks.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
```diff
@@ -117,25 +117,17 @@ internal void ValidateCreateContext(bool isServer, string targetHost, SslProtoco
             if (targetHost == null)
             {
                 throw new ArgumentNullException("targetHost");
-            }
-            
-            if (isServer)
-            {
-                enabledSslProtocols &= (SslProtocols)Interop.SChannel.ServerProtocolMask;
```
CIPop: Breaking: where did this code end up?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
```diff
@@ -117,25 +117,17 @@ internal void ValidateCreateContext(bool isServer, string targetHost, SslProtoco
...
-                if (serverCertificate == null)
-                {
-                    throw new ArgumentNullException("serverCertificate");
-                }
-            }
-            else
-            {
-                enabledSslProtocols &= (SslProtocols)Interop.SChannel.ClientProtocolMask;
```
CIPop: Breaking: where did this code end up?

src/Common/src/Interop/Unix/Interop.Libraries.cs
```diff
@@ -6,13 +6,15 @@ internal static partial class Interop
     private static partial class Libraries
     {
         internal const string Libc = "libc";                   // C library
-        internal const string LibCoreClr= "libcoreclr";        // CoreCLR runtime
+        internal const string LibCoreClr = "libcoreclr";        // CoreCLR runtime
```
stephentoub: Nit: alignment of comments is off

src/Common/src/Interop/Unix/Interop.Libraries.cs
```diff
@@ -6,13 +6,15 @@ internal static partial class Interop
...
         internal const string LibCrypto = "libcrypto";         // OpenSSL crypto library
+        internal const string LibSsl = "libssl";               // OpenSSL library
         internal const string Zlib = "libz";                   // zlib compression library
 
         // Shims
         internal const string SystemNative = "System.Native";
-        internal const string HttpNative   = "System.Net.Http.Native";
+        internal const string HttpNative = "System.Net.Http.Native";
+        internal const string SslNative = "System.Net.Security.Native";
```
stephentoub: Do we need a separate native library from System.Security.Cryptography.Native for this?

src/Common/src/Interop/Unix/System.Net.Security.Native/Interop.Initialization.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        internal static void Initialize()
+        {
+            // No-op that exists to provide a hook for other static constructors
+        }
+
+        [DllImport(Libraries.LibSsl, SetLastError = true, EntryPoint = "SSL_load_error_strings")]
+        internal static extern void SSL_load_error_strings();
+
+        [DllImport(Libraries.LibSsl, SetLastError = true, EntryPoint = "SSL_library_init")]
```
stephentoub: Why are the entry point names repeated when there's no change?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+        #region structures
+        [StructLayout(LayoutKind.Sequential)]
+        private struct SslContext
+        {         
+            internal IntPtr sslPtr;
+            internal IntPtr readBioPtr;
+            internal IntPtr writeBioPtr;
+            internal bool isServer;          
+        }
+        #endregion
```
stephentoub: I'll reiterate the comments I made in the other PR: in summary, I don't understand why we're using a struct for this.  We're allocating native memory to store it, and then any time we need the data, we're having to marshal the data back to managed.  We should instead just use a class, using a GCHandle to keep it alive and have a pointer we can pass around in native.  This will also help with ensuring we don't accidentally double-free the pointers contained, as we can null them out after we've free'd them.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+
+        //TODO Set remote certificate options
+        internal static IntPtr AllocateSslContext(long options, SafeX509Handle certHandle, SafeEvpPKeyHandle certKeyHandle, bool isServer, bool remoteCertRequired)
+        {
+            IntPtr sslContextPtr = Marshal.AllocHGlobal(Marshal.SizeOf<SslContext>());
+            SslContext sslContext = new SslContext
+            {              
+                sslPtr = IntPtr.Zero,
+                readBioPtr = IntPtr.Zero,
+                writeBioPtr = IntPtr.Zero,
```
stephentoub: These are already IntPtr.Zero and don't need to be initialized to it.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+                if (IntPtr.Zero == contextPtr)
+                {
+                    throw CreateSslException("Failed to allocate SSL/TLS context");
+                }
+
+                libssl.SSL_CTX_ctrl(contextPtr, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
+
+                libssl.SSL_CTX_set_quiet_shutdown(contextPtr, 1);
+
+                if (certHandle != null && certKeyHandle != null)
```
stephentoub: If these are non-null, are they also !IsInvalid?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+                {
+                    SetSslCertificate(contextPtr, certHandle, certKeyHandle);
+                }
+
+                sslContext.sslPtr = libssl.SSL_new(contextPtr);
+
+                libssl.SSL_CTX_free(contextPtr);
```
stephentoub: It's ok to do this because... SSL_new incremented a ref count on contextPtr?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+
+                libssl.SSL_CTX_free(contextPtr);     
+
+                if (IntPtr.Zero == sslContext.sslPtr)
+                {
+                    throw CreateSslException("Failed to create SSSL object from SSL context");
+                }      
+
+                IntPtr memMethodRead = libcrypto.BIO_s_mem();
+                IntPtr memMethodWrite = libcrypto.BIO_s_mem();
```
stephentoub: We shouldn't need to request this twice.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+                if ((IntPtr.Zero == memMethodWrite) || (IntPtr.Zero == memMethodRead))
+                {
+                    throw CreateSslException("Failed to return memory BIO method function");
+                }
+
+                sslContext.readBioPtr = libssl.BIO_new(memMethodRead);
+                sslContext.writeBioPtr = libssl.BIO_new(memMethodWrite);
+                if ((IntPtr.Zero == sslContext.readBioPtr) || (IntPtr.Zero == sslContext.writeBioPtr))
+                {
+                    throw CreateSslException("Failed to retun new BIO for a given method type");
```
stephentoub: If readBioPtr != null but writeBioPtr == null, we end up leaking readBioPtr.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+                    libssl.SSL_set_accept_state(sslContext.sslPtr);
+                }
+                else
+                {
+                    libssl.SSL_set_connect_state(sslContext.sslPtr);
+                }
+
+                libssl.SSL_set_bio(sslContext.sslPtr, sslContext.readBioPtr, sslContext.writeBioPtr);
+            }
+            catch
```
stephentoub: We shouldn't be eating the exception here.

src/System.Net.Security/src/System/Net/_SslSessionsCache.cs
```diff
@@ -51,8 +53,10 @@ internal SslCredKey(byte[] thumbPrint, int allowedProtocols, EncryptionPolicy en
 
                 _HashCode ^= allowedProtocols;
                 _HashCode ^= (int)encryptionPolicy;
-                _AllowedProtocols = allowedProtocols;
+                _HashCode ^= serverMode ? 5 : 7; //TODO used a prime number here as it's a XOR. Figure out appropriate value.
```
CIPop: Potentially breaking: why did the cert comparison hash change from the original algorithm shipped in .Net Desktop?

src/System.Net.Security/src/System/Net/_SslSessionsCache.cs
```diff
@@ -51,8 +53,10 @@ internal SslCredKey(byte[] thumbPrint, int allowedProtocols, EncryptionPolicy en
...
+                _AllowedProtocols = allowedProtocols;
```
CIPop: Breaking: the allowed protocols used to include the Client/Server flags. Now it doesn't.

src/System.Net.Security/src/System/Net/_SslSessionsCache.cs
```diff
@@ -18,6 +19,7 @@ private struct SslCredKey
         {
             private byte[] _CertThumbPrint;
             private int _AllowedProtocols;
+            private bool isServerMode;
```
CIPop: Nit: `_isServerMode`

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -0,0 +1,408 @@
...
+
+        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int count, int bufferCapacity)
+        {
+            SslContext context = Marshal.PtrToStructure<SslContext>(handlePtr);  
+
+            var retVal = libssl.SSL_write(context.sslPtr, new IntPtr(buffer.ToInt64() + offset), count);
+
+            if (retVal != count)
+            {
+                int error = GetSslError(context.sslPtr, retVal, "");
```
stephentoub: Maybe I should stop commenting on things previously commented on in that other PR?  GetSslError doesn't need to take a string message.

src/System.Net.Security/src/System/Net/_SslSessionsCache.cs
```diff
@@ -96,6 +100,11 @@ public override bool Equals(Object y)
                     return false;
                 }
 
+                if (isServerMode != other.isServerMode)
```
CIPop: Now I understand how this works but I don't understand the reason for this getting changed from the original implementation.
The object's now bigger and requires extra operations to determine equality.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+namespace System.Net
+{
+    // For SSL connections:
+    internal class SSPISecureChannelType : SSPIInterface
+    {
+        public Exception GetException(SecurityStatus status)
+        {
+            // TODO: To be implemented
+            throw new Exception("status = " + status);
```
stephentoub: return instead of throw?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                count = resultSize;
+            }
+            return retVal;
+        }
+
+        public int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute,
+            out SafeFreeContextBufferChannelBinding refHandle)
+        {
+            throw NotImplemented.ByDesignWithMessage(SR.net_MethodNotImplementedException);
+        }
```
stephentoub: Other places that have not-yet-implemented portions have a TODO comment.

src/System.Net.Security/src/System/Net/CertWrapper.cs
CIPop: I don't really understand the purpose of this indirection. What is the design intent?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        public int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
+        {
+            bool gotReference = false;
+            connectionInfo = null;
+            try
+            {
+                securityContext.DangerousAddRef(ref gotReference);
+                Interop.libssl.SSL_CIPHER cipher = Interop.OpenSsl.GetConnectionInfo(securityContext.DangerousGetHandle());
```
stephentoub: GetConnectionInfo can't just take the SafeHandle rather than doing this AddRef/GetHandle/Release stuff in the caller?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+
+                if (null == inputBuffer)
+                {
+                    done = Interop.OpenSsl.DoSslHandshake(context.DangerousGetHandle(), inputPtr, 0, out outputPtr, out outputSize);
+                }
+                else
+                {
+                    inputHandle = GCHandle.Alloc(inputBuffer.token, GCHandleType.Pinned);
+                    inputPtr = Marshal.UnsafeAddrOfPinnedArrayElement(inputBuffer.token, inputBuffer.offset);
+                    done = Interop.OpenSsl.DoSslHandshake(context.DangerousGetHandle(), inputPtr, inputBuffer.size, out outputPtr, out outputSize);
```
stephentoub: Can DoSslHandshake just take the SafeHandle?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                outputBuffer.offset = 0;
+                if (outputSize > 0)
+                {
+                    outputBuffer.token = new byte[outputBuffer.size];
+                    Marshal.Copy(outputPtr, outputBuffer.token, 0, outputBuffer.size);
+                    Marshal.FreeHGlobal(outputPtr);
+                }
+                else
+                {
+                    outputBuffer.token = null;
```
stephentoub: Assert that outputPtr is IntPtr.Zero?  If it's not we're leaking.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                    inputPtr = Marshal.UnsafeAddrOfPinnedArrayElement(inputBuffer.token, inputBuffer.offset);
+                    done = Interop.OpenSsl.DoSslHandshake(context.DangerousGetHandle(), inputPtr, inputBuffer.size, out outputPtr, out outputSize);
+                }           
+
+                outputBuffer.size = outputSize;
+                outputBuffer.offset = 0;
+                if (outputSize > 0)
+                {
+                    outputBuffer.token = new byte[outputBuffer.size];
+                    Marshal.Copy(outputPtr, outputBuffer.token, 0, outputBuffer.size);
```
stephentoub: We're copying _from_ the _output_Ptr?

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+                securityContext.DangerousAddRef(ref gotReference);
+                inputHandle = GCHandle.Alloc(buffer, GCHandleType.Pinned);
+                IntPtr inputPtr = Marshal.UnsafeAddrOfPinnedArrayElement(buffer, 0);
+                if (encrypt)
+                {
+                    resultSize = Interop.OpenSsl.Encrypt(securityContext.DangerousGetHandle(), inputPtr, offset, size, buffer.Length);
+                }
+                else
+                {
+                    resultSize = Interop.OpenSsl.Decrypt(securityContext.DangerousGetHandle(), inputPtr, size);
```
stephentoub: I'll stop asking about these for each of the methods, but it'd be great if we could pass the SafeHandle down as close to the native calls as possible.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,227 @@
...
+            bool ignore = false;
+            target.DangerousAddRef(ref ignore);
+            _Target = target;
+            SetHandle(new IntPtr(0));   // make this handle valid
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            SafeFreeCredentials target = _Target;
+            if (target != null)
```
stephentoub: In what situation would it be null?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -0,0 +1,227 @@
...
+            Interop.OpenSsl.FreeSslContext(handle);
+            Debug.Assert((null != _credential) && !_credential.IsInvalid, "Invalid credential saved in SafeDeleteContext");
+            _credential.DangerousRelease();
+            return true;
+        }
+
+        public override string ToString()
+        {
+            return IsInvalid ? String.Empty : handle.ToString();
+        }
```
stephentoub: Is this just for debugging purposes?

src/System.Net.Security/src/System/Net/Windows/CertModule.cs
```diff
@@ -0,0 +1,270 @@
...
+            Interop.Secur32.IssuerListInfoEx issuerList = (Interop.Secur32.IssuerListInfoEx)outObj;
+
+            try
+            {
+                if (issuerList.cIssuers > 0)
+                {
+                    unsafe
+                    {
+                        uint count = issuerList.cIssuers;
+                        issuers = new string[issuerList.cIssuers];
```
stephentoub: count instead of issuerList.cIssuers?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
bartonjs: It looks like it's in /src/System.Net.Security/src/System/Net/Windows/SSPISecureChannelType.cs `CreateSecureCredential`

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
bartonjs: Ooh, ooh, I know this one :).  The constructor that assigned them checked them for a valid response; and SetSslCertificate also asserts that they're valid.

src/System.Net.Security/src/System/Net/Windows/SSPIAuthType.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Collections;
+using System.Diagnostics;
+using Microsoft.Win32.SafeHandles;
+using System.Security.Authentication;
+
+namespace System.Net
+{ 
+
+    // For Authentication (Kerberos, NTLM, Negotiate and WDigest):
+    internal class SSPIAuthType: SSPIInterfaceNego
```
CIPop: This code has changed from the original, oddly enough, without any apparent reason.
I diffed by extracting to another file and diffing that way.

src/System.Net.Security/src/System/Net/Windows/SSPIAuthType.cs
```diff
@@ -0,0 +1,213 @@
...
+                bool ignore = false;
+                context.DangerousAddRef(ref ignore);
+                status = Interop.Secur32.DecryptMessage(ref context._handle, inputOutput, sequenceNumber, &qop);
+            }
+            finally
+            {
+                context.DangerousRelease();
+            }
+
+            const uint SECQOP_WRAP_NO_ENCRYPT = 0x80000001;
```
CIPop: Original code in master:          
`            if (status == 0 && qop == Interop.Secur32.SECQOP_WRAP_NO_ENCRYPT)`

I don't understand why this is changed. Maybe this was an older version, prior to my commit to master?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslState.cs
bartonjs: Fixed.

bartonjs:
@stephentoub Most of the comments from you appear to be the same as the ones that have yet to have merged resolution from #3345.  We discussed these the other day, and you said you were okay with merging today provided that we track and drive down the outstanding issues... is that still the case?

stephentoub:
> is that still the case?

Yeah, as long as:
- @CIPop is comfortable that Windows functionality isn't being regressed
- We'll address the outstanding issues soon after this all ends up in master

then LGTM.

CIPop:
@stephentoub: I can't say it looks good, but it's good progress. I can't also say that the Windows functionality (or even security) hasn't regressed just yet: I'm still reviewing.

@bartonjs: So far in the CR I see nothing of great concern. Given the complexity of changes and code that was moved around then changed (no diffs available from existing GitHub tooling) I will need more time to leave comments. Since this work will unblock partners and devs wanting to try out Security on non-Windows machines, let's merge for beta8 and continue work for beta9.

Until we make the full split from System.Private.Networking, existing developers working on ASP.Net on Windows are _not_ affected.

:shipit:

src/System.Net.Security/src/System/Net/CertInterface.cs
CIPop: Actually this is an interface and so it should be `ICertInterfacePAL`

src/System.Net.Security/src/System/Net/CertWrapper.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        static CertInterface certModule = new CertModule();
```
CIPop: Missing `private` access modifier.

src/System.Net.Security/src/System/Net/CertWrapper.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Win32.SafeHandles;
```
CIPop: Nit: `SafeHandles` using statement actually not required.

src/System.Net.Security/src/System/Net/GlobalSSPI.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    // Need a global so we can pass the interfaces as variables.
+    internal static class GlobalSSPI
+    {
+        internal static SSPIInterface SSPISecureChannel = new SSPISecureChannelType();      
+    }
+
```
CIPop: Nit: extra new line, missing new line at EOF.

CIPop: NIT: extra new line, no new line at EOF.

src/System.Net.Security/src/System/Net/GlobalSSPI.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    // Need a global so we can pass the interfaces as variables.
+    internal static class GlobalSSPI
+    {
+        internal static SSPIInterface SSPISecureChannel = new SSPISecureChannelType();
```
CIPop: Please either:
a. Leave the old code in.
b. Comment it out with a `TODO`:

`internal static SSPIInterface SSPISecureChannel = new SSPISecureChannelType();`

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+        SecurityStatus InitializeSecurityContext(SafeFreeCredentials credential, ref SafeDeleteContext context, string targetName, SecurityBuffer[] inputBuffers, SecurityBuffer outputBuffer);
+        SecurityStatus EncryptMessage(SafeDeleteContext securityContext, byte[] buffer, int size, int headerSize, int trailerSize, out int resultSize);
+        SecurityStatus DecryptMessage(SafeDeleteContext securityContext, byte[] buffer, ref int offset, ref int count);     
+        int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute, out SafeFreeContextBufferChannelBinding refHandle);
+        int QueryContextStreamSizes(SafeDeleteContext securityContext, out StreamSizes streamSizes);
+        int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo conectionInfo);
+        int QueryContextRemoteCertificate(SafeDeleteContext securityContext, out SafeFreeCertContext remoteCert);
+        int QueryContextIssuerList(SafeDeleteContext securityContext, out Object issuerList);
+    }   
+}
```
CIPop: NIT: missing new line at EOF.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+        SafeFreeCredentials AcquireCredentialsHandle(X509Certificate certificate, SslProtocols protocols, EncryptionPolicy policy, bool isServer);
+        SecurityStatus AcceptSecurityContext(ref SafeFreeCredentials credential, ref SafeDeleteContext context, SecurityBuffer inputBuffer, SecurityBuffer outputBuffer, bool remoteCertRequired);
+        SecurityStatus InitializeSecurityContext(ref SafeFreeCredentials credential, ref SafeDeleteContext context, string targetName, SecurityBuffer inputBuffer, SecurityBuffer outputBuffer);
+        SecurityStatus InitializeSecurityContext(SafeFreeCredentials credential, ref SafeDeleteContext context, string targetName, SecurityBuffer[] inputBuffers, SecurityBuffer outputBuffer);
+        SecurityStatus EncryptMessage(SafeDeleteContext securityContext, byte[] buffer, int size, int headerSize, int trailerSize, out int resultSize);
+        SecurityStatus DecryptMessage(SafeDeleteContext securityContext, byte[] buffer, ref int offset, ref int count);     
+        int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute, out SafeFreeContextBufferChannelBinding refHandle);
```
CIPop: Design: the definitions in this interface are not cohesive: some members are returning `SecurityStatus` while others are just returning an `int`.
Returning `int` is generally unacceptable for managed code.

This used to make sense since the original code was part of the Interop layer (the interface to the Native SSPI defined by Schannel).

src/System.Net.Security/src/System/Net/GlobalSSPI.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net
+{
+    // Need a global so we can pass the interfaces as variables.
+    internal static class GlobalSSPI
```
CIPop: Design: Given the changes, this class doesn't make sense now and it should be removed.

`GlobalSSPI` was the factory class for the two Windows SSPIs. Since SSPIInterface (PAL) is now TLS-specific.

Suggested plan:
  1. Push GlobalSSPI and the two SSPIInterface implementations based on Schannel down in the Windows-specific implementation. Regardless of changes at this layer, this abstraction still captures the essence of how Schannel SSPIs work.
  2. Redesign a cleaner PAL that would satisfy the needs for TLS
  3. (Future) design a new PAL for Nego
  4. Do a commonality/variability analysis between the two TLS and Nego PALs. If it makes sense anymore, bring the factory back.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,290 @@
...
+    // For SSL connections:
+    internal class SSPISecureChannelType : SSPIInterface
+    {
+        public Exception GetException(SecurityStatus status)
+        {
+            // TODO: To be implemented
+            throw new Exception("status = " + status);
+        }
+
+        public void VerifyPackageInfo()
```
CIPop: Is this supposed to be empty?

src/System.Net.Security/src/System/Net/Windows/SSPIAuthType.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Security.Authentication;
+
+namespace System.Net
+{ 
+
+    // For Authentication (Kerberos, NTLM, Negotiate and WDigest):
+    internal class SSPIAuthType: SSPIInterfaceNego
+    {
+
+        public void VerifyPackageInfo()
```
CIPop: I don't think this needs to be in the interface or this class.

src/System.Net.Security/src/System/Net/SSPIInterfaceNego.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Authentication;
+
+namespace System.Net
+{
+    // Used to define the interface for security to use.
+    internal interface SSPIInterfaceNego
+    {
+        void VerifyPackageInfo();
```
CIPop: Why is this part of the interface?

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Authentication;
+
+namespace System.Net
+{
+    // Used to define the interface for security to use.
+    internal interface SSPIInterface
+    {
+        void VerifyPackageInfo();
```
CIPop: Design: This looks like it's Windows specific: Schannel enumeration of available SSPIs and their properties.
Should not be part of this interface.

src/System.Net.Security/src/System/Net/Unix/SSPISecureChannelType.cs
CIPop: Why is this different from the original code anyway? AFAIK Unix code will still supposed to throw Win32Exception correctly setting the OS-level error code.

src/System.Net.Security/src/System/Net/SSPIInterface.cs
```diff
@@ -0,0 +1,31 @@
...
+        Exception GetException(SecurityStatus status);
```
CIPop: Design: I don't see the purpose for this `GetException` method. Shouldn't the code just throw Win32Exception setting the correct OS error level?

src/System.Net.Security/src/System/Net/SSPIInterfaceNego.cs
```diff
@@ -0,0 +1,39 @@
...
+        Exception GetException(SecurityStatus status);
```
CIPop: See above: `GetException()` could be replaced by throw new `Win32Exception(...)`.

src/System.Net.Security/src/System/Net/SSPIInterfaceNego.cs
```diff
@@ -0,0 +1,39 @@
...
+        int VerifySignature(SafeDeleteContext context, byte[] buffer, uint sequenceNumber);
+        int QueryContextChannelBinding(SafeDeleteContext phContext, ChannelBindingKind attribute, out SafeFreeContextBufferChannelBinding refHandle);
+        int QueryContextStreamSizes(SafeDeleteContext securityContext, out StreamSizes streamSizesContext);
+        int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo conectionInfoContext);
+        int QueryContextRemoteCertificate(SafeDeleteContext securityContext, out SafeFreeCertContext remoteCertContext);
+        int QueryContextIssuerList(SafeDeleteContext securityContext, out Object issuerListContext);
+        int QuerySecurityContextToken(SafeDeleteContext phContext, out SecurityContextTokenHandle phToken);
+        int CompleteAuthToken(ref SafeDeleteContext refContext, SecurityBuffer[] inputBuffers);
+    }   
+}
```
CIPop: NIT: no new line at EOF.

src/System.Net.Security/src/System/Net/Windows/SSPIAuthType.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+            finally
+            {
+                phContext.DangerousRelease();
+            }
+
+            return status;
+        }
+    }
+
```
CIPop: NIT: please remove extra new lines.

src/System.Net.Security/src/System/Net/Windows/SSPIAuthType.cs
```diff
@@ -0,0 +1,213 @@
...
+
+}
```
CIPop: NIT: missing new line at EOF.

src/System.Net.Security/src/System/Net/Windows/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,790 @@
...
+        public const Interop.Secur32.ContextFlags RequiredFlags = Interop.Secur32.ContextFlags.ReplayDetect |  Interop.Secur32.ContextFlags.SequenceDetect |
+                                                                  Interop.Secur32.ContextFlags.Confidentiality | Interop.Secur32.ContextFlags.AllocateMemory;
+
+        public const Interop.Secur32.ContextFlags ServerRequiredFlags =  RequiredFlags | Interop.Secur32.ContextFlags.AcceptStream;
+
+        private static volatile SecurityPackageInfoClass[] securityPackages;   
+
+        public void VerifyPackageInfo()
+        {
+            bool found = false;
```
CIPop: Breaking: Please change this back to the original code.
Please _do not_ change existing code during PAL work unless necessary.

```C#
     internal static SecurityPackageInfoClass GetVerifyPackageInfo(SSPIInterface secModule, string packageName, bool throwIfMissing)
        {
            SecurityPackageInfoClass[] supportedSecurityPackages = EnumerateSecurityPackages(secModule);
            if (supportedSecurityPackages != null)
            {
                for (int i = 0; i < supportedSecurityPackages.Length; i++)
                {
                    if (string.Compare(supportedSecurityPackages[i].Name, packageName, StringComparison.OrdinalIgnoreCase) == 0)
                    {
                        return supportedSecurityPackages[i];
                    }
                }
            }

            if (Logging.On)
            {
                Logging.PrintInfo(Logging.Web, SR.Format(SR.net_log_sspi_security_package_not_found, packageName));
            }

            if (throwIfMissing)
            {
                throw new NotSupportedException(SR.net_securitypackagesupport);
            }

            return null;
        }
```

src/System.Net.Security/src/System/Net/Windows/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,790 @@
...
+
+        #region Private Methods
+
+        private enum OP
+        {
+            Encrypt = 1,
+            Decrypt
+        }
+
+        private SecurityStatus MapToSecurityStatus(Interop.SecurityStatus secStatus)
```
CIPop: This needs to either get better or removed.

We can't rely on converting an enum to string then back to an int... Perf will be horrible.

src/System.Net.Security/src/System/Net/_SSPIWrapper.cs
```diff
@@ -5,229 +5,43 @@
 using System.Globalization;
 using System.Net.Security;
 using System.Runtime.InteropServices;
-
+using System.Security.Authentication;
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
 namespace System.Net
 {
     internal static class SSPIWrapper
```
CIPop: Design: (I think this is against what I've originally proposed... sorry.)
We need to get code back in here and push this in PAL. I need this for NegotiateStream, NTAuthentication, etc.

src/System.Net.Security/src/System/Net/Windows/SSPISecureChannelType.cs
```diff
@@ -0,0 +1,790 @@
...
+using System.Runtime.InteropServices;
+using System.Security.Authentication;
+using System.Security.Authentication.ExtendedProtection;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Principal;
+
+namespace System.Net
+{
+    // For SSL connections:
+    internal partial class SSPISecureChannelType : SSPIInterface
```
CIPop: For NegotiateStream it looks like this is better back the way it was. The PAL layer should probably be set higher than *SSPI* classes that are Windows specific.

PR #3466, opened 2015-09-25
Fix the Globalization test with checking date/time formats

tarekgh:
All globalization tests shouldn't be written assuming any fixed culture data. 
culture data (including the date/time formats) can be changed by the OS or
can be overriden by the users which make the tests to fail.

### DIFF
--- a/src/System.Globalization/tests/CultureInfo/CultureInfoDateTimeFormat.cs
+++ b/src/System.Globalization/tests/CultureInfo/CultureInfoDateTimeFormat.cs
@@ -39,55 +39,14 @@ namespace System.Globalization.Tests
         }
 
         [Theory]
-        [InlineData("en-US", "h:mm tt")]
-        public void TestShortTimePatternLocale(string locale, string expected)
+        [InlineData("fi-FI")]
+        public void TestLongTimePatternLocale2(string locale)
         {
             CultureInfo myTestCulture = new CultureInfo(locale);
-            DateTimeFormatInfo dtf = myTestCulture.DateTimeFormat;
-            Assert.Equal(expected, dtf.ShortTimePattern);
-        }
-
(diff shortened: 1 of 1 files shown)


src/System.Globalization/tests/CultureInfo/CultureInfoDateTimeFormat.cs
```diff
@@ -39,55 +39,14 @@ public void PosTest2()
...
-                Assert.True(dtf.LongTimePattern == expectedWindows || dtf.LongTimePattern == expectedIcu);
-            }
-            else
-            {
-                Assert.Equal(expectedIcu, dtf.LongTimePattern);
-            }
+            DateTimeFormatInfo myDateTimeFormat = new DateTimeFormatInfo();
+            myDateTimeFormat.LongTimePattern = "H:mm:ss";
+            myTestCulture.DateTimeFormat = myDateTimeFormat;
+            Assert.Equal("H:mm:ss", myTestCulture.DateTimeFormat.LongTimePattern);
```
stephentoub: So are we now effectively just testing the roundtripping of the pattern?

tarekgh: previously we were literally testing the pattern itself from the culture data which is wrong because we cannot assume the data will never change. what I am testing now is explicitly setting a custom data (like H:mm:ss) and checking this format persist. 
We have other tests already testing the formatting and the parsing which does the round tripping checks for the date/time values

stephentoub:
> what I am testing now is explicitly setting a custom data (like H:mm:ss) and checking this format persist.

Right, that's what I meant by roundtripping.

tarekgh:
thanks @stephentoub.

Issue #3441, opened 2015-09-24
Follow up on CR feedback for System.Net.Sockets.


pgavlin:
See #3422.

PR #3469, opened 2015-09-25
Fix a behavior mismatch in Linux Sockets.

pgavlin:
Winsock tolerates a call to shutdown[1, 2] on a TCP socket whose underlying
connection has already transitioned to the CLOSED state[3]. Linux, however,
returns ENOTCONN, since the connection is already closed.

This change fixes the mismatch by exposing the socket instance's connection
state (i.e. Socket._isConnected) to the PAL and changing the *nix
implementation of SocketPal.Shutdown to ignore ENOTCONN if the socket is
connected.

### DIFF
--- a/src/Common/tests/System.Net/Sockets/SocketTestExtensions.cs
+++ b/src/Common/tests/System.Net/Sockets/SocketTestExtensions.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net.Sockets.Tests
+{
+    static class SocketTestExtensions
+    {
+        // Binds to an IP address and OS-assigned port. Returns the chosen port.
+        public static int Bind(this Socket socket, IPAddress address)
+        {
+            socket.Bind(new IPEndPoint(address, 0));
+            return ((IPEndPoint)socket.LocalEndPoint).Port;
+        }
+    }
+}
--- a/src/Common/tests/System.Net/Sockets/SocketTestExtensions.cs
+++ b/src/Common/tests/System.Net/Sockets/SocketTestExtensions.cs
@@ -6,7 +6,7 @@ namespace System.Net.Sockets.Tests
     static class SocketTestExtensions
     {
         // Binds to an IP address and OS-assigned port. Returns the chosen port.
-        public static int Bind(this Socket socket, IPAddress address)
+        public static int BindToAnonymousPort(this Socket socket, IPAddress address)
         {
             socket.Bind(new IPEndPoint(address, 0));
             return ((IPEndPoint)socket.LocalEndPoint).Port;
--- a/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs
+++ b/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs
@@ -4432,7 +4432,7 @@ namespace System.Net.Sockets
             GlobalLog.Print("Socket#" + Logging.HashString(this) + "::Shutdown() how:" + how.ToString());
 
             // This can throw ObjectDisposedException.
-            SocketError errorCode = SocketPal.Shutdown(_handle, _isConnected, how);
+            SocketError errorCode = SocketPal.Shutdown(_handle, _isConnected, _isDisconnected, how);
 
             GlobalLog.Print("Socket#" + Logging.HashString(this) + "::Shutdown() Interop.Winsock.shutdown returns errorCode:" + errorCode);
 
@@ -5516,7 +5516,7 @@ namespace System.Net.Sockets
                     else
                     {
                         // Since our timeout is in ms and linger is in seconds, implement our own sortof linger here.
-                        errorCode = SocketPal.Shutdown(_handle, true, SocketShutdown.Send);
+                        errorCode = SocketPal.Shutdown(_handle, _isConnected, _isDisconnected, SocketShutdown.Send);
                         GlobalLog.Print("SafeCloseSocket::Dispose(handle:" + _handle.DangerousGetHandle().ToString("x") + ") shutdown():" + (errorCode == SocketError.SocketError ? SocketPal.GetLastSocketError() : errorCode).ToString());
 
                         // This should give us a timeout in milliseconds.
@@ -5604,7 +5604,7 @@ namespace System.Net.Sockets
 
             try
             {
-                SocketPal.Shutdown(_handle, _isConnected, how);
+                SocketPal.Shutdown(_handle, _isConnected, _isDisconnected, how);
             }
             catch (ObjectDisposedException) { }
         }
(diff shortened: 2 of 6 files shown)


pgavlin:
@CIPop @stephentoub PTAL.

src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs
```diff
@@ -5516,7 +5516,7 @@ protected virtual void Dispose(bool disposing)
                     else
                     {
                         // Since our timeout is in ms and linger is in seconds, implement our own sortof linger here.
-                        errorCode = SocketPal.Shutdown(_handle, SocketShutdown.Send);
+                        errorCode = SocketPal.Shutdown(_handle, true, SocketShutdown.Send);
```
stephentoub: Just my personal preference, but consider naming the arguments so that it's clear at the call site what "true" here means.

stephentoub:
LGTM

src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs
```diff
@@ -4432,7 +4432,7 @@ public void Shutdown(SocketShutdown how)
             GlobalLog.Print("Socket#" + Logging.HashString(this) + "::Shutdown() how:" + how.ToString());
```
CIPop: Nit: please correct spacing above: new line after the }.

src/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Windows.cs
```diff
@@ -802,7 +802,7 @@ public static SocketError Select(IList checkRead, IList checkWrite, IList checkE
             return (SocketError)socketCount;
         }
 
-        public static SocketError Shutdown(SafeCloseSocket handle, SocketShutdown how)
+        public static SocketError Shutdown(SafeCloseSocket handle, bool isConnected, SocketShutdown how)
         {
             SocketError err = Interop.Winsock.shutdown(handle, (int)how);
```
CIPop: This should make build do a warning since isConnected isn't used. Please add a comment to clarify why it isn't used for Windows.

src/System.Net.Sockets/tests/FunctionalTests/Shutdown.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                int received = 0;
+                do
+                {
+                    received = client.Receive(buffer);
+                } while (received != 0);
+
+                // Wait for the underlying connection to transition from TIME_WAIT to
+                // CLOSED.
+                Task.Delay(5000).Wait();
```
CIPop: Note: This is actually not the same across all platforms.
Please name the 5000 value.

src/System.Net.Sockets/tests/FunctionalTests/Shutdown.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        [Fact]
+        public void Shutdown_TCP_CLOSED_Success()
+        {
+            using (Socket server = new Socket(SocketType.Stream, ProtocolType.Tcp))
+            using (Socket client = new Socket(SocketType.Stream, ProtocolType.Tcp))
+            {
+                int port = server.Bind(IPAddress.IPv6Any);
```
CIPop: Nit: personally, I find that in this case the extension is more confusing and hiding the intent.

CIPop: Maybe a better name for the extension method?

src/Common/tests/System.Net/Sockets/SocketTestExtensions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Net.Sockets.Tests
+{
+    static class SocketTestExtensions
+    {
+        // Binds to an IP address and OS-assigned port. Returns the chosen port.
+        public static int Bind(this Socket socket, IPAddress address)
```
CIPop: Looks like an overload for Bind. In my opinion this is confusing. Maybe using a simple static helper function would be easier to understand.

CIPop:
LGTM (left a few observations)

src/System.Net.Sockets/tests/FunctionalTests/Shutdown.cs
pgavlin: > Note: This is actually not the same across all platforms.

That's correct. I'm not entirely sure that a 5-second wait is generally sufficient, in fact: technically, this should be twice the max segment lifetime, which would be 4 minutes. Experimentally, however, this works.

> Please name the 5000 value.

Okay.

src/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Windows.cs
pgavlin: It does not produce a warning: the C# compiler does not warn on unused parameters, only unused variables. I'll go ahead and add a comment.

src/System.Net.Sockets/tests/FunctionalTests/Shutdown.cs
pgavlin: What would you propose? `BindToAddress`? I personally don't find the code as written confusing, but I'm sure we can find some middle ground.

src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs
pgavlin: I'm going to leave that as-is as it's unrelated to this change. Please file an issue for follow-up (or add this to #3441).

src/System.Net.Sockets/tests/FunctionalTests/Shutdown.cs
CIPop: Since we have no documentation, we'll point our customers to the tests. They'll see this and think: why can't I do `Bind(IPAddress)`?

If I read the line it looks like that Bind is actually a Socket member, which is not. Maybe `TestBind` or even `SocketTestServer.BindToAnyIPV6(socket)`.

pgavlin: > Since we have no documentation, we'll point our customers to the tests

Can you clarify this? Don't we have documentation on MSDN?

> If I read the line it looks like that Bind is actually a Socket member, which is not

Yes, this is a problem with extension methods. I think the general consensus, however, is that the readability gains are worth the potential confusion. That said, I'll rename this to `BindAndAssignPort`.

bartonjs:
@dotnet-bot test this please (and try to use consistent package versions on all flavors)

bartonjs:
The CI system is having a package cache problem, reverting to an older build, we're overriding the build failure by committee.

PR #3470, opened 2015-09-25
Disable tests failing with ICU collation

ellismg:
We will need to investigate these tests to understand if the ICU
behavior is expected or not (it's possible that the sort weight
differences between Windows and ICU is problematic here).

Commiting these now before the ICU collation work lands in CoreCLR to
keep CI happy.


ellismg:
/cc @stephentoub @bartonjs @nguerrera 

This will need to be pulled into any branches off of master so the tests don't break once the collation stuff is merged.  My plan is to commit this before the CoreCLR changes are merged and then have a follow up commit that enables a bunch of tests so that master stays green.

nguerrera:
LGTM. (I'll pull it down to dev/metadata)

stephentoub:
LGTM

PR #3471, opened 2015-09-25
Shim'ing out GetNameInfo, get/free addrinfo, and GetHostName

sokket:
Shim'ing out GetNameInfo, get/free addrinfo, and GetHostName. The calls to getaddrinfo are not 1-to-1 with the native calls due to
complexities around the native structs. Calls using sockaddr and
addrinfo have nested structs pointing to different data depending
on if the IP is IPv4 or IPv6. To mitigate the complexities around
this code and around the underlying struct differences on the
currently supported platforms, I kept the same calling pattern (as
in Get* and Free*) but converted them to use an intermediate data
structure that is defined in the PAL. We can then take in the
required data from the Managed World and determine which native
types are required, handle all the native structs and casts, and
give back only the information that the managed code needs in a
much simpler format.

/cc @nguerrera @stephentoub @jasonwilliams200OK @pgavlin


sokket:
Note - will run the formatting script for the managed code before committing

sokket:
This should fix #2988 :)

jasonwilliams200OK:
Thank you! @sokket :)

This was supposed to be the last major stopper to build CoreFX on FreeBSD! Now BSD support should be completed. One minor note for those who would want to try to build on FreeBSD 10.2; if you are using latest Mono from portsnap, you should add `TZ=UTC` before [./build.sh#L132](https://github.com/dotnet/corefx/blob/391d5c7509b95b025b6f0f30a180438a4c9be0b2/build.sh#L132) like so: `TZ=UTC Mono29679..` ; as pointed out by @saper here: https://github.com/dotnet/coreclr/issues/602#issuecomment-108672899. This is most probably fixed in Mono master (by https://github.com/mono/mono/pull/1860 and https://github.com/mono/mono/pull/1870 but the current mono master is still not building on FreeBSD as is due to some linux specific GCC macro usage..), so until vNext of Mono is landed for FreeBSD, this patch is needed. 

After that I tried `cert-sync /usr/local/share/certs/ca-root-nss.crt` as well as `mozroots --import --sync` and `mozroots --import --sync --machine`, but it fails on `Restoring all packages...` step:

```csh
BatchRestorePackages:
  Restoring all packages...
EXEC : error : SendFailure (Error writing headers) [/root/projects/corefx/build.proj]
EXEC : error : SendFailure (Error writing headers) [/root/projects/corefx/build.proj]
EXEC : error : SendFailure (The object was used after being disposed.) [/root/projects/corefx/build.proj]
...
...
```

maybe someone can assist how to get around this issue.. /cc @akoeplinger.

src/Common/src/Interop/Unix/System.Native/Interop.GetHostName.cs
```diff
@@ -27,11 +25,10 @@ internal static unsafe string gethostname()
                 // which should only happen if the buffer we supply isn't big
                 // enough, and we're using a buffer size that the man page
                 // says is the max for POSIX (and larger than the max for Linux).
-                Debug.Fail("gethostname failed");
+                System.Diagnostics.Debug.Fail("gethostname failed");
```
stephentoub: Nit: Why remove the using and then fully qualify this?

sokket: Ah sorry this was a new file and I copy+paste the header and using's from a template. Will add the using back to keep it a simpler change to track

src/Common/src/Interop/Unix/System.Native/Interop.HostEntries.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal unsafe struct PalIpAddress
+        {
+            internal byte* Address;     // Buffer to fit IPv4 or IPv6 address
+            internal int Count;         // Number of bytes in the address buffer
+            internal bool IsIpv6;       // If this is an IPv6 Address or IPv4
+            internal byte reserved1;    // Bits here and below are reserved and called
+            internal byte reserved2;    // out individually instead of as an array
```
stephentoub: > called out individually

Could also be:
```C#
private fixed byte padding[3];
```
?

src/Common/src/Interop/Unix/System.Native/Interop.HostEntries.cs
```diff
@@ -0,0 +1,58 @@
...
+            internal byte reserved3;    // so we can use pointers to this struct
+        }
+        
+        [StructLayout(LayoutKind.Sequential)]
+        internal unsafe struct HostEntry
+        {
+            internal byte* CanonicalName;       // Canonical Name of the Host
+            internal PalIpAddress* Addresses;   // List of IP Addresses associated with this host
+            internal int Count;                 // Number of IP Addresses associated with this host
+            internal int reserved;
```
stephentoub: private instead of internal?

src/Common/src/Interop/Unix/System.Native/Interop.HostEntries.cs
sokket: I always forget about fixed arrays...that's much simpler, thanks!

src/Common/src/Interop/Unix/System.Native/Interop.HostEntries.cs
sokket: Good idea

src/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs
```diff
@@ -110,134 +110,80 @@ private static unsafe IPHostEntry CreateHostEntry(Interop.libc.hostent* hostent)
...
+            if (result != 0)
             {
-                ai_family = Interop.libc.AF_UNSPEC, // Get all address families
-                ai_flags = Interop.libc.AI_CANONNAME
+                hostinfo = NameResolutionUtilities.GetUnresolvedAnswer(name);
+                nativeErrorCode = result;
+                return GetSocketErrorForNativeError(result);
+            }
+
+            String canonicalName = Marshal.PtrToStringAnsi((IntPtr)entry->CanonicalName);
```
stephentoub: All of this until the free'ing of entry should be in a try block, with the free'ing in a finally.

sokket: Ah right, fixed

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+        lambda(*info);
+        freeaddrinfo(info);
+    }
+    else
+    {
+        *err = result;
+    }
+}
+
+static void ConvertByteArrayToV6SockAddrIn(sockaddr_in6& addr, const char* buffer, int32_t bufferLength)
```
pgavlin: s/char/uint8_t/

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+#else
+        addr.sin6_addr.__u6_addr.__u6_addr32[i] = reinterpret_cast<const uint32_t*>(addressBuffer)[i];
+#endif
+    }
+
+    // Mark that this is INET6
+    addr.sin6_family = AF_INET6;
+}
+
+static void ConvertV6SockAddrInToByteArray(char* buffer, int32_t bufferLength, const sockaddr_in6& addr)
```
pgavlin: s/char/uint8_t/

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+static void ConvertByteArrayToV6SockAddrIn(sockaddr_in6& addr, const char* buffer, int32_t bufferLength)
+{
+    assert(bufferLength == NUM_BYTES_IN_IPV6_ADDRESS);
+    (void) bufferLength;
+
+    // IPv6 should have 4 32-bit ints, copy them to our buffer
+    const uint32_t* addressBuffer = reinterpret_cast<const uint32_t*>(buffer);
+    for (int i = 0; i < 4; i++)
```
pgavlin: Is there a reason not to use `memcpy`?

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+static void ConvertV6SockAddrInToByteArray(char* buffer, int32_t bufferLength, const sockaddr_in6& addr)
+{
+    assert(bufferLength == NUM_BYTES_IN_IPV6_ADDRESS);
+    (void) bufferLength;
+
+    // IPv6 should have 4 32-bit ints, copy them to our buffer
+    uint32_t* output = reinterpret_cast<uint32_t*>(buffer);
+    for (int i = 0; i < 4; i++)
```
pgavlin: See above comment re: `memcpy`

src/Native/System.Native/pal_networking.cpp
pgavlin: I see a number of instances of this, actually. Is there a reason that you're using `char`?

src/Native/System.Native/pal_networking.h
```diff
@@ -0,0 +1,73 @@
...
+enum GetHostErrorCodes
+{
+    PAL_HOST_NOT_FOUND = 1,
+    PAL_TRY_AGAIN = 2,
+    PAL_NO_RECOVERY = 3,
+    PAL_NO_DATA = 4,
+    PAL_NO_ADDRESS = PAL_NO_DATA,
+};
+
+struct PalIpAddress
```
pgavlin: You might instead define this struct as:

```
struct HostEntry
{
    int32_t Count;
    uint8_t Data[NUM_BYTES_IN_IPV6_ADDRESS];
}
```

This would avoid an extra indirection when accessing the data, and `IsIpv6` can be derived from the value of `Count`.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+#else
+        output[i] = addr.sin6_addr.__u6_addr.__u6_addr32[i];
+#endif
+    }
+}
+
+static void ConvertByteArrayToSockAddrIn(sockaddr_in& addr, const char* buffer, int32_t bufferLength)
+{
+    assert(bufferLength == NUM_BYTES_IN_IPV4_ADDRESS);
+    (void) bufferLength;
```
pgavlin: This probably deserves a comment along the lines of "silence a compiler warning".

pgavlin: (ditto for other instances of the same)

bartonjs:
This one is failing with some segfaulting tests:

```
22:13:21 ./run-test.sh: line 172: 58393 Aborted                 (core dumped) ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCategory
22:13:21    System.Diagnostics.FileVersionInfo.Tests  Total: 3, Errors: 0, Failed: 0, Skipped: 0, Time: 0.873s
22:13:21 ./run-test.sh: line 172: 58391 Aborted                 (core dumped) ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCatego
```

http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2749/console

sokket:
I believe its the System.Net.Primitives tests due to a native static assert. Looks like our assumption of max IP address string lengths is different in managed than the define in system headers

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+const int NUM_BYTES_IN_IPV4_ADDRESS = 4;
+const int NUM_BYTES_IN_IPV6_ADDRESS = 16;
+
+static_assert(PAL_HOST_NOT_FOUND == HOST_NOT_FOUND, "");
+static_assert(PAL_TRY_AGAIN == TRY_AGAIN, "");
+static_assert(PAL_NO_RECOVERY == NO_RECOVERY, "");
+static_assert(PAL_NO_DATA == NO_DATA, "");
+static_assert(PAL_NO_ADDRESS == NO_ADDRESS, "");
+
+static void IpStringToAddressHelper(const char* address, const char* port, bool isIpV6, int32_t* err, std::function<void(const addrinfo& info)> lambda)
```
Maxwe11: AFAIK `std::function` uses type erasure which may/should result in additional allocation via operator `new` so I think this argument should be passed as `const&` to avoid additional allocations during the copying.

src/Native/System.Native/pal_networking.cpp
sokket: Nope, good catch. I changed all char's to uint8_t, put reinterpret_cast<char*> as necessary, as well as put a static assertion to make sure sizeof(char) == sizeof(uint8_t) so our casts are valid. If sizeof(uint8_t) != sizeof(char) then we'll need to revisit this file for that platform

src/Native/System.Native/pal_networking.cpp
sokket: Fixed, see above

src/Native/System.Native/pal_networking.cpp
sokket: Good idea, fixed

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+    hint.ai_family = isIpV6 ? AF_INET6 : AF_INET;
+    hint.ai_socktype = 0;
+    hint.ai_protocol = 0;
+
+    addrinfo* info = nullptr;
+    int result = getaddrinfo(address, port, &hint, &info);
+    if (result == 0)
+    {
+        *err = 0;
+        lambda(*info);
```
pgavlin: Seems simpler to just key off of `isIpv6` here and avoid the callback.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+    assert((stringLength >= INET_ADDRSTRLEN) && (stringLength <= INET6_ADDRSTRLEN));
+    (void) addressLength; // We only need this to be sure the address is the right length
+
+    int32_t result;
+    socklen_t len = UnsignedCast(stringLength);
+
+    if (isIpv6)
+    {
+        sockaddr_in6 addr = {};
+        addr.sin6_scope_id = scope;
```
pgavlin: Nit: you could use a designated initializer here:

```
sockaddr_in6 addr = {
    .sin6_scope_id = scope
};
```

pgavlin: (as well as in other situations where this pattern is repeated)

pgavlin:
There are a number of places in this change that manually pad record types (incl. automatically-generated closures) out to avoid a compiler warning on 64-bit platforms. IIUC the warning is "the compiler is padding this structure out for alignment". Have we considered disabling this warning (esp. given that different platforms are going to require different alignments, so the decisions we make now may not be universally applicable)?

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+        // Skip non-IPv4 and non-IPv6 addresses
+        if ((ai->ai_family != AF_INET) && (ai->ai_family != AF_INET6))
+            continue;
+
+        // Only increment the count of addresses here; don't try to parse on the first
+        // walk since we would (most likely) need to realloc the array, which is more
+        // expensive than just incrementing the limited number of IPs twice.
+        he->Count++;
+        infos.push_back(ai);
```
pgavlin: I'd prefer that we avoid the vector here and just repeat the check for `AF_INET` and `AF_INET6` in the loop below.

src/Native/System.Native/pal_networking.h
sokket: I started down that path and hit issues around determining length since IPv6 addresses can be variable size (::1 to full IPv6 with IPv4 mappings). Instead of dealing with that, this seemed like a cleaner and more easily understood solution, if it does have an extra indirection and one extra bit in the struct

src/Native/System.Native/pal_networking.cpp
```diff
@@ -0,0 +1,332 @@
...
+        if ((he->CanonicalName == nullptr) && (ai->ai_canonname != nullptr))
+        {
+            size_t len = strlen(ai->ai_canonname) + 1; // Name + null
+            he->CanonicalName = new char[len];
+            SafeStringCopy(he->CanonicalName, len, ai->ai_canonname);
+        }
+
+        // Skip non-IPv4 and non-IPv6 addresses
+        if ((ai->ai_family != AF_INET) && (ai->ai_family != AF_INET6))
+            continue;
```
pgavlin: Nit: this please surround the `continue` with braces:

```
        if ((ai->ai_family != AF_INET) && (ai->ai_family != AF_INET6))
        {
            continue;
        }
```

src/Native/System.Native/pal_networking.cpp
sokket: Good idea

src/Native/System.Native/pal_networking.cpp
sokket: Good idea

src/Native/System.Native/pal_networking.cpp
pgavlin: (there are a few other places in this file that have the same issue, e.g. line 291: https://github.com/dotnet/corefx/pull/3471/files#diff-a533f7324e00c23d36788e9084520906R291)

src/Native/System.Native/pal_networking.cpp
sokket: Good to know, thanks! I added that fix

src/Native/System.Native/pal_networking.cpp
sokket: The lambda just points to more unmanaged code. I contemplated merging the IPv4 and IPv6 code into one function, like the IpAddressToString function, but it would turn into a messy prototype with a multiple optional parameters that would be used depending on the version; due to this, I figured the minuscule overhead of another function call was worth it for the code cleanliness, but if you think the optional parameters would be better then I'd be happy to change it :)

src/Native/System.Native/pal_networking.cpp
sokket: Good idea

src/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs
```diff
@@ -110,134 +110,80 @@ private static unsafe IPHostEntry CreateHostEntry(Interop.libc.hostent* hostent)
...
-                    nativeErrorCode = errorCode;
-                    return GetSocketErrorForNativeError(errorCode);
-                }
+                SocketAddress sockaddr = new SocketAddress(entry->Addresses[i].IsIpv6 ? AddressFamily.InterNetworkV6 : AddressFamily.InterNetwork);
 
-                int numAddresses = 0;
-                for (Interop.libc.addrinfo* ai = addrinfo; ai != null; ai = ai->ai_next)
+                // Convert the bytes from native into an IP address
+                // Start at offset 2 in order to skip the address family bytes
+                for (int d = 2; d < entry->Addresses[i].Count; d++)
```
pgavlin: This doesn't look right: IIRC `Addresses[i].Address` contains either an `in_addr` or an `in6_addr` (from the appropraite `sockaddr`), neither of which have an address family at the start.

src/Native/System.Native/pal_networking.cpp
pgavlin: What optional parameters? Can't the signature just be `static void IpStringToAddressHelper(const char* address, const char* port, bool isIpV6, char* buffer, int bufferLength, int32_t* err)`? Or better yet, just return the error and make it `static int IpStringToAddressHelper(const char* address, const char* port, bool isIpV6, char* buffer, int bufferLength)`.

sokket: No, because the IPv6 version needs to pass in a port where as the IPv4 version needs to pass a port out; also, in IPv6, the scope needs to be passed out whereas scope doesn't exist in IPv4. Having one function to do all of this looked messy and makes the function multi-use and less clear of what parameter is used where and by whom.

src/Native/System.Native/pal_networking.cpp
sokket: Sure

src/Native/System.Native/pal_networking.cpp
pgavlin: Ah, I missed those lines. It's up to you, then: personally, I'd toss the callback and leave it up to the caller to do the `freeaddrinfo` when necessary, but I don't feel too strongly.

src/Native/System.Native/pal_networking.cpp
sokket: Fixed

src/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs
sokket: Sorry if the comment isn't clear; d is being used as an input into the SocketAddress byte array, which needs to start at 2 since the SocketAddress internal buffer has the AddressFamily stored at bytes 0 and 1. The loop does [d - 2] to grab bytes from offset 0 of the input array but puts the values into offset [d] of the output array (the SocketAddress)

sokket: Spoke offline and found the issue, fixed. Thanks Pat!

janhenke:
@dotnet-bot test this please.

sokket:
Looks like more of the new socket tests are failing. I'm investigating now, the previous failures were fixed with trivial changes so hopefully these will be the same

sokket:
@pgavlin Have you seen these failures before?

pgavlin:
@dotnet-bot test this please

pgavlin:
Yes, I have.

src/Native/System.Native/pal_networking.h
pgavlin: Good point re: deriving `IsIpv6` from the size. Perhaps this would be better, then:

```
struct PalIpAddress
{
    bool IsIpv6;
    uint8_t Count;
    uint8_t Data[NUM_BYTES_IN_IPV6_ADDRESS];
}
```

pgavlin: ...on second thought, the `PalIpAddress` isn't holding any text, right? `Data` is either an IPv4 address (always 4 bytes) or an IPv6 address (always 16 bytes). Am I missing something?

src/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs
```diff
@@ -110,134 +110,99 @@ private static unsafe IPHostEntry CreateHostEntry(Interop.libc.hostent* hostent)
...
+                    SocketAddress sockaddr;
+                    IPEndPoint factory;
+                    int bufferLength;
+                    if (entry->Addresses[i].IsIpv6)
                     {
-                        canonicalName = Marshal.PtrToStringAnsi((IntPtr)ai->ai_canonname);
-                    }
+                        sockaddr = new SocketAddress(AddressFamily.InterNetworkV6);
+                        factory = IPEndPointStatics.IPv6Any;
+                        bufferLength = SocketAddressPal.IPv6AddressSize;
```
pgavlin: Please add a comment or file an issue to track updating this once `sockaddr` et. al. are shimmed out.

pgavlin:
This is looking good aside from some comments re: `PalIpAddress`.

src/Native/System.Native/pal_networking.h
sokket: I'd prefer to keep an explicit flag to say if the address is V6 or not rather than rely on checking byte counts. This keeps the passthrough opaque to implementation and just a passthrough, letting the managed class and OS determine if the buffer actually contains valid data.

sokket:
@dotnet-bot test this please

src/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs
sokket: Opened issue #3570

sokket:
Looks like the test failure is unrelated and the assertions are a known issue that @pgavlin is investigating now. Once this build completes and I verify the same issues, this will be good to merge

sokket:
Same, expected, failures. Any further comments?

stephentoub:
@dotnet-bot test this please (I disabled the failing AssemblyLoadContext test)

sokket:
@dotnet-bot test this please (looks like a Windows timer test failed...?)

stephentoub:
> looks like a Windows timer test failed...?

Did you open an issue for it? :wink:

sokket:
Yup :) #3583

jasonwilliams200OK:
:white_check_mark: :+1:

sokket:
Green run - any more comments or can this be merged?

pgavlin:
LGTM. We can address any other feedback during follow-up changes.

sokket:
Cool, merging. Thanks everyone!

PR #3478, opened 2015-09-26
Additional String.StartsWith tests

bbowyersmyth:
Greater coverage and StringComparision usage.

"Soft Hyphen" is ignored on all Windows cultures. I'll see from the CI build whether this is also holds true on Linux. Otherwise I might have to use a fixed culture and chars known to work.

### DIFF
--- a/src/System.Runtime/tests/System/String.cs
+++ b/src/System.Runtime/tests/System/String.cs
@@ -8,6 +8,8 @@ using Xunit;
 
 public static unsafe class StringTests
 {
+    private const string c_SoftHyphen = "\u00AD";
+
     [Fact]
     public static void TestWithEmptyString()
     {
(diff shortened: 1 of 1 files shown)


bbowyersmyth:
@dotnet-bot test this please

src/System.Runtime/tests/System/String.cs
```diff
@@ -1137,42 +1139,54 @@ public static void TestReplaceString()
...
-        b = s.StartsWith("", StringComparison.CurrentCultureIgnoreCase);
-        Assert.True(b);
-        b = s.StartsWith("HEL", StringComparison.CurrentCultureIgnoreCase);
-        Assert.True(b);
-        b = s.StartsWith("Goodbye", StringComparison.CurrentCultureIgnoreCase);
-        Assert.False(b);
+        if (comparisionType == StringComparison.CurrentCulture)
+        {
+            Assert.Equal(expected, text.StartsWith(value));
+        }
```
stephentoub: Why have this special case?  Wouldn't be sufficient to just always do:
```C#
Assert.Equal(expected, text.StartsWith(value, comparisonType);
```

src/System.Runtime/tests/System/String.cs
```diff
@@ -1137,42 +1139,54 @@ public static void TestReplaceString()
...
+    [InlineData(StringComparison.Ordinal, "Hello", "Abc", false)]
+    [InlineData(StringComparison.Ordinal, "Hello", c_SoftHyphen + "Hel", false)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "Hel", true)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "Hello", true)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "Hello Larger", false)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "", true)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "HEL", true)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", "Abc", false)]
+    [InlineData(StringComparison.OrdinalIgnoreCase, "Hello", c_SoftHyphen + "Hel", false)]
+    public static void TestStartsWith(StringComparison comparisionType, string text, string value, bool expected)
```
stephentoub: Nit: comparisionType => comparisonType

stephentoub:
cc; @ellismg, @tarekgh

src/System.Runtime/tests/System/String.cs
ellismg: I've seen this pattern done (and done it my self) in a few places for code coverage purposes (now you get `String.StartsWith(String)` covered)).

Personally, I could go either way.

ellismg:
LGTM.  Nice to see ICU passing on this as well!  Thanks very much for these additions!

src/System.Runtime/tests/System/String.cs
JonHanna: With that many InlineData attributes, maybe a MemberData would be worthwhile?

src/System.Runtime/tests/System/String.cs
stephentoub: Ah, makes sense. Ok.

bbowyersmyth:
Fixed comparisonType typo

src/System.Runtime/tests/System/String.cs
bbowyersmyth: Yep that was the reason. Just to get it included in the testing.

PR #3479, opened 2015-09-26
Native: Multiple fixes

jasonwilliams200OK:
##### <del>067ea11</del> d9cf2d8

* Fix wrong cast. Clang on FreeBSD caught this as error as `id_t` is not the type `[get/set]priority` functions are expecting:
  * Linux man (uses `id_t who`): http://man7.org/linux/man-pages/man2/getpriority.2.html
  * FreeBSD man (uses `int who`): https://www.freebsd.org/cgi/man.cgi?query=getpriority&sektion=2 (this applies to other BSDs as well, such as [DragonFly](https://www.dragonflybsd.org/cgi/web-man?command=getpriority&section=2) and [NetBSD](http://nixdoc.net/man-pages/NetBSD/man2/getpriority.2.html), except for [OpenBSD](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/getpriority.2?query=getpriority), hence I've used <del>`#if !defined(BSD) ..`</del> `cmake_prototype_definition` to detect the correct prototype).

##### <del>207d186</del> e348434

* Removes unrequired dependencies from native ` src/Native/gen-buildsys-clang.sh`.
  * It builds without llvm-* dependencies.
  * Cleanup unused code blocks.
  * Minor conditions and comment fixes.

### DIFF
--- a/src/Native/System.Native/pal_process.cpp
+++ b/src/Native/System.Native/pal_process.cpp
@@ -389,12 +389,20 @@ extern "C" int32_t GetPriority(PriorityWhich which, int32_t who)
 {
     // GetPriority uses errno 0 to show succes to make sure we don't have a stale value
     errno = 0;
+#if PRIORITY_REQUIRES_INT_WHO
+    return getpriority(which, who);
+#else
     return getpriority(which, static_cast<id_t>(who));
+#endif
 }
 
 extern "C" int32_t SetPriority(PriorityWhich which, int32_t who, int32_t nice)
 {
+#if PRIORITY_REQUIRES_INT_WHO
+    return setpriority(which, who, nice);
+#else
     return setpriority(which, static_cast<id_t>(who), nice);
+#endif
 }
 
 extern "C" char* GetCwd(char* buffer, int32_t bufferSize)
--- a/src/Native/gen-buildsys-clang.sh
+++ b/src/Native/gen-buildsys-clang.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# This file invokes cmake and generates the build system for gcc.
+# This file invokes cmake and generates the build system for Clang.
 #
 
 if [ $# -lt 3 -o $# -gt 4 ]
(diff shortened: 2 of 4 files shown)


stephentoub:
cc: @nguerrera, @sokket

src/Native/System.Native/pal_process.cpp
```diff
@@ -11,6 +11,7 @@
 #include <limits.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <sys/param.h>
```
nguerrera: Out of curiosity, why is this needed?

src/Native/System.Native/pal_process.cpp
```diff
@@ -389,12 +390,20 @@ extern "C" int32_t GetPriority(PriorityWhich which, int32_t who)
 {
     // GetPriority uses errno 0 to show succes to make sure we don't have a stale value
     errno = 0;
+#if !defined(BSD) || defined(__OpenBSD__) || defined(__APPLE__)
```
nguerrera: We should use CMAKE config or a pattern that would let the compiler infer the right thing here. (Also, pre-existing, but ideally, that cast to id_t would be checked with at least an assert if id_t is not wider than int32_t).  

We should not have #ifdef SPECIFIC_OS in the cpp files. (As a last, resort it can be used in config.h.in if the config test is unwieldy, see HAVE_STAT64 vs. __APPLE__)

nguerrera:
Thanks, @jasonwilliams200OK. I had two comments, but the rest looks good.

src/Native/gen-buildsys-clang.sh
```diff
@@ -46,11 +46,11 @@ OS=`uname`
...
+# number postfixed (i.e. llvm-3.5), so we check for that first.
 # On FreeBSD the version number is appended without point and dash (i.e.
-# llvm-ar35).
+# llvm-35).
 # Additionally, OSX doesn't use the llvm- prefix.
-if [ $OS = "Linux" -o $OS = "FreeBSD" -o $OS = "OpenBSD" -o $OS = "NetBSD" ]; then
+if [ $OS = "Linux" -o $OS = "DragonFly" -o $OS = "FreeBSD" -o $OS = "NetBSD" -o $OS = "OpenBSD" ]; then
   llvm_prefix="llvm-"
 elif [ $OS = "Darwin" ]; then
   llvm_prefix=""
```
nguerrera: Just found a third issue: Are we even using locate_llvm_exec or llvm_prefix anymore?

src/Native/System.Native/pal_process.cpp
jasonwilliams200OK: This is for the check `#ifdef BSD`:
`sys/param.h` is available in all POSIX and `BSD` macro is only defined in BSD-backed OSes.

src/Native/gen-buildsys-clang.sh
jasonwilliams200OK: Nice catch! I thought it is used to check llvm existence and early exit. Fixed by 9e22f86.

src/Native/System.Native/pal_process.cpp
```diff
@@ -389,12 +389,20 @@ extern "C" int32_t GetPriority(PriorityWhich which, int32_t who)
...
+#ifdef PRIORITY_REQUIRES_INT_WHO
+    return getpriority(which, who);
+#else
     return getpriority(which, static_cast<id_t>(who));
+#endif
 }
 
 extern "C" int32_t SetPriority(PriorityWhich which, int32_t who, int32_t nice)
 {
+#ifdef PRIORITY_REQUIRES_INT_WHO
```
nguerrera: The build is failing on Linux and Mac. This needs to be #if not #ifdef since we're using #cmakedefine01.

jasonwilliams200OK: Thanks @nguerrera! I was just trying to figure out what's going wrong (by debugging with the old fashion print way using `MESSAGE(STATUS ..)` construct). :)
This is fixed by d9cf2d8.
Incidentally, do you still want me to add that assertion for `id_t` width vs. `int32_t` (using `sizeof` operator)?

jasonwilliams200OK:
Thanks for the review @nguerrera. I have fixed the issues and re-pushed.

Regarding the `assert` comment, do you want me to address it by checking `id_t` width vs. `int32_t` (using `sizeof` operator) as part of this PR?

jasonwilliams200OK:

nguerrera:
Nevermind the assert since it was already missing. (Aside: Note that sizeof() check isn't sufficient since it doesn't account for possible difference in signedness between int32_t and id_t). 

I'm going to make another change to make potentially narrowing conversions like that go through a helper and also avoids having to cmake configure every occurrence like this. It'll also help clean up some things in the crypto shim.

This LGTM once it gets through a successful CI.

PR #3486, opened 2015-09-27
Remove platform specific hack in test projects

ellismg:
Previously CI would use the output of the Windows build to grab the test
assemblies for all the cross plat builds. To work around this some test
projects had a csproj per platform which set properties and then
included a common file. The CI system no longer has this limitation (it
now uses the Linux build to get the Linux tests, etc.) so we can remove
this hack.


ellismg:
@pgavlin PTAL

pgavlin:
LGTM. Thanks for doing this.

PR #3487, opened 2015-09-27
Native: Disables a warning for clang v3.6.

jasonwilliams200OK:
While building with clang36 on FreeBSD (`./build native clang3.6`),
I was getting this warning: `macro name is a reserved identifier`.

### DIFF
--- a/src/Native/configure.cmake
+++ b/src/Native/configure.cmake
@@ -6,7 +6,7 @@ include(CheckPrototypeDefinition)
 
 #CMake does not include /usr/local/include into the include search path
 #thus add it manually. This is required on FreeBSD.
-include_directories(/usr/local/include)
+include_directories(SYSTEM /usr/local/include)
 
 check_function_exists(
     stat64


akoeplinger:
Out of curiosity: which macro name is reserved?

jasonwilliams200OK:
@akoeplinger, 

```

Executing make install -j 2
Scanning dependencies of target System.Native
Scanning dependencies of target System.Net.Http.Native
[ 20%] Building CXX object System.Net.Http.Native/CMakeFiles/System.Net.Http.Native.dir/pal_curlinit.cpp.o
[ 10%] Building CXX object System.Native/CMakeFiles/System.Native.dir/pal_errno.cpp.o
In file included from /root/corefx/src/Native/System.Net.Http.Native/pal_curlinit.cpp:8:
/usr/local/include/curl/curl.h:2:9: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#define __CURL_CURL_H
        ^
In file included from /root/corefx/src/Native/System.Net.Http.Native/pal_curlinit.cpp:8:
In file included from /usr/local/include/curl/curl.h:33:
/usr/local/include/curl/curlver.h:2:9: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#define __CURL_CURLVER_H
        ^
In file included from /root/corefx/src/Native/System.Net.Http.Native/pal_curlinit.cpp:8:
In file included from /usr/local/include/curl/curl.h:34:
/usr/local/include/curl/curlbuild.h:3:9: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#define __CURL_CURLBUILD_H
        ^
In file included from /root/corefx/src/Native/System.Net.Http.Native/pal_curlinit.cpp:8:
In file included from /usr/local/include/curl/curl.h:35:
/usr/local/include/curl/curlrules.h:2:9: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#define __CURL_CURLRULES_H
        ^
/usr/local/include/curl/curlrules.h:228:13: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#    define __CURL_OFF_T_C_HLPR2(Val,Suffix) Val ## Suffix
            ^
/usr/local/include/curl/curlrules.h:232:11: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#  define __CURL_OFF_T_C_HLPR1(Val,Suffix) __CURL_OFF_T_C_HLPR2(Val,Suffix)
          ^
```

It was building with clang 3.4 and 3.5 but not 3.6 and 3.7. With the conditional flag, it now builds with 3.6 and 3.7 as well. 

Weirdly I don't get clang37 in FreeBSD 10.2 via `pkg search clang` (on FreeBSD 10.1, it is available though !? had to boot into that one to verify..) 

Note that I had to check if compiler has this option because on Mac, AppleClang v6x is used and it doesn't require / recognize this flag, hence CMake was failing. Same the case with (regular) Clang < v3.6 on other Unices..

akoeplinger:
Interesting that this warning fires since that isn't even our code. Looks good to me otherwise.

stephentoub:
cc: @nguerrera, @sokket

nguerrera:
Hmm, it seems the system header  logic is of f somehow. Rather than disabling the warning, can we give clang a nudge that curl headers are "system".

http://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers

jasonwilliams200OK:
@nguerrera, nice idea. Fixed by 40ee8c0.

nguerrera:
Thanks.

PR #3493, opened 2015-09-28
Exclude NotImplemented from code coverage.

JonHanna:
NotImplemented.ByDesign is by its nature not often hit by useful tests.
Including it in code coverage reduces how useful code coverage is in
identifying areas that need greater testing. Therefore it should be
excluded from coverage.

### DIFF
--- a/src/System.Xml.XPath.XmlDocument/src/System.Xml.XPath.XmlDocument.csproj
+++ b/src/System.Xml.XPath.XmlDocument/src/System.Xml.XPath.XmlDocument.csproj
@@ -83,6 +83,9 @@
     <Compile Include="$(CommonPath)\System\NotImplemented.cs">
       <Link>Common\System\NotImplemented.cs</Link>
     </Compile>
+    <Compile Include="$(CommonPath)\System\Diagnostics\CodeAnalysis\ExcludeFromCodeCoverageAttribute.cs">
+      <Link>Common\System\Diagnostics\CodeAnalysis\ExcludeFromCodeCoverageAttribute.cs</Link>
+    </Compile>
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
--- a/src/Common/src/System/NotImplemented.cs
+++ b/src/Common/src/System/NotImplemented.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace System
 {
     //
@@ -14,6 +16,7 @@ namespace System
     //
     // This makes it distinguishable both from human eyes and CCI from NYI's that truly represent undone work.
     //
+    [ExcludeFromCodeCoverage]
     internal static class NotImplemented
     {
         internal static Exception ByDesign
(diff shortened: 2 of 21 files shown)


stephentoub:
> Including it in code coverage reduces how useful code coverage is in
identifying areas that need greater testing

How much of a difference did this actually make in having the numbers better reflect reality?  Many of these uses are in tests, which are already excluded.  Others are in Unix implementations, for which we don't currently measure coverage.  Etc.

JonHanna:
Very little, because it's small, but it changes eyeballing for red as a sign that something could benefit from some work.

src/System.Xml.XPath.XmlDocument/src/System.Xml.XPath.XmlDocument.csproj
```diff
@@ -83,6 +83,10 @@
     <Compile Include="$(CommonPath)\System\NotImplemented.cs">
       <Link>Common\System\NotImplemented.cs</Link>
     </Compile>
+    <Compile Include="$(CommonPath)\System\Diagnostics\CodeAnalysis\ExcludeFromCodeCoverageAttribute.cs">
+      <Link>Common\System\Diagnostics\CodeAnalysis\ExcludeFromCodeCoverageAttribute.cs</Link>
+    </Compile>
+    <Compile
```
stephentoub: This won't compile.

JonHanna: It certainly won't. Now, however did that slip in?

nguerrera:
Hmm, if it throws NotImplemented *by design*, why can't we test that? The idea with NotImplemented.ByDesign is to distinguish the case where we really want to ship throwing NotImplemented (compat generally) and it's not e.g. an unfinished port/todo.

JonHanna:
@nguerrera possibly because it's unreachable.

Of course, it could be reachable, and then should be tested, but then the site of it being called will still lack coverage it should have.

stephentoub:
> but it changes eyeballing for red as a sign that something could benefit from some work.

I'm not understanding this part.  This change is going to make it so that the NotImplemented class itself is excluded from coverage, not the call sites to it.  The call sites are what would show up in red (and are what we should keep red because they're legitimate lines that we should be considering for coverage), unless you actually navigated to the NotImplemented.cs file, which there'd be no reason to do.  

Are you just referring to the summary table, which would have a line entry for NotImplemented.cs that would potentially show 0% coverage?

JonHanna:
> Ah, I think I understand, you meant eyeballing for red in the summary table for the whole assembly, which not only shows numbers but also a green/red bar depicting that number? I thought you meant green/red for individual lines in a .cs file.

Yes. 0% or low % mostly means somewhere that needs work in ensuring that they have good test cases, but NotImplemented is just a distraction.

stephentoub:
I've been holding off on merging this because I was concerned that this would break our internal TFS builds once it mirrored down.  I finally got a chance to check, and indeed there are other internal projects not yet in the open that depend on NotImplemented.cs and that will break the build due to the projects not also including ExcludeFromCodeCoverageAttribute.cs.

I understand your desire to make this change, but given the limited value it provides and the logistics involved in fixing this (the internal projects would also need to verified on .NET Native), I'm going to close this out for now, and we can potentially revisit it later.

Thanks for the suggestion.

Issue #3502, opened 2015-09-28
Windows: new X509Certificate2(byte[]).GetRSAPrivateKey returns RSACSP instead of RSACng


bartonjs:
The idea/thought was that we'd use CNG for everything except for hardware providers that were linked through CAPI.  In particular, this would prevent the SHA256 lookup failure that occurs when using CAPI's PROV_RSA_FULL instead of PROV_RSA_AES.

I'm pretty sure that we made the software keys be returned via RSACng on desktop (4.6).  This was accomplished by calling CryptAcquireCertificatePrivateKey with the "ncrypt only" flag (see the code on [reference source](http://referencesource.microsoft.com/#System.Core/System/Security/Cryptography/X509Certificates/X509Native.cs,2e55137aa586849f)), since that can up-convert software based CAPI keys.

bartonjs:
This was fixed by #3507, but didn't tag appropriately.

PR #3507, opened 2015-09-28
Fix issue #3502

AtsushiKan:
cc @bartonjs 

- The "ProviderType == 0" path is now is probably unreachable and merits an assert. But leaving it alone is cheap insurance against our being around about the "probably."


bartonjs:
LGTM (looks rather like what is in referencesource :)).

As a sanity test, can you edit PfxTests::TestPrivateKey, remove the Windows ActiveIssue attribute, and change the hash algorithm to SHA256, then `msbuild /p:BuildAndTest` from X509\tests?  That's what I did to confirm the issue this morning.  Since we're still having intermittent PFX conflicts on Windows that test change shouldn't be checked in :)

AtsushiKan:
Done – test works (well, it fails the expectedsig comparison but that’s expected since we changed the hash size and input…)


.

bartonjs:
Thanks for jumping on this so quickly, @AtsushiKan.

PR #3508, opened 2015-09-28
Change S.N.Sockets/tests to use (mostly) anonymous ports.

pgavlin:
This change only touches tests in System.Net.Sockets/tests; those under
S.N.Sockets.Legacy/tests will be taken care of in a follow-up commit.

### DIFF
--- a/src/Common/tests/System.Net/Sockets/Performance/SocketPerformanceTests.cs
+++ b/src/Common/tests/System.Net/Sockets/Performance/SocketPerformanceTests.cs
@@ -26,20 +26,30 @@ namespace System.Net.Sockets.Performance.Tests
             SocketImplementationType clientType, 
             int iterations, 
             int bufferSize, 
-            int socket_instances, 
+            int socketInstances, 
             long expectedMilliseconds)
         {
             long milliseconds;
 
 #if arm
-        iterations /= 100;
+            iterations /= 100;
 #endif
 
-            using (SocketTestServer.SocketTestServerFactory(
-                serverType,
-                socket_instances * 5,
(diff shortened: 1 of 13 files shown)


pgavlin:
@CIPop @stephentoub @davidsh PTAL

pgavlin:
Ping: I think that this should help with some of the stability issues we've been seeing, so I'd like to get this in ASAP. I have a follow-up change waiting in the wings as well (see https://github.com/pgavlin/corefx/tree/SocketsAnonymousPorts2).

stephentoub:
Thanks, Pat.  LGTM, but @cipop should also take a look.

src/Common/tests/System.Net/Sockets/Performance/SocketPerformanceTests.cs
```diff
@@ -55,13 +65,24 @@ public SocketPerformanceTests(ITestOutputHelper log)
...
 
+        public void ClientServerTest(
+            SocketImplementationType serverType, 
+            SocketImplementationType clientType, 
+            int iterations, 
+            int bufferSize, 
+            int socketInstances, 
+            long expectedMilliseconds)
+        {
+            ClientServerTest(0, serverType, clientType, iterations, bufferSize, socketInstances, expectedMilliseconds);
```
CIPop: Nit: I was a bit confused on what 0 meant here (since this is a _Client_ / Server test and client port 0 doesn't make sense). It would be nice to add a comment to explain that 0 means automatic port assignment.

CIPop:
LGTM
Thanks @pgavlin for adding this enhancement!

PR #3509, opened 2015-09-28
Improve efficiency of BigInteger.ToByteArray()

dtrebbien:
ToByteArray() was previously implemented in two passes. The first pass
would create a byte array with space for 4 * dwords.Length bytes and
write all uint elements of dwords into the byte array in little Endian
form. The second pass would create a new array to store the result,
which would be the array created by the first pass, trimmed to remove
unnecessary high bytes, and padded with a high byte if the extra byte
was needed.

This commit makes BigInteger.ToByteArray() more efficient by combining
the two passes and performing two's complement on the stack without
creating a copy of _bits.


dtrebbien:
cc: @stephentoub @KrzysztofCwalina @axelheer

src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
```diff
@@ -275,59 +275,107 @@ int IComparable.CompareTo(Object obj)
         // return an array of one byte whose element is 0x00.
         public byte[] ToByteArray()
         {
-            if (_bits == null && _sign == 0)
+            int sign = _sign;
+            if (sign == 0)
```
KrzysztofCwalina: I prefer to wrap even single-statement ifs in braces. But I am not sure if it's our general convention.

KrzysztofCwalina:
I would like @axelheer to review it, but otherwise looks good.

src/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs
```diff
@@ -275,59 +275,107 @@ int IComparable.CompareTo(Object obj)
...
             {
-                dwords = _bits;
+                Debug.Assert(sign == 1);
                 highByte = 0x00;
+                highDword = bits[bits.Length - 1];
+            }
+
+            byte msb;
+            int msbIndex;
+            if ((msb = (byte)((highDword & 0xff000000U) >> 24)) != highByte)
```
axelheer: The binary ands shouldn't be necessary, the shift/cast remove the other bits anyway.

axelheer:
+1

stephentoub:
@dotnet-bot test this please

dtrebbien:
I am working on another commit to add braces around single-line `if` statements and to remove the unnecessary bitwise-ANDs per @axelheer's feedback.  Do I need to wait for dotnet-bot to finish?  How can I see the test results?

stephentoub:
> Do I need to wait for dotnet-bot to finish? 

Nope, though the run is done anyway.

> How can I see the test results?

Click on the Details link in the bottom section of the page. That'll take you to the associated job, and from there you can branch off to view the individual legs.

dtrebbien:
New commit added and rebased for @979026e.

src/System.Runtime.Numerics/tests/BigInteger/ToByteArray.cs
```diff
@@ -0,0 +1,181 @@
...
+            Assert.Equal(new byte[] { 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x01, 0xff }, bi.ToByteArray());
+
+            bi = new BigInteger(-0xffffff00000000L);
+            Assert.Equal(new byte[] { 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0xff }, bi.ToByteArray());
+
+            bi = BigInteger.Parse("-18374686475376656384");
+            Assert.Equal(new byte[] { 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x01, 0xff }, bi.ToByteArray());
+
+            bi = BigInteger.Parse("-18446744069414584320");
+            Assert.Equal(new byte[] { 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0xff }, bi.ToByteArray());
```
stephentoub: I realize this value doesn't fit in an Int64, but could you add a couple of Parse-based tests for much bigger values as well?

Also, we should probably verify not only the expected by array but that roundtripping works correctly.  Do we have any such tests elsewhere? e.g. passing bi back to the ctor and verifying that it and the original BigInteger are the same value?

Given that doing so would expand the number of lines per test case a bit, you might also consider doing this as a [Theory] with [InlineData] instead, such that we end up with a reusable test body applied to a bunch of one-line [InlineData] attributes that contain the input and the expected output.

stephentoub:
One additional suggestion, but otherwise LGTM.  Could you also squash this down to a single commit? Thanks!

dtrebbien:
Updated & squashed commit pushed

stephentoub:
@dotnet-bot test this please

stephentoub:
Thanks, @dtrebbien.

PR #3511, opened 2015-09-28
Improve UnixFileStream.Read/WriteAsync consistency and perf

stephentoub:
This commit addresses two related issues.

1) When a FileStream is created on Windows with FileOptions.Asynchronous (the default in corefx), synchronous Read and Write operations are actually implemented as blocking wrappers for ReadAsync and WriteAsync.  Since in FileOptions.Asynchronous mode ReadAsync and WriteAsync can be used to issue requests that run concurrently, this also means that a call to Read/Write could safely be made on the stream while a ReadAsync/WriteAsync operation was in progress.  In contrast in the current Unix implementation, Read/Write are not implemented as wrappers for ReadAsync/WriteAsync, so although ReadAsync/WriteAsync can have operations overlapped, using Read/Write while such an operation was outstanding could result in corruption of the FileStream, and in particular of its internal buffer.  In general, such usage patterns are not encouraged, but we should aim to be consistent across platforms in how we handle them and whether corruption is possible.

2) On Windows, when in FileOptions.Asynchronous mode, ReadAsync/WriteAsync use the stream's internal buffer directly, potentially avoiding making any asynchronous calls and just returning synchronously if the request can be satisfied using the buffer.  On Unix we were never doing that.

This commit addresses these as follows:
- One extra field is added to UnixFileStream that stores additional state used when FileOptions.Asynchronous is set.  This state includes a semaphore used to synchronize operations, as well as storage for the buffer/offset/count from the currently active async call, and a cached task that can be handed back for some ReadAsync operations.
- When FileOptions.Asynchronous is not set, little changes: Read/Write/ReadAsync/WriteAsync/ReadByte/WriteByte all check _useAsyncIO, and do exactly what they do today if it's false.
- When FileOptions.Asynchronous is set, Read/Write/ReadByte/WriteByte wrap their current behavior with calls to the semaphore's Wait/Release, such that they will be appropriately serialized with other outstanding asynchronous operations happening on the stream.  This addresses issue (1).
- For (2), when FileOptions.Asynchronous is set, ReadAsync/WriteAsync check whether there is enough data/space in the internal buffer to satisfy the request synchronously, and if there is they do the buffer copy synchronously and return an already completed task.  Since it's very common to ask for the same amount of data on each call, and thus if it's satisified by the buffer get exactly the same amount of data on each call, in ReadAsync we cache a task with a Result of the last number of bytes returned, and we reuse that task the next time if it contains the appropriate count.  This same optimization is already employed in the Windows implementation. When we do need to schedule work asynchronously, we also do so explicitly using a continuation off of the semaphore's WaitAsync rather than using the base implementation.  Due to the base Stream's implementation and its need to account for legacy async models, this ends up reducing the amount of memory used to schedule the async operation by ~50%.

cc: @ellismg, @sokket, @ianhays, @ericstj

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
+++ b/src/System.IO.FileSystem/src/System/IO/UnixFileStream.cs
@@ -1,13 +1,12 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Win32.SafeHandles;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-
-using Microsoft.Win32.SafeHandles;
 
 namespace System.IO
 {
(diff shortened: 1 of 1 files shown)


stephentoub:
@dotnet-bot test this please ("unable to connect to github.com")

sokket:
Thanks for the great explanation Stephen! LGTM

stephentoub:
Thanks, Jon.

PR #3513, opened 2015-09-28
Update all CoreFx package references to 23328 build.

bartonjs:
`build.cmd /p:LockDependencies=true` was the start of this change, but some errors with xunit build numbers led to all xunit upgrades being reset back to what they had in master.

project.json changes:
* System.Net.Security\src Switch off of explicit beta versioning, use the same baseline as everyone else.  The S.S.C.Algorithms explicit upgrade is no longer required.
* System.Net.Requests\tests didn't build.  Looking at the project.lock.json change, it appeared that a System.Net.Http reference was lost, restored it.
* System.Net.Http\tests didn't build. Looking at the project.lock.json change, it appeared that a System.IO.FileSystem reference was lost, restored it.
* System.Net.WebSockets.Client\src didn't build. Looking at the project.lock.json change, it appeared that a System.Collections (4.0.10) reference was lost, restored it.

### DIFF
--- a/src/Common/tests/System/Diagnostics/RemoteExecutorConsoleApp/project.lock.json
+++ b/src/Common/tests/System/Diagnostics/RemoteExecutorConsoleApp/project.lock.json
@@ -3,7 +3,7 @@
   "version": 1,
   "targets": {
     "DNXCore,Version=v5.0": {
-      "System.Console/4.0.0-beta-23311": {
+      "System.Console/4.0.0-beta-23328": {
         "type": "package",
         "dependencies": {
           "System.IO": "4.0.0",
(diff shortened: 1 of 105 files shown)


bartonjs:
cc: @stephentoub @ericstj

weshaggard:
@pallavit was also in the process of updating the lock files.

bartonjs:
@weshaggard Mainly my interest was in undoing the unusual beta build specifiers that I added to System.Net.Sockets.  @pallavit if you want to take that edit into your version I'll happily close this one out :smile:.

I'm having trouble understanding why `System.Diagnostics.TextWriterTraceListenerTests.CtorsStreamTests.TestConstructorWithStreamAndName(testName: "MyXMLTraceWriter")` can't load `System.Diagnostics.TraceSource`.  I sort of see it in the project.lock.json change; but sort of don't; it should just be a runtime package now.

weshaggard:
If you just want to limit your update to System.Net.Sockets, which is probably best, then change just those projects and rebuild just those projects before you do a full rebuild of the repo. This is currently a bug in dnu (https://github.com/aspnet/dnx/issues/2771) where if it finds one invalidated lock file it will invalidate and re-evaluate all the lock files in the repo. That is what you are hitting and what @pallavit was hitting. At this point I think it is a secondary goal to update all the lock files. 

As for the TraceSource issue that is a side-effect of our recent package refactoring which would require the test project that depends on this to add a "runtimes" section to get it pulled in correctly.

bartonjs:
Synced up with @pallavit, and got the System.Diagnostics.TextWriterTraceListenerTests problem worked out (hopefully).

Since what she was trying to do was scoped, and I was (among other things) trying to update the baseline to be consistent, I think we should go ahead and take this one (while she continues to wrestle with her share of problems here).

bartonjs:
@weshaggard please take a look.  Though if you're disagreeing with the principle of upgrading the baseline to Beta8 then that's a different kettle of fish :smile:.

bartonjs:
@dotnet-bot test this please

weshaggard:
@bartonjs I don't have any concerns with the overall update, I just wanted to make sure you guys are sync on the changes. As long as everything builds and works I'm fine with the refresh.

PR #3514, opened 2015-09-28
Fix Markdown typo in the coding style

Alcaro:
The closing accent of a monospace block got lost somewhere, leading to the obvious incorrect results.

I'm not sure if "any Cpp or H file" at the top is the correct capitalization, but I can't think of anything better.


stephentoub:
LGTM. Thanks for the fix.

PR #3517, opened 2015-09-29
Add TypeNameParser tests

### LINKED ISSUE #534: Port and commit type-name parsing tests
@nguerrera just committed vnext of the metadata reader (https://github.com/nguerrera/corefx/tree/mdreader-vnext) with the new signature and attribute parsing support. We need to port the Reflection-style type-naming parsing tests to XUnit and check them in.

nguerrera:
@davkean Adding tests for your type name parser. I marked you as author on the commit with your original VS unit tests, then followed it up with commits to port to xunit. 

The port is minimal, and I have not for example bothered to convert to [Theory] etc. yet. 

I've also separated out the part that uses desktop-only API to compare against reflection type name parsing in to the .Desktop test project. Mind you, that project is temporary until we get test explorer working with portable tests, so I'm not sure we can keep that long term.

cc @tmat

Fix #534


nguerrera:
Merging to dev branch optimistically. Need this to unblock follow-up changes to parser. Please do review and I'll fix things up in follow-up commits if necessary.

PR #3519, opened 2015-09-29
Use anonymous ports in S.N.Sockets.Legacy/test/FunctionalTests. 

pgavlin:
This is a follow-up to #3508. With this change, all of the legacy functional
tests use anonymous ports.

This also includes a small change to `SocketPerformanceTests.cs` to address PR feedback from #3508.


pgavlin:
@cipop @davidsh @stephentoub PTAL

stephentoub:
LGTM

src/System.Net.Sockets.Legacy/tests/FunctionalTests/AgnosticListenerTest.cs
```diff
@@ -23,7 +21,7 @@ public AgnosticListenerTest(ITestOutputHelper _log)
         [Fact]
         public void Create_Success()
         {
-            TcpListener listener = TcpListener.Create(TestPortBase);
+            TcpListener listener = TcpListener.Create(0);
```
CIPop: Nit: Please add a comment on what 0 means.

CIPop:
Thanks Pat! I really like how the code ended up looking: this and the previous PR were both very clean changes. 
I already have it in my plan to port the Perf tests back to Desktop. I think this change is also worth porting back to ensure 100% app-compat. I'll add it to my list.

LGTM

PR #3521, opened 2015-09-29
Disable OSX tests failing in CI

ellismg:


src/System.Threading.Tasks.Dataflow/tests/Dataflow/EtwTests.cs
```diff
@@ -9,6 +9,7 @@ namespace System.Threading.Tasks.Dataflow.Tests
     public class EtwTests
     {
         [Fact]
+        [ActiveIssue(3520, PlatformID.OSX)]
```
stephentoub: Can you verify that the right project is being built and used for OS X? The .WP8.csproj doesn't include the ETW support and would result in this failure. I'm surprised if this attribute is actually needed.

ellismg: I see.  This could be an interaction between my recent changes to speed up the test runs and how run-tests.sh builds out the overlay.  I think that perhaps what is happening is that previously we would never copy down the WP8 version and now we are and run-test.sh's way of building the overlay could be picking it up and using it instead of the regular version.  Thanks for the pointer.

stephentoub: Thanks. Otherwise, LGTM.

bartonjs:
X509 LGTM.

src/System.Threading.Tasks.Dataflow/tests/Dataflow/EtwTests.cs
stephentoub: Matt, the EtwTests is now failing on Linux as well, e.g. http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst/1052/testReport/junit/System.Threading.Tasks.Dataflow.Tests/EtwTests/TestEtw/.  Were you able to track down the script problem?

stephentoub:
I'm going to go ahead and check this in to cover us until a script fix is available, assuming that's the issue.

PR #3522, opened 2015-09-29
Simple removal one line redundant code.

JonHanna:
ReadOnlyCollection is used to create an array, but that array is then never
used.

As the collection cannot be null, doesn't even change exception timing of null
reference to remove it, just removes redundant array construction.


stephentoub:
LGTM

PR #3523, opened 2015-09-29
Updated performance tests

ianhays:
Made some changes to the perf tests so that the data was more easily comparable and made some minor cleanliness updates while I was at it.

- PerfUtils is no longer static and is instantiated on a test-by-test basis
- Greatly increased the size of the testable area for a number of the tests
- Tweaked the collection and innerIteration sizes for a number of tests
- Fixed some namespace bugs

Still to do:
- Add perf tests that exist to "lock in" optimizations we have in place. For example, Trim is optimized to return the same string instance if nothing needs to be trimmed.
- Add Dictionary test that has a fixed-time GetHashCode/Equals type so that we can test 
- Make System.IO.Compression tests inherit from FileCleanupTestBase

@stephentoub

### DIFF
--- a/src/Common/tests/Performance/PerfUtils.cs
+++ b/src/Common/tests/Performance/PerfUtils.cs
@@ -8,30 +8,44 @@ using System.IO;
 namespace System
 {
     /// <summary>Static helper class for performance tests</summary>
-    public static class PerfUtils
+    public class PerfUtils
     {
-        private static Random rand = new Random(1234132);
+        private Random rand;
+
+        /// <summary>
+        /// Initializes a new PerfUtils object with the default random seed.
+        /// </summary>
+        public PerfUtils()
+        {
+            rand = new Random(1234132);
+        }
+
+        /// <summary>
+        /// Initializes a new PerfUtils object with the given seed. Use this if also
+        /// using MemberData that was created using PerfUtils to avoid possible collision
+        /// errors.
+        /// </summary>
+        public PerfUtils(int seed)
+        {
+            rand = new Random(seed);
+        }
 
         /// <summary>
         /// Helper method to create a string containing a number of random
         /// characters equal to the specified length
         /// </summary>
-        public static string CreateString(int length)
+        public string CreateString(int length)
         {
-            // Random is not thread safe and will cause intermittent thread deadlock if used without being locked
-            lock (rand)
-            {
-                byte[] bytes = new byte[length];
-                rand.NextBytes(bytes);
-                return System.Convert.ToBase64String(bytes);
-            }
+            byte[] bytes = new byte[length];
+            rand.NextBytes(bytes);
+            return System.Convert.ToBase64String(bytes);
         }
 
         /// <summary>Gets a test file full path that is associated with the call site.</summary>
         /// <param name="index">An optional index value to use as a suffix on the file name.  Typically a loop index.</param>
         /// <param name="memberName">The member name of the function calling this method.</param>
         /// <param name="lineNumber">The line number of the function calling this method.</param>
-        public static string GetTestFilePath(int? index = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
+        public string GetTestFilePath(int? index = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
         {
             return Path.Combine(Path.GetTempPath(), string.Format(
                 index.HasValue ? "{0}_{1}_{2}_{3}" : "{0}_{1}_{2}",
--- a/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
+++ b/src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
@@ -16,8 +16,8 @@ namespace System.Collections.Tests
             if (_testData == null)
             {
                 _testData = new List<object[]>();
-                _testData.Add(new object[] { CreateHashtable(100) });
-                _testData.Add(new object[] { CreateHashtable(1000) });
+                _testData.Add(new object[] { CreateHashtable(10000) });
+                _testData.Add(new object[] { CreateHashtable(1000000) });
             }
             return _testData;
         }
@@ -25,8 +25,9 @@ namespace System.Collections.Tests
         public static Hashtable CreateHashtable(int size)
         {
             Hashtable ht = new Hashtable();
+            PerfUtils utils = new PerfUtils();
             for (int i = 0; i < size; i++)
-                ht.Add(PerfUtils.CreateString(50), PerfUtils.CreateString(50));
+                ht.Add(utils.CreateString(50), utils.CreateString(50));
             return ht;
         }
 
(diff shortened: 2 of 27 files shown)


src/System.Collections.Specialized/tests/project.lock.json
```diff
@@ -1,5 +1,5 @@
 {
```
ianhays: Will remove this file in the rebase or PR feedback commit.

src/Common/tests/Performance/PerfUtils.cs
```diff
@@ -8,15 +8,25 @@
...
 
         /// <summary>
         /// Helper method to create a string containing a number of random
         /// characters equal to the specified length
         /// </summary>
-        public static string CreateString(int length)
+        public string CreateString(int length)
         {
             // Random is not thread safe and will cause intermittent thread deadlock if used without being locked
             lock (rand)
```
ianhays: This lock is no longer necessary now that PerfUtils is on a per-test basis

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
```diff
@@ -48,16 +52,20 @@ public void ctor()
         public void GetItem(Hashtable table)
         {
             object result;
-            string key = PerfUtils.CreateString(50);
+            PerfUtils utils = new PerfUtils(983452);
```
stephentoub: Why the special seed passed in?

src/System.Collections.Specialized/tests/project.lock.json
```diff
@@ -1,5 +1,5 @@
...
-  "locked": true,
+  "locked": false,
```
stephentoub: true?

src/System.Collections/tests/Performance/Perf.Dictionary.cs
```diff
@@ -51,14 +53,16 @@ public void Add(Dictionary<string, string> dict)
...
-                }
-                for (int i = 1; i <= 9; i++)
-                    Assert.True(dict.Remove("key" + i));
+                    for (int i = 0; i <= 20000; i++)
+                    {
+                        copyDict.Add("TestString1." + i, "string"); copyDict.Add("TestString2." + i, "string");
+                        copyDict.Add("TestString3." + i, "string"); copyDict.Add("TestString4." + i, "string");
+                        copyDict.Add("TestString5." + i, "string"); copyDict.Add("TestString6." + i, "string");
+                        copyDict.Add("TestString7." + i, "string"); copyDict.Add("TestString8." + i, "string");
+                        copyDict.Add("TestString9." + i, "string");
```
stephentoub: Each of these is now also boxing an integer, calling ToString on that boxed int, and doing a string.Concat.

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
ianhays: Otherwise it will use the same seed that is used in the generation of the collections used as MemberData which will cause the key being generated to already exist within the Hashtable

src/System.Collections.Specialized/tests/project.lock.json
ianhays: This lock file slipped through and will be removed when I next update this PR.

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
stephentoub: Ok, thanks. That's worth a comment.

src/System.Diagnostics.Process/tests/Performance/Perf.Process.cs
```diff
@@ -109,15 +180,23 @@ public void GetStandardOutput()
...
-                    proc.Kill();
+                    processes[i] = CreateProcess();
+                    processes[i].StartInfo.RedirectStandardOutput = true;
+                    processes[i].Start();
                 }
+
+                // Begin Testing
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < 500; i++)
+                        result = processes[i].StandardOutput;
```
stephentoub: I think the important thing here is not how fast we can get StandardOutput, but how fast we can read all of the data sent through standard output.

src/System.Runtime/tests/Performance/Perf.Guid.cs
```diff
@@ -25,13 +26,14 @@ public void ctor_str()
...
-                    new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
-                }
+                    for (int i = 0; i < 10000; i++)
+                    {
+                        new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff"); new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
+                        new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff"); new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
+                        new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff"); new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
+                        new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff"); new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
+                        new Guid("a8a110d5-fc49-43c5-bf46-802db8f843ff");
+                    }
```
stephentoub: Nit: rather than repeating the guid so many times, you might just have it in a const string local.

src/System.Collections/tests/Performance/Perf.Dictionary.cs
ianhays: Yeah, it's not great; I do this in a few other places as well. I could just make the key an integer and do a bit of arithmetic (key = 1 + i*9, key = 2 + i*9, etc.) on it to make sure there are no collisions. Still need to add some tests with a measured-time Equals/GetHashCode.

stephentoub:
Left some comments, but otherwise LGTM.

src/System.Diagnostics.Process/tests/Performance/Perf.Process.cs
ianhays: Update pushed; mind taking a look at the GetStandardOutput method? It essentially measures ReadToEnd() now, but it seems like it might be dependent on the execution time of the process and not be testing the correct thing.

src/System.Runtime/tests/Performance/Perf.Guid.cs
ianhays: done

src/System.Collections.NonGeneric/tests/Performance/Perf.HashTable.cs
```diff
@@ -69,14 +78,16 @@ public void Add(Hashtable table)
...
                 using (iteration.StartMeasurement())
                 {
-                    table.Add("key1", "value"); table.Add("key2", "value"); table.Add("key3", "value");
-                    table.Add("key4", "value"); table.Add("key5", "value"); table.Add("key6", "value");
-                    table.Add("key7", "value"); table.Add("key8", "value"); table.Add("key9", "value");
+                    for (int i = 0; i < 40000; i++)
+                    {
+                        tableCopy.Add(i * 10 + 1, "value"); tableCopy.Add(i * 10 + 2, "value"); tableCopy.Add(i * 10 + 3, "value");
+                        tableCopy.Add(i * 10 + 4, "value"); tableCopy.Add(i * 10 + 5, "value"); tableCopy.Add(i * 10 + 6, "value");
+                        tableCopy.Add(i * 10 + 7, "value"); tableCopy.Add(i * 10 + 8, "value"); tableCopy.Add(i * 10 + 9, "value");
```
stephentoub: Note that Hashtable works in terms of objects, so each of these Add calls is boxing its Int32 argument.  Don't know if that'll affect your perf results.

ianhays: Yeah.. that's expected. It was either boxed ints, string concatenation, or pre-building 360000 boxed ints/strings in a local array before the measurement starts and then add those within the loops. I may switch to the pre-built method later down the line if results are being skewed too much, but I think it is fine for now.

stephentoub: >  It was either boxed ints, string concatenation, or ...

Or making this a test of Add/Remove rather than just Add.

Issue #3362, opened 2015-09-22
System.Net.Security related Pending PR coments/TODO


rajansingh10:
1. Interop.OpenSsl TODO - See if SSL_CTX_Set_quite_shutdown can be used

vijaykota:
@rajansingh10, seems like a copy-paste error. There should be more unaddressed comments here

rajansingh10:
2. Trailer size requirement changes based on Protocol. Currently it's hard coded. (Interop.libssl_types.cs)
3. SSL3_RT_MAX_COMPRESSED_LENGTH will depend on whether compression is allowed or not. (Interop.libssl_types.cs)
4. AllocateSslContext - Set remote certificate options (Interop.OpenSsl.cs)
5. Disconnect -  check the SSL_shutdown error (Interop.OpenSsl.cs)
6. BioRead/BioWrite - Should bio retry, investigate & fix? (Interop.OpenSsl.cs)
7. SslCredKey - used a prime number here as it's a XOR. Figure out appropriate value. (_SslSessionsCache.cs)
8. GetRemoteCertificate - Fetch remoteCertificateStore, if applicable for unix. (CertModule.cs)
9. GetRequestCertificateAuthorities - populate issuers. (CertModule.cs)
10. EnsureStoreOpened - Implement the method. (CertModule.cs)
11. SafeFreeContextBufferChannelBinding - Implement the class (SecuritySafeHandles.cs)
12.  SslConnectionInfo -  map key sizes (SslConnectionInfo.cs)
13.  MapProtocol - Figure out correct value for default - (SslConnectionInfo.cs)
14. MapCipherAlgorithmType/MapExchangeAlgorithmType/MapHashAlgorithmType - Handle other values - (SslConnectionInfo.cs)
15. GetException - Implement the method. (SSPISecureChannelType.cs)
16. AcquireCredentialsHandle -  retVal is not getting released now, need to be fixed.  (SSPISecureChannelType.cs)
17. QueryContextChannelBinding & QueryContextIssuerList - Implement the method. (SSPISecureChannelType.cs)

rajansingh10:
17. Add error messages to *.resx files for localization.

rajansingh10:
18. Use SslContext as a class rather than struct & use a GCHandle rather than allocating native memory every time.

rajansingh10:
Evaluate if various SSL methods in GetSslMethod of Openssl.cs, can be cached one time & use always.

[...]

PR #3525, opened 2015-09-29
Updating SSLConnectionInfo with correct protocol version and Cipher algorithms

shrutigarg:
It includes 
a. mapping the Cipher / KeyExchange / DataHash algorithms mapping logic to the corresponding enums , In case for a few algorithms if enum is not already defined, it will return the numeric value representing that type in Openssl. (this behavior is similar to the Windows in which for unsupported enums we get the numeric value as it is).
 b. Protocol Mapping

### DIFF
--- a/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
@@ -184,15 +184,17 @@ internal static partial class Interop
             return libssl.SSL_get_peer_cert_chain(context);
         }
 
-        internal static libssl.SSL_CIPHER GetConnectionInfo(SafeSslHandle context)
+        internal static libssl.SSL_CIPHER GetConnectionInfo(SafeSslHandle sslHandle, out string protocolVersion)
         {
-            IntPtr cipherPtr = libssl.SSL_get_current_cipher(context);
+            IntPtr cipherPtr = libssl.SSL_get_current_cipher(sslHandle);
             var cipher = new libssl.SSL_CIPHER();
             if (IntPtr.Zero != cipherPtr)
             {
                 cipher = Marshal.PtrToStructure<libssl.SSL_CIPHER>(cipherPtr);
             }
 
+            IntPtr versionPtr = libssl.SSL_get_version(sslHandle);
+            protocolVersion = Marshal.PtrToStringAnsi(versionPtr);
             return cipher;
         }
 
--- a/src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
@@ -23,7 +23,7 @@ internal static partial class Interop
         internal static extern IntPtr TLSv1_1_method();
 
         [DllImport(Interop.Libraries.LibSsl)]
-        internal static extern IntPtr TLSv1_2_method();     
+        internal static extern IntPtr TLSv1_2_method();
 
         [DllImport(Interop.Libraries.LibSsl)]
         internal static extern IntPtr SSL_CTX_new(IntPtr meth);
@@ -122,5 +122,8 @@ internal static partial class Interop
 
             return handle;
         }
+
+        [DllImport(Interop.Libraries.LibSsl)]
+        internal static extern IntPtr SSL_get_version(SafeSslHandle ssl);
     }
 }
--- a/src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
+++ b/src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
@@ -76,11 +76,21 @@ namespace System.Net
             streamSizes = new StreamSizes(Interop.libssl.SslSizes.HEADER_SIZE, Interop.libssl.SslSizes.TRAILER_SIZE, Interop.libssl.SslSizes.SSL3_RT_MAX_PLAIN_LENGTH);
         }
 
-        public static void QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
+        public static int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
         {
+            string protocolVersion;
             connectionInfo = null;
-            Interop.libssl.SSL_CIPHER cipher = Interop.OpenSsl.GetConnectionInfo(securityContext.SslContext);
-            connectionInfo =  new SslConnectionInfo(cipher);
+            try
+            {
+                Interop.libssl.SSL_CIPHER cipher = Interop.OpenSsl.GetConnectionInfo(securityContext.SslContext, out protocolVersion);
+                connectionInfo =  new SslConnectionInfo(cipher, protocolVersion);
+               
+                return 0;
+            }
+            catch
+            {
+                return -1;
+            }
         }
 
         private static long GetOptions(SslProtocols protocols)
--- a/src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
@@ -101,5 +101,57 @@ internal static partial class Interop
         {
             SSL_ST_OK = 3
         }
+
+        internal enum CipherAlgorithm
+        {
+            SSL_DES = 1,
+            SSL_3DES = 2,
+            SSL_RC4 = 4,
+            SSL_RC2 = 8,
+            SSL_IDEA = 16,
+            SSL_eNULL = 32,
+            SSL_AES128 = 64,
+            SSL_AES256 = 128,
+            SSL_CAMELLIA128 = 256,
+            SSL_CAMELLIA256 = 512,
+            SSL_eGOST2814789CNT = 1024,
+            SSL_SEED = 2048,
+            SSL_AES128GCM = 4096,
+            SSL_AES256GCM = 8192
+          
+        }
+
+        internal enum KeyExchangeAlgorithm
+        {
+            SSL_kRSA = 1,
+            /* DH cert, RSA CA cert */
+            SSL_kDHr = 2,
+            /* DH cert, DSA CA cert */
+            SSL_kDHd = 4,
+            /* tmp DH key no DH cert */
+            SSL_kEDH = 8,
+            /* Kerberos5 key exchange */
+            SSL_kKRB5 = 16,
+            /* ECDH cert, RSA CA cert */
+            SSL_kECDHr = 32,
+            /* ECDH cert, ECDSA CA cert */
+            SSL_kECDHe = 64,
+            SSL_kEECDH = 128,
+            SSL_kPSK = 256,
+            SSL_kGOST = 512,
+            SSL_kSRP = 1024,
+        }
+
+        internal enum DataHashAlgorithm
+        {
+            SSL_MD5 = 1,
+            SSL_SHA1 = 2,
+            SSL_GOST94 = 4,
+            SSL_GOST89MAC = 8,
+            SSL_SHA256 = 16,
+            SSL_SHA384 = 32,
+            SSL_AEAD = 64
+        }
+
     }
 }
--- a/src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
+++ b/src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
@@ -7,6 +7,30 @@ namespace System.Net
 {
     internal class SslConnectionInfo
     {
+        //Windows constants for algorithms without matching enums in SslStream
+        //Taken from wincrypt.h 
+        internal const int SSL_SHA256 = 32780;
+        internal const int SSL_SHA384 = 32781;
+        internal const int SSL_ECDH = 43525;
+        internal const int SSL_ECDSA = 41475;
+
+        //Algorithm constants which are not present in wincrypt.h(not supported in Windows)
+        //These constants values are not conflicting with existing constants in wincrypt.h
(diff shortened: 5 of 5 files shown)


shrutigarg:
cc: @stephentoub, @bartonjs, @vijaykota, @rajansingh10, @pgavlin , @CIPop , @davidsh, @SidharthNabar

shrutigarg:
@dotnet-bot test this please

jasonwilliams200OK:
:bulb: presuming this fixes issue #3362, if you specify `Fix #3362` or better yet `Fixes #3362` as a separate line in the description (first post) of this PR, it will automatically close the issue as this PR gets merged. Here is an example PR: https://github.com/dotnet/corefx/pull/3417.
If there are multiple issues fixed by one PR, we can add multiple `Fixes #___` on separate lines in order to have them auto-closed on merge.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -236,6 +236,7 @@ internal static libssl.SSL_CIPHER GetConnectionInfo(IntPtr sslContextPtr)
                 cipher = Marshal.PtrToStructure<libssl.SSL_CIPHER>(cipherPtr);
             }
 
+            prtocolVersion = libssl.SSL_get_version(sslPtr);
```
stephentoub: Typo: prtocol => protocol

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -100,7 +100,10 @@ internal static partial class libssl
         [DllImport(Interop.Libraries.LibSsl)]
         internal static extern int BIO_ctrl_pending(IntPtr bio);
 
-        [DllImport(Interop.Libraries.LibSsl, EntryPoint = "SSL_CTX_set_quiet_shutdown")]
+        [DllImport(Interop.Libraries.LibSsl)]
         internal static extern void SSL_CTX_set_quiet_shutdown(IntPtr ssl, int mode);
+
+        [DllImport(Interop.Libraries.LibSsl, CharSet = CharSet.Ansi)]
+        internal static extern string SSL_get_version(IntPtr ssl);
```
stephentoub: ssl_get_version returns a const char ptr. By having the DllImport defined to return string, the marshaller is going to try to free the returned ptr... that's a problem. This should instead be defined to return an IntPtr; you can then use the Marshal class to get the string.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -15,30 +15,31 @@ internal class SslConnectionInfo
...
+                    return SslProtocols.Tls;
+                case "TLSv1.1":
                     return SslProtocols.Tls11;
-                case 4:
+                case "TLSv1.2":
                     return SslProtocols.Tls12;
                 default:
-                    // TODO (Issue #3362) Figure out correct value for this
-                    return SslProtocols.Tls;
+                    return SslProtocols.None;
```
stephentoub: Why it's the string-based version, which seems to involve extra P/Invokes, better than the previous int-based version?

bartonjs: I agree with @stephentoub: This would be better as an enum->enum mapping function, instead of a string->enum mapping function.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -54,16 +55,21 @@ private static CipherAlgorithmType MapCipherAlgorithmType(long encryption)
...
                 case 128:
                     return CipherAlgorithmType.Aes256;
+                case 16:
+                case 256:
+                case 512:
+                case 1024:
+                case 2048:
+                case 4096:
+                case 8192:
+                    return (CipherAlgorithmType)encryption;
```
bartonjs: * Why do all of these values just inject onto the enum?
* Please create an enum to name the case labels in this switch; I have no idea what any of these values are supposed to be.
* Since this is a library with a separate contract (the `ref` directory project) the implementation library can add new enum values without their names being public.  So if Windows equivalents of these 7 values can be found, they can be correctly populated (which would match the Windows code's blind cast).
    * Failing that, making some local consts for what the Windows equivalents would be is better than having competing direct-cast implementations.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -74,12 +80,19 @@ private static ExchangeAlgorithmType MapExchangeAlgorithmType(long mac)
...
+                case 8:
                     return ExchangeAlgorithmType.DiffieHellman;
+                case 16:
+                case 32:
+                case 64:
+                case 128:
+                case 256:
+                case 512:
+                case 1024:
+                    return (ExchangeAlgorithmType) mac;
```
bartonjs: Same as MapCipherAlgorithmType.  We need to not have two different unmapped values for the same thing.

For both of them, if OpenSSL exposes a concept Windows doesn't, we should find a way to update the enum (it's a mostly-non-breaking change, but the public names need to go through API review).  The only breaking-ness is that someone referencing CoreFX wouldn't be able to go back to Desktop with that identifier.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -92,8 +105,13 @@ private static HashAlgorithmType MapHashAlgorithmType(long mac)
                     return HashAlgorithmType.Md5;
                 case 2:
                     return HashAlgorithmType.Sha1;
+                case 4:
+                case 8:
+                case 16:
+                case 32:
+                case 64:
+                    return (HashAlgorithmType) mac;
```
bartonjs: Same as MapCipherAlgorithmType. We need to not have two different unmapped values for the same thing.

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -100,7 +100,10 @@ internal static partial class libssl
...
     }
 }
```
CIPop: Nit: missing new line at EOF.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
```diff
@@ -15,30 +15,31 @@ internal class SslConnectionInfo
...
         }
 
-        private static SslProtocols MapProtocol(long algorithm)
+        private SslProtocols MapProtocolVersion(string protocolVersion)
         {
-            switch (algorithm)
+            switch (protocolVersion)
             {
-                case 1:
+                case "SSLv2":
```
CIPop: Note: in Windows SSLv2 is prohibited OS-wide (although via Registry it can be brought back AFAIK).
The code looks good (we should support SSLv2 in case someone really needs to use it (insecurely) but, for security purposes, I'd like to understand if SSLv2 is, by default, disabled in OpenSSL.

CIPop:
+1 on all @bartonjs 's comments regarding the enums. 
We should also explore the option of adding the missing `CipherAlgorithmType`s to the contract and then back to Desktop.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
vijaykota: Yes. libssl needs to be explicitly rebuilt from sources enabling it. But I don't see what this comment has to do with the change referenced.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
vijaykota: Sounds like a biggish change since we are not familiar with this process. @shrutigarg, please file a separate issue for this and the review can be paused till we update this PR with the required changes

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
vijaykota: > if OpenSSL exposes a concept Windows doesn't

@shrutigarg can confirm this. I believe even on Windows, values outside the defined enum set are possible. Your point about updating the enums still stands but I wanted to clarify that it is not specific to OpenSSL

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
shrutigarg: will fix

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
shrutigarg: the Protocols returning from  get_version() are in string format which are getting mapped here directly,
Previously we were relying on the SSL_CIPHER->algoVersion for protocols (represented in int) which were not returning all protocols version correctly .
So moved to this.

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
shrutigarg: There are few algorithms that Windows/Unix can support but are not listed in ENUMs. (ie. ECDH, SHA256 and many more) 
In windows also some number is returned representing that value , Same thing we followed here ,

bartonjs: > In windows also some number is returned representing that value , Same thing we followed here , 

Yes, that's the problem.  You need to return the SAME "some number" for the same concept.

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
shrutigarg: resolved

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
shrutigarg: fixed

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
shrutigarg: fixed

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
shrutigarg: Created the enums for the case label values.
As discussed ,local enums are not present in Security dll so cannot modify it . Going ahead with the constant values for unsupported enums

src/System.Net.Security/src/System/Net/Unix/SslConnectionInfo.cs
shrutigarg: fixed

shrutigarg:
@bartonjs @stephentoub @CIPop @vijaykota 
I have incorporated the comments and after offline discussions, going ahead with returning constant numbers for the algorithms without matching enums.


PS: by mistake two commits got squashed.

bartonjs:
Structurally looks reasonable to me, but it would be good for @CIPop to sign off on the new consts (and the belief that they aren't likely to collide with Windows)

CIPop:
/cc @SidharthNabar

> Structurally looks reasonable to me, but it would be good for @CIPop to sign off on the new consts (and the belief that they aren't likely to collide with Windows)

To prove this, we'll need to see tests that execute against Desktop, CoreFX on Windows and CoreFX on *NIX. These are not present in the PR and so I can't validate any app-compat with Windows.

In the future we'll need to ensure (through tests) that the PAL does proper adaptation and that the values returned at the public API are identical with the ones .Net Desktop currently returns even for the ones that don't have enum names.

shrutigarg:
>In the future we'll need to ensure (through tests) that the PAL does proper adaptation and that the values returned at the public API are identical with the ones .Net Desktop currently returns even for the ones that don't have enum names

Manually i tried this  by setting up the Openssl server to support only that cipher( which needs to be tested) which can become the shared cipher with the client..  We can add the similar testcases to verify the cipher.. 
Opened issue: https://github.com/dotnet/corefx/issues/3832

shrutigarg:
Fixed the merge issue .

src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
```diff
@@ -76,11 +76,21 @@ public static void QueryContextStreamSizes(SafeDeleteContext securityContext, ou
             streamSizes = new StreamSizes(Interop.libssl.SslSizes.HEADER_SIZE, Interop.libssl.SslSizes.TRAILER_SIZE, Interop.libssl.SslSizes.SSL3_RT_MAX_PLAIN_LENGTH);
         }
 
-        public static void QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
+        public static int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
```
CIPop: @shrutigarg @vijaykota @bartonjs The PAL signature cannot change except if the common code is changed as well. 
The assumption that the return code is actually inspected is wrong: the Windows signature is still void and the error is actually ignored which might be a sign of a security issue.

vijaykota:
@shrutigarg PTAL. Looks like a bad rebase to me

src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
shrutigarg: Will Fix this.

PR #3527, opened 2015-09-29
Remove some redundancies from System.Linq.Queryable

JonHanna:
Don't cache Func<T> produced in field in EnumerableExecutor<T> as it's never used twice.

Conversely store local values of some repeated method/property calls.

Don't create array for GetStaticMethods as no calling code benefits from it being an array.

Should probably be squashed into a single commit before merge, but may be easier to review separately.

### DIFF
--- a/src/System.Linq.Queryable/src/System/Linq/EnumerableExecutor.cs
+++ b/src/System.Linq.Queryable/src/System/Linq/EnumerableExecutor.cs
@@ -22,7 +22,6 @@ namespace System.Linq
     public class EnumerableExecutor<T> : EnumerableExecutor
     {
         private Expression _expression;
-        private Func<T> _func;
 
         // Must remain public for Silverlight
         public EnumerableExecutor(Expression expression)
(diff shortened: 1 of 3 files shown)


JonHanna:

JonHanna:

stephentoub:
Other than the squashing, LGTM.

cc: @VSadov

JonHanna:
Squashed down now.

VSadov:
I wonder why the func is not reused. There was clearly an intent to keep it, since compiling a new one is expensive. Could it be a bug somewhere else that we end up not reusing the func?

JonHanna:
@VSadov considered in isolation it makes sense to cache the Func for precisely that reason. The only two uses of the class though only call `Execute` or `ExecuteBoxed` once each in the instance's lifetime, so while the memoisation is a sensible optimisation for the class, it gains nothing as actually used.

VSadov:
@JonHanna - that seems unfortunate. EnumerableExecutor is just as shortlived, so caching something in executor gains nothing.

PR #3528, opened 2015-09-29
Using SafeHandles instead of IntPtrs in Interop.libssl

vijaykota:
Changed the Unix code to use SafeHandles

cc: @stephentoub @bartonjs @rajansingh10 @shrutigarg

### DIFF
--- a/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
@@ -8,153 +8,106 @@ using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Win32.SafeHandles;
 
+using SafeSslHandle = Interop.libssl.SafeSslHandle;
+
 internal static partial class Interop
 {
     internal static class OpenSsl
     {
-        #region structures
-        [StructLayout(LayoutKind.Sequential)]
-        private struct SslContext
-        {         
-            internal IntPtr sslPtr;
-            internal IntPtr readBioPtr;
-            internal IntPtr writeBioPtr;
-            internal bool isServer;
-        }
-        #endregion
-
         #region internal methods
 
-
         //TODO (Issue #3362) Set remote certificate options
-        internal static IntPtr AllocateSslContext(long options, SafeX509Handle certHandle, SafeEvpPKeyHandle certKeyHandle, bool isServer, bool remoteCertRequired)        
+        internal static SafeSslHandle AllocateSslContext(long options, SafeX509Handle certHandle, SafeEvpPKeyHandle certKeyHandle, bool isServer, bool remoteCertRequired)
         {
-            SslContext sslContext = new SslContext
+            SafeSslHandle context = null;
+
+            IntPtr method = GetSslMethod(isServer, options);
+
+            using (libssl.SafeSslContextHandle innerContext = new libssl.SafeSslContextHandle(method))
             {
-                isServer = isServer,
-            };
-
-            try
-            {
-                IntPtr method = GetSslMethod(isServer, options);
-
-                IntPtr contextPtr = libssl.SSL_CTX_new(method);
-
-                if (IntPtr.Zero == contextPtr)
+                if (innerContext.IsInvalid)
                 {
                     throw CreateSslException("Failed to allocate SSL/TLS context");
                 }
 
-                libssl.SSL_CTX_ctrl(contextPtr, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
+                libssl.SSL_CTX_ctrl(innerContext, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
 
-                libssl.SSL_CTX_set_quiet_shutdown(contextPtr, 1);
+                libssl.SSL_CTX_set_quiet_shutdown(innerContext, 1);
 
                 if (certHandle != null && certKeyHandle != null)
                 {
-                    SetSslCertificate(contextPtr, certHandle, certKeyHandle);
+                    SetSslCertificate(innerContext, certHandle, certKeyHandle);
                 }
 
-                sslContext.sslPtr = libssl.SSL_new(contextPtr);
-
-                libssl.SSL_CTX_free(contextPtr);
-
-                if (IntPtr.Zero == sslContext.sslPtr)
+                context = SafeSslHandle.Create(innerContext, isServer);
+                Debug.Assert(context != null, "Expected non-null return value from SafeSslHandle.Create");
+                if (context.IsInvalid)
                 {
-                    throw CreateSslException("Failed to create SSSL object from SSL context");
+                    context.Dispose();
+                    throw CreateSslException("Failed to create SSL object from SSL context");
                 }
-
-                IntPtr memMethod = libcrypto.BIO_s_mem();
-
-                if (IntPtr.Zero == memMethod)
-                {
-                    throw CreateSslException("Failed to return memory BIO method function");
-                }
-
-                sslContext.readBioPtr = libssl.BIO_new(memMethod);
-                sslContext.writeBioPtr = libssl.BIO_new(memMethod);
-
-                if ((IntPtr.Zero == sslContext.readBioPtr) || (IntPtr.Zero == sslContext.writeBioPtr))
-                {
-                    FreeBio(sslContext);
-                    throw CreateSslException("Failed to retun new BIO for a given method type");
-                }
-
-                if (isServer)
-                {
-                    libssl.SSL_set_accept_state(sslContext.sslPtr);
-                }
-                else
-                {
-                    libssl.SSL_set_connect_state(sslContext.sslPtr);
-                }
-
-                libssl.SSL_set_bio(sslContext.sslPtr, sslContext.readBioPtr, sslContext.writeBioPtr);
-            }
-            catch
-            {
-                Disconnect(sslContext.sslPtr);
-                throw;
             }
 
-            IntPtr sslContextPtr = Marshal.AllocHGlobal(Marshal.SizeOf<SslContext>());
-            Marshal.StructureToPtr(sslContext, sslContextPtr, false);
-            return sslContextPtr;
+            return context;
         }
 
-        internal static bool DoSslHandshake(IntPtr sslContextPtr, IntPtr recvPtr, int recvCount, out IntPtr sendPtr, out int sendCount)
+        internal static bool DoSslHandshake(SafeSslHandle context, IntPtr recvPtr, int recvCount, out IntPtr sendPtr, out int sendCount)
         {
             sendPtr = IntPtr.Zero;
             sendCount = 0;
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-            bool isServer = context.isServer;
             if ((IntPtr.Zero != recvPtr) && (recvCount > 0))
             {
-                BioWrite(context.readBioPtr, recvPtr, recvCount);
+                BioWrite(context.InputBio, recvPtr, recvCount);
             }
 
-            int retVal = libssl.SSL_do_handshake(context.sslPtr);
-
-            if (retVal != 1)
+            libssl.SslErrorCode error;
+            int retVal = libssl.SSL_do_handshake(context);
+            if (retVal == 1)
             {
-                libssl.SslErrorCode error = GetSslError(context.sslPtr, retVal);
+                // In case of a client, this indicates successful handshake completion
+                if (!context.IsServer)
+                {
+                    return true;
+                }
+            }
+            else
+            {
+                error = GetSslError(context, retVal);
 
                 if ((retVal != -1) || (error != libssl.SslErrorCode.SSL_ERROR_WANT_READ))
                 {
-                    throw CreateSslException(context.sslPtr, "SSL Handshake failed: ", retVal);
+                    throw CreateSslException(context, "SSL Handshake failed: ", retVal);
                 }
             }
 
-            sendCount = libssl.BIO_ctrl_pending(context.writeBioPtr);
+            sendCount = libssl.BIO_ctrl_pending(context.OutputBio);
 
             if (sendCount > 0)
             {
                 sendPtr = Marshal.AllocHGlobal(sendCount);
-                sendCount = BioRead(context.writeBioPtr, sendPtr, sendCount);
+                sendCount = BioRead(context.OutputBio, sendPtr, sendCount);
                 if (sendCount <= 0)
                 {
                     int errorCode = sendCount;
                     Marshal.FreeHGlobal(sendPtr);
                     sendPtr = IntPtr.Zero;
-                    sendCount = 0;
-                    throw CreateSslException(context.sslPtr, "Read Bio failed: ", errorCode);                   
+                    sendCount = 0;                  
+                    throw CreateSslException(context, "Read Bio failed: ", errorCode);                   
                 }
             }
         
-            return ((libssl.SSL_state(context.sslPtr) == (int)libssl.SslState.SSL_ST_OK));
+            return ((libssl.SSL_state(context) == (int)libssl.SslState.SSL_ST_OK));
 
         }
 
-        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int count, int bufferCapacity, out libssl.SslErrorCode errorCode)
+        internal static int Encrypt(SafeSslHandle context, IntPtr buffer, int offset, int count, int bufferCapacity, out libssl.SslErrorCode errorCode)
         {
             errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
 
-            SslContext context = Marshal.PtrToStructure<SslContext>(handlePtr);
-
-            var retVal = libssl.SSL_write(context.sslPtr, new IntPtr(buffer.ToInt64() + offset), count);
-
+            int retVal = libssl.SSL_write(context, new IntPtr(buffer.ToInt64() + offset), count);
             if (retVal != count)
             {
-                errorCode = GetSslError(context.sslPtr, retVal);
+                errorCode = GetSslError(context, retVal);
                 retVal = 0;
 
                 switch (errorCode)
@@ -170,14 +123,14 @@ internal static partial class Interop
             }
             else
             {
-                int capacityNeeded = libssl.BIO_ctrl_pending(context.writeBioPtr);      
+                int capacityNeeded = libssl.BIO_ctrl_pending(context.OutputBio);      
 
                 if (capacityNeeded > bufferCapacity)
                 {
                     throw CreateSslException("OpenSsl::Encrypt capacity needed is more than buffer capacity. capacityNeeded = " + capacityNeeded + "," + "bufferCapacity = " + bufferCapacity);
                 }             
 
-                retVal = BioRead(context.writeBioPtr, buffer, capacityNeeded);
+                retVal = BioRead(context.OutputBio, buffer, capacityNeeded);
 
                 if (retVal < 0)
                 {
@@ -188,17 +141,15 @@ internal static partial class Interop
             return retVal;
         }
 
-        internal static int Decrypt(IntPtr sslContextPtr, IntPtr outBufferPtr, int count, out libssl.SslErrorCode errorCode)
+        internal static int Decrypt(SafeSslHandle context, IntPtr outBufferPtr, int count, out libssl.SslErrorCode errorCode)
         {
             errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
 
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-
-            int retVal = BioWrite(context.readBioPtr, outBufferPtr, count);
+            int retVal = BioWrite(context.InputBio, outBufferPtr, count);
 
             if (retVal == count)
             {
-                retVal = libssl.SSL_read(context.sslPtr, outBufferPtr, retVal);
+                retVal = libssl.SSL_read(context, outBufferPtr, retVal);
 
                 if (retVal > 0)
                 {
@@ -208,7 +159,7 @@ internal static partial class Interop
 
             if (retVal != count)
             {
-                errorCode = GetSslError(context.sslPtr, retVal);
+                errorCode = GetSslError(context, retVal);
                 retVal = 0;
 
                 switch (errorCode)
@@ -219,7 +170,7 @@ internal static partial class Interop
 
                     case libssl.SslErrorCode.SSL_ERROR_WANT_READ:
                         // update error code to renegotiate if renegotiate is pending, otherwise make it SSL_ERROR_WANT_READ
-                        errorCode = libssl.SSL_renegotiate_pending(context.sslPtr) == 1 ?
+                        errorCode = libssl.SSL_renegotiate_pending(context) == 1 ?
                                     libssl.SslErrorCode.SSL_ERROR_RENEGOTIATE :
                                     libssl.SslErrorCode.SSL_ERROR_WANT_READ;
                         break;
@@ -232,27 +183,19 @@ internal static partial class Interop
             return retVal;
         }
 
-        internal static IntPtr GetPeerCertificate(IntPtr sslContextPtr)
+        internal static SafeX509Handle GetPeerCertificate(SafeSslHandle context)
         {
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-            IntPtr sslPtr = context.sslPtr;
-            IntPtr certPtr = libssl.SSL_get_peer_certificate(sslPtr);
-            return certPtr;
+            return libssl.SSL_get_peer_certificate(context);
         }
 
-        internal static SafeSharedX509StackHandle GetPeerCertificateChain(IntPtr sslContextPtr)
+        internal static SafeSharedX509StackHandle GetPeerCertificateChain(SafeSslHandle context)
         {
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-            IntPtr sslPtr = context.sslPtr;
-
-            return libssl.SSL_get_peer_cert_chain(sslPtr);
+            return libssl.SSL_get_peer_cert_chain(context);
         }
 
-        internal static libssl.SSL_CIPHER GetConnectionInfo(IntPtr sslContextPtr)
+        internal static libssl.SSL_CIPHER GetConnectionInfo(SafeSslHandle context)
         {
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-            IntPtr sslPtr = context.sslPtr;
-            IntPtr cipherPtr = libssl.SSL_get_current_cipher(sslPtr);
+            IntPtr cipherPtr = libssl.SSL_get_current_cipher(context);
             var cipher = new libssl.SSL_CIPHER();
             if (IntPtr.Zero != cipherPtr)
             {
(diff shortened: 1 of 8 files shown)


src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,111 +8,69 @@
...
-                libssl.SSL_CTX_set_quiet_shutdown(contextPtr, 1);
-
-                if (certHandle != null && certKeyHandle != null)
+                using (libssl.SafeSslContextHandle sslContext = new libssl.SafeSslContextHandle(method))
                 {
-                    SetSslCertificate(contextPtr, certHandle, certKeyHandle);
-                }
-
-                sslContext.sslPtr = libssl.SSL_new(contextPtr);
+                    if ((null == sslContext) || sslContext.IsInvalid)
```
bartonjs: If we've moved to a world where C#'s `new` can return `null` we should be sad :)  (null check is redundant)

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,111 +8,69 @@
...
-                {
-                    throw CreateSslException("Failed to allocate SSL/TLS context");
-                }
-
-                libssl.SSL_CTX_ctrl(contextPtr, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
-
-                libssl.SSL_CTX_set_quiet_shutdown(contextPtr, 1);
-
-                if (certHandle != null && certKeyHandle != null)
+                using (libssl.SafeSslContextHandle sslContext = new libssl.SafeSslContextHandle(method))
```
bartonjs: Personally I find it a bit hard to read with `context` and `sslContext` and they're different types.  Perhaps `context` could be `ssl`, or `sslHandle`, or something within this method?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,111 +8,69 @@
...
             }
             catch
             {
-                Disconnect(sslContext.sslPtr);
-                throw;
+                if (null != context)
+                {
+                    context.Dispose();
+                    context = null;
+                }
```
bartonjs: This used to throw (continue) the exception, now it's eating it.  If eating it is intentional (which I'm sure will get lots of disagreement) that should have a comment explaining why.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,107 @@
...
+                    return;
+                }
+
+                // After SSL_set_bio, the BIO handles are owned by SSL pointer
+                // and are automatically freed by SSL_free. To prevent a double
+                // free, we need to keep the ref counts bumped up till SSL_free
+                bool gotReadRef = false, gotWriteRef = false;
+                _readBio.DangerousAddRef(ref gotReadRef);
+                _writeBio.DangerousAddRef(ref gotWriteRef);
+                SSL_set_bio(tempHandle, _readBio.DangerousGetHandle(), _writeBio.DangerousGetHandle());
```
bartonjs: If we were to get a ThreadAbortException between the DangerousAddRefs and the setting of the handle field the BIOs would never get freed.

```C#
bool gotReadRef = false, gotWriteRef = false;

try
{
    _readBio.DangerousAddRef(ref gotReadRef);
    ...
    handle = tempHandle;
}
catch
{
    if (gotWriteRef)
    {
        _writeBio.DangerousRelease();
    }

    ...

    throw;
}
```

Though you'd have better tracking, I think, if you made this ctor be a static method, and make SSL_new return a `SafeSslHandle`.  Then there's not also the problem of a ThreadAbortException needing to handle the situation where `tempHandle` was assigned but it never made it to `handle`.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            public override bool IsInvalid
+            {
+                get { return handle == IntPtr.Zero; }
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                SSL_free(handle);
+                _readBio.SetHandleAsInvalid();  // BIO got freed in SSL_free
```
bartonjs: Instead of doing SetHandleAsInvalid here, and persisting the object references, you should be able to do SetHandleAsInvalid and Dispose the object as a local variable in the allocator.  Then the managed overhead of the object tracking can be eliminated after the pointer is transferred.

src/Common/src/Interop/Unix/libssl/Interop.libssl.cs
```diff
@@ -56,51 +56,48 @@ internal static partial class libssl
...
-        internal static extern int BIO_read(IntPtr bio, IntPtr buf, int num);
+        internal static extern int BIO_read(SafeBioHandle bio, IntPtr buf, int num);
 
         [DllImport(Interop.Libraries.LibSsl)]
-        internal static extern int BIO_write(IntPtr bio, IntPtr buf, int num);
+        internal static extern int BIO_write(SafeBioHandle bio, IntPtr buf, int num);
 
         [DllImport(Interop.Libraries.LibSsl)]
-        internal static extern IntPtr SSL_get_peer_certificate(IntPtr ssl);
+        internal static extern SafeX509Handle SSL_get_peer_certificate(SafeSslHandle ssl);
```
bartonjs: I checked (again), and this method does require an X509_free call to be applied to the return value, so I'm very happy this is a SafeHandle'd method now :).

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -19,24 +19,31 @@ internal sealed class SafeFreeCertContext : DebugSafeHandle
...
     {
 #endif
-        public SafeFreeCertContext(IntPtr handle) : base(handle, true)
+        private readonly SafeX509Handle _certificate;
+
+        public SafeFreeCertContext(SafeX509Handle certificate) : base(IntPtr.Zero, true)
         {
+            if ((null != certificate) && !certificate.IsInvalid)
+            {
+                _certificate = Interop.libcrypto.X509_dup(certificate);
```
bartonjs: I don't think you need to dup it.  `SSL_get_peer_certificate` already bumps the refcount on the certificate.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: So SafeSslHandle should throw an exception. What would an appropriate exception type for this be?

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Thanks.. I knew there would be a better way to do this.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: I didn't understand this fully. In a static method, the try/catch would not be needed since ThreadAbortEx is not possible, is that right?

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: My opinion is that S.N.Security PAL should not rely on how Interop.OpenSsl implements its methods. Unless you feel strongly about this, I would like to leave this as "won't fix"

vijaykota:
@bartonjs, apologies for squashing the commit - I didn't realize you were reviewing it. The changes in  https://github.com/vijaykota/corefx/commit/04a1c741639b471222f7985a22ffd6c6f36fb379 vs earlier commits are:
1. Got rid of the DangerousAddRefs() in SSPISecureChannelType.cs
2. Fixed a bug in SecureChannel.cs

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: Another option is persisting the returned SafeX509Handle by AddRef'ing it

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Sorry.. this was a mistake. Will fix this

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Agreed :) I dug deep but came up with nothing! Will try again to come up with some other names

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Valid point.. will fix this

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
bartonjs: > do you mean SetHandleAsInvalid after the ssl_set_bio

Yep.

> But a valid handle is required in InputBio and OutputBio

Calling `SSL_set_bio` transferred the ownership into the SSL*.  `SetHandleAsInvalid` doesn't free the resource, it tells the CLR to _not_ free the resource on `Finalize` or `Dispose`.  Then, the subsequent call to `Dispose` would suppress finalization, and let the managed object be freed.  The native resource would still be valid.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
bartonjs: You very much need to know what OpenSSL did/does.  I'm pretty sure the code as-is will leave the native certificate at the end with a tracking reference count of 1, so it won't be freed.  This means every SslStream leaks a native resource to the tune of an X509* and all of its references.

bartonjs: Actually, no, since it's a SafeHandle returned now it's... probably fine.

* SSL_get_peer_certificate bumps the refcount to 2, that is encapsulated in the returned SafeHandle.
* This code calls X509_dup, bumping the refcount to 3.
* The originally returned SafeHandle gets Finalized (since you're not Disposing it), decrementing the refcount to 2.
* The output of X509_dup is eventually Disposed (or FInalized), decrementing the refcount to 1.
* The SafeSslHandle is eventually Disposed (or Finalized), decrementing the refcount to 0 and freeing the native resource.

So really the problem is that you're sending a SafeHandle off to the finalizer queue on every call.

The point about needing to know what OpenSSL does still stands, though; if the p/invoke target return type was IntPtr instead of SafeHandle there'd be a decrement missing from that series.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
bartonjs: Oh... I think I understand the problem; it's that you need to write into / read from them still.

Right.  Nevermind :)

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
bartonjs: > In a static method, the try/catch would not be needed since ThreadAbortEx is not possible, is that right?

I don't know of any reason why a static method would be immune to a ThreadAbortException.  The example on [MSDN](https://msdn.microsoft.com/en-us/library/system.threading.threadabortexception%28v=vs.110%29.aspx) even uses only static methods for its demonstration.  The only thing that I know of that's immune is a Constrained Execution Region, which we don't have in CoreFX/CoreCLR.

That said, it seems like we might not have ThreadAbortException in CoreFX/CoreCLR (@janvorli or @stephentoub am I missing something here?).

> if SSL_new returns a SafeSslHandle, ctor signature would change, correct?

Yes, the constructor would likely go to `private SafeSslHandle()`.

> I don't see a clear benefit.

Getting native resources tracked by a SafeHandle as early as possible is always good.  The window from the time where you get the IntPtr to the time that it's properly tracked by a SafeHandle is a window for leaking native resources. The more code that can go between them, the more chances of the leak.  By having the Marshaller do the SafeHandle wrapping during the P/Invoke return you're about as immune to leaking as you can get.

stephentoub: > it seems like we might not have ThreadAbortException in CoreFX/CoreCLR (@janvorli or @stephentoub am I missing something here?).

Correct. We don't worry about hardening code against TAEs.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Renamed the SSL_CTX to innerContext. Not sure if it helps though.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Thanks, @stephentoub and @bartonjs. Sorry it took me a long time to understand that p/invoke can work with a <b><I>private</I></b> ctor. The benefit is blindingly clear now :)

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: I changed the "contract" of SafeFreeCertContext to own the SafeX509Handle passed to it. So instead of X509_dup, it now does an AddRef and also eventually Disposes it.

> The point about needing to know what OpenSSL does still stands, though

Understood. I should qualify my earlier statement as "caller needs to know that OpenSsl (the class) yields ownership of the returned SafeX509Handle but doesn't need to know how it was constructed:

vijaykota:
@dotnet-bot, test this please

stephentoub:
@vijaykota, this will need to be rebased due to other changes that have come into the repo.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,111 +8,63 @@
...
-                {
-                    libssl.SSL_set_connect_state(sslContext.sslPtr);
-                }
-
-                libssl.SSL_set_bio(sslContext.sslPtr, sslContext.readBioPtr, sslContext.writeBioPtr);
+                context = SafeSslHandle.Create(innerContext, isServer);
             }
-            catch
+
+            if ((null != context) && !context.IsInvalid)
```
stephentoub: In what situation will Create return null?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,111 +8,63 @@
...
         }
 
-        internal static bool DoSslHandshake(IntPtr sslContextPtr, IntPtr recvPtr, int recvCount, out IntPtr sendPtr, out int sendCount)
+        internal static bool DoSslHandshake(SafeSslHandle context, IntPtr recvPtr, int recvCount, out IntPtr sendPtr, out int sendCount)
         {
             sendPtr = IntPtr.Zero;
             sendCount = 0;
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-            bool isServer = context.isServer;
+            bool isServer = context.IsServer;
```
stephentoub: Nit: is the local actually needed?  It's used in just one place, which could just be:
```C#
if (retVal == 1 && context.IsServer)
```

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -122,43 +74,41 @@ internal static bool DoSslHandshake(IntPtr sslContextPtr, IntPtr recvPtr, int re
...
             {
-                error = GetSslError(context.sslPtr, retVal);
+                error = GetSslError(context, retVal);
 
                 if ((retVal != -1) || (error != libssl.SslErrorCode.SSL_ERROR_WANT_READ))
                 {
-                    throw CreateSslException(context.sslPtr, "SSL Handshake failed: ", retVal);
+                    throw CreateSslException(context, "SSL Handshake failed: ", retVal);
                 }
             }
```
stephentoub: Rather than checking retVal twice, could you just do:
```C#
int retVal = libssl.SSL_do_handshake(context);
if (retVal == 1)
{
    if (!context.IsServer)
        return true;
}
else
{
    int error = GetSslError(context, retVal);
    ... // etc.
}
```
?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -322,25 +241,23 @@ private static IntPtr GetSslMethod(bool isServer, long options)
             return method;
         }
 
-        private static void Disconnect(IntPtr sslPtr)
+        private static void Disconnect(SafeSslHandle context)
         {
-            if (IntPtr.Zero != sslPtr)
+            if ((null != context) && !context.IsInvalid)
```
stephentoub: Will context ever be null?

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        internal sealed class SafeSslHandle : SafeHandle
+        {
+            private SafeBioHandle _readBio;
+            private SafeBioHandle _writeBio;
+            private bool _isServer;
+
+            public bool IsServer
+            {
+                get {return _isServer;}
```
stephentoub: Nit: spacing

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            public static SafeSslHandle Create(SafeSslContextHandle context, bool isServer)
+            {
+                IntPtr memMethod = libcrypto.BIO_s_mem();
+                SafeBioHandle readBio = libcrypto.BIO_new(memMethod);
+                if (readBio.IsInvalid)
+                {
+                    return new SafeSslHandle();
+                }
+                SafeBioHandle writeBio = libcrypto.BIO_new(memMethod);
```
stephentoub: Nit: please add a blank line above this

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,126 @@
...
+                    return handle;
+                }
+                handle._isServer = isServer;
+
+                // After SSL_set_bio, the BIO handles are owned by SSL pointer
+                // and are automatically freed by SSL_free. To prevent a double
+                // free, we need to keep the ref counts bumped up till SSL_free
+                bool gotRef = false;
+                readBio.DangerousAddRef(ref gotRef);
+                Debug.Assert(gotRef, "Unexpected failure in AddRef of read Bio");
```
stephentoub: DangerousAddRef throws if it can't get the reference; this Assert is unnecessary and will never fail.  Same for the one below.

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
```diff
@@ -0,0 +1,126 @@
...
+                handle._readBio = readBio;
+                gotRef = false;
+                writeBio.DangerousAddRef(ref gotRef);
```
stephentoub: If this throws, we're going to leak the ref on readBio.

stephentoub: That said, I see now that we're just allocating these above, so no one else has a reference to these, so no one else could be disposing of them, so these will always succeed.  Ok.

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -19,24 +19,32 @@ internal sealed class SafeFreeCertContext : DebugSafeHandle
     internal sealed class SafeFreeCertContext : SafeHandle
     {
 #endif
-        public SafeFreeCertContext(IntPtr handle) : base(handle, true)
+        private readonly SafeX509Handle _certificate;
+
+        public SafeFreeCertContext(SafeX509Handle certificate) : base(IntPtr.Zero, true)
         {
+            if ((null != certificate) && !certificate.IsInvalid)
```
stephentoub: Will certificate ever be null?  If so, isn't that an error?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Sorry.. it was due to an earlier avatar of Create which was returning null. Fixed this now

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Should never be. Changed this to an Assert up the call stack

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Done

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Done

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Thanks.. code looks cleaner now after removing the Asserts

src/Common/src/Interop/Unix/libssl/Interop.SafeSslHandle.cs
vijaykota: Added an explicit catch. I think there was a leak because once handle.ReleaseHandle is executed, the readBio handle would be set to zero

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: When server queries for the client certificate, it may be null and that is ok. Added a comment

vijaykota:
@dotnet-bot, test this please

stephentoub:
@dotnet-bot test this please

vijaykota:
Sorry.. had to rebase after https://github.com/dotnet/corefx/pull/3638 Hitting build failures that don't seem to be related to this PR

vijaykota:
Sorry.. had to rebase after https://github.com/dotnet/corefx/pull/3638 I am not able to figure out the build failures. They seem to be connected to this PR

vijaykota:
@dotnet-bot, test this please

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,153 +8,106 @@
...
-        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int count, int bufferCapacity, out libssl.SslErrorCode errorCode)
+        internal static int Encrypt(SafeSslHandle context, IntPtr buffer, int offset, int count, int bufferCapacity, out libssl.SslErrorCode errorCode)
         {
             errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
 
-            SslContext context = Marshal.PtrToStructure<SslContext>(handlePtr);
-
-            var retVal = libssl.SSL_write(context.sslPtr, new IntPtr(buffer.ToInt64() + offset), count);
-
+            var retVal = libssl.SSL_write(context, new IntPtr(buffer.ToInt64() + offset), count);
```
stephentoub: This should use int rather than var, assuming that's what the type is.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -188,17 +141,15 @@ internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int cou
...
 
-        internal static int Decrypt(IntPtr sslContextPtr, IntPtr outBufferPtr, int count, out libssl.SslErrorCode errorCode)
+        internal static int Decrypt(SafeSslHandle context, IntPtr outBufferPtr, int count, out libssl.SslErrorCode errorCode)
         {
             errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
 
-            SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
-
-            int retVal = BioWrite(context.readBioPtr, outBufferPtr, count);
+            var retVal = BioWrite(context.InputBio, outBufferPtr, count);
```
stephentoub: int rather than var

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
```diff
@@ -19,24 +19,34 @@ internal sealed class SafeFreeCertContext : DebugSafeHandle
...
+
+        public SafeFreeCertContext(SafeX509Handle certificate) : base(IntPtr.Zero, true)
         {
+            // In certain scenarios (eg. server querying for a client cert), the
+            // input certificate may be invalid and this is OK
+            if ((null != certificate) && !certificate.IsInvalid)
+            {
+                bool gotRef = false;
+                certificate.DangerousAddRef(ref gotRef);
+                Debug.Assert(gotRef, "Unexpected failure in AddRef of certificate");
```
stephentoub: Unnecessary assert

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -8,153 +8,106 @@
...
-            try
-            {
-                IntPtr method = GetSslMethod(isServer, options);
+            SafeSslHandle context = null;
 
-                IntPtr contextPtr = libssl.SSL_CTX_new(method);
+            IntPtr method = GetSslMethod(isServer, options);
 
-                if (IntPtr.Zero == contextPtr)
+            using (libssl.SafeSslContextHandle innerContext = new libssl.SafeSslContextHandle(method))
```
stephentoub: This will Dispose of innerContext at the end of the using block.  That's ok?  I see it's being passed to methods like SafeSslHandle.Create... it's not needed after those calls?  Or they're AddRef'ing to keep it alive?

stephentoub:
Left a few more comments.  Feel free to merge after addressing them.  You might also want to squash.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: The SSL_CTX object (inside SafeSslContextHandle) is not required after SSL_new(). So the Dispose is by design.

stephentoub: Ok, good, thanks.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
vijaykota: Done

src/System.Net.Security/src/System/Net/Unix/SecuritySafeHandles.cs
vijaykota: Sorry.. I forgot to remove this. Fixed now

vijaykota:
Thanks for the review, @stephentoub  and @bartonjs. All comments have been addressed.

PR #3536, opened 2015-09-29
Remove System.Security.SecureString from .NetCore

pallavit:
Since the change forced lock files update, I also updated the xunit version from beta to rc1 for the affected files.


pallavit:
@weshaggard Could you please review this change.

nguerrera:
I think 2.1.0 stable is out. Should we use that?

mellinoe:
It may make sense to try to jump straight to 2.1.0 stable. Assuming there's no other changes that we need to make along with that, it's probably better. If there's some breaking changes that we'd need to account for, we can put it off.

weshaggard:
If we try and jump to stable we need to also update the build tools test-runtime version to match. I wouldn't do it only here.

pallavit:
I would like to scope the change here for SecureString removal purposes. I agree we should try to move to the higher xuinit version across the board, however we should do that separately.

stephentoub:
LGTM

weshaggard:
LGTM. Please make sure you also unlist the package from the feeds.

pallavit:
Thanks for the review. Merging and closing.

PR #3539, opened 2015-09-29
Bump buildtools to 00097

ellismg:


tarekgh:
LGTM. thanks for CC'ing me.

dir.props
```diff
@@ -67,6 +67,7 @@
   <!-- list of nuget package sources passed to dnu -->
   <ItemGroup>
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
+
```
weshaggard: nit: was there a reason for the extra line?

ellismg: I had put another package source there for testing on Linux and when I removed it I missed deleting everything.

PR #3547, opened 2015-09-29
Prevent stack overflow from EnumerableQuery of null enumerable

### LINKED ISSUE #3546: Stack overflow enumerating EnumerableQuery created from null enumerable
`EnumerableQuery<T>` can be created passing null a `Expression`, an inappropriately typed `Expression`, a null `IEnumerable<T>`, or a null value of a derived type.

Blocking this would be a breaking change, especially since such objects can be usefully used as query providers.

Calling `GetEnumerator()` on most such cases result in an `ArgumentNullException` or `ArgumentException`, which could probably be improved (another exception might be more appropriate), but if the constructor was passed a null `IEnumerable<T>` then `GetEnumerator()` hits two cases where it would go into infinite recursion.

First when the expression is visited, it is then recursively visited again until stack overflow. Potentially there are other ways this path could be hit, and so other bugs it could cause.

If this is fixed, then `GetEnumerator()` itself will then delegate to its own `GetEnumerator()` causing infinite recursion again.

```C#
[Fact]
public void NullEnumerableConstantNullExpression()
{
  IQueryable<int> query = new EnumerableQuery<int>((IEnumerable<int>)null);
  var exp = (ConstantExpression)query.Expression;
  query.GetEnumerator(); // stack overflow.
}

### PR DESCRIPTION
**Pull request by JonHanna**
Fixes #3546

If an EnumerableQuery is created from a null IEnumerable<T>, calling
GetEnumerator() will cause infinite recursion. Change this to throw
immediately.

### DIFF
--- a/src/System.Linq.Queryable/src/Resources/Strings.resx
+++ b/src/System.Linq.Queryable/src/Resources/Strings.resx
@@ -135,4 +135,7 @@
   <data name="NoMethodOnTypeMatchingArguments" xml:space="preserve">
     <value>There is no method '{0}' on type '{1}' that matches the specified arguments</value>
   </data>
+  <data name="EnumeratingNullEnumerableExpression" xml:space="preserve">
+    <value>Cannot enumerate a query created from a null IEnumerable&lt;&gt;</value>
+  </data>
 </root>
--- a/src/System.Linq.Queryable/src/System/Linq/EnumerableQuery.cs
+++ b/src/System.Linq.Queryable/src/System/Linq/EnumerableQuery.cs
@@ -135,7 +135,10 @@ namespace System.Linq
                 EnumerableRewriter rewriter = new EnumerableRewriter();
                 Expression body = rewriter.Visit(_expression);
                 Expression<Func<IEnumerable<T>>> f = Expression.Lambda<Func<IEnumerable<T>>>(body, (IEnumerable<ParameterExpression>)null);
-                _enumerable = f.Compile()();
+                IEnumerable<T> enumerable = f.Compile()();
+                if (enumerable == this)
+                    throw Error.EnumeratingNullEnumerableExpression();
+                _enumerable = enumerable;
             }
             return _enumerable.GetEnumerator();
         }
--- a/src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs
+++ b/src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs
@@ -144,7 +144,9 @@ namespace System.Linq
                     Type t = GetPublicType(sq.Enumerable.GetType());
                     return Expression.Constant(sq.Enumerable, t);
                 }
-                return this.Visit(sq.Expression);
+                Expression exp = sq.Expression;
+                if (exp != c)
+                    return Visit(exp);
             }
             return c;
         }
--- a/src/System.Linq.Queryable/src/System/Linq/Error.cs
+++ b/src/System.Linq.Queryable/src/System/Linq/Error.cs
@@ -44,5 +44,10 @@ namespace System.Linq
         {
             return new InvalidOperationException(Strings.NoMethodOnTypeMatchingArguments(name, type));
         }
+
+        internal static Exception EnumeratingNullEnumerableExpression()
+        {
+            return new InvalidOperationException(Strings.EnumeratingNullEnumerableExpression());
+        }
     }
 }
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by JonHanna** [comment]

**Comment by JonHanna** [comment]

File: src/System.Linq.Queryable/src/System/Linq/EnumerableQuery.cs [hunk 1]
```diff
@@ -135,7 +135,10 @@ private IEnumerator<T> GetEnumerator()
                 EnumerableRewriter rewriter = new EnumerableRewriter();
                 Expression body = rewriter.Visit(_expression);
                 Expression<Func<IEnumerable<T>>> f = Expression.Lambda<Func<IEnumerable<T>>>(body, (IEnumerable<ParameterExpression>)null);
-                _enumerable = f.Compile()();
+                IEnumerable<T> enumerable = f.Compile()();
+                if (enumerable == this)
+                    throw new InvalidOperationException();
```
stephentoub: This should probably have a descriptive message associated with it (unless we don't do that throughout this project?)

stephentoub:
cc: @VSadov

src/System.Linq.Queryable/src/System/Linq/EnumerableQuery.cs
JonHanna: I'll add a message.
In similar cases,incidentally, we throw a rather cryptic ArgumentException. I'm not sure how much it would break to change those to InvalidOperationException.

src/System.Linq.Queryable/src/Resources/Strings.resx
```diff
@@ -135,4 +135,7 @@
   <data name="NoMethodOnTypeMatchingArguments" xml:space="preserve">
     <value>There is no method '{0}' on type '{1}' that matches the specified arguments</value>
   </data>
+  <data name="EnumeratingNullEnumerableExpression" xml:space="preserve">
+    <value>Cannot enumerate a query created from a null IEnumerable&lt;&gt;</value>
+  </data>
 </root>
```
jasonwilliams200OK: nit: missing EOF LF

JonHanna: Are we paying attention to that, when these files are often created by the resource editor tool rather than as text?

jasonwilliams200OK: Yes it is tool's fault. I think it is always better to have LF, for example on linux mono develop doesn't make this mistake. :)

JonHanna: As the tool does it, I'm not clear if the style guides apply. (Though I actually did use a text editor, but I left what wasn't relevant to my changes alone).

ellismg: My general preference is that for tool generated files like this we just leave whatever the tool produces.  I don't want us to spend time fighting VS over this stuff.

src/System.Linq.Queryable/src/System/Linq/EnumerableQuery.cs
```diff
@@ -135,7 +135,10 @@ private IEnumerator<T> GetEnumerator()
                 EnumerableRewriter rewriter = new EnumerableRewriter();
                 Expression body = rewriter.Visit(_expression);
                 Expression<Func<IEnumerable<T>>> f = Expression.Lambda<Func<IEnumerable<T>>>(body, (IEnumerable<ParameterExpression>)null);
-                _enumerable = f.Compile()();
+                IEnumerable<T> enumerable = f.Compile()();
+                if (enumerable == this)
+                    throw Error.EnumeratingNullEnumerableExpression();
```
VSadov: Is it possible to make the check and throw an error earlier - when query is created?

JonHanna: When the query is created would be ideal, but because the same class serves as both query and query provider there could be working code that creates a query from a null enumerable and then only ever uses it as a provider, and throwing on creation would hence break running code.
The check could be replaced with a null check in this method, but it wouldn't make much difference in the common case, and this way if there's some other pattern that could have the same effect it would also be caught.

VSadov: Ah, I see. I did not realize the breaking change aspect of that.

JonHanna: As often, it's a pity no IDE comes with a time-machine plugin ;)

PR #3549, opened 2015-09-30
Address perf for GetFullPath and defer to OS.

JeremyKuhne:
@vancem, @KrzysztofCwalina, @stephentoub, @Priya91 

This change attempts to limit the number of allocations made when evaluating
full paths. This is particularly important when Windows paths can be way over
the historical max of 260 characters. (The new max is over 32K)

We no longer kick paths in Windows that are valid, outside of paths with colons
that are not directory separators. (This has the effect of blocking alternate
data stream access- which is a legacy behavior.)

When creating directories, we still drop the trailing separator, which causes
trailing whitespace to be trimmed. This makes it particularly difficult to create
paths with trailing spaces in .NET, which is still desirable as most code has
a tough time dealing with this scenario. We'll open them fine now, which hits
the above point of not kicking valid legacy paths (paths that don't require
extended syntax).

Several simplifications to the logic were made, including moving relative segment
evaluation to shared code and calling it directly when getting directory names
rather than relying on the not-so-obvious prior behavior of GetFullPath without
"full checks".


stephentoub:
@dotnet-bot test this please

JeremyKuhne:
@stephentoub I'll be pushing an update that fixes some issues. I put this up so I could review the strategy with Vance (which I did off line).

JeremyKuhne:
@weshaggard, @stephentoub, @Priya91
This should be ready now. If there are any other people I should tag please let me know.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -235,6 +232,52 @@ private static bool StartsWithOrdinal(this StringBuilder builder, string value)
         }
 
         /// <summary>
+        /// Returns true if the given StringBuilder starts with the given value at the given index.
```
KrzysztofCwalina: The docs seem to be out of sync from the implementation.

src/Common/src/System/IO/PathInternal.Windows.cs
```diff
@@ -235,6 +232,52 @@ private static bool StartsWithOrdinal(this StringBuilder builder, string value)
...
+        /// </summary>
+        /// <param name="value">The string to compare against the start of the StringBuilder.</param>
+        /// <param name="index">The starting index to look for a match.</param>
+        internal static bool StartsWithOrdinal(this string source, string value, int index = 0)
+        {
+            if (value == null || source.Length < value.Length || index < 0)
+                return false;
+
+            for (int i = 0; i < value.Length; i++)
+            {
```
KrzysztofCwalina: This seems to be doing full string ordinal comparison. Should the loop start with index?

JeremyKuhne: Yes- I'll update it, thanks.

stephentoub:
@dotnet-bot test this please

src/Common/src/System/IO/PathInternal.cs
```diff
@@ -23,5 +25,67 @@ internal static void CheckInvalidPathChars(string path, bool checkAdditional = f
...
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Returns true if the given string starts with the given value.
+        /// </summary>
+        /// <param name="value">The string to compare against the start of the source string.</param>
+        internal static bool StartsWithOrdinal(this string source, string value)
+        {
```
stephentoub: Could this implementation just be:
```C#
return source.StartsWith(value, StringComparison.Ordinal);
```
?

src/Common/src/System/IO/PathInternal.cs
```diff
@@ -23,5 +25,67 @@ internal static void CheckInvalidPathChars(string path, bool checkAdditional = f
...
+            if (trimChars == null || trimChars.Length == 0)
+                return builder;
+
+            int end = builder.Length - 1;
+
+            for (; end >= 0; end--)
+            {
+                int i = 0;
+                char ch = builder[end];
```
stephentoub: This is probably completely fine, but at some point it'd be interesting to do some profiling with long paths to see if this shows up at all.  StringBuilder is now implemented as a linked list of builders, so indexing into it actually needs to walk that list to find the desired index.

src/System.IO.FileSystem/src/System/IO/Directory.cs
```diff
@@ -27,7 +27,7 @@ public static DirectoryInfo GetParent(String path)
                 throw new ArgumentException(SR.Argument_PathEmpty, "path");
             Contract.EndContractBlock();
 
-            String fullPath = PathHelpers.GetFullPathInternal(path);
+            String fullPath = Path.GetFullPath(path);
```
stephentoub: Excellent; I love that we're switching to use the public API surface area rather than an internal helper.

src/System.Runtime.Extensions/src/System/IO/Path.cs
```diff
@@ -455,5 +435,82 @@ private static string ToBase32StringSuitableForDirName(byte[] buff)
...
             return StringBuilderCache.GetStringAndRelease(sb);
         }
+
+        /// <summary>
+        /// Try and remove relative segments from the given path (without combining with a root).
+        /// </summary>
+        /// <param name="skip">Skip the specified number of characters before evaluating.</param>
+        private static string RemoveRelativeSegments(string path, int skip = 0)
+        {
+            // Remove "//", "/./", and "/../" from the path.  We would ideally use realpath
```
stephentoub: The realpath part of this comment is only relevant to Unix.

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
```diff
@@ -55,82 +57,21 @@ public static string GetFullPath(string path)
...
                 throw new PathTooLongException(SR.IO_PathTooLong);
             }
 
             string result =
-                sb.Length == 0 ? (fullCheck ? DirectorySeparatorCharAsString : string.Empty) :
-                sb.Length == path.Length ? path :
-                sb.ToString();
-            StringBuilderCache.Release(sb);
+                collapsedString.Length == 0 ? (fullChecks ? DirectorySeparatorCharAsString : string.Empty) :
+                collapsedString.Length == path.Length ? path :
```
stephentoub: Extracting the normalization code into RemoveRelativeSegments and having it return string rather than the StringBuilder nullified this optimization.  Previously, if no normalization was performed such that the input string was the same as the output string (which is pretty common), we would just return the original string and wouldn't need to convert the StringBuilder into a string.  But now with this change, we're always converting the StringBuilder to a string, even if it's identical to the original.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
```diff
@@ -130,469 +121,75 @@ private static bool ValidateExtendedPath(string path, bool fullCheck)
...
+                // This is because the nulls will signal the end of the string to Win32 and therefore have
+                // unpredictable results. Other invalid characters we give a chance to be normalized out.
+                if (path.IndexOf('\0') != -1)
+                    throw new ArgumentException(SR.Argument_InvalidPathChars, "path");
+
+                // Toss out paths with colons that aren't a valid drive specifier.
+                // Cannot start with a colon and can only be of the form "C:" or "\\?\C:".
+                // (Note that we used to explicitly check "http:" and "file:"- these are caught by this check now.)
+                int startIndex = PathInternal.PathStartSkip(path);
+                startIndex += isExtended ? PathInternal.ExtendedPathPrefix.Length + 2 : 2;
```
stephentoub: Should this instead be:
```C#
int startIndex = 2 + PathInternal.PathStartSkip(path);
if (isExtended)
{
    startIndex += PathInternal.ExtendedPathPrefix.Length;
}
```
?

src/System.Runtime.Extensions/src/System/IO/Path.cs
```diff
@@ -189,8 +168,9 @@ public static string GetFileNameWithoutExtension(string path)
         public static string GetPathRoot(string path)
         {
             if (path == null) return null;
-            path = NormalizePath(path, fullCheck: false);
-            return path.Substring(0, PathInternal.GetRootLength(path));
+            PathInternal.CheckInvalidPathChars(path);
+            int pathRoot = PathInternal.GetRootLength(path);
+            return pathRoot <= 0 ? String.Empty : path.Substring(0, pathRoot);
```
stephentoub: We should probably move GetPathRoot into the PAL layer.  On Unix, GetPathRoot is going to either return the empty string for relative paths or "/" for absolute paths, which means we should never need to be allocating any new strings.  If you'd prefer this be done as a separate change, that's fine... just open an issue to me to fix it.

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,373 @@
...
-
-            this.m_arrayPtr = charArrayPtr;
-            this.m_capacity = length;
-            this.m_maxPath = Path.MaxPath;
-            useStackAlloc = true;
-            doNotTryExpandShortFileName = false;
-        }
+    {
+        // This should cover the vast majority of usages without adding undue memory pressure
+        private const int MAX_BUILDER_SIZE = 1024;
```
stephentoub: Nit: MaxBuilderSize

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,373 @@
...
-        // Instantiates a PathHelper with a heap alloc'd array of ints. Will create a StringBuilder
-        internal PathHelper(int capacity, int maxPath)
-        {
-            this.m_sb = new StringBuilder(capacity);
-            this.m_capacity = capacity;
-            this.m_maxPath = maxPath;
-            doNotTryExpandShortFileName = false;
-        }
+        // Can't be over 8.3 and be a short name
+        private const int MAX_SHORT_NAME = 12;
```
stephentoub: Nit: MaxShortName

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,373 @@
...
-        internal int Capacity
-        {
-            get
-            {
-                return m_capacity;
-            }
-        }
+        // Trim trailing white spaces, tabs etc but don't be aggressive in removing everything that has UnicodeCategory of trailing space.
+        // string.WhitespaceChars will trim more aggressively than what the underlying FS does (for ex, NTFS, FAT).
+        private static readonly char[] TrimEndChars = { (char)0x9, (char)0xA, (char)0xB, (char)0xC, (char)0xD, (char)0x20, (char)0x85, (char)0xA0 };
```
stephentoub: Nit: s_trimEndChars

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,373 @@
...
-            set
+            finally
             {
-                Contract.Requires(index >= 0 && index < Length);
-                if (useStackAlloc)
-                {
-                    m_arrayPtr[index] = value;
-                }
-                else
+                if (t_cachedOutputBuffer != null && t_cachedOutputBuffer.Capacity > MAX_BUILDER_SIZE)
```
stephentoub: For multiple accesses to a thread local, we should cache it into a local first, e.g.
```C#
StringBuilder local = t_cachedOutputBuffer;
if (local != null && local.Capacity > MaxBuilderSize)
{
    t_cachedOutputBuffer = null;
}
```

stephentoub: Separately, though, why do we store the buffer into t_cachedOutputBuffer if it's too big in the first place?

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,373 @@
...
+                t_cachedOutputBuffer.EnsureCapacity(minimumCapacity);
+            }
 
-                // If success, the return buffer length does not account for the terminating null character.
-                // If in-sufficient buffer, the return buffer length does account for the path + the terminating null character.
-                // If failure, the return buffer length is zero 
-                if (r >= Path.MaxPath)
-                    throw new PathTooLongException(SR.IO_PathTooLong);
+            return t_cachedOutputBuffer;
+        }
```
stephentoub: It'd be cheaper to write this as:
```C#
StringBuilder local = t_cachedOutputBuffer;
if (local == null)
{
    t_cachedOutputBuffer = local = new StringBuilder(minimumCapacity);
}
else
{
    local.Clear();
    local.EnsureCapacity(minimumCapacity);
}
return local;
```
That said, it'd be better to not even store the builder into t_cachedOutputBuffer until we're done with it and know that it's not too big.

stephentoub:
> This has the effect of blocking alternate data stream access- which is a legacy behavior.

I didn't understand this comment.  Are you saying that alternate data streams were previously allowed and now aren't, were previously not allowed and now are, or something else?

> We'll open them fine now

I'm not quite following this, either.  Does this change impact existing behavior for non-extended paths at all, e.g. are we now throwing in places we didn't previously, not throwing in places we used to, not normalizing in places we used to, etc.?

@ericstj should also review.

Thanks.

src/Common/src/System/IO/PathInternal.cs
JeremyKuhne: It would differ in the behavior around null. I could change it to check null then call the other- thoughts?

src/Common/src/System/IO/PathInternal.cs
JeremyKuhne: Definitely want to keep digging. In our usages it should be flat- from what I can tell, the builders get collapsed when it gets P/Invoked.

src/System.Runtime.Extensions/src/System/IO/Path.cs
JeremyKuhne: I'll update the comment.

src/System.Runtime.Extensions/src/System/IO/Path.Unix.cs
JeremyKuhne: Whoops- I'll move this to the helper and keep track of whether or not we actually modify the string.

src/System.Runtime.Extensions/src/System/IO/Path.Windows.cs
JeremyKuhne: I suppose that is clearer. I'll change it.

src/System.Runtime.Extensions/src/System/IO/Path.cs
JeremyKuhne: I'll go ahead and change it.

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
JeremyKuhne: Was copying from StringBuilderCache- I'll change it.

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
JeremyKuhne: It is all contained here so it shouldn't be possible to hit multiple accesses, right? I was trying to avoid an extra nest on the main loop for clarity. I'll flip things around.

JeremyKuhne:
@stephentoub 
I'll try and clarify the comments a bit. We didn't allow colons before, which blocked alternate data streams. If we want to allow them we would need to add significantly more logic or drop the colon check altogether.

And yes, we no longer throw for some standard (non-extended) valid paths that we used to. The only difference in our behavior from Windows now is that we'll helpfully trim some leading whitespace in some cases (as before) and we'll kick potentially valid paths with colons in them (as before).

src/Common/src/System/IO/PathInternal.cs
stephentoub: Looking at the marshaling implementation for StringBuilder, it actually looks pretty expensive.  When marshaling in, we allocate native memory and potentially a managed array, potentially with multiple copies.  And when marshaling out, it looks like we allocate a new char[] to store the results and substitute that into the StringBuilder.  @vancem, is that correct?  If so, I'm concerned our use of StringBuilder in all of this path manipulation code is going to perform poorly.

src/System.Runtime.Extensions/src/System/IO/PathHelper.Windows.cs
```diff
@@ -2,457 +2,370 @@
...
-                Contract.Requires(index >= 0 && index < Length);
-                if (useStackAlloc)
-                {
-                    m_arrayPtr[index] = value;
-                }
-                else
+                if (fullPath != null && fullPath.Capacity <= MaxBuilderSize)
                 {
-                    m_sb[index] = value;
+                    // Too big, let it go
```
stephentoub: This comment no longer makes sense.

JeremyKuhne: Fixed

src/Common/src/System/IO/PathInternal.cs
JeremyKuhne: I'll walk through the P/Invoke various scenarios in the debugger. If I have to I'll switch this out with some other buffer.

JeremyKuhne:
@dotnet-bot test this please

ericstj:
I reviewed the code and it looks OK.  So long as we pass all existing tests and don't introduce new or different exceptions for existing behavior I'm OK with it (seems to be true).  LGTM.

PR #3553, opened 2015-09-30
Extract the whole cert chain during SSL handshake.

bartonjs:
When the server sends the whole cert chain, extract the elements after the leaf/EE certificate to be passed into X509Chain.ChainPolicy.ExtraStore; this will prevent failures due to unknown intermediate CAs (and may also prevent the need to download that intermediate).

This is a partial fix for #3362 (GetRemoteCertificate - Fetch remoteCertificateStore, if applicable for unix. (CertModule.cs)).

### DIFF
--- a/src/System.Net.Security/src/System/Net/Unix/CertModule.cs
+++ b/src/System.Net.Security/src/System/Net/Unix/CertModule.cs
@@ -76,6 +76,33 @@ namespace System.Net
                     remoteContext.DangerousAddRef(ref gotReference);
                     result = new X509Certificate2(remoteContext.DangerousGetHandle());
                 }
+
+                remoteCertificateStore = new X509Certificate2Collection();
+
+                SafeSharedX509StackHandle chainStack =
+                    Interop.OpenSsl.GetPeerCertificateChain(securityContext.DangerousGetHandle());
+
+                GC.KeepAlive(securityContext);
+
+                using (chainStack)
+                {
+                    if (!chainStack.IsInvalid)
+                    {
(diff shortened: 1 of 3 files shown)


bartonjs:
cc: @stephentoub @vijaykota @rajansingh10 @shrutigarg

This will need to be merged with Vijay's SafeHandle change, depending on which one goes in first.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -76,6 +76,33 @@ internal override X509Certificate2 GetRemoteCertificate(SafeDeleteContext securi
                     remoteContext.DangerousAddRef(ref gotReference);
                     result = new X509Certificate2(remoteContext.DangerousGetHandle());
                 }
+
+                remoteCertificateStore = new X509Certificate2Collection();
+
+                SafeSharedX509StackHandle chainStack =
+                    Interop.OpenSsl.GetPeerCertificateChain(securityContext.DangerousGetHandle());
```
rajansingh10: Q:- Should not it be DangerousAddRef way?

rajansingh10:
LGTM.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
vijaykota: GC.KeepAlive is sufficient. If https://github.com/dotnet/corefx/pull/3528 goes after this, I will fix this up

vijaykota:
LGTM. Thanks for fixing this, @bartonjs

PR #3560, opened 2015-09-30
Add portable PDB reader to System.Reflection.Metadata

nguerrera:
Submitting on behalf of @tmat

See Portable PDB specification: https://github.com/dotnet/roslyn/blob/master/docs/specs/PortablePdb-Metadata.md

**TODO**
- [x] API review: https://github.com/dotnet/apireviews/pull/17
- [x] Respond to API review feedback


src/System.Reflection.Metadata/src/System/Reflection/Metadata/BlobReader.cs
```diff
@@ -364,6 +394,7 @@ public int ReadCompressedInteger()
         /// </summary>
         /// <param name="value">The value of the compressed integer that was read.</param>
         /// <returns>true if the value was read successfully. false if the data at the current position was not a valid compressed integer.</returns>
+        /// <exception cref="BadImageFormatException">The data at the current position was not a valid compressed integer.</exception>
```
nguerrera: This looks wrong.

src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/PooledStringBuilder.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            // do not store builders that are too large.
+            if (builder.Capacity <= 1024)
+            {
+                builder.Clear();
+                _pool.Free(this);
+            }
+        }
+
+        [Obsolete("Consider calling ToStringAndFree instead.")]
```
nguerrera: We don't need this internal obsolete method. 

Also, should we switch to StringBuilderCache in Common\?

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.WinMD.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace System.Reflection.Metadata
 {
-    partial class MetadataReader
+    public partial class MetadataReader
```
nguerrera: If we're going to put 'public' here, we should also put sealed IMHO to matched the other partial declaration.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
```diff
@@ -816,6 +945,59 @@ internal void GetParameterRange(MethodDefinitionHandle methodDef, out int firstP
...
+            {
+                lastConstantRowId = this.LocalConstantTable.NumberOfRows;
+            }
+            else
+            {
+                lastConstantRowId = this.LocalScopeTable.GetConstantStart(scopeRowId + 1) - 1;
+            }
+        }
+
+        // TODO: move throw helpers to common place.
```
nguerrera: We have that common place now, but not caught in merge from master.

stephentoub:
@dotnet-bot test this please

src/System.Reflection.Metadata/src/System/Reflection/Metadata/BlobReader.cs
tmat: Yes, it doesn't throw.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.WinMD.cs
tmat: Let's just revert to ```partial class```

src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/PooledStringBuilder.cs
tmat: Feel free to improve this -- i think i just copied Roslyn impl.

src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs
tmat: ok, let's clean it up after it's merged.

tmat:
:+1:

src/System.Reflection.Metadata/src/System/Reflection/Metadata/PortablePdb/SequencePointCollection.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                return _reader.ReadCompressedInteger();
+            }
+
+            private ushort ReadColumn()
+            {
+                int column = _reader.ReadCompressedInteger();
+                if (column > ushort.MaxValue)
+                {
+                    // TODO:
```
nguerrera: Fixing these simple TODOs now.

nguerrera:
tagging @weshaggard

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Handles.cs
```diff
@@ -652,6 +652,18 @@ public override int GetHashCode()
...
         }
+
+        /// <summary>
+        /// Returns a handle to <see cref="MethodDebugInformation"/> corresponding to this handle.
+        /// </summary>
+        /// <remarks>
+        /// The resulting handle is only valid within the context of a <see cref="MetadataReader"/> open on the Portable PDB blob,
+        /// which in case of standalone PDB file is a different reader than the one containing this method definition.
+        /// </remarks>
+        public MethodDebugInformationHandle ToDebugInformationHandle()
```
weshaggard: I don't remember seeing this in the APIReview (although I could have missed it). Is this something that we want directly on Handle, as opposed to on MethodDebugInformationHandle?

tmat: This is new since the API review. It is on MethodDefinitionHandle, it's little confusing in the github review, since all handles are in a single file. MethodDebugInformationHandle has the reverse translation method.

weshaggard: Is there any real reason to have it on Handle then? I would prefer to not have the dependency in Handle and keep the dependency from MethodDebugInformationHandle to Handle instead of the reverse.

weshaggard:
I had a quick look over this and it looks good. Nice job @tmat and @nguerrera for getting this done.

@nguerrera can you please file a tracking issue for getting test coverage of the PortablePDB format in our repo.

nguerrera:
@tmat can you look at f9202f8 as well?

nguerrera:
@weshaggard I filed #3787 earlier

src/System.Reflection.Metadata/src/System/Reflection/Metadata/Handles.cs
weshaggard: Sorry I missed the fact that this method is on MethodDefinitionHandle as opposed to Handle like I thought. However my comment still stands as I think it would be better to have this as an extension method or a FromMethodDefintion method on MethodDebugInformationHandle.

nguerrera: I don't think we need to worry about the coupling here. By design all the handles are known to the metadata reader and I can't imagine splitting them up. A static From would be less discoverable and an extension would create a new type just for that, which would add clutter IMHO.

weshaggard: I'm fine with that, it was just something I noticed and wanted to call out.

jasonwilliams200OK:
@nguerrera, should we remove the following line:

> PDBs aren't generated by Roslyn on Unix. https://github.com/dotnet/roslyn/issues/2449

from [unix-instructions#known-issues](https://github.com/dotnet/corefx/blob/0c42646/Documentation/building/unix-instructions.md#known-issues)?

davidfowl:
This is pretty awesome :smile:

PR #3561, opened 2015-09-30
Removed dependency on System.Runtime 4.0.20

KrzysztofCwalina:
This component needs to depend on System.Runtime 4.0.10,
which does not have Buffer.MemoryCopy exposed.
This change adds an implementation of Buffer.MemoryCopy directly to this component.

### DIFF
--- a/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
+++ b/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
@@ -1,0 +1,201 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System {
+
+    // These sources are taken from corclr repo (src\mscorlib\src\System\Buffer.cs with x64 path removed)
+    // The reason for this duplication is that System.Runtime.dll 4.0.10 did not expose Buffer.MemoryCopy,
+    // but we need to make this component work with System.Runtime.dll 4.0.10
+    // The methods AreOverlapping and SlowCopyBackwards are not from Buffer.cs. Buffer.cs does an internal CLR call for these.
+    static class BufferInternal
+    {       
+        // This method has different signature for x64 and other platforms and is done for performance reasons.
+        [System.Security.SecurityCritical]
+        private unsafe static void Memmove(byte* dest, byte* src, uint len)
+        {
+            if (AreOverlapping(dest, src, len))
+            {
+                SlowCopyBackwards(dest, src, len);
+                return;
+            }
+
+            // This is portable version of memcpy. It mirrors what the hand optimized assembly versions of memcpy typically do.
+            switch (len)
+            {
+                case 0:
+                    return;
+                case 1:
+                    *dest = *src;
+                    return;
+                case 2:
+                    *(short*)dest = *(short*)src;
+                    return;
+                case 3:
+                    *(short*)dest = *(short*)src;
+                    *(dest + 2) = *(src + 2);
+                    return;
+                case 4:
+                    *(int*)dest = *(int*)src;
+                    return;
+                case 5:
+                    *(int*)dest = *(int*)src;
+                    *(dest + 4) = *(src + 4);
+                    return;
+                case 6:
+                    *(int*)dest = *(int*)src;
+                    *(short*)(dest + 4) = *(short*)(src + 4);
+                    return;
+                case 7:
+                    *(int*)dest = *(int*)src;
+                    *(short*)(dest + 4) = *(short*)(src + 4);
+                    *(dest + 6) = *(src + 6);
+                    return;
+                case 8:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    return;
+                case 9:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(dest + 8) = *(src + 8);
+                    return;
+                case 10:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(short*)(dest + 8) = *(short*)(src + 8);
+                    return;
+                case 11:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(short*)(dest + 8) = *(short*)(src + 8);
+                    *(dest + 10) = *(src + 10);
+                    return;
+                case 12:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(int*)(dest + 8) = *(int*)(src + 8);
+                    return;
+                case 13:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(int*)(dest + 8) = *(int*)(src + 8);
+                    *(dest + 12) = *(src + 12);
+                    return;
+                case 14:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(int*)(dest + 8) = *(int*)(src + 8);
+                    *(short*)(dest + 12) = *(short*)(src + 12);
+                    return;
+                case 15:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(int*)(dest + 8) = *(int*)(src + 8);
+                    *(short*)(dest + 12) = *(short*)(src + 12);
+                    *(dest + 14) = *(src + 14);
+                    return;
+                case 16:
+                    *(int*)dest = *(int*)src;
+                    *(int*)(dest + 4) = *(int*)(src + 4);
+                    *(int*)(dest + 8) = *(int*)(src + 8);
+                    *(int*)(dest + 12) = *(int*)(src + 12);
+                    return;
+                default:
+                    break;
+            }
+
+            if (((int)dest & 3) != 0)
+            {
+                if (((int)dest & 1) != 0)
+                {
+                    *dest = *src;
+                    src++;
+                    dest++;
+                    len--;
+                    if (((int)dest & 2) == 0)
+                        goto Aligned;
+                }
+                *(short*)dest = *(short*)src;
+                src += 2;
+                dest += 2;
+                len -= 2;
+            Aligned:;
+            }
+
+            uint count = len / 16;
+            while (count > 0)
+            {
+                ((int*)dest)[0] = ((int*)src)[0];
+                ((int*)dest)[1] = ((int*)src)[1];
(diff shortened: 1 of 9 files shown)


KrzysztofCwalina:
@muratg, @Tratcher, @ericstj, @tarekgh

src/System.Text.Encodings.Web/src/project.lock.json
```diff
@@ -1,18 +1,15 @@
 {
-  "locked": true,
+  "locked": false,
```
mellinoe: I think this should technically be locked, although you don't have any non-specific versions listed.

mellinoe:
Why does this need to support System.Runtime 4.0.10.0?

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
```diff
@@ -0,0 +1,271 @@
...
+
+    // These sources are taken from corclr repo (src\mscorlib\src\System\Buffer.cs)
+    // The reason for this duplication is that System.Runtime.dll 4.0.10 did not expose Buffer.MemoryCopy,
+    // but we need to make this component work with System.Runtime.dll 4.0.10
+    // The methods AreOverlapping and SlowCopyBackwards are not from Buffer.cs. Buffer.cs does an internal CLR call for these.
+    static class BufferInternal
+    {       
+        // This method has different signature for x64 and other platforms and is done for performance reasons.
+        [System.Security.SecurityCritical]
+#if WIN64
```
tarekgh: do we need to keep WIN64/X86 def'ed compilation? I expect the component will be always copmiled for AnyCpu. right?

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+
+        private static unsafe bool AreOverlapping(byte* dest, byte* src, ulong len)
+        {
+            ulong destAddress = (ulong)dest;
+            ulong srcAddress = (ulong)src;
+            ulong srcEnd = srcAddress + len;
+            ulong destEnd = destAddress + len;
+            if (srcEnd >= destAddress && srcEnd <= destEnd)
```
tarekgh: do we count for other case of the overlapping which is srcAddress lay between destAddress and destEnd but srcEnd is greater than destEnd?

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
KrzysztofCwalina: Possibly, but there is so much pointer arithmetic here that I really did not want to touch this code. If you feel strongly about it, I will remove the ifdefs.

src/System.Text.Encodings.Web/src/project.lock.json
KrzysztofCwalina: How do I lock it?

KrzysztofCwalina:
@mellinoe, see https://github.com/aspnet/HttpAbstractions/issues/391#issuecomment-144246862 for reasons for the change.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
KrzysztofCwalina: I don't this we need to copy backwards in this case. By the time forward copy gets to override the source, these source bytes are not needed anymore (as they have already been moved)

src/System.Text.Encodings.Web/src/project.lock.json
Tratcher: Why is this lock file checked in? Our recommendation is to not check them in.

KrzysztofCwalina: All (most I looked at) corefx projects seem to have lock files checked in.

mellinoe: Yeah, all of our projects right now have them checked in. I think we should move away from that myself, but right now that's how we're doing it, in this repo at least. Since none of the dependencies are non-specific, I don't think it technically matters, although I'm not familiar enough with how NuGet works to be sure about that.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
tarekgh: I am not feeling strongly about it.

tarekgh:
LGTM.

src/System.Text.Encodings.Web/src/project.lock.json
davidfowl: Yea it shouldn't have the lock file checked in. We're actually going to be removing this property.

KrzysztofCwalina:
test this please

KrzysztofCwalina:
@dotnet-bot test this please

src/System.Text.Encodings.Web/src/project.lock.json
mellinoe: @davidfowl Should we be removing all of the lock files we have checked in? All of the projects we build today have lock files in the repository here.

KrzysztofCwalina:
@dotnet-bot test this please

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
```diff
@@ -0,0 +1,271 @@
...
+        public static unsafe void MemoryCopy(void* source, void* destination, long destinationSizeInBytes, long sourceBytesToCopy)
+        {
+            if (sourceBytesToCopy > destinationSizeInBytes)
+            {
+                throw new ArgumentOutOfRangeException("sourceBytesToCopy");
+            }
+#if WIN64
+            Memmove((byte*)destination, (byte*)source, checked((ulong)sourceBytesToCopy));
+#else
+            Memmove((byte*)destination, (byte*)source, checked((uint)sourceBytesToCopy));
```
ericstj: Casting long to uint: overflow?

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
ericstj: Have you tested that it'll do the right thing when compiled as AnyCPU w/o win64 defined and JIT'ed as 64-bit?  I did a quick scan and it seems that this would be a) slower than it has to be b) restricted to uint lengths.  I couldn't find any correctness issues but I didn't scan everything.

IIRC this piece of code was very finely tuned to do the right tradeoff between perf of native vs overhead of p-invoke.  Using 32bit always might invalidate that.  We could achieve the 64/32 behavior on anycpu by turning those ifdefs into a runtime check.
 
@jkotas what do you think?

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
```diff
@@ -0,0 +1,271 @@
...
+                dest[i] = src[i];
+            }
+        }
+
+        private static unsafe bool AreOverlapping(byte* dest, byte* src, ulong len)
+        {
+            ulong destAddress = (ulong)dest;
+            ulong srcAddress = (ulong)src;
+            ulong srcEnd = srcAddress + len;
+            ulong destEnd = destAddress + len;
```
jkotas: You should do this math using pointers - without casting to ulong.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
KrzysztofCwalina: This is code taken as-is from Buffer.cs. I assume the cast is checked for the very reason that it can but should not overflow in practice, as the integer is supposed to be the length of contiguous memory block.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
jkotas: I am assuming that you would like to keep the library as AnyCPU and portable - no PInvokes:

If you do not want to give away performance, you should create two copies of this method - one for 32-bit and second for 64-bit (with the right sides of WIN64 #if deleted), switch between them at call site using `if (sizeof(void*) == 8)`.

Using 32-bit version always will work too but it will be slower, in particular for the larger lengths. You should measure whether it is acceptable.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
jkotas: It is confusing because the original in Buffer.cs is compiled platform specific, but this library is compiled as AnyCPU. It may be useful to delete all WIN64 parts and all "long" overloads to make the code easier to understand - assuming that you would like to keep this library as AnyCPU.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
KrzysztofCwalina: Makes sense.

src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/BufferInternal.cs
KrzysztofCwalina: Yeah, but I wanted to minimize changes (i.e. make no changes) to the code I copy-pasted.

KrzysztofCwalina:
By popular demand removed the x64 ifdefs

src/System.Text.Encodings.Web/src/project.json
```diff
@@ -1,16 +1,16 @@
...
-    "System.IO": "4.0.10",
-    "System.Reflection": "4.0.10",
+    "System.Globalization": "4.0.0",
+    "System.IO": "4.0.0",
+    "System.Reflection": "4.0.0",
     "System.Reflection.Primitives": "4.0.0",
     "System.Resources.ResourceManager": "4.0.0",
-    "System.Runtime": "4.0.20",
-    "System.Runtime.Extensions": "4.0.10",
+    "System.Runtime": "4.0.10",
```
ericstj: You can go back as far as 4.0.0 and have this work on net 4.5 as well.  Make sure to do so for all dependencies.

src/System.Text.Encodings.Web/src/project.json
```diff
@@ -1,16 +1,16 @@
...
+    "System.Runtime.Extensions": "4.0.0",
     "System.Text.Encoding": "4.0.10",
     "System.Threading": "4.0.10"
```
ericstj: Both System.Text.Encoding 4.0.10 and System.Threading 4.0.10 are only supported on >= .Net 4.6.  See previous comment.

KrzysztofCwalina: I changed them all to 4.0.0

ericstj:
LGTM

src/System.Text.Encodings.Web/src/project.json
akoeplinger: @KrzysztofCwalina commit message still mentions "System.Runtime 4.0.10" which might be misleading.

PR #3567, opened 2015-10-01
Adding the Managed SNI for System.Data.SqlClient for xplat

saurabh500:
The code contains the Network interaction of the SqlDriver ported to Managed code.

The code supports tcp connectivity scenario.

### DIFF
--- a/src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
+++ b/src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
@@ -8,9 +8,9 @@ namespace System.Data.SqlClient
 {
     internal static class SNINativeMethodWrapper
     {
+#if !MANAGED_SNI
         private const string SNI = "sni.dll";
 
-#if !MANAGED_SNI
         private static int s_sniMaxComposedSpnLength = -1;
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
@@ -228,10 +228,10 @@ namespace System.Data.SqlClient
 
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SNIWaitForSSLHandshakeToCompleteWrapper")]
         internal static extern uint SNIWaitForSSLHandshakeToComplete([In] SNIHandle pConn, int dwMilliseconds);
-#endif
+
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         internal static extern uint UnmanagedIsTokenRestricted([In] IntPtr token, [MarshalAs(UnmanagedType.Bool)] out bool isRestricted);
-#if !MANAGED_SNI
+
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         private static extern uint GetSniMaxComposedSpnLength();
 
@@ -416,6 +416,9 @@ namespace System.Data
     {
         internal static bool IsTokenRestrictedWrapper(IntPtr token)
         {
+#if MANAGED_SNI
+            throw new PlatformNotSupportedException("The Win32NativeMethods.IsTokenRestrictedWrapper is not supported on non-Windows platform");
+#else
             bool isRestricted;
             uint result = SNINativeMethodWrapper.UnmanagedIsTokenRestricted(token, out isRestricted);
 
@@ -425,6 +428,7 @@ namespace System.Data
             }
 
             return isRestricted;
+#endif
         }
     }
 }
--- a/src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
+++ b/src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
@@ -10,6 +10,7 @@ namespace System.Data.SqlClient
     {
         private const string SNI = "sni.dll";
 
+#if !MANAGED_SNI
         private static int s_sniMaxComposedSpnLength = -1;
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
@@ -26,8 +27,10 @@ namespace System.Data.SqlClient
                 return s_sniMaxComposedSpnLength;
             }
         }
+#endif
 
         #region Structs\Enums
+#if !MANAGED_SNI
         [StructLayout(LayoutKind.Sequential)]
         internal struct ConsumerInfo
         {
@@ -163,6 +166,7 @@ namespace System.Data.SqlClient
             internal string function;
             internal uint lineNumber;
         }
+#endif
 
         internal enum SniSpecialErrors : uint
         {
@@ -179,6 +183,7 @@ namespace System.Data.SqlClient
         #endregion
 
         #region DLL Imports
+#if !MANAGED_SNI
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SNIAddProviderWrapper")]
         internal static extern uint SNIAddProvider(SNIHandle pConn, ProviderEnum ProvNum, [In] ref uint pInfo);
 
@@ -223,8 +228,10 @@ namespace System.Data.SqlClient
 
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SNIWaitForSSLHandshakeToCompleteWrapper")]
         internal static extern uint SNIWaitForSSLHandshakeToComplete([In] SNIHandle pConn, int dwMilliseconds);
+#endif
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         internal static extern uint UnmanagedIsTokenRestricted([In] IntPtr token, [MarshalAs(UnmanagedType.Bool)] out bool isRestricted);
+#if !MANAGED_SNI
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         private static extern uint GetSniMaxComposedSpnLength();
 
@@ -272,7 +279,9 @@ namespace System.Data.SqlClient
 
         [DllImport(SNI, CallingConvention = CallingConvention.Cdecl)]
         private static extern uint SNIWriteSyncOverAsync(SNIHandle pConn, [In] SNIPacket pPacket);
+#endif
         #endregion
+#if !MANAGED_SNI
         internal static uint SniGetConnectionId(SNIHandle pConn, ref Guid connId)
         {
             return SNIGetInfoWrapper(pConn, QTypes.SNI_QUERY_CONN_CONNID, out connId);
@@ -388,6 +397,7 @@ namespace System.Data.SqlClient
                 : IntPtr.Zero;
             native_consumerInfo.ConsumerKey = consumerInfo.key;
         }
+#endif
     }
 }
 
--- a/src/System.Data.SqlClient/src/System.Data.SqlClient.csproj
+++ b/src/System.Data.SqlClient/src/System.Data.SqlClient.csproj
@@ -9,7 +9,9 @@
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PlatformTarget>AnyCPU</PlatformTarget>
-	<DefineConstants>$(DefineConstants);PROJECTK</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetsWindows)' != 'true' ">
+    <DefineConstants>$(DefineConstants);MANAGED_SNI</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
   </PropertyGroup>
@@ -141,6 +143,21 @@
     <Compile Include="System\Data\SqlTypes\SQLUtility.cs" />
     <Compile Include="System\Data\SqlTypes\SqlXml.cs" />
 	<Compile Include="$(CommonPath)\System\NotImplemented.cs"/>
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetsWindows)' != 'true' ">
+    <Compile Include="System\Data\SqlClient\SNI\LimitedConcurrencyLevelTaskScheduler.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIError.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIHandle.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNILoadHandle.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIMarsConnection.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIMarsHandle.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIMarsManager.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIMarsQueuedPacket.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIPacket.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNIProxy.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNITcpHandle.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SslOverTdsStream.cs" />
+    <Compile Include="System\Data\SqlClient\SNI\SNICommon.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
--- a/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/LimitedConcurrencyLevelTaskScheduler.cs
+++ b/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/LimitedConcurrencyLevelTaskScheduler.cs
@@ -1,0 +1,186 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// Limited concurrent level task scheduler
+    /// </summary>
+    internal class LimitedConcurrencyLevelTaskScheduler : TaskScheduler
+    {
+        [ThreadStatic]
+        private static bool s_currentThreadIsProcessingItems;
+        private readonly LinkedList<Task> _tasks = new LinkedList<Task>();
+        private readonly int _maxDegreeOfParallelism;
+        private int _delegatesQueuedOrRunning = 0;
+
+        /// <summary>
+        /// Gets the maximum concurrency level supported by this scheduler.  
+        /// </summary>
+        public sealed override int MaximumConcurrencyLevel
+        {
+            get
+            {
+                return _maxDegreeOfParallelism;
+            }
+        }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="maxDegreeOfParallelism">Number of tasks to allow to run concurrently</param>
+        public LimitedConcurrencyLevelTaskScheduler(int maxDegreeOfParallelism)
+        {
+            if (maxDegreeOfParallelism < 1)
+            {
+                throw new ArgumentOutOfRangeException("maxDegreeOfParallelism");
+            }
+
+            _maxDegreeOfParallelism = maxDegreeOfParallelism;
+        }
+
+        /// <summary>
+        /// Queues a task to the scheduler
+        /// </summary>
+        /// <param name="task">Task</param>
+        protected sealed override void QueueTask(Task task)
+        {
+            // Add the task to the list of tasks to be processed.  If there aren't enough  
+            // delegates currently queued or running to process tasks, schedule another.  
+            lock (_tasks)
+            {
+                _tasks.AddLast(task);
+                if (_delegatesQueuedOrRunning < _maxDegreeOfParallelism)
+                {
+                    ++_delegatesQueuedOrRunning;
+                    NotifyThreadPoolOfPendingWork();
+                }
+            }
+        }
+
+        /// <summary>
+        /// Inform thread pool there is pending work to do.
+        /// </summary>
+        private void NotifyThreadPoolOfPendingWork()
+        {
+            ThreadPool.QueueUserWorkItem(_ =>
+            {
+                // Note that the current thread is now processing work items. 
+                // This is necessary to enable inlining of tasks into this thread.
+                s_currentThreadIsProcessingItems = true;
+
+                try
+                {
+                    // Process all available items in the queue. 
+                    while (true)
+                    {
+                        Task item;
+                        lock (_tasks)
+                        {
+                            // When there are no more items to be processed, 
+                            // note that we're done processing, and get out. 
+                            if (_tasks.Count == 0)
+                            {
+                                --_delegatesQueuedOrRunning;
+                                break;
+                            }
+
+                            // Get the next item from the queue
+                            item = _tasks.First.Value;
+                            _tasks.RemoveFirst();
+                        }
+
+                        // Execute the task we pulled out of the queue 
+                        base.TryExecuteTask(item);
+                    }
+                }
+
+                // We're done processing items on the current thread 
+                finally
+                {
+                    s_currentThreadIsProcessingItems = false;
+                }
+            }, null);
+        }
+
+        /// <summary>
+        /// Attempts to execute the specified task on the current thread.  
+        /// </summary>
+        /// <param name="task">Task</param>
+        /// <param name="taskWasPreviouslyQueued">true if task was previously queued</param>
+        /// <returns>If task was executed.</returns>
+        protected sealed override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQueued)
+        {
+            // If this thread isn't already processing a task, we don't support inlining 
+            if (!s_currentThreadIsProcessingItems)
+            {
+                return false;
+            }
+
+            // If the task was previously queued, remove it from the queue 
+            if (taskWasPreviouslyQueued)
+            {
+                // Try to run the task.  
+                if (TryDequeue(task))
+                {
+                    return base.TryExecuteTask(task);
+                }
+                else
+                {
+                    return false;
+                }
+            }
+            else
+            {
+                return base.TryExecuteTask(task);
+            }
+        }
+
+        /// <summary>
+        /// Remove previously scheduled task.
+        /// </summary>
+        /// <param name="task">Task</param>
+        /// <returns>If task was removed</returns>
+        protected sealed override bool TryDequeue(Task task)
+        {
+            lock (_tasks)
+            {
+                return _tasks.Remove(task);
+            }
+        }
+
+        /// <summary>
+        /// Gets an enumerable of the tasks currently scheduled on this scheduler.  
+        /// </summary>
+        /// <returns>Tasks scheduled</returns>
+        protected sealed override IEnumerable<Task> GetScheduledTasks()
+        {
+            bool lockTaken = false;
+
+            try
+            {
+                Monitor.TryEnter(_tasks, ref lockTaken);
+
+                if (lockTaken)
+                {
+                    return _tasks;
+                }
+                else
+                {
+                    throw new NotSupportedException();
+                }
+            }
+            finally
+            {
+                if (lockTaken)
+                {
+                    Monitor.Exit(_tasks);
+                }
+            }
+        }
+    }
+}
--- a/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
@@ -19,8 +19,9 @@ namespace System.Data.SqlClient.SNI
     internal enum SNIProviders
     {
         NP_PROV = 1,        // Named Pipes Provider
+        SMUX_PROV = 5,      // MARS Provider
         TCP_PROV = 7,       // TCP Provider
-        INVALID_PROV = 10,   // Invalid Provider
+        INVALID_PROV = 10,  // Invalid Provider
     }
 
     /// <summary>
@@ -50,13 +51,6 @@ namespace System.Data.SqlClient.SNI
         SMUX_DATA = 8       // SMUX data packet
     }
 
-    /// <summary>
-    /// Consumer info for SNI
-    /// </summary>
-    internal class SNIConsumerInfo
-    {
-    }
-
     internal class SNICommon
     {
         /// <summary>
@@ -70,7 +64,6 @@ namespace System.Data.SqlClient.SNI
         /// <returns>True if certificate is valid</returns>
         internal static bool ValidateSslServerCertificate(string targetServerName, object sender, X509Certificate cert, X509Chain chain, SslPolicyErrors policyErrors)
         {
-            // If the certificate is not a valid, signed certificate, then fail validation.
             if (policyErrors == SslPolicyErrors.None)
             {
                 return true;
--- a/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
+++ b/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
@@ -1,0 +1,135 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// SNI Asynchronous callback
+    /// </summary>
+    /// <param name="packet">SNI packet</param>
+    /// <param name="sniErrorCode">SNI error code</param>
+    internal delegate void SNIAsyncCallback(SNIPacket packet, uint sniErrorCode);
+
+    /// <summary>
+    /// SNI provider identifiers
+    /// </summary>
+    internal enum SNIProviders
+    {
+        NP_PROV = 1,        // Named Pipes Provider
+        TCP_PROV = 7,       // TCP Provider
+        INVALID_PROV = 10,   // Invalid Provider
+    }
+
+    /// <summary>
+    /// SMUX packet header
+    /// </summary>
+    internal class SNISMUXHeader
+    {
+        public const int HEADER_LENGTH = 16;
+
+        public byte SMID;
+        public byte flags;
+        public ushort sessionId;
+        public uint length;
+        public uint sequenceNumber;
+        public uint highwater;
+    }
+
+    /// <summary>
+    /// SMUX packet flags
+    /// </summary>
+    [Flags]
+    internal enum SNISMUXFlags
+    {
+        SMUX_SYN = 1,       // Begin SMUX connection
+        SMUX_ACK = 2,       // Acknowledge SMUX packets
+        SMUX_FIN = 4,       // End SMUX connection
+        SMUX_DATA = 8       // SMUX data packet
+    }
+
+    /// <summary>
+    /// Consumer info for SNI
+    /// </summary>
+    internal class SNIConsumerInfo
+    {
+    }
+
+    internal class SNICommon
+    {
+        /// <summary>
+        /// Validate server certificate callback for SSL
+        /// </summary>
+        /// <param name="targetServerName">Server that client is expecting to connect to</param>
+        /// <param name="sender">Sender object</param>
+        /// <param name="cert">X.509 certificate</param>
+        /// <param name="chain">X.509 chain</param>
+        /// <param name="policyErrors">Policy errors</param>
+        /// <returns>True if certificate is valid</returns>
+        internal static bool ValidateSslServerCertificate(string targetServerName, object sender, X509Certificate cert, X509Chain chain, SslPolicyErrors policyErrors)
+        {
+            // If the certificate is not a valid, signed certificate, then fail validation.
+            if (policyErrors == SslPolicyErrors.None)
+            {
+                return true;
+            }
+
+            if ((policyErrors & SslPolicyErrors.RemoteCertificateNameMismatch) != 0)
+            {
+                string certServerName = cert.Subject.Substring(cert.Subject.IndexOf('=') + 1);
+
+                // Verify that target server name matches subject in the certificate
+                if (targetServerName.Length > certServerName.Length)
+                {
+                    return false;
+                }
+                else if (targetServerName.Length == certServerName.Length)
+                {
+                    // Both strings have the same length, so targetServerName must be a FQDN
+                    if (!targetServerName.Equals(certServerName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    if (string.Compare(targetServerName, 0, certServerName, 0, targetServerName.Length, StringComparison.OrdinalIgnoreCase) != 0)
+                    {
+                        return false;
+                    }
+
+                    // Server name matches cert name for its whole length, so ensure that the
+                    // character following the server name is a '.'. This will avoid
+                    // having server name "ab" match "abc.corp.company.com"
+                    // (Names have different lengths, so the target server can't be a FQDN.)
+                    if (certServerName[targetServerName.Length] != '.')
+                    {
+                        return false;
+                    }
+                }
+            }
+            else
+            {
+                // Fail all other SslPolicy cases besides RemoteCertificateNameMismatch
+                return false;
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Sets last error encountered for SNI
+        /// </summary>
+        /// <param name="provider">SNI provider</param>
+        /// <param name="nativeError">Native error code</param>
+        /// <param name="sniError">SNI error code</param>
+        /// <param name="errorMessage">Error message</param>
+        /// <returns></returns>
+        internal static uint ReportSNIError(SNIProviders provider, uint nativeError, uint sniError, string errorMessage)
+        {
+            SNILoadHandle.SingletonInstance.LastError = new SNIError(provider, nativeError, sniError, errorMessage);
+            return TdsEnums.SNI_ERROR;
+        }
+    }
+}
--- a/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
+++ b/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
@@ -9,14 +9,14 @@ namespace System.Data.SqlClient.SNI
     /// </summary>
     internal class SNIError
     {
-        public SNIProviders provider;
-        public String errorMessage;
-        public uint nativeError;
-        public uint sniError;
-        public String function;
-        public uint lineNumber;
+        public readonly SNIProviders provider;
+        public readonly string errorMessage;
+        public readonly uint nativeError;
+        public readonly uint sniError;
+        public readonly string function;
+        public readonly uint lineNumber;
 
-        public SNIError(SNIProviders provider, uint nativeError, uint sniErrorCode, String errorMessage)
+        public SNIError(SNIProviders provider, uint nativeError, uint sniErrorCode, string errorMessage)
         {
             this.lineNumber = 0;
             this.function = "";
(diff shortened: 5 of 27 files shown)


saurabh500:
Build failed due to test failure in System.Net.Sockets.Legacy test failed.

src/System.Data.SqlClient/src/System.Data.SqlClient.csproj
```diff
@@ -9,7 +9,9 @@
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PlatformTarget>AnyCPU</PlatformTarget>
-	<DefineConstants>$(DefineConstants);PROJECTK</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetsWindows)' != 'true' ">
+    <DefineConstants>$(DefineConstants);MANAGED_SNI</DefineConstants>
```
stephentoub: Why not use it everywhere?  What features are we missing on Windows that prevent this?

src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
```diff
@@ -10,6 +10,7 @@ internal static class SNINativeMethodWrapper
     {
         private const string SNI = "sni.dll";
 
+#if !MANAGED_SNI
```
stephentoub: In general for corefx we've been trying to move away from having these kinds of ifdefs sprinkled throughout the code.  Instead, we've opted for a scheme where we have different files containing the different implementations (either as static members called to or members on partial classes), and then the different builds include the files they need.  We should consider such an approach here as well.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/LimitedConcurrencyLevelTaskScheduler.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// Limited concurrent level task scheduler
+    /// </summary>
+    internal class LimitedConcurrencyLevelTaskScheduler : TaskScheduler
```
stephentoub: I don't think you should need this.  If your goal is just to be able to control the max number of tasks that can run concurrently, you can construct a ConcurrentExclusiveTaskScheduler with that max number, and then schedule to its ConcurrentScheduler.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="serverName">Server name</param>
+        /// <param name="port">TCP port number</param>
+        /// <param name="timerExpire">Connection timer expiration</param>
+        /// <param name="callbackObject">Callback object</param>
+        public SNITCPHandle(string serverName, int port, long timerExpire, object callbackObject)
+        {
+            _writeScheduler = new LimitedConcurrencyLevelTaskScheduler(1);
```
stephentoub: Per my previous comment, this could be:
```C#
_writeScheduler = new ConcurrentExclusiveSchedulerPair(1).ConcurrentScheduler;
```
or because you're actually just trying to serialize execution, making everything exclusive, you could do:
```C#
_writeScheduler = new ConcurrentExclusiveSchedulerPair().ExclusiveScheduler;
```

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNITcpHandle.cs
```diff
@@ -0,0 +1,421 @@
...
+        private object _callbackObject;
+        private Stream _stream;
+        private Socket _socket;
+        private NetworkStream _tcpStream;
+        private SslStream _sslStream;
+        private SslOverTdsStream _sslOverTdsStream;
+        private TcpClient _tcpClient;
+        private SNIAsyncCallback _receiveCallback;
+        private SNIAsyncCallback _sendCallback;
+        private TaskScheduler _writeScheduler;
```
stephentoub: Looks like more of these fields can be made readonly.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// SNI Asynchronous callback
+    /// </summary>
+    /// <param name="packet">SNI packet</param>
+    /// <param name="sniErrorCode">SNI error code</param>
+    internal delegate void SNIAsyncCallback(SNIPacket packet, uint sniErrorCode);
```
stephentoub: Why a custom delegate rather than ```Action<SNIPacket, uint>```?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
```diff
@@ -0,0 +1,135 @@
...
+        SMUX_SYN = 1,       // Begin SMUX connection
+        SMUX_ACK = 2,       // Acknowledge SMUX packets
+        SMUX_FIN = 4,       // End SMUX connection
+        SMUX_DATA = 8       // SMUX data packet
+    }
+
+    /// <summary>
+    /// Consumer info for SNI
+    /// </summary>
+    internal class SNIConsumerInfo
```
stephentoub: What's the purpose of this type?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            this.lineNumber = 0;
+            this.function = "";
+            this.provider = provider;
+            this.nativeError = nativeError;
+            this.sniError = sniErrorCode;
+
+            // For now hardcode "TCP Provider" to keep TDS parser happy
+            //
+            this.errorMessage = "TCP Provider: " + errorMessage + "\r\n";
```
stephentoub: Does this need to be localized?  Does the "\r\n" need to be a platform-agnostic newline?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal class SNIError
+    {
+        public SNIProviders provider;
+        public String errorMessage;
+        public uint nativeError;
+        public uint sniError;
+        public String function;
+        public uint lineNumber;
+
+        public SNIError(SNIProviders provider, uint nativeError, uint sniErrorCode, String errorMessage)
```
stephentoub: Nit: we're not yet consistent across the whole codebase, but we're trying to use the C# keywords instead of type names where possible, e.g. string instead of String

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading;
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// Global SNI settings and status
+    /// </summary>
+    internal class SNILoadHandle
+    {
+        static public SNILoadHandle SingletonInstance = new SNILoadHandle();
```
stephentoub: readonly?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private ThreadLocal<SNIError> _lastError = new ThreadLocal<SNIError>(() => { return new SNIError(SNIProviders.INVALID_PROV, 0, 0, "No SNI Error has been reported yet."); });
```
stephentoub: I'm not sure what's exposed publically and what's not... does this string need to be in the .resx?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+        private readonly UInt32 _status = TdsEnums.SNI_SUCCESS;
```
stephentoub: Nit: another example where it'd be preferable to use a C# keyword instead of type name, e.g. uint instead of UInt32.  I'll stop commenting on these.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// SNI MARS connection. Multiple MARS streams will be overlayed on this connection.
+    /// </summary>
+    internal class SNIMarsConnection
+    {
+        private SNIHandle _lowerHandle;
+        private ushort _nextSessionId = 0;
+        private Guid _connectionId = Guid.NewGuid();
```
stephentoub: Some of these fields look like they could be readonly.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="lowerHandle">Lower handle</param>
+        public SNIMarsConnection(SNIHandle lowerHandle)
+        {
+            _lowerHandle = lowerHandle;
+            _lowerHandle.SetAsyncCallbacks(HandleReceiveComplete, HandleSendComplete);
```
stephentoub: As is, this is going to allocate delegates for both HandleReceiveComplete and HandleSendComplete.  If you were to change the signature of the delegate to accept some state object as well, then you could declare these to be lambdas that take that state and the C# compiler will cache the delegates so they won't be reallocated each time.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        public SNIMarsHandle CreateSession(Object callbackObject, bool async)
+        {
+            lock (this)
+            {
+                ushort sessionId = _nextSessionId++;
```
stephentoub: It's ok if session IDs are eventually reused?  If not, should this be a ulong instead of a ushort?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+        public uint StartReceive()
+        {
+            SNIPacket packet = null;
+
+            if (ReceiveAsync(ref packet) == TdsEnums.SNI_SUCCESS_IO_PENDING)
+            {
+                return TdsEnums.SNI_SUCCESS_IO_PENDING;
+            }
+
+            SNILoadHandle.SingletonInstance.LastError = new SNIError(SNIProviders.TCP_PROV, 0, 0, "Could not start MARS async receive");
```
stephentoub: I'm used to file I/O where an async operation can actually return synchronously, in which case it won't return "pending".  That's not the case here, right?  ReceiveAsync will return PENDING even if it actually completed synchronously (or maybe synchronous completion isn't possible)?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+                return _lowerHandle.CheckConnection();
+            }
+        }
+
+        /// <summary>
+        /// Process a receive error
+        /// </summary>
+        public void HandleReceiveError()
+        {
+            foreach (SNIMarsHandle handle in _sessions.Values)
```
stephentoub: _sessions is protected elsewhere by a ```lock(this)```... that's not necessary here?  If it's not necessary because you expect HandleReceiveError will only ever be called while holding the lock, I'd recommend asserting that, e.g. ```Debug.Assert(Monitor.IsEntered(this));```

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
```diff
@@ -0,0 +1,314 @@
...
+                            }
+                        }
+
+                        _currentHeader = new SNISMUXHeader();
+                        _currentHeader.SMID = _headerBytes[0];
+                        _currentHeader.flags = _headerBytes[1];
+                        _currentHeader.sessionId = BitConverter.ToUInt16(_headerBytes, 2);
+                        _currentHeader.length = BitConverter.ToUInt32(_headerBytes, 4) - SNISMUXHeader.HEADER_LENGTH;
+                        _currentHeader.sequenceNumber = BitConverter.ToUInt32(_headerBytes, 8);
+                        _currentHeader.highwater = BitConverter.ToUInt32(_headerBytes, 12);
```
stephentoub: Just as a style thing, consider using object initialization syntax to highlight the configuration being done here, e.g.
```C#
_currentHeader = new SNISMUXHeader()
{
    SMID = _headerBytes[0],
    flags = _headerBytes[1],
    sessionId = BitConverter.ToUInt16(_headerBytes, 2);
    length = BitConverter.ToUInt32(_headerBytes, 4) - SNISMUXHeader.HEADER_LENGTH,
    sequenceNumber = BitConverter.ToUInt32(_headerBytes, 8),
    highwater = BitConverter.ToUInt32(_headerBytes, 12)
};

File: src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs [hunk 19]
```diff
@@ -0,0 +1,57 @@
...
+        public SNIError LastError
+        {
+            get
+            {
+                return _lastError.Value;
+            }
+
+            set
+            {
+                _lastError = new ThreadLocal<SNIError>(() => { return value; });
```
stephentoub: Why not:
```C#
_lastError.Value = value;
```
?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+        /// </summary>
+        public override void Dispose()
+        {
+            try
+            {
+                SendControlPacket(SNISMUXFlags.SMUX_FIN, false);
+            }
+            catch (Exception)
+            {
+            }
```
stephentoub: Should we at least Debug.Fail or something?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+            {
+                GetSMUXHeaderBytes(0, (byte)flags, ref headerBytes);
+            }
+
+            SNIPacket packet = new SNIPacket(null);
+            packet.SetData(headerBytes, SNISMUXHeader.HEADER_LENGTH);
+
+            if (async)
+            {
+                _connection.SendAsync(packet, (sentPacket, error) => { });
```
stephentoub: Do we need to verify anything about the return value?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+        private Object _callbackObject;
+        private Guid _connectionId = Guid.NewGuid();
+        private ushort _sessionId;
+        private uint _sequenceNumber;
+        private uint _receiveHighwater = 4;
+        private uint _receiveHighwaterLastAck = 4;
+        private uint _sendHighwater = 4;
+        private int _asyncReceives = 0;
+        private ManualResetEvent _packetEvent = new ManualResetEvent(false);
+        private ManualResetEvent _ackEvent = new ManualResetEvent(false);
```
stephentoub: Consider using ManualResetEventSlim instead of ManualResetEvent.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+
+                encapsulatedPacket = GetSMUXEncapsulatedPacket(packet);
+
+                if (callback != null)
+                {
+                    encapsulatedPacket.SetCompletionCallback(callback);
+                }
+                else
+                {
+                    encapsulatedPacket.SetCompletionCallback(HandleSendComplete);
```
stephentoub: I assume InternalSendAsync is called relatively frequently?  Each call that results in this SetCompletionCallback call is going to result in allocating a delegate for HandleSendComplete.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+        /// Handle send completion
+        /// </summary>
+        /// <param name="packet">SNI packet</param>
+        /// <param name="sniErrorCode">SNI error code</param>
+        public void HandleSendComplete(SNIPacket packet, uint sniErrorCode)
+        {
+            lock (this)
+            {
+                Debug.Assert(_callbackObject != null);
+#if MANAGED_SNI
```
stephentoub: This file appears to only be included when MANAGED_SNI is being set.  Are these #ifs needed?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
```diff
@@ -0,0 +1,521 @@
...
+        }
+
+        /// <summary>
+        /// Set async callbacks
+        /// </summary>
+        /// <param name="receiveCallback">Receive callback</param>
+        /// <param name="sendCallback">Send callback</param>
+        public override void SetAsyncCallbacks(SNIAsyncCallback receiveCallback, SNIAsyncCallback sendCallback)
+        {
+        }
```
stephentoub: Is this purposefully a nop?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsManager.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace System.Data.SqlClient.SNI
+{
+    /// <summary>
+    /// Singleton to manage all MARS connection
+    /// </summary>
+    internal class SNIMarsManager
+    {
+        public static SNIMarsManager Singleton = new SNIMarsManager();
+        private Dictionary<SNIHandle, SNIMarsConnection> _connections = new Dictionary<SNIHandle, SNIMarsConnection>();
```
stephentoub: These should likely be readonly.

stephentoub: Rather than locking on _connections, should this be a ConcurrentDictionary?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIPacket.cs
```diff
@@ -0,0 +1,322 @@
...
+
+                    if (_length == 0)
+                    {
+                        SNILoadHandle.SingletonInstance.LastError = new SNIError(SNIProviders.TCP_PROV, 0, 0, "Connection was terminated");
+                        error = true;
+                    }
+                }
+
+                callback(this, error ? TdsEnums.SNI_ERROR : TdsEnums.SNI_SUCCESS);
+            });
```
stephentoub: This should probably explicitly specify ```, CancellationToken.None, TaskContinuationOptions.DenyChildAttach, TaskScheduler.Default``` unless you really want to be scheduling to the current scheduler.  

Separately, you could then also improve perf a bit by passing object state into the lambda and using that to provide the arguments to the function rather than using closed over state.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        /// <summary>
+        /// Enable MARS support on a connection
+        /// </summary>
+        /// <param name="handle">Connection handle</param>
+        /// <returns>SNI error code</returns>
+        public uint EnableMars(SNIHandle handle)
+        {
+            if (SNIMarsManager.Singleton.CreateMarsConnection(handle) == TdsEnums.SNI_SUCCESS_IO_PENDING)
```
stephentoub: CreateMarsConnection is going to throw an exception if handle was previously added already.  Is that expected, or did you want it to return SNI_ERROR in that case?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
```diff
@@ -0,0 +1,354 @@
...
+        /// </summary>
+        /// <param name="handle">SNI handle</param>
+        /// <param name="bufferSize">Buffer size</param>
+        /// <returns>SNI error code</returns>
+        public uint SetConnectionBufferSize(SNIHandle handle, uint bufferSize)
+        {
+            if (handle is SNITCPHandle)
+            {
+                (handle as SNITCPHandle).SetBufferSize((int)bufferSize);
+            }
```
stephentoub: It'd be a bit cleaner to do:
```C#
var snitcpHandle = handle as SNITCPHandle;
if (snitcpHandle != null)
{
    snitcpHandle.SetBufferSize((int)bufferSize);
}
```

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+                    _packetBytes -= 8;
+                }
+
+                if (count > _packetBytes)
+                {
+                    count = _packetBytes;
+                }
+            }
+
+            readBytes = _stream.Read(scratch, 0, count);
```
stephentoub: Do we completely control the stream that's being used here?  I'm just wondering if we need to be concerned with bad/maliicious streams returning bad data, e.g. a readBytes value that's negative or that's larger than count?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+
+            if (_encapsulate)
+            {
+                _packetBytes -= readBytes;
+            }
+
+            Buffer.BlockCopy(scratch, 0, buffer, offset, readBytes);
+            return readBytes;
+        }
```
stephentoub: Is this the stream type on which we're calling ReadAsync in some other places, e.g. in ReadFromStreamAsync?  If so, we should be overriding ReadAsync (and presumably WriteAsync and FlushAsync as well).  The base implementation will work, but it does so by simply queueing a work item that invokes Read/Write/Flush synchronously, which means blocking a thread, which defeats the scalability benefits of async.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+            while (count > 0)
+            {
+                // During the SSL negotiation phase, SSL is tunnelled over TDS packet type 0x12. After
+                // negotiation, the underlying socket only sees SSL frames.
+                //
+                if (_encapsulate)
+                {
+                    if (count > 4088)
+                    {
+                        currentCount = 4088;
```
stephentoub: What is this number 4088?  I'd suggest putting it into a named constant.  Same goes for some of the other magic numbers floating around the implementation.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Set stream length. 
+        /// </summary>
+        /// <param name="value">Length</param>
+        public override void SetLength(long value)
+        {
+            throw NotImplemented.ByDesignWithMessage("Not required");
```
stephentoub: Probably should be NotSupportedException.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// </summary>
+        public override long Position
+        {
+            get
+            {
+                throw NotImplemented.ByDesignWithMessage("Not required");
+            }
+            set
+            {
+                throw NotImplemented.ByDesignWithMessage("Not required");
```
stephentoub: Both of these should probably be NotSupportedException.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        /// <summary>
+        /// Seek in stream
+        /// </summary>
+        /// <param name="offset">Offset</param>
+        /// <param name="origin">Origin</param>
+        /// <returns>Position</returns>
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            throw NotImplemented.ByDesignWithMessage("Not required");
```
stephentoub: Ditto.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            get { return _stream.CanWrite; }
+        }
+
+        /// <summary>
+        /// Check if stream can be seeked
+        /// </summary>
+        public override bool CanSeek
+        {
+            get { return _stream.CanSeek; }
```
stephentoub: Since we're throwing from all of the seeking-related members, this should probably just return false.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        /// <summary>
+        /// Get stream length
+        /// </summary>
+        public override long Length
+        {
+            get
+            {
+                throw NotImplemented.ByDesignWithMessage("Not required");
```
stephentoub: Ditto on exception type.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
```diff
@@ -301,7 +304,11 @@ internal void Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandle
             _connHandler = connHandler;
             _loginWithFailover = withFailover;
 
+#if MANAGED_SNI
+            UInt32 sniStatus = SNILoadHandle.SingletonInstance.Status;
```
stephentoub: Can we just name the properties the same so that we don't need the #if: we just end up including the right file and the right things happen?  Or is there some reason we can't do that?

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
```diff
@@ -311,6 +318,8 @@ internal void Connect(ServerInfo serverInfo, SqlInternalConnectionTds connHandle
                 Debug.Assert(false, "SNI returned status != success, but no error thrown?");
             }
 
+#if MANAGED_SNI
+#else
```
stephentoub: #if !MANAGED_SNI ?

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
```diff
@@ -6243,7 +6306,11 @@ private void LoadSSPILibrary()
                     {
                         // use local for ref param to defer setting s_maxSSPILength until we know the call succeeded.
                         UInt32 maxLength = 0;
+#if MANAGED_SNI
+                        if (0 != SNIProxy.Singleton.InitializeSspiPackage(ref maxLength))
+#else
                         if (0 != SNINativeMethodWrapper.SNISecInitPackage(ref maxLength))
+#endif // MANAGED_SNI
```
stephentoub: In all of these #if...else...endif cases, it'd be nice if we could follow the same scheme we do elsewhere in the repo.  You have two files that expose the same surface area, you make calls to that surface area, and then you just include whichever file is appropriate for a given build.  That way, the calling code remains clean, it's obvious from the .csproj which support is being used, etc.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParserStateObject.cs
```diff
@@ -2596,7 +2724,16 @@ private void ReadSniError(TdsParserStateObject stateObj, UInt32 error)
                                 }
                                 else
                                 {
+#if MANAGED_SNI
+#if DEBUG
+                                    if (syncReadPacket != null)
+                                    {
+                                        Debug.Assert(syncReadPacket.IsInvalid, "unexpected syncReadPacket without corresponding SNIPacketRelease");
+                                    }
+#endif
```
stephentoub: Seems like this could all just be simplified to:
```C#
Debug.Assert(syncreadPacket == null || syncreadPacket.IsInvalid, ...);
```

src/System.Data.SqlClient/src/project.json
```diff
@@ -13,7 +13,7 @@
     "System.IO.FileSystem": "4.0.0",
     "System.Net.Primitives": "4.0.10",
     "System.Net.Security": "4.0.0-beta-*",
-    "System.Net.Sockets": "4.0.10",
+    "System.Net.Sockets": "4.0.10-beta-*",
```
stephentoub: Why did this change?

stephentoub:
Thanks, @saurabh500.  Did a pass through and left some comments.

stephentoub:
@dotnet-bot test this please (#3580)

src/System.Data.SqlClient/src/Interop/SNINativeMethodWrapper.cs
corivera: If we have multiple files laying around, then we'll start encountering a lot of maintenance issues later on. The code would also have to be pulled apart in such a way that it becomes un unreadable spaghetti mess. Using #ifdef controls is a much more elegant solution for us, and is what we do in Azure code anyway.

src/System.Data.SqlClient/src/System.Data.SqlClient.csproj
corivera: It's actually the opposite. There are a lot of features in SqlClient like LocalDB, Integrated Auth, and Named Pipes that are all Windows specific. So we're currently using the native codebase to get around these issues until we create managed solutions.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
corivera: Old leftover code. Removed.

corivera: Removed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNITcpHandle.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParserStateObject.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
corivera: Once the managed code has feature parity with our native code, these #if checks will no longer be necessary and can be removed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Yes, we control the underlying stream. This class is a wrapper around the streams used in TCPHandle.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNICommon.cs
corivera: We use this form of delegate in a lot places, so using a more descriptive type helps with readability. Instead of having to parse out what the Action does, you know it's supposed to be a callback.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
corivera: Fixed.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
divega: Would 

``` C#
(handle as SNITCPHandle)?.SetBufferSize((int)bufferSize);
```
be too much?

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
stephentoub: We can't use C# 6 in the repo yet.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsManager.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/LimitedConcurrencyLevelTaskScheduler.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNITcpHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIError.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Mars connections have their own callbacks. See constructor for SNIMarsConnection.

src/System.Data.SqlClient/src/project.json
corivera: This change isn't showing up in my local version, and there are no other changes mentioned in its history.

src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParser.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNILoadHandle.cs
corivera: Removed that string

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Added an issue for this.
https://github.com/dotnet/corefx/issues/3620

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs
corivera: Added an issue for this:
https://github.com/dotnet/corefx/issues/3621

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SslOverTdsStream.cs
corivera: Added an issue for this:
https://github.com/dotnet/corefx/issues/3622

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIPacket.cs
corivera: Fixed

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: ReceiveAsync currently always returns IO_PENDING. Packets could return synchronously, but we return IO_PENDING to handle all the async callbacks.

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: This session count gets reset for each new login, since a new Mars connection is setup each time. Let me find out what the max number of allowable parallel MARS queries is.

corivera: Looks like the max in native code is also just ushort.MAX

corivera: I'll add a size check for the session count, and throw an error if it's higher than ushort.Max

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsConnection.cs
corivera: Added an issue for this:
https://github.com/dotnet/corefx/issues/3624

src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIMarsHandle.cs
corivera: Added an issue for this:
https://github.com/dotnet/corefx/issues/3624

src/System.Data.SqlClient/src/project.json
saurabh500: @stephentoub The compilation with 4.0.10.0 goes well but we get runtime errors with 4.0.10.0 while the System.Net.Sockets is being looked for. 
For now I will leave this as is and I will open an issue to migrate this version to a higher version after testing.

stephentoub:
I left a few more small comments, but otherwise the updates LGTM.

saurabh500:
@stephentoub 
I have undone the changes to SqlMetadata which had reverted the Array.Copy changes.
I have opened an issue to handle the string movement to resx in https://github.com/dotnet/corefx/issues/3642

Other comments mentioned are taken care of.

Once the build passes,  I will merge the changes.

stephentoub:
Sounds good.  Thanks, @saurabh500.

natemcmaster:
@saurabh500 So does this mean SNI is working on xplat now? Or is there more remaining work?

saurabh500:
@natemcmaster I have tested the new managed functionality to connect to Azure SQL on Ubuntu to read Data from the server.
The testing was based on the CoreFx test execution instructions. https://github.com/dotnet/corefx/blob/master/Documentation/building/cross-platform-testing.md

I will be testing it with the sample project that you have created to see where things stand currently.
The managed SNI puts SqlClient in a better position compared to its earlier state. 
I will keep you updated.

PR #3582, opened 2015-10-01
Fix seg faults due to incorrect crypto DllImport signatures

### LINKED ISSUE #3581: Seg fault in X509Certificates in Linux CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2908/console
```
13:23:39 Starting:    System.Security.Cryptography.X509Certificates.Tests
13:23:40 ./run-test.sh: line 172:  8479 Segmentation fault      (core dumped) ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCategory
```

stephentoub:
Three DllImports were declared to return strings rather than IntPtr.  Returning a string like this should only be done when the native code is allocating a ```char*``` that must be freed by the consumer, and the CLR dutifully attempts to do so after marshaling the ```char*``` to the managed string.  However, these three DllImports all return ```const char*```s that shouldn't be freed, which was resulting in seg faults.

The fix is simply to declare the DllImports to return IntPtr, and then use Marshal.PtrToStringAnsi to do the marshaling.

Fix #3581
cc: @bartonjs, @nguerrera

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Crypto.cs
@@ -94,8 +94,16 @@ internal static partial class Interop
         [DllImport(Libraries.CryptoNative)]
         internal static extern void RecursiveFreeX509Stack(IntPtr stack);
 
-        [DllImport(Libraries.CryptoNative, CharSet = CharSet.Ansi)]
-        internal static extern string GetX509RootStorePath();
+        internal static string GetX509RootStorePath()
+        {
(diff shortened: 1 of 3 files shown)


stephentoub:
Build passed. Merging pre-review to unblock CI. I'll address any feedback separately.

bartonjs:
LGTM, thanks @stephentoub.

nguerrera:
LGTM  too.

nguerrera:
@yizhang82 Has a feature ever been considered for declaring masrhaling information such that a string is marshaled from a P/Invoke without being freed?

PR #3584, opened 2015-10-01
Make X509FindType.FindBySubjectKeyIdentifier correctly check for equality

bartonjs:
FindBySubjectKeyIdentifier only checked the byte[].Length and the value of the first byte; it returned false matches for an incremented key value.

Added a test that fails without this change to keep it from regressing.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/tests/FindTests.cs
+++ b/src/System.Security.Cryptography.X509Certificates/tests/FindTests.cs
@@ -387,6 +387,15 @@ namespace System.Security.Cryptography.X509Certificates.Tests
 
         [Fact]
         [ActiveIssue(1993, PlatformID.AnyUnix)]
+        public static void TestBySubjectKeyIdentifier_NoMatch_RightLength()
+        {
+            RunZeroMatchTest(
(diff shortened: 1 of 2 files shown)


bartonjs:
cc @AtsushiKan

src/System.Security.Cryptography.X509Certificates/tests/FindTests.cs
```diff
@@ -385,6 +385,15 @@ public static void TestBySubjectKeyIdentifier_NoMatch()
             RunZeroMatchTest(X509FindType.FindBySubjectKeyIdentifier, "");
         }
 
+		[Fact]
```
stephentoub: Nit: indentation

stephentoub:
LGTM

src/System.Security.Cryptography.X509Certificates/tests/FindTests.cs
bartonjs: Apparently on _this_ computer it was set to tabs, so there's a mix of whitespace where I pasted from a much bigger change.  Oops, fixed :).

AtsushiKan:
LGTM.

Issue #2604, opened 2015-08-03
run-test.sh deploys WP8 dataflow if present, fails ETW test


nguerrera:
On my machine (might have different behavior based on file system order), running with System.Threading.Tasks.Dataflow and System.Threading.Tasks.Dataflow.WP8 dlls in the corefx bin path, the test overlay gets the WP8 build, which doesn't have FEATURE_TRACING and the ETW test fails

nguerrera:
Can we get rid of the WP8 specifc build and make vNext of dataflow drop support for WP8?

stephentoub:
cc: @weshaggard

weshaggard:
I don't know the exact implications of why this was done so I don't have a particular opinion. In general I'm in favor of eliminating forks of the libraries if it is possible.

nguerrera:
Fixed by #3585.

PR #3585, opened 2015-10-01
Don't copy WP8 aseemblies to test overlay

### LINKED ISSUE #3520: System.Threading.Tasks.Dataflow.Tests.EtwTests.TestEtw failed in CI on OSX
This is a deterministic failure as far as I can tell.

```
MESSAGE:
Assert.Equal() Failure Expected: 0 Actual: 2
+++++++++++++++++++
STACK TRACE:
at System.Threading.Tasks.Dataflow.Tests.EtwTests.<>c__DisplayClass7.<TestEtw>b__1() at System.Diagnostics.Tracing.TestEventListener.RunWithCallback(Action`1 handler, Action body) at System.Threading.Tasks.Dataflow.Tests.EtwTests.TestEtw()
```

http://dotnet-ci.cloudapp.net/job/dotnet_corefx_mac_debug_tst/50/testReport/junit/System.Threading.Tasks.Dataflow.Tests/EtwTests/TestEtw/

ellismg:
Before we moved to using packer to slim down our test drop, we had
Jenkins exclude the System.Threading.Tasks.Dataflow.WP8 assemblies when
uploading build artifacts. This meant that the OSX and Linux jobs would
never copy them down and hence they could not be picked up by
run-test.sh (and overwrite the correct version of
System.Threading.Tasks.Dataflow when building the test overlay).

With packer, we now archive everything, so depending on the order in
which find discovered files, we may have copied over the WP8 version,
which does not have ETW support.

For now, update run-test.sh to exclude this folder when copying stuff
over. The correct long term fix is to just do what we do on
Windows (have the build produce the correct test layouts and get out of
this run-test.sh test overlay game).

Fixes #3520

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -148,7 +148,7 @@ create_test_overlay()
 	echo "Corefx binaries not found at $CoreFxBins"
 	exit 1
   fi
-  find $CoreFxBins -path "*ToolRuntime*" -prune -o -name '*.dll' -and -not -name "*Test*" -exec cp '{}' "$OverlayDir" ";"
+  find $CoreFxBins \( -path "*ToolRuntime*" -or -path "*System.Threading.Tasks.Dataflow.WP8*" \) -prune -o -name '*.dll' -and -not -name "*Test*" -exec cp '{}' "$OverlayDir" ";"
 
   # Then the native CoreFX binaries
   if [ ! -d $CoreFxNativeBins ]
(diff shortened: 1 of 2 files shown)


ellismg:
/cc @stephentoub 

This should address the WP8 issue.

stephentoub:
LGTM.  

(I don't know the history for why we have a separate WP8 project for this assembly; it'd be nice if we could just get rid of it and avoid such problems.)

ellismg:
@dotnet-bot test this please (test run hung due to #3592)

nguerrera:
This was also #2604. I'd like to just get rid of the WP8 assembly so not sure if I should keep that open.

ellismg:
@dotnet-bot test this please

ellismg:
@dotnet-bot test this please (maybe now all the lock files are updated correctly?)

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #3586, opened 2015-10-01
Increasing the positive case timeout for an unreliable Timer test

### LINKED ISSUE #3533: Timer_Change_BeforeDueTime_ChangesWhenTimerWillFire failed in CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_debug/1368/console
```
 TimerChangeTests.Timer_Change_BeforeDueTime_ChangesWhenTimerWillFire [FAIL]
05:19:31         Should have received a timer event after this new duration
05:19:31         Expected: True
05:19:31         Actual:   False
05:19:31         Stack Trace:
05:19:31            d:\j\workspace\dotnet_corefx_windows_debug\src\System.Threading.Timer\tests\TimerChangeTests.cs(94,0): at TimerChangeTests.Timer_Change_BeforeDueTime_ChangesWhenTimerWillFire()
```

sokket:
This test fails randomly and unreliably; it looks to be a timing issue so increasing the positive-case timeout should add some buffer time for the timer to fire. Since this is the positive case, if the timer fires quickly then the wait will be cut short and not slow down the run. This fixes #3533

/cc: @stephentoub

### DIFF
--- a/src/System.Threading.Timer/tests/TimerChangeTests.cs
+++ b/src/System.Threading.Timer/tests/TimerChangeTests.cs
@@ -91,7 +91,7 @@ public class TimerChangeTests
         {
             Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(100)), "The reset event should not have been set yet");
             t.Change(TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(-1));
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(500)), "Should have received a timer event after this new duration");
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(TimerFiringTests.MaxPositiveTimeoutInMs)), "Should have received a timer event after this new duration");
         }
     }
 }
--- a/src/System.Threading.Timer/tests/TimerFiringTests.cs
+++ b/src/System.Threading.Timer/tests/TimerFiringTests.cs
@@ -8,7 +8,7 @@ using Xunit;
 
 public class TimerFiringTests
 {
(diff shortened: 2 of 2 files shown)


src/System.Threading.Timer/tests/TimerChangeTests.cs
```diff
@@ -91,7 +91,7 @@ public void Timer_Change_BeforeDueTime_ChangesWhenTimerWillFire()
         {
             Assert.False(are.WaitOne(TimeSpan.FromMilliseconds(100)), "The reset event should not have been set yet");
             t.Change(TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(-1));
-            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(500)), "Should have received a timer event after this new duration");
+            Assert.True(are.WaitOne(TimeSpan.FromMilliseconds(750)), "Should have received a timer event after this new duration");
```
stephentoub: Since we expect the event to fire and will fail the test if we don't, should we use a much larger value? e.g. in TimerFiringTests we use MaxPositiveTimeoutInMs set to 30 seconds I believe for cases like this.

stephentoub:
One question/comment on the value chosen, otherwise LGTM.

src/System.Threading.Timer/tests/TimerChangeTests.cs
sokket: Good idea, I'll make that value visible internally and use it instead

sokket:
Merging, thanks Stephen!

PR #3602, opened 2015-10-02
Update lock files

ellismg:


bartonjs:
LGTM.  It sort of looks like you may have missed the one for SecureString (the only other hit for "00085" on search), but since we're trying to delete that library I'm not bothered by it much.

ellismg:
@bartonjs I don't even understand why that file is still there...

ellismg:
Opened #3603 to track cleaning up SecureString stuff.  I would delete it myself right now except I know that @pallavit had to do some work with the type internally in TFS, so I don't want to delete it because I don't understand if that would break some stuff on the TFS side.

PR #3608, opened 2015-10-02
Add tests for EnumerableQuery created from expression.

JonHanna:
Try to cover a large range of possible expressions.

### DIFF
--- a/src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
+++ b/src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
@@ -1,0 +1,509 @@
+﻿// Copyright (c) Jon Hanna. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Reflection;
+using Xunit;
+
+namespace System.Linq.Tests
+{
+    public class QueryFromExpressionTests
+    {
+        private class SimplePair : IEnumerable<int>
+        {
+            public int First { get; set; }
+
+            public int Second { get; set; }
+
+            public IEnumerator<int> GetEnumerator()
+            {
+                yield return First;
+                yield return Second;
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
+            }
+        }
+
+        private static IQueryProvider _prov = Enumerable.Empty<int>().AsQueryable().Provider;
+
+        [Fact]
+        public void ExpressionToQueryFromProvider()
+        {
+            Expression exp = Expression.Constant(Enumerable.Range(0, 2).AsQueryable());
+            IQueryable<int> q = _prov.CreateQuery<int>(exp);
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @VSadov

src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        {
+            Expression cond = Expression.Condition(
+                Expression.Equal(
+                    Expression.Modulo(Expression.Constant(1), Expression.Constant(2)),
+                    Expression.Constant(0)
+                   ),
+                Expression.Constant(Enumerable.Range(0, 2).AsQueryable()),
+                Expression.Constant(Enumerable.Range(3, 2).AsQueryable())
+            );
+            IQueryable<int> q = new EnumerableQuery<int>(cond);
```
stephentoub: Since the only difference between the two tests is a few characters on this line, you might consider making the test a Theory would a Boolean argument and just switching on that arg to determine whether to do new EnumerableQuery or _prov.CreateQuery.  I had to look hard at both tests to figure out what was different.

src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
```diff
@@ -0,0 +1,508 @@
...
+                    .Where(c => char.IsControl(c));
+            }
+        }
+
+        [Fact]
+        public void PropertyAccess()
+        {
+            Expression access = Expression.Property(null, typeof(QueryFromExpressionTests), "AsciiControlCharacters");
+            IQueryable<char> q = _prov.CreateQuery<char>(access);
+            Assert.Equal(Enumerable.Range(0, 32).Concat(Enumerable.Range(127, 1)).Select(i => (char)i), q);
```
stephentoub: Would it not be cleaner to just compare against the same ```Enumerable.Range(0, 128).Select(i => (char)i).Where(c => char.IsControl(c))``` so that we can easily see that we expect as output what was given as input?

src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
JonHanna: I think instead I'll just use a simpler expression here. What I intended to test was the expression working when passed to a constructor, whereas of the two ways to turn an expression into an `EnumerableQuery` I use the other (more provider-agnostic) approach in the rest of the tests. I copy-pasted the expression from the previous test, but since that's not what is being tested, it should be simpler.

src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
JonHanna: Yes, it would.

src/System.Linq.Queryable/tests/QueryFromExpressionTests.cs
JonHanna: I removed this one and start the set with an `ExpressionToQueryFromProvider` and an `ExpressionToQueryByConstructor` to cover the two types of query creation as simply as possible.

stephentoub:
LGTM. Thanks.

Issue #3580, opened 2015-10-01
NullReferenceException in System.Net.Sockets.Tests.Disconnect.Success on Windows


stephentoub:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_debug_prtest/4466/console
```C#
Discovering: System.Net.Sockets.AsyncServer.Tests
  Discovered:  System.Net.Sockets.AsyncServer.Tests
  Starting:    System.Net.Sockets.AsyncServer.Tests
     System.Net.Sockets.Tests.Disconnect.Success [FAIL]
        System.NullReferenceException : Object reference not set to an instance of an object.
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Net.Sockets\src\System\Net\Sockets\Socket.Windows.cs(86,0): at System.Net.Sockets.Socket.DisconnectExBlocking(IntPtr socketHandle, IntPtr overlapped, Int32 flags, Int32 reserved)
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Net.Sockets\src\System\Net\Sockets\SocketPal.Windows.cs(156,0): at System.Net.Sockets.SocketPal.Disconnect(Socket socket, SafeCloseSocket handle, Boolean reuseSocket)
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Net.Sockets\src\System\Net\Sockets\Socket.cs(2600,0): at System.Net.Sockets.Socket.Disconnect(Boolean reuseSocket)
           d:\j\workspace\dotnet_corefx_windows_debug_prtest\src\System.Net.Sockets.Legacy\tests\FunctionalTests\Disconnect.cs(42,0): at System.Net.Sockets.Tests.Disconnect.Success()
  Finished:    System.Net.Sockets.AsyncServer.Tests
```
cc: @davidsh, @cipop, @pgavlin

stephentoub:
Failed again here: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/4552/console

stephentoub:
Failed again here: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/4775/console
```
System.Net.Sockets.Tests.Disconnect.Success [FAIL]
        System.NullReferenceException : Object reference not set to an instance of an object.
        Stack Trace:
[...]

PR #3613, opened 2015-10-02
Fix and enable the DateTime parsing tests

tarekgh:
The test used to fail on Windows because we didn't have the update coreclr
package. after updating the package the test work fine.
The test used to fail on Linux because the Serbian cultures Date patterns
are wrong and not matching CLDR as it include spaces after the date separator
and in same time the date separator and time separartor are start same with the
same char ".".
that causing the parser to get confused and fail. we need to fix the date
pattern for such cultures when reading it from ICU or fix the parser to handle
such case but this can be risky as it can break other cases.

### DIFF
--- a/src/System.Runtime/tests/System/DateTime.cs
+++ b/src/System.Runtime/tests/System/DateTime.cs
@@ -340,7 +340,7 @@ public static unsafe class DateTimeTests
     [InlineData("sr-Latn-ME")]
     [InlineData("sr-Latn-RS")]
     [InlineData("sr-Latn-XK")]
-    [PlatformSpecific(PlatformID.Windows)] // ToDo: Serbian cultures in Linux have wrong date patterns. need to fix that.
+    [ActiveIssue(3616, PlatformID.AnyUnix)] 
     public static void TestSerbianCulturesParsing(string cultureName)
     {
         TestDateTimeParsingWithSpecialCultures(cultureName);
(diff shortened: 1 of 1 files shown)


src/System.Runtime/tests/System/DateTime.cs
```diff
@@ -326,17 +326,27 @@ public static void TestGetDateTimeFormats()
...
+    }
+
+    [Theory]
     [InlineData("sr-Cyrl-XK")]
     [InlineData("sr-Latn-ME")]
     [InlineData("sr-Latn-RS")]
     [InlineData("sr-Latn-XK")]
-    [ActiveIssue(3391)]
-    public static void TestDateTimeParsingWithSpecialCultures(string cultureName)
+    [PlatformSpecific(PlatformID.Windows)] // ToDo: Serbian cultures in Linux have wrong date patterns. need to fix that.
```
stephentoub: Rather than using PlatformSpecific, it seems like we should open an issue to track the fix for Unix, and here use ```[ActiveIssue(issueNumber, PlatformID.AnyUnix]```.

stephentoub:
Other than my one comment, LGTM.

src/System.Runtime/tests/System/DateTime.cs
tarekgh: That is good idea. I'll do that.

stephentoub:
@dotnet-bot test this please (https://github.com/dotnet/corefx/issues/3580)

PR #3615, opened 2015-10-02
Add explicit 0 index args to Array.Copy usage

stephentoub:
```Array.Copy``` is implemented to work with any ```System.Array```, not just ```T[]```.  When ```Array.Copy(sourceArr, destArray, length)``` is used, it delegates to ```Array.Copy(sourceArr, sourceIndex, destArray, destIndex, length)```, and as such needs to query for the lower-bound of the supplied arrays, since although a ```T[]``` in C# always has a lower bound of 0, an arbitrary ```Array``` may not.  Given this, instead of calling ```Array.Copy(src, dest, length)``` where ```src``` and ```dest``` are both known to be ```T[]```, it's slightly more efficient to just call ```Array.Copy(src, 0, dest, 0, length)```, avoiding the need for the ```Array.GetLowerBound``` calls made while delegating to the 5-arg version.

I noticed one occurrence of this in some code I was looking at, and while fixing that, I fixed all of them across corefx\src.

### DIFF
--- a/src/Common/src/Interop/Windows/BCrypt/BCryptAlgorithmCache.cs
+++ b/src/Common/src/Interop/Windows/BCrypt/BCryptAlgorithmCache.cs
@@ -39,7 +39,7 @@ internal partial class Interop
     
                 Entry[] newCache = new Entry[cache.Length + 1];
                 Entry newEntry = new Entry(hashAlgorithmId, flags, safeBCryptAlgorithmHandle);
-                Array.Copy(cache, newCache, cache.Length);
+                Array.Copy(cache, 0, newCache, 0, cache.Length);
                 newCache[newCache.Length - 1] = newEntry;
     
                 // Atomically overwrite the cache with our new cache. It's possible some other thread raced to add a new entry with us - if so, one of the new entries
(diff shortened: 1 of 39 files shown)


mellinoe:
LGTM

nguerrera:
LGTM too. 

Possible follow-up: should we consider adding generic T[] overloads? Or maybe code generators could make this transformation as an optimization?

stephentoub:
> should we consider adding generic T[] overloads?

Yes :smile:

stephentoub:
Thanks, guys.

weshaggard:
@stephentoub what kind of performance improvements do we see here? Is it enough to start using an analyzer or something else that helps us ensure always do this?

stephentoub:
> what kind of performance improvements do we see here? 

On my machine, for a micro-benchmark that just does Array.Copy with int[]s and comparing without or with the explicit 0 args:
- Length = 1, ~33% faster with the explicit 0 args
- Length = 10, ~28% faster with the explicit 0 args
- Length = 100, ~17% faster with the explicit 0 args
- Length = 1000, ~5% faster with the explicit 0 args

> Is it enough to start using an analyzer or something else that helps us ensure always do this?

Obviously most uses of Array.Copy aren't in a situation that would see improvements akin to this micro-benchmark, and it's likely that most of the changes I made have no measurable difference on the all-up scenario.  However, there's essentially no downside (two additional ldc.i4.0 IL instructions at the call site), and falls into the bucket of simply eliminating unnecessary work (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Array.cs#L268-L280), and given these numbers it can make a difference for certain uses, so as we're looking at analyzers to add for best usage of the BCL, I do think an analyzer for this would be good.

PR #3619, opened 2015-10-02
Updated perf tests and documentation

ianhays:
- Updated getting started perf docs
- Removed caching of collections for tests; my results found the caching of negligible benefit
- Changed from MemberData to InlineData where possible to simplify matching method signatures for the perf analysis tools.
- Fixed the namespace for FileSystem tests to be the same as the other FS tests

Note: Will also include new build tools version after merging of https://github.com/dotnet/buildtools/pull/299

### DIFF
--- a/Documentation/project-docs/performance-tests.md
+++ b/Documentation/project-docs/performance-tests.md
@@ -5,16 +5,89 @@ This document contains instructions for building, running, and adding Performanc
 
 Requirements
 --------------------
-
-To run performance tests, .NET portable v5.0 is required. This library is included in [the Visual Studio Community 2015 download](https://www.visualstudio.com/products/visual-studio-community-vs). To get the correct packages during installation, follow these steps after opening the installer:
+### Windows
+To run performance tests on Windows, .NET portable v5.0 is required. This library is included in [the Visual Studio Community 2015 download](https://www.visualstudio.com/products/visual-studio-community-vs). To get the correct packages during installation, follow these steps after opening the installer:
 1. Select "Custom Installation" if no installation is present, or "Modify" otherwise
 2. Check the "Universal Windows App Development Tools" box under the "Windows and Web Development" menu
 3. Install
+### Linux
+Performance tests on Linux require all of the same steps as they do for regular xunit tests - check out the linux instructions [here](https://github.com/dotnet/corefx/blob/master/Documentation/building/unix-instructions.md). Once you can have a directory on your Linux machine with a working corerun and xunit.console.netcore.exe (as well as the test dll containing your perf tests!), you only need to run the following command:
+
+`dnu commands install Microsoft.DotNet.xunit.performance.runner.dnx 1.0.0-alpha-build0021 -f https://www.myget.org/F/dotnet-buildtools/`
+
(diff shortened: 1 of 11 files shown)


Documentation/project-docs/performance-tests.md
```diff
@@ -5,16 +5,89 @@ This document contains instructions for building, running, and adding Performanc
...
+### Windows
+To run performance tests on Windows, .NET portable v5.0 is required. This library is included in [the Visual Studio Community 2015 download](https://www.visualstudio.com/products/visual-studio-community-vs). To get the correct packages during installation, follow these steps after opening the installer:
 1. Select "Custom Installation" if no installation is present, or "Modify" otherwise
 2. Check the "Universal Windows App Development Tools" box under the "Windows and Web Development" menu
 3. Install
+### Linux
+Performance tests on Linux require all of the same steps as they do for regular xunit tests - check out the linux instructions [here](https://github.com/dotnet/corefx/blob/master/Documentation/building/unix-instructions.md). Once you can have a directory on your Linux machine with a working corerun and xunit.console.netcore.exe (as well as the test dll containing your perf tests!), you only need to run the following command:
```
stephentoub: Nit: I'd suggest changing "check out" to "see" or something like that... "check out" might be confusing with regards to a source control system.

Documentation/project-docs/performance-tests.md
```diff
@@ -5,16 +5,89 @@ This document contains instructions for building, running, and adding Performanc
...
+
+`dnu commands install Microsoft.DotNet.xunit.performance.runner.dnx 1.0.0-alpha-build0021 -f https://www.myget.org/F/dotnet-buildtools/`
+
+Be careful that your mscorlib and test dlls were compiled using the "/p:Configuration=Release" property. Otherwise you may get skewed results.
```
stephentoub: And libcoreclr

Documentation/project-docs/performance-tests.md
ianhays: Good point! Changing.

src/System.Console/tests/Performance/Perf.Console.cs
```diff
@@ -0,0 +1,91 @@
...
+                    {
+                        streams[0 + i * 4] = Console.OpenStandardInput(); streams[1 + i * 4] = Console.OpenStandardInput();
+                        streams[2 + i * 4] = Console.OpenStandardInput(); streams[3 + i * 4] = Console.OpenStandardInput();
+                    }
+                }
+                foreach (Stream s in streams)
+                {
+                    Assert.NotNull(s);
+                }
+            }
```
stephentoub: We should Dispose of all of the Streams here.

Documentation/project-docs/performance-tests.md
ianhays: Worth adding, thanks

src/System.Console/tests/Performance/Perf.Console.cs
ianhays: We most certainly should. I removed the disposal and added that assert to debug a problem i was having with the test (was doing my math wrong). Looks like I forgot to revert that change. Fixing.

stephentoub:
A few comments, otherwise LGTM.

ianhays:
Thanks @stephentoub, updated with feedback

PR #3627, opened 2015-10-02
Implement X509Certificate2Collection.Find for Unix.

bartonjs:
* All of the object->rich type parsing code is in the common FindPal class.
* Each of the X509FindType types dispatches pre-parsed data to an IFindPal instance.
* There are still two sources of disabled tests for the Unix port
    * Issue 1985: Distingushed Names have (e.g.) "S=Washington" on Windows, "ST=Washington" on Unix.
    * Issue 3099: The Unix and Windows implementations are choosing different certificates out of a multi-file PFX
* Across the various [Theory] components, a little over 100 test cases were added to verify consistent behavior


bartonjs:
The tests claimed to be disabled as a part of #1993, but that really tracks `NotImplementedException`s, so there's no issue paperwork to really deal with here.

bartonjs:
cc: @AtsushiKan

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
```diff
@@ -0,0 +1,275 @@
...
+                return (X509KeyUsageFlags)(int)findValue;
+
+            if (findValue is uint)
+                return (X509KeyUsageFlags)(uint)findValue;
+
+            string findValueString = findValue as string;
+
+            if (findValueString != null)
+            {
+                if (findValueString.Equals("DigitalSignature", StringComparison.OrdinalIgnoreCase))
```
justinvp: Could these mappings be in a static readonly Dictionary for faster lookups? (I suspect the Dictionary lookup would be faster than all the if checks, especially for strings towards the end of the if checks, but haven't measured to confirm).

justinvp: Or a `switch` on `findValueString.Length` would be really fast (though, it's more verbose, so might not be worth it depending on how often this code is called)...

Something like:

```c#
switch (findValueString.Length)
{
    case 16:
        switch (char.ToUpperInvariant(findValueString[1]))
        {
            case 'I':
                return ConfirmedX509KeyUsage(findValueString, "DigitalSignature", X509KeyUsageFlags.DigitalSignature);

            case 'A':
                return ConfirmedX509KeyUsage(findValueString, "DataEncipherment", X509KeyUsageFlags.DataEncipherment);
        }
        break;

    case 15:
        return ConfirmedX509KeyUsage(findValueString, "KeyEncipherment", X509KeyUsageFlags.KeyEncipherment);

    case 14:
        return ConfirmedX509KeyUsage(findValueString, "NonRepudiation", X509KeyUsageFlags.NonRepudiation);

    case 12:
        switch (char.ToUpperInvariant(findValueString[0]))
        {
            case 'K':
                return ConfirmedX509KeyUsage(findValueString, "KeyAgreement", X509KeyUsageFlags.KeyAgreement);

            case 'E':
                return ConfirmedX509KeyUsage(findValueString, "EncipherOnly", X509KeyUsageFlags.EncipherOnly);

            case 'D':
                return ConfirmedX509KeyUsage(findValueString, "DecipherOnly", X509KeyUsageFlags.DecipherOnly);
        }
        break;

    case 11:
        return ConfirmedX509KeyUsage(findValueString, "KeyCertSign", X509KeyUsageFlags.KeyCertSign);

    case 7:
        return ConfirmedX509KeyUsage(findValueString, "CrlSign", X509KeyUsageFlags.CrlSign);
}
```

With a helper like this:

```c#
private static X509KeyUsageFlags ConfirmedX509KeyUsage(string findValueString, string flagString, X509KeyUsageFlags flags)
{
    Debug.Assert(findValueString != null);
    Debug.Assert(flagString != null);

    if (flagString.Equals(findValueString, StringComparison.OrdinalIgnoreCase))
    {
        return flags;
    }

    throw new CryptographicException(SR.Cryptography_X509_InvalidFindValue);
}
```

Perhaps with more `Debug.Asserts` to verify the lengths and passed in `flagString` and `flags` match.

src/System.Security.Cryptography.X509Certificates/tests/FindTests.cs
```diff
@@ -84,9 +86,13 @@ private static void RunSingleMatchTest_PfxCer(X509FindType findType, object find
                 Assert.Equal(expected, match);
                 Assert.NotSame(expected, match);
 
-                // Verify that the find result and original are linked, not just equal.
-                match.FriendlyName = "HAHA";
-                Assert.Equal("HAHA", expected.FriendlyName);
+                // FriendlyName is Windows-only.
```
AtsushiKan: Could we use some other property to test the linkage, then?

bartonjs: Nope, FriendlyName and Archived are the only mutatable properties on Windows, and there aren't any on non-Windows.  So there's not really a notion of being linked that needs to be preserved on Unix.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: While I agree that the jump table is probably a lot faster, dictionary's probably sufficient :)

src/Common/src/Interop/Unix/libcrypto/Interop.ASN1.Print.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        [DllImport(Libraries.LibCrypto)]
+        private static extern int ASN1_STRING_print_ex(SafeBioHandle bio, SafeAsn1StringHandle str, Asn1StringPrintFlags flags);
+
+        internal static unsafe string DerStringToManagedString(
+            byte[] derString,
+            Asn1StringTypeFlags flags=AnyTextStringType)
+        {
+            SafeAsn1StringHandle asn1String = OpenSslD2I(
+                (zero, ppin, len) => d2i_ASN1_type_bytes(zero, ppin, len, flags),
```
stephentoub: This is still going to create a closure and allocate a delegate due to ```flags```... is that intentional?  This might not be a critical path we need to worry about, so up to you.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        private static string ConfirmedOidValue(IFindPal findPal, object findValue, OidGroup oidGroup)
+        {
+            string maybeOid = ConfirmedCast<string>(findValue);
+
+            if (maybeOid.Length == 0)
+            {
+                throw new ArgumentException(SR.Argument_InvalidOidValue);
+            }
```
stephentoub: Just verifying that ArgumentException is the right exception type for this situation?  Is the OID actually coming in through some public API, and it's the argument to that method we're saying is invalid?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
```diff
@@ -0,0 +1,268 @@
...
+                throw new ArgumentException(SR.Argument_InvalidOidValue);
+            if (keyValue[1] != '.' || keyValue[len - 1] == '.') // should not end in a dot
+                throw new ArgumentException(SR.Argument_InvalidOidValue);
+
+            // While characters 0 and 1 were both validated, start at character 1 to validate
+            // that there aren't two dots in a row.
+            for (int i = 1; i < len; i++)
+            {
+                // ensure every character is either a digit or a dot
+                if (char.IsDigit(keyValue[i]))
```
stephentoub: IsDigit returns true for characters besides '0'-'9'... is that ok?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
```diff
@@ -0,0 +1,268 @@
...
+                    throw new ArgumentException(SR.Argument_InvalidOidValue);
+            }
+        }
+
+        internal static BigInteger PositiveBigIntegerFromByteArray(byte[] bytes)
+        {
+            // To prevent the big integer from misinterpreted as a negative number,
+            // add a "leading 0" to the byte array.
+            // Since BigInteger(bytes[]) requires a little-endian byte array,
+            // the "leading 0" actually goes at the end of the array.
```
stephentoub: Why isn't this done when the last character is < 0x80?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
```diff
@@ -0,0 +1,268 @@
...
+            BigInteger ten = new BigInteger(10);
+            BigInteger accum = BigInteger.Zero;
+
+            foreach (char c in decimalString)
+            {
+                if (c >= '0' && c <= '9')
+                {
+                    accum = BigInteger.Multiply(accum, ten);
+                    accum = BigInteger.Add(accum, c - '0');
+                }
```
stephentoub: else?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
```diff
@@ -0,0 +1,331 @@
...
+                return oid.Value;
+            }
+
+            FindPal.ValidateOidValue(maybeOid);
+            return maybeOid;
+        }
+
+        public void FindByThumbprint(byte[] thumbprint)
+        {
+            FindCore(cert => cert.GetCertHash().ContentsEqual(thumbprint));
```
stephentoub: Again, may not matter at all, but this is going to allocate a closure/delegate due to capturing thumbprint.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+            fixed (char* pSubjectName = subjectName)
+            {
+                FindCore(CertFindType.CERT_FIND_SUBJECT_STR, pSubjectName);
+            }
+        }
+
+        public void FindBySubjectDistinguishedName(string subjectDistinguishedName)
+        {
+            FindCore(
+                delegate(SafeCertContextHandle pCertContext)
```
stephentoub: Nit: Why the anonymous method syntax here when you used the lambda syntax elsewhere?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+                {
+                    byte[] actual = pCertContext.CertContext->pCertInfo->SerialNumber.ToByteArray();
+
+                    // Convert to BigInteger as the comparison must not fail due to spurious leading zeros
+                    BigInteger actualAsBigInteger = PositiveBigIntegerFromByteArray(actual);
+
+                    GC.KeepAlive(pCertContext);
```
stephentoub: Why the KeepAlive?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+            FindCore(
+                delegate(SafeCertContextHandle pCertContext)
+                {
+                    // The template name can have 2 different formats: V1 format (<= Win2K) is just a string
+                    // V2 format (XP only) can be a friendly name or an OID.
+                    // An example of Template Name can be "ClientAuth".
+
+                    bool foundMatch = false;
+                    CERT_INFO* pCertInfo = pCertContext.CertContext->pCertInfo;
+                    {
```
stephentoub: Why the extra scope?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+                    int numOids;
+                    int cbData = 0;
+                    if (!Interop.crypt32.CertGetValidUsages(1, ref pCertContext, out numOids, null, ref cbData))
+                        return false;
+
+                    // -1 means the certificate is good for all usages.
+                    if (numOids == -1)
+                        return true;
+
+                    fixed (byte* pOidsPointer = new byte[cbData])
```
stephentoub: Is there any chance cbData here is 0?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+        }
+
+        public void Dispose()
+        {
+            _storePal.Dispose();
+        }
+
+        private void FindCore(Func<SafeCertContextHandle, bool> filter)
+        {
+            unsafe
```
stephentoub: Nit: You have it in the method signature in other FindCore methods... why not here, too?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+        private unsafe void FindCore(CertFindType dwFindType, void* pvFindPara, Func<SafeCertContextHandle, bool> filter = null)
+        {
+            SafeCertStoreHandle findResults = Interop.crypt32.CertOpenStore(
+                CertStoreProvider.CERT_STORE_PROV_MEMORY,
+                CertEncodingType.All,
+                IntPtr.Zero,
+                CertStoreFlags.CERT_STORE_ENUM_ARCHIVED_FLAG | CertStoreFlags.CERT_STORE_CREATE_NEW_FLAG,
+                null);
+            if (findResults.IsInvalid)
+                throw Marshal.GetHRForLastWin32Error().ToCryptographicException(); ;
```
stephentoub: Nit: extra ;

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
```diff
@@ -0,0 +1,405 @@
...
+                new TimeSpan(0, 0, 0));
+
+            if (chainPal == null)
+                return false;
+
+            using (chainPal)
+            {
+                Exception verificationException;
+                bool? verified = chainPal.Verify(X509VerificationFlags.NoFlag, out verificationException);
+                if (!(verified.HasValue && verified.Value))
```
stephentoub:
```C#
if (!verified.GetValueOrDefault())
```
?

src/Common/src/Interop/Unix/libcrypto/Interop.ASN1.Print.cs
bartonjs: It there a way that wouldn't, without inlining the code from OpenSslD2I?

stephentoub: You'd need to change the signature of OpenSslD2I, such that both OpenSslD2I and the delegate provided to it both take an additional state argument... then you can pass flags as that state argument and have it threaded through to the delegate invocation.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: Yep.  `certs.Find(X509FindType.FindByExtension, "", false);`

stephentoub: Ok, cool.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: Probably not what the original author intended, but it's the behavior desktop has. (http://referencesource.microsoft.com/#System/security/system/security/cryptography/x509/x509utils.cs,6472ac0799e53d68).  I'll have to try a non-ascii value and see if it matches on desktop, I suppose.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: Because it's already going to be interpreted as positive.  Just ask you asked for back in June :) https://github.com/dotnet/corefx/pull/1990/files#r32145804

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: Else ignored, sadly.  It's one of the desktop weird input compatibility tests.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
stephentoub: Cool; in that case, can you add that to the comment?  The 0x80 looks magical.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
bartonjs: > Why the anonymous method syntax here when you used the lambda syntax elsewhere

These were just moved from the file where they were to this new location during the PAL restructuring.  They were declared delegate in the desktop code, probably because they predated lambdas by about 3 major releases :).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
bartonjs: > Why the extra scope?

Looks like it's just showing the V1 case as a self-contained block.  I just moved it :).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
bartonjs: In context, I feel like 'no'.  If there were no valid usages then numOids would have been -1, otherwise there would be non-zero space required for the OID extraction.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs
bartonjs: > You have [unsafe] in the method signature in other FindCore methods... why not here, too?

The other methods take a void* parameter, so they require it.  By not putting `unsafe` in the declaration here it allows the callers using this one to not also need to say unsafe. 

At least, that's the motivation I assume, since I just moved the existing code to a new file :)

stephentoub: > By not putting unsafe in the declaration here it allows the callers using this one to not also need to say unsafe.

unsafe in the declaration has no impact on callers; it's functionally identical to unsafe instead being used around the entire body of the method.

bartonjs: ... I thought something sounded off when I said that.  Justifying other people's code is hard :).

I've made this change, and a few other of the style comments from the Windows code, in a separate commit; to try to keep "move" vs "change" as separate.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/FindPal.cs
bartonjs: \u0662 (arabic 2) doesn't work instead of '2' on Windows, or Unix, returning the empty set instead of complaining the OID is invalid.  Adding a compatibility test, and just leaving this as matching Desktop's behavior.

bartonjs:
Pushed some new commits.
* Corrected the typo in the base commit title
* Broke the PositiveBigIntegerFromByteArray memory improvement out to a separate commit, to track move vs-change
* Added @justinvp's suggested initial capacity and assert to @justinvp's suggested dictionary :smile:.
* Made some style changes to the Windows FindPal, left them in a separate commit from the base.

stephentoub:
LGTM

PR #3631, opened 2015-10-02
Add test coverage for ReadOnlyDictionary

### LINKED ISSUE #990: Improve code coverage for System.ObjectModel
According to the current code coverage report, this library is only at 30%. Our general goal is closer to 90%, though this will depend on the library. We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

alexperovich:
Add additional test coverage for ReadOnlyDictionary improving System.ObjectModel assembly coverage from 54% to 73%.

Fix #990

### DIFF
--- a/src/Common/tests/Collections/ICollectionTest.cs
+++ b/src/Common/tests/Collections/ICollectionTest.cs
@@ -307,10 +307,7 @@ namespace Tests.Collections
             ICollection<T> collection = GetCollection(items) as ICollection<T>;
             if (collection == null)
                 return;
-            foreach (var item in items)
-            {
-                Assert.True(collection.Contains((T)item));
-            }
+            Assert.All(items, item => Assert.True(collection.Contains((T) item)));
         }
 
         internal class MyInvalidReferenceType
--- a/src/Common/tests/Collections/ICollectionTest.cs
+++ b/src/Common/tests/Collections/ICollectionTest.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using Xunit;
 using Xunit.Sdk;
@@ -296,6 +297,19 @@ namespace Tests.Collections
                     collection.Count);
                 collection.CopyTo(itemArray, 0);
                 CollectionAssert.Equal(items, itemArray);
+            }
+        }
+
+        [Fact]
+        public void CollectionShouldContainAllItems()
+        {
+            object[] items = GenerateItems(16);
+            ICollection<T> collection = GetCollection(items) as ICollection<T>;
+            if (collection == null)
+                return;
+            foreach (var item in items)
+            {
+                Assert.True(collection.Contains((T)item));
             }
         }
 
--- a/src/Common/tests/Collections/IDictionaryTest.cs
+++ b/src/Common/tests/Collections/IDictionaryTest.cs
@@ -28,12 +28,13 @@ namespace Tests.Collections
         public void KeysShouldBeCorrect(int count)
         {
             object[] items = GenerateItems(count);
-            IDictionary<TKey, TValue> dict = GetDictionary(items);
             var expectedKeys = new TKey[items.Length];
             for (int i = 0; i < items.Length; ++i)
             {
                 expectedKeys[i] = ((KeyValuePair<TKey, TValue>)items[i]).Key;
             }
+
+            IDictionary<TKey, TValue> dict = GetDictionary(items);
 
             CollectionAssert.Equal(expectedKeys, dict.Keys);
 
@@ -51,12 +52,13 @@ namespace Tests.Collections
         public void ValuesShouldBeCorrect(int count)
         {
             object[] items = GenerateItems(count);
-            IDictionary<TKey, TValue> dict = GetDictionary(items);
             var expectedValues = new TValue[items.Length];
             for (int i = 0; i < items.Length; ++i)
             {
                 expectedValues[i] = ((KeyValuePair<TKey, TValue>)items[i]).Value;
             }
+
+            IDictionary<TKey, TValue> dict = GetDictionary(items);
 
             CollectionAssert.Equal(expectedValues, dict.Values);
 
@@ -175,6 +177,7 @@ namespace Tests.Collections
                         Assert.Equal(pair.Key, entry.Key);
                         Assert.Equal(pair.Value, entry.Value);
                     }
+                    Assert.False(dictEnumerator.MoveNext());
                     dictEnumerator.Reset();
                     enumerator.Reset();
                 }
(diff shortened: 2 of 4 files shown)


src/Common/tests/Collections/ICollectionTest.cs
```diff
@@ -299,6 +300,19 @@ public void CopyToValidTypes()
             }
         }
 
+        [Fact]
+        public void CollectionShouldContainAllItems()
+        {
+            object[] items = GenerateItems(16);
+            ICollection<T> collection = GetCollection(items) as ICollection<T>;
+            if (collection == null)
+                return;
```
stephentoub: Why are these type checks necessary?

src/Common/tests/Collections/ICollectionTest.cs
```diff
@@ -299,6 +300,19 @@ public void CopyToValidTypes()
...
+            foreach (var item in items)
+            {
+                Assert.True(collection.Contains((T)item));
```
stephentoub: Could you do something like:
```C#
Assert.All(items, item => Assert.True(collection.Contains(item)));
```
instead?

src/Common/tests/Collections/IDictionaryTest.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            object[] items = GenerateItems(count);
+            IDictionary<TKey, TValue> dict = GetDictionary(items);
+            var expectedKeys = new TKey[items.Length];
+            for (int i = 0; i < items.Length; ++i)
+            {
+                expectedKeys[i] = ((KeyValuePair<TKey, TValue>)items[i]).Key;
+            }
+
+            CollectionAssert.Equal(expectedKeys, dict.Keys);
```
stephentoub: Is this different from Assert.Equal that takes two IEnumerables? e.g. does this not care about ordering?

src/Common/tests/Collections/IDictionaryTest.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(16)]
+        [InlineData(100)]
+        public void KeysShouldBeCorrect(int count)
+        {
+            object[] items = GenerateItems(count);
+            IDictionary<TKey, TValue> dict = GetDictionary(items);
```
stephentoub: This ```dict``` isn't used until the assert below... consider moving it down closer to where it's used.

src/Common/tests/Collections/IDictionaryTest.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(16)]
+        [InlineData(100)]
+        public void ValuesShouldBeCorrect(int count)
+        {
+            object[] items = GenerateItems(count);
+            IDictionary<TKey, TValue> dict = GetDictionary(items);
```
stephentoub: Same comments/questions as above.  I see these same patterns are repeated, so I won't comment on them further.

src/Common/tests/Collections/IDictionaryTest.cs
```diff
@@ -0,0 +1,266 @@
...
+                        Assert.True(dictEnumerator.MoveNext());
+                        var pair = (KeyValuePair<TKey, TValue>) enumerator.Current;
+                        Assert.Equal(dictEnumerator.Current, dictEnumerator.Entry);
+                        var entry = dictEnumerator.Entry;
+                        Assert.Equal(pair.Key, dictEnumerator.Key);
+                        Assert.Equal(pair.Value, dictEnumerator.Value);
+                        Assert.Equal(pair.Key, entry.Key);
+                        Assert.Equal(pair.Value, entry.Value);
+                    }
+                    dictEnumerator.Reset();
```
stephentoub: Do you want to assert here that dictEnumerator.MoveNext is false?

src/System.ObjectModel/tests/ReadOnlyDictionary/ReadOnlyDictionaryTests.cs
```diff
@@ -217,6 +218,186 @@ public static void DebuggerAttributeTests()
...
+        }
+
+        public ICollection<TValue> Values
+        {
+            get { return _inner.Values; }
+        }
+    }
+
+    public class ReadOnlyDictionaryOverNonGenericTests
+        : IDictionaryTest<string, int>
```
stephentoub: You probably are just following an existing pattern, but I find it very confusing that the base test class names begin with an ```I```.

stephentoub:
Left some comments, but otherwise LGTM.  My biggest complaints are to do with the existing patterns/confusion employed by the existing tests, and I realize you're just expanding on what's already there.

src/Common/tests/Collections/ICollectionTest.cs
alexperovich: The collection type is not required to implement generic ICollection. I didn't want to assume that it would.

src/Common/tests/Collections/ICollectionTest.cs
alexperovich: done

src/Common/tests/Collections/IDictionaryTest.cs
alexperovich: This is the same as Assert.Equal, except this one accepts non generic IEnumerable. Assert.Equal only accepts generic IEnumerable.

src/Common/tests/Collections/IDictionaryTest.cs
alexperovich: done

src/Common/tests/Collections/IDictionaryTest.cs
alexperovich: done

src/System.ObjectModel/tests/ReadOnlyDictionary/ReadOnlyDictionaryTests.cs
alexperovich: The intent here is that the base test class defines tests over the IDictionary interface that should hold true for any given implementation of IDictionary. What would be a better name?

src/Common/tests/Collections/ICollectionTest.cs
stephentoub: Ok.

src/System.ObjectModel/tests/ReadOnlyDictionary/ReadOnlyDictionaryTests.cs
stephentoub: DictionaryTests?  BaseIDictionaryTests? etc.  To a casual reader/reviewer, it just looks like the class is implementing an interface rather than deriving from a base class that actually exposes [Facts].

stephentoub:
Since it's just following existing conventions, I'm going to go ahead and merge this.  There's a bunch of other things to be cleaned up in these tests; this naming can be one piece of that.

Issue #3569, opened 2015-10-01
System.Runtime.Loader.Tests.AssemblyLoadContextTest.InitializeDefaultContextTest failing on Linux in CI


pgavlin:
For example:

http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2896/

pgavlin:
@stephentoub: who should take a look at this one?

stephentoub:
@kouvel, I think you recently made some changes here.  Could you please take a look?  This is causing a bunch of corefx runs to fail, e.g. http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst/1052/testReport/junit/System.Runtime.Loader.Tests/AssemblyLoadContextTest/InitializeDefaultContextTest/

kouvel:
It looks like this was caused by my change e780495d6e7575812fd5c42d4125c35b6054fa95. I had a clean test run on the PR though, so maybe in combination with another change. I'll investigate.

kouvel:
It certainly was my change. The issue is that removing the PKT check in the binder when binding to assemblies on the TPA list means when the app executable is unsigned and alongside corerun, it's treated as a platform assembly and the assembly load context doesn't get locked.

This seems to be by design as far as the binder is concerned. If the app executable were signed or in a different directory, the behavior would differ. @jkotas suggested a way to make the behavior more consistent would be to add assemblies in the app executable's directory to the TPA list.

I'm going to try that, meanwhile I'll disable the test to unblock the CI.

stephentoub:
Thanks, @kouvel. I already disabled the test this morning.

kouvel:
Oh that's great, thanks!

kouvel:
Fixed

Issue #3439, opened 2015-09-24
Replace Microsoft.DotNet.CoreCLR dependency with Microsoft.NETCore.Runtime.CoreCLR


nguerrera:
Remove dotnet-coreclr feed from dir.props and NuGet.config once this is done as well.

See https://github.com/dotnet/corefx/pull/3413#discussion_r40368451

cc @weshaggard

weshaggard:
We have replaced all know usages of this so closing.

PR #3632, opened 2015-10-02
Fix assembly load context InitializeDefaultContext test

kouvel:
Updated the test and enabled it on Unixes. Disabled it on Windows, as the CoreCLR binaries being used to run the tests on Windows appear to be old. Once the repo starts pulling a more recent version of CoreCLR binaries, I'll re-enable the test on Windows.

Related to #3569

### DIFF
--- a/src/System.Runtime.Loader/tests/AssemblyLoadContextTest.cs
+++ b/src/System.Runtime.Loader/tests/AssemblyLoadContextTest.cs
@@ -109,16 +109,22 @@ namespace System.Runtime.Loader.Tests
             Assert.NotNull(context);
         }
 
-        [ActiveIssue(3569, PlatformID.AnyUnix)]
+        [ActiveIssue(3569, PlatformID.Windows)]
         [Fact]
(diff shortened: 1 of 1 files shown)


kouvel:
@stephentoub, regarding the CoreCLR binaries that are used for running CoreFX tests on Windows, where do they come from?

I noticed that the test was disabled only outside Windows, and was wondering why it's passing on Windows, as the behavior there should be the same. It looks like the CoreCLR binaries the tests are using are old and don't have my changes. When I replace them with my own built binaries the test fails on Windows too, as expected.

kouvel:
CC @sergiy-k

mellinoe:
I guess the old CoreCLR binaries is related to https://github.com/dotnet/corefx/issues/3439 ?

stephentoub:
On Windows we pull the runtime from a NuGet package. On Linux we pull the latest successful build from the coreclr repo.

stephentoub:
LGTM

PR #3637, opened 2015-10-04
Improve Process.GetProcesses* performance on Linux

### LINKED ISSUE #3505: Process.GetProcessesByName significantly slower on Linux than Windows for nonexistent process
The performance test being run:
```
[Benchmark]
[InlineData(1)]
[InlineData(2)]
[InlineData(3)]
public void GetProcessesByName(int innerIterations)
{
    foreach (var iteration in Benchmark.Iterations)
        {
            for (int i = 0; i < innerIterations; i++)
            {
                Process.GetProcessesByName("1"); Process.GetProcessesByName("1"); Process.GetProcessesByName("1");
                Process.GetProcessesByName("1"); Process.GetProcessesByName("1"); Process.GetProcessesByName("1");
                Process.GetProcessesByName("1"); Process.GetProcessesByName("1"); Process.GetProcessesByName("1");
            }
        }
}
```

Linux perf results (38.455 total seconds):
```
<assemblies>
  <assembly name="System.Diagnostics.Process.Tests.dll" environment="64-bit .NET (unknown version) [collection-per-assembly, parallel (1 threads)]" test-framework="xUnit.net 2.1.0.3168" run-date="2015-09-29" run-time="03:16:25" total="3" passed="3" failed="0" skipped="0" time="38.455" errors="0">
    <errors />
    <collection total="3" passed="3" failed="0" skipped="0" name="Test collection for System.Diagnostics.Process.Tests.dll" time="35.243">
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 1)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="6.5254048" result="Pass" />
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 2)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="11.3182348" result="Pass" />
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 3)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="17.3998431" result="Pass" />
    </collection>
  </assembly>
</assemblies>
```

Windows perf results (3.355 total seconds):
```
<assemblies>
  <assembly name="System.Diagnostics.Process.Tests.dll" environment="64-bit .NET (unknown version) [collection-per-assembly, parallel (8 threads)]" test-framework="xUnit.net 2.1.0.3168" run-date="2015-09-28" run-time="12:22:29" total="3" passed="3" failed="0" skipped="0" time="3.355" errors="0">
    <errors />
    <collection total="3" passed="3" failed="0" skipped="0" name="Test collection for System.Diagnostics.Process.Tests.dll" time="3.183">
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 1)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="1.0872349" result="Pass" />
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 2)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="1.040529" result="Pass" />
      <test name="System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 3)" type="System.Diagnostics.Tests.Perf_Process" method="GetProcessesByName" time="1.0556937" result="Pass" />
    </collection>
  </assembly>
</assemblies>
```

Further increasing the number of calls to GetProcessesByName makes comparatively little difference on Windows. For example, with an ```InnerIterations``` of 5000 (45000 total function calls), the elapsed time on Windows is only 106 seconds. In that same time Linux can only complete ~20 InnerIterations (180 total function calls). This suggests the Linux implementation takes roughly 250 times as long as the Windows implementation.

stephentoub:
On Linux we parse procfs to get information about the various processes on the system.  With our current implementation, this is significantly slower for an operation like Process.GetProcessesByName than that same operation on Windows.

Some of this is due to the core source of the information and needing to enumerate directories and open/read/close procfs files to get at the information for the processes; not much we can do about that.  However, a large chunk of it is due simply to the implementation, which we can fix.  This PR fixes the implementation to significantly lessen the gap.  For a test that does Process.GetProcessesByName and that will return just a couple of processes, this improves throughput by ~6x and reduces gen0 GCs by ~24x.  It's still ~2x slower on my Linux VM than it is on my Windows host, but my Linux VM also has about 50% more active processes than does my Windows host.

Three primary sets of changes divided into three commits:
- We previously enumerated all of the threads for a process prior to filtering by the specified name.  Now we delay doing that enumeration and associated processing until we know the process is going to match the filter.
- We use a StringParser type to parse the files in procfs.  We currently extract primitive types by first getting a substring and then parsing that substring; with this change, we do the parsing without first extracting the string.
- We currently read each file from procfs using File.ReadAllText.  Since we're doing this many times as part of the same operaiton, I added a simple reusable stream reader that lets us cache and reuse most of the objects involved in ReadAllText.

Fixes #3505
cc: @ianhays, @pallavit, @mellinoe, @ellismg

### DIFF
--- a/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
+++ b/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
@@ -3,10 +3,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
+using System.Text;
 
 internal static partial class Interop
 {
@@ -180,26 +180,29 @@ internal static partial class Interop
             return GetTaskDirectoryPathForProcess(pid) + tid.ToString(CultureInfo.InvariantCulture) + StatFileName;
         }
 
-        internal static bool TryReadStatFile(int pid, out ParsedStat result)
+        internal static bool TryReadStatFile(int pid, out ParsedStat result, ReusableTextReader reusableReader)
         {
-            bool b = TryParseStatFile(GetStatFilePathForProcess(pid), out result);
+            bool b = TryParseStatFile(GetStatFilePathForProcess(pid), out result, reusableReader);
             Debug.Assert(!b || result.pid == pid, "Expected process ID from stat file to match supplied pid");
             return b;
         }
 
-        internal static bool TryReadStatFile(int pid, int tid, out ParsedStat result)
+        internal static bool TryReadStatFile(int pid, int tid, out ParsedStat result, ReusableTextReader reusableReader)
         {
-            bool b = TryParseStatFile(GetStatFilePathForThread(pid, tid), out result);
+            bool b = TryParseStatFile(GetStatFilePathForThread(pid, tid), out result, reusableReader);
             Debug.Assert(!b || result.pid == tid, "Expected thread ID from stat file to match supplied tid");
             return b;
         }
 
-        private static bool TryParseStatFile(string statFilePath, out ParsedStat result)
+        private static bool TryParseStatFile(string statFilePath, out ParsedStat result, ReusableTextReader reusableReader)
         {
             string statFileContents;
             try
             {
-                statFileContents = File.ReadAllText(statFilePath);
+                using (var source = new FileStream(statFilePath, FileMode.Open, FileAccess.Read, FileShare.Read, bufferSize: 1, useAsync: false))
+                {
+                    statFileContents = reusableReader.ReadAllText(source);
+                }
             }
             catch (IOException)
             {
--- a/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
+++ b/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
@@ -182,12 +182,16 @@ internal static partial class Interop
 
         internal static bool TryReadStatFile(int pid, out ParsedStat result)
         {
-            return TryParseStatFile(GetStatFilePathForProcess(pid), out result);
+            bool b = TryParseStatFile(GetStatFilePathForProcess(pid), out result);
+            Debug.Assert(!b || result.pid == pid, "Expected process ID from stat file to match supplied pid");
+            return b;
         }
 
         internal static bool TryReadStatFile(int pid, int tid, out ParsedStat result)
         {
-            return TryParseStatFile(GetStatFilePathForThread(pid, tid), out result);
+            bool b = TryParseStatFile(GetStatFilePathForThread(pid, tid), out result);
+            Debug.Assert(!b || result.pid == tid, "Expected thread ID from stat file to match supplied tid");
+            return b;
         }
 
         private static bool TryParseStatFile(string statFilePath, out ParsedStat result)
--- a/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
+++ b/src/Common/src/Interop/Linux/procfs/Interop.ProcFsStat.cs
@@ -99,8 +99,7 @@ internal static partial class Interop
             {
                 return ParseMapsModulesCore(File.ReadLines(GetMapsFilePathForProcess(pid)));
             }
-            catch (FileNotFoundException) { }
-            catch (DirectoryNotFoundException) { }
+            catch (IOException) { }
             catch (UnauthorizedAccessException) { }
 
             return Array.Empty<ParsedMapsModule>();
@@ -181,19 +180,30 @@ internal static partial class Interop
             return GetTaskDirectoryPathForProcess(pid) + tid.ToString(CultureInfo.InvariantCulture) + StatFileName;
         }
 
-        internal static ParsedStat ReadStatFile(int pid)
+        internal static bool TryReadStatFile(int pid, out ParsedStat result)
         {
-            return ParseStatFile(GetStatFilePathForProcess(pid));
+            return TryParseStatFile(GetStatFilePathForProcess(pid), out result);
         }
 
-        internal static ParsedStat ReadStatFile(int pid, int tid)
+        internal static bool TryReadStatFile(int pid, int tid, out ParsedStat result)
         {
-            return ParseStatFile(GetStatFilePathForThread(pid, tid));
+            return TryParseStatFile(GetStatFilePathForThread(pid, tid), out result);
         }
 
-        private static ParsedStat ParseStatFile(string statFilePath)
+        private static bool TryParseStatFile(string statFilePath, out ParsedStat result)
         {
-            string statFileContents = File.ReadAllText(statFilePath);
+            string statFileContents;
+            try
+            {
+                statFileContents = File.ReadAllText(statFilePath);
+            }
+            catch (IOException)
+            {
+                // Between the time that we get an ID and the time that we try to read the associated stat
+                // file(s), the process could be gone.
+                result = default(ParsedStat);
+                return false;
+            }
 
             var parser = new StringParser(statFileContents, ' ');
             var results = default(ParsedStat);
@@ -256,7 +266,8 @@ internal static partial class Interop
             parser.MoveNextOrFail(); // guest_time
             parser.MoveNextOrFail(); // cguest_time
 
-            return results;
+            result = results;
+            return true;
         }
     }
 }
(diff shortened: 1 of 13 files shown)


stephentoub:
@mellinoe, I think you're using procfs for NetworkInformation on Linux? Some of these changes could help that as well.

mellinoe:
Yeah, I can certainly use the StringParser improvements, and I can refactor some things to use the ReusableTextReader. LGTM.

ianhays:
My tests show that this PR provided roughly a 3x perf increase to Linux timing.

```
Test Name	Date	Windows	Linux	Linux / Windows
System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 1)	10/6/2015	13.61966048	398.7071	29.2743788
System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 1)	10/5/2015	19.00326243	1725.0231	90.77510275

System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 2)	10/6/2015	24.02649406	874.0489999	36.37854935
System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 2)	10/5/2015	37.28179277	3230.6164	86.65399809

System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 3)	10/6/2015	41.42263029	1317.2985	31.80142089
System.Diagnostics.Tests.Perf_Process.GetProcessesByName(innerIterations: 3)	10/5/2015	55.19539114	4967.2278	89.99352478

```

Thanks, Steve!

stephentoub:
Thanks, @ianhays.  It might be interesting for this test to try to normalize the perf results based on the number of processes on the machine, as there's a per process cost to these APIs.  I'm also still interested in understanding why your machines see such a significant difference between Windows and Linux... I see a difference, but closer to 2x rather than 30x.

ianhays:
I'm guessing that it is due to interference from the other tests. When I run the GetProcessesByName test alone I get more reasonable results (closer to yours). I need to go back over my tests and determine if I'm missing some cleanup that may be leaving zombie process info that GetProcessesByName is then enumerating over. My understanding of how we work with processes on Linux is fairly shallow, so it's quite possible I missed something.

ianhays:
When GetProcessByName is being run on Linux with other tests, there are 3000 processes to iterate over. When it runs on Linux by itself, there are 300 processes to iterate over.

Windows by itself has 120 processes to iterate over. Windows with other tests has 120 processes to iterate over.

So definitely a cleanup issue on Linux then.

On Linux, I'm getting continuous printout of "Managed assembly not found: No such file or directory" and I assume that is related. Once I figure that out and cleanup the extraneous processes, I'm going to standardize the test by creating dummy processes up to a constant number before running GetProcesses. This way, Linux and Windows will have the same number of processes to iterate through for the test.

stephentoub:
> So definitely a cleanup issue on Linux then.

Yeah, I just took a quick look.  It looks like in https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Process/tests/Performance/Perf.Process.cs we're not waiting on the processes after starting them; as these are child processes, they'll end up sticking around as zombies until either the parent goes away or the parent waits on them to get their completion information.

If you add calls to WaitForExit for each process that you Start before exiting the test, does the issue go away?

ianhays:
That was it all right. Still having the hundreds of printouts for "Managed assembly not found: No such file or directory", but the results are now consistent regardless of whether theyre being run alone or with other tests.

With your PR and the tests change along with standardizing the number of processes on the machine, results are now within expected deviation and near 1:1.

Linux:
```
<performance runid="System.Diagnostics.Process.Tests.dll-Linux" etl="/home/ianha/alliswell/results/System.Diagnostics.Process.Tests.dll-Linux.csv">
  <metrics>
	<Duration displayName="Duration" unit="msec" />
  </metrics>
  <iterations>
	<iteration index="0" Duration="37.160000085830688" />
	<iteration index="1" Duration="41.426999926567078" />
	<iteration index="2" Duration="31.929900050163269" />
	<iteration index="3" Duration="27.401799917221069" />
	<iteration index="4" Duration="27.794100046157837" />
	<iteration index="5" Duration="28.112800002098083" />
	<iteration index="6" Duration="28.28849995136261" />
	<iteration index="7" Duration="28.073700070381165" />
	<iteration index="8" Duration="27.770399928092957" />
	<iteration index="9" Duration="27.35699999332428" />
	<iteration index="10" Duration="27.958299994468689" />
	<iteration index="11" Duration="53.862499952316284" />
	<iteration index="12" Duration="44.656000018119812" />
	<iteration index="13" Duration="28.961400032043457" />
	<iteration index="14" Duration="28.649800062179565" />
	<iteration index="15" Duration="27.925600051879883" />
	<iteration index="16" Duration="27.952899932861328" />
	<iteration index="17" Duration="103.09859991073608" />
	<iteration index="18" Duration="29.175100088119507" />
	<iteration index="19" Duration="27.065999984741211" />
	<iteration index="20" Duration="28.677299976348877" />
	<iteration index="21" Duration="36.407299995422363" />
	<iteration index="22" Duration="30.009999990463257" />
	<iteration index="23" Duration="27.950099945068359" />
	<iteration index="24" Duration="29.955699920654297" />
	<iteration index="25" Duration="41.308300018310547" />
	<iteration index="26" Duration="29.133000016212463" />
	<iteration index="27" Duration="39.77209997177124" />
	<iteration index="28" Duration="28.331000089645386" />
	<iteration index="29" Duration="25.055299997329712" />
	<iteration index="30" Duration="27.51419997215271" />
  </iterations>
</performance>
```

Windows core:
```
        <performance runid="System.Diagnostics.Process.Tests.dll-WindowsCore" etl="D:\git\corefx\bin\tests\Windows_NT.AnyCPU.Release\System.Diagnostics.Process.Tests\dnxcore50\System.Diagnostics.Process.Tests.dll-WindowsCore.etl">
          <metrics>
            <Duration displayName="Duration" unit="msec" />
          </metrics>
          <iterations>
            <iteration index="0" Duration="59.282092863266826" />
            <iteration index="1" Duration="50.452444632232528" />
            <iteration index="2" Duration="25.796428853477209" />
            <iteration index="3" Duration="24.46984334430681" />
            <iteration index="4" Duration="26.45444719759189" />
            <iteration index="5" Duration="25.923014704771049" />
            <iteration index="6" Duration="25.489087529839708" />
            <iteration index="7" Duration="25.535559362634558" />
            <iteration index="8" Duration="29.739692162596384" />
            <iteration index="9" Duration="24.91831078594123" />
            <iteration index="10" Duration="24.082958208707623" />
            <iteration index="11" Duration="26.0735492306344" />
            <iteration index="12" Duration="24.978467575633204" />
            <iteration index="13" Duration="24.523442758818874" />
            <iteration index="14" Duration="24.604126983855622" />
            <iteration index="15" Duration="24.793150450990652" />
            <iteration index="16" Duration="24.332423568803279" />
            <iteration index="17" Duration="26.008260582107141" />
            <iteration index="18" Duration="28.423370371098827" />
            <iteration index="19" Duration="26.15138242298417" />
            <iteration index="20" Duration="24.832494702068743" />
            <iteration index="21" Duration="23.943542710322617" />
            <iteration index="22" Duration="24.585595271391412" />
            <iteration index="23" Duration="25.150669949915937" />
            <iteration index="24" Duration="25.258153882208262" />
            <iteration index="25" Duration="25.985452320612239" />
            <iteration index="26" Duration="24.596144092332452" />
            <iteration index="27" Duration="24.832209598799636" />
            <iteration index="28" Duration="25.335701971289382" />
            <iteration index="29" Duration="25.25530284952174" />
            <iteration index="30" Duration="25.041475398011244" />
            <iteration index="31" Duration="24.511468421534573" />
            <iteration index="32" Duration="25.234775414176511" />
            <iteration index="33" Duration="24.906621551925582" />
            <iteration index="34" Duration="25.3838844236966" />
            <iteration index="35" Duration="25.330285009184536" />
            <iteration index="36" Duration="24.30476855174129" />
            <iteration index="37" Duration="24.335559704758452" />
            <iteration index="38" Duration="24.17533166776002" />
            <iteration index="39" Duration="25.141546645318158" />
          </iterations>
        </performance>
```

stephentoub:
Excellent! But before we declare success, we should understand that not found failure. It's very likely coming from https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/coreclr/hosts/unixcorerun/corerun.cpp#L118, which means the process being started isn't running much and is exiting very quickly. That shouldn't influence these results, but we should verify and fix the problem regardless.

PR #3638, opened 2015-10-05
Fixed Renegotiate & sslV3 issue in System.Net.Security xplat. Made op…

rajansingh10:
1. Fixed xplat Renegotiate in both code path, as server & as client.
2. Fixed SSLv3 Decrypt. Now we understand Error_Want_Read & handle accordingly.
3. Various SSL method IntPtr are now being referenced once statically.

### DIFF
--- a/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
@@ -113,16 +113,10 @@ internal static partial class Interop
             }
 
             int retVal = libssl.SSL_do_handshake(context.sslPtr);
-            if ((retVal == 1) && !isServer)
-            {            
-                return true;
-            }
-
-            int error;
 
             if (retVal != 1)
             {
-                error = GetSslError(context.sslPtr, retVal);
+                libssl.SslErrorCode error = GetSslError(context.sslPtr, retVal);
 
                 if ((retVal != -1) || (error != libssl.SslErrorCode.SSL_ERROR_WANT_READ))
                 {
@@ -138,11 +132,11 @@ internal static partial class Interop
                 sendCount = BioRead(context.writeBioPtr, sendPtr, sendCount);
                 if (sendCount <= 0)
                 {
-                    error = sendCount;
+                    int errorCode = sendCount;
                     Marshal.FreeHGlobal(sendPtr);
                     sendPtr = IntPtr.Zero;
-                    sendCount = 0;                  
-                    throw CreateSslException(context.sslPtr, "Read Bio failed: ", error);                   
+                    sendCount = 0;
+                    throw CreateSslException(context.sslPtr, "Read Bio failed: ", errorCode);                   
                 }
             }
         
@@ -150,34 +144,40 @@ internal static partial class Interop
 
         }
 
-        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int count, int bufferCapacity)
+        internal static int Encrypt(IntPtr handlePtr, IntPtr buffer, int offset, int count, int bufferCapacity, out libssl.SslErrorCode errorCode)
         {
+            errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
+
             SslContext context = Marshal.PtrToStructure<SslContext>(handlePtr);
 
             var retVal = libssl.SSL_write(context.sslPtr, new IntPtr(buffer.ToInt64() + offset), count);
 
             if (retVal != count)
             {
-                int error = GetSslError(context.sslPtr, retVal);
-                if (libssl.SslErrorCode.SSL_ERROR_ZERO_RETURN == error)
+                errorCode = GetSslError(context.sslPtr, retVal);
+                retVal = 0;
+
+                switch (errorCode)
                 {
-                    return 0; // indicate end-of-file
+                    // indicate end-of-file
+                    case libssl.SslErrorCode.SSL_ERROR_ZERO_RETURN:
+                    case libssl.SslErrorCode.SSL_ERROR_WANT_READ:
+                        break;
+
+                    default:
+                        throw CreateSslException("OpenSsl::Encrypt failed");
                 }
-                throw CreateSslException("OpenSsl::Encrypt failed");
             }
-
-            int capacityNeeded = libssl.BIO_ctrl_pending(context.writeBioPtr);      
-
-            if (retVal == count)
+            else
             {
+                int capacityNeeded = libssl.BIO_ctrl_pending(context.writeBioPtr);      
+
                 if (capacityNeeded > bufferCapacity)
                 {
                     throw CreateSslException("OpenSsl::Encrypt capacity needed is more than buffer capacity. capacityNeeded = " + capacityNeeded + "," + "bufferCapacity = " + bufferCapacity);
-                }
+                }             
 
-                IntPtr outBufferPtr = buffer;
-
-                retVal = BioRead(context.writeBioPtr, outBufferPtr, capacityNeeded);
+                retVal = BioRead(context.writeBioPtr, buffer, capacityNeeded);
 
                 if (retVal < 0)
                 {
@@ -188,8 +188,10 @@ internal static partial class Interop
             return retVal;
         }
 
-        internal static int Decrypt(IntPtr sslContextPtr, IntPtr outBufferPtr, int count)
+        internal static int Decrypt(IntPtr sslContextPtr, IntPtr outBufferPtr, int count, out libssl.SslErrorCode errorCode)
         {
+            errorCode = libssl.SslErrorCode.SSL_ERROR_NONE;
+
             SslContext context = Marshal.PtrToStructure<SslContext>(sslContextPtr);
 
             int retVal = BioWrite(context.readBioPtr, outBufferPtr, count);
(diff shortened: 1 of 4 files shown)


rajansingh10:
cc: @stephentoub, @bartonjs, @davidsh, @CIPop, @SidharthNabar, @pgavlin , @vijaykota,

stephentoub:
Thanks, @rajansingh10.  I don't see any tests being added or changed as part of this.  Are there no tests that were failing before the change and are now passing?  If not, shouldn't we add some?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -206,12 +208,23 @@ internal static int Decrypt(IntPtr sslContextPtr, IntPtr outBufferPtr, int count
...
+                switch (errorCode)
                 {
-                    return 0; // indicate end-of-file
+                    // indicate end-of-file
+                    case libssl.SslErrorCode.SSL_ERROR_ZERO_RETURN:                      
+                        break;
+
+                    case libssl.SslErrorCode.SSL_ERROR_WANT_READ:
+                        // update error code to renegotiate if renegotiate is pending, otherwise make it SSL_ERROR_WANT_READ
+                        errorCode = libssl.SSL_renegotiate_pending(context.sslPtr) == 1 ? libssl.SslErrorCode.SSL_ERROR_RENEGOTIATE : libssl.SslErrorCode.SSL_ERROR_WANT_READ;
```
stephentoub: Nit: please split this across multiple lines... it's a very long line

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -295,23 +308,23 @@ private static IntPtr GetSslMethod(bool isServer, long options)
...
             }
             else if (noSsl2 && noTls10 && noTls11 && noTls12)
             {
-                method = libssl.SSLv3_method();
+                method = libssl.SslMethods.SSLv3_method;
             }
             else
             {
-                method = libssl.SSLv23_method();
+                method = libssl.SslMethods.SSLv23_method;
```
stephentoub: You could restructure this a bit to avoid checking the same conditions repeatedly, e.g.
```C#
IntPtr method = libssl.SslMethods.SSLv23_method; // default
if (noSsl2)
{
    if (noSsl3)
    {
        if (noTls11 && noTls12)
        {
            method = libssl.SslMethods.TLSv1_method;
        }
        else if (noTls10 && noTls12)
        {
            method = libssl.SslMethods.TLSv1_1_method;
        }
        else if (noTls10 && noTls11)
        {
            method = libssl.SslMethods.TLSv1_2_method;
        }
    }
    else if (noTls10 && noTls11 && noTls12)
    {
        method = libssl.SslMethods.SSLv3_method;
    }
}
```
Just something to consider.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -295,23 +308,23 @@ private static IntPtr GetSslMethod(bool isServer, long options)
 
             if (noSsl2 && noSsl3 && noTls11 && noTls12)
             {
-                method = libssl.TLSv1_method();
+                method = libssl.SslMethods.TLSv1_method;
```
stephentoub: ```protocolMask``` above should be declared ```const```.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
stephentoub: That said, rather than declaring these as ```no``` something, would it make sense to invert them and have them declare what's available rather than what's not available, e.g.
```C#
bool ssl2 = (options & libssl.Options.SSL_OP_NO_SSLv2) == 0;
```
?  I personally have an easier time with conditionals like this when I don't have to repeatedly think through double-negatives.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -400,20 +413,19 @@ private static SslException CreateSslException(string message)
...
-                return new SslException(innerEx.Message, innerEx);
-            }
-            else
-            {
-                return new SslException(message + ": " + error, error);
+                case libssl.SslErrorCode.SSL_ERROR_SYSCALL:
+                    return new SslException(message, error);
+
+                case libssl.SslErrorCode.SSL_ERROR_SSL:
+                        Exception innerEx = Interop.libcrypto.CreateOpenSslCryptographicException();
```
stephentoub: Nit: indentation here is off

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
```diff
@@ -61,12 +62,43 @@ internal static class SslSizes
...
-            internal const int SSL_ERROR_WANT_READ = 2;
-            internal const int SSL_ERROR_SYSCALL = 5;
-            internal const int SSL_ERROR_ZERO_RETURN = 6;
+            internal static readonly IntPtr TLSv1_method;
+            internal static readonly IntPtr TLSv1_1_method;
+            internal static readonly IntPtr TLSv1_2_method;
+            internal static readonly IntPtr SSLv3_method;
+            internal static readonly IntPtr SSLv23_method;
+
+            private static IntPtr checkMethodPtr(IntPtr methodPtr)
```
stephentoub: Without line numbers in stack traces right now on Unix, we won't know which of the calls to checkMethodPtr is causing the assert.  Since the whole point of this method is the Debug.Fail, I'd suggest adding a [CallerLineNumber] argument to the method, and using that line number in the call to Debug.Fail, so that we know exactly which one fails if it does.

stephentoub: Actually, instead, I'd suggest doing this:
```C#
internal static readonly IntPtr TLSv1_method = libssl.TLSv1_method();
internal static readonly IntPtr TLSv1_1_method = libssl.TLSv1_1_method();
internal static readonly IntPtr TLSv1_2_method = libssl.TLSv1_2_method();
internal static readonly IntPtr SSLv3_method = libssl.SSLv3_method();
internal static readonly IntPtr SSLv23_method = libssl.SSLv23_method();

#if DEBUG
static SslMethods()
{
    Debug.Assert(TLSv1_method != IntPtr.Zero, "TLSv1_method is null");
    Debug.Assert(TLSv1_1_method != IntPtr.Zero, "TLSv1_1_method is null");
    Debug.Assert(TLSv1_2_method != IntPtr.Zero, "TLSv1_2_method is null");
    Debug.Assert(SSLv3_method != IntPtr.Zero, "SSLv3_method is null");
    Debug.Assert(SSLv23_method != IntPtr.Zero, "SSLv23 method is null");
}
#endif
```
That way, in release builds, SslMethods will be a beforefieldinit type, we won't end up paying anything for checkMethodPtr in release builds, and the assert will tell us in debug builds exactly which method is null.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.libssl_types.cs
rajansingh10: yep, makes sense. Done.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: done, made it simpler.

rajansingh10:
@stephentoub Thanks for the review. Regarding the tests, I will check what is the current status.
I will add more coverage, in case this is yet not covered. Currently I do not have much insight into existing xplat test. IMO, if this PR looks fine, merge can go ahead & I will send the test update tomorrow in an new PR?

rajansingh10:
Incorporated comments.

rajansingh10:
@stephentoub yeah it will cause an issue, I just recall the reason for putting it that way. It have to fall back to v23 method. I am making it as suggested with modified variables but keeping order & similarity as it is.

stephentoub:
> I will send the test update tomorrow in an new PR?

Ok, sounds fine.

stephentoub:
> yeah it will cause an issue

Ok, let's get https://github.com/rajansingh10/corefx/commit/2e807bbfd3da74a685a2dead77f75fdf1ee3c60e#commitcomment-13598666 fixed before merging this.

rajansingh10:
@stephentoub fixed & uploaded. I will squash the commits, once signed off. thanks.

stephentoub:
LGTM

bartonjs:
LGTM2

vijaykota:
> Are there no tests 

cc: @shrutigarg 

@stephentoub, the checked in test cases are failing due to <i>other</i> missing features like algorithm strength not being set, encryption policy not being honored etc. @shrutigarg is working thru these failures. She will also try to increase the existing test coverage

rajansingh10:
@dotnet-bot test this please

rajansingh10:
@dotnet-bot test this please

stephentoub:
The tests are failing due to https://github.com/dotnet/corefx/issues/3658... retesting won't help right now.

stephentoub:
@dotnet-bot test this please

rajansingh10:
@stephentoub Thanks.

PR #3639, opened 2015-10-05
Fix and enable a few GC tests

### LINKED ISSUE #2938: GC tests GetTotalMemoryTest_ForceCollection,  GetGenerationTest test failed in CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release/1057/console
```
GCTests.GetTotalMemoryTest_ForceCollection [FAIL]
        Assert.True() Failure
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release\src\System.Runtime\tests\System\GC.cs(295,0): at GCTests.GetTotalMemoryTest_ForceCollection()
```
We should probably just delete these tests.

stephentoub:
Several GC tests were previously disabled due to poor assumptions about several GC behaviors, e.g. assuming that objects always move to a higher generation up to the max when a full collection occurs, assuming that the GC won't kick in due to other things happening concurrently in the process, etc.

This commit deletes one test and revamps another two that were previously disabled.

Fixes #2938
cc: @Maoni0, @Sridhar-MS

### DIFF
--- a/src/System.Runtime/tests/System/GC.cs
+++ b/src/System.Runtime/tests/System/GC.cs
@@ -270,43 +270,42 @@ public class GCTests
         }
     }
 
-    [ActiveIssue(2938)]
-    [Fact]
-    public static void GetTotalMemoryTest_NoForceCollection()
-    {
-        GC.Collect();
-
-        byte[] bytes = new byte[50000];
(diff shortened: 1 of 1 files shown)


stephentoub:
@dotnet-bot test this please (#3641)

PR #3640, opened 2015-10-05
Implement AesCng and TripleDESCng

AtsushiKan:
Implement AesCng and TripleDESCng.

This code is derived from the internal CNG-based Aes implementation in SSC.Algorithms
but the factoring has been redone so that the padding logic, cipher algorithm and the
underlying provider are now properly separated.

UniversalCryptoEncryptor and UniversalCryptoDecryptor implement the public-facing
ICryptoTransform interface and do the padding/depadding.

They are parameterized by a BasicSymmetricCipher which implements the unpadded transform
and encapsulates the cipher algorithm and underlying crypto technology (CNG or OpenSSL or
whatever.)

BasicSymmetricCipherCng is the CNG-based implementation of that, and
is parameterized by the algorithm (AES or 3DES.)

(The existing Aes implementations in Algorithms could be rewritten in terms of these
but I'm not doing that now. We don't even know that we'll keep those inside
SSC.Algorithms, in which case, redoing them would be throwaway work. I've put the agnostic
code in the Common folder just in case.)

Beefed up the shared Aes tests with a few more code-coverage tests, but what's there
seems pretty comprehensive already.

TripleDES tests exercise the basic cipher. No need to do more code-coverage with these
as AesCng and TripleDES are the virtually the same class on the managed side.

cc @bartonjs

### DIFF
--- a/src/Common/src/Internal/Cryptography/UniversalCryptoDecryptor.cs
+++ b/src/Common/src/Internal/Cryptography/UniversalCryptoDecryptor.cs
@@ -1,0 +1,199 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Security.Cryptography;
+
+namespace Internal.Cryptography
+{
+    //
+    // A cross-platform ICryptoTransform implementation for decryption. 
+    //
+    //  - Implements the various padding algorithms (as we support padding algorithms that the underlying native apis don't.)
+    //
+    //  - Parameterized by a BasicSymmetricCipher which encapsulates the algorithm, key, IV, chaining mode, direction of encryption
+    //    and the underlying native apis implementing the encryption.
+    //
+    internal sealed class UniversalCryptoDecryptor : UniversalCryptoTransform
+    {
+        public UniversalCryptoDecryptor(PaddingMode paddingMode, BasicSymmetricCipher basicSymmetricCipher)
+            : base(paddingMode, basicSymmetricCipher)
+        {
+        }
+
+        protected sealed override int UncheckedTransformBlock(byte[] inputBuffer, int inputOffset, int inputCount, byte[] outputBuffer, int outputOffset)
+        {
+            //
+            // If we're decrypting, it's possible to be called with the last blocks of the data, and then
+            // have TransformFinalBlock called with an empty array. Since we don't know if this is the case,
+            // we won't decrypt the last block of the input until either TransformBlock or
+            // TransformFinalBlock is next called.
+            //
+            // We don't need to do this for PaddingMode.None because there is no padding to strip, and
+            // we also don't do this for PaddingMode.Zeros since there is no way for us to tell if the
+            // zeros at the end of a block are part of the plaintext or the padding.
+            //
+            int decryptedBytes = 0;
+            if (DepaddingRequired)
+            {
+                // If we have data saved from a previous call, decrypt that into the output first
+                if (_heldoverCipher != null)
+                {
+                    int depadDecryptLength = BasicSymmetricCipher.Transform(_heldoverCipher, 0, _heldoverCipher.Length, outputBuffer, outputOffset);
+                    outputOffset += depadDecryptLength;
+                    decryptedBytes += depadDecryptLength;
+                }
+                else
+                {
+                    _heldoverCipher = new byte[InputBlockSize];
+                }
+
+                // Postpone the last block to the next round.
+                Debug.Assert(inputCount >= _heldoverCipher.Length, "inputCount >= _heldoverCipher.Length");
+                int startOfLastBlock = inputOffset + inputCount - _heldoverCipher.Length;
+                Buffer.BlockCopy(inputBuffer, startOfLastBlock, _heldoverCipher, 0, _heldoverCipher.Length);
+                inputCount -= _heldoverCipher.Length;
+                Debug.Assert(inputCount % InputBlockSize == 0, "Did not remove whole blocks for depadding");
+            }
+
+            if (inputCount > 0)
+            {
+                decryptedBytes += BasicSymmetricCipher.Transform(inputBuffer, inputOffset, inputCount, outputBuffer, outputOffset);
+            }
+
+            return decryptedBytes;
+        }
+
+        protected sealed override byte[] UncheckedTransformFinalBlock(byte[] inputBuffer, int inputOffset, int inputCount)
+        {
+            // We can't complete decryption on a partial block
+            if (inputCount % InputBlockSize != 0)
+                throw new CryptographicException(SR.Cryptography_PartialBlock);
+
+            //
+            // If we have postponed cipher bits from the prior round, copy that into the decryption buffer followed by the input data.
+            // Otherwise the decryption buffer is just the input data.
+            //
+
+            byte[] ciphertext = null;
+
+            if (_heldoverCipher == null)
+            {
+                ciphertext = new byte[inputCount];
+                Buffer.BlockCopy(inputBuffer, inputOffset, ciphertext, 0, inputCount);
+            }
+            else
+            {
+                ciphertext = new byte[_heldoverCipher.Length + inputCount];
+                Buffer.BlockCopy(_heldoverCipher, 0, ciphertext, 0, _heldoverCipher.Length);
+                Buffer.BlockCopy(inputBuffer, inputOffset, ciphertext, _heldoverCipher.Length, inputCount);
+            }
+
+            // Decrypt the data, then strip the padding to get the final decrypted data. Note that even if the cipherText length is 0, we must
+            // invoke TransformFinal() so that the cipher object knows to reset for the next cipher operation.
+            byte[] decryptedBytes = BasicSymmetricCipher.TransformFinal(ciphertext, 0, ciphertext.Length);
+            byte[] outputData;
+            if (ciphertext.Length > 0)
+            {
+                outputData = DepadBlock(decryptedBytes, 0, decryptedBytes.Length);
+            }
+            else
+            {
+                outputData = Array.Empty<byte>();
+            }
+
+            Reset();
+            return outputData;
+        }
+
+        protected sealed override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                byte[] heldoverCipher = _heldoverCipher;
+                _heldoverCipher = null;
+                if (heldoverCipher != null)
+                {
+                    Array.Clear(heldoverCipher, 0, heldoverCipher.Length);
+                }
+            }
+
+            base.Dispose(disposing);
+        }
+
+        private void Reset()
+        {
+            if (_heldoverCipher != null)
+            {
+                Array.Clear(_heldoverCipher, 0, _heldoverCipher.Length);
+                _heldoverCipher = null;
+            }
+        }
+
+        private bool DepaddingRequired
+        {
+            get
+            {
+                // Aes automatically strips padding after decryption when PKCS7 padding is in effect.
+                // It does not do so when PaddingMode.Zeroes in effect as that padding mode is not sufficiently
+                // self-describing to do the operation safely.
+                return PaddingMode == PaddingMode.PKCS7;
+            }
+        }
+
+        /// <summary>
+        ///     Remove the padding from the last blocks being decrypted
+        /// </summary>
+        private byte[] DepadBlock(byte[] block, int offset, int count)
+        {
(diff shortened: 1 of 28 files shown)


src/Common/src/Internal/Cryptography/UniversalCryptoDecryptor.cs
```diff
@@ -0,0 +1,200 @@
...
+            return decryptedBytes;
+        }
+
+        protected sealed override byte[] UncheckedTransformFinalBlock(byte[] inputBuffer, int inputOffset, int inputCount)
+        {
+            // We can't complete decryption on a partial block
+            if (inputCount % InputBlockSize != 0)
+                throw new CryptographicException(SR.Cryptography_PartialBlock);
+
+            byte[] outputData;
```
stephentoub: Nit: this should be declared down closer to where it's actually set.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngSymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,194 @@
...
+            outer.BaseKeySize = keySize;
+            _keyName = null; // Setting _keyName to null signifies that this object is now based on a plaintext key, not a stored CNG key.
+        }
+
+        public void GenerateKey()
+        {
+            byte[] key = new byte[_outer.BaseKeySize.BitSizeToByteSize()];
+            do
+            {
+                RandomNumberGenerator.Create().GetBytes(key);
```
stephentoub: RandomNumberGenerator is IDisposable. We should Dispose of the instance when done with it, even if the current implementations have Dispose as a nop.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngSymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            while (_outer.IsWeakKey(key));
+            SetKey(key);
+        }
+
+        public void GenerateIV()
+        {
+            byte[] iv = new byte[_outer.BlockSize.BitSizeToByteSize()];
+            RandomNumberGenerator.Create().GetBytes(iv);
```
stephentoub: Ditto.

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/SymmetricImportExportExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+                    bw.Write((char)0);
+
+                    // Write out a BCRYPT_KEY_DATA_BLOB_HEADER
+                    bw.Write((int)Interop.BCrypt.BCRYPT_KEY_DATA_BLOB_MAGIC);          // BCRYPT_KEY_DATA_BLOB_HEADER.dwMagic
+                    bw.Write((int)Interop.BCrypt.BCRYPT_KEY_DATA_BLOB_VERSION1);       // BCRYPT_KEY_DATA_BLOB_HEADER.dwVersion
+                    bw.Write((int)(key.Length));                                       // BCRYPT_KEY_DATA_BLOB_HEADER.cbKeyData
+
+                    bw.Write((byte[])key);                                             // Write out the key data.
+                }
+                ms.Flush();
```
stephentoub: MemoryStream.Flush is a nop

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/SymmetricImportExportExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Globalization;
+using System.Security.Cryptography;
+
+namespace Internal.Cryptography
+{
+    internal static class SymmetricImportExportExtensions
+    {
+        public static CngKey ToCngKey(this byte[] key, string algorithm)
+        {
+            using (MemoryStream ms = new MemoryStream())
```
stephentoub: Consider passing in the expected capacity here so that the MemoryStream doesn't need to continually grow its array.

src/Common/src/Interop/Windows/Crypt32/Interop.NCryptAlgorithms.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+internal static partial class Interop
+{
+    internal static partial class Crypt32
```
bartonjs: Wouldn't NCrypt algorithms be in NCrypt, not Crypt32?

src/Common/tests/Cryptography/AlgorithmImplementations/AES/AesCornerTests.cs
```diff
@@ -0,0 +1,328 @@
...
+        [Fact]
+        public static void ZeroLengthTransformNoPad()
+        {
+            // Use no TransformBlock calls() - do the entire transform using only TransformFinalBlock().
+            byte[] key = "1ed2f625c187b993256a8b3ccf9dcbfa5b44b4795c731012f70e4e64732efd5d".HexToByteArray();
+            byte[] iv = "47d1e060ba3c8643f9f8b65feeda4b30".HexToByteArray();
+
+            byte[] plainText = "".HexToByteArray();
+            // CBC, Padding.None
+            byte[] expectedCipher = "".HexToByteArray();
```
bartonjs: Why not just Array.Empty\<byte\>?

src/Common/src/Interop/Windows/Crypt32/Interop.NCryptAlgorithms.cs
AtsushiKan: Perhaps but that follows existing naming pattern in Common\Src\Interop.

src/Common/tests/Cryptography/AlgorithmImplementations/TripleDES/TripleDESCipherTests.cs
```diff
@@ -0,0 +1,161 @@
...
+using Xunit;
+
+namespace System.Security.Cryptography.Encryption.TripleDes.Tests
+{
+    public static class TripleDESCipherTests
+    {
+        [Fact]
+        public static void TripleDESDefaults()
+        {
+            TripleDES des = TripleDESFactory.Create();
```
bartonjs: TripleDESRoundTrip... uses a using, why doesn't this?

src/Common/src/Interop/Windows/Crypt32/Interop.NCryptAlgorithms.cs
AtsushiKan: On second thought, I think I’m confused. Wait a sec…

src/Common/tests/Cryptography/AlgorithmImplementations/AES/AesCornerTests.cs
AtsushiKan: Symmetry with the other crytotests that use the <string>.ToHexByteArray() pattern.

src/Common/tests/Cryptography/CryptoUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Globalization;
+using System.Security.Cryptography;
+
+namespace Test.Cryptography
+{
+    internal static class CryptoUtils
+    {
+        internal static byte[] Encrypt(this SymmetricAlgorithm alg, byte[] plainText, int blockSizeMultipler = 1)
+        {
+            return alg.CreateEncryptor().Transform(plainText, blockSizeMultipler);
```
bartonjs: the output of CreateEncryptor is disposable, and should be in a using.

src/Common/tests/Cryptography/CryptoUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        internal static byte[] Decrypt(this SymmetricAlgorithm alg, byte[] cipher, int blockSizeMultipler = 1)
+        {
+            return alg.CreateDecryptor().Transform(cipher, blockSizeMultipler);
```
bartonjs: the output of CreateDecryptor is disposable, and should be in a using.

src/Common/tests/Cryptography/CryptoUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+                if (count >= blockSize)
+                {
+                    byte[] buffer = new byte[blockSize];
+                    int numBytesWritten = transform.TransformBlock(input, i, count, buffer, 0);
+                    Array.Resize(ref buffer, numBytesWritten);
+                    output.AddRange(buffer);
+                }
+                else
+                {
+                    byte[] finalBlock = transform.TransformFinalBlock(input, i, count);
```
bartonjs: I guess I don't understand what blockSizeMultiplier is supposed to be doing.

Transform("hi".ToBytes(), 10) is the same as Transform("hi".ToBytes(), 1), in that both just call TransformFinalBlock.

And since blocksize isn't decremented during looping, Transform(new byte[4096], 1) is just going to call TransformBlock over and over, but never TransformFinalBlock.

AtsushiKan: “I” is the loop induction variable, not “blockSize”.

The multiplier controls the size of the chunks we feed to Transform(). Purpose is to let us test multiple block transforms, and edge cases (like never calling Transform but only TransformFinalBlock.)

src/System.Security.Cryptography.Cng/src/Internal/Cryptography/CngSymmetricAlgorithmCore.cs
```diff
@@ -0,0 +1,194 @@
...
+                {
+                    return cngKey.GetSymmetricKeyDataIfExportable(_algorithm);
+                }
+            }
+        }
+
+        public void SetKey(byte[] key)
+        {
+            _outer.BaseKey = key;
+            _keyName = null; // Setting _keyName to null signifies that this object is now based on a plaintext key, not a stored CNG key.
```
bartonjs: Opening an AesCng with a named key, then assigning the Key property switches to a software ephemeral key.  Is that what we'd expect, or should it just throw?

AtsushiKan: Fair question. I’m no fan of the way these Crypto classes let you reassign everything under the sun instead of specifying things once and for all in a constructor.

On the other hand, that is the pattern, so I didn’t want to change it for this one case.

bartonjs: Well, the existing things let you change a software key to another software key.  In this case, it feels like it should either
a) Update the key in the named key container
b) throw

But
c) switch from a saved (and maybe hardware) key to an unsaved software key

doesn't seem very good.  I'm a fan of b; personally.

AtsushiKan: I would definitely not vote for (a). Exchanging one software key to another doesn’t propagate the change backward to any other component that happened to be storing the old key. Nor would I expect it to.

(b) Throwing a “Make up your mind” exception if you attempt to set Key or KeySize after constructing the SA with a named key is fine by me
      philosophically, but it isn’t consistent with the overall flavor of SymmetricAlgorithm-based classes so I’d like to get an opinion from the api guys here.

AtsushiKan: cc @nguerrera 

Do you have an opinion on this?

bartonjs: @nguerrera Here's the summary:

```
using (Aes maybeHardwareAes = new AesCng(keyAddress, keyProvider))
{
    // What should this line do?
    maybeHardwareAes.Key = new byte[128 / 8];
}
```

nguerrera: Looked at this a bit offline with @bartonjs. We already have the same situation in the RSA/symmetric case, where you can create an RSACng from a hardware key and then change the key to a new software one using KeySize setter or Import. We should be consistent so I think the way @AtsushiKan is fine unless you guys want to make an all-up proposal for changing the behavior elsewhere too.

bartonjs:
LGTM

stephentoub:
LGTM

PR #3643, opened 2015-10-05
Remove dummy Console.cs from Common\tests

stephentoub:
Several of the test projects are doing Console.WriteLine, but aren't actually using System.Console.dll, and are instead including a dummy Console.cs file where WriteLine just calls Debug.WriteLine.  This commit deletes that dummy file and updates the consuming test projects to just use Debug.WriteLine. (We don't want our tests to be using Console.WriteLine for this kind of debug spew, anyway.)


stephentoub:
cc: @mellinoe

mellinoe:
LGTM

stephentoub:
Thanks, Eric.

Issue #71, opened 2014-11-13
Behaviour of `Quaternion.CreateFromAxisAngle` when axis is not a unit vector


CodesInChaos:
The resulting quaternion depends on the length of `axis`. This does not describe a rotation and thus does not match the documentation.

There are a few ways to handle this issue:

1. Normalize `axis`, but this incurs a performance hit
2. Add a precondition that `axis.Length()` ≈ 1.

    But enforcing that pre-condition is about as expensive as 1), so it'd probably need to be a documentation-only precondition. Unfortunately this means that consumers will rely on the current behaviour even if it's undocumented, so I don't think that this is a good solution.
3. Document the current behaviour

----

Just for convenience, the relevant source code:

    /// <summary>
    /// Creates a Quaternion from a vector and an angle to rotate about the vector.
    /// </summary>
    /// <param name="axis">The vector to rotate around.</param>
    /// <param name="angle">The angle, in radians, to rotate around the vector.</param>
    /// <returns>The created Quaternion.</returns>
    public static Quaternion CreateFromAxisAngle(Vector3 axis, float angle)
    {
        Quaternion ans;

        float halfAngle = angle * 0.5f;
        float s = (float)Math.Sin(halfAngle);
        float c = (float)Math.Cos(halfAngle);

        ans.X = axis.X * s;
        ans.Y = axis.Y * s;
        ans.Z = axis.Z * s;
        ans.W = c;

        return ans;
    }

jasonwilliams200OK:
This is fixed by https://github.com/dotnet/corefx/pull/3649.

PR #3649, opened 2015-10-05
Improve Quaternion.CreateFromAxisAngle docs [Issue #71]

mellinoe:
This adds some documentation around the fact that the "axis" parameter needs to be of unit length before calling the method. Addresses #71 .

@rpetrusha Any way we could get this same info into the MSDN docs of Quaternion?
https://msdn.microsoft.com/en-us/library/system.numerics.quaternion.createfromaxisangle(v=vs.111).aspx

### DIFF
--- a/src/System.Numerics.Vectors/src/System/Numerics/Quaternion.cs
+++ b/src/System.Numerics.Vectors/src/System/Numerics/Quaternion.cs
@@ -155,9 +155,10 @@ namespace System.Numerics
         }
 
         /// <summary>
-        /// Creates a Quaternion from a vector and an angle to rotate about the vector.
+        /// Creates a Quaternion from a normalized vector axis and an angle to rotate about the vector.
         /// </summary>
-        /// <param name="axis">The vector to rotate around.</param>
+        /// <param name="axis">The unit vector to rotate around.
+        /// This vector must be normalized before calling this function or the resulting Quaternion will be incorrect.</param>
         /// <param name="angle">The angle, in radians, to rotate around the vector.</param>
         /// <returns>The created Quaternion.</returns>
         public static Quaternion CreateFromAxisAngle(Vector3 axis, float angle)


src/System.Numerics.Vectors/src/System/Numerics/Quaternion.cs
```diff
@@ -155,9 +155,9 @@ public static Quaternion Inverse(Quaternion value)
         }
 
         /// <summary>
-        /// Creates a Quaternion from a vector and an angle to rotate about the vector.
+        /// Creates a Quaternion from a unit vector axis and an angle to rotate about the vector.
         /// </summary>
-        /// <param name="axis">The vector to rotate around.</param>
+        /// <param name="axis">The unit vector to rotate around. This must be a vector of length 1.</param>
```
stephentoub: Is this "must" verified anywhere in the implementation?

mellinoe: It is not verified anywhere, if you pass in a vector that hasn't been normalized, you will probably not get the result you expect.

Perhaps "the given vector must be normalized prior to calling this function" would be better?

stephentoub: Either of those are ok, but I think we should briefly explain what happens if you don't do that, e.g. "This should be a vector of length 1 or else the result will be non-sensical." or something better worded.

mellinoe:
I changed the comment a little bit, to note that the result will be incorrect if you don't pre-normalize the vector.

src/System.Net.Primitives/src/System/Net/IWebProxy.cs
```diff
@@ -1,18 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: Why is this file getting deleted?  Looks like a mistake.

stephentoub:
> I changed the comment a little bit

LGTM

src/System.Net.Primitives/src/System/Net/IWebProxy.cs
mellinoe: Ugh, didn't mean to commit that... this keeps getting automatically deleted in my local repository for some reason...

mellinoe:
I'll let the CI go again. I don't know why that random file keeps showing up as deleted in my local repo ><

stephentoub:
@dotnet-bot test this please

PR #3650, opened 2015-10-05
Expose Console.Is[Input\Output]Redirected APIs

pallavit:


pallavit:
@stephentoub , @weshaggard Could you please take a look at this?

src/System.Console/src/System/Console.cs
```diff
@@ -73,6 +90,41 @@ private static TextWriter CreateOutputWriter(Stream outputStream)
...
+        private static bool _isStdInRedirected;
+        private static bool _isStdOutRedirected;
+        private static bool _isStdErrRedirected;
+
+        public static bool IsInputRedirected
+        {
+            get
+            {
+                return Volatile.Read(ref _stdInRedirectQueried) ? _isStdInRedirected
+                    : EnsureInitialized(ref _isStdInRedirected, ref _stdInRedirectQueried, () => { return ConsolePal.IsInputRedirectedCore(); });
```
stephentoub: The way this is constructed, we're going to end up calling EnsureInitialized on every access to IsInputRedirected when input isn't redirected.  How about just boxing the result into a ```StrongBox<T>``` and using the existing EnsureInitialized method?  e.g.
```C#
private static StrongBox<bool> _isStdInRedirected;
...
public static bool IsInputRedirected
{
    get
    {
        StrongBox<bool> redirected = Volatile.Read(ref _isStdInRedirected) ??
            EnsureInitialized(ref _isStdInRedirected, () => new StrongBox<bool>(ConsolePal.IsInputRedirectedCore()));
        return redirected.Value;
    }
}
```

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -63,7 +63,8 @@ public static ConsoleColor BackgroundColor
 
         public static void ResetColor()
         {
-            if (!ConsoleOutIsTerminal)
+            // We only want to reset colors if we're targeting a TTY device
+            if (Console.IsOutputRedirected)
```
stephentoub: I don't think this is correct.  IsOutputRedirected is caching the result, but that will cause problems here if someone replaces the output stream with Console.SetOut.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -176,10 +210,9 @@ private static void ChangeColor(bool foreground, ConsoleColor color)
             }
 
             // Changing the color involves writing an ANSI character sequence out to the output stream.
-            // We only want to do this if we know that sequence will be interpreted by the output
-            // rather than simply displayed visibly.  A reasonable approximation for this is whether
-            // the underlying stream is our UnixConsoleStream and it's wrapping a character device.
-            if (!ConsoleOutIsTerminal)
+            // We only want to do this if we know that sequence will be interpreted by the output.
+            // rather than simply displayed visibly.
+            if (Console.IsOutputRedirected)
```
stephentoub: Same concern as the earlier case.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: You are right. I was still in the windows world where Console.SetOut() does not affect the IsXXXRedirected. I will change this to IsOutputRedirectedCore().

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: Same as above.

src/System.Console/tests/RedirectedStream.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using Xunit;
+
+public class RedirectedStream : RemoteExecutorTestBase
+{
+    [Fact] // the CI system redirects stdout, so we can only really test the redirected behavior.
```
stephentoub: I'm not clear on why the CI system affects our ability to test it not being redirected.  Can't we schedule a process that sends the result back through some other mechanism, e.g. writing the result to a pipe that we then read and verify here?

src/System.Console/tests/System.Console.Tests.csproj
```diff
@@ -46,6 +59,14 @@
       <Project>{F9DF2357-81B4-4317-908E-512DA9395583}</Project>
       <Name>System.Console</Name>
     </ProjectReference>
+	<ProjectReference Include="$(CommonTestPath)\System\Diagnostics\RemoteExecutorConsoleApp\RemoteExecutorConsoleApp.csproj">
+      <Project>{69e46a6f-9966-45a5-8945-2559fe337827}</Project>
+      <Name>RemoteExecutorConsoleApp</Name>
+    </ProjectReference>
+	<ProjectReference Include="..\..\System.Diagnostics.Process\src\System.Diagnostics.Process.csproj">
```
stephentoub: Nit: indentation seems off here

src/System.Console/tests/project.lock.json
```diff
@@ -424,11 +406,11 @@
           "lib/DNXCore50/System.Threading.ThreadPool.dll": {}
         }
       },
-      "xunit/2.1.0": {
+      "xunit/2.1.0-beta3-build3029": {
```
stephentoub: Why are these changing back to an earlier release?

src/System.Console/tests/RedirectedStream.cs
pallavit: I tried it, but even in that scenario the newly created process was also getting piped.

stephentoub: Really?  I don't know how that would even happen... what was it getting piped to?

src/System.Console/tests/project.lock.json
pallavit: XUnit.NetCore.Extensions most recent version depended on higher version of XUnit.NetCore than the one used by XUnit package, which was causing a build failure. So, as work-around I manually bumped the netcore extension version. This surfaces only after my change, since project.json file is updated which forces an update to project.lock.json

src/System.Console/tests/project.json
```diff
@@ -15,8 +16,8 @@
     "System.Threading.Tasks.Parallel": "4.0.0",
     "System.Threading.Thread": "4.0.0-beta-*",
     "System.Threading.ThreadPool": "4.0.10-beta-*",
-    "xunit": "2.1.0",
-    "xunit.netcore.extensions": "1.0.0-prerelease-*"
+    "xunit": "2.1.0-beta3-*",
```
weshaggard: looks like a stale reference This should be undone.

src/System.Console/tests/RedirectedStream.cs
pallavit: That got me curious as well and I am looking at it in parallel. However, my guess is that this happens because the parent process handles will also be inherited and so, if the parent process is redirected it affects the child. Note that if I run this explicitly by calling
CoreRun.exe xunit.console.netcore.exe System.Console.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests  -notrait category=OuterLoop -notrait category=failing
things work as expected.

src/System.Console/src/System/Console.cs
pallavit: I am not sure how StrongBox<T> is implemented but  won't we pay unboxing penalty with it every time?
Sorry, I am missing your concern here, after the first call to EnsureInitialized _stdXXXRedirectQueried should be set to true and so every subsequent call should return stdXXXRedirect directly w/o calling the initialization again, irrespective of it being true or false. Is not it?

src/System.Console/tests/System.Console.Tests.csproj
pallavit: Taken care.

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: We actually want the IsXxRedirected behavior to be the same across the platforms, right?  In which case SetOut should not affect it on Unix either?

Seems like we have two cases:
- IsXxRedirected checks the file descriptors for 0/1/2 to see if they're character devices
- Color ANSI color checking checks the file descriptors underlying the output stream to see if they're character devices

src/System.Console/src/System/Console.cs
stephentoub: > I am not sure how StrongBox is implemented but won't we pay unboxing penalty with it every time?

StrongBox is just a simple class like:
```C#
public class StrongBox<T>
{
    public T Value;
    public StrongBox() { }
    public StrongBox(T value) { Value = value; }
}
```

> Is not it?

Sorry, you're right, I was missing the difference in naming between _stdInRedirectQueried and _stdInRedirected.  Reading it quickly, I read them as the same value.

However, if you do stick with what you've done, you have a threading bug in your implementation.  You're setting _stdInRedirectQueried to true before you're setting _stdInRedirected; another thread could come along and see _stdInRedirectQueried as true and then proceed to read _stdInRedirected before you've actually initialized it.  That's one of the benefits of the ```StrongBox<T>``` approach: it's a single object, making it more difficult to get the threading matters wrong.  Normally I'd be against an approach that involved an extra allocation, but there's just a max of three of them for the whole process, so it's not a big deal, and it makes the code simpler IMHO, eliminated several fields, eliminating the extra initialization function, etc.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: > We actually want the IsXxRedirected behavior to be the same across the platforms, right? In which case SetOut should not affect it on Unix either?
I thought we intentionally do this on Unix? Why do we make the Color behave this way? On windows even Color is based on standard input, output and error.

pallavit: >We actually want the IsXxRedirected behavior to be the same across the platforms, right? In which case SetOut should not affect it on Unix either?

I thought we intentionally do this on Unix? Why do we make the Color behave this way? On windows even Color is based on standard input, output and error.

stephentoub: > I thought we intentionally do this on Unix?

We do, for the color changing stuff.  We didn't have IsXxRedirected before, so there was no intenionality about that :wink:.

> Why do we make the Color behave this way? On windows even Color is based on standard input, output and error.

Because on Unix we change the color by actually writing to the output stream.  If it's not going to something that knows how to interpret that output and remove the output from what's displayed, you end up with lots of funky characters in the output stream.  You could easily see this before with Jenkins, where these ended up on every line of the output log where the color was changed by xunit.

src/System.Console/tests/project.lock.json
ellismg: Can you please rebase your changes on top of master and see if that fixes the issue?  We worked through all of this stuff on last Friday.  These lock file updates are going to cause problems if we merge them as is.

pallavit: Taking a look at it now :)

stephentoub:
@dotnet-bot test this please (CI failed to copy a file)

src/System.Console/tests/System.Console.Tests.csproj
```diff
@@ -33,6 +33,19 @@
     <Compile Include="ThreadSafety.cs" />
     <Compile Include="XunitAssemblyAttributes.cs" />
     <Compile Include="TermInfo.cs" />
+	<Compile Include="RedirectedStream.cs" />
```
stephentoub: Nit: spacing

stephentoub:
LGTM

PR #3651, opened 2015-10-05
build: Unset %Platform% on invoking build.cmd

### LINKED ISSUE #69: build.cmd does not build solution on HP laptop (when Platform=MCD is pre-set)
I've got HP laptop with windows 7. When I run build.cmd, I've got the error "configuration is invalid"

    C:\Projects\dotnet\corefx>build.cmd
    C:\Projects\dotnet\corefx\src\System.Collections.Immutable.sln.metaproj : error MSB4126: указанная конфигурация решения "Release|MCD" недопустима. Укажите допустимую конфигурац
    ию решения с помощью свойств Configuration и Platform (например, MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") или оставьте эти свойства пустыми, чтобы
    использовать конфигурацию решения по умолчанию. [C:\Projects\dotnet\corefx\src\System.Collections.Immutable.sln]

The reason of this error is that environment variable `Platform` is set to `MCD` on the laptop by default. I have to change build command in the build.cmd by adding `/p:Platform="Any CPU"` to compile the project.

Build script should check for allowed platforms and generate user-friendly error message, which says what to do to successfully compile the project when the platform is not supported.

jasonwilliams200OK:
<del>If `%Platform%` is set to MCD, on invoking `build.cmd` it unsets the
var for local scope.</del>

Unset `%Platform%`, so build don't get disrupted by the incorrect value
set in evn variable.

Fixes #69.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -23,6 +23,11 @@ if not defined VisualStudioVersion (
 )
 
 :EnvSet
+:: Clear the 'Platform' env variable for this session,
+:: as it's a per-project setting within the build, and
+:: misleading value (such as 'MCD' in HP PCs) may lead
+:: to build breakage (issue: #69).
+set Platform=
 
 :: Log build command line
 set _buildproj=%~dp0build.proj


stephentoub:
cc: @nguerrera, @weshaggard

weshaggard:
@jasonwilliams200OK is this actually common enough to special case in build.cmd? I personally feel that we shouldn't special case this and people just need to ensure their Platform isn't set to some unexpected value.

nguerrera:
I'm OK with this one fix because it seems MCD is a common annoyance. However, I'll change my mind if we start special casing other msbuild variables like this.

jasonwilliams200OK:
@weshaggard, @nguerrera, I totally agree that this is the last thing we should be doing. But "Platform variable set to MCD in HP machines causing issues with MSBuild" is officially a very annoying issue. So, instead of doing defensive filtering in script, we should add a note in build guide for folks using HP systems (like @xplicit) to not to run:

```
.\build.cmd
```

on HP systems, but instead unset env var to achieve the default behavior:

```cmd
:: from cmd
SET %Platform%= && .\build.cmd
```

or


```powershell
# from powershell
$env:Platform=$null; .\build.cmd
```

With that said, should I amend the commit revoking this change and instead adding a note in build markdown in order to close #69?

weshaggard:
@jasonwilliams200OK I think clearing platform via `set Platform=` is better then special casing this. I would also be OK with checking for valid platforms, as I think there is another issue asking for that.

jasonwilliams200OK:
@weshaggard, I have removed the special casing. PTAL.

akoeplinger:
@jasonwilliams200OK can you update the commit message to match?

jasonwilliams200OK:
@akoeplinger, done. Also updated first comment and title of PR. :)

stephentoub:
@dotnet-bot test this please

PR #3653, opened 2015-10-05
Shim OpenSSL Evp_* methods.

eerhardt:
Adding shims for all OpenSSL methods in Interop.EVP.cs.


src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.EVP.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [DllImport(Libraries.CryptoNative)]
+        internal extern static void DestroyEvpMdContext(IntPtr ctx);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal extern static int GetEvpMdSize(IntPtr md);
+
+
+        [DllImport(Libraries.CryptoNative)]
+        internal extern static IntPtr CreateEvpMd5();
```
bartonjs: `Create` bothers me here, that feels like it's building something.  Instead it's accessing a shared pointer.  OpenSharedEvpMd5?  OpenEvpMd5Algorithm?

bartonjs:
Structurally LGTM; if no one else objects to the naming of the algorithm pointers I won't put up a fight :)

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.EVP.cs
eerhardt: I can rename them. I didn't realize it wasn't creating a new instance.

How about just GetEvpMd5?

bartonjs: Seems happier than "Create" :).  Good with me (it feels a bit lacking, but I'm drawing a blank on a better verb)

nguerrera: Get works for me.

src/Native/System.Security.Cryptography.Native/evp.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+/*
+Function:
+InitializeEvpDigest
+
+Direct shim to EVP_DigestInit_ex.
+*/
+extern "C" int32_t InitializeEvpDigest(EVP_MD_CTX* ctx, const EVP_MD* type, ENGINE* impl)
+{
+    return EVP_DigestInit_ex(ctx, type, impl);
```
nguerrera: For 1:1 functions, we should just use our casing and otherwise leave the name alone: EvpDigestInitEx, etc.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.EVP.cs
nguerrera: Actually, see my other comment re 1:1 if that applies here.

src/Native/System.Security.Cryptography.Native/evp.cpp
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Microsoft. All rights reserved.
```
nguerrera: Use the name pal_evp.cpp to match the pattern we've got elsewhere (except for the grandfathered openssl.c)

nguerrera: Also, another pattern we have in the shims is to have a header documenting the public interface. I actually went back and forth on this and considered just having .cpp since we only P/Invoke, but when I discussed it last with @socket, he felt there was value in keeping the headers, so we stuck with it.

nguerrera:
The change looks good to me semantically. I only had the comments about naming and use of a header.

eerhardt:
Thanks for the great feedback. I've updated the code as per suggestions. Let me know what you think.

nguerrera:
Thanks. LGTM.

sokket:
LGTM

stephentoub:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

PR #3654, opened 2015-10-06
Editing Socket References

josguil:
Some constructors that are doing complex operations (like interop or
trying to connect over the network) are removed.

APM methods are removed and replaced by TPL equivalent methods.


josguil:
While I correctly sync my branch so it doesn't show 125 commits, PTAL at the changed file:

@CIPop , @davidsh , @SidharthNabar , @pgavlin

PR #3655, opened 2015-10-06
Editing Socket References

josguil:
Some constructors that are doing time consuming operations for constructors (like interop or
trying to connect over the network) are removed.

APM methods are removed and replaced by TPL equivalent methods.

### DIFF
--- a/src/System.Net.Sockets/ref/System.Net.Sockets.cs
+++ b/src/System.Net.Sockets/ref/System.Net.Sockets.cs
@@ -388,19 +388,19 @@ namespace System.Net.Sockets
 
     public static partial class SocketTaskExtensions
     {
-        public static System.Threading.Tasks.Task<Socket> AcceptAsync(this System.Net.Sockets.Socket thisSocket) { return default(System.Threading.Tasks.Task<Socket>); }
-        public static System.Threading.Tasks.Task<Socket> AcceptAsync(this System.Net.Sockets.Socket thisSocket, System.Net.Sockets.Socket acceptSocket) { return default(System.Threading.Tasks.Task<Socket>); }
-        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket thisSocket, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task); }
-        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket thisSocket, System.Net.IPAddress address, int port) { return default(System.Threading.Tasks.Task); }
-        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket thisSocket, System.Net.IPAddress[] addresses, int port) { return default(System.Threading.Tasks.Task); }
-        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket thisSocket, string host, int port) { return default(System.Threading.Tasks.Task); }
-        public static System.Threading.Tasks.Task<int> ReceiveAsync(this System.Net.Sockets.Socket thisSocket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
-        public static System.Threading.Tasks.Task<int> ReceiveAsync(this System.Net.Sockets.Socket thisSocket, System.Collections.Generic.IList<System.ArraySegment<byte>> buffers, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
-        public static System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveFromResult> ReceiveFromAsync(this System.Net.Sockets.Socket thisSocket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveFromResult>); }
-        public static System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveMessageFromResult> ReceiveMessageFromAsync(this System.Net.Sockets.Socket thisSocket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveMessageFromResult>); }
-        public static System.Threading.Tasks.Task<int> SendAsync(this System.Net.Sockets.Socket thisSocket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
-        public static System.Threading.Tasks.Task<int> SendAsync(this System.Net.Sockets.Socket thisSocket, System.Collections.Generic.IList<System.ArraySegment<byte>> buffers, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
-        public static System.Threading.Tasks.Task<int> SendToAsync(this System.Net.Sockets.Socket thisSocket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<int>); }
+        public static System.Threading.Tasks.Task<Socket> AcceptAsync(this System.Net.Sockets.Socket socket) { return default(System.Threading.Tasks.Task<Socket>); }
+        public static System.Threading.Tasks.Task<Socket> AcceptAsync(this System.Net.Sockets.Socket socket, System.Net.Sockets.Socket acceptSocket) { return default(System.Threading.Tasks.Task<Socket>); }
+        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket socket, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task); }
+        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket socket, System.Net.IPAddress address, int port) { return default(System.Threading.Tasks.Task); }
+        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket socket, System.Net.IPAddress[] addresses, int port) { return default(System.Threading.Tasks.Task); }
+        public static System.Threading.Tasks.Task ConnectAsync(this System.Net.Sockets.Socket socket, string host, int port) { return default(System.Threading.Tasks.Task); }
+        public static System.Threading.Tasks.Task<int> ReceiveAsync(this System.Net.Sockets.Socket socket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
+        public static System.Threading.Tasks.Task<int> ReceiveAsync(this System.Net.Sockets.Socket socket, System.Collections.Generic.IList<System.ArraySegment<byte>> buffers, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
+        public static System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveFromResult> ReceiveFromAsync(this System.Net.Sockets.Socket socket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveFromResult>); }
+        public static System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveMessageFromResult> ReceiveMessageFromAsync(this System.Net.Sockets.Socket socket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<System.Net.Sockets.SocketReceiveMessageFromResult>); }
+        public static System.Threading.Tasks.Task<int> SendAsync(this System.Net.Sockets.Socket socket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
+        public static System.Threading.Tasks.Task<int> SendAsync(this System.Net.Sockets.Socket socket, System.Collections.Generic.IList<System.ArraySegment<byte>> buffers, System.Net.Sockets.SocketFlags socketFlags) { return default(System.Threading.Tasks.Task<int>); }
+        public static System.Threading.Tasks.Task<int> SendToAsync(this System.Net.Sockets.Socket socket, System.ArraySegment<byte> buffer, System.Net.Sockets.SocketFlags socketFlags, System.Net.EndPoint remoteEndPoint) { return default(System.Threading.Tasks.Task<int>); }
     }
     public enum SocketType
     {
@@ -454,6 +454,7 @@ namespace System.Net.Sockets
         public UdpClient() { }
         public UdpClient(int port) { }
         public UdpClient(int port, System.Net.Sockets.AddressFamily family) { }
+        public UdpClient(System.Net.IPEndPoint localEP) { }
         public UdpClient(System.Net.Sockets.AddressFamily family) { }
         protected bool Active { get { return default(bool); } set { } }
         public int Available { get { return default(int); } }
@@ -472,7 +473,6 @@ namespace System.Net.Sockets
         public void JoinMulticastGroup(System.Net.IPAddress multicastAddr, int timeToLive) { }
         public void JoinMulticastGroup(System.Net.IPAddress multicastAddr, System.Net.IPAddress localAddress) { }
         public System.Threading.Tasks.Task<System.Net.Sockets.UdpReceiveResult> ReceiveAsync() { return default(System.Threading.Tasks.Task<System.Net.Sockets.UdpReceiveResult>); }
-        public System.Threading.Tasks.Task<int> SendAsync(byte[] datagram, int bytes) { return default(System.Threading.Tasks.Task<int>); }
         public System.Threading.Tasks.Task<int> SendAsync(byte[] datagram, int bytes, System.Net.IPEndPoint endPoint) { return default(System.Threading.Tasks.Task<int>); }
         public System.Threading.Tasks.Task<int> SendAsync(byte[] datagram, int bytes, string hostname, int port) { return default(System.Threading.Tasks.Task<int>); }
     }
--- a/src/System.Net.Sockets/ref/System.Net.Sockets.cs
+++ b/src/System.Net.Sockets/ref/System.Net.Sockets.cs
@@ -182,27 +182,9 @@ namespace System.Net.Sockets
         public short Ttl { get { return default(short); } set { } }
         public System.Net.Sockets.Socket Accept() { return default(System.Net.Sockets.Socket); }
         public bool AcceptAsync(System.Net.Sockets.SocketAsyncEventArgs e) { return default(bool); }
-        public System.Threading.Tasks.Task<Socket> AcceptAsync() { return default(System.Threading.Tasks.Task<Socket>); }
-        public System.Threading.Tasks.Task<Socket> AcceptAsync(int receiveSize) { return default(System.Threading.Tasks.Task<Socket>); }
-        public System.Threading.Tasks.Task<Socket> AcceptAsync(System.Net.Sockets.Socket acceptSocket, int receiveSize) { return default(System.Threading.Tasks.Task<Socket>); }
-        public System.Threading.Tasks.Task ConnectAsync(System.Net.EndPoint remoteEP) { return default(System.Threading.Tasks.Task); }
-        public System.Threading.Tasks.Task ConnectAsync(System.Net.IPAddress address, int port) { return default(System.Threading.Tasks.Task); }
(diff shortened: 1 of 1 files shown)


josguil:
PTAL:

@CIPop , @davidsh , @SidharthNabar , @pgavlin

stephentoub:
cc: @ericstj

pgavlin:
Is this intended to be the final shape of the contract? I see that the .ctor that takes a `SocketInformation` instance is gone, but the related methods/types (`Socket.DuplicateAndClose`, `SocketInformation`, and `SocketInformationOptions`) still exist. These APIs should also be removed.

FWIW, here are a few more problematic APIs from a cross-platform point-of-view:
- `SocketFlags.Broadcast`, `SocketFlags.MaxIOVectorLength`, `SocketFlags.Multicast`, `SocketFlags.Partial`: AFAICT these flags have no equivalent on other platforms.
- A number of the `ProtocolType` entries have no equivalent on other platforms.
- `Socket.IOControl`, `Socket.Disconnect`, `Socket.DisconnectAsync`, : these APIs are simply are not supportable in any reasonable fashion on non-Windows platforms.
- The instance-method variants of `Socket.Connect` are not generally cross-platform: there are important platforms that do not support multiple connection attempts on the same socket (e.g. OS X). We can support these APIs halfway by probing each address on a new socket until a usable address is found and then connecting to that address with the user's socket, but this isn't really a fix, as the socket will become unusable if the final connection attempt fails. To make a long story short: these methods expose differences in OS semantics that cannot be papered over. Replacing these APIs with static variants that return a connected `Socket` would be sufficient.
- `Socket.GetSocketOption`, `Socket.SetSocketOption`, `SocketOptionLevel`, and `SocketOptionName`: these aren't necessarily candidates for removal, but they are pretty liberal w.r.t. the inputs they accept. Ideally we'd remove these APIs or limit the set of acceptable (Level, Name) pairs to only those that are known to work on all platforms (e.g. [those specified by the Open Group](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html))
- `Socket.SendPacketsAsync` and `SendPacketsElement`: these are a bit fuzzy. It's probably possible to fake these APIs up on other platforms, but I'm not sure that it's worth the effort. Do we have usage data for these?

I'll add further comments to this PR for anything else that pops up.

pgavlin:
A few more thoughts re: `Connect`: we might be able to support multiple connection attempts on the same socket on other platforms by creating a new OS socket per-attempt and using the first OS socket that connects successfully (if any) as the underlying OS socket for the `Socket` instance. This may come with significant implementation challenges, however, esp. around errors returned for operations on disconnected sockets and sockets for non-connection-oriented protocols.

One other behavior that is slightly problematic is the ability to send/receive a first blob of data with a connect/accept, respectively. This is also supportable cross-platform with extra work (it's not an OS-provided behavior on non-Windows OSes), but I'd prefer to cut the API if the capability isn't often used.

josguil:
Hi Pat, thanks for your input.

This is not intended to be the final version of the contract. There are still missing all the changes you mentioned. I'm only doing changes in constructors and APM methods. For all the other changes, we'll need to coordinate to either add them to this PR or deciding on having them in a different one.

stephentoub:
@dotnet-bot test this please

src/System.Net.Sockets/ref/System.Net.Sockets.cs
```diff
@@ -156,7 +156,6 @@ public partial class SendPacketsElement
     public partial class Socket : System.IDisposable
     {
         public Socket(System.Net.Sockets.AddressFamily addressFamily, System.Net.Sockets.SocketType socketType, System.Net.Sockets.ProtocolType protocolType) { }
-        public Socket(System.Net.Sockets.SocketInformation socketInformation) { }
```
CIPop: @josguil What was the blocking Win32 API in the code-path that caused this to be removed?

src/System.Net.Sockets/ref/System.Net.Sockets.cs
```diff
@@ -210,23 +208,8 @@ public partial class Socket : System.IDisposable
         public void Connect(string host, int port) { }
         public bool ConnectAsync(System.Net.Sockets.SocketAsyncEventArgs e) { return default(bool); }
         public static bool ConnectAsync(System.Net.Sockets.SocketType socketType, System.Net.Sockets.ProtocolType protocolType, System.Net.Sockets.SocketAsyncEventArgs e) { return default(bool); }
-        public void Disconnect(bool reuseSocket) { }
```
CIPop: @josguil If we remove `Disconnect` and `DisconnectAsync` we'll need to also remove `SocketAsyncEventArgs.DisconnectReuseSocket`.

src/System.Net.Sockets/ref/System.Net.Sockets.cs
```diff
@@ -321,7 +303,6 @@ public enum SocketFlags
         Broadcast = 1024,
         ControlDataTruncated = 512,
         DontRoute = 4,
-        MaxIOVectorLength = 16,
         Multicast = 2048,
         None = 0,
         OutOfBand = 1,
```
CIPop: Please also remove `SocketInformationOptions.UseOnlyOverlappedIO`.

src/System.Net.Sockets/ref/System.Net.Sockets.cs
```diff
@@ -455,16 +426,8 @@ public partial class TcpListener
...
         public System.Threading.Tasks.Task<System.Net.Sockets.TcpClient> AcceptTcpClientAsync() { return default(System.Threading.Tasks.Task<System.Net.Sockets.TcpClient>); }
-        public void AllowNatTraversal(bool allowed) { }
-        public System.IAsyncResult BeginAcceptSocket(System.AsyncCallback callback, object state) { return default(System.IAsyncResult); }
-        public System.IAsyncResult BeginAcceptTcpClient(System.AsyncCallback callback, object state) { return default(System.IAsyncResult); }
-        public static System.Net.Sockets.TcpListener Create(int port) { return default(System.Net.Sockets.TcpListener); }
-        public System.Net.Sockets.Socket EndAcceptSocket(System.IAsyncResult asyncResult) { return default(System.Net.Sockets.Socket); }
-        public System.Net.Sockets.TcpClient EndAcceptTcpClient(System.IAsyncResult asyncResult) { return default(System.Net.Sockets.TcpClient); }
         public bool Pending() { return default(bool); }
         public void Start() { }
         public void Start(int backlog) { }
```
CIPop: @josguil @pgavlin @SidharthNabar We probably want to remove `TransmitFileOptions` and the `SendPacketsFlags` property on `SocketAsyncEventArgs`.

src/System.Net.Sockets/ref/System.Net.Sockets.cs
```diff
@@ -483,11 +446,7 @@ public enum TransmitFileOptions
     public partial class UdpClient : System.IDisposable
     {
         public UdpClient() { }
-        public UdpClient(int port) { }
-        public UdpClient(int port, System.Net.Sockets.AddressFamily family) { }
-        public UdpClient(System.Net.IPEndPoint localEP) { }
         public UdpClient(System.Net.Sockets.AddressFamily family) { }
-        public UdpClient(string hostname, int port) { }
```
CIPop: @josguil We need at least one of the ctors to Bind the UDP connectionless socket. Otherwise ReceiveAsync won't be able to function at all.
The underlying Win32 API shouldn't be blocking since no actual I/O is performed.

src/System.Net.Sockets/ref/System.Net.Sockets.cs
pgavlin: Whether or not there is a blocking Win32 call here, this API is a Win32-ism and should be removed (along with `Socket.DuplicateAndClose`, `class SocketInformation`, and `enum SocketInformationOptions`).

josguil: @CIPop  It was:
                    UnsafeSocketsNativeMethods.OSSOCK.WSAPROTOCOL_INFO protocolInfo = 
                        (UnsafeSocketsNativeMethods.OSSOCK.WSAPROTOCOL_INFO)Marshal.PtrToStructure<UnsafeSocketsNativeMethods.OSSOCK.WSAPROTOCOL_INFO>((IntPtr)pinnedBuffer);

src/System.Net.Sockets/ref/System.Net.Sockets.cs
josguil: Done for next PR

src/System.Net.Sockets/ref/System.Net.Sockets.cs
josguil: Done

src/System.Net.Sockets/ref/System.Net.Sockets.cs
josguil: Done

src/System.Net.Sockets/ref/System.Net.Sockets.cs
josguil: Ok, I thought we didn't want any of the constructors to do any connect-like operations, do we have any preference on the one that should be added back? What is the most used?

src/System.Net.Sockets/ref/System.Net.Sockets.cs
CIPop: That's not really a Win32 call but @pgavlin is right: let's remove this and related types.

CIPop:
@josguil Except the first comment which still requires changes, this looks good to me.
Given that we first need to change the current System.Private.Networking to avoid any broken builds internally, we can't merge this right away (marking it to avoid accidental merges).
We'll also need to work out the current SqlClient dependency and ensure that CoreFX will build correctly once the package is created.

@pgavlin please take another look and let us know if there are any other APIs that we should be removing.

@terrajobst @weshaggard @ericstj : FXDC, please let us know if there are any additional concerns over the proposed contract surface for the final Sockets 4.1 contract.

At this point, my personal recommendation is not to add a legacy contract with the APM APIs. Instead, our team could provide a simple extension class that allows conversion from TPL to APM.

josguil:
@stephentoub  @CIPop @davidsh @pgavlin  
Please take one last look. 
As discussed, I'll not be merging right now this PR, before I'll be applying the related changes in source code in project K.

pgavlin:
@terrajobst @weshaggard @ericstj @stephentoub: any opinions on the proposed surface area here?

weshaggard:
We should work with @terrajobst to get this through the API Review http://aka.ms/apireview.

markmnl:
#2936 is resolved by changing the APM signatures to TPL

`SetIPProtectionLevel(IPProtectionLevel)` Can be covered by `SetSocketOption()`?
`Blocking`
`RecieveFrom()` blocking, unless non blocking socket or connection-less type (i.e. UDP)
`SendTo()` blocking, unless non blocking socket or connection-less type (i.e. UDP)

Are all missing which we rely on in .NET Frameworks  `System.Net.Sockets.Socket` today.

`SetIPProtectionLevel(IPProtectionLevel)` there is not equivalent on other OS's that I am aware of however if we do not set it in a desktop .NET app today we cannot traverse NAT even if only IPv4 on Windows. I raised the issue on [Connect here](https://connect.microsoft.com/VisualStudio/feedback/details/1431526/nat-traversal-udp-hole-punching-not-possible-in-windows-store-apps-even-with-internet-client-server-capability). (Using Mono on other platforms `SetIPProtectionLevel(IPProtectionLevel)` is not implemented but traversal works without it anyway). *My concern is if `SetIPProtectionLevel(IPProtectionLevel)` or equivalent is not implemented in .NET Core, NAT traversal on Windows will not be possible thus P2P will not be possible.*

`Blocking`, `RecieveFrom()` and `SendTo()` would be sorely missed - we send and receive datagrams up to 60 times a second from our main game loop - *if the only way to send and receive will be using `async` TPL methods I imagine that would add significant overhead*. You can see how we use  .NET Framework's `System.Net.Sockets.Socket` today here: https://github.com/markmnl/FalconUDP/blob/single-threaded/SocketTransceiver.cs

CIPop:
@markmnl 
>  Blocking ,  RecieveFrom()  and  SendTo()  would be sorely missed - we send and receive datagrams up to 60 times a second from our main game loop 

It is by design that all blocking I/O have been removed. Blocking a thread for I/O is a bad design in _any_ situation and on any kernel/OS. In your case, given the high-perf aspect I recommend redesigning your code to use `SocketAsyncEventArgs`. TPL will work too although the simpler code comes with certain overheads as well.

> SetIPProtectionLevel(IPProtectionLevel)  Can be covered by  SetSocketOption() ?
Yes: that was what the original implementation was using. 

> My concern is if  SetIPProtectionLevel(IPProtectionLevel)  or equivalent is not implemented in .NET Core, NAT traversal on Windows will not be possible thus P2P will not be possible.

This is a good point. /cc: @pgavlin, @SidharthNabar  for the outlined scenario. 

We may need to ensure that either:
    a. We make the SetSocketOption available on other OSs that support it.
    b. We provide multiple OS-specific tests in our FunctionalTests that demonstrate how the scenario is still achievable. (The tests should include documentation on what OSs don't support the scenario.)

markmnl:
@CIPop 

`RecieveFrom()` and `SendTo()` are not blocking if `Socket.Blocking` is set to false or if the protocol type is connectionless - this makes sense as sending UDP datagrams is not an asynchronous operation - there is nothing to await. 

`SocketAsyncEventArgs` methods runs callbacks on a thread pool thread... this may be good for web server but in a single threaded application one now has to take care of marshalling everything in a thread safe manner so your application can use it... 

TPL methods are not suited to cyclic architectures, games, where everything runs at a recurring rate frequently (granted most of your use cases will be event based architectures e.g. UI or web, i.e. do something when clicked or an http request received). In cyclic architecture when will my `ReceiveAsync()` call run - I need to run at a specific point in the game loop so I would have to stash the result in order to do that or I would have to hack around the `TPL` and block on waiting for the `Task` to complete....

Please please include `ReceiveFrom()` and `SendTo()` :S

markmnl:
I suppose I could workaround it by spawning a new `Thread` that just sat there dealing with the net IO - enquing sends by the application and received datagrams so I can replicate the `ReceiveFrom()` and `SendTo()` functionality in my own wrapper. Though that does come with the overhead of an extra `Thread`.

CIPop:
>  RecieveFrom()  and  SendTo()  are not blocking if  Socket.Blocking  is set to false or if the protocol type is connectionless.

You're right! We did include all other non-blocking I/O BSD-like APIs but somehow we missed these. Thanks for clarifying!

@josguil, @SidharthNabar please review the contract surface to ensure we didn't miss any other non-blocking BSD style APIs.

>  SocketAsyncEventArgs  methods runs callbacks on a thread pool thread... this may be good for web server but in a single threaded application one now has to take care of marshalling everything in a thread safe manner so your application can use it... 

There's an (old, yet probably not very far from truth with respect to blocking/non-blocking) I/O perf comparison published in http://www.microsoft.com/mspress/books/5726.aspx

CIPop:
> I suppose I could workaround it by spawning a new  Thread  that just sat there dealing with the net IO - enquing sends by the application and received datagrams so I can replicate the  ReceiveFrom()  and  SendTo()  functionality in my own wrapper. Though that does come with the overhead of an extra  Thread .

Actually you should try to perform the work in the IOCP callbacks as operations complete. That way you span across all available cores and you don't keep any thread blocked for I/O.

Please start a different thread for this discussion. It's falling out of scope from the original PR.

Thank you again for pointing out that the missing 2 APIs can be non-blocking!

markmnl:
@CIPop Thanks! 

(I was wrong saying: "...not blocking if `Blocking` is set to false or connectionless...", it is only when `Blocking` is set to false. Typical usage case is here: https://github.com/markmnl/FalconUDP/blob/single-threaded/SocketTransceiver.cs.

With regards to `SetIPProtectionLevel(IPProtectionLevel)` setting up tests may be tricky as you have to physically have 3 machines to test NAT traversal with 2 behind different NAT. The technique used (same used by STUN before resorting to relay) is best described in RFC 5128: https://tools.ietf.org/html/rfc5128#page-11

josguil:
@markmnl 

Hi, I'm looking at the current reference, I could not understand your first comment.  ReceiveFrom and SendTo are included in this contract. Were you perhaps referring to the BeginReceiveFrom and BeginSendTo methods?

markmnl:
@josguil 

So it they are! My bad - I must have looked at only the first commit's diff instead of the whole file. Apologies :S (I do not care for Begin, End ASM methods).

markmnl:
@josguil Although `Socket.Blocking` is not there so how would I use `SentTo()` and `ReceiveFrom()` so they are non-blocking?

SidharthNabar:
@markmnl - Just going back to your point about `SetIPProtectionLevel` - I'm not sure I follow. What we are proposing is that you can just replace calls to `Socket.SetIPProtectionLevel` with calls to `Socket.SetSockOption` with `SocketOptionName = IPProtectionLevel`. We are not proposing removing that functionality. 

I think the challenges that you mention should would apply for either of these approaches, so the proposed revision doesn't affect it. Please let me know if I am missing something. Thanks!

josguil:
@markmnl  

Blocking is also there. It's hidden by GitHub but if you expand the code you can see it.

SidharthNabar:
@pgavlin - Just want to quickly recap your feedback in the initial discussion to make sure we have captured everything

>  SocketFlags.Broadcast ,  SocketFlags.MaxIOVectorLength ,  SocketFlags.Multicast ,  SocketFlags.Partial : AFAICT these flags have no equivalent on other platforms

We have removed SocketFlags.MaxIOVectorLength, but other than that, we really want to avoid sub-setting enumerations as much as possible. We should be able to ignore/throw PNSE on other platforms, that don't support these values.

> A number of the  ProtocolType  entries have no equivalent on other platforms.

The ProtocolType enum values are based on the IANA assigned list of protocols, and as such, is not tied to a Windows-specific concept. In this case, we want to avoid sub-setting the enumeration and continue to include all current values.

> Socket.IOControl ,  Socket.Disconnect ,  Socket.DisconnectAsync , : these APIs are simply are not supportable in any reasonable fashion on non-Windows platforms.

@CIPop mentioned that IOControl is a fairly well established concept even outside of Windows, and as such, we believe we should continue to support it. For SocketDisconnect, DisconnectAsync - agreed. We have removed those.

> The instance-method variants of  Socket.Connect  are not generally cross-platform: there are important platforms that do not support multiple connection attempts on the same socket (e.g. OS X). We can support these APIs halfway by probing each address on a new socket until a usable address is found and then connecting to that address with the user's socket, but this isn't really a fix, as the socket will become unusable if the final connection attempt fails. To make a long story short: these methods expose differences in OS semantics that cannot be papered over. Replacing these APIs with static variants that return a connected  Socket  would be sufficient.

These were already shipped as part of the 4.0 version and cannot be removed now due to backward compatibility requirement.

> Socket.GetSocketOption ,  Socket.SetSocketOption ,  SocketOptionLevel , and  SocketOptionName : these aren't necessarily candidates for removal, but they are pretty liberal w.r.t. the inputs they accept. Ideally we'd remove these APIs or limit the set of acceptable (Level, Name) pairs to only those that are known to work on all platforms (e.g. those specified by the Open Group)

As mentioned above, we want to avoid sub-setting an enum as much as possible. Hence, we propose to keep supporting all enum values in SocketOptionName and we can either no-op or throw PNSE on platforms where they aren't supported as of now. 

> Socket.SendPacketsAsync  and  SendPacketsElement : these are a bit fuzzy. It's probably possible to fake these APIs up on other platforms, but I'm not sure that it's worth the effort. Do we have usage data for these?

@CIPop mentioned that these are very performant and easy-to-use methods. As such, we would like to keep including them.

pgavlin:
> mentioned that IOControl is a fairly well established concept even outside of Windows

Yes, it is a fairly well-established concept, but it is very much platform-specific: there is no way to write platform-independent code that uses ioctls, as each platform has its own set. This is my objection to the API.

As far as the enum entries: the real problem here is not the range of values exactly. It is that the APIs that accept these values are expected to work with values that are valid within the scope of winsock but that are not named in the enumerated types (in general, the Sockets API is intended to be pretty transparent to the underlying platform). This runs directly against our ability to PAL out these APIs: in order to be compatible with code that was written for Windows, we need to translate as many of these values as possible, but in order to be faithful to the spirit of the API (and maximally useful), we must let unknown values pass through to the underlying platform. There is no solution I can see other than either
a) breaking code that expects to be able to pass values outside the range of the relevant enums through to the underlying platform, or
b) adding new APIs that explicitly make no promise of passing values to the underlying platform (e.g. `Socket.ctor(int, int, int)` alongside `Socket.ctor(AddressFamily, SocketType, ProtocolType)`.

On its face (b) is a reasonable enough compromise, but it would *substanitally* increase the surface area for S.N.Sockets if taken to its natural extreme (every API that accepts any of `SocketFlags`, `AddressFamily`, `SocketType`, `ProtocolType`, `SocketOptionName`, `SocketOptionLevel`, et. al. would need to have a version that accepted appropriately-sized integers).

pgavlin:
I just realized that I forgot to list one other approach: deny the ability to pass unknown values through to the underlying platform on *nix. This is a pragmatic approach, and this is what we currently plan on doing.

markmnl:
@SidharthNabar @josguil If `Blocking` is there and the equivalent of `SetIPProtectionLevel()` is too, I apologise for the unnecessary debate. (I used `Socket` in a UWP app recently and `DatagramSocket` in Win RT both of which had neither making them unusable for our purpose - which is why I was concerned - looking forward to using `Socket` in .NET Core).

josguil:
@CIPop @davidsh @SidharthNabar @pgavlin 

After some more in depth analysis of the contract, we've discovered some changes needed in order to support better TPL to APM (Begin/End pattern) wrappers. 
By @stephentoub 's suggestion, we decided to make the return type in the ``ReceiveFromAsync`` and ``ReceiveMessageFromAsync`` methods a struct that includes all the ref and out parameters as well as the already returned int type that we had in the previous version. 

We've also discovered that the SendAsync method in UdpClient with only datagram and bytes as parameters can not be supported since we are removing the Connect related APIs it depends on. Additionally, the constructor for UDPClient receiving an IPEndPoint is added back after some agreement that the bind operation it does it's acceptable and similar to other constructors in the same class.

josguil:
Change done in this commit https://github.com/josguil/corefx/commit/a3050bacc9b716f52f086751bfb4bf0384019f53.

PR #3670, opened 2015-10-06
Remove project references in crypto libraries

### LINKED ISSUE #2521: Remove crypto P2P references when possible
The crypto contract refresh required changing the package references to project references because they were interdependent.  They should go back to package references when able.

bartonjs:
* Move ProjectReference to project.json
* RSA's derived types were using a protected field that isn't in the contract, so now they implement it with no secret ability.
* Removed the Console references from the crypto test libraries that were not using Console
* Had to split the project.json for X509Certificates into two, because project.json doesn't support the conditional compilation dependencies required for X509
    * Windows: S.S.C.Cng + S.S.C.Csp
    * Unix: S.S.C.OpenSsl

Fixes #2521.

### DIFF
--- a/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
+++ b/src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
@@ -27,7 +27,6 @@ namespace System.Security.Cryptography
         /// <exception cref="CryptographicException">if <paramref name="keySize" /> is not valid</exception>
         public RSACng(int keySize)
         {
-            _legalKeySizesValue = s_legalKeySizes;
             KeySize = keySize;
         }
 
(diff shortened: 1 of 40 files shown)


bartonjs:
cc: @ericstj

stephentoub:
@dotnet-bot test this please

jasonwilliams200OK:
@dotnet-bot, test this please. (`A process is hung on the target machine and preventing it from cleaning up the workspace`)

bartonjs:
cc: @weshaggard @stephentoub

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
```diff
@@ -48,24 +47,28 @@ public RSACng(CngKey key)
             if (key.AlgorithmGroup != CngAlgorithmGroup.Rsa)
                 throw new ArgumentException(SR.Cryptography_ArgRSAaRequiresRSAKey, "key");
 
-            _legalKeySizesValue = s_legalKeySizes;
             Key = CngAlgorithmCore.Duplicate(key);
         }
 
+        public override KeySizes[] LegalKeySizes
```
ericstj: Can you update the reference assembly?  There are some edge cases where the compiler needs to know that a method is overridden.

ericstj:
LGTM

weshaggard:
LGTM thanks @bartonjs for following up and removing the P2P references.

src/System.Security.Cryptography.Cng/src/System/Security/Cryptography/RSACng.cs
bartonjs: > There are some edge cases where the compiler needs to know that a method is overridden.

Oh, right, "base.Member" being statically bound, and all that.  Good catch!

ericstj: We need to get APIcompat running and add a rule for this.

stephentoub:
@dotnet-bot test this please (all of the tests seem to have passed, but then Jenkins seems to have failed)

bartonjs:
@dotnet-bot test this please (looks like some Azure communications errors, maybe?)

PR #3675, opened 2015-10-06
Remove dead code from S.N.{Security,Sockets}.

pgavlin:
This change removes unused, internal-only APM APIs from the aforementioned assemblies as well as related types.


pgavlin:
@CIPop @davidsh PTAL. Please let me know if the removed APM APIs in S.N.Security are simply latent--there isn't anything checked in that was using them.

CIPop:
/cc @vijaykota @bartonjs 

Did a quick lookup in desktop code. The internal Security APIs were previously used by TlsStream and provide an APM Async implementation of SslStream. It doesn't appear to be usable via public APIs.
I think it's ok to remove at this time: we can always bring it back if we need to optimize.

TlsStream was the only reason for `NetworkStream::UnsafeBeginMultipleWrite`.

LGTM

Thanks @pgavlin for detecting and removing the unused code!

bartonjs:
Nothing here was something I was expecting to modify, so fine with me :)

pgavlin:
@dotnet-bit test this please

pgavlin:
@dotnet-bot test this please

PR #3676, opened 2015-10-06
Use custom test-runtime project.json

ellismg:
Today, the test-runtime project.json and project.lock.json we have comes
from the buildtools package. This is less than ideal for a few reasons:

 - To change anything in here requires that we roll a new version of
   build tools and consume that.
 - The version in buildtools does not have runtime sections for either
   Ubuntu 14.04 or OSX 10.10, so when we restore a test runtime we
   always get the windows version. This leads us to having to do upstack
   hacks like "run-test.sh" which tries to replace the windows versions
   of things with correct cross platform versions.

Now that we are publishing cross platform packages, we should be using
them when we run tests. This change introduces a copy of the test
runtime project.json we can version at a different cadence than build
tools and updates the build to use it instead.

One artifact of this change is you can now pass
/p:TestNugetRuntimeId="ubuntu.14.04-x64" to the build and the test
folders produced will actually be runable on Ubuntu 14.04 (and you could
do a similar thing for osx.10.10-x64).

The next step will be to start ripping out code in run-test.sh so we
don't actually do a bunch of extra copies we don't need (e.g. we can
remove all of the CoreFX managed copying).

This also moves us one step closer to being able to easily update the
version of the runtime we test against.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -43,6 +43,12 @@
     <ToolsDir Condition="'$(ToolsDir)'==''">$(PackagesDir)Microsoft.DotNet.BuildTools.$(BuildToolsVersion)/lib/</ToolsDir>
   </PropertyGroup>
 
+  <!-- Test runtime -->
+  <PropertyGroup>
+    <TestRuntimeProjectJson Condition="'$(TestRuntimeProjectJson)' == ''">$(SourceDir)Common/test-runtime/project.json</TestRuntimeProjectJson>
+    <TestRuntimeProjectLockJson Condition="'$(TestRuntimeProjectLockJson)' == ''">$(SourceDir)Common/test-runtime/project.lock.json</TestRuntimeProjectLockJson>
+  </PropertyGroup>
+
   <!-- list of nuget package sources passed to nuget.exe -->
   <ItemGroup>
     <NuGetSourceList Include="https:%2F%2Fwww.myget.org/F/dotnet-buildtools" />
--- a/src/Common/test-runtime/project.json
+++ b/src/Common/test-runtime/project.json
@@ -1,0 +1,27 @@
+{
+    "dependencies": {
+      "Microsoft.NETCore.Platforms": "1.0.1-beta-23328",
+      "Microsoft.NETCore.TestHost": "1.0.0-beta-23328",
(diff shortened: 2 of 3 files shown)


ellismg:
/cc @weshaggard @stephentoub @naamunds 

I have corresponding changes to run-test.sh coming to consume this, but I think I need to get this in first to stage things correctly with CI.

weshaggard:
LGTM assuming CI passes.

ellismg:
@dotnet-bot test this please (failure was due to new worker machines not having some software installed, I fixed the machines up and will update the base image shortly).

joperezr:
@ellismg does this mean that we will remove the project.json from buildtools as well? or will we keep both of them?

ellismg:
@joperezr: We should keep it for now to not impact other projects @weshaggard would really like to centralize stuff, but maybe this is a file that will need to be end up being per repo in the limit.

ellismg:
@dotnet-bot test this please (machine was in a bad state due to previous tests holding stuff open).

dir.props
```diff
@@ -43,6 +43,12 @@
     <ToolsDir Condition="'$(ToolsDir)'==''">$(PackagesDir)Microsoft.DotNet.BuildTools.$(BuildToolsVersion)/lib/</ToolsDir>
   </PropertyGroup>
 
+  <!-- Test runtime -->
+  <PropertyGroup>
```
naamunds: How are these new properties used, or how are they intended to be used?

ellismg: They are overriding [properties that buildtools uses](https://github.com/dotnet/buildtools/blob/b668d329ae4950ba1500cc4a8d480188123856cb/src/Microsoft.DotNet.Build.Tasks/Targets/publishtest.targets).

naamunds:
LGTM.

mellinoe:
Do we need the lock.json file?

ellismg:
@mellinoe We need a lock file somewhere in order to figure out what assemblies to copy into the test runtime.  My preference is to check in the lock file instead of trying to generate it on the fly by restoring project.json.  This is similar to what buildtools does (it deploys both the json and .lock.json files).

Issue #1963, opened 2015-06-08
Add EVP_CIPHER_CTX_new to System.Security.Cryptography.Native


bartonjs:
Interop.EVP.Cipher.cs:
```
        // This was computed by sizeof(EVP_CIPHER_CTX) on a 64-bit Ubuntu
        // machine running OpenSSL 1.0.1f.  If we end up making a native
        // interop boundary library for OpenSSL then a very good candidate
        // method would be EVP_CIPHER_CTX_new, so the size can be computed
        // to match the platform.
        internal const int EVP_CIPHER_CTX_SIZE = 168;
```

Currently the native library is in dotnet/coreclr; so some mild gymnastics might be required.

EVP_CIPHER_CTX_new should allocate the memory and call (OpenSSL) EVP_CIPHER_CTX_init, and have a well-defined free.

If gymnastics are required for checking this in, a followup issue to use the new method should be filed.  If the native library has moved to dotnet/corefx by the time this is done, then the managed code to consume it should potentially be part of the same change.

PR #3677, opened 2015-10-06
Shim OpenSSL Evp_Cipher* methods.

eerhardt:
Adding shims for all OpenSSL methods in Interop.EVP.Cipher.cs.


src/Common/src/Microsoft/Win32/SafeHandles/SafeEvpCipherCtxHandle.Unix.cs
```diff
@@ -37,11 +37,11 @@ internal static SafeEvpCipherCtxHandle Create()
 
             try
             {
-                memPtr = Marshal.AllocHGlobal(Interop.libcrypto.EVP_CIPHER_CTX_SIZE);
+                memPtr = Marshal.AllocHGlobal(Interop.Crypto.EVP_CIPHER_CTX_SIZE);
```
bartonjs: Actually, while you're in here; could you take care of #1963, and make Create and Free methods for this type?  Then we can get rid of EVP_CIPHER_CTX_SIZE.

bartonjs: I guess you sort of did already, by making GetEvpCipherCtxSize.  I think I'm okay with that...

Should still tag the issue as fixed, though :)

nguerrera:
Rather than XxxGetSize(), AllocHGlobal(), XxxInit(), XxxFinal(), FreeHGlobal() sequence, I think there's value in a slight 1:1 departure with XxxCreate(), XxxDestroy(). That was what I found nicer for HmacXxx(). I forget if I missed that same pattern on last review, but it would be good to be consistent whenever OpenSSL has a similar pattern.

src/Common/src/Microsoft/Win32/SafeHandles/SafeEvpCipherCtxHandle.Unix.cs
nguerrera: I personally like making less chatty Create/Destroy shims like what I did for HMAC, but if you guys think 1:1 with GetSize and Alloc/FreeHGlobal is better, I won't push back, but then you should probably switch my HMAC code to match this pattern...

nguerrera:
LGTM apart from the create/destroy vs 1:1 + get size/allochglobal/freehglobal discussion. I'll leave it to @bartonjs and your judgment, which patter you prefer.

stephentoub:
@dotnet-bot test this please (https://github.com/dotnet/corefx/issues/3682)

src/Common/src/Microsoft/Win32/SafeHandles/SafeEvpCipherCtxHandle.Unix.cs
bartonjs: I think I'd prefer a Create and Free/Destroy/whatever within the PAL.  In particular, if OpenSSL got around to adding methods later we'd (maybe) want to opportunistically just shim it.

eerhardt:
@nguerrera, @bartonjs - the EVP_MD_CTX* uses a very slightly different pattern.

Example:
```C++
EVP_MD_CTX *mdctx = EVP_MD_CTX_create();
EVP_DigestInit_ex(mdctx, md, NULL);
EVP_DigestUpdate(mdctx, mess1, strlen(mess1));
EVP_DigestUpdate(mdctx, mess2, strlen(mess2));
EVP_DigestFinal_ex(mdctx, md_value, &md_len);
EVP_MD_CTX_destroy(mdctx);
```

Here, they have Create() & Destroy() methods, and you need to call the Init_ex method after Create().  So I could combine Create() and Init_ex() into a single P/Invoke.  Do you think that is beneficial?  It would be one less P/Invoke.  I could then turn our Init_ex() method into a "Reset()", similar to HMAC and what I did for Interop.EVP.Cipher.

If you want, I can add that change to this PR.  Let me know.

bartonjs:
LGTM :)

stephentoub:
LGTM2

PR #3678, opened 2015-10-06
Implement EnsureStoreOpened for the Unix PAL

bartonjs:
Now that X509Certificate2Collection.Find has been implemented on Unix, it makes sense to wire this up.

There isn't yet a notion of LM\My on Unix, and rather than eating the exception over and over this code just doesn't bother trying.  It does, however, set itself up to do so in the future, by making the code that does the store opening take in a parameter for the store name (and a ref parameter for the object reference field).

This is a part of uber-issue #3362 (EnsureStoreOpened - Implement the method. (CertModule.cs))

### DIFF
--- a/src/System.Net.Security/src/System/Net/Unix/CertModule.cs
+++ b/src/System.Net.Security/src/System/Net/Unix/CertModule.cs
@@ -4,13 +4,17 @@
 using System.Diagnostics;
 using Microsoft.Win32.SafeHandles;
 using System.Net.Security;
-using System.Runtime.InteropServices;
+using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
+using System.Threading;
 
 namespace System.Net
 {   
     internal partial class CertModule : CertInterface
     {
+        private static readonly object s_lockObject = new object();
+        private static X509Store s_userCertStore;
+
         #region internal Methods
         internal override SslPolicyErrors VerifyCertificateProperties(
             X509Chain chain,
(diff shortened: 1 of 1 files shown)


bartonjs:
cc: @stephentoub @vijaykota @rajansingh10 @shrutigarg

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -140,8 +143,46 @@ internal override string[] GetRequestCertificateAuthorities(SafeDeleteContext se
...
+
+                            storeField = store;
+
+                            GlobalLog.Print(
+                                "CertModule::EnsureStoreOpened() storeLocation:" + storeLocation +
+                                    " returned store:" + store.GetHashCode().ToString("x"));
+                        }
+                        catch (CryptographicException e)
+                        {
+                            GlobalLog.Assert(
```
nguerrera: Is this disabled in release, doesn't seem safe to ignore.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -140,8 +143,46 @@ internal override string[] GetRequestCertificateAuthorities(SafeDeleteContext se
 
         internal override X509Store EnsureStoreOpened(bool isMachineStore)
         {
-            // TODO (Issue #3362) do the implementation
-            return null;
+            if (isMachineStore)
+            {
+                // There's not currently a LocalMachine\My store on Unix, so don't bother trying
+                // and having to deal with the exception.
```
stephentoub: Is there an issue# we should include in a TODO here?

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -140,8 +143,46 @@ internal override string[] GetRequestCertificateAuthorities(SafeDeleteContext se
...
+                lock (s_lockObject)
+                {
+                    if (storeField == null)
+                    {
+                        try
+                        {
+                            X509Store store = new X509Store(StoreName.My, storeLocation);
+                            store.Open(OpenFlags.ReadOnly);
+
+                            storeField = store;
```
stephentoub: Technically I believe storeField should be volatile for this double-checked locking to be correct.

stephentoub:
Other than Nick's and my comments, LGTM.

vijaykota:
Thanks, @bartonjs. LGTM

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
bartonjs: I just re-checked the Windows version, thinking I'd gone for parity, but theirs does end in `throw;`; so I'll need to beef that up.

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
bartonjs: Well, to keep the handy ref-based helper apparently I'd need to throw a MemoryBarrier() call in right before the field assignment.  If we didn't bring MemoryBarrier into CoreFX; I guess I'll have to un-helper and slap on the volatile keyword.

stephentoub: Why not Volatile.Read and Volatile.Write?

bartonjs: Went with Volatile.Read and Volatile.Write, hopefully it looks more hardened against IA64 levels of weak memory model now :).

stephentoub:
LGTM

Issue #3629, opened 2015-10-02
Fix the System.Net.Sockets version on SqlClient


saurabh500:
We have changed the version of the System.Data.SqlClient to use System.Net.Sockets 4.0.10-beta-* from 4.0.10.0 
This was happening due to a runtime error while resolving 4.0.10.0
This needs to be investigated.

https://github.com/dotnet/corefx/pull/3567

saurabh500:
The version has been upgraded to 4.1.0-beta-* to take care of the xplat dependencies

PR #3681, opened 2015-10-06
Upgrading the System.Net.Sockets dependency

saurabh500:
The dependency of System.Net.Sockets should be upgraded to 4.1.0-beta-* to utilize the xplat capabilities of Networking.
https://github.com/dotnet/corefx/issues/3629

### DIFF
--- a/src/System.Data.SqlClient/src/project.json
+++ b/src/System.Data.SqlClient/src/project.json
@@ -13,7 +13,7 @@
     "System.IO.FileSystem": "4.0.0",
     "System.Net.Primitives": "4.0.10",
     "System.Net.Security": "4.0.0-beta-*",
-    "System.Net.Sockets": "4.0.10-beta-*",
+    "System.Net.Sockets": "4.1.0-beta-*",
     "System.Reflection": "4.0.0",
     "System.Reflection.TypeExtensions": "4.0.0",
     "System.Resources.ResourceManager": "4.0.0",
--- a/src/System.Data.SqlClient/src/project.lock.json
+++ b/src/System.Data.SqlClient/src/project.lock.json
@@ -70,7 +70,7 @@
           "lib/dotnet/System.ComponentModel.EventBasedAsync.dll": {}
         }
       },
-      "System.Console/4.0.0-beta-23401": {
+      "System.Console/4.0.0-beta-23406": {
         "type": "package",
         "dependencies": {
           "System.IO": "4.0.0",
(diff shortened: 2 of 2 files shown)


stephentoub:
cc: @pgavlin, @cipop, @davidsh, @ericstj

src/System.Data.SqlClient/src/project.json
```diff
@@ -13,7 +13,7 @@
     "System.IO.FileSystem": "4.0.0",
     "System.Net.Primitives": "4.0.10",
     "System.Net.Security": "4.0.0-beta-*",
-    "System.Net.Sockets": "4.0.10-beta-*",
+    "System.Net.Sockets": "4.1.0-beta-*",
```
CIPop: We are making big changes to this contract. Please note that the shape will change shortly:
- Removing Windows-only APIs (DuplicateAndClose, etc)
- Removing APM APIs and adding TPL equivalents
- Removing some of the constructors that make blocking I/O calls.

saurabh500: @CIPop What APIs are changing exactly? 
We are using Socket, NetworkStream, TcpClient from the Sytem.Net.Sockets

They are referenced in 
System\Data\SqlClient\SNI\SNITcpHandle.cs

Are there going to be any changes there?

CIPop: Yes, Begin/End (APM) will be replaced by Task Async in the final contract. We may end up providing the legacy APM apis in a different contract or an adapter that converts from TPL to APM.
/cc: @SidharthNabar @davidsh @josguil

CIPop: Proposed changes are discussed in #3655.

saurabh500: What's your plan to manage the change distribution? Create a new version for Net.Sockets? 
What does it mean for us as the consumers? 
We need to have Data.SqlClient run on the Networking for XPlat for which the version is 4.1.0-beta-*

CIPop: Our plan is for consumers to have minimal to no changes short term. Longer term we encourage all I/O consumers to switch to either TPL or higher performance APIs (such as `SocketAsyncEventArgs` in the Sockets case).

Short term options:
* Option 1: Take dependency on the Common\src\System\Threading\TasksTaskToAPM.cs file and keep the current implementation.
* Option 2: We provide a legacy package containing the Begin/End APIs as extension methods. No changes are required to your code except taking a new dependency to the new contract.

saurabh500: When are these changes targeted to be released? 
Will there be sufficient time for us to react and make changes in the RC1 timeframe?

saurabh500: I will merge this request for SqlClient and we can sync up on the next wave of changes in System.Net.Sockets with higher performance APIs

Issue #3559, opened 2015-09-30
Expose System.Console.CursorVisible


pallavit:
Azure CLI would like us to expose this property

Issue #3558, opened 2015-09-30
Expose System.Console.WindowWidth


pallavit:
Azure CLI would like us to expose WindowWidth property

PR #3683, opened 2015-10-07
Expose Console.WindowWidth and Console.CursorVisible

pallavit:
Based on request #3558 and #3559 exposing Console.WindowWidth and Console.CursorVisible.

On windows this is a simple port.
On Unix, there is no good way to set WindowWidth. tput supports cols capability but it is non-POSIX and so I have chosen to throw PlatformNotSupportedException instead. Similarly I could not find any capability that supports getting the cursor visibility. However, there are capabilities for making it visible (cursor_normal) and invisible(cursor_invisible) and so get_CursorVisible throws PlatformNotSupportedException while set_CursorVisible works as expected. 
Also both these APIs are not meant to work in case the process is redirected. Unfortunately windows does not explicitly test this scenario and so there is no consistent exception type thrown in this scenario (also msdn does state that these properties should be used along with IsOutputRedirected). For the fear of different exception type on Unix and Windows, for now, I have explicitly thrown exception on Unix either, however we can consider throwing IOException().


pallavit:
@stephentoub, @weshaggard, could you please take a look?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -73,6 +73,36 @@ public static void ResetColor()
             }
         }
 
+        public static int WindowWidth
+        {
+            get
+            {
+                if (!ConsoleOutIsTerminal) return 0; // Should we rather throw exception here? On windows we get IOException from the native code, do we need to match for code portability?
```
stephentoub: Throwing seems reasonable.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -73,6 +73,36 @@ public static void ResetColor()
...
+                return TerminalBasicInfo.Instance.ColumnFormat;
```
stephentoub: I don't think this works.  The terminfo database is static; it doesn't change based on the current state of the windows.  This is going to just return the default value of 80.

pallavit: I checked tput cols changes on resizing and was under the impression that TermInfo db and tput give the same result but seems like I was wrong. Hmm.. the only other way I know this can be done is using Ioctl and tput cols and none is POSIX compliant.

stephentoub: We strive to use things that are POSIX compliant, but in this case there isn't, so we do the best we can.  I'd suggest doing adding something to the native shim that uses ioctl with TIOCGWINSZ  if it's available, or else returns an error indicating it's not supported, and we can either throw a PlatformNotSupported or return a sentiinel value from WindowSize.  That should work for both Linux and OS X, and also FreeBSD.  The native shim's implementation can be augmented in the future with other implementations as we find it necessary.

pallavit: By native shim do you mean the System.Native? Also, I was thinking we could try to read the environment variable COLUMNS, which does get updated on resizing the terminal. I have not yet tested it by using Environment.GetEnvironmentVariable though.

stephentoub: > By native shim do you mean the System.Native?

Yes

> I have not yet tested it by using Environment.GetEnvironmentVariable though

I don't think that'll work:
https://github.com/dotnet/coreclr/issues/635

> Also, I was thinking we could try to read the environment variable COLUMNS, which does get updated on resizing the terminal

That could be an ok fallback to try in the case of ioctl/TIOCGWINSZ not being available or not working.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -279,6 +309,33 @@ private TerminalColorInfo(TermInfo.Database db)
...
+            public static TerminalBasicInfo Instance { get { return _instance.Value; } }
+
+            private TerminalBasicInfo(TermInfo.Database db)
+            {
+                ColumnFormat = db != null ? db.GetNumber(TermInfo.Database.ColumnIndex) : 0;
+                CursorVisibleFormat = db != null ? db.GetString(TermInfo.Database.CursorVisibleIndex) : string.Empty;
+                CursorInvisibleFormat = db != null ? db.GetString(TermInfo.Database.CursorInvisibleIndex) : string.Empty;
+            }
+
+            /// <summary>Lazy initialization of the terminal color information.</summary>
```
stephentoub: Copy-paste error.

src/System.Console/src/System/ConsolePal.Windows.cs
```diff
@@ -218,6 +329,12 @@ private static ConsoleColor ColorAttributeToConsoleColor(Interop.mincore.Color c
             return (ConsoleColor)c;
         }
 
+        private static Interop.mincore.CONSOLE_SCREEN_BUFFER_INFO GetBufferInfo()
+        {
+            bool unused;
+            return GetBufferInfo(true, out unused);
+        }
+
```
stephentoub: Are there any tests we can write?  If nothing else the tests could at least execute the code paths to ensure they don't throw when they're not supposed to.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: On windows, if the handle is redirected the native API that we use for this would INVALID_HANDLE. However, we try to use all the 3 handles, input, output and error to get the information. On Unix, we might not have to do this actually. I am yet to test the o/p of Ioctl but based on its o/p I will change this.

src/System.Console/src/System/ConsolePal.Windows.cs
pallavit: I have manually tested the scenarios. However, since most of the scenarios don't work in case of redirection, the value add is minimum.

stephentoub: > I have manually tested the scenarios

After fixing WindowWidth to work? :)

>  However, since most of the scenarios don't work in case of redirection, the value add is minimum.

How about on Unix creating a child process that opens a chr device (e.g. /dev/null) and dups it onto fd 1; then we should detect that console out is a terminal and proceed as if we're not redirected.

Even if we don't do that, though, we should still test what we can, that we throw PlatformNotSupported in the appropriate cases, that we don't throw exceptions in other cases, etc.

src/Native/System.Native/pal_io.cpp
```diff
@@ -793,3 +794,17 @@ extern "C" int32_t Write(int32_t fd, const void* buffer, int32_t bufferSize)
     assert(count >= -1 && count <= bufferSize);
     return static_cast<int32_t>(count);
 }
+
+extern "C" int32_t GetWindowSize(WinSize* windowSize)
+{
+    assert(windowSize != nullptr);
+
+    int error = ioctl(1/*STDOUT_FILENO*/, TIOCGWINSZ, windowSize);
```
stephentoub: STDOUT_FILENO should be defined in unistd.h; it's not working?

stephentoub: Do we need to do cmake feature detection for ioctl and TIOCGWINSZ and then make this conditional on their availability, otherwise setting errno to ENOTSUP and returning -1?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -73,6 +73,36 @@ public static void ResetColor()
             }
         }
 
+        public static int WindowWidth
+        {
+            get
+            {
+                int cols = Interop.Sys.GetWindowWidth();
+                return (cols != -1) ? cols : TerminalBasicInfo.Instance.ColumnFormat;
```
stephentoub: Nice; I think that makes sense as the fallback value.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -73,6 +73,36 @@ public static void ResetColor()
...
+        public static bool CursorVisible
+        {
+            get
+            {
+                throw new PlatformNotSupportedException();
+            }
+            set
+            {
+                if (!ConsoleOutIsTerminal) return;
+                string formatString = value ? TerminalBasicInfo.Instance.CursorVisibleFormat : TerminalBasicInfo.Instance.CursorInvisibleFormat;
```
stephentoub: Nit: would you mind formatting this slightly differently, as:
```C#
string formatString = value ?
    TerminalBasicInfo.Instance.CursorVisibleFormat :
    TerminalBasicInfo.Instance.CursorInvisibleFormat;
```
I think that just makes it a bit easier to understand.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -73,6 +73,36 @@ public static void ResetColor()
...
+                if (!String.IsNullOrEmpty(formatString))
```
stephentoub: Nit: String => string

src/System.Console/ref/System.Console.cs
```diff
@@ -24,6 +25,7 @@ public static partial class Console
         public static void SetError(System.IO.TextWriter newError) { }
         public static void SetIn(System.IO.TextReader newIn) { }
         public static void SetOut(System.IO.TextWriter newOut) { }
+        public static bool WindowWidth { get { return default(bool); } set { } }
```
stephentoub: bool isn't the right return type

src/Native/System.Native/pal_io.h
```diff
@@ -610,3 +621,11 @@ extern "C" void Sync();
...
  */
 extern "C" int32_t Write(int32_t fd, const void* buffer, int32_t bufferSize);
+
+
+/**
+* Gets the windows size of the terminal
+*
+* Returns 0 on success; otherwise, returns errorNo.
+*/
+extern "C" int32_t GetWindowSize(WinSize* windowsSize);
```
stephentoub: Missing newline at end of file; I believe we'll fail to compile without it.

src/Native/System.Native/pal_io.cpp
```diff
@@ -793,3 +794,17 @@ extern "C" int32_t Write(int32_t fd, const void* buffer, int32_t bufferSize)
...
+
+    if (error != 0)
+    {
+        *windowSize = {}; // managed out param must be initialized
+    }
+
+    return error;
+}
```
stephentoub: Missing newline at end of file; I believe we'll fail to compile without it.

src/System.Console/ref/System.Console.cs
weshaggard: Looks like we need to get our ApiCompat tooling working to catch these type of errors.

src/System.Console/src/System/ConsolePal.Windows.cs
stephentoub: > After fixing WindowWidth to work?

Just tried it... it works :smile: Nice.

pallavit: > How about on Unix creating a child process that opens a chr device (e.g. /dev/null) and dups it onto fd 1; then we 
should detect that console out is a terminal and proceed as if we're not redirected.

WIth <3 days experience on Unix, I am not sure I fully grasp this :). I will take a closer look. Also, I was looking at some of the Console tests especially for TermInfo where we have removed the test from CI system but still added some tests which we can run manually. We can add some like that :)

> Even if we don't do that, though, we should still test what we can, that we throw PlatformNotSupported in the appropriate cases, that we don't throw exceptions in other cases, etc
Yes, I will add these scenarios..

src/System.Console/ref/System.Console.cs
pallavit: oops! Thanks for catching :).

src/Native/System.Native/pal_io.h
pallavit: Really?.. I did not get compilation error. But, it could be because I copy-pasted my changes from dev box to a unix box and it might have taken care of this automatically. Changing it now.

src/Native/System.Native/pal_io.cpp
pallavit: > STDOUT_FILENO should be defined in unistd.h; it's not working?

I just checked it works :). Taken care.

> Do we need to do cmake feature detection for ioctl and TIOCGWINSZ and then make this conditional on their availability, otherwise setting errno to ENOTSUP and returning -1?

I am not sure how to do that. Could you please point me in the right direction?

stephentoub:
cc: @nguerrera for the shim addition

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: Sure :)

src/System.Console/src/System/ConsolePal.Windows.cs
pallavit: I have added some tests for Unix. I will add for Windows once IsXXXRedirected is checked in. It will make writing tests easier :)

src/System.Console/tests/WindowAndCursorProps.cs
```diff
@@ -0,0 +1,109 @@
...
+public class WindowAndCursorProps
+{
+    [Fact]
+    [PlatformSpecific(PlatformID.AnyUnix)]
+    public static void WindowWidth()
+    {
+        Assert.Throws<PlatformNotSupportedException>(() => Console.WindowWidth = 100);
+
+        // Validate that Console.WindowWidth returns some value in a non-redirected o/p.
+        RunInNonRedirectedOutput((data) => Console.WriteLine(Console.WindowWidth));
```
stephentoub: I'm a little confused about this RunInNonRedirectedOutput due to the comment below citing that the CI system makes it difficult.  Is it just that this usage of it isn't actually problematic but it's also not really testing what it's designed to test?

src/System.Console/tests/WindowAndCursorProps.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+        finally
+        {
+            Console.SetOut(savedOut);
+        }
+
+        return data;
+    }
+
+    private sealed class InterceptStreamWriter : StreamWriter
```
stephentoub: We have something like this elsewhere in the Console tests, right?  If so, could we consolidate to a single definition?  This could even move to Common\tests\ if we think it might be useful elsewhere at some point, too.

src/System.Console/src/System/ConsolePal.Windows.cs
stephentoub: Thanks :)

stephentoub:
@dotnet-bot test this please (looks like CI machine ran out of disk space)

src/Native/System.Native/pal_io.cpp
stephentoub: > I am not sure how to do that

See https://github.com/dotnet/corefx/blob/master/src/Native/configure.cmake for a variety of examples.

src/System.Console/tests/WindowAndCursorProps.cs
pallavit: CI system will always run these tests with a redirected o/p. And so RunInNonRedirectedOutput would still be running in redirected mode unless we manually test it by explicitly running the corerun command.

src/System.Console/tests/WindowAndCursorProps.cs
pallavit: Yes, I will move this code to common :).

src/Native/configure.cmake
```diff
@@ -31,6 +32,16 @@ check_function_exists(
...
     HAVE_POSIX_ADVISE)
+	
+check_function_exists(
+    ioctl
+    HAVE_IOCTL)
+	
+ check_symbol_exists (
+    TIOCGWINSZ
+	"sys/ioctl.h"
+	HAVE_TIOCGWINSZ
```
stephentoub: Nit: I think these two lines have tabs instead of spaces

src/System.Console/tests/Common/System/IO/InterceptStreamWriter.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: How about putting it in src/Common/tests instead of in src/System.Console/tests/Common?

stephentoub:
A few more minor comments, but LGTM

pallavit:
Thanks for the review. Merging the changes :)

PR #3685, opened 2015-10-07
Mark corerun as executable

ellismg:
Make sure corerun is marked as executable (since the version we pull down might not have the right bits set).  This is fall out from my recent refactoring of how we build the test layouts.


stephentoub:
LGTM

ellismg:
Rolled into #3684, so no need to merge this one.

Issue #3297, opened 2015-09-17
Improve optimization strategy for BlockExpression


bartdesmet:
The check for optimization to N-ary specialized nodes in the Block factories only happens for the Expression[] overload. With the addition of IReadOnlyList&lt;T&gt; to the framework, we can also optimize the IEnumerable&lt;Expression&gt; overloads, similar to the changes proposed in #3288.

VSadov:
related - https://github.com/dotnet/coreclr/issues/1592

JonHanna:
Related: #3046

PR #3686, opened 2015-10-07
Use optimized N-ary BlockExpression nodes from all Block factories

bartdesmet:
This addresses issue #3297 by using Block2-Block5 types when applicable, regardless of the factory method used to create the BlockExpression.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
@@ -848,10 +848,14 @@ namespace System.Linq.Expressions
         {
             ContractUtils.RequiresNotNull(expressions, "expressions");
             var expressionList = expressions.ToReadOnly();
-            ContractUtils.RequiresNotEmpty(expressionList, "expressions");
-            RequiresCanRead(expressionList, "expressions");
+            var variableList = variables.ToReadOnly();
 
-            return Block(expressionList.Last().Type, variables, expressionList);
+            if (variableList.Count == 0)
+            {
+                return GetOptimizedBlockExpression(expressionList);
+            }
+
+            return BlockCore(expressionList.Last().Type, variableList, expressionList);
         }
 
         /// <summary>
@@ -869,6 +873,30 @@ namespace System.Linq.Expressions
             var expressionList = expressions.ToReadOnly();
             var variableList = variables.ToReadOnly();
 
+            if (variableList.Count == 0)
+            {
+                var expressionsList = expressions as IReadOnlyList<Expression>;
+                if (expressionsList != null)
+                {
+                    var expressionCount = expressionsList.Count;
+
+                    if (expressionCount > 0)
+                    {
+                        var lastExpression = expressionsList[expressionCount - 1];
+
+                        if (lastExpression.Type == type)
+                        {
+                            return GetOptimizedBlockExpression(expressionsList);
+                        }
+                    }
+                }
+            }
+
+            return BlockCore(type, variableList, expressionList);
+        }
+
+        private static BlockExpression BlockCore(Type type, ReadOnlyCollection<ParameterExpression> variableList, ReadOnlyCollection<Expression> expressionList)
+        {
             ContractUtils.RequiresNotEmpty(expressionList, "expressions");
             RequiresCanRead(expressionList, "expressions");
             ValidateVariables(variableList, "variables");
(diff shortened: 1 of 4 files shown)


bartdesmet:
Not sure what I can do to get commits 7792073 and 2b7903f out of sight in the PR; those are the result of pulling in the code that changed in the meantime in dotnet/corefx.

Maxwe11:
>Not sure what I can do to get commits 7792073 and 2b7903f out of sight in the PR; those are the result of pulling in the code that changed in the meantime in dotnet/corefx.

git rebase

stephentoub:
cc: @VSadov

src/System.Linq.Expressions/src/System/Dynamic/Utils/CollectionExtensions.cs
```diff
@@ -11,5 +14,25 @@ public static T[] Copy<T>(this T[] array)
...
+                return Copy(array);
+            }
+
+            ICollection<T> collection = enumerable as ICollection<T>;
+            if (collection != null)
+            {
+                array = new T[collection.Count];
+                collection.CopyTo(array, 0);
+                return array;
+            }
```
VSadov: I think enumerable.ToArray now does all the checks above and more.

VSadov:
LGTM, but I think CopyToArray is no longer necessary. Just use .ToArray

VSadov:
You can get rid of merging commits by rebasing these commits on top of the recent master while Skipping merging commits.

bartdesmet:
Grmbl I've made it worse again :-(. 5 minutes coding, 50 minutes git.

bartdesmet:
Changed to using ToArray.

VSadov:
something failed in C++ on Linux. Could not be possible because of this change.

VSadov:
seems like some file sharing conflict during build

VSadov:
test this please

bartdesmet:
@dotnet-bot test this please

src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
```diff
@@ -937,5 +955,20 @@ internal static void ValidateVariables(ReadOnlyCollection<ParameterExpression> v
...
+            switch (expressions.Count)
+            {
+                case 2: return Block(expressions[0], expressions[1]);
+                case 3: return Block(expressions[0], expressions[1], expressions[2]);
+                case 4: return Block(expressions[0], expressions[1], expressions[2], expressions[3]);
+                case 5: return Block(expressions[0], expressions[1], expressions[2], expressions[3], expressions[4]);
+                default:
+                    ContractUtils.RequiresNotEmptyList(expressions, "expressions");
+                    RequiresCanRead(expressions, "expressions");
+                    return new BlockN(expressions.ToArray());
```
JonHanna: Could `ToReadOnly()` serve better here? It likewise promises that changes to the source enumerable won't change the copy held in the `BlockN`, but won't create a new collection if `Expressions` is accessed (very likely) at the point of its access, and won't create a new collection or if the source enumerable was already a `ReadOnlyCollection`, as happens with the default `Visit` behaviour.

JonHanna:
I'm closing #3046 since this does the same thing, but if any of the tests it added seem still worthwhile they could be added to this.

src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
JonHanna: Indeed, the changed factory also goes through `ToReadOnly()`, so that case would definitely benefit from `ToReadOnly()` being used here. As is, this PR could make some uses heavier than they were before.

VSadov: Agreed. I will make the change.

VSadov: Actually, no, it makes sense to create node with an expression array. 
The array will be automatically wrapped into ReadOnly, but only if GetOrMakeExpressions is called, which is not considered common. Otherwise expressions will be stored as array, which is leaner

JonHanna: It will be hit if `Expressions` is accessed, which will definitely happen if it gets rewritten or visited, which is one of the reasons why the method affected by this change might be hit rather than the `params` method. Since `ToReadOnly` returns the argument if its already a `ReadOnlyCollection` and uses a single instance for all empty collections of the same element type, the consistent use of it throughout means there's a high chance of instances being shared, which is hence lighter than using an array.

A middle ground would be to use:

```C#
return new BlockN(expressions as ReadOnlyCollection<Expression> ?? (IList<Expression>)expressions.ToArray());
```

As then the method that takes an `IEnumerable<Expression>` and which is hit in the instance-sharing cases will still have the previous behaviour of using the `ReadOnlyCollection` it passes and, the method with the single `params` methods will keep its previous behaviour of passing a new copy of the array, which as you say is lighter if `Expressions` is never hit.

PR #3696, opened 2015-10-07
Added exception messages to *.resx file. using Debug.Fail wherever ap…

rajansingh10:
1. Added exception messages to *.resx. Using the SR now.
2. using Debug.Fail/Assert wherever deemed fit.
3. Fixed GetException() method.

### DIFF
--- a/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
+++ b/src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
@@ -27,7 +27,7 @@ internal static partial class Interop
             {
                 if (innerContext.IsInvalid)
                 {
-                    throw CreateSslException("Failed to allocate SSL/TLS context");
+                    throw CreateSslException(SR.net_allocate_ssl_context_failed);
                 }
 
                 libssl.SSL_CTX_ctrl(innerContext, libssl.SSL_CTRL_OPTIONS, options, IntPtr.Zero);
@@ -44,7 +44,7 @@ internal static partial class Interop
                 if (context.IsInvalid)
                 {
                     context.Dispose();
-                    throw CreateSslException("Failed to create SSL object from SSL context");
+                    throw CreateSslException(SR.net_allocate_ssl_context_failed);
                 }
             }
 
@@ -62,21 +62,14 @@ internal static partial class Interop
 
             libssl.SslErrorCode error;
             int retVal = libssl.SSL_do_handshake(context);
-            if (retVal == 1)
-            {
-                // In case of a client, this indicates successful handshake completion
-                if (!context.IsServer)
-                {
-                    return true;
-                }
-            }
-            else
+
+            if (retVal != 1)
             {
                 error = GetSslError(context, retVal);
 
                 if ((retVal != -1) || (error != libssl.SslErrorCode.SSL_ERROR_WANT_READ))
                 {
-                    throw CreateSslException(context, "SSL Handshake failed: ", retVal);
+                    throw CreateSslException(context, SR.net_ssl_handshake_failed_error, retVal);
                 }
             }
 
@@ -85,14 +78,19 @@ internal static partial class Interop
             if (sendCount > 0)
             {
                 sendPtr = Marshal.AllocHGlobal(sendCount);
-                sendCount = BioRead(context.OutputBio, sendPtr, sendCount);
-                if (sendCount <= 0)
+
+                try
                 {
-                    int errorCode = sendCount;
-                    Marshal.FreeHGlobal(sendPtr);
-                    sendPtr = IntPtr.Zero;
-                    sendCount = 0;                  
-                    throw CreateSslException(context, "Read Bio failed: ", errorCode);                   
+                    sendCount = BioRead(context.OutputBio, sendPtr, sendCount);
+                }
+                finally
+                {
+                    if (sendCount <= 0)
+                    {
+                        Marshal.FreeHGlobal(sendPtr);
+                        sendPtr = IntPtr.Zero;
+                        sendCount = 0;
+                    }
                 }
             }
         
@@ -118,24 +116,17 @@ internal static partial class Interop
                         break;
 
                     default:
-                        throw CreateSslException("OpenSsl::Encrypt failed");
+                        throw CreateSslException(SR.net_ssl_encrypt_failed, errorCode);
                 }
             }
             else
             {
-                int capacityNeeded = libssl.BIO_ctrl_pending(context.OutputBio);      
+                int capacityNeeded = libssl.BIO_ctrl_pending(context.OutputBio);
 
-                if (capacityNeeded > bufferCapacity)
-                {
-                    throw CreateSslException("OpenSsl::Encrypt capacity needed is more than buffer capacity. capacityNeeded = " + capacityNeeded + "," + "bufferCapacity = " + bufferCapacity);
-                }             
+                Debug.Assert(bufferCapacity >= capacityNeeded, "Input buffer of size " + bufferCapacity +
+                                                              " bytes is insufficient since encryption needs " + capacityNeeded + " bytes.");
 
                 retVal = BioRead(context.OutputBio, buffer, capacityNeeded);
-
-                if (retVal < 0)
-                {
-                    throw CreateSslException("OpenSsl::Encrypt failed");
-                }
             }
 
             return retVal;
@@ -176,7 +167,7 @@ internal static partial class Interop
                         break;
 
                     default:
-                        throw CreateSslException("OpenSsl::Decrypt failed");
+                        throw CreateSslException(SR.net_ssl_decrypt_failed, errorCode);
                 }
             }
 
@@ -253,7 +244,7 @@ internal static partial class Interop
 
             if (IntPtr.Zero == method)
             {
-                throw CreateSslException("Failed to get SSL method");
+                throw CreateSslException(SR.net_get_ssl_method_failed);
             }
 
             return method;
@@ -275,7 +266,7 @@ internal static partial class Interop
             int bytes = libssl.BIO_read(bio, buffer, count);
             if (bytes != count)
             {
-                throw CreateSslException("Failed in Read BIO");
+                throw CreateSslException(SR.net_ssl_read_bio_failed_error);
             }
             return bytes;
         }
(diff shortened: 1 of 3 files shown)


rajansingh10:
cc: @stephentoub, @bartonjs, @davidsh, @CIPop, @SidharthNabar, @pgavlin , @vijaykota

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -76,7 +76,7 @@ internal static bool DoSslHandshake(SafeSslHandle context, IntPtr recvPtr, int r
 
                 if ((retVal != -1) || (error != libssl.SslErrorCode.SSL_ERROR_WANT_READ))
                 {
-                    throw CreateSslException(context, "SSL Handshake failed: ", retVal);
+                    Debug.Fail("SSL Handshake failed with error:" + error);
```
stephentoub: What happens here for release?  We still want to proceed to do all of the stuff below?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -91,8 +91,8 @@ internal static bool DoSslHandshake(SafeSslHandle context, IntPtr recvPtr, int r
                     int errorCode = sendCount;
                     Marshal.FreeHGlobal(sendPtr);
                     sendPtr = IntPtr.Zero;
-                    sendCount = 0;                  
-                    throw CreateSslException(context, "Read Bio failed: ", errorCode);                   
+                    sendCount = 0;
+                    Debug.Fail("SSL Read BIO failed with error:" + errorCode);
```
stephentoub: Same question as above... we want to continue running even if sendCount <= 0?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -118,23 +118,21 @@ internal static int Encrypt(SafeSslHandle context, IntPtr buffer, int offset, in
...
             {
-                int capacityNeeded = libssl.BIO_ctrl_pending(context.OutputBio);      
+                int capacityNeeded = libssl.BIO_ctrl_pending(context.OutputBio);
 
-                if (capacityNeeded > bufferCapacity)
-                {
-                    throw CreateSslException("OpenSsl::Encrypt capacity needed is more than buffer capacity. capacityNeeded = " + capacityNeeded + "," + "bufferCapacity = " + bufferCapacity);
-                }             
+                Debug.Assert(bufferCapacity >= capacityNeeded, "Input buffer of size " + bufferCapacity +
+                                                              " bytes is insufficient since encryption needs " + capacityNeeded + " bytes.");
```
stephentoub: Asserts should only be used if the situation can never occur in practice.  Does BIO_ctrl_pending always return an appropriate value here?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -251,10 +249,7 @@ private static IntPtr GetSslMethod(bool isServer, long options)
                 }
             }
 
-            if (IntPtr.Zero == method)
-            {
-                throw CreateSslException("Failed to get SSL method");
-            }
+            Debug.Assert(IntPtr.Zero != method, "Failed to get SSL method");
```
stephentoub: Seems like this could actually be zero if the static cctor failed to get any of these.  Shouldn't we still throw in that case?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -275,7 +270,7 @@ private static int BioRead(SafeBioHandle bio, IntPtr buffer, int count)
             int bytes = libssl.BIO_read(bio, buffer, count);
             if (bytes != count)
             {
-                throw CreateSslException("Failed in Read BIO");
+                Debug.Fail("BIO read failed.");
```
stephentoub: BIO_read and BIO_write are documented to be able to fail.  Shouldn't we throw in such situations?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -307,45 +302,40 @@ private static void SetSslCertificate(libssl.SafeSslContextHandle contextPtr, Sa
...
+            Debug.Assert(1 == retVal, "Failed to use SSL certificate private key.");
 
             //check private key
             retVal = libssl.SSL_CTX_check_private_key(contextPtr);
-            if (1 != retVal)
-            {
-                throw CreateSslException("Certificate pivate key check failed");
-            }
+
+            Debug.Assert(1 == retVal, "SSL certificate pivate key check failed.");
```
stephentoub: Same question about why these have been changed from exceptions to asserts.  I'll stop asking about such things now.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: In remote scenario if static cctor fails, this method will return null. As a result untimately innerContext will become invalid in allocateSSLContext & will throw the exception there.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: if ssl_write call before is a success, BIO_ctrl_pending suppose to always return proper value. OpenSSL documentation does not mention it failing or possible return error values like it does for other APIs. Checked the implementation as well. It will return -2 only if the operation is NOT supported.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: Will make it an Exception.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: Actually it should throw as we won't be throwing inside of BioRead & Write. Will update.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: user of these methods does the relevant check & throws the exception. So exception not needed here.

stephentoub: If the check here really isn't necessary, then these should be Debug.Asserts, not Debug.Fails.

stephentoub: Also, if all of the callers do the relevant checks, then what's the purpose of these wrappers anymore?

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: throwing exception now.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: modified to throw exception in BioRead & BioWrite. Removed calling code exception.
Debug.Assert was my first choice as well, but left Debug.Fail because the condition check might come back due to a bio_retry related //todo which I am going to work next. So thought of leaving the same as it is.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: removed it as now BioRead is throwing the exception.

rajansingh10:
@stephentoub thanks for the review. I have addressed the comments.

rajansingh10:
Marked "Needs extra testing before merge" as after recent change, yet to validate few unit scenarios.

src/System.Net.Security/src/Resources/Strings.resx
```diff
@@ -330,4 +330,37 @@
...
+  </data>
+  <data name="net_ssl_use_cert_failed" xml:space="preserve">
+    <value>Using SSL certificate failed with OpenSSL error - {0}.</value>
+  </data>
+  <data name="net_ssl_use_private_key_failed" xml:space="preserve">
+    <value>Using SSL certificate private key failed with OpenSSL error - {0}.</value>
+  </data>
+  <data name="net_ssl_write_bio_failed_error" xml:space="preserve">
+    <value>SSL Write BIO failed with OpenSSL error - {0}.</value>
+  </data>
```
stephentoub: I'm curious, do we know what the OpenSSL error messages look like in many of these cases?  I'm wondering if we could just have a generic message like "OpenSSL operation failed: {0}" rather than having all of these different varieties.  The stack trace will also provide context as to where the error is coming from.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
```diff
@@ -307,45 +305,49 @@ private static void SetSslCertificate(libssl.SafeSslContextHandle contextPtr, Sa
...
         {
             ulong errorVal = libssl.ERR_get_error();
-            string msg = message + ": " + Marshal.PtrToStringAnsi(libssl.ERR_reason_error_string(errorVal));
+            string msg = SR.Format(message, Marshal.PtrToStringAnsi(libssl.ERR_reason_error_string(errorVal)));
             return new SslException(msg, (int)errorVal);
         }
 
         private static SslException CreateSslException(string message, libssl.SslErrorCode error)
         {
+			string msg = SR.Format(message, error);
```
stephentoub: Nit: spacing

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
stephentoub: Ok, thanks.

src/System.Net.Security/src/Resources/Strings.resx
rajansingh10: OpenSsl gives very plain error. For eg. if certfile does not exists in above APIs below is the error:-
"No such file or directory".
Does not give any more detail. So I think we should keep the detailed API specific messages.

src/Common/src/Interop/Unix/libssl/Interop.OpenSsl.cs
rajansingh10: Done.

rajansingh10:
@stephentoub I have squashed the commits as there was no pending comments.

rajansingh10:
Fixed one merge issue caused between #3638 & other PR that went after it.

Issue #1991, opened 2015-06-10
Improve memory allocations of DeflateStream


redknightlois:
We use GZipStream to handle the data sent on the wire to a WebAPI with both big and small requests. The allocation stress is more evident with small packages though. 

On a sample of 60 seconds in an stress test we measured that
System.IO.Compression.OutputWindow.ctor() and System.IO.Compression.DefrateStream.ctor() would be consistently in the top 5 allocators. Between the two 1.8Gb in allocations in 60 seconds.

The root is that DefrateStream will create on every construction (1 per request) an Infrater (https://github.com/dotnet/corefx/blob/master/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs#L62) and the Infrater will also build an OutputWindow (https://github.com/dotnet/corefx/blob/master/src/System.IO.Compression/src/System/IO/Compression/OutputWindow.cs). Which in turn will allocate 32Kb to use as a buffer for each request. 

Reusing those buffers and/or OutputWindow would aleviate a lot the allocation stress GZipStream causes when used in small streams contexts like Web APIs.

stephentoub:
Could you share some more details? e.g. how many requests were there that represented those allocations?  How many were running concurrently on average and at the limit? (And what do you expect in a normal use case vs stress?) etc.  I ask because any pooling strategy would need to be tuned appropriately.

redknightlois:
Sure, I can even send you the code to repro, or the profiler session dump if you want it. We were analyzing our own hotspots when we stumbled upon this (completely unintented finding), we usually just go and fix them (sometimes using another library or writing our own components if we need to) but wanted to report it so you can have it in your radar as the current API does not provide a way to fix it at all. 

The test is synthetic so message size and some other parameters are controlled for what we wanted to break, but it roughly represents a typical heavy write use case scenario (small documents puts, deletes and updates at high frequency). 
Code: https://gist.github.com/redknightlois/b0f38b4fac93af4fd137

Some details:

- Test time: Aprox 60 seconds (under the profiler so those may be a bit less in a real scenario)
- Mean Message Size (including HTTP headers): 300 bytes
- Number of requests: 52486 request (compress & decompress)
- Number of decompress requests: 26243 requests (aka WindowOutput.ctor calls)
- Number of client threads: 14 (initiators)
- Number of handling threads: 8 (server managed)
- Total memory allocated: 17Gb 
[...]

PR #3702, opened 2015-10-07
#1991 - ObjectPool approach

redknightlois:
Allow DefrateStream and OutputWindow to pool internal resources that causes stress allocation when DefrateStream is used in high throughput small data packages. Those commonly happen in web servers which accept gzipped incoming connections and must respond accordingly. DefrateStream will allocate for reuse up to 256Kb of memory (or up to 32 concurrent streams without allocations) and will apply both to ZLib and Managed versions. On the managed implementation (InfrateManaged), there will exist a reservation of up to 512Kb of memory (or up to 16 concurrent compression streams without allocations). The rationale is that on the web server scenario messages are paired, one incoming and one outgoing, therefore there are half as many InfrateManaged than total DefrateStreams.

Reservations happen lazily, therefore in low throughput scenarios the probability of allocating the whole 768Kb is pretty low.

Fix: #1991

### DIFF
--- a/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
@@ -5,11 +5,14 @@ using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Diagnostics.Contracts;
+using System.IO.Compression.Internal;
 
 namespace System.IO.Compression
 {
     public partial class DeflateStream : Stream
     {
+        private static readonly ObjectPool<byte[]> s_byteBufferPool = new ObjectPool<byte[]>(() => new byte[DefaultBufferSize], 32);  
+
         internal const int DefaultBufferSize = 8192;
 
         private Stream _stream;
@@ -17,7 +20,7 @@ namespace System.IO.Compression
         private bool _leaveOpen;
         private IInflater _inflater;
         private IDeflater _deflater;
-        private byte[] _buffer;
+        private readonly byte[] _buffer;
 
         private int _asyncOperations;
 
@@ -50,8 +53,8 @@ namespace System.IO.Compression
             _inflater = CreateInflater(reader);
             _stream = stream;
             _mode = CompressionMode.Decompress;
-            _leaveOpen = leaveOpen;
-            _buffer = new byte[DefaultBufferSize];
+            _leaveOpen = leaveOpen;            
+            _buffer = s_byteBufferPool.Allocate();
         }
 
 
@@ -85,7 +88,7 @@ namespace System.IO.Compression
             _stream = stream;
             _mode = mode;
             _leaveOpen = leaveOpen;
-            _buffer = new byte[DefaultBufferSize];
+            _buffer = s_byteBufferPool.Allocate();
         }
 
         // Implies mode = Compress
@@ -115,7 +118,7 @@ namespace System.IO.Compression
 
             _deflater = CreateDeflater(compressionLevel);
 
-            _buffer = new byte[DefaultBufferSize];
+            _buffer = s_byteBufferPool.Allocate();
         }
 
         private static IDeflater CreateDeflater(CompressionLevel? compressionLevel)
@@ -611,6 +614,8 @@ namespace System.IO.Compression
                         base.Dispose(disposing);
                     }
                 }  // finally
+
+                s_byteBufferPool.Free(_buffer);
             }  // finally
         }  // Dispose
 
--- a/src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
+++ b/src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
@@ -1,0 +1,115 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+
+namespace System.IO.Compression.Internal
+{
+    /// <summary>
+    /// Generic implementation of object pooling pattern with predefined pool size limit. The main
+    /// purpose is that limited number of frequently used objects can be kept in the pool for
+    /// further recycling.
+    /// 
+    /// Notes: 
+    /// 1) it is not the goal to keep all returned objects. Pool is not meant for storage. If there
+    ///    is no space in the pool, extra returned objects will be dropped.
+    /// 
+    /// 2) it is implied that if object was obtained from a pool, the caller will return it back in
+    ///    a relatively short time. Keeping checked out objects for long durations is ok, but 
+    ///    reduces usefulness of pooling. Just new up your own.
+    /// 
+    /// Not returning objects to the pool in not detrimental to the pool's work, but is a bad practice. 
+    /// Rationale: 
+    ///    If there is no intent for reusing the object, do not use pool - just use "new". 
+    /// </summary>
+    internal sealed class ObjectPool<T> where T : class
+    {
+        private struct Element
+        {
+            internal T Value;
+        }
+
+        // storage for the pool objects.
+        private readonly Element[] _items;
+
+        // factory is stored for the lifetime of the pool. We will call this only when pool needs to
+        // expand. compared to "new T()", Func gives more flexibility to implementers and faster
+        // than "new T()".
+        private readonly Func<T> _factory;
+
+
+        internal ObjectPool(Func<T> factory)
+            : this(factory, Environment.ProcessorCount * 2)
+        { }
+
+        internal ObjectPool(Func<T> factory, int size)
+        {
+            _factory = factory;
+            _items = new Element[size];
+        }
+
+        private T CreateInstance()
+        {
+            var inst = _factory();
+            return inst;
+        }
+
+        /// <summary>
+        /// Produces an instance.
+        /// </summary>
+        /// <remarks>
+        /// Search strategy is a simple linear probing which is chosen for it cache-friendliness.
+        /// Note that Free will try to store recycled objects close to the start thus statistically 
+        /// reducing how far we will typically search.
+        /// </remarks>
+        internal T Allocate()
+        {
+            var items = _items;
+            T inst;
+
+            for (int i = 0; i < items.Length; i++)
+            {
+                // Note that the read is optimistically not synchronized. That is intentional. 
+                // We will interlock only when we have a candidate. in a worst case we may miss some
+                // recently returned objects. Not a big deal.
+                inst = items[i].Value;
+                if (inst != null)
+                {
+                    if (inst == Interlocked.CompareExchange(ref items[i].Value, null, inst))
+                    {
+                        goto gotInstance;
+                    }
+                }
+            }
+
+            inst = CreateInstance();
+            gotInstance:
+
+            return inst;
+        }
+
+        /// <summary>
+        /// Returns objects to the pool.
+        /// </summary>
+        /// <remarks>
+        /// Search strategy is a simple linear probing which is chosen for it cache-friendliness.
+        /// Note that Free will try to store recycled objects close to the start thus statistically 
+        /// reducing how far we will typically search in Allocate.
+        /// </remarks>
+        internal void Free(T obj)
+        {
+            var items = _items;
+            for (int i = 0; i < items.Length; i++)
+            {
+                if (items[i].Value == null)
+                {
+                    // Intentionally not using interlocked here. 
+                    // In a worst case scenario two objects may be stored into same slot.
+                    // It is very unlikely to happen and will only mean that one of the objects will get collected.
+                    items[i].Value = obj;
+                    break;
+                }
+            }
+        }
+    }
+}
--- a/src/System.IO.Compression/System.IO.Compression.sln
+++ b/src/System.IO.Compression/System.IO.Compression.sln
@@ -45,4 +45,7 @@ Global
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
+	GlobalSection(CodealikeProperties) = postSolution
+		SolutionGuid = a7e376ea-0587-47fe-88c2-fd75be4b20bb
+	EndGlobalSection
 EndGlobal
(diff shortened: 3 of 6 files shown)


src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Produces an instance.
+        /// </summary>
+        /// <remarks>
+        /// Search strategy is a simple linear probing which is chosen for it cache-friendliness.
+        /// Note that Free will try to store recycled objects close to the start thus statistically 
+        /// reducing how far we will typically search.
+        /// </remarks>
+        internal T Allocate()
+        {
+            var items = _items;
```
Maxwe11: :exclamation: coding style violation: rule [# 10](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                // Note that the read is optimistically not synchronized. That is intentional. 
+                // We will interlock only when we have a candidate. in a worst case we may miss some
+                // recently returned objects. Not a big deal.
+                inst = items[i].Value;
+                if (inst != null)
+                {
+                    if (inst == Interlocked.CompareExchange(ref items[i].Value, null, inst))
+                    {
+                        goto gotInstance;
```
Maxwe11: IMO It would be better to use boolean flag

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
redknightlois: @maxwe11 I actually took this verbatim from: https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/ObjectPool%601.cs

I was signing the CLA and then opening a new ticket to make ObjectPool<T> available throught all CoreFX to avoid redundant code-copying.

Maxwe11: In that case I'd rather apply on this file `git mv` to `src/Common` folder and reuse it :smile:

src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
```diff
@@ -611,6 +614,8 @@ protected override void Dispose(bool disposing)
                         base.Dispose(disposing);
                     }
                 }  // finally
+
+                s_byteBufferPool.Free(_buffer);
```
VSadov: I am not familiar with the deflater codebase, but quite familiar with this pool. One thing to watch out when using ObjectPool, especially if it is static, is to not return back large objects. 
If _buffer may expand, make sure you are not returning really big ones.

redknightlois: It does not, they are all the same size. 32Kb and 8Kb (all below the LOH threshold).

VSadov: Pool limits the number of retained objects, but you also need to ensure a limit for the size of stored objects. Otherwise the cache is unbounded - also known as memory leak.

redknightlois: I selected the size to ensure that a max of 1MB of memory is consumed, as arrays are all of the same size (and will never change, _buffer is readonly just in case) and we can know that only 32 are going to be retained.

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
VSadov: We are competing with new/GC here on the basis that we know the scenario and can do better.
To be profitable this piece needs to be fast and that often requires some code patterns you would not otherwise use. I'd ignore goto here.

BTW, Roslyn has a slightly newer version that has special treatment for situations when a single instance is sufficient in most cases, which is relatively common. It basically adds a singleton stash in addition to an array.

https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/SharedCollections/ObjectPool%601.cs

src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs
VSadov: Sounds good then.

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
redknightlois: I am fairly acquainted with the Roslyn ObjectPool, it is the general consensus to use that one instead on CoreFX? It feels a bit too broad a change for me to do on my first PR :)

VSadov:
CC @stephentoub

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
VSadov: All the leak tracking goo is unnecessary. You can get same data from any profiler if needed. The only interesting difference there is that additional  single item cache. 
Yes, that can be added later.

stephentoub:
cc: @KrzysztofCwalina, @vancem 

I applaud the goal, but I have concerns about adding such an object pool under DeflateStream.  We could end up significantly increasing the amount of space taken up by the process; in some apps that maintain a steady state of use, that's probably fine, but in other apps that use a DeflateStream only here and there, or worse use a bunch in a quick burst and then never again, this could effectively be a big leak.  We'd probably need to invest in much better object pooling before being comfortable having such a pool hidden beneath such a common stream.

Personally, I think I'd prefer to see a different approach taken:
- Introduce some sort of buffer pool interface to the framework (e.g. IBufferPool), with some good default implementation that's also exposed (e.g. BufferPool)
- Add ctors to streams like DeflateStream and FileStream, to methods like Stream.CopyToAsync, etc. that take an IBufferPool instead of or in addition to a buffer size
- Leave it up to the consumer of the class/API whether to pass in a pool and, if it does, which instance of which kind to use

We'd want to design such a solution holistically across the set of types that would use it rather than one-off for an individual stream.  @KrzysztofCwalina has already started experimenting with buffer pools in corefxlabs at https://github.com/dotnet/corefxlab/tree/master/src/System.Buffers.

redknightlois:
@stephentoub While I agree that a more holistical solution, but those would not improve the life either of those that have already code out there either (not a pit of success as @friedroman commented). This solution (we should think about how big a footprint make sense) is an stopgap there. That's essentially the reason behind limiting the memory footprint to under 1Mb. 

We dont know where CoreCLR is going to be used, but it seems a small price to pay from the eventual usage scenarios into the mid-range to heavy-range usage scenarios with also immediate improvement to all users of the class without changes. In the case of the single usage and even many uses in short intervals (not multithreaded apps) there is no reason why the pool will grow over a few instances (the amount of simultaneously used DefrateStream instances); so the effective footprint for all quick burst and fire-and-forget cases should be low enough (except in a few degenerated cases). While the improvement on heavy-usage and to current code is still retained.

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
jasonwilliams200OK: > I was signing the CLA and then opening a new ticket to make...

Apparently the first part didn't went through. :smile:

redknightlois: No it didnt. Sent an email to the foundation requesting support. Not even using the Access documents option worked.

redknightlois: @stephentoub I didn't get any response from either the ticket I opened to DocuSign and the email I sent to the DotNet Foundation that is specified in the email. Do you know who I could contact to be able to sign the CLA? Because it is not working, tried in Chrome and Edge, no luck.

stephentoub: @richlander, @terrajobst, can you help @redknightlois?

stephentoub:
> stopgap

I understand.  But the problem with a "stopgap" like this is that we very likely won't be able to remove it, once folks rely on its being there.

redknightlois:
I meant "stopgap" as in: it works as it is, if you are not in the 90% (I am making this number up, but we should aim at least that coverage), then you should use a different solution. Not as in something thansitory. 

Currently if you rely on GZipStream to do anything outside of unzipping a few files the GC pressure is not negligible. Every time we create a GZipStream we are allocating up to 40kb just to do nothing (and 8K if you are using the zlib version). Then there are internal allocation in the managed provider in the range of 32kb plus other allocations on the process of compressing and decompressing itself (which I didn't consider in this PR).

stephentoub:
> Every time we create a GZipStream we are allocating up to 40kb just to do nothing (and 8K if you are using the zlib version)

The zlib version should be what's basically used everywhere.  In what situation with corefx are you not using the zlib implementation?  As I've stated multiple times, I fundamentally agree there's a problem that needs to be solved here, e.g. an 8KB buffer being allocated for each stream.  I just want to make sure we're not overstating the problem.
e
> I meant "stopgap" as in

This is a very slippery slope.  If every stream type maintained its own buffer pool, that could get very expensive very quickly.  What about FileStream?  NetworkStream?  PipeStream? BufferedStream? What about methods that do a copy and allocate a buffer, e.g. Stream.CopyTo?  Etc.  They're all over the place, and if we tackled them each independently, I really fear we're going to create a mess.

@vancem, opinions?

KrzysztofCwalina:
We need to get and use a shared buffer pool. We are getting very close to having cycles to look into implementing such shared pool.

redknightlois:
@stephentoub They are not nearly close, most do not even allocate at all. The only one that does allocate heavily is MemoryStream (nothing that it makes sense to do there) and BufferedStream. BufferedStream is nowhere near, it uses by default 4K and it also allows you to decide to use less buffer in cases where you know you are not going to gain either (not as great, but it can help on some scenarios as the one I discovered through profiling).

stephentoub:
> They are not nearly close, most do not even allocate at all

I don't understand this comment.  
FileStream by default allocates a 4K buffer.  
BufferedStream by default allocates a 4K buffer. 
MemoryStream obviously allocates as much as is necessary to store all the data.
StreamReader/StreamWriter by default allocate a 1K buffer.
Stream.CopyTo by default allocates an 80K buffer.
Etc.

Even just looking at the 4K ones, I don't understand why 4K is so different from 8K that the 8K one would warrant a solution and the 4K one wouldn't.  Sure, with a stream like FileStream you can lower the size of the buffer, but in doing so you sacrifice other performance benefits.

redknightlois:
For most high performance scenarios, 4K is bad too, but it is half as bad as 8K (sorry I couldnt resist the attempt at humor :D). 

```Stream.CopyTo``` indeed as you noted is very bad. Didnt lookup the code for that one, but I would have never though about it, as for ages I have been using custom implementations for it. As you noted the stock one is quite bad if you are calling it too many times. https://github.com/dotnet/coreclr/issues/1494#issuecomment-138495504
Our implementation uses an internal internal memory pool to keep allocations in check :) . 

Even in the case of ```FileStream``` (as in ```BufferedStream```) the constructor allows you to control the buffer size and therefore the allocations. https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/FileStream.cs#L493

Most of the time is a game of trying to figure out how big a buffer would make sense trying to get the most out of the allocation vs. read performance function.

src/System.IO.Compression/src/System/IO/Compression/Internal/ObjectPool`1.cs
stephentoub: @redknightlois, I spoke with @martinwoodward  at the foundation, and he's concerned because he didn't see any email from you nor any issues with signing.  He asked that you resend your email and include him directly: martin@dotnetfoundation.org.  Sorry for the hassle, and thanks for persisting.

redknightlois: Done. Forwarded the original email to Martin.

KrzysztofCwalina:
@sokket, could you look into this and see if the buffer pool you are working on coul dbe used here.

martinwoodward:
@stephentoub - we have a signed CLA back from @redknightlois now (thanks for bearing with us, still not sure what was going wrong there but we did it manually to unblock this one). Any new PR's created by him in the future should come up as cla-signed automatically. I manually changed this one to avoid it having to be closed and re-opened.

stephentoub:
Thanks, @martinwoodward and @redknightlois!

sokket:
@KrzysztofCwalina it looks like it could since this is using a byte[] buffer pool; either of my implementations could work here. I'll sync up with you offline to figure out next steps

stephentoub:
@redknightlois, thanks for the example PR.  I'm going to close this now.  We still have https://github.com/dotnet/corefx/issues/1991 tracking improvements around this, and we have https://github.com/dotnet/corefx/issues/4547 tracking adding a shared buffer pool to corefx, which we'd want to use instead of a custom implementation unique to this assembly.  Part of @sokket's work on https://github.com/dotnet/corefx/issues/4547 will be in validating the pool in use by assemblies like this and others in corefx.

KrzysztofCwalina:
@sokket , @Priya91, @redknightlois, who will do a PR to use the shared pool in compression?

sokket:
@KrzysztofCwalina I opened issue #4875 to track this for RC2, assigned to myself

kerryjiang:
Hello guys, It's pretty good to have a buffer pool in the assembly System.Buffers.
But ObjectPool is required in lots of situations and I also found many different internal ObjectPool implementations in CoreCLR and CoreFx.

So, do you have a plan to make one of the ObjectPools public and move it to a common place?

PR #3703, opened 2015-10-07
#1991 - Reusable/Reset approach

redknightlois:
Allow DefrateStream to be able to be closed and reset to allow users to avoid the heavy allocations that happen with the managed implementation and the construction costs.

Alternative approach: #1991

### DIFF
--- a/src/System.IO.Compression/System.IO.Compression.sln
+++ b/src/System.IO.Compression/System.IO.Compression.sln
@@ -45,4 +45,7 @@ Global
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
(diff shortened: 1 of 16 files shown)


stephentoub:
Thanks, @redknightlois.  I'm going to close this one out for now.  With new surface area, it would need to go through the API review process, and assuming it succeeded in that, would need to be submitted to future rather than to master.  I think your primary goal was simply to highlight how the approach might be implemented, and the PR will still be viewable after it's closed.

redknightlois:
Yes, the idea was to show the potential approaches. Either way I would like to sign the CLA, which DocuSign is not allowing me to.

stephentoub:
> Either way I would like to sign the CLA, which DocuSign is not allowing me to.

I've pinged @richlander and @terrajobst again.

PR #3704, opened 2015-10-07
Don't make imported key automatically exportable.

AtsushiKan:
cc @bartonjs 

Removing the "set to exportable" behavior of the import helper I just added.

- It's an unnecessary hiccup in sharing the code with desktop (since the ExportPolicy setter is private, not internal there.)

- We don't need it. (We only use this to translate plaintext keys to Cng in AesCng - we still store the plaintext in the base class Key property so it's still retrievable.) So it's just wasted cycles.

- The decision should be up to the guy who called Import.


bartonjs:
Makes sense, LGTM

Issue #3149, opened 2015-09-10
Expression compiler - Inconsistency with language specification


bartdesmet:
Section 7.9 of the C# language specification describes the number of bits to shift. It looks like this rule is not honored by the expression compiler which omits the `and` instruction to perform the mask, prior to performing the `shl` or `shr` instruction. The following repros the discrepancy:

Expression&lt;Func&lt;long?, int?, long?&gt;&gt; e = (x, y) => x &lt;&lt; y;

Func&lt;long?, int?, long?&gt; f = (x, y) => x &lt;&lt; y;
Func&lt;long?, int?, long?&gt; g = e.Compile();

var c = f(-9223372036854775808, -2147483648);
var d = g(-9223372036854775808, -2147483648);

Console.WriteLine(c);
Console.WriteLine(d);

Note that the implementation of the expression interpreter relies on C# compiled code fragments for << and >>, so this will cause a different result for .Compile() between frameworks that use the expression compiler versus those that use the interpreter.

mikedn:
> It looks like this rule is not honored by the expression compiler which omits the `and` instruction to perform the mask, prior to performing the `shl` or `shr` instruction. 

Hmm, is this a bug in the expression compiler or is this a bug in the C# compiler? The [LeftShift](https://msdn.microsoft.com/en-us/library/bb360628(v=vs.110).aspx) documentation doesn't say anything about this so one could conclude that it's the C# compiler obligation to generate the necessary `and`, the same way it does when it emits IL.

VSadov:
more interestingly the repro prints different results for me only on x86 /o+

VSadov:
IL spec says:
```
III.3.58 shl – shift integer left
. . . 
The return value is
unspecified if shiftAmount is greater than or equal to the width of value
```
So this is a bug in the ET compiler. It needs to mask or the op will result in undefined behavior

bartdesmet:
Also, the C# compiler emitting the And node would be a breaking change and it would not be something an expression analyzer would expect to see.

Sent from my Windows Phone
________________________________

VSadov:
@bartdesmet - yes, that is correct. Changing ET shape is a breaking change. No need to look far for examples.
https://github.com/dotnet/roslyn/issues/4471 - that is basically a reintroduction of a subtle semantical bug, just to keep clients from breaking over changes in the tree shape.

mikedn:
> more interestingly the repro prints different results for me only on x86 /o+

[...]

PR #3706, opened 2015-10-07
Fix for LeftShift and RightShift code generation in LambdaCompiler

bartdesmet:
This fixes issue #3149 in the LambdaCompiler by inserting the masking behavior similar to what the C# and VB compilers do.


VSadov:
LGTM

VSadov:
I assume interpreter naturally does the right thing so does not need a similar fix.

bartdesmet:
Correct; it relies on statically compiled C# code that emits the same masks.

bartdesmet:
I've removed the stale comments in the interpreter that were the result of some copy/paste when the instructions were implemented.

Issue #3244, opened 2015-09-14
Consider having expression interpreter available in all build flavors


bartdesmet:
For some of our Bing service, we often run into the case where expression compilation is too expense, either because it's on a request/response path, or because it's statically known that the expression is small and will be evaluated only once. I can provide more details on such scenarios if needed.

Unfortunately, expression interpretation is a property of the .NET build flavor, e.g. for .NET Native the Compile method uses the interpreter.

It'd be nice to have an Interpret method alongside Compile to enable those scenarios. It'd also help a lot in writing parity tests for the expression compiler and interpreter because their outcomes can be cross-checked.

bartdesmet:
Discussed offline with @VSadov; broadening the audience here now.

VSadov:
@jaredpar @madstorgersen - FYI.

VSadov:
I think allowing compiling and interpreting behavior to be both possible in the same binary would be useful. Currently these are mutually exclusive - I.E. you can have System.Linq.Expressions that either has compiler or interpreter. 

There could be some technical difficulties in implementation, but hopefully these can be resolved.

We would need some new API to get to the interpreter. Most obvious would be ".Interpret" alongside with Compile, but it also could be an overload of Compile like Compile(bool preferInterpretation) or something like that.

In terms of testing - We need to test both compiler and interpreter anyways and this change will only make that easier.

antiufo:
I recently noticed that Expression.Compile() was in the hot path of my software project (many small, single-use compilations). I refactored out the interpreter and made it available to Net45/Dnxcore50 (nuget: Shaman.System.Linq.Expressions.Interpreter). The whole thing now runs much faster.
I agree with the idea of making the interpreter publically available.

VSadov:
We have discussed this and it seems a good thing to have.
As for the entry point, the preferred shape would be "Compile(bool preferInterpretation)"

The reasons are:
1) Compile() does not necessary compile. It basically prefers compiling if available, but otherwise will interpret.
2) The method does not really start interpretation. Technically it still compiles into a delegate, just with a different form that will be executed by the means of interpretation.

Compile(bool preferInterpretation) should use interpreter when preferInterpretation == true and interpreter is available (I see no reasons for it to be unavailable, but who knows..)

When preferInterpretation == false, it behaves like the old Compile() - it will use compiler if available and interpreter otherwise.

VSadov:
[...]

PR #3714, opened 2015-10-08
Initial work to expose lambda interpreter alongside lambda compiler

bartdesmet:
See issue #3244. This is a first stab at exposing an Interpret method alongside the Compile method on LambdaExpression and Expression<T> so the user can choose whether to interpret (e.g. for small expression fragments that need to be evaluated once) or to compile (when available on the target platform).

Whether or not the Interpret method is included is based on the FeatureInterpret build flag. This is different from the IsInterpreting build flag which changes the behavior of Compile to use Interpret on platforms that don't have dynamic code gen. Note that I've switched the defined build-time constants to be a bit more descriptive, e.g. FEATURE_COMPILER (meaning the lambda compiler is available and should be used for the Compile method) and FEATURE_INTERPRETER (meaning the Interpret method is exposed) rather than being tied to the target platform e.g. FEATURE_CORECLR.

By adding this public API surface, we can cross-check the compiler and interpreter in tests. A simple example is provided in this initial PR. It can be used to repro a variety of issues with the interpreter that were reported before.

### DIFF
--- a/src/System.Dynamic.Runtime/src/System.Dynamic.Runtime.csproj
+++ b/src/System.Dynamic.Runtime/src/System.Dynamic.Runtime.csproj
@@ -9,7 +9,7 @@
     <AssemblyName>System.Dynamic.Runtime</AssemblyName>
     <RootNamespace>System.Dynamic.Runtime</RootNamespace>
     <AssemblyVersion>4.0.11.0</AssemblyVersion>
-    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_CORECLR</DefineConstants>
+    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_COMPILER</DefineConstants>
   </PropertyGroup>
   <!-- Default configurations to help VS understand the configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
--- a/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
+++ b/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
@@ -9,7 +9,8 @@
     <AssemblyName>System.Linq.Expressions</AssemblyName>
     <RootNamespace>System.Linq.Expressions</RootNamespace>
     <AssemblyVersion>4.0.11.0</AssemblyVersion>
-    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_CORECLR</DefineConstants>
+    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_COMPILE</DefineConstants>
+    <DefineConstants Condition=" '$(FeatureInterpret)' == 'true' ">$(DefineConstants);FEATURE_INTERPRET</DefineConstants>
   </PropertyGroup>
   <!-- Default configurations to help VS understand the configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
@@ -137,7 +138,6 @@
     <Compile Include="System\Linq\Expressions\Compiler\CompilerScope.cs" />
     <Compile Include="System\Linq\Expressions\Compiler\CompilerScope.Storage.cs" />
     <Compile Include="System\Linq\Expressions\Compiler\ConstantCheck.cs" />
-    <Compile Include="System\Linq\Expressions\Compiler\DelegateHelpers.Generated.cs" />
     <Compile Include="System\Linq\Expressions\Compiler\HoistedLocals.cs" />
     <Compile Include="System\Linq\Expressions\Compiler\ILGen.cs" />
     <Compile Include="System\Linq\Expressions\Compiler\KeyedQueue.cs" />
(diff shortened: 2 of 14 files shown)


src/System.Linq.Expressions/src/System/Linq/Expressions/LambdaExpression.cs
```diff
@@ -113,16 +113,27 @@ public bool TailCall
...
         }
 
-#if FEATURE_CORECLR
+#if FEATURE_INTERPRET
+        /// <summary>
+        /// Produces a delegate that represents the lambda expression.
+        /// </summary>
+        /// <returns>A delegate containing the compiled version of the lambda.</returns>
+        public Delegate Interpret()
+        {
```
VSadov: We discussed this a bit before (I've lost where I placed the comments, perhaps I forgot to publish..).
It seems better if the entry point has shape   "Compile(bool preferInterpretation)"

The reasons are:
1) Compile()   does not necessary compile. It basically prefers compiling if available, but otherwise will interpret.
2) The method does not really start interpretation. Technically it still compiles into a delegate, just with a different form that will be executed by the means of interpretation.

src/System.Linq.Expressions/src/System/Linq/Expressions/LambdaExpression.cs
```diff
@@ -113,16 +113,27 @@ public bool TailCall
         /// <returns>A delegate containing the compiled version of the lambda.</returns>
         public Delegate Compile()
         {
-#if FEATURE_CORECLR
+#if FEATURE_COMPILE
             return Compiler.LambdaCompiler.Compile(this);
 #else
             return new System.Linq.Expressions.Interpreter.LightCompiler().CompileTop(this).CreateDelegate();
 #endif
```
VSadov: This should just forward to   Compile(preferInterpretation: true)

we cannot support case where both compiling and interpreting are unavailable anyways.

src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
```diff
@@ -9,7 +9,8 @@
     <AssemblyName>System.Linq.Expressions</AssemblyName>
     <RootNamespace>System.Linq.Expressions</RootNamespace>
     <AssemblyVersion>4.0.11.0</AssemblyVersion>
-    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_CORECLR</DefineConstants>
+    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_COMPILE</DefineConstants>
+    <DefineConstants Condition=" '$(FeatureInterpret)' == 'true' ">$(DefineConstants);FEATURE_INTERPRET</DefineConstants>
```
VSadov: I assume FEATURE_INTERPRET will be set true unconditionally at some point. We definitely need to set it when IsInterpreting or we would have no compiler and no interpreter.

src/System.Linq.Expressions/src/System/Linq/Expressions/LambdaExpression.cs
VSadov: Compile(bool preferInterpretation)   should use interpreter when preferInterpretation == true and interpreter is available (I see no reasons for it to be unavailable, but who knows..)

When preferInterpretation == false, it behaves like the old Compile() - it will use compiler if available and interpreter otherwise.

bartdesmet: Agree on the method signature change. Should we have it available unconditionally already (i.e. the public API has been blessed)? If so, I'll get rid of the #if there.

VSadov: It has not been formally reviewed, but it makes the most sense if it is unconditional, so it shoud be ok to remove #if for that.

src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
bartdesmet: Still need to make this change, if it's ok for the "preferInterpretation" overload to Compile to show up as a public API when IsInterpreting is set to true (if which case, should it be there all the time, so users that start off on a platform with only the interpreter available don't see that overload go missing when moving to a platform that historically only had the compiler)?

VSadov: I think that is the point for the flag to be suggestive - you can move between interpreter-only platforms and platforms that can use (or have to use) compiling.

bartdesmet:
Notes for the latest iteration:
- Compile(bool) is unconditionally available in order to avoid inconsistencies in public APIs across different build flavors of the framework
- Compile() uses the compiler if available, otherwise falls back to using the interpreter (which should be included in the build if the compiler is not available)
- Build flag "FeatureInterpret" controls whether the interpreter is included and can be used when using Compile(true); this enables builds where both compiler and interpreter are usable from the public APIs and is also used for cross-testing the compiler and the interpreter
- Build flag "IsInterpreting" uses the interpreter everywhere and is used for platforms where the compiler is not available at all; this flag is compatible with the build configuration prior to those changes (modulo the addition of the Compile(bool) public API)

The main thing to review here is the public API addition and its behavior of honoring the flag only when the interpreter is available.

VSadov:
LGTM
@bartdesmet - do you want to make any more changes to this? 
You've mentioned that it could be simplified by removing FeatureInterpret

bartdesmet:
I think we can integrate this now, so we can unblock the work on testing. We can simplify the build flavors once everything starts to look great in terms of testing. Sounds reasonable?

src/System.Dynamic.Runtime/src/System.Dynamic.Runtime.csproj
```diff
@@ -9,7 +9,7 @@
     <AssemblyName>System.Dynamic.Runtime</AssemblyName>
     <RootNamespace>System.Dynamic.Runtime</RootNamespace>
     <AssemblyVersion>4.0.11.0</AssemblyVersion>
-    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_CORECLR</DefineConstants>
+    <DefineConstants Condition=" '$(IsInterpreting)' != 'true' ">$(DefineConstants);FEATURE_COMPILER</DefineConstants>
```
weshaggard: FEATURE_COMPILER => FEATURE_COMPILE

This looks like the root cause for issue https://github.com/dotnet/corefx/pull/4862. While that issue was fixed by removing the #ifdef it looks like we might also have some other bugs in this library. @DinoV @VSadov @bartdesmet I'm going to file a separate issue to fix this.

PR #3721, opened 2015-10-08
Shim OpenSSL ERR* methods.

eerhardt:
Adding shims for all OpenSSL methods in Interop.ERR.cs.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
@@ -12,7 +12,7 @@ internal static partial class Interop
     internal static partial class Crypto
     {
         [DllImport(Libraries.CryptoNative)]
-        private static extern ulong ErrGetError();
+        private static extern ulong ErrGetError([MarshalAs(UnmanagedType.Bool)] out bool isAllocFailure);
 
         [DllImport(Libraries.CryptoNative, CharSet = CharSet.Ansi)]
         private static extern void ErrErrorStringN(ulong e, StringBuilder buf, int len);
@@ -27,9 +27,6 @@ internal static partial class Interop
 
         internal static Exception CreateOpenSslCryptographicException()
         {
-            const uint ReasonMask = 0x0FFF;
-            const uint ERR_R_MALLOC_FAILURE = 64 | 1;
-
             // The Windows cryptography library reports error codes through
             // Marshal.GetLastWin32Error, which has a single value when the
             // function exits, last writer wins.
@@ -43,15 +40,19 @@ internal static partial class Interop
             // Windows code and the OpenSSL-calling code, drain the queue
             // whenever an Exception is desired, and report the exception
             // related to the last value in the queue.
-            ulong error = ErrGetError();
+            bool isAllocFailure;
+            ulong error = ErrGetError(out isAllocFailure);
             ulong lastRead = error;
+            bool lastIsAllocFailure = isAllocFailure;
 
             // 0 (there's no named constant) is only returned when the calls
             // to ERR_get_error exceed the calls to ERR_set_error.
             while (lastRead != 0)
             {
                 error = lastRead;
-                lastRead = ErrGetError();
+                isAllocFailure = lastIsAllocFailure;
+
+                lastRead = ErrGetError(out lastIsAllocFailure);
             }
 
             // If we're in an error flow which results in an Exception, but
@@ -62,17 +63,14 @@ internal static partial class Interop
                 return new CryptographicException();
             }
 
-            // Even though ErrGetError returns ulong (C++ unsigned long), we 
-            // really only expect error codes in the UInt32 range
-            Debug.Assert(error <= uint.MaxValue, "ErrGetError should only return error codes in the UInt32 range.");
-
-            // Inline version of the ERR_GET_REASON macro.
-            ulong reason = error & ReasonMask;
-
-            if (reason == ERR_R_MALLOC_FAILURE)
+            if (isAllocFailure)
             {
                 return new OutOfMemoryException();
             }
+
+            // Even though ErrGetError returns ulong (C++ unsigned long), we 
+            // really only expect error codes in the UInt32 range
+            Debug.Assert(error <= uint.MaxValue, "ErrGetError should only return error codes in the UInt32 range.");
 
             // If there was an error code, and it wasn't something handled specially,
             // use the OpenSSL error string as the message to a CryptographicException.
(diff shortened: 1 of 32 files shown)


stephentoub:
LGTM. Thanks, Eric.

stephentoub:
(Looks like a "libcrypto" to "Crypto" change may have been missed in one place: "Internal\Cryptography\AesOpenSslCryptoTransform.cs(137,41): error CS0117: 'Interop.libcrypto' does not contain a definition for 'CreateOpenSslCryptographicException' ")

eerhardt:
:(.  I missed that when I merged with the latest changes.  I'll get it fixed.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
```diff
@@ -2,25 +2,26 @@
...
-    internal static partial class libcrypto
+    internal static partial class Crypto
     {
         [DllImport(Libraries.LibCrypto)]
-        private static extern uint ERR_get_error();
+        private static extern ulong ErrGetError();
 
         [DllImport(Libraries.LibCrypto, CharSet = CharSet.Ansi)]
-        private static extern void ERR_error_string_n(uint e, StringBuilder buf, int len);
+        private static extern void ErrErrorStringN(ulong e, StringBuilder buf, int len);
```
bartonjs: I feel like @stephentoub just recently marked `buf` as `[In]` here for marshal performance.

stephentoub: My PR is still pending.  I'll handle the merge conflicts after this goes in.

bartonjs: Ah, well, at least I remembered seeing it; thanks for confirming :)

stephentoub: (That said, if we care about performance here, I think we'd change this to just stackalloc a byte*, change the signature to take a byte* instead of a StringBuilder, and then use Marshal.PtrToStringAnsi to extract the string.)

bartonjs:
LGTM.  Continued thanks, @eerhardt!

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
eerhardt: @stephentoub is that something you want me to take care of in this change? Or something you will do in the other PR? Or something that requires its own change?
For this case, it seems minor, since this is only called when we get an error from OpenSSL.

stephentoub: Don't worry about it for this change, unless you really feel like it.  As you say, it's an error path; if it were anything else, I'd be jumping up and down to fix it immediately :smile:  At some point soon we should go through all such usage and see whether it makes sense to optimize.

src/Native/System.Security.Cryptography.Native/pal_err.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <openssl/err.h>
+
+extern "C" uint64_t ErrGetError()
+{
+    return ERR_get_error();
+}
+
+extern "C" void ErrErrorStringN(uint64_t e, char* buf, int32_t len)
+{
+    ERR_error_string_n(e, buf, static_cast<size_t>(len));
```
nguerrera: Use pal_utilities.h and UnsignedCast(len). Will assert on < 0 and will let the > 0 -> size_t conversion be implicit in the (exceedingly unlikely. I admit) case that positive int32_t is not a valid size_t. Or just using uint32_t len but that will just move casts to managed code.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
```diff
@@ -61,8 +62,12 @@ internal static Exception CreateOpenSslCryptographicException()
                 return new CryptographicException();
             }
 
+            // Even though ErrGetError returns ulong (C++ unsigned long), we 
+            // really only expect error codes in the UInt32 range
+            Debug.Assert(error <= uint.MaxValue, "ErrGetError should only return error codes in the UInt32 range.");
```
nguerrera: Do we need this assert? I'm not seeing what we're doing with reason where we'd care.

bartonjs: We're unchecked casting error as `int` and making it be ex.HResult.  So we're asserting it fits in 32-bits, so there's no dataloss.

nguerrera: I see that now, not sure how I missed it. ;)

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
```diff
@@ -61,8 +62,12 @@ internal static Exception CreateOpenSslCryptographicException()
...
+
             // Inline version of the ERR_GET_REASON macro.
-            uint reason = error & ReasonMask;
+            ulong reason = error & ReasonMask;
```
nguerrera: Where is ERR_R_MALLOC_FAILURE below defined? If it's a constant in C#, there should be a static assert in shim that we have the value right.

nguerrera:
LGTM. Just a few minor comments. Thanks.

sokket:
LGTM 2

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ERR.cs
eerhardt: > Where is ERR_R_MALLOC_FAILURE below defined?

It is defined in C# at the beginning of the method.

> there should be a static assert in shim that we have the value right.

How about instead ErrGetError returns a "out bool isAllocFailure"?  Then we can get rid of all the special knowledge of OpenSSL values in C# code:  ReasonMask, ERR_R_MALLOC_FAILURE, and the simulation of ERR_GET_REASON macro.  If/when we need to add more, we can change the bool into an enum that we define/control in both managed and the shim code.  The shim code would convert OpenSSL reasons into our own reasons before passing it back to managed code. But since we only need 1 now, we can just use a bool and change it later, if needed. Thoughts?

eerhardt:
@nguerrera - How about something like that?

nguerrera:
LGTM

Issue #2070, opened 2015-06-16
How to implement System.Diagnostics.Process on FreeBSD?


ghuntley:
This issue is to discuss how we can implement `System.Diagnostics.Process` on FreeBSD which does not have procfs by enabled by default as [it should be considered deprecated.](http://lists.freebsd.org/pipermail/freebsd-fs/2011-February/010760.html)
# Reference Material
- [Implement basic System.Diagnostics.Process.Modules support on Linux #2188](https://github.com/dotnet/corefx/pull/2188#issuecomment-116671221)
- The System.Diagnostics.Process [specification on MSDN.](https://msdn.microsoft.com/en-us/library/system.diagnostics.process%28v=vs.110%29.aspx) 
- The [FreeBSD process information implementation within Mono](https://github.com/mono/mono/blob/master/mono/io-layer/processes.c#L1888-L1910) (LGPL)
- The [FreeBSD process information implementation within IO.JS](https://github.com/nodejs/io.js/blob/master/deps/uv/src/unix/freebsd.c) (MIT)
- https://www.freebsd.org/cgi/man.cgi?procstat and https://www.freebsd.org/cgi/man.cgi?sysctl(3)
  
  ```
       KERN_PROC
   Return selected information about specific running processes.
  
   For the following names, an array of struct kinfo_proc structures
   is returned, whose size depends on the current number of such
   objects in the system.
  
     Third level name      Fourth level is:
     KERN_PROC_ALL         None
     KERN_PROC_PID         A process ID
     KERN_PROC_PGRP        A process group
     KERN_PROC_TTY         A tty device
     KERN_PROC_UID         A user ID
     KERN_PROC_RUID        A real user ID
  
   If the third level name is KERN_PROC_ARGS then the command line
   argument array is returned in a flattened form, i.e., zero-termi-
   nated arguments follow each other.  The total size of array is
   returned.  It is also possible for a process to set its own
   process title this way.  If the third level name is
   KERN_PROC_PATHNAME, the path of the process' text file is stored.
   For KERN_PROC_PATHNAME, a process ID of -1 implies the current
   process.
  
     Third level name      Fourth level is:
     KERN_PROC_ARGS        A process ID
     KERN_PROC_PATHNAME        A process ID
  ```

/cc: @janhenke @saper @jasonwilliams200OK @josteink

stephentoub:
> The FreeBSD process information implementation within Mono

[...]

PR #3730, opened 2015-10-08
Moving libproc calls in System.Diagnostics.Process over to use shims

sokket:
Moving libproc calls in System.Diagnostics.Process over to use shims

/cc @nguerrera @stephentoub


src/Native/System.Native/pal_process.cpp
```diff
@@ -59,6 +65,9 @@ static_assert(PAL_PRIO_PROCESS == static_cast<int>(PRIO_PROCESS), "");
 static_assert(PAL_PRIO_PGRP == static_cast<int>(PRIO_PGRP), "");
 static_assert(PAL_PRIO_USER == static_cast<int>(PRIO_USER), "");
 
+// We make assumptions based on these being the same
+static_assert(sizeof(char) == sizeof(uint8_t), "");
```
nguerrera: Can't hurt to be paranoid, I guess, but this is guaranteed.

http://en.cppreference.com/w/cpp/language/types
>Besides the minimal bit counts, the C++ Standard guarantees that 
>1 == sizeof(char) <= sizeof(short) <= sizeof(int) <= sizeof(long) <= sizeof(long long).

src/Native/System.Native/pal_process.cpp
```diff
@@ -417,3 +426,23 @@ extern "C" char* GetCwd(char* buffer, int32_t bufferSize)
 
     return getcwd(buffer, UnsignedCast(bufferSize));
 }
+
+extern "C" int32_t ProcListAllPids(int32_t* buffer, int32_t buffersize)
+{
+    return proc_listallpids(reinterpret_cast<void*>(buffer), buffersize);
```
nguerrera: This cast should be allowed implicitly.

src/Native/System.Native/pal_process.h
```diff
@@ -5,6 +5,9 @@
 
 #include "pal_types.h"
 #include <string.h>
+#include <sys/param.h>
+
+struct rusage_info_v3;
```
nguerrera: what's the significance of _v3?

src/Native/System.Native/pal_process.cpp
sokket: Yeah, I just put that in to protect us in case something really wonky happens...it would save a bunch of debugging time :)

src/Native/System.Native/pal_process.h
sokket: The v3 is the latest rusage_info struct version number; it gives us more data and the data that we want

src/Native/System.Native/pal_process.cpp
sokket: Oops, good catch

sokket:
Any other comments?

src/Common/src/Interop/OSX/System.Native/Interop.ProcPidInfo.cs
```diff
@@ -0,0 +1,239 @@
...
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        // Constants from sys/param.h
+        private const int MAXCOMLEN = 16;
```
stephentoub: Are we validating these values somewhere in the native shim?  Seems like otherwise this code still has an implicit dependency on the OS implementation.

src/Native/System.Native/pal_process.h
nguerrera: I don't understand why we have a forward declaration to something we only cast to something else.

src/Common/src/Interop/OSX/System.Native/Interop.ProcRUsage.cs
```diff
@@ -0,0 +1,87 @@
...
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        // Constants from sys/resource.h
+        private const int RUSAGE_SELF = 0;
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal unsafe struct RUsageInfo
```
nguerrera: Are we assuming that these have a fixed layout because they're only used on OS X? Does FreeBSD have these structs, functions?

sokket: These structs do have a fixed layout and are OS X only (as far as I know). This pack started off abstracting away how generic Process Information was retrieved and passing back an intermediate struct, but after we decided to go in a different direction I figured I could just check in the shim'd OS X portions.

src/Common/src/Interop/OSX/System.Native/Interop.ProcPidInfo.cs
sokket: Good catch, thanks. Added static_asserts for each constant and enum value declared on the C# side

sokket:
Note - I ran the code formatter which fixed some non-format-compliant code in other files

src/Native/System.Native/pal_process.cpp
```diff
@@ -17,6 +17,12 @@
 #include <syslog.h>
 #include <unistd.h>
 
+#if HAVE_LIBPROC
+#include <libproc.h>
+#include <sys/proc_info.h>
+#include <sys/resource.h>
+#endif
```
jasonwilliams200OK: As noted by @saper in a related thread (#2070); at some point, would it make sense to use [`sysctl`](https://en.wikipedia.org/wiki/Sysctl) interface explicitly for `__APPLE__` and `BSD`? (note that `BSD` macro is defined in `<sys/param.h>` which is available in all Unices but this macro is defined only in BSD OS's - [source](http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system))

sokket: Yes, it will definitely be beneficial to switch over to sysctl in order to support FreeBSD and not have an OS X specific interface. This pack is more about shim'ing out native calls to make it easier for that to happen down the road. I'm trying to get all PInvoke code switched over to the PAL in time for a release candidate; once the PInvoke calls have been shim'd, I'd be happy to switch this over to sysctl (if it hasn't been done already) :)

jasonwilliams200OK: Thanks. :+1: 
I was feeling an urge to point it out again so it doesn't get forgotten. :smile:

sokket: No problem, I opened issue #3753 to track this work and marked as up-for-grabs in case anyone has time before I do :smiley:

stephentoub: @jasonwilliams200OK, our current focus for the initial release is on Windows, Linux, and OS X.  Given that, and given that we have a working implementation on OS X, it's not a priority at present to spend time rewriting the existing working OS X-specific implementation.  If you or another member of the community would like to provide a sysctl-based FreeBSD implementation, we'd happily review the PR, and if that implementation also provided as good or better support for OS X as does the current OS X implementation, I think we'd happily get rid of the current implementation and switch to the new one.  I wouldn't expect such a PR to come from someone on the corefx team at Microsoft in the near future, though.  Thanks for your continued interest.

src/Native/System.Native/pal_process.h
sokket: The underlying function takes in a ```rusage_info_t``` which is typedef'd to ```void*``` but is actually an ```rusage_info_v3``` under the covers; there is also v1 and v2 versions but those are deprecated.

sokket:
Anything else or can I merge? :)

sokket:
Spoke with @nguerrera offline and we agreed that this shim doesn't have a lot of value add considering the calls here are OS X specific. If work is done to move over to ```sysctl``` then we can re-visit shim'ing this, but otherwise this is not necessary

PR #3733, opened 2015-10-09
Fix dependency conflicts by explicitly matching them

weshaggard:
We had a number of conflicts that could cause issues in the future
plus I wanted to clean out all the warnings.

This should take care of all the warnings of the form:

 C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1819,5):
 warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.
 These reference conflicts are listed in the build log when log verbosity is set to detailed.


weshaggard:
@dotnet-bot test this please

stephentoub:
LGTM

PR #3734, opened 2015-10-09
Fix test RID in build.sh.

pgavlin:
build.sh was not properly setting the test RID when invoked without
a target OS parameter.


pgavlin:
@mmitche @ellismg @stephentoub PTAL

PR #3735, opened 2015-10-09
Shim socket addresses.

pgavlin:
The socket address shim adds native accessors for the various bits that
managed code needs to access, namely:
- Address families
- Port numbers
- IP addresses

This change also updates the name resolution shim to use an iterator-based
design for walking host entries. This interface trades chattiness (a
P/Invoke per IP address) for memory footprint (various native and managed
allocations per IP address).

### DIFF
--- a/src/Common/src/System/Net/SocketAddressPal.Unix.cs
+++ b/src/Common/src/System/Net/SocketAddressPal.Unix.cs
@@ -10,156 +10,129 @@ namespace System.Net
 {
     internal static class SocketAddressPal
     {
-        public const int IPv6AddressSize = Interop.libc.sockaddr_in6.Size;
-        public const int IPv4AddressSize = Interop.libc.sockaddr_in.Size;
         public const int DataOffset = 0;
+
+        public readonly static int IPv6AddressSize;
+        public readonly static int IPv4AddressSize;
+
+        static SocketAddressPal()
+        {
+            int ipv4AddressSize, ipv6AddressSize;
+            unsafe
+            {
+                Interop.Error err = Interop.Sys.GetIPSocketAddressSizes(&ipv4AddressSize, &ipv6AddressSize);
+                Debug.Assert(err == Interop.Error.SUCCESS);
+            }
+
+            IPv4AddressSize = ipv4AddressSize;
+            IPv6AddressSize = ipv6AddressSize;
+        }
+
+        private static void ThrowOnFailure(Interop.Error err)
+        {
+            switch (err)
+            {
+                case Interop.Error.SUCCESS:
+                    return;
+
+                case Interop.Error.EFAULT:
+                    // The buffer was either null or too small.
+                    throw new IndexOutOfRangeException();
+
+                case Interop.Error.EAFNOSUPPORT:
+                    // There was no appropriate mapping from the platform address family.
+                    throw new PlatformNotSupportedException();
+
+                default:
+                    Debug.Fail("Unexpected failure in GetAddressFamily");
+                    throw new PlatformNotSupportedException();
+            }
+        }
 
         public static unsafe AddressFamily GetAddressFamily(byte[] buffer)
         {
+            AddressFamily family;
+            Interop.Error err;
             fixed (byte* rawAddress = buffer)
             {
-                var sockaddr = (Interop.libc.sockaddr*)rawAddress;
-                switch (Interop.CheckedRead((byte*)sockaddr, buffer.Length, &sockaddr->sa_family))
-                {
-                    case Interop.libc.AF_UNSPEC:
-                        return AddressFamily.Unspecified;
-
-                    case Interop.libc.AF_UNIX:
-                        return AddressFamily.Unix;
-
-                    case Interop.libc.AF_INET:
-                        return AddressFamily.InterNetwork;
-
-                    case Interop.libc.AF_INET6:
-                        return AddressFamily.InterNetworkV6;
-
-                    default:
-                        throw new PlatformNotSupportedException();
-                }
+                err = Interop.Sys.GetAddressFamily(rawAddress, buffer.Length, (int*)&family);
             }
+
+            ThrowOnFailure(err);
+            return family;
         }
 
         public static unsafe void SetAddressFamily(byte[] buffer, AddressFamily family)
         {
+            Interop.Error err;
             fixed (byte* rawAddress = buffer)
             {
-                var sockaddr = (Interop.libc.sockaddr*)rawAddress;
-                Interop.CheckBounds((byte*)sockaddr, buffer.Length, &sockaddr->sa_family);
-
-                switch (family)
-                {
-                    case AddressFamily.Unspecified:
-                        sockaddr->sa_family = Interop.libc.AF_UNSPEC;
-                        break;
-
-                    case AddressFamily.Unix:
-                        sockaddr->sa_family = Interop.libc.AF_UNIX;
-                        break;
-
-                    case AddressFamily.InterNetwork:
-                        sockaddr->sa_family = Interop.libc.AF_INET;
-                        break;
-
-                    case AddressFamily.InterNetworkV6:
-                        sockaddr->sa_family = Interop.libc.AF_INET6;
-                        break;
-
-                    default:
-                        Debug.Fail("Unsupported addres family");
-                        throw new PlatformNotSupportedException();
-                }
+                err = Interop.Sys.SetAddressFamily(rawAddress, buffer.Length, (int)family);
             }
+
+            ThrowOnFailure(err);
         }
 
         public static unsafe ushort GetPort(byte[] buffer)
         {
             ushort port;
-
+            Interop.Error err;
             fixed (byte* rawAddress = buffer)
             {
-                var sockaddr = (Interop.libc.sockaddr*)rawAddress;
-                switch (sockaddr->sa_family)
-                {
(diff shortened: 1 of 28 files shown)


pgavlin:
@stephentoub @sokket PTAL

src/Common/src/System/Net/SocketAddressPal.Unix.cs
```diff
@@ -10,156 +10,129 @@ namespace System.Net
...
-                        return AddressFamily.Unspecified;
+                Interop.Error err = Interop.Sys.GetIPSocketAddressSizes(&ipv4AddressSize, &ipv6AddressSize);
+                Debug.Assert(err == Interop.Error.SUCCESS);
+            }
 
-                    case Interop.libc.AF_UNIX:
-                        return AddressFamily.Unix;
+            IPv4AddressSize = ipv4AddressSize;
+            IPv6AddressSize = ipv6AddressSize;
+        }
```
stephentoub: We should double-check the JIT's behavior with regards to treating readonly static Int32s as constants for non-beforefieldinit types.  I think it still will, but on the off-chance it doesn't, it'd probably be more efficient overall to have two static methods, each of which calls GetIPSocketAddressSizes, and use those in the declarations of the field to let the type be beforefieldinit.  Might be worthwhile to do that anyway for slightly more efficient access to the class in general.

src/Common/src/System/Net/SocketAddressPal.Unix.cs
```diff
@@ -10,156 +10,129 @@ namespace System.Net
...
             }
         }
 
+        public static unsafe AddressFamily GetAddressFamily(byte[] buffer)
+        {
+            AddressFamily family;
+            Interop.Error err;
+            fixed (byte* rawAddress = buffer)
+            {
+                err = Interop.Sys.GetAddressFamily(rawAddress, buffer.Length, (int*)&family);
```
stephentoub: Could the AddressFamily just be an ```out AddressFamily``` arg rather than an ```int*```?

src/Common/src/System/Net/SocketAddressPal.Unix.cs
```diff
@@ -10,156 +10,129 @@ namespace System.Net
...
-
-                    case Interop.libc.AF_INET6:
-                        port = Interop.CheckedRead((byte*)sockaddr, buffer.Length, &((Interop.libc.sockaddr_in6*)sockaddr)->sin6_port);
-                        break;
-
-                    default:
-                        Debug.Fail("Unsupported address family");
-                        throw new PlatformNotSupportedException();
-                }
+                err = Interop.Sys.GetPort(rawAddress, buffer.Length, &port);
```
stephentoub: Same general question... if it's possible to use ref/out instead of pointers, that'd be great.  Similarly, is there a performance advantage to defining these signatures to use ```byte*``` instead of ```byte[]``` and letting the runtime do the marshaling?  Or maybe there are other call sites where we only have a ```byte*```?

pgavlin: I thought that the general guidance was to be honest about the types of P/Invoke parameters, even if that necessitated a bit of extra marshalling in consuming code. Personally, I prefer that these definitions remain as close as possible to the native definitions. What's the motivation for relying on the runtime's marshallers?

stephentoub: I don't have a strong preference either way.  For something like a ```byte[]```, where there's a clear mapping between it and a ```byte*```, I have no problem with ```byte[]``` being used in the P/Invoke signature, as it's really just a managed representation of the native concept; if it's the same performance to use a ```byte[]``` as it is to manually use fixed and pass a ```byte*```, then using a ```byte[]``` would have no real downside and an upside of less explicitly unsafe code and less code to be read and maintained.  But if there are other advantages, like maybe this is faster for some reason, or maybe we have other call sites that would benefit more from it being a ```byte*```, then it's fine.

src/Native/System.Native/pal_networking.h
```diff
@@ -40,43 +44,65 @@ enum GetHostErrorCodes
     PAL_NO_ADDRESS = PAL_NO_DATA,
 };
 
-struct PalIpAddress
+/**
+ * Address families recognized by {Get,Set}AddressFamily
```
stephentoub: Would be worth adding a comment that these values must be the same as those in System.Net.Sockets.AddressFamily.

src/Common/src/System/Net/SocketAddressPal.Unix.cs
pgavlin: Using `byte[]` rather than `byte*` is less flexible--we do have call sites that `stackalloc` a `byte` buffer and pass that in (this is another reason that I prefer pointer to arrays/ref/out: pointers imply neither heap allocation nor pinning).

stephentoub: > we do have call sites that stackalloc a byte buffer and pass that in 

That's a good reason to stick with ```byte*```.

nguerrera: I would tend to agree with @stephentoub. I don't see the downside of byte[] (or ref/out to blittable struct) vs byte*, and it does make the code easier to read and maintain IMHO.

I don't believe there would be any perf loss, but tagging @yizhang82 to confirm.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -392,3 +360,270 @@ extern "C" int32_t GetHostName(uint8_t* name, int32_t nameLength)
...
+
+    switch (sockAddr->sa_family)
+    {
+        case AF_UNSPEC:
+            *addressFamily = PAL_AF_UNSPEC;
+            return PAL_SUCCESS;
+
+        case AF_UNIX:
+            *addressFamily = PAL_AF_UNIX;
+            return PAL_SUCCESS;
```
stephentoub: With this, do Socket's "just work" with domain sockets, or is there work necessary elsewhere in the pal to make them work?  If they just work, great!  If not, I'm wondering whether it'd be better to remove this case for now and just have PAL_EAFNOSUPPORT returned until such time that appropriate changes are made across the layer to enable them.

stephentoub:
A few comments, but otherwise LGTM.  Thanks, Pat.

src/Common/src/System/Net/SocketAddressPal.Unix.cs
nguerrera: Missed the extra discussion. Certainly, if we need to stackalloc in some instances, then byte* is of course fine and can't be undone without a perf hit.

FWIW, in some places, I've seen us have multiple declarations to same function with e.g. byte*, byte[] overloads. I'm not sure if that's worth it or not, though.

src/Native/System.Native/pal_networking.h
pgavlin: Yes--I meant to have a comment (and perhaps assertions in the managed code for validation).

src/Native/System.Native/pal_networking.cpp
pgavlin: Yes, domain sockets should "just work" assuming that the consuing managed code has built a `System.Net.SocketAddress` instance that contains a properly laid-out [sockaddr_un](http://man7.org/linux/man-pages/man7/unix.7.html).

sokket:
LGTM - please run the format-code.sh file to make sure all C++ code adheres to our code style :smiley:

src/Common/src/System/Net/SocketAddressPal.Unix.cs
pgavlin: Good suggestion. I'll make the change.

src/Native/System.Native/pal_networking.cpp
stephentoub: Great!  Not as part of this PR, but it'd be good to add a Unix-only test for that, both to prove to ourselves that it does work and to make sure that we don't inadvertently regress it in the future.

pgavlin:
@sokket: I'll run the formatter in a follow-up PR; there's code that wasn't touched by this PR that's being reformatted.

src/Native/System.Native/pal_networking.cpp
pgavlin: That could be tricky: the layout of a `sockaddr_un` may differ between *nix platforms (as these things do). I'll think about it some more.

sokket:
@pgavlin Those files should be fixed up by my PR that should be going in momentarily

sokket:
@pgavlin, @stephentoub brought up a good point that my formatting change might cause merge conflict headaches for you. I'll revert them and do a PR after both of our packs are in to format everything and change the build script to also run the formatter when anyone builds native, preventing these issues

src/Common/src/System/Net/SocketAddressPal.Unix.cs
yizhang82: In most cases, byte[] would be a reasonable choice for the majority of people because it is easier concept for most developers, where perf isn't super critical. But it does have its down sides - it is less flexible, pinning makes memory more fragmented and makes GC less efficient, and might lead to extra memory allocation/free in out/ref cases, etc. In potentially perf critical scenarios in our network libraries, I would agree using byte* is probably the better choice here. And there is no perf loss - we are just passing a pointer. Well, in the case of this change, there is no practical difference since you are doing your own pinning.

PR #3736, opened 2015-10-09
Read the allowed certificate issuers list on Unix SslStream

bartonjs:
The Unix PAL version of GetRequestCertificateAuthorities.

Development:
* Split out SafeInteriorHandle to facilitate creating two more SafeHandle types following that pattern.
* Applied the new shim style for the two macro-functions that were needed for reading STACK_OF(X509_NAME)
* Created two new internal SafeHandle types.  While in the past these were in isolated files in Common\src\Microsoft\Win32\SafeHandles, I wanted to try something new: keeping them in the file with their create/free/primary-usage.

Testing:
I couldn't figure out how to cause the flow that results in the PAL method getting called, so I inserted an artificial call to it in CompleteHandshake, and saw that it was decoding as expected.

This is yet another partial fix for issue #3362 (GetRequestCertificateAuthorities - populate issuers. (CertModule.cs))

cc: @stephentoub @vijaykota @rajansingh10 @shrutigarg @CIPop

### DIFF
--- a/src/Common/src/Microsoft/Win32/SafeHandles/SafeInteriorHandle.cs
+++ b/src/Common/src/Microsoft/Win32/SafeHandles/SafeInteriorHandle.cs
@@ -1,0 +1,52 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Win32.SafeHandles
+{
+    internal abstract class SafeInteriorHandle : SafeHandle
+    {
+        private SafeHandle _parent;
+
+        protected SafeInteriorHandle(IntPtr invalidHandleValue, bool ownsHandle)
+            : base(invalidHandleValue, ownsHandle)
+        {
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            SafeHandle parent = _parent;
+
+            if (parent != null)
+            {
+                parent.DangerousRelease();
+            }
+
+            _parent = null;
+            SetHandle(IntPtr.Zero);
+            return true;
+        }
+
+        public override bool IsInvalid
+        {
+            get
+            {
+                // If handle is 0, we're invalid.
+                // If we have a _parent and they're invalid, we're invalid.
+                return handle == IntPtr.Zero || (_parent != null && _parent.IsInvalid);
+            }
+        }
+
+        internal void SetParent(SafeHandle parent)
+        {
+            bool addedRef = false;
+            parent.DangerousAddRef(ref addedRef);
+            Debug.Assert(addedRef);
+
+            _parent = parent;
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


src/System.Net.Security/src/System/Net/Unix/CertModule.cs
```diff
@@ -132,10 +132,33 @@ internal override X509Certificate2 GetRemoteCertificate(SafeDeleteContext securi
...
+                }
 
-            return issuers;
+                string[] clientAuthorityNames = new string[nameCount];
+
+                for (int i = 0; i < nameCount; i++)
+                {
+                    using (SafeSharedX509NameHandle nameHandle = Interop.Crypto.GetX509NameStackField(names, i))
+                    {
+                        X500DistinguishedName dn = Interop.Crypto.LoadX500Name(nameHandle);
```
vijaykota: One doubt about this name. Will it match the chain.ChainElements[].Certificate.Issuer field? I am trying to confirm that the name returned here would pass the checks at around line 515 in _SecureChannel.cs

src/Common/src/Microsoft/Win32/SafeHandles/SafeInteriorHandle.cs
```diff
@@ -0,0 +1,52 @@
...
+                // If we have a _parent and they're invalid, we're invalid.
+                return handle == IntPtr.Zero || (_parent != null && _parent.IsInvalid);
+            }
+        }
+
+        internal void SetParent(SafeHandle parent)
+        {
+            bool addedRef = false;
+            parent.DangerousAddRef(ref addedRef);
+            Debug.Assert(addedRef);
```
vijaykota: I believe Assert is not needed because DangerousAddRef throws

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
bartonjs: Looking at [the referenced code](https://github.com/dotnet/corefx/blob/master/src/System.Net.Security/src/System/Net/_SecureChannel.cs#L511-L524), yes.  [The code for get_Issuer](https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs#L54-L57) is the same: `LoadX500Name(nameHandle).Name`

src/Common/src/Microsoft/Win32/SafeHandles/SafeInteriorHandle.cs
vijaykota: Nit: Assert is not needed since DangerousAddRef() throws

src/System.Net.Security/src/System/Net/Unix/CertModule.cs
vijaykota: Ok.. thanks

vijaykota:
LGTM and thanks, @bartonjs. The new SafeHandle will save @shrutigarg time in her set_CA_list changes.

stephentoub:
This has a build error:
```
src\Common\src\Interop\Unix\libssl\Interop.libssl.cs(114,31): error CS0117: 'Interop.libcrypto' does not contain a definition for 'CheckValidOpenSslHandle' [d:\j\workspace\dotnet_corefx_linux_debug_prtest\src\System.Net.Security\src\System.Net.Security.csproj]
```

stephentoub:
LGTM

bartonjs:
> This has a build error:

Apparently after rebasing I left that line merge in unstaged changes, and force pushed merely to move the timestamp on the commit.  Oops :)

PR #3742, opened 2015-10-09
Fixed second write issue, introduced by PR#3675

rajansingh10:
1. PR #3675 broke the scenario where there will be successive read/write calls on stream (>1).
2. older code in finally was like, where asyncRequest will always be null.
 finally
            {
                if (asyncRequest == null || failed)
                {
                    _NestedWrite = 0;
                }
            }
3. After removing dead code i.e. asyncRequest in PR #3675 ,the flow will never set _NestedWrite==0 in positive scenario, causing failure when read/write is called next time by the user.
4. Added multi write testcase as well for this.

### DIFF
--- a/src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
+++ b/src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
@@ -196,7 +196,7 @@ namespace System.Net.Security
         }
 
         //
-        // Combined sync/async write method. For sync case asyncRequest==null.
+        // Sync write method.
         //
         private void ProcessWrite(byte[] buffer, int offset, int count)
         {
@@ -207,7 +207,6 @@ namespace System.Net.Security
                 throw new NotSupportedException(SR.Format(SR.net_io_invalidnestedcall, "Write", "write"));
             }
 
-            bool failed = false;
             try
             {
                 StartWriting(buffer, offset, count);
@@ -216,7 +215,6 @@ namespace System.Net.Security
             {
                 _SslState.FinishWrite();
 
-                failed = true;
                 if (e is IOException)
                 {
                     throw;
@@ -226,10 +224,7 @@ namespace System.Net.Security
             }
             finally
             {
-                if (failed)
-                {
-                    _NestedWrite = 0;
-                }
+                _NestedWrite = 0;
             }
         }
 
@@ -307,9 +302,7 @@ namespace System.Net.Security
         }
 
         //
-        // Combined sync/async read method. For sync request asyncRequest==null
-        // There is a little overhead because we need to pass buffer/offset/count used only in sync.
-        // Still the benefit is that we have a common sync/async code path.
+        // Sync read method.
         //
         private int ProcessRead(byte[] buffer, int offset, int count)
         {
@@ -320,7 +313,6 @@ namespace System.Net.Security
                 throw new NotSupportedException(SR.Format(SR.net_io_invalidnestedcall, "Read", "read"));
             }
 
-            bool failed = false;
             try
             {
                 int copyBytes;
@@ -341,7 +333,6 @@ namespace System.Net.Security
             catch (Exception e)
             {
                 _SslState.FinishRead(null);
-                failed = true;
                 if (e is IOException)
                 {
                     throw;
@@ -351,11 +342,7 @@ namespace System.Net.Security
             }
             finally
             {
-                // If sync request or exception.
-                if (failed)
-                {
-                    _NestedRead = 0;
-                }
+                _NestedRead = 0;
             }
         }
 
--- a/src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
+++ b/src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
@@ -3,8 +3,10 @@
 
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -14,6 +16,11 @@ namespace System.Net.Security.Tests
 {
     public class SslStreamStreamToStreamTest
     {
+        private readonly byte[] sampleMsg = Encoding.UTF8.GetBytes("Sample Test Message");
+        private readonly TimeSpan TaskTimeSpan = TimeSpan.FromSeconds(15);
+
+        #region Tests
+
         [Fact]
         public void SslStream_StreamToStream_Authentication_Success()
         {
@@ -32,20 +39,6 @@ namespace System.Net.Security.Tests
                 bool finished = Task.WaitAll(auth, TimeSpan.FromSeconds(3));
                 Assert.True(finished, "Handshake completed in the allotted time");
             }
-        }
-
-        public bool AllowAnyServerCertificate(
-            object sender,
-            X509Certificate certificate,
-            X509Chain chain,
-            SslPolicyErrors sslPolicyErrors)
-        {
-            if (sslPolicyErrors != SslPolicyErrors.RemoteCertificateNameMismatch)
-            {
-                return true;
-            }
-
-            return false;
         }
 
         [Fact]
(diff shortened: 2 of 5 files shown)


rajansingh10:
cc: @stephentoub, @bartonjs, @davidsh, @CIPop, @SidharthNabar, @pgavlin , @vijaykota

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -71,6 +63,79 @@ public void SslStream_StreamToStream_Authentication_IncorrectServerName_Fail()
...
+        {
+            SslStream clientSSLStream = null;
+            SslStream ServerSSLStream = null;
+
+            try
+            {
+                GetServerClientSSLStream(out clientSSLStream, out ServerSSLStream);
+                clientSSLStream.Write(sampleMsg);
+                clientSSLStream.Write(sampleMsg);
+            }
```
stephentoub: There's nothing to verify or assert?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
```diff
@@ -207,7 +207,6 @@ private void ProcessWrite(byte[] buffer, int offset, int count)
                 throw new NotSupportedException(SR.Format(SR.net_io_invalidnestedcall, "Write", "write"));
             }
 
-            bool failed = false;
```
stephentoub: Nit: the comment above about asyncRequest==null looks out of date... there is no asyncRequest here.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
```diff
@@ -320,7 +315,6 @@ private int ProcessRead(byte[] buffer, int offset, int count)
                 throw new NotSupportedException(SR.Format(SR.net_io_invalidnestedcall, "Read", "read"));
             }
 
-            bool failed = false;
```
stephentoub: Nit: same issue with the comment above being out of date

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: all we needed to verify that calling write >=twice does not throw net_io_invalidnestedcall exception. so there is no assert/check required here.

stephentoub: And we don't want to verify that the write actually happened, e.g. by reading the data from the server stream and verifying the second write wasn't a nop for example?

src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
rajansingh10: done.

src/System.Net.Security/src/System/Net/SecureProtocols/_SslStream.cs
rajansingh10: done

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -71,6 +63,79 @@ public void SslStream_StreamToStream_Authentication_IncorrectServerName_Fail()
             }
         }
 
+        [Fact]
+        public void SslStream_StreamToStream_Back2Back_ClientWrite_Success()
+        {
+            SslStream clientSSLStream = null;
+            SslStream ServerSSLStream = null;
```
bartonjs: Naming: the server variable should start with a lowercase letter.
Naming: The long-series-of-capitals rule says it should be `clientSslStream`

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -71,6 +63,79 @@ public void SslStream_StreamToStream_Authentication_IncorrectServerName_Fail()
...
+        {
+            if (sslPolicyErrors != SslPolicyErrors.RemoteCertificateNameMismatch)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private void GetServerClientSSLStream(out SslStream clientSSLStream, out SslStream ServerSSLStream)
```
bartonjs: Naming: GetServerClientSslStream, clientSslStream, serverSslStream

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: since, I am assuming it to be a white box validation & this testcase was mainly to cover the fix, I purposefully chose not to verify the data. I can extend it though as there is no harm in extra validation :)

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -71,6 +63,79 @@ public void SslStream_StreamToStream_Authentication_IncorrectServerName_Fail()
...
+            var clientStream = new FakeNetworkStream(false, network);
+            var serverStream = new FakeNetworkStream(true, network);
+
+            clientSSLStream = new SslStream(clientStream, false, AllowAnyServerCertificate);
+            ServerSSLStream = new SslStream(serverStream);
+
+            X509Certificate2 certificate = TestConfiguration.GetServerCertificate();
+            Task[] auth = new Task[2];
+
+            auth[0] = clientSSLStream.AuthenticateAsClientAsync(certificate.Subject);
```
bartonjs: This would be better as certificate.GetNameInfo(X509NameType.SimpleName, false).  Then it'd be the CN value from the subject (if present) vs the whole subject line (which is a guaranteed non-match)

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: ok

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: done

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: done

rajansingh10:
@dotnet-bot test this please

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
...
     {
+        private readonly byte[] sampleMsg = Encoding.UTF8.GetBytes("Sample Test Message");
+        private const int TaskWaitTime = 15; // seconds
+
+        #region Tests
+
         [Fact]
         public void SslStream_StreamToStream_Authentication_Success()
         {
+            SslStream clientSslStream = null;
```
CIPop: Please make this code consistent with other tests. The helper functions, while enabling code reuse make this code harder to read and so I would remove them.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -48,29 +125,56 @@ public void SslStream_StreamToStream_Authentication_Success()
...
-                    auth[0].GetAwaiter().GetResult();
-                });
+            if (serverSslStream != null)
+            {
+                serverSslStream.Dispose();
+            }
+        }
 
-                auth[1].GetAwaiter().GetResult();
+        private void DisposeNetworkStream(FakeNetworkStream clientStream, FakeNetworkStream serverStream)
```
CIPop: This should be replaced with a _using_ statement.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
...
+                {
+                    auth[0].GetAwaiter().GetResult();
+                });
+
+                auth[1].GetAwaiter().GetResult();
+            }
+        }
+
+        [Fact]
+        public void SslStream_StreamToStream_Back2Back_ClientWrite_Success()
```
CIPop: Test names need to be very clear: what does "Back2Back" mean?

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
...
         public void SslStream_StreamToStream_Authentication_Success()
         {
+            SslStream clientSslStream = null;
+            SslStream serverSslStream = null;
+            FakeNetworkStream clientStream = null;
+            FakeNetworkStream serverStream = null;
+
+            try
+            {
+                bool result = GetServerClientSslStream(out clientSslStream, out serverSslStream, out clientStream, out serverStream, 3);
```
CIPop: Having so many parameters and so many of them marked as `out` is a [code smell](https://en.wikipedia.org/wiki/Code_smell). This particular case shows that the code doesn't encapsulate well and doesn't respect the single responsibility principle.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
...
+
+                clientSslStream.Write(sampleMsg);
 
-                bool finished = Task.WaitAll(auth, TimeSpan.FromSeconds(3));
-                Assert.True(finished, "Handshake completed in the allotted time");
+                // The condition on which read method (UpdateReadStream) in FakeNetworkStream does a network read is flawed.
+                // That works fine in single read/write but fails in multi read write as stream size can be more, but real data can be < stream size.
+                // So I am doing an explicit read here. This issue is specific to test only & irrespective of xplat.
+
+                serverStream.DoNetworkRead();
```
CIPop: This isn't something that you can actually do with a NetworkStream. The comment above sounds like you're hitting a limitation of the FakeNetworkStream which should be addressed there instead of breaking encapsulation.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
 {
     public class SslStreamStreamToStreamTest
     {
+        private readonly byte[] sampleMsg = Encoding.UTF8.GetBytes("Sample Test Message");
+        private const int TaskWaitTime = 15; // seconds
```
CIPop: Why not directly defining a timespan instead of allocating a new one below?

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
```diff
@@ -14,27 +16,102 @@ namespace System.Net.Security.Tests
...
+
+        #region Tests
```
CIPop: Regions in my opinion are another code smell that shows that there's just too much in a single file. Since this grew bigger, we now need to extract the two classes below in their own files for clarity.

CIPop:
Please refactor and redesign the object model. We should aim towards having tests written as sample code that non-platform devs can refer to.

It's usually better to have more verbose tests that can be easily read instead of factoring into multiple simple functions. If small variations are needed, constructs such as XUnit's Theory can make code reuse simpler while maintaining readability.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: Regions not necessarily are code smell. Infact they are used to organize the code in a better way for the sake of readability. While putting more classes in single class & making it a master sort of class is definitely a bad design. I will take out the stream & network classes out. to another files

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: existing code was like that. I will define one.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: removed. Not touching this tc anymore.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: done

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: done.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: done

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: yep, this is a work-around to unblock the test. Could not find the actual fix after spending some time. So putting the work-around. I can open an issue to track this & fix it as per the priority.

rajansingh10:
@CIPop I have made relevent changes. Factored out the exsiting classes into seperate files.

stephentoub:

stephentoub:
@dotnet-bot test this please (SendPingAsyncWithHostAndTimeoutAndBufferAndPingOptions timed out)

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
CIPop: Yes, please add a TODO and add this to a new or existing tracking issue.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
CIPop: I used to think the same way. See the following article presenting some interesting points on why regions are always bad:
   * http://programmers.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell
   * StyleCop rule# SA1124: DoNotUseRegions

CIPop:
Thanks for refactoring the test code @rajansingh10! 
Only one nit about a missing TODO to fix the network behavior encapsulation of the fake network issue that we should look at in the future. 

LGTM

rajansingh10:
Test Issue #3802 to track the network read issue. Squashed the commits. Will merge once check passes. Thanks @CIPop @stephentoub @bartonjs for the review.

src/System.Net.Security/tests/FunctionalTests/SslStreamStreamToStreamTest.cs
rajansingh10: ok. thats a new learning for me :)

PR #3746, opened 2015-10-09
 Fix ThreadLocal test failures

### LINKED ISSUE #3689: RunThreadLocalTest8_Values failed in CI
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_windows_release_prtest/4721/console

```
System.Threading.Tests.ThreadLocalTests.RunThreadLocalTest8_Values [FAIL]
        RunThreadLocalTest8_Values: Expected values to contain both previous values and 1000 new values
        Expected: True
        Actual:   False
        Stack Trace:
           d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Threading\tests\ThreadLocalTests.cs(259,0): at System.Threading.Tests.ThreadLocalTests.RunThreadLocalTest8_Values()
```

ericeil:
RunThreadLocalTest8_Values had a couple of issues:

1. It expected ThreadLocal values to be retained for threads that had
exited, even when new threads stored their own values.  This cannot be
the case, as it would eventually exhaust memory.

2. The Tasks created in the "for" loop were capturing the iteration
variable, not the particular value for each iteration.  So the
ThreadLocal values were not getting set correctly.

I've fixed these by adding some synchronization, capturing the correct
value, and removing a redundant test that interfered with the later
logic.

Fixes #3689.


stephentoub:
Thanks for fixing this, Eric, but I have some questions...

> It expected ThreadLocal values to be retained for threads that had exited, even when new threads stored their own values. This cannot be the case, as it would eventually exhaust memory.

Isn't this the purpose, though, of passing ```true``` for the ```trackAllValues``` to the ThreadLocal's ctor?
http://blogs.msdn.com/b/pfxteam/archive/2011/11/11/10236000.aspx
If the test is failing because a value is missing due to a thread going away, that seems like it'd be a bug in the actual product rather than in the test.

> The Tasks created in the "for" loop were capturing the iteration variable, not the particular value for each iteration. So the ThreadLocal values were not getting set correctly.

The body of that loop looks like this:
```C#
var task = Task.Factory.StartNew(() => threadLocal.Value = i, CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default);
task.Wait();
```
Doesn't the presence of that Wait mean that the thread local value would actually be correct, since ```i``` wouldn't be updated until after usage of it for that iteration completed?

stephentoub:
@dotnet-bot test this please (#3748, #3749)

ericeil:
@stephentoub, you're right that the variable capture was correct prior to my change.  

I guess I'll need to look into the intended behavior of trackAllValues a little more.  I'll close this PR for now.

PR #3751, opened 2015-10-09
Use new visitor class for EnumerableQuery expressions.

### LINKED ISSUE #3607: QueryEnumerable only accepts expression features prior to Framework version 4.0
Any expression type introduced after 3.5 will fail with an `ArgumentException`:

```C#
[Fact]
public void Block()
{
    Expression block = Expression.Block(
        Expression.Empty(),
        Expression.Constant(Enumerable.Range(0, 2).AsQueryable())
        );
    IQueryable<int> q = _prov.CreateQuery<int>(block);
    Assert.Equal(Enumerable.Range(0, 2), q); // "Unhandled Expression Type: Block"
}
```

Expressions that existed with 3.5, but using features from 4.0 will also fail with an `ArgumentException`:

```C#
[Fact]
public void ExplicitlyTypedConditional()
{
    Expression call = Expression.Call(
        typeof(Queryable),
        "OfType",
        new[] { typeof(long) },
        Expression.Condition(
            Expression.Constant(true),
            Expression.Constant(new long?[] { 2, 3, null, 1 }.AsQueryable()),
            Expression.Constant(Enumerable.Range(0, 3).AsQueryable().Select(i => (long)i)),
            typeof(IQueryable)
            )
        );
    IQueryable<long> q = _prov.CreateQuery<long>(call);
    Assert.Equal(new long[] { 2, 3, 1 }, q); // "Argument types do not match"
}

### PR DESCRIPTION
**Pull request by JonHanna**
Fixes #3607

Have EnumerableRewriter inherit from ExpressionVisitor, hence allowing those
Expression classes introduced after 3.5 to be used. Also handle explicit
typing of those expressions that can be explicitly typed.

### CODE REVIEW

**Comment by stephentoub** [comment]
cc: @VSadov

**Comment by JonHanna** [comment]

**Comment by JonHanna** [comment]

**Comment by stephentoub** [comment]

File: src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs [hunk 1]
```diff
@@ -4,21 +4,27 @@
...
 
 namespace System.Linq
 {
-    internal class EnumerableRewriter : OldExpressionVisitor
+    internal class EnumerableRewriter : ExpressionVisitor
     {
-        internal EnumerableRewriter()
+        // We must ensure that if a LabelTarget is rewritten that it is always rewritten to the same new target
+        // or otherwise expressions using it won't match correctly.
+        private Dictionary<LabelTarget, LabelTarget> _targetCache = new Dictionary<LabelTarget, LabelTarget>();
```
stephentoub: readonly?

src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs
```diff
@@ -4,21 +4,27 @@
...
+        private Dictionary<Type, Type> _equivalentTypeCache = new Dictionary<Type, Type>
```
stephentoub: readonly?

src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs
```diff
@@ -4,21 +4,27 @@
...
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
 
 namespace System.Linq
 {
-    internal class EnumerableRewriter : OldExpressionVisitor
+    internal class EnumerableRewriter : ExpressionVisitor
```
stephentoub: How often / in what situations are EnumerableRewriters constructed?  This is adding two dictionaries to every construction.

stephentoub:
@VSadov, @JonHanna, what's the status of this PR?  I couldn't tell from https://github.com/dotnet/corefx/issues/3607 whether this is something that @VSadov actually wants changed.

src/System.Linq.Queryable/src/System/Linq/EnumerableRewriter.cs
JonHanna: If you call `Execute` or `Execute<TResult>` on the linq-to-objects `IQueryProvider` implementation that will cause it to be created once per call. If you enumerate a linq-to-objects query then it is called once and the resulting `IEnumerable<T>` is cached.

The target dictionary is necessary for any rewriting of targets, as they must be replaced consistently (each reference of the same instance either not changed or always replaced with the same new instance). The type cache is memoising what can be relatively expensive. Both of them could perhaps be lazy-constructed though, and probably should be. In particular while the target dictionary is vital when touched on, it isn't touched on at all by the majority of cases (all currently working cases don't use it, as the old version doesn't support label expressions at all).

VSadov:
I looked around for any info on why OldExpressionVisitor is used here, but could not find much. I have two theories - It could have been to preserve some parity with old query providers, to avoid cases where something runs in IL and does not translate to SQL for example, but that is a possibility anyways (not everything can be translated to SQL), so this would not be a strong reason.

The other possibility is that new visitor is more complex because of double dispatch. In particular it would likely be somewhat slower and have noticeably deeper stack depth when visiting same expressions. If there were real examples where switching to new visitor would cause perf degrades or stack overflows, I can see how keeping the old visitor would be a solution for robustness/perf issues.

Perhaps it makes sense to use new visitor, but override Visit to skip Accept for nodes that were known to the old visitor?

JonHanna:
The first of those possibilities would have the flaw that applying the policy to its logical conclusion would mean also never updating any other provider to ensure parity in all directions, and hence never having any use of the new functionality at all. That's not to say it wasn't the idea in mind.

A big possibility is that for a while it simply wouldn't have worked—the new class has an exposed virtual generic method, and virtual generic methods weren't working—though that isn't necessarily the case (which came when?).

It could also have been simply to reduce the impact of the change to `ExpressionVisitor`. This PR isn't rocket science, but it isn't trivial either, so the simpler approach of reusing the old version makes some sense in that regard.

> Perhaps it makes sense to use new visitor, but override Visit to skip Accept for nodes that were known to the old visitor?

The change to `ExpressionVisitor` took on the cost of dispatching via a virtual method that calls back to that visitor, but doing so gained (as well as flexibility and easier future extension) the loss of a lookup based on another virtual member to dispatch internally. At first blush at least, testing type in an override of `Visit()` seems to me to contain the worse of both worlds, unless I'm missing something.

In any case, I'm going to go ahead with the idea of making the label cache lazy-constructed for now, as that is definitely a clear win on what I have here so far.

VSadov:
@JonHanna i was just theorizing on purpose of keeping the old visitor around. Virtual generic methods were always functional on desktop CLR, but they could have been slow. They are not very fast even now, comparatively. Extra stack frame could also be noticeable.

Anyways, the change looks good.

Issue #3750, opened 2015-10-09
Expression interpreter - NegateChecked doesn't throw OverflowException


bartdesmet:
Repro case:

`Expression.NegateChecked(Expression.Constant(short.MinValue))`

Issue seems to be that the checked(...) expression does not incorporate the conversion back after performing the negation.

I'll have a look at fixing it.

PR #3754, opened 2015-10-09
Fixing expression interpreter for NegateChecked overflow behavior

bartdesmet:
This fixes issue #3750.

cc @VSadov


bartdesmet:
Note this also removes a few Negate instructions for arithmetic types that are not supported through the factory methods anyway.

VSadov:
LGTM.

VSadov:
looks like this needs rebasing

bartdesmet:
Should be good to go now.

PR #3756, opened 2015-10-09
Shim ASN1 native calls to OpenSSL.

eerhardt:
Added shims for all native calls in Interop.ASN1.cs.

@nguerrera, @bartonjs, @sokket


sokket:
LGTM!

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void Asn1StringFree(IntPtr o);
+
+        internal static string OBJ_obj2txt_helper(IntPtr asn1ObjectPtr)
+        {
+            // OBJ_obj2txt returns the number of bytes that should have been in the answer, but it does not accept
+            // a NULL buffer.  The documentation says "A buffer length of 80 should be more than enough to handle
+            // any OID encountered in practice", so start with a buffer of size 80, and try again if required.
+            // Therefore, 512 should be quite sufficient.
```
stephentoub: I don't see 512 used anywhere.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
```diff
@@ -174,7 +174,10 @@ public unsafe void DecodeX509KeyUsageExtension(byte[] encoded, out X509KeyUsageF
                 if (data->pathlen != IntPtr.Zero)
                 {
                     hasPathLengthConstraint = true;
-                    pathLengthConstraint = Interop.libcrypto.ASN1_INTEGER_get(data->pathlen).ToInt32();
+                    long pathLengthConstraint64 = Interop.Crypto.Asn1IntegerGet(data->pathlen);
+
+                    Debug.Assert(pathLengthConstraint64 < int.MaxValue, "pathLengthConstraint needs to be in the Int32 range.");
```
stephentoub: Should the shim return an int32_t then?  Or should this be an exception if it's possible it actually could be larger than  Int32.MaxValue?

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -0,0 +1,112 @@
...
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void Asn1OctetStringFree(IntPtr o);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void Asn1StringFree(IntPtr o);
+
+        internal static string OBJ_obj2txt_helper(IntPtr asn1ObjectPtr)
+        {
+            // OBJ_obj2txt returns the number of bytes that should have been in the answer, but it does not accept
+            // a NULL buffer.  The documentation says "A buffer length of 80 should be more than enough to handle
```
stephentoub: Below there's a comment about bytesNeeded not included the null terminator... does the 80?

stephentoub:
Compilation failure on OS X:
```
/Users/dotnet-bot/j/workspace/dotnet_corefx_mac_nativecomp_release_prtest/src/Native/System.Security.Cryptography.Native/pal_asn1.cpp:15:12: error: no matching function for call to 'ASN1_INTEGER_get'
    return ASN1_INTEGER_get(a);
           ^~~~~~~~~~~~~~~~
/usr/include/openssl/asn1.h:900:6: note: candidate function not viable: 1st argument ('const ASN1_INTEGER *' (aka 'const asn1_string_st *')) would lose const qualifier
long ASN1_INTEGER_get(ASN1_INTEGER *a) DEPRECATED_IN_MAC_OS_X_VERSION_10_7_AND_LATER;
     ^
1 error generated.
```

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
bartonjs: Yeah, that stale comment is my bad.

eerhardt: I'll update the whole thing.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -0,0 +1,112 @@
...
+        [DllImport(Libraries.CryptoNative, CharSet = CharSet.Ansi)]
+        private static extern int ObjObj2Txt([Out] StringBuilder buf, int buf_len, IntPtr a);
+
+        [DllImport(Libraries.CryptoNative, CharSet = CharSet.Ansi)]
+        internal static extern IntPtr GetFriendlyNameSharedObject(string friendlyName);
+
+        [DllImport(Libraries.CryptoNative, CharSet = CharSet.Ansi)]
+        internal static extern int ObjSn2Nid(string sn);
+
+        // Returns shared pointers, should not be tracked as a SafeHandle.
```
bartonjs: I've since learned the art of non-freeing SafeHandle types, so maybe this comment is a bit outdated.  But that's not something that needs to change right now.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    return d2i_ASN1_BIT_STRING(zero, ppin, len);
+}
+
+extern "C" void Asn1BitStringFree(ASN1_STRING* a)
+{
+    ASN1_BIT_STRING_free(a);
+}
+
+extern "C" ASN1_OCTET_STRING* D2IAsn1OctetString(ASN1_OCTET_STRING** zero, const unsigned char** ppin, int32_t len)
```
bartonjs: The `char**` part of d2i is a pain in managed code.  Instead of direct shimming here, could we try to move the second \* into the native layer?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
eerhardt: Since ASN1_INTEGER_get returns a 'long' and the shim is a direct shim, it feels more appropriate for the shim to return an Int64.  In this case, "pathlen" is pretty much guaranteed to fit inside an Int32, which makes the assert and cast valid, IMO.

The cast will throw an exception right? It wouldn't be the best exception possible, but it still would fail if there was ever a case where path length could be greater than 2 billion.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    return OBJ_txt2obj(s, true);
+}
+
+extern "C" int32_t ObjObj2Txt(char* buf, int32_t buf_len, const ASN1_OBJECT* a)
+{
+    return OBJ_obj2txt(buf, buf_len, a, true);
+}
+
+extern "C" const ASN1_OBJECT* GetFriendlyNameSharedObject(const char* friendlyName)
```
bartonjs: `GetObjectDefinitionByName`, maybe?  This name feels like it's getting the friendly name out, vs using the friendly name as the input.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -0,0 +1,112 @@
...
+        [return: MarshalAs(UnmanagedType.Bool)]
+        internal static extern bool Asn1OctetStringSet(SafeAsn1OctetStringHandle o, byte[] d, int len);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void Asn1OctetStringFree(IntPtr o);
+
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void Asn1StringFree(IntPtr o);
+
+        internal static string OBJ_obj2txt_helper(IntPtr asn1ObjectPtr)
```
bartonjs: I did a terrible job of naming this; and since it's not directly calling OBJ_obj2txt anymore, maybe the name should be spruced up.

GetOidValue, perhaps?

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
eerhardt: I was going to clean up all the d2i string functions once they are all shimmed. The problem is that I was going file-by-file and not all the methods passed into `Interop.libcrypto.OpenSslD2I` are shimmed yet.
So I was going to direct shim these until all the d2i functions were done, and then remove the "zero" parameter from the shim and move the second * into native.  Does that sound OK?

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
eerhardt: Done.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
eerhardt: Done.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
bartonjs: I don't know if our default is checked, or unchecked, casts.

If we really want it to throw, making it an explicitly checked cast is good.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
bartonjs: It feels like they could be done just one at a time for incremental goodness:

* d2i_ASN1_OCTET_STRING => DecodeAsn1OctetString(const unsigned char* buf, int32_t len)
* Interop.libcrypto.OpenSslD2I(Interop.libcrypto.d2i_ASN1_OCTET_STRING, data) => Interop.Crypto.DecodeAsn1OctetString(data)

But if it feels easier to tackle in phases, I'll take your word on it.  (Since it's always easier to critique than to do :)).

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
nguerrera: It won't throw. We have unchecked everywhere. I want to make debug checked, but there's a lot of test failures when I tried it. If we want to throw always, we should use explicit 'checked'.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pal_asn1.h"
+
+#include <openssl/objects.h>
+
+static_assert(NID_undef == 0, "Managed code relies on NID_undef == 0");
+static_assert(NID_secp224r1 == 713, "Managed code relies on NID_secp224r1 == 713");
```
nguerrera: I would put PAL_NID_* constants in the .h. The idea should be that by reading the .h, you get everything you need to write the interop in C#. And then assert vs. the PAL_* constant.

nguerrera:
LGTM, just a two very small comments.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
eerhardt: So the idea is to basically move the whole OpenSslD2I function from managed into native? I didn't think about that.

bartonjs: Yeah, it feels like the native would be
```C++
extern "C" ASN1_OCTET_STRING* DecodeAsn1OctetString(const unsigned char* buf, int32_t len)
{
    if (!buf || !len)
    {
        return nullptr;
    }

    return d2i_ASN1_OCTET_STRING(nullptr, &buf, len);
}
```

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
eerhardt: We make a buffer of size 80 and try it.  If the resulting string is exactly 80 characters long, the `if (bytesNeeded >= buf.Capacity)` check will kick in, and we will make a new buffer of 81 characters long, where the null terminator will fit.  So we should be fine with the existing code.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
eerhardt: I've added 'checked' explicitly.

src/Native/System.Security.Cryptography.Native/pal_asn1.cpp
eerhardt: I'm going to do these all in 1 shot in my next change.

eerhardt:
@dotnet-bot test this please

stephentoub:
@dotnet-bot test this please

stephentoub:

PR #3761, opened 2015-10-09
Fixing the build.sh script to cause a native build to automatically format the C++ source code

sokket:
Fixing the build.sh script to cause a native build to automatically format the C++ source code

@stephentoub  @weshaggard

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -141,8 +141,12 @@ build_managed_corefx()
 
 build_native_corefx()
 {
+    # Format the Native Code to make sure it is format-compliant
+    echo "Formatting native source code..."
+    "$__nativeroot/format-code.sh"
+    echo "Done"
+    
     # All set to commence the build
-
     echo "Commencing build of corefx native components for $__BuildOS.$__BuildArch.$__BuildType"
     cd "$__IntermediatesDir"
 


sokket:
The two changes are adding the CS file for the build script and the build.sh file, everything is a change from the format script running and bringing the C++ code into style-compliance

stephentoub:
@dotnet-bot test this please (failure due to System.Native.so appearing to be out-of-date with respect to the corefx lib using it)

build.sh
```diff
@@ -221,20 +225,7 @@ esac
...
-    OSX)
-        __monoroot=/Library/Frameworks/Mono.framework/Versions/Current
-        ;;
-    *)
-        __monoroot=/usr
-        ;;
-esac
-
-__referenceassemblyroot=$__monoroot/lib/mono/xbuild-frameworks
+__referenceassemblyroot="$(csharp FindMonoXBuildFrameworks.cs)"
```
stephentoub: Why use C# for this?  If we can do it in script instead, seems better to keep it simple.

src/Native/Common/pal_utilities.h
```diff
@@ -61,8 +61,8 @@ inline typename std::make_unsigned<T>::type UnsignedCast(T value)
  * MSDN blogs: http://blogs.msdn.com/b/the1/archive/2004/05/07/128242.aspx
  */
 template <typename T, size_t N>
-char ( &_ArraySizeHelper( T (&array)[N] ))[N];
-#define ARRAY_SIZE( array ) (sizeof( _ArraySizeHelper( array ) ))
+char(&_ArraySizeHelper(T(&array)[N]))[N];
```
stephentoub: Nit: I'd suggest doing these formatting changes in a separate commit from the build script changes.

build.sh
akoeplinger: "Hardcoding" the location of the system-level Mono was actually intentional. The reason is that only the Mono from Xamarin packages ship with the PCL reference assemblies, so e.g. if I have a custom built Mono in my PATH I still want the corefx build to use those PCL assemblies from the regular Mono install. This change would break that workflow.

I'd prefer having a way to override `__referenceassemblyroot` instead.

src/Native/Common/pal_utilities.h
akoeplinger: +1

build.sh
sokket: @stephentoub I could not find a way to determine the GAC location reliably via regular scripting. Mono has some environment variables that *might* be there aren't guaranteed.

I understand the use of hardcoding the path, but the problem is that this path isn't guaranteed; this will pick up the changes to whatever Mono the user is using by default since it uses the same algorithm that the terminal uses to find what the user means by typing ```mono```. I think it would be more confusing to have this hardcoded and require everyone to have the exact same configuration rather than follow the user preference and functionality set by the path; ie when you run ```mcs``` or ```mono``` it will use a certain PCL and we will adhere to that. It seems like if you have multiple versions installed and this path hardcoded, it becomes a debugging issue to figure out which version CoreFX used and which version your test program used; I'm not tied to either way of doing this, I was simply trying to make it fit with how ```which``` works and how the system will find the PCL.

akoeplinger: @sokket well, the thing is you can't get the PCL assemblies unless you install the package from the Xamarin repository, which always installs into the paths hardcoded above (in theory you could get the ZIP file from somewhere and unpack it to the right location, but I doubt the average developer knows where to put it).

A **much** better solution would be to just ship the PCL assemblies in a NuGet package (if it's possible to wire that into MSBuild somehow).

sokket: Gotcha. I'll revert this part since it seems unnecessary / controversial / not how we want to do it long-term anyway. Thanks for the feedback guys :)

src/Native/Common/pal_utilities.h
sokket: Done, PR #3782 is fixing the formatting issues and this will just be adding the line to run the format script for each build

sokket:
Fixed to only add the format-code.sh call in the build.sh script with no other changes

build.sh
```diff
@@ -141,8 +141,12 @@ build_managed_corefx()
 
 build_native_corefx()
 {
+    # Format the Native Code to make sure it is format-compliant
+    echo "Formatting native source code..."
+    "$__nativeroot/format-code.sh"
```
nguerrera: Won't this eliminate incremental compilation or is the format script smart enough to only affect the modification time of a file if there is an actual change?

nguerrera:
Please update the PR name and description to match the actual commit for future reference.

build.sh
sokket: If you run format-code.sh and clang-format doesn't find any issue with a file, the file will not be changed and it does not show up in git status or change the compilation

sokket:
Done

sokket:
Any further comments or can I merge? :)

build.sh
nguerrera: You're right, just did a test and modification times are not altered and incremental build show up-to-date. :)

nguerrera:
I'm not crazy about formatting during a build, particularly any official build. I would consider having a strict mode that triggers a build error if there would be format changes and have CI and official builds use that. I would rather reject PRs with bad formatting than silently build from something that isn't identical to the source tree.

sokket:
There were a couple questions raised about potential issues down the road, so closing this until we can come up with a good long-term solution. Opened issue #3789 to track

weshaggard:
While I like the idea of auto-formatting I agree with @nguerrera that I'm not crazy about it happening as part of the build. I think of as something we get folks used to before committing like we tell folks to squash before submitting a PR.

PR #3762, opened 2015-10-09
Consume build tools 00106, which adds support for auto test packages

FiveTimesTheFun:
Consume build tools 00106, which adds support for automatically
generating test nuget packages.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,12 +10,13 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00104</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00106</BuildToolsVersion>
     <DnxVersion>1.0.0-beta7</DnxVersion>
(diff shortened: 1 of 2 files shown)


FiveTimesTheFun:
Goes with https://github.com/dotnet/buildtools/pull/307

dir.props
```diff
@@ -41,6 +42,7 @@
     <!-- Input Directories -->
     <PackagesDir Condition="'$(PackagesDir)'==''">$(ProjectDir)packages/</PackagesDir>
     <ToolsDir Condition="'$(ToolsDir)'==''">$(PackagesDir)Microsoft.DotNet.BuildTools.$(BuildToolsVersion)/lib/</ToolsDir>
+    <NuProjDir Condition="'$(NuProjDir)'==''">$(PackagesDir)NuProj.$(NuProjVersion)/tools/</NuProjDir>
```
weshaggard: I see this is where NuProjDir is getting set. I'm not sure if I like the fact this this is required to be setup and restored by all the consuming repo's of BuildTools. We should probably include the library directly in the BuildTools package if we want to depend on it.

FiveTimesTheFun: Thanks Wes, will reply over at https://github.com/dotnet/buildtools/pull/307.

stephentoub:
Closing this out, as buildtools has already progressed past 106.

PR #3764, opened 2015-10-10
Update and addition to Performance tests

ianhays:
- Made the existing perf tests for Linq follow the new syntax and use the new runner
- Removed all references to v5.0 libraries being required as they no longer are
- Added new tests using usage data from aspnet/benchmarks
- Modified some existing tests to be more easily comparable using analysis tools that look at inlinedata/memberdata (e.g. System.IO.Compression)
- Updated the lock files to use the latest xunit-performance package
- Modified perf tests for System.Diagnostics.Process to provide more consistent results.


src/Common/tests/System.Net/Sockets/Performance/SocketPerformanceTests.cs
```diff
@@ -8,7 +8,7 @@
 using Xunit;
```
ianhays: The SocketPerformanceTest changes shouldnt be in here. I'm going to put those tests in a separate PR shortly.

src/System.Globalization/tests/Performance/Perf.StringCultureInfo.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
ianhays: This file won't be merged, but I did have a question on it. My builds for it were failing with the notice that the overloads I'm trying to use (ones that take a CultureInfo) didn't exist. Any ideas?

ianhays:
@stephentoub

ianhays:
Depends on https://github.com/dotnet/buildtools/pull/308

src/System.Collections/tests/Performance/Perf.List.cs
```diff
@@ -186,6 +186,21 @@ public void Enumerator(int size)
...
         [InlineData(10000)]
         [InlineData(100000)]
+        public void SetCapacity(int size)
+        {
+            List<object> list = CreateList(size);
+            foreach (var iteration in Benchmark.Iterations)
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < 10000; i++)
+                    {
+                        list.Capacity = size + 1;
```
stephentoub: This is measuring the performance of Capacity when it doesn't actually need to change the size of the list, since it's always being set to the same value.  Is that intentional?

src/System.Globalization/tests/Performance/Perf.StringCultureInfo.cs
stephentoub: They don't exist in the .NET Core surface area:
https://github.com/dotnet/corefx/blob/master/src/System.Runtime/ref/System.Runtime.cs#L1588-L1589

ianhays: I see. I was looking at https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs to determine if they were present, so that makes sense. Will remove them.

src/System.Collections/tests/Performance/Perf.List.cs
ianhays: The docs [here](https://msdn.microsoft.com/en-us/library/y52x03h2(v=vs.110).aspx) only mention that setting the capacity is O(n) where n is the new size. They did not mention the special case of if the new capacity == the old capacity, but thinking on it now there's no reason to reallocate a new internal array if the desired size is already met with the old array. I'll modify this test to set the capacity to a different value each iteration.

src/System.Console/tests/Performance/Perf.Console.cs
```diff
@@ -85,5 +84,77 @@ public void OpenStandardError()
...
+            }
+            Console.BackgroundColor = beforeTest;
+        }
+
+        [Benchmark]
+        [PlatformSpecific(PlatformID.Windows)]
+        public void ResetColor()
+        {
+            ConsoleColor beforeTestBackground = Console.BackgroundColor; // In case it isn't set to the default.
+            ConsoleColor beforeTestForeground = Console.ForegroundColor; // In case it isn't set to the default.
```
stephentoub: This is going to fail on Unix.

src/System.Console/tests/Performance/Perf.Console.cs
```diff
@@ -85,5 +84,77 @@ public void OpenStandardError()
                     s.Dispose();
             }
         }
+
+        [Benchmark]
+        [PlatformSpecific(PlatformID.Windows)]
```
stephentoub: You've made these platform specific because of the call to get_ForegroundColor/get_BackgroundColor?  How about just using ConsoleColor.ResetColor() at the end instead?

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -10,75 +10,70 @@ namespace System.IO.Compression.Tests
...
-                    for (int i = 0; i < 1000; i++)
+            byte[] bytes;
+            const int fileSize = 100000000;
+            PerfUtils utils = new PerfUtils();
+            string filePath = utils.GetTestFilePath() + ".gz";
+            switch (type)
+            {    
+                case CompressionType.CryptoRandom:
+                    bytes = new byte[fileSize];
+                    var rand = RandomNumberGenerator.Create();
```
stephentoub: Technically this should be Dispose'd of when you're done with it.

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -10,75 +10,70 @@ namespace System.IO.Compression.Tests
...
+                    rand.GetBytes(bytes);
+                    using (FileStream output = File.Create(filePath))
+                    using (GZipStream zip = new GZipStream(output, CompressionMode.Compress))
+                        zip.Write(bytes, 0, bytes.Length);
+                    break;
+                case CompressionType.RepeatedSegments:
+                    bytes = Text.Encoding.UTF8.GetBytes(utils.CreateString(fileSize / 1000));
+                    using (FileStream output = File.Create(filePath))
+                    using (GZipStream zip = new GZipStream(output, CompressionMode.Compress))
+                        for (int i = 0; i < 1000; i++)
```
stephentoub: I don't remember what the compression window size is, but you might want to check to see whether the algorithm will actually see these as repeated segments, or whether the segments are too long for it to see the pattern.

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -10,75 +10,70 @@ namespace System.IO.Compression.Tests
...
-
-                // Highly repeated data
-                {
-                    byte[] bytes = new byte[101000000];
+                case CompressionType.CryptoRandom:
+                    bytes = new byte[100000000];
+                    RandomNumberGenerator.Create().GetBytes(bytes);
+                    break;
+                case CompressionType.RepeatedSegments:
+                    bytes = new byte[101000000];
```
stephentoub: Why is this size different from the others?

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
```diff
@@ -10,75 +10,70 @@ namespace System.IO.Compression.Tests
...
-                    rand.GetBytes(bytes);
-                    _byteArraysToCompress.Add(new object[] { bytes });
-                }
-
-                // Highly repeated data
-                {
-                    byte[] bytes = new byte[101000000];
+                case CompressionType.CryptoRandom:
+                    bytes = new byte[100000000];
+                    RandomNumberGenerator.Create().GetBytes(bytes);
```
stephentoub: Should be disposed.

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
```diff
@@ -0,0 +1,62 @@
...
+        public void Write(int size)
+        {
+            string filePath = GetTestFilePath();
+            byte[] bytes = CreateBytesToWrite(size);
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (FileStream writer = File.Create(filePath))
+                using (iteration.StartMeasurement())
+                {
+                        writer.Write(bytes, 0, size);
```
stephentoub: All of the sizes are larger than FileStream's default buffer capacity, so this test is skipping optimizations FileStream has to make Write's faster.  That's fine, but you might want to test the optmizations as well.  Often it's good to have perf tests that "lock in" optimizations.

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
```diff
@@ -0,0 +1,62 @@
...
+        public void Read(int size)
+        {
+            string filePath = CreateFile(size);
+            byte[] bytes = new byte[size];
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (FileStream reader = File.OpenRead(filePath))
+                using (iteration.StartMeasurement())
+                {
+                    reader.Read(bytes, 0, size);
```
stephentoub: Ditto.

src/System.IO.MemoryMappedFiles/tests/Performance/Perf.MemoryMappedFile.cs
```diff
@@ -0,0 +1,56 @@
...
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                MemoryMappedFile f1, f2, f3, f4, f5;
+                using (iteration.StartMeasurement())
+                {
+                    f1 = MemoryMappedFile.CreateNew("newTestMap_1", capacity);
+                    f2 = MemoryMappedFile.CreateNew("newTestMap_2", capacity);
+                    f3 = MemoryMappedFile.CreateNew("newTestMap_3", capacity);
+                    f4 = MemoryMappedFile.CreateNew("newTestMap_4", capacity);
+                    f5 = MemoryMappedFile.CreateNew("newTestMap_5", capacity);
```
stephentoub: This is going to fail on Unix.  Names are not supported.  It'd be better to just pass null for the name, unless you actually want to test named maps, in which case that could be a separate, Windows-specific test.

src/System.IO.MemoryMappedFiles/tests/Performance/Perf.MemoryMappedFile.cs
```diff
@@ -0,0 +1,56 @@
...
+        public void Dispose(int capacity)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                MemoryMappedFile f1, f2, f3, f4, f5;
+                f1 = MemoryMappedFile.CreateNew("newTestMap_1", capacity);
+                f2 = MemoryMappedFile.CreateNew("newTestMap_2", capacity);
+                f3 = MemoryMappedFile.CreateNew("newTestMap_3", capacity);
+                f4 = MemoryMappedFile.CreateNew("newTestMap_4", capacity);
+                f5 = MemoryMappedFile.CreateNew("newTestMap_5", capacity);
```
stephentoub: Ditto.

src/System.IO.MemoryMappedFiles/tests/Performance/Perf.MemoryMappedFile.cs
```diff
@@ -0,0 +1,56 @@
...
+
+                using (iteration.StartMeasurement())
+                {
+                    f1.Dispose(); f2.Dispose(); f3.Dispose(); f4.Dispose(); f5.Dispose();
+                }
+            }
+        }
+    }
+}
```
stephentoub: You might also want a test for CreateFromFile

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+            File.Delete(filePath);
+        }
+    }
+}
```
stephentoub: How about WriteAsync and ReadAsync?  Are those coming later?

src/System.IO.Pipes/tests/Performance/Perf_PipeTest.cs
```diff
@@ -0,0 +1,30 @@
...
+            PerfUtils utils = new PerfUtils();
+            byte[] sent = Text.Encoding.UTF8.GetBytes(utils.CreateString(size));
+            byte[] received = new byte[size];
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (ServerClientPair pair = CreateServerClientPair())
+                using (iteration.StartMeasurement())
+                {
+                    Task.Run(() => { pair.writeablePipe.Write(sent, 0, sent.Length); });
+                    pair.readablePipe.Read(received, 0, size);
```
stephentoub: We should wait for the Task after the Read.

src/System.Linq/tests/Performance/Perf.Linq.cs
```diff
@@ -0,0 +1,199 @@
...
+namespace System.Linq.Tests
+{
+    public class Perf_Linq
+    {
+        #region Helper Methods
+
+        /// <summary>
+        /// Provides TestInfo data to xunit performance tests
+        /// </summary>
+        public static IEnumerable<object[]> TestData()
```
stephentoub: Consider a more descriptive name, like IterationSizeWrapperData.

src/System.Numerics.Vectors/tests/Performance/Perf.Vector.cs
```diff
@@ -0,0 +1,79 @@
...
+                    {
+                        Vector2.Dot(two_1, two_2); Vector2.Dot(two_1, two_2); Vector2.Dot(two_1, two_2);
+                        Vector3.Dot(three_1, three_2); Vector3.Dot(three_1, three_2); Vector3.Dot(three_1, three_2);
+                        Vector4.Dot(four_1, four_2); Vector4.Dot(four_1, four_2); Vector4.Dot(four_1, four_2);
+                    }
+                }
+            }
+        }
+    }
+}
```
stephentoub: @mellinoe might want to suggest what else should be tested here.

stephentoub:
Left some comments, but generally looks good.

src/System.Numerics.Vectors/tests/Performance/Perf.Vector.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using Microsoft.Xunit.Performance;
+
+namespace System.Numerics.Tests
+{
+    // Equals, DotProduct, ctor
+    public class Perf_Vector
```
mellinoe: The most interesting things to test here are probably the methods under the "Intrinsics" files. Those are operations that are specially handled by the JIT. The most interesting ones are probably:

* Operator +, Operator -, Operator *, and the named versions, Add, Subtract, Multiply
* Dot()
* SquareRoot
* Length(Squared)
* Normalize
* Cross (for Vector3)
* Distance(Squared)

I'd also like to have them separated out by type (i.e. Vector2, Vector3, Vector4), because the code generated for them is probably different enough that we want to separate the results.

src/System.Console/tests/Performance/Perf.Console.cs
ianhays: I was concerned that doing ResetColor may overwrite the user-set console color with the default console color. For example if the user console was set to red but the program default is black, running the tests would change the console to black if we used ResetColor.

After looking at the code again it seems like the colorinfo used in ResetColor is actually set whenever we call the setters of Foreground/Background, so my concerns weren't valid. I'll modify these tests to use ResetColor and be xplat capable, thanks.

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
ianhays: I messed up the char->byte conversion and bandaided the problem instead of resolving it. Fixed.

src/System.IO.Compression/tests/Performance/Perf.DeflateStream.cs
ianhays: Good idea!

We initiate zlib with a windowBits of -15 which zlib interprets as a memory window size of of (2^15 - 262) = 32506 bytes. That should give enough leeway when writing 1000, but I'll add a test for 100 bytes at a time to see those results as well.

stephentoub: > size being written in the current test is 100000 not 1000, so it's not small enough for the window

:wink:

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
ianhays: Good idea. I think I'll rewrite these tests to cover some cases around the optimizations:

- Perform 1 read for full desired size. Skips the buffer entirely.
- Perform x reads each of size _bufferSize - 1 (4095 bytes). Uses the buffer but forces a read every time.
- Perform x reads each of size _bufferSize / 8 (512 bytes). Buffer is useful.

How do those sound?

stephentoub:

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
stephentoub: I'm not sure we need the 2nd case, but the 1st and 3rd sound good.  You might also consider one for ReadByte, which is essentially size == 1.

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
ianhays: Adding them now. How is this structure?

```
    if (useAsync)
    {
        // perform a number of writes of size bufferSize asynchronously, then waits for them all to complete
        for (int i = 0; i < innerIterations; i++)
            asyncCalls[i] = reader.ReadAsync(bytes, 0, bufferSize);
        Task.WaitAll(asyncCalls);
    }
    else
    {
        for (int i = 0; i < innerIterations; i++)
            reader.Read(bytes, 0, bufferSize);
    }
```
The results show the async versions slower in most cases, so I'm wondering if that structure is adding too much overhead in the WaitAll or something. I could await on every indiividual async call, but that just be the same test as the synchronous reads.

stephentoub: It usually makes sense that an asynchronous MethodAsync is slower than its synchronous Method counterpart, as the former usually needs to do more work to set up the asynchrony and manage it.  The benefits of the async methods are around scalability and end-to-end performance benefits to the all-up application.

As for the structure, two things:
1) I think it'd be better to structure the async one as:
```C#
await ReadAsyncCore(innerIterations, bytes, bufferSize);
...
private static async Task ReadAsyncCore(int innerIterations, byte[] bytes, int bufferSize)
{
    for (int i = 0; i < innerIterations; i++)
        await reader.ReadAsync(bytes, 0, bufferSize).ConfiguratAwait(false);
}
```
2) We should probably ensure that the FileStream is constructed with FileOptions.None in the sync case and FileOptions.Asynchronous in the asynchronous case.  Otherwise, on Windows at least, with FileOptions.Asynchronous set, Read just ends up wrapping ReadAsync, and with FileOptions.Onone, ReadAsync ends up wrapping Read.

ianhays: Done and done! I added a WriteCore/ReadCore with the same structure as WriteAsyncCore/ReadAsyncCore for consistency and ensured the streams were constructed with the proper attributes. Will include in next PR update.

src/System.IO.FileSystem/tests/Performance/Perf.FileStream.cs
ianhays: That sounds useful. Added.

ianhays:
@mellinoe added some new tests for Vector, separated into Vector2, 3, 4 like you wanted :)

@stephentoub responded to all of your feedbacks. Thanks for the great comments.

I still need to:
- remove sockets changes
- remove stringcultureinfo tests
- Update buildtools PR and get that merged

mellinoe:
@ianhays The vector tests look good overall to me. My only concern is that the test structure might have a non-negligible effect on the results. I think the method call to ExecuteTest and then the switch jump might add some overhead that could affect results, since the vector operations themselves often get JIT'd down to a single instruction, or a few instructions.

Anyways, I think they are fine to merge right now, the content of the tests looks good to me and should give accurate enough results. When I have some time I may go through them and implement them using T4 templates to avoid the two points of overhead above. That way, we could basically move all of the switching/method calling to code-generation time, and at run-time only execute the inner benchmark logic.

ianhays:
Thanks for the review and direction for the tests Eric!

>My only concern is that the test structure might have a non-negligible effect on the results.

The call and switch have no appreciable effect on the results when the number of innerIterations as high as I set it (1000000). The innerIterations loop, however, adds roughly .3 ms to the test time. For the Vector2 tests for Add_operator, sub_operator, mul_operator, Dot, SquareRoot, and Length_Squared this inner loop makes up the entirety of the test timing. The best solution to this would be a massive amount of loop unrolling, but I agree with you that templating may be a better answer. While accuracy is absolutely the primary concern, I don't want to sacrifice too much in succinctness if we can avoid it with these microbenchmarks.

Either way, the results for Linux vs Windows are far enough off that the switch overhead is a low priority for the time being. I'll post an info dump with the results today :)

ianhays:
Rebased and moved around a few numbers.

stephentoub:
LGTM. Thanks.

PR #3768, opened 2015-10-10
Refactoring System.Net.Security PAL

CIPop:
Refactoring System.Net.Security PAL in preparation to add back the Schannel Interop multi-interface SSPI object model required by NegotiateStream.
 
The PAL design is similar to the one @pgavlin designed for System.Net.Primitives, NameResolution and Sockets: partial static classes replaced the interface/class/factory pattern that was derived from the SSPI object model. Semantics should remain the same.

1. <os>\CertModule.cs was renamed to CertificateValidationPal.<os>.cs. CertWrapper and CertInterface have been removed.
2. <os>\SSPISecureChannelType.cs was renamed to SslStreamPal.<os>.cs. Modified GlobalSSPI, SSPIInterface have been removed temporarily. The plan is to add them back with the original semantics for NegotiateStream support via Win32 SSPI.
  * Small change in QueryContextChannelBinding to simplify the PAL.
  * Added a static readonly sync object in the Windows implementation to preserve interlocked behavior.
  * Removed Win32Exception from MapToSecurityStatus. (Still needs to be redesigned to avoid string parsing.)
3. Rest of \<os>\* have been moved to Interop. I did my best to integrate with the latest sources in master.
4. Removed NegotiateStream / Auth code. This will be added back with NegotiateStream support.

@bartonjs @vijaykota @rajansingh10 PTAL
/cc: @davidsh @stephentoub @pgavlin


src/System.Net.Security/src/System.Net.Security.csproj
```diff
@@ -178,6 +184,15 @@
...
       <Link>Interop\Unix\libssl\Interop.SafeSslHandle.cs</Link>
     </Compile>
+    <Compile Include="$(CommonPath)\Interop\Unix\libssl\SecuritySafeHandles.cs">
+      <Link>Interop\Unix\libssl\SecuritySafeHandles.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Unix\libssl\StreamSizes.cs">
+      <Link>Interop\Unix\libssl\StreamSizes.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Unix\libssl\SslConnectionInfo.cs">
+      <Link>Interop\Unix\libssl\SslConnectionInfo.cs</Link>
```
vijaykota: Note that these were deliberately not put into Interop\libssl earlier because they are specific to System.Net.Security.SslStream. There is a danger that due to dev carelessness, the non-Security specific code might start getting mixed with this (eg. adding mapping code in Interop.Openssl.cs). Will leave it to you if you still want to organize code like this.

src/System.Net.Security/src/System.Net.Security.csproj
```diff
@@ -154,17 +144,33 @@
...
+      <Link>Interop\Windows\SChannel\UnmanagedCertificateContext.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Windows\SChannel\SecurityPackageInfoClass.cs">
+      <Link>Interop\Windows\SChannel\SecurityPackageInfoClass.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Windows\SChannel\SslConnectionInfo.cs">
+      <Link>Interop\Windows\SChannel\SslConnectionInfo.cs</Link>
+    </Compile>
+    <Compile Include="$(CommonPath)\Interop\Windows\SChannel\SecSizes.cs">
+      <Link>Interop\Windows\SChannel\SecSizes.cs</Link>
```
vijaykota: Nit: SecurityPackageInfo and SecSizes are used in NegotiateStream as well so maybe they should go into a Secur32 folder

vijaykota:
LGTM

rajansingh10:
LGTM

src/System.Net.Security/src/System.Net.Security.csproj
CIPop: For the original Win32 implementation these are structs extracted from schannel.h and so they must be in the Windows\Interop. The requirement is to rename them to the original form in the .h include (not done in this PR).

This is only phase 1 of the refactoring required for the Nego work for Schannel. We can further refine this afterwards. My recommendation is to remove these from the common code (possibly replacing them with a PAL equivalent).

src/System.Net.Security/src/System.Net.Security.csproj
CIPop: Thanks! Moving to secur32. 
These are defined in sspi.h and are common to all the SSPI modules.

PR #3770, opened 2015-10-11
Fix inconsistent behavior of UnmanagedMemoryAccessor.ReadDecimal()

jamesqo:
Noticed a few discrepancies in behavior from the CoreCLR and CoreFX versions of `UnmanagedMemoryStream` recently:

- If the decimal read has an invalid value, the [mscorlib](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/IO/UnmanagedMemoryAccessor.cs#L311) version will throw because it calls the decimal constructor, which [checks](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Decimal.cs#L278) that the decimal read is valid. We don't.

- If the target machine is big endian, `ReadDecimal` will return a corrupt decimal value since the fields are arranged in reverse order.

This PR fixes those issues.

### DIFF
--- a/src/System.IO.UnmanagedMemoryStream/src/Resources/Strings.resx
+++ b/src/System.IO.UnmanagedMemoryStream/src/Resources/Strings.resx
@@ -198,4 +198,7 @@
   <data name="IO_StreamTooLong" xml:space="preserve">
     <value>Stream was too long.</value>
   </data>
+  <data name="Arg_BadDecimal" xml:space="preserve">
+    <value>Read an invalid decimal value from the buffer.</value>
+  </data>
 </root>
--- a/src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
+++ b/src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
@@ -401,10 +401,12 @@ namespace System.IO
         [System.Security.SecuritySafeCritical]  // auto-generated
         public Decimal ReadDecimal(Int64 position)
         {
+            const int ScaleMask = 0x00FF0000;
+            const int SignMask = unchecked((int)0x80000000);
+
             int sizeOfType = sizeof(Decimal);
             EnsureSafeToRead(position, sizeOfType);
 
-            Decimal d;
             unsafe
             {
                 byte* pointer = null;
@@ -418,11 +420,16 @@ namespace System.IO
                     int hi = UnsafeReadInt32(pointer + 8);
                     int flags = UnsafeReadInt32(pointer + 12);
 
-                    int* ptr = (int*)(&d);
-                    *ptr++ = flags;
-                    *ptr++ = hi;
-                    *ptr++ = lo;
-                    *ptr = mid;
+                    // Check for invalid Decimal values
+                    if (!((flags & ~(SignMask | ScaleMask)) == 0 && (flags & ScaleMask) <= (28 << 16)))
+                    {
+                        throw new ArgumentException(SR.Arg_BadDecimal); // Throw same Exception type as Decimal(int[]) ctor for compat
+                    }
+
+                    bool isNegative = (flags & SignMask) != 0;
+                    byte scale = (byte)(flags >> 16);
+
+                    return new decimal(lo, mid, hi, isNegative, scale);
                 }
                 finally
                 {
@@ -432,8 +439,6 @@ namespace System.IO
                     }
                 }
             }
-
-            return d;
         }
 
         /// <summary>
--- a/src/System.IO.UnmanagedMemoryStream/tests/UmaTests.cs
+++ b/src/System.IO.UnmanagedMemoryStream/tests/UmaTests.cs
@@ -40,6 +40,23 @@ namespace System.IO.Tests
         }
 
         [Fact]
+        public static void UmaInvalidReadDecimal()
+        {
+            const int capacity = 16; // sizeof(decimal)
+
+            using (var buffer = new TestSafeBuffer(capacity))
+            using (var uma = new UnmanagedMemoryAccessor(buffer, 0, capacity, FileAccess.ReadWrite))
+            {
+                // UMA should throw when reading bad decimal values (some bits of flags are reserved and must be 0)
+                uma.Write(0, 0); // lo
+                uma.Write(4, 0); // mid
+                uma.Write(8, 0); // hi
+                uma.Write(12, -1); // flags (all bits are set, so this should raise an exception)
+                Assert.Throws<ArgumentException>(() => uma.ReadDecimal(0)); // Should throw same exception as decimal(int[]) ctor for compat
+            }
+        }
+
+        [Fact]
         public static void UmaReadWrite()
         {
             const int capacity = 199;


stephentoub:
cc: @KrzysztofCwalina

src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
```diff
@@ -418,11 +421,29 @@ public Decimal ReadDecimal(Int64 position)
...
+                    {
+                        // Follow same order of fields declared in Decimal.cs
+                        *ptr++ = flags;
+                        *ptr++ = hi;
+                        *ptr++ = lo;
+                        *ptr = mid;
+                    }
+                    else
+                    {
+                        // Use reverse order
```
nguerrera: This doesn't seem right to me. Big endian machine would have each int's bytes in reverse order, but the four int fields would be in the same order.

nguerrera: And we don't have to swap the int bytes either, because everything in UnmanagedMemoryAccessor reads/writes in native endianness.

src/System.IO.UnmanagedMemoryStream/src/Resources/Strings.resx
```diff
@@ -198,4 +198,7 @@
   <data name="IO_StreamTooLong" xml:space="preserve">
     <value>Stream was too long.</value>
   </data>
+  <data name="Arg_DecBitCtor" xml:space="preserve">
+    <value>Decimal byte array constructor requires an array of length four containing valid decimal bytes.</value>
```
nguerrera: This message is not quite correct since we didn't call the byte array constructor.

src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
nguerrera: We should probably call this `Decimal(int lo, int mid, int hi, bool isNegative, byte scale)` ctor. We'd be just as efficient (no array alloc), not need unsafe code, and get the validation done by Decima instead of us. We'd still have to split up the flags in to sign and scale, but we do that anyway in your proposal to do the validation, so it still seems better.

stephentoub:

src/System.IO.UnmanagedMemoryStream/src/Resources/Strings.resx
jamesqo: Ah, hadn't realized that (was just copying over the old code from mscorlib). Fixed.

src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
jamesqo: Done! That was a pretty good idea TBH, I should have thought of that earlier.

Removed the `System.Runtime.Extensions` dependency as well.

src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
```diff
@@ -418,11 +420,16 @@ public Decimal ReadDecimal(Int64 position)
...
 
-                    int* ptr = (int*)(&d);
-                    *ptr++ = flags;
-                    *ptr++ = hi;
-                    *ptr++ = lo;
-                    *ptr = mid;
+                    // Check for invalid Decimal values
+                    if (!((flags & ~(SignMask | ScaleMask)) == 0 && (flags & ScaleMask) <= (28 << 16)))
+                    {
+                        throw new IOException(SR.IO_BadDecimal);
```
stephentoub: This is a different exception type than is thrown from the mscorlib version, right?  If the goal of this change is to bring the behaviors in sync, seems we should throw the same exception type.

stephentoub:
Doesn't look like we have much in the way of existing tests for ReadDecimal.  With a change like this, we really need some that a) validates the failure path, and b) ensures proper round-tripping of a variety of different decimal values through WriteDecimal/ReadDecimal.  Can you add some?

stephentoub:
@KrzysztofCwalina, thoughts on the PR?

@jamesqo, are you planning to add tests?

Thanks.

src/System.IO.UnmanagedMemoryStream/src/System/IO/UnmanagedMemoryAccessor.cs
KrzysztofCwalina: I agree.

jamesqo: @KrzysztofCwalina [Yep.](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Decimal.cs#L278)

jamesqo: Awesome, I'll change the type to `ArgumentException` and leave a comment why.

jamesqo:
@stephentoub Going to add tests, yep.

jamesqo:
Done.

KrzysztofCwalina:
Looks good!

src/System.IO.UnmanagedMemoryStream/tests/UmaTests.cs
```diff
@@ -23,6 +23,13 @@ public static void UmaInvalidReadWrite()
                 Assert.Throws<ArgumentOutOfRangeException>(() => uma.Write(capacity, 12345));
                 Assert.Throws<ArgumentException>(() => uma.Write(capacity - 1, 0.123));
 
+                // UMA should throw when reading bad decimal values (some bits of flags are reserved and must be 0)
+                uma.Write(0, 0); // lo
```
stephentoub: Did these tests pass for you locally?  They're null ref'ing.

jamesqo: [They did, actually](http://i.imgur.com/WhD0Us8.png)-- I don't know why they're failing with CI. Also, I checked line 838 (where Jenkins said the null ref happened) and there's no way an exception could have been thrown there:

```csharp
*((Int32*)pointer) = value;
```

I don't know, perhaps it's a problem with CI?

nguerrera: If pointer has a small value (including null of course!), that certainly could show up as NullReferenceException.

stephentoub: > they did, actually

That screen shot shows a build, but no tests as part of it. Try cd'ing into the tests directory and doing msbuild /t:rebuildandtest

> If pointer has a small value (including null of course!), 

Yes, and in fact note that the test is using
"FakeSafeBuffer", which IIRC is true to its name (it's fake) and doesn't actually have any memory associated with it. So attempting to write to it will blow up.

jamesqo: @stephentoub Yep, my bad. I forgot to run MSBuild from the tests directory, sorry about that.

As for the tests, I've fixed the null refs by using a `TestSafeBuffer` instead of a fake one.

stephentoub:
Thanks, @jamesqo.  LGTM.  Could you please squash this?  You could have two commits, one for the src changes and one for the new test, or just a single commit for everything.

jamesqo:
@stephentoub Sure, no problem.

stephentoub:
Thanks, @jamesqo.

Issue #3769, opened 2015-10-11
ECDsa Sign and Verify expect different data forms between Windows and Unix


bartonjs:
ECDSA's Sign operation, algorithmically, emits two values, `r` and `s`.

The values produced by Desktop, and CoreFX-Windows are equivalent to `r.Concat(s)`.

CoreFX-Unix is emitting (and expecting) the signature block:
```
   ECDSA-Sig-Value ::= SEQUENCE {
     r  INTEGER,
     s  INTEGER
   }
```
These values will never be equal, differing by the ASN/DER structural information.  Since Desktop is the de facto standard here, we need to make the Unix version match.

PR #3773, opened 2015-10-11
Implement GetECDsaPublicKey for Unix.

### LINKED ISSUE #3383: X509Certificate2.GetECDsaPublicKey() on Unix still unimplemented.
(from Jeremy)

As for GetECDsaPublicKey, I think that you can get there with EVP_PKEY* X509_PUBKEY_get(X509_PUBKEY*). Probably want to just add it to the native interop library as EVP_PKEY* GetX509EvpPublicKey(X509* x509). Then you can just unpack the EC_KEY the same way you did for the private key here.

bartonjs:
Unfortunately, the test for this was the first place where Windows ECDsa and Unix ECDsa were in the same place with the same key, and it was observed that they do not use the same data format for Sign/Verify.  This problem is tracked in issue #3769.

Fixes #3383.

cc: @AtsushiKan

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs
+++ b/src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.IO;
+using System.Text;
 using Test.Cryptography;
 using Xunit;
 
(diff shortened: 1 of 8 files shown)


stephentoub:
LGTM

stephentoub:

stephentoub:
@dotnet-bot test this please (#3785)

src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs
```diff
@@ -128,6 +129,56 @@ public static void TestKey_RSA()
...
+            }
+        }
+
+        [Fact]
+        [ActiveIssue(3769, PlatformID.AnyUnix)]
+        public static void TestECDsaPublicKey_ValidatesSignature()
+        {
+            byte[] existingSignature =
+            {
+                // r:
```
joshfree: where did we get this data blob from? could we add comment

stephentoub:
@dotnet-bot test this please ("Timeout cloning repo from GitHub")

stephentoub:
Compilation failure on Unix:
```
Internal\Cryptography\Pal.Unix\OpenSslX509Encoder.cs(180,21): error CS0103: The name 'Debug' does not exist in the current context [d:\j\workspace\dotnet_corefx_linux_debug_prtest\src\System.Security.Cryptography.X509Certificates\src\System.Security.Cryptography.X509Certificates.csproj]
```

src/System.Security.Cryptography.X509Certificates/tests/PublicKeyTests.cs
bartonjs: > where did we get this data blob from?

Attributed:
```
// This signature was produced as the output of ECDsaCng.SignData with the same key
// on .NET 4.6.  Ensure it is verified here as a data compatibility test.
//
// Note that since ECDSA signatures contain randomness as an input, this value is unlikely
// to be reproduced by another equivalent program.
```

bartonjs:
> Compilation failure on Unix

That's what I get for removing at-the-time unused using import statements :smile:.  Rebased, fixed, republished.

PR #3776, opened 2015-10-12
Portable PDB v1.0

tmat:
Finalizes Portable PDB v1.0:

1) Adds Document column to MethodBody table -- it stores the document of the first sequence point. Allows the consumer to construct Document to Method map w/o decoding all sequence points (which is potentially MBs of data)
2) Calculate PDB ID as a hash of the PDB content and store it to #Pdb stream.
3) Change version in headers to 1.0


tmat:
Disregard please.

PR #3781, opened 2015-10-12
Allow recursive Directory.Delete of paths with trailing slash

### LINKED ISSUE #3780: IOException on recursive Directory.Delete of path with trailing backslash
I observed this after we took a recent beta System.IO over in MSBuild.  Many of our tests create a directory and have a `finally` that deletes it, which suddenly started failing.  After debugging, I think this is a regression caused by 0721434, which added a call in [`RemoveDirectory`](https://github.com/dotnet/corefx/blob/0721434e19e7f9ecf8bf6dba7485f6c86827388c/src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs#L491) to canonicalize the path to extended form.  The following call then adds a backslash before calling `FindFirstFile`, but the extended path has different behavior in Win32: it no longer silently condenses double backslashes to a single path separator.

While proving to myself that I understood the bug I wrote a fix; I'll send a PR shortly.

rainersigwald:
Fixes #3780.

FindFirstFile with a normal path will happily collapse multiple
backslashes into a single logical path separator, but when presented
with a `\\?\` path, that doesn't happen.  As a result, unconditionally
concatenating a path-separator was fine until RemoveDirectoryHelper was
always called with EnsureExtendedPrefix (which is needed for extended
path support, introduced in 0721434).  With that, Win32 FindFirstFile
returns an error if there's a double backslash in the path beyond the
prefix.

@JeremyKuhne, are you a good person to review this?

### DIFF
--- a/src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs
+++ b/src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs
@@ -513,7 +513,7 @@ namespace System.IO
                 Interop.mincore.WIN32_FIND_DATA data = new Interop.mincore.WIN32_FIND_DATA();
 
                 // Open a Find handle
-                using (SafeFindHandle hnd = Interop.mincore.FindFirstFile(fullPath + PathHelpers.DirectorySeparatorCharAsString + "*", ref data))
+                using (SafeFindHandle hnd = Interop.mincore.FindFirstFile(Directory.EnsureTrailingDirectorySeparator(fullPath) + "*", ref data))
                 {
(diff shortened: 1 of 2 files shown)


JeremyKuhne:
@rainersigwald LGTM. I'll spend some more quality time with this to try and make sure there are no other regressions. Thanks!

JeremyKuhne:
@dotnet-bot test this please

stephentoub:

stephentoub:
LGTM. Thanks, @rainersigwald.

PR #3782, opened 2015-10-12
Fixing up the formatting of the native code

sokket:
Fixing up the formatting of the native code (everything under src/Native) to adhere to our style guidelines.
There aren't any functional changes in this PR, only formatting changes

/cc @stephentoub

### DIFF
--- a/src/Native/System.Native/pal_networking.cpp
+++ b/src/Native/System.Native/pal_networking.cpp
@@ -21,16 +21,13 @@ static_assert(PAL_TRY_AGAIN == TRY_AGAIN, "");
 static_assert(PAL_NO_RECOVERY == NO_RECOVERY, "");
 static_assert(PAL_NO_DATA == NO_DATA, "");
 static_assert(PAL_NO_ADDRESS == NO_ADDRESS, "");
-static_assert(sizeof(uint8_t) == sizeof(char), ""); // We make casts from uint8_t to char for OS functions, make sure it's legal
+static_assert(sizeof(uint8_t) == sizeof(char), ""); // We make casts from uint8_t to char so make sure it's legal
 
 static int IpStringToAddressHelper(const uint8_t* address, const uint8_t* port, bool isIPv6, addrinfo*& info)
 {
     assert(address != nullptr);
 
-    addrinfo hint = {
-        .ai_family = isIPv6 ? AF_INET6 : AF_INET,
-        .ai_flags = AI_NUMERICHOST | AI_NUMERICSERV
-    };
+    addrinfo hint = {.ai_family = isIPv6 ? AF_INET6 : AF_INET, .ai_flags = AI_NUMERICHOST | AI_NUMERICSERV};
 
     info = nullptr;
     return getaddrinfo(reinterpret_cast<const char*>(address), reinterpret_cast<const char*>(port), &hint, &info);
@@ -73,7 +70,7 @@ static void ConvertByteArrayToSockAddrIn(sockaddr_in& addr, const uint8_t* buffe
 static void ConvertSockAddrInToByteArray(uint8_t* buffer, int32_t bufferLength, const sockaddr_in& addr)
 {
     assert(bufferLength == NUM_BYTES_IN_IPV4_ADDRESS);
-    (void)bufferLength;  // Silence compiler warnings about unused variables on release mode
+    (void)bufferLength; // Silence compiler warnings about unused variables on release mode
 
     uint32_t* output = reinterpret_cast<uint32_t*>(buffer);
     *output = addr.sin_addr.s_addr; // Send back in network byte order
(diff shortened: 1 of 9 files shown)


src/Native/System.Native/pal_networking.cpp
```diff
@@ -21,16 +21,14 @@ static_assert(PAL_TRY_AGAIN == TRY_AGAIN, "");
 static_assert(PAL_NO_RECOVERY == NO_RECOVERY, "");
 static_assert(PAL_NO_DATA == NO_DATA, "");
 static_assert(PAL_NO_ADDRESS == NO_ADDRESS, "");
-static_assert(sizeof(uint8_t) == sizeof(char), ""); // We make casts from uint8_t to char for OS functions, make sure it's legal
+static_assert(sizeof(uint8_t) == sizeof(char),
+              ""); // We make casts from uint8_t to char for OS functions, make sure it's legal
```
stephentoub: Why did this move to its own line, because of the comment?  Not a big deal, but seems like this actually makes the code more difficult to read because this line ends up differing from all the ones before it.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -21,16 +21,14 @@ static_assert(PAL_TRY_AGAIN == TRY_AGAIN, "");
...
 
 static int IpStringToAddressHelper(const uint8_t* address, const uint8_t* port, bool isIPv6, addrinfo*& info)
 {
     assert(address != nullptr);
 
-    addrinfo hint = {
-        .ai_family = isIPv6 ? AF_INET6 : AF_INET,
-        .ai_flags = AI_NUMERICHOST | AI_NUMERICSERV
-    };
+    addrinfo hint = {.ai_family = isIPv6 ? AF_INET6 : AF_INET, .ai_flags = AI_NUMERICHOST | AI_NUMERICSERV};
```
stephentoub: I found the previous code easier to visually parse.  Is this new code really the desired format?

src/Native/System.Native/pal_networking.cpp
```diff
@@ -218,11 +209,7 @@ extern "C" int32_t GetHostEntryForName(const uint8_t* address, HostEntry* entry)
         return ConvertGetAddrInfoAndGetNameInfoErrorsToPal(result);
     }
 
-    *entry = {
-        .CanonicalName = nullptr,
-        .AddressListHandle = reinterpret_cast<void*>(info),
-        .IPAddressCount = 0
-    };
+    *entry = {.CanonicalName = nullptr, .AddressListHandle = reinterpret_cast<void*>(info), .IPAddressCount = 0};
```
stephentoub: Same comment as earlier... I realize this is being done by a tool following a rule, so I'll stop commenting on it... I'm just questioning the rule.

src/Native/System.Native/pal_networking.cpp
sokket: I believe this was moved due to the line-length limitation where the comment pushed the line over the character limit so it moved the argument and the comment to a new line

stephentoub:
cc: @nguerrera 

A few comments on changes that seem to make the code harder to read, but otherwise LGTM.

src/Native/System.Native/pal_networking.cpp
nguerrera: Agree is there an option for this by chance?

src/Native/System.Native/pal_networking.cpp
nguerrera: Shorten the comment maybe? 

FWIW, we don't really need this check. sizeof(uint8_t) == sizeof(char) == 1 is guaranteed by the standard.

nguerrera:
+1 to @stephentoub's minor comments, but LGTM otherwise.

src/Native/System.Native/pal_networking.cpp
sokket: I'm trying to keep this pack from having any code changes at all, so if we need to remove the check then that should be in a different pack...I don't want this to have any side effects. I can shorten the comment to keep it on one line, but I would really prefer to keep the check and not remove it in this pack

src/Native/System.Native/pal_networking.cpp
sokket: I believe it is this rule ```ConstructorInitializerAllOnOneLineOrOnePerLine``` which we have set to ```true``` to enforce rules around constructors. Initializer lists are old C-style things that are disabled via clang by default and throw a warning so we should think about if we want to use them or not. In either case, this pack is more about running the formatting tool to bring us into style-conformance rather than changing the style; I'd be more than happy to change the style if we so choose, but I think we should do that in a different pack. 

Thoughts?

sokket:
@stephentoub @nguerrera Fixed the comment to reduce line length and responded to comments. Any other thoughts? :)

src/Native/System.Native/pal_networking.cpp
nguerrera: OK.

src/Native/System.Native/pal_networking.cpp
mellinoe: Agreed as well, is there any way to just loosen the strictness of whichever rule did this?

sokket: From my initial investigation, it's all-or-nothing (see the comment below). We can definitely change this rule if we so choose, but let's chat about it in a separate issue and just get the code into current compliance; this PR and #3761 combined will make it so any changes we make to the formatting style will happen seamlessly, so it wouldn't be a problem to change this in the future

Issue #3494, opened 2015-09-28
ConnectV4MappedIPAddressToV4Host_Success failed in CI on Linux


stephentoub:
http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2785/console
```
09:31:53    System.Net.Sockets.Tests.DualMode.ConnectV4MappedIPAddressToV4Host_Success [FAIL]
09:31:53       System.Net.Sockets.SocketException : Unknown error 10054
09:31:53       Stack Trace:
09:31:54             at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
09:31:54             at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
09:31:54             at System.Net.Sockets.Socket.Connect(IPAddress address, Int32 port)
09:31:54             at System.Net.Sockets.Tests.DualMode.DualModeConnect_IPAddressToHost_Helper(IPAddress connectTo, IPAddress listenOn, Boolean dualModeServer, Int32 port)
09:31:54             at System.Net.Sockets.Tests.DualMode.ConnectV4MappedIPAddressToV4Host_Success()
```

stephentoub:
Also at http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2785/

stephentoub:
And the related ```ConnectV6IPEndPointToDualHost_Success``` failed at http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2783/

stephentoub:
Again here: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/2810/

stephentoub:
Again here: http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst/1128/console

I'm going to disable the test.

ericeil:
Error 10054 is "connection reset."  What all these cases seem to have in common is that they involve accepting a connection, and then not cleanly closing the newly-accepted socket.  Perhaps Linux is a little less forgiving of this in the client-side `connect` call, resulting in an error where Windows reports success.

One solution might be to explicitly call `Shutdown` and `Dispose` on the accepted Socket.  Another would be to set the accepted socket to "linger" before discarding it.  Between those, I might favor the "linger" option, as it avoids a dependency on the more complex Shutdown code.

[...]

PR #3791, opened 2015-10-12
Shim all the d2i_* functions in OpenSSL.

eerhardt:
Adding shims for all the d2i functions.

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/pal_x509.cpp
+++ b/src/Native/System.Security.Cryptography.Native/pal_x509.cpp
@@ -13,3 +13,23 @@ extern "C" EVP_PKEY* GetX509EvpPublicKey(X509* x509)
     // X509_get_X509_PUBKEY returns an interior pointer, so should not be freed
     return X509_PUBKEY_get(X509_get_X509_PUBKEY(x509));
 }
+
+extern "C" X509_CRL* DecodeX509Crl(const unsigned char* buf, int32_t len)
+{
+    if (!buf || !len)
+    {
+        return nullptr;
+    }
+
+    return d2i_X509_CRL(nullptr, &buf, len);
+}
+
+extern "C" X509* DecodeX509(const unsigned char* buf, int32_t len)
+{
+    if (!buf || !len)
+    {
+        return nullptr;
+    }
+
+    return d2i_X509(nullptr, &buf, len);
+}
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
@@ -20,12 +20,9 @@ namespace Internal.Cryptography.Pal
             _pkcs12Handle = pkcs12Handle;
         }
 
-        public unsafe static bool TryRead(byte[] data, out OpenSslPkcs12Reader pkcs12Reader)
+        public static bool TryRead(byte[] data, out OpenSslPkcs12Reader pkcs12Reader)
         {
-            SafePkcs12Handle handle = Interop.libcrypto.OpenSslD2I(
-                (ptr, b, i) => Interop.libcrypto.d2i_PKCS12(ptr, b, i),
-                data,
-                checkHandle: false);
+            SafePkcs12Handle handle = Interop.Crypto.DecodePkcs12(data, data.Length);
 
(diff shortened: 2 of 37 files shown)


src/Native/System.Security.Cryptography.Native/pal_x509.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pal_x509.h"
+
+extern "C" X509_NAME* DecodeX509Name(const unsigned char* buf, int32_t len)
```
bartonjs: Probably better belongs in pal_x509_name.cpp/h

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
```diff
@@ -22,10 +22,7 @@ private OpenSslPkcs12Reader(SafePkcs12Handle pkcs12Handle)
 
         public unsafe static bool TryRead(byte[] data, out OpenSslPkcs12Reader pkcs12Reader)
```
bartonjs: Do these methods that no longer need to call OpenSslD2I still need to be declared unsafe?

bartonjs:
Minor feedback, otherwise LGTM.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
eerhardt: I've removed all the 'unsafe' declarations except one - it is still using pointers.

src/Native/System.Security.Cryptography.Native/pal_x509.cpp
eerhardt: I agree it logically does belong there, but following the [Interop Guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/interop-guidelines.md):

>- Minimize unnecessary IL in assemblies.
  - Interop signatures should only be compiled into the assemblies that
    actually consume them. Having extra signatures bloats assemblies and
    makes it more difficult to do static analysis over assemblies to
    understand what they actually use. It also leads to problems when such
    static verification is used as a gate, e.g. if a store verifies that
    only certain APIs are used by apps in the store.

If I was to put the signature in Interop.X509Name.cs, then System.Net.Security would have a new method compiled in it that it doesn't use, and System.Security.Cryptography.X509Certificates would have 2 new methods that it doesn't use.
(Note: I like to keep the Interop.X.cs file 1:1 with pal_x.cpp/h, as I think this makes the most sense when trying to find where methods are declared.)

eerhardt:
@dotnet-bot test this please (Socket test failed, which I don't think I broke)

src/Native/System.Security.Cryptography.Native/pal_x509.cpp
bartonjs: X509Certificates will end up using the methods that are in X509Name as you make further shimming progress; and the SafeHandle should probably move into Interop.X509Name.cs as well (or have the file further exploded).

So my preferences would be
1. Move this to pal_x509_name, and Interop.X509Name.cs
2. Move this to pal_x509_name, move the extern to Interop.X509Name_Decode.cs or Interop.DecodeX509Name.cs
3. Move this to pal_x509_name, leave it in Interop.X509.cs
4. Leave the native method in pal_x509, the extern method in Interop.X509.cs
5. Leave the native method in pal_x509, move the extern to Interop.X509Name.cs

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslPkcs12Reader.cs
bartonjs: Danke :)

nguerrera:
LGTM.

stephentoub:
@dotnet-bot test this please (#3494)

stephentoub:
LGTM

eerhardt:
FYI - I still have 1 comment to address from Jeremy, so please don't merge this yet.

eerhardt:
All code review comments have been addressed and squashed.

PR #3792, opened 2015-10-12
Removing stale interop file that is no longer called

sokket:
Removing stale interop file (Interop.getsetrlimit.cs) that is no longer called due to an existing shim (Interop.ResourceLimits.cs)

/cc @nguerrera


nguerrera:
:+1: Thanks

PR #3795, opened 2015-10-12
Enable support for secp256r1 (NIST-P256) in Unix ECDSA

### LINKED ISSUE #3772: Unix ECDsa doesn't support P-256
Windows supports { 256, 384, 521 }
Unix supports { 224, 384, 521 }

I suspect the 224 was meant to be 256 (since 256 is a SuiteB curve).  256 should be added.

bartonjs:
As of FIPS 186-3, the [SuiteB ECDsa curves](https://www.nsa.gov/ia/_files/ecdsa.pdf) are P-256 and P-384.  The Unix implementation doesn't allow P-256, which seems fairly limiting in interoperability.

Fixes #3772.

cc: @AtsushiKan


stephentoub:
LGTM

PR #3796, opened 2015-10-12
Add perf test to deserialize large object graph

### LINKED ISSUE #3302: Perf tests that deserialize large object graphs

khdang:
Fix #3302

### DIFF
--- a/src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
+++ b/src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
@@ -15,26 +15,25 @@ namespace System.Runtime.Tests
         [Benchmark]
         public void DeserializeLargeObjectGraph()
         {
-            var obj = CreateObject(SimpleType.GraphSize, 0);
+            SimpleType obj = CreateObject(0, 0, -1);
             var serializer = new DataContractSerializer(typeof(SimpleType));
             using (var stream = new MemoryStream())
             {
                 serializer.WriteObject(stream, obj);
-                stream.Position = 0;
 
                 foreach (var iteration in Benchmark.Iterations)
                     using (iteration.StartMeasurement())
                         for (int i = 0; i < 3; i++)
                         {
-                            var obj2 = serializer.ReadObject(stream);
                             stream.Position = 0;
+                            SimpleType obj2 = (SimpleType)serializer.ReadObject(stream);
                         }
             }
         }
 
-        public SimpleType CreateObject(int height, int id)
+        public SimpleType CreateObject(int height, int parentId, int currentId)
         {
-            var index = height * 10 + id;
+            int index = parentId * SimpleType.SimpleTypeListSize + (currentId + 1);
             var obj = new SimpleType()
             {
                 IntProperty = index,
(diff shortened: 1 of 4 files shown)


khdang:
cc: @shmao @SGuyGe @ianhays

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                serializer.WriteObject(stream, obj);
+                stream.Position = 0;
+
+                foreach (var iteration in Benchmark.Iterations)
+                    using (iteration.StartMeasurement())
+                        for (int i = 0; i < 3; i++)
+                        {
+                            var obj2 = serializer.ReadObject(stream);
+                            stream.Position = 0;
```
stephentoub: If you move this up one line, you could get rid of the earlier stream.Position = 0 call.

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
```diff
@@ -0,0 +1,75 @@
...
+                        }
+            }
+        }
+
+        public SimpleType CreateObject(int height, int id)
+        {
+            var index = height * 10 + id;
```
stephentoub: What's the purpose of index here?  Multiple nodes in this tree are going to end up with the same height and same id values, and thus the same index...is that ok?

If nothing else, this should be int rather than var.

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
```diff
@@ -0,0 +1,75 @@
...
+            var obj = new SimpleType()
+            {
+                IntProperty = index,
+                StringProperty = index + " string value",
+                EnumProperty = (MyEnum)(index % (Enum.GetNames(typeof(MyEnum)).Length)),
+                CollectionProperty = new List<string>(),
+                SimpleTypeList = new List<SimpleType>()
+            };
+            for (var i = 0; i < SimpleType.CollectionSize; ++i)
```
stephentoub: var => int

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                obj.CollectionProperty.Add(index + "." + i);
+            }
+            if (height > 0)
+            {
+                for (var i = 0; i < SimpleType.SimpleTypeListSize; ++i)
```
stephentoub: var => int

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Runtime.Serialization;
+
+namespace System.Runtime.Tests
+{
+    public class Perf_Deserialization
+    {
+        [Benchmark]
+        public void DeserializeLargeObjectGraph()
+        {
+            var obj = CreateObject(SimpleType.GraphSize, 0);
```
stephentoub: var => SimpleType

src/System.Runtime.Serialization.Xml/tests/System.Runtime.Serialization.Xml.Tests.csproj
```diff
@@ -19,6 +19,18 @@
     <Compile Include="MyResolver.cs" />
     <Compile Include="XmlDictionaryWriterTest.cs" />
   </ItemGroup>
+  <!-- Performance tests require v5.0 portable tools -->
```
ianhays: This section is no longer needed. Some changes to buildtools were made that allowed us to leave out this entire section:
```
  <!-- Performance tests require v5.0 portable tools -->  
  <PropertyGroup Condition="'$(RunPerfTestsForProject)' == 'true'">  
    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>  
    <TargetFrameworkProfile />  
  </PropertyGroup>  
```
I'll update the docs accordingly.

src/System.Runtime.Serialization.Xml/tests/Performance/Perf.Deserialization.cs
khdang: My intention was just to have fixed string values in each node so having the same values in the nodes are fine. Nonetheless I've updated so that the index value is unique for each node.

khdang:
Thanks for reviewing. I've made the changes based on the comments.

PR #3798, opened 2015-10-13
Shim'ing out the processor affinity function calls for the Linux kernel

sokket:
Shim'ing out the processor affinity function calls for the Linux kernel

/cc @nguerrera @stephentoub


nguerrera:
LGTM

src/Common/src/Interop/Unix/System.Native/Interop.SchedGetSetAffinity.cs
```diff
@@ -0,0 +1,29 @@
...
+        internal static unsafe extern int SchedSetAffinity(int pid, CpuSetBits* mask);
+
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal static unsafe extern int SchedGetAffinity(int pid, CpuSetBits* mask);
+
+        [DllImport(Libraries.SystemNative)]
+        internal static unsafe extern void CpuSet(int cpu, CpuSetBits* set);
+
+        [DllImport(Libraries.SystemNative)]
+        internal static unsafe extern bool CpuIsSet(int cpu, CpuSetBits* set);
```
stephentoub: Can these functions use ref/out rather than pointers?

src/Native/System.Native/pal_process.cpp
```diff
@@ -417,3 +427,51 @@ extern "C" char* GetCwd(char* buffer, int32_t bufferSize)
...
+    cpu_set_t set = {};
+    ConvertPalToNativeCpuSet(*mask, set);
+    return sched_setaffinity(pid, sizeof(cpu_set_t), &set);
+}
+#endif
+
+#if HAVE_SCHED_GETAFFINITY
+static void ConvertNativeToPalCpuSet(const cpu_set_t& native, CpuSetBits& pal)
+{
+    memset(pal.Bits, 0, ARRAY_SIZE(pal.Bits));
```
stephentoub: Elsewhere I think we've written this as:
```C
pal = { };
```
Does that not work here for some reason?

src/Native/System.Native/pal_process.cpp
```diff
@@ -417,3 +427,51 @@ extern "C" char* GetCwd(char* buffer, int32_t bufferSize)
 
     return getcwd(buffer, UnsignedCast(bufferSize));
 }
+
+#if HAVE_SCHED_SETAFFINITY
+static void ConvertPalToNativeCpuSet(const CpuSetBits& pal, cpu_set_t& native)
+{
+    memcpy(native.__bits, pal.Bits, ARRAY_SIZE(native.__bits));
```
stephentoub: Earlier we statically asserted that sizeof(CpuSetBits) >= sizeof(cpu_set_t).  So while unlikely, it's possible that CpuSetBits is larger than cpu_set_t, which means it's possible that this call could be trying to set affinity to a value that's out of range for the native side, and because we're only coping ARRAY_SIZE(native.__bits), we'll end up trimming those off and losing the information.  Are we ok with that?

stephentoub:
A few comments/questions, but otherwise LGTM.

src/Common/src/Interop/Unix/System.Native/Interop.SchedGetSetAffinity.cs
sokket: Sure

src/Native/System.Native/pal_process.cpp
sokket: Sure, I'll change it

src/Native/System.Native/pal_process.cpp
sokket: I'll add a check when sizeof(pal) > sizeof(native) that will look at the bits from the end of pal until the end of native and make sure they are all 0 so that we won't truncate data. If we would truncate, I'll set errno to EINVAL so it will bubble up to the managed layer

src/Common/src/Interop/Unix/System.Native/Interop.SchedGetSetAffinity.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal unsafe struct CpuSetBits
```
stephentoub: One more thought.  Since the Bits field is never used and is here purely to impact the size of the struct, you could just add a StructLayout attribute to the type to set its size to 128.  Then the type wouldn't need to be unsafe.  But if you'd prefer to stick with what you have, that's fine, though you might chaynge the visibility of Bits to be private.

stephentoub:
LGTM

src/Common/src/Interop/Unix/System.Native/Interop.SchedGetSetAffinity.cs
sokket: I'll change the field to be private so that it's still visible what the struct contains and how it fits with the native implementation.

stephentoub: Ok.

sokket:
Everything is green with LGTM's....merging. Thanks folks!

PR #3800, opened 2015-10-13
Make Unix ECDSA support the Windows signature format

### LINKED ISSUE #3769: ECDsa Sign and Verify expect different data forms between Windows and Unix

bartonjs:
OpenSSL's ECDSA signature routine returns a signature block conforming to the structure embedded in a certificate:
```
SEQUENCE(
  r INTEGER,
  s INTEGER
)
```
Windows creates an array of BitsToBytes(KeySize) for each of r and s, then returns the concatenation of their values.

Since ECDsaCng was added in .NET 3.5 as a thin wrapper over the Windows API, that's the .NET data format.

ECDsaOpenSsl converts the Windows format to the OpenSSL format on VerifyHash, and the OpenSSL format to the Windows format on SignHash.

Signatures produced with values prior to this commit will no longer verify, only the Windows format is supported.

DerSequenceReader was moved into Common (from X509Certificates) and a DerEncoder was added; both of which now have tests in the System.Security.Cryptography.Encoding test library.

Fixes #3769.

cc: @AtsushiKan @stephentoub


bartonjs:
@dotnet-bot test this please (I believe you that it failed, but it works locally, and I'm kinda hoping you had a stale copy when you ran...)

stephentoub:
Still fails:
```
23:06:16    System.Security.Cryptography.X509Certificates.Tests.PublicKeyTests.TestECDsaPublicKey_ValidatesSignature [FAIL]
23:06:16       isSignatureValid
23:06:16       Expected: True
23:06:16       Actual:   False
23:06:16       Stack Trace:
23:06:16             at System.Security.Cryptography.X509Certificates.Tests.PublicKeyTests.TestECDsaPublicKey_ValidatesSignature()
23:06:16             at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
23:06:16             at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
23:06:16             at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

src/Common/src/System/Security/Cryptography/DerEncoder.cs
```diff
@@ -0,0 +1,197 @@
...
+            // So:
+            // 0 => 0x00.
+            // 1 => 0x01.
+            // 127 => 0x7F.
+            // 128 => 0x81 0x80
+            // 255 => 0x81 0xFF
+            // 256 => 0x82 0x01 0x00
+            // 65535 => 0x82 0xFF 0xFF
+            // 65536 => 0x83 0x01 0x00 0x00
+            // ...
```
stephentoub: Consider adding an example for int.MaxValue just to highlight the upper bound:
```
int.MaxValue => 0x84 0x7F 0xFF 0xFF 0xFF
```

src/System.Security.Cryptography.Encoding/tests/System.Security.Cryptography.Encoding.Tests.csproj
```diff
@@ -17,11 +17,22 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AsnEncodedData.cs" />
+    <Compile Include="DerEncoderTests.cs" />
+    <Compile Include="DerSequenceReaderTests.cs" />
     <Compile Include="Oid.cs" />
     <Compile Include="OidCollection.cs" />
+    <Compile Include="$(CommonPath)\System\Security\Cryptography\DerEncoder.cs">
+      <Link>Common\System\Security\Cryptography\DerEncoder.cs</Link>
+    </Compile>
```
stephentoub: Weren't you just expressing concern about this practice? :smile:

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/ECDsaOpenSsl.cs
```diff
@@ -177,6 +192,77 @@ protected override void Dispose(bool disposing)
...
+        {
+            int size = GetSignatureFieldSize();
+
+            try
+            {
+                DerSequenceReader reader = new DerSequenceReader(input, inputOffset, inputCount);
+                byte[] rDer = reader.ReadIntegerBytes();
+                byte[] sDer = reader.ReadIntegerBytes();
+                byte[] response = new byte[2 * size];
+
```
stephentoub: Assert that rDer.Length == sDer.Length == size?

stephentoub:
LGTM other than whatever's causing the test failure.

src/System.Security.Cryptography.Encoding/tests/System.Security.Cryptography.Encoding.Tests.csproj
bartonjs: Ish :) This is a class (well, two) instead of the whole library; and I actually want to make what they're doing public and just use the public surface area.  Hopefully this is a point-in-time solution :).

bartonjs:
@stephentoub (or other smart people), can you take a look and see what I'm missing?

* http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/3347/testReport/junit/System.Security.Cryptography.X509Certificates.Tests/PublicKeyTests/TestECDsaPublicKey_ValidatesSignature/ claims it's still returning false.
* https://github.com/dotnet/corefx/commit/2bb54b055ac3bfa09679bfe11b42e05ec8df7cd8 made it such that false can no longer be returned, instead it throws an exception (so I could figure out what's going on)
* http://dotnet-ci.cloudapp.net/job/dotnet_corefx_linux_debug_tst_prtest/3347/testReport/junit/System.Security.Cryptography.EcDsaOpenSsl.Tests/EcDsaOpenSslTests/CtorHandle384/ shows that that works.


Wait a minute... is this because I switched from P2P to the package references, so I have a fix-lag?

stephentoub:
> is this because I switched from P2P to the package references, so I have a fix-lag?

Likely :smile:

bartonjs:
Ugh.  Okay, leaving the X509Certificates test that was failing disabled; and I'll give myself an issue to re-enable it in a couple of days.

bartonjs:
@dotnet-bot test this please (and try to not let the flaky sockets test fail)

PR #3803, opened 2015-10-13
Remove needless checks on creating ReadOnlyCollection in Linq.Queryable

JonHanna:
System.Linq.Queryable uses ReadOnlyCollectionExtensions to create
ReadOnlyCollection instances in a way that should increase aliasing (hence
reducing allocations), and guarantee that the underlying collection is not
mutated elsewhere.

However, the only paths that hit it do so with collections that will never
be already ReadOnlyCollection (one possible alias opportunity gone) or empty
or null (the other gone). Nor will the collections ever be accessed by any
other code (removing the risk of mutation).

Futher, while the use of `ToArray()` also gives the lightest possible choice
for contained collection, with an advantage on memory pressure if the instance
was long-lasting, all collections so-produced are soon eligible for collection,
so calling `ToArray()` just leads to more allocations unnecessarily, along with
having a more costly interface invocation.

Therefore, replace such uses with `List<T>`'s `AsReadOnly()`.


JonHanna:

stephentoub:
LGTM

cc: @VSadov

PR #3804, opened 2015-10-13
Reduce allocations in Enumerable.Block and remove always-true test

JonHanna:
Don't call `ToArray()` in `GetOptimizedBlockExpression()` if the argument is
a `ReadOnlyCollection`, to work with the attempts throughout the project to
reduce allocations of such collections.

Conversely, don't call `ToReadOnlyCollection()` in one of the `Block()`
overloads if the source already implements `IReadOnlyList` to allow lighter
array objects to be used more often.

One overload of `Block` uses `as IReadOnlyList<Expression>` and a null-check on
a variable that will never be null, and of a type that implements
`IReadOnlyList<Expression>` and hence will always take that branch.

Change to just use that variable directly.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
@@ -847,14 +847,14 @@ namespace System.Linq.Expressions
         public static BlockExpression Block(IEnumerable<ParameterExpression> variables, IEnumerable<Expression> expressions)
         {
             ContractUtils.RequiresNotNull(expressions, "expressions");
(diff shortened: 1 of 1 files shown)


JonHanna:
@VSadov the first of these commits is a compromise in terms of https://github.com/dotnet/corefx/pull/3686#discussion_r41542478 in that it only uses `ReadOnlyCollection` if it already has one, and reduces how often that will happen, to gain the advantage of holding an array you mentioned, while still using that type if it receives one, to gain the advantage of reduced allocation in some cases that I talked about.

stephentoub:
LGTM

VSadov:
LGTM

PR #3809, opened 2015-10-13
Fix bugs in IPPacketInformation.

pgavlin:
- GetHashCode() and Equals() both stood the chance of dereferencing a null
  pointer when called on default values.
- GetHashCode() was using a weak method of combining hash codes. This has
  been updated to use the same technique used by Roslyn.
- The equality and inequality operators were unnecessarily boxing input
  parameters.

### DIFF
--- a/src/System.Net.Sockets/src/System/Net/Sockets/IPPacketInformation.cs
+++ b/src/System.Net.Sockets/src/System/Net/Sockets/IPPacketInformation.cs
@@ -32,14 +32,15 @@ namespace System.Net.Sockets
             }
         }
 
-        public static bool operator ==(IPPacketInformation packetInformation1, IPPacketInformation packetInformation2)
+        public static bool operator ==(IPPacketInformation left, IPPacketInformation right)
         {
-            return packetInformation1.Equals(packetInformation2);
+            return left._networkInterface == right._networkInterface &&
+                (left._address == null && right._address == null || left._address.Equals(right._address));
         }
 
-        public static bool operator !=(IPPacketInformation packetInformation1, IPPacketInformation packetInformation2)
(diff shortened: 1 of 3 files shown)


pgavlin:
@CIPop @davidsh PTAL

stephentoub:
LGTM

src/System.Net.Sockets/src/System/Net/Sockets/IPPacketInformation.cs
```diff
@@ -49,13 +50,13 @@ public override bool Equals(object comparand)
...
             }
 
-            IPPacketInformation obj = (IPPacketInformation)comparand;
-            return _address.Equals(obj._address) && _networkInterface == obj._networkInterface;
+            return this == (IPPacketInformation)comparand;
         }
 
         public override int GetHashCode()
         {
-            return _address.GetHashCode() + _networkInterface.GetHashCode();
```
CIPop: If the tests have different results, for support purposes, please add a comment explaining how this is behaving different from desktop.

CIPop:
Looks like we're going to get different behavior from Desktop. In all cases where that is the case, please add a test describing how the behavior is different. That is going to help with future support cases.

LGTM

pgavlin:
Yes, we will: in cases where the desktop framework would have thrown a `NullReferenceException`, this code will succeed. This change contains tests that pin this behavior (`GetHashCode_DefaultValues_Success` and `Equals_DefaultValues_Success`).

CIPop:
> in cases where the desktop framework would have thrown a  NullReferenceException , this code will succeed. This change contains tests that pin this behavior ( GetHashCode_DefaultValues_Success  and  Equals_DefaultValues_Success ).

Please add above as a comment in the Test. It makes it much easier to answer support questions in the future.

PR #3812, opened 2015-10-13
Shim Interop.Bignum and Interop.BIO methods.

eerhardt:


src/Native/System.Security.Cryptography.Native/pal_bignum.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pal_bignum.h"
+
+extern "C" void BigNumDestroy(BIGNUM* a)
```
bartonjs: Why `Destroy` instead of `Free`? (Just curious)

src/Native/System.Security.Cryptography.Native/pal_bio.cpp
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pal_bio.h"
+
+#include <assert.h>
+
+extern "C" BIO* BioCreateNew()
```
bartonjs: I'd go with `CreateMemoryBio`, `BioNewMemory` or something; to indicate that it's already been assigned a BIO type.  (BIO is similar to Stream; this is effectively MemoryStream)

src/Native/System.Security.Cryptography.Native/pal_bio.h
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pal_types.h"
+
+#include <openssl/bio.h>
+
+/*
+Creates a new BIO instance.
```
bartonjs: "Creates a new memory-backed BIO instance."

src/Native/System.Security.Cryptography.Native/pal_bio.h
```diff
@@ -0,0 +1,47 @@
...
+extern "C" int32_t BioRead(BIO* b, void* buf, int32_t len);
+
+/*
+Direct shim to BIO_write.
+*/
+extern "C" int32_t BioWrite(BIO* b, const void* buf, int32_t len);
+
+/*
+Shims the BIO_get_mem_data to get the size of data available in the BIO.
+*/
```
bartonjs: Are these using the same documentation style as the rest of the shims? I thought they had more structure?  Just making sure we're being consistent.

bartonjs:
Some small things, but LGTM; modulo the compile failure.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.BIO.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+
+using Microsoft.Win32.SafeHandles;
+
+using NativeLong=System.IntPtr;
```
nguerrera: This can be removed now, right?

nguerrera:
LGTM apart from one small comment.

src/Native/System.Security.Cryptography.Native/pal_bignum.cpp
eerhardt: Just following the naming pattern from Interop.Hmac (HmacDestroy).

src/Native/System.Security.Cryptography.Native/pal_bio.h
eerhardt: Done.

src/Native/System.Security.Cryptography.Native/pal_bio.cpp
eerhardt: Went with CreateMemoryBio .

src/Native/System.Security.Cryptography.Native/pal_bio.h
eerhardt: I've seen some with "Function: *functionName*" in the documentation, but this seems overly redundant and prone to getting out-of-date.

sokket: We've been using a comment header like so:
```
/**
 * Quick description of function
 *
 * What function returns
 */
```

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.BIO.cs
eerhardt: Yes, I missed this one.  Removed.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Bignum.cs
```diff
@@ -3,33 +3,23 @@
...
-
-        [DllImport(Libraries.LibCrypto)]
-        private static extern IntPtr BN_bin2bn(byte[] s, int len, IntPtr zero);
+        [DllImport(Libraries.CryptoNative)]
+        internal static extern void BigNumDestroy(IntPtr a);
 
-        [DllImport(Libraries.LibCrypto)]
-        private static extern unsafe int BN_bn2bin(SafeBignumHandle a, byte* to);
+        [DllImport(Libraries.CryptoNative)]
+        private static extern IntPtr BigNumFromBinary(byte[] s, int len);
```
stephentoub: Just curious, why doesn't this return a SafeBignumHandle?

src/Native/System.Security.Cryptography.Native/pal_bio.cpp
```diff
@@ -0,0 +1,44 @@
...
+extern "C" int32_t BioWrite(BIO* b, const void* buf, int32_t len)
+{
+    return BIO_write(b, buf, len);
+}
+
+extern "C" int32_t GetMemoryBioSize(BIO* bio)
+{
+    long ret = BIO_get_mem_data(bio, nullptr);
+    
+    assert(ret <= INT32_MAX);
```
stephentoub: Why is this the case?  If it's due to a guarantee made elsewhere, great.

src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/OpenSslAsnFormatter.cs
```diff
@@ -48,11 +48,11 @@ protected override string FormatNative(Oid oid, byte[] rawData, bool multiLine)
                         return null;
                     }
 
-                    int printLen = Interop.libcrypto.GetMemoryBioSize(bio);
+                    int printLen = Interop.Crypto.GetMemoryBioSize(bio);
 
                     // Account for the null terminator that it'll want to write.
                     StringBuilder builder = new StringBuilder(printLen + 1);
-                    Interop.libcrypto.BIO_gets(bio, builder, builder.Capacity);
+                    Interop.Crypto.BioGets(bio, builder, builder.Capacity);
```
stephentoub: @bartonjs, separate from this change, we should audit the various places StringBuilder is being used in these signatures.  All of the uses of BioGets, for example, are of the form:
```C#
StringBuilder sb = new StringBuilder(size);
NativeCall(sb, sb.Capacity);
return sb.ToString();
```
This not only allocates the underlying char[], but also the StringBuilder object, as well as additional copy costs to marshal out and in the current StringBuilder implementation another char[] allocation on the way back.  Seems it could be implemented just as easily as:
```C#
char[] chars = new char[size];
NativeCall(chars, chars.Length);
return Encoding.UTF8.GetString(chars);
```
or something like that.

stephentoub:
LGTM

src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/OpenSslAsnFormatter.cs
eerhardt: @stephentoub - What's the advantage to ever using a StringBuilder in an interop call?  In mscorlib, we use the StringBuilderCache, which saves on allocation if the cached builder is large enough.  But outside of caching, why else would we use a StringBuilder?

stephentoub: That's a good question :wink:  With the current marshaling implementation (which I think we should look at fixing), I'm having trouble coming up with a good argument for why you'd ever want to use one, at least as [Out].  For [In], it might be a little bit more reasonable, but even there the current marshaling implementation has some costs that you could likely workaround using something other than StringBuilder.

src/Native/System.Security.Cryptography.Native/pal_bio.cpp
eerhardt: BIO_get_mem_data is a macro over top of `BIO_ctrl(b,BIO_CTRL_INFO,0,(char *)pp)`.  When BIO_ctrl called with BIO_CTRL_INFO, it returns 
```C++
case BIO_CTRL_INFO:
        ret = (long)ctx->cont;
        break;
```
where ctx->cont is an int.

Also, all the callers of GetMemoryBioSize use the result for the capacity to create arrays and StringBuilders, so it needs to get casted to an int32 somewhere.

I've added a comment.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Bignum.cs
eerhardt: This method is called from `internal static IntPtr CreateBignumPtr` below. CreateBignumPtr is called from https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs#L291 and put on the "RSA_ST" struct.  The RSA_ST struct is eventually freed by SafeRsaHandle, which calls [RSA_free](http://openssl.org/docs/manmaster/crypto/RSA_new.html) which 
>RSA_free() frees the RSA structure and its components.

This will take care of freeing the Bignum memory.

eerhardt: Maybe when RSAOpenSsl is shimmed this will get cleaner...?

bartonjs: Yeah, I think I wrote that before System.Security.Cryptography.Native existed.

And I was probably being lazy with "oh, the lifetime threat is small" vs calling SetHandleAsInvalid.  Not things I'd let others get away with.

::smacks self on the hand:: Bad Jeremy.

Issue #2025, opened 2015-06-12
DeflateStream's crc32 calculation consuming up to 50% of GZipStream write time


stephentoub:
Writing to a GZipStream is spending the majority of its execution time in zlib's deflate implementation, but the remainder is mostly spent in a crc32 calculation, anywhere from 0 to 50% depending on how complicated the compression is.  We should investigate whether using zlib's crc32 would result in measurably faster results.

Maxwe11:
How to measure this? If any perf tests present in repo how to run it?

stephentoub:
I wrote some benchmarks locally and was profiling those with VS.

Though I just realized that this would be a difficult issue to grab given that we haven't yet published clrcompression.dll and it doesn't currently export crc32.  I'm going to remove the up for grabs and assign it to myself for now.  Thanks for the interest, though!

justinvp:
FYI for potential future optimizations... Intel implemented some patches to zlib and one of the changes was to make use of the SSE4.2 crc32 instruction on systems that support it. http://www.intel.com/content/dam/www/public/us/en/documents/white-papers/zlib-compression-whitepaper-copy.pdf

stephentoub:
Thanks, Justin.  Yeah, I saw that. All the more reason to use it :smile:

bjjones:
I’m happy to see this issue being raised. I identified this myself a few months back and spent some time working to find a solution. I agree utilizing the hardware instruction would be optimal.

I’d be very interested in taking ownership of this issue when clrcompression.dll becomes available.

stephentoub:
@bjjones, you're welcome to give it a go.  I already updated clrcompression.dll to export crc32, and that updated version should already be in the NuGet packages now being used by corefx.  Thanks!

stephentoub:
Fixed by #3817 and #3987

PR #3817, opened 2015-10-13
Improved crc32 generation using improved algorithm in managed code

bjjones:
In reference to issue #2025.

This pull request replaces the implementation of the Sarwate crc32 algorithm in crc32helper.cs with Intel's improved "Slicing-by-8" algorithm described in this IEEE paper: http://ieeexplore.ieee.org/xpls/abs_all.jsp?arnumber=4531728&tag=1

The algorithm improves performance by increasing the size of the lookup table, and allowing crc32 calculation on 8 bytes at a time. 

I have tested extensively, with a 1.92x speedup in the method using an Intel® 5th Generation Core (tm) i5 Processor based system, as well as on an Intel® Xeon® Processor E5 v2 Family based server system.

This is the first of two related pull requests. This will improve the managed crc32 implementation, and will affect platforms that do not have clrcompression.dll available for use. My next pull request will make use of the native crc32 function included with clrcompression.dll. I have tested the native function as well, and it shows a slightly higher performance increase than the managed algorithm described in this pull request.


stephentoub:
Thanks for the PR, @bjjones.

A couple of high-level questions:

1) We're mostly phasing out the managed deflate implementation (though it's still there as a backup), but the managed crc32 calculation is actually used in the current implementation even when zlib/clrcompression is in use.  Do you have any measurements for how this change affects GZipStream all-up when using the zlib implementation, on various kinds of inputs?  In a previous test I'd found that depending on the inputs, anywhere from 0 to 50% of the time could be spent in the crc32 calculation, so if we could be cutting half of that away, that'd be a great improvement.

2) To your knowledge, is the algorithm you implement public domain, without any kind of restrictions?

bjjones:
1) Yes, I have tested with GZipStream and the 1.92x gain holds over the workloads I've used. In my own workloads, crc32 generation accounted for ~5-11% of time, with the optimization improving the workloads by 2.5-5.5%, respectively. I personally am not aware of a case where crc32 generation takes 50% of compression time, but I'm confident you would see a significant gain in that case.

2) Yes, this has been approved by Intel to be contributed to the .NET Foundation as per the CLA.

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
```diff
@@ -62,22 +63,428 @@ internal static class Crc32Helper
...
+                       && (offset <= buffer.Length - length), "check the caller");
 
-            crc32 ^= 0xffffffffU;
+            uint term1, term2, term3 = 0;
+            int running_length;
+            int end_bytes;
+
+            crc32 ^= 0xFFFFFFFFU;
+            running_length = (length / 8) * 8;
+            end_bytes = length - running_length;
```
stephentoub: Nit: just a style thing, but could you please define the locals here rather than separately above, and name them with camel casing rather than with underscore separators, e.g.
```C#
int runningLength = (length / 8) * 8;
int endBytes = length - runningLength;
```

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
```diff
@@ -62,22 +63,428 @@ internal static class Crc32Helper
...
+
+            for (int i = 0; i < running_length / 8; i++)
+            {
+                crc32 ^= (uint)(buffer[offset] | buffer[offset + 1] << 8 | buffer[offset + 2] << 16 | buffer[offset + 3] << 24);
```
stephentoub: Here and below, it appears we're assuming a particular endianess.  Does that matter?  If it does, maybe we should at least add a ```Debug.Assert(BitConverter.IsLittleEndian)``` to help us catch future issues related to this?

Also, I wonder if you experimented with pinning the buffer for the duration of the loop and using unsafe code to access it; that would help to avoid these multiple indexings and shifts and whatnot, as you could just index into the buffer as an ```int*```.

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
```diff
@@ -62,22 +63,428 @@ internal static class Crc32Helper
             0x5d681b02u, 0x2a6f2b94u, 0xb40bbe37u, 0xc30c8ea1u, 0x5a05df1bu,
             0x2d02ef8du
         };
+        private static readonly uint[] s_crcTable_1 = new uint[256] {
```
stephentoub: Could you add a comment citing how these tables were generated and including a citation for the paper that introduces the algorithm?

stephentoub:
Thanks for the responses, @bjjones.

>  I personally am not aware of a case where crc32 generation takes 50% of compression time, but I'm confident you would see a significant gain in that case.

I tried out your changes with a variety of synthetic cases.  With an obviously-contrived-but-nonetheless-interesting case of repeated sequences of all bytes 0 to 255 and writing out 1K at a time to the GZipStream, on my machine I see the crc32 computation taking up between 40-50% of the time, and your PR improves the overall throughput of the GZipStream compression by ~20%.  That's with the zlib implementation.  Nice.

I left a few small comments/questions on the changes, but overall it LGTM.

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
bjjones: Added a citation at the top, as well as a brief description on how the tables we're generated. Let me know if the description is not adequate.

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
bjjones: Yes, this will only work on little endian systems. I added a Debug.Assert.

I experimented with setting the buffer as fixed today, but I did not see a noticeable change in performance.

src/System.IO.Compression/src/System/IO/Compression/Crc32Helper.cs
bjjones: Fixed.

bjjones:
Thanks for the feedback @stephentoub.  I've addressed each note. 

I'm curious that it currently says "All checks have failed"? Is this definitely an issue on my end? I am not having an issue building and running the included tests on my system, and there was very little change from the last successful build.

stephentoub:
@dotnet-bot test this please (NuGet failures: FindPackagesById)

stephentoub:
Thanks, @bjjones.

> I'm curious that it currently says "All checks have failed"?

Looks like there was a problem downloading NuGet packages (for future reference, you can click on the details link to take you to the Jenkins reports about each leg and what went wrong).  I've restarted the job.

stephentoub:
LGTM.  Thanks for doing this, @bjjones.

PR #3822, opened 2015-10-13
System.Diagnostics.Tracing.Logger implementation

vancem:
This is the new Logger API, that is intended to be used by the ASP.NET team.    See the comments in Logger.cs for why we need this in addition to EventSource.   

Because Logger is implemented using DiagnosticSource, it will leverage the (upcoming) bridge that allows all EventListeners (including ETW), to access data logged with DiagnosticSource (and thus Logger).   

This code is intended to replace a chunk of what is currently in Microsoft.Extensions.Logging (see https://github.com/aspnet/Logging).   Logger and ILogger however where moved into their on DLL in coreFX because it they are generally useful

This pull request is tentative.   It is here to get feedback from the ASP.NET team and others and as part of that feedback we may move things around.

### DIFF
--- a/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
+++ b/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
@@ -30,13 +30,5 @@ namespace System.Diagnostics.Tracing
         /// <param name="arguments"></param>
         /// <returns></returns>
         IDisposable ActivityStart(string activityName, LogLevel level = LogLevel.Critical, object arguments = null);
-
-        /// <summary>
-        /// Add an observer
-        /// </summary>
-        /// <param name="observer"></param>
-        /// <param name="filter"></param>
-        /// <returns></returns>
-        IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, Func<string, LogLevel, bool> filter);
     }
 }
--- a/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
+++ b/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
@@ -1,0 +1,12 @@
+﻿using System.Collections.Generic;
+
+namespace System.Diagnostics.Tracing
+{
+    interface ILogger
+    {
+        bool IsEnabled(LogLevel level);
+        void Log(string logItemName, LogLevel level, object arguments = null);
+        IDisposable ActivityStart(string activityName, LogLevel level = LogLevel.Critical, object arguments = null);
+        IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, LogLevel level);
+    }
+}
--- a/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
+++ b/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
@@ -2,11 +2,41 @@
 
 namespace System.Diagnostics.Tracing
 {
-    interface ILogger
+    /// <summary>
+    /// A generic interface for logging.
+    /// </summary>
+    public interface ILogger
     {
+        /// <summary>
+        /// Checks if the given LogLevel is enabled.
+        /// </summary>
+        /// <param name="level"></param>
+        /// <returns></returns>
         bool IsEnabled(LogLevel level);
+
+        /// <summary>
+        /// Log a single item
+        /// </summary>
+        /// <param name="logItemName"></param>
+        /// <param name="level"></param>
+        /// <param name="arguments"></param>
         void Log(string logItemName, LogLevel level, object arguments = null);
+
+        /// <summary>
+        /// Start of new Activity
+        /// </summary>
+        /// <param name="activityName"></param>
+        /// <param name="level"></param>
+        /// <param name="arguments"></param>
+        /// <returns></returns>
         IDisposable ActivityStart(string activityName, LogLevel level = LogLevel.Critical, object arguments = null);
+
+        /// <summary>
+        /// Add an observer
+        /// </summary>
+        /// <param name="observer"></param>
+        /// <param name="level"></param>
+        /// <returns></returns>
         IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, LogLevel level);
     }
 }
--- a/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
+++ b/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
@@ -35,8 +35,8 @@ namespace System.Diagnostics.Tracing
         /// Add an observer
         /// </summary>
         /// <param name="observer"></param>
-        /// <param name="level"></param>
+        /// <param name="filter"></param>
         /// <returns></returns>
-        IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, LogLevel level);
+        IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, Func<string, LogLevel, bool> filter);
     }
 }
--- a/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
+++ b/src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
@@ -3,32 +3,35 @@
 namespace System.Diagnostics.Tracing
 {
     /// <summary>
-    /// A generic interface for logging.
+    /// A interface for logging.  It is supported by System.Diagnostics.Tracing.Logger, but it could in theory have
+    /// other implementations.   See Logger for more details on expected use.  
     /// </summary>
     public interface ILogger
     {
         /// <summary>
-        /// Checks if the given LogLevel is enabled.
+        /// returns true if the given verbosity level is active (will be logged if 'Log' is called).  
         /// </summary>
-        /// <param name="level"></param>
-        /// <returns></returns>
         bool IsEnabled(LogLevel level);
 
         /// <summary>
-        /// Log a single item
+        /// Log a single message.   A message has a name, a Verbosity level and a object that contains any other
+        /// information needed by the message.   Typical this object is created using an anonymous type.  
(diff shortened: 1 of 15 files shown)


stephentoub:
@dotnet-bot test this please

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        /// <summary>
+        /// Subscribe to the Logger, sending events to 'observer' for any logging that is enabled for verbosity 'level'.  
+        /// 
+        /// Note that in multi-subscriber cases you MAY get messages that have a more verbose level than you asked for.
+        /// You can filter them out yourself.   If there is only one subscriber (or all subscribers subscribe at the 
+        /// same verbosity.  It works fine.   
+        /// </summary>
+        public virtual IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, Func<string, LogLevel, bool> filter)
```
davidfowl: Is this the replacement for the LoggerProvider? I see the comment about using `DiagnosticListener.AllListeners` to get at a particular instance. I assume there's going to be an example of that somewhere else?

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// returns true if the given verbosity level is active (will be logged if 'Log' is called).  
+        /// </summary>
+        bool IsEnabled(LogLevel level);
+
+        /// <summary>
+        /// Log a single message.   A message has a name, a Verbosity level and a object that contains any other
+        /// information needed by the message.   Typical this object is created using an anonymous type.  
+        /// </summary>
+
+        void Log(string messageName, LogLevel level, object arguments = null);
```
nblumhardt: One of the goals stated in the comment below is the ability to use this as a facade on log4net, NLog etc., but none of those frameworks use this style of "name + payload" logging. Compatibility is probably not worth chasing if the experience is second-class/inconsistent with native use of those frameworks.

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
```diff
@@ -0,0 +1,37 @@
...
+    public interface ILogger
+    {
+        /// <summary>
+        /// returns true if the given verbosity level is active (will be logged if 'Log' is called).  
+        /// </summary>
+        bool IsEnabled(LogLevel level);
+
+        /// <summary>
+        /// Log a single message.   A message has a name, a Verbosity level and a object that contains any other
+        /// information needed by the message.   Typical this object is created using an anonymous type.
```
nblumhardt: In heavily instrumented apps, will this lead to the generation of a unique anonymous type for every structured log event? Since the `arguments` payload is the way to get event-specific data though the API, it seems like the number of types generated could be problematic. (Those types will also have to retain full metadata since the only way to retrieve their properties is via reflection - not sure how this fits with .NET Native.)

nblumhardt: Also, I'm not deeply familiar with the language but believe F# may lack anonymous types.

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// argument object, however it returns an IDisposable, that is indended to be used in a 'using' 
+        /// clause that establishes the scope for the activity.   When the returns IDisposable is disposed
+        /// the 'Stop' message is sent.  
+        /// 
+        /// Typically the intent of an activity is to form this scope, so that information either logged
+        /// as part of the start, stop, (or any message in between), can be assciated with the activity
+        /// (and thus with each other).    It is expected that Activities nest, and thus form stacks. 
+        /// Typically activities that do not propery nest are treated as errors.   
+        /// </summary>
+        IDisposable ActivityStart(string activityName, LogLevel level = LogLevel.Critical, object arguments = null);
```
nblumhardt: `level` is defaulted to `Critical` here but not on `Log()` - not sure if there's a semantic difference.

stephentoub:
Closing and reopening to trigger Jenkins to pick up the PR

stephentoub:
@dotnet-bot test this please (https://github.com/dotnet/corefx/issues/3844)

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// should be used on all hot code paths before 'Log' is called so that even argument setup is skipped if
+        /// logging is off.   
+        /// </summary>
+        public virtual bool IsEnabled(LogLevel level)
+        {
+            if (level > _maxLevel)
+                return false;
+            if (_filter != null)
+            {
+                foreach (Func<string, LogLevel, bool> filter in _filter.GetInvocationList())
```
stephentoub: Every invocation of _filter.GetInvocationList() is going to allocate a new array, and since IsEnabled is meant to be cheap, that seems relatively expensive.  Instead of having _filter be ```Func<string, LogLevel, bool>```, how about making it a ```List<Func<string, LogLevel, bool>>```?

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <summary>
+        /// Subscribe to the Logger, sending events to 'observer' for any logging that is enabled for verbosity 'level'.  
+        /// 
+        /// Note that in multi-subscriber cases you MAY get messages that have a more verbose level than you asked for.
+        /// You can filter them out yourself.   If there is only one subscriber (or all subscribers subscribe at the 
+        /// same verbosity.  It works fine.   
+        /// </summary>
+        public virtual IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, Func<string, LogLevel, bool> filter)
+        {
+            _filter += filter;
```
stephentoub: Is this a leak?  It seems like every call to Subscribe is going to add another filter, but Dispose'ing of the returned IDisposable doesn't remove that filter.

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+                if (!Disposed)
+                {
+                    Disposed = true;
+                    _logger.Log(_activityName + ".Stop", _level, _arguments);
+                }
+            }
+            Logger _logger;
+            LogLevel _level;
+            string _activityName;
+            object _arguments;
```
stephentoub: Nit: to conform with style guidelines, these should all be explicitly annotated as ```private```

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// 
+    /// Logger leverages DiagnosticListener as its 'plumbing'.   Thus Every Logger is a DiagnosticListener which  
+    /// means that you can discover them using DiagnosticListener.AllListeners.   Once you have the Logger you want, 
+    /// you can use the Subscribe() method to connect to the stream of data.  
+    /// 
+    /// System.Diagnostics.Tracing.EventSource will have plumbing to treat all DiagnosticSource (and thus all 
+    /// Loggers), as EventSources (as long as the payloads are serializable).   Thus it is also possible access
+    /// the data stream of Loggers from an EventListener and therefore ETW and LTTng.  
+    /// </summary>
+    public class Logger : DiagnosticListener, ILogger
```
stephentoub: It seems strange that a Logger is a DiagnosticListener... since the purpose of logging is writing to it, wouldn't it be a DiagnosticSource that others listen to, just as EventSources write and are listened to by EventListeners?

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
stephentoub: Or maybe you're doing it this way for thread-safety reasons, as multiple threads could end up finding the listener in AllListeners and subscribing and writing to it concurrently?  If so, that's worth calling out in a comment, and you probably want to use some sort of immutable list for the delegates, one that's cheap to iterate over unlike multicast delegates.

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/Logger.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// Creates a new logger with the given name.   A Logger is a DiagnosticSource which understands LogLevel (verbosity levels)
+        /// </summary>
+        /// <param name="name">The name of the logger.  This is often the comonent name.</param>
+        /// <param name="maxLevel">The most verbose message that will ever be logged.  It defaults to Debug which, being the most
+        /// verbose level, means any message will be logged.   Useful in the context of security to make it less likely that 
+        /// private (debug) information does not get logged.  It also increases the efficienctly of IsEnabled if you only wish
+        /// to EVER turn low-verbosity events.</param>
+        public Logger(string name, LogLevel maxLevel = LogLevel.Debug) : base(name)
+        {
+            _maxLevel = maxLevel;
```
stephentoub: Should validate that maxLevel is in the allowed range.

vancem:
CC @rahku

src/System.Diagnostics.Tracing.Logger/src/System/Diagnostics/Tracing/ILogger.cs
vancem: That was by design (although it is a policy decision).  It was meant to mimic the behavior of the ILogger.BeginScope (which is unconditional).

stephentoub:
What's the status of this PR? Next steps? Etc.?

vancem:
We decided to take a different direction.   We will not be proceeding with this PR.

stephentoub:
Ok, thanks.

PR #3825, opened 2015-10-14
Adding original SSPI object model back to the Windows SslStream implementation

CIPop:
1. In preparation for NegotiateStream I'm adding back the original SSPI object model. Code was brought in from c74b6f2 with cosmetic or no modifications (simplest way to review is to diff against that). SslStreamPal.Windows has been updated to match the original code in SecureChannel.

2. Updated both Windows and Unix PAL to match the interface (a few methods got moved, no substantial changes in non-Windows code).

3. Adding common test timeout configuration.

@bartonjs @vijaykota @rajansingh10 PTAL
/cc: @DavidSh, @josguil @SidharthNabar @pgavlin @stephentoub

### DIFF
--- a/src/System.Net.Security/src/System/Net/CertificateValidationPal.Unix.cs
+++ b/src/System.Net.Security/src/System/Net/CertificateValidationPal.Unix.cs
@@ -17,14 +17,14 @@ namespace System.Net
 
         internal static SslPolicyErrors VerifyCertificateProperties(
             X509Chain chain,
-            X509Certificate2 certificate,
+            X509Certificate2 remoteCertificate,
             bool checkCertName,
             bool isServer,
             string hostName)
         {
             SslPolicyErrors sslPolicyErrors = SslPolicyErrors.None;
 
-            if (!chain.Build(certificate))
+            if (!chain.Build(remoteCertificate))
             {
                 sslPolicyErrors |= SslPolicyErrors.RemoteCertificateChainErrors;
             }
@@ -39,7 +39,7 @@ namespace System.Net
                 {
                     int hostnameMatch;
 
-                    using (SafeX509Handle certHandle = Interop.libcrypto.X509_dup(certificate.Handle))
+                    using (SafeX509Handle certHandle = Interop.libcrypto.X509_dup(remoteCertificate.Handle))
                     {
                         IPAddress hostnameAsIp;
 
@@ -83,12 +83,12 @@ namespace System.Net
                 return null;
             }
 
-            GlobalLog.Enter("SecureChannel#" + Logging.HashString(securityContext) + "::RemoteCertificate{get;}");
+            GlobalLog.Enter("CertificateValidationPal.Unix SecureChannel#" + Logging.HashString(securityContext) + "::GetRemoteCertificate()");
             X509Certificate2 result = null;
             SafeFreeCertContext remoteContext = null;
             try
             {
-                int errorCode = SslStreamPal.QueryContextRemoteCertificate(securityContext, out remoteContext);
+                int errorCode = QueryContextRemoteCertificate(securityContext, out remoteContext);
 
                 if (remoteContext != null && !remoteContext.IsInvalid)
                 {
@@ -137,7 +137,7 @@ namespace System.Net
                 Logging.PrintInfo(Logging.Web, SR.Format(SR.net_log_remote_certificate, (result == null ? "null" : result.ToString(true))));
             }
 
-            GlobalLog.Leave("SecureChannel#" + Logging.HashString(securityContext) + "::RemoteCertificate{get;}", (result == null ? "null" : result.Subject));
+            GlobalLog.Leave("CertificateValidationPal.Unix SecureChannel#" + Logging.HashString(securityContext) + "::GetRemoteCertificate()", (result == null ? "null" : result.Subject));
 
             return result;
         }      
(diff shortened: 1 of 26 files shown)


src/System.Net.Security/src/System/Net/CertificateValidationPal.Windows.cs
```diff
@@ -187,12 +186,12 @@ internal static X509Store EnsureStoreOpened(bool isMachineStore)
...
             X509Store store = isMachineStore ? s_myMachineCertStoreEx : s_myCertStoreEx;
 
-            if (store == null)
+            if (Volatile.Read(ref store) == null)
             {
                 lock (s_syncObject)
                 {
                     store = isMachineStore ? s_myMachineCertStoreEx : s_myCertStoreEx;
-                    if (store == null)
+                    if (Volatile.Read(ref store) == null)
```
pgavlin: IIRC the second volatile read is not necessary--the lock above provides the necessary fences.

pgavlin:
LGTM at a glance, but the named reviewers (@bartonjs, @vijaykota, @rajansingh10) should take a look as well.

src/System.Net.Security/src/System/Net/CertificateValidationPal.Windows.cs
bartonjs: I had that belief as well when writing the Unix version of this; but decided that it read as less confusing to treat it the same way both times; and the second volatile read should only ever happen once :).

bartonjs:
The Unix changes look like they're either just moving code, or removing some scary looking try+catch-everythings; so LGTM.  (Didn't look at anything Windowsy)

src/System.Net.Security/src/System/Net/CertificateValidationPal.Windows.cs
CIPop: Various articles mention that the variable that's being checked needs to be `volatile`. To quote exactly:
> Without any memory barriers, it's broken in the ECMA CLI specification too. It's possible that under the .NET 2.0 memory model (which is stronger than the ECMA spec) it's safe, but I'd rather not rely on those stronger semantics, especially if there's any doubt as to the safety. Making the instance variable volatile can make it work, as would explicit memory barrier calls, although in the latter case even experts can't agree exactly which barriers are required. I tend to try to avoid situations where experts don't agree what's right and what's wrong!
Taken from: http://csharpindepth.com/Articles/General/Singleton.aspx

Most likely all instances of the double-checked lock mechanisms used throughout our code could be changed to either the `static` or the `Lazy<T>` implementations. I'll leave a `TODO` here and create a task.

pgavlin: That note refers to fields with the `volatile` specifier, which is overly strong in this case (as the second read doesn't need its own memory barrier). Before `Volatile.{Read,Write}`, however, there was no other way to perform volatile reads/writes.

Regardless, it's worth investigating whether or not Lazy<T> is a better choice.

CIPop:
Thanks @pgavlin and @bartonjs for the quick review!

If anyone didn't get to properly review this PR, please leave your comments here. I will address them in a future iteration.

src/System.Net.Security/src/System/Net/CertificateValidationPal.Unix.cs
```diff
@@ -227,5 +227,27 @@ private static X509Store EnsureStoreOpened(ref X509Store storeField, StoreLocati
...
+                remoteCertContext = new SafeFreeCertContext(remoteCertificate);
+                return 0;
+            }
+            catch
+            {
+                return -1;
+            }
+        }
+
+        private static int QueryContextIssuerList(SafeDeleteContext securityContext, out Object issuerList)
```
vijaykota: @rajansingh10 , please remove this in your next PR. I believe we don't use this anywhere

src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
```diff
@@ -70,55 +70,17 @@ public static SafeFreeContextBufferChannelBinding QueryContextChannelBinding(Saf
...
-            {              
-                return -1;
-            }
+            streamSizes = new StreamSizes(Interop.libssl.SslSizes.HEADER_SIZE, Interop.libssl.SslSizes.TRAILER_SIZE, Interop.libssl.SslSizes.SSL3_RT_MAX_PLAIN_LENGTH);
         }
 
-        public static int QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
+        public static void QueryContextConnectionInfo(SafeDeleteContext securityContext, out SslConnectionInfo connectionInfo)
         {
             connectionInfo = null;
```
vijaykota: Nit: not needed since try/catch is removed

vijaykota:
Unix changes look ok to me. Didn't review changes in *.Windows.cs

src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
CIPop: Sorry about the delay: the code changed since this PR: https://github.com/dotnet/corefx/pull/3525/files

Regarding this particular code: `connectionInfo` is an `out` parameter and it needs to be populated within the method.

CIPop:
Thanks @vijaykota for the review!

src/System.Net.Security/src/System/Net/SslStreamPal.Unix.cs
vijaykota: My comment (outdated, maybe) was that it was getting set at line 83. Earlier exceptions were handled so the out parameter had to be set.

PR #3829, opened 2015-10-14
Unix SslStream: Implement RFC2812 IP Address matching for HTTP over TLS.

### LINKED ISSUE #3445: CheckX509Hostname does not match iPAddress subjectAltName fields
[RFC 2818](http://www.ietf.org/rfc/rfc2818.txt) says:
```
   In some cases, the URI is specified as an IP address rather than a
   hostname. In this case, the iPAddress subjectAltName must be present
   in the certificate and must exactly match the IP in the URI.
```

So, we should do that.

bartonjs:
If the hostname looks like an IP Address then do exact matches against GEN_IPADDR Subject Alternative Name entries.  Unlike the GEN_DNS matching of text hostnames, the IP Address form always falls back to the CN text comparison if no other match has been found.

Fixes #3445.

cc: @vijaykota @stephentoub @rajansingh10 @shrutigarg @CIPop @davidsh

### DIFF
--- a/src/Native/System.Security.Cryptography.Native/openssl.c
+++ b/src/Native/System.Security.Cryptography.Native/openssl.c
@@ -881,7 +881,6 @@ CheckX509Hostname(
     // RFC2818 says that if ANY dNSName alternative name field is matched then we
     // should ignore the subject common name.
 
-    // TODO (3445): Match an input IP Address against the iPAddress SAN entries
     // TODO (3446): Match using IDNA rules.
 
     if (san)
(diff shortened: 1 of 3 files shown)


src/Native/System.Security.Cryptography.Native/openssl.c
```diff
@@ -939,6 +938,100 @@ CheckX509Hostname(
...
+            }
+
+            ipAddr = sanEntry->d.iPAddress;
+
+            if (!ipAddr || !ipAddr->data || ipAddr->length != addressBytesLen)
+            {
+                continue;
+            }
+
+            if (!memcmp(addressBytes, ipAddr->data, (size_t)addressBytesLen))
```
stephentoub: I assume we don't have to worry about byte order?

stephentoub:
LGTM

src/Native/System.Security.Cryptography.Native/openssl.c
bartonjs: RFC 3280 defines that iPAddress values within the subjectAltName extension are in network-byte-order.  And while the MSDN documentation is a bit sparse for IPAddress.GetAddressBytes(), it seems to be returning things in network-byte-order.

stephentoub: Sounds good.

vijaykota:
LGTM. Sorry about the delay

PR #3835, opened 2015-10-14
Adding InstructionName overrides where missing.

### LINKED ISSUE #3834: Expression interpreter - Some instructions lacking an InstructionName
Makes it hard to debug the emitted code. E.g. AssignLocalInstruction lacks an override for InstructionName.

bartdesmet:
This fixes #3834.

We could also reduce code size (if we care about that) by making Instruction.InstructionName use this.GetType to distill a meaningful name (e.g. ignoring nested private types such as AddInt32 and traverse to the first type in the hierarchy that ends with "Instruction"). Given this is only used at debug time, the added cost wouldn't be an issue.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Instruction.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/Interpreter/Instruction.cs
@@ -36,9 +36,9 @@ namespace System.Linq.Expressions.Interpreter
 
         public abstract int Run(InterpretedFrame frame);
 
-        public virtual string InstructionName
+        public abstract string InstructionName
         {
-            get { return "<Unknown>"; }
+            get;
         }
 
         public override string ToString()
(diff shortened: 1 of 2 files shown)


bartdesmet:
CC @VSadov

stephentoub:
> We could also reduce code size (if we care about that)

It's not our top priority, but we do care about it.  More than code size, however, it seems more resilient against refactorings and the like if we don't have to hardcode such names.  If we can avoid doing so, decrease code size, and not impact release performance, sounds like a win-win.

VSadov:
I am not sure how that would work on ProjectN. By default the metadata is stripped off if proven to be not necessary. Once we start reflecting over instruction types, we will have to directly or indirectly force the inclusion of metadata for these.

I think returning strings is better form this prospective. 

As for the resilience to refactoring. Once we can use C#6  nameof feature, it will likely be less of a concern

stephentoub:
> I am not sure how that would work on .Net native

That's fair. Ok.

VSadov:
The change seems pretty straightforward. 
LGTM

PR #3847, opened 2015-10-14
Add SecurityTransparentAttribute to System.Collections.Immutable

### LINKED ISSUE #1512: System.Collections.Immutable and APTCA/SecurityTransparentAttribute
If possible please add one of attributes forcing assembly to be Security Transparent (AllowPartiallyTrustedCallersAttribute or SecurityTransparentAttribute).

This will help using this library in some quite uncommon partial trust scenarios.

### LINKED ISSUE #3293: System.Collections.Immutable assembly considered Security Critical
I'm writing a program where some untrusted code is running in a separate app domain.  I want this untrusted code to use various Immutables, but I get the following error:

`Attempt by security transparent method 'MyMethod' to access security critical method 'System.Collections.Immutable.ImmutableList.CreateBuilder<MyObject>()' failed.`

`Assembly 'MyAssembly, Version=1.0.0.0, Culture=neutral, PublicKeyToken=**** is partially trusted, which causes the CLR to make it entirely security transparent regardless of any transparency annotations in the assembly itself.  In order to access security critical code, this assembly must be fully trusted.`

Why is the Immutable assembly considered security critical?  I would imaging this is an oversight and it should be Security Transparent.

AArnott:
I'm currently soliciting verification results from @NArnott 

Fix #3293
Fix #1512

### DIFF
--- a/src/System.Collections.Immutable/src/Properties/SecurityTransparent.cs
+++ b/src/System.Collections.Immutable/src/Properties/SecurityTransparent.cs
@@ -1,0 +1,6 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Security;
+
+[assembly: SecurityTransparent]
--- a/src/System.Collections.Immutable/src/Properties/InternalsVisibleTo.cs
+++ b/src/System.Collections.Immutable/src/Properties/InternalsVisibleTo.cs
@@ -1,8 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-//------------------------------------------------------------
-
-//------------------------------------------------------------
 
 using System.Runtime.CompilerServices;
 
--- a/src/System.Collections.Immutable/src/System.Collections.Immutable.csproj
+++ b/src/System.Collections.Immutable/src/System.Collections.Immutable.csproj
@@ -36,6 +36,7 @@
   <ItemGroup>
     <Compile Include="Properties\InternalsVisibleTo.cs" />
     <Compile Include="GlobalSuppressions.cs" />
+    <Compile Include="Properties\SecurityTransparent.cs" />
     <Compile Include="System\Collections\Generic\IHashKeyCollection.cs" />
     <Compile Include="System\Collections\Generic\ISortKeyCollection.cs" />
     <Compile Include="System\Collections\Immutable\AllocFreeConcurrentStack.cs" />


src/System.Collections.Immutable/src/Properties/SecurityTransparent.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Security;
+
+[assembly: SecurityTransparent]
```
stephentoub: @weshaggard, @joshfree, are there any other corefx assemblies we should be doing this with?  I'm wondering if we should move this file to src/Common.

stephentoub:
LGTM

src/System.Collections.Immutable/src/Properties/SecurityTransparent.cs
weshaggard: I'm not entirely sure what the implications of this are so I cannot say definitively. Unless someone does some proper test validation and understands the scope I don't think I'm comfortable with blindly doing this.

stephentoub: > I don't think I'm comfortable with blindly doing this

Thanks... just to be clear, are you referring to this specific change to immutable collections, or are you referring to my question about doing it elsewhere as well?

weshaggard: I was referring to the general case. I'm OK with individual libraries adding this if the right testing and security audits are done (i.e. we know we don't call critical code without proper checks and annotations).

weshaggard: Actually thinking a little more I kind of recall @jkotas suggesting we remove all security annotations for our libraries so perhaps he has an opinion on whether we should be marking this transparent or not.

stephentoub: This library (and some of the others in corefx) seem like special-cases, though, as they are meant to be usable with runtimes that do enforce security annotations.

jkotas: Supporting partial trust is not as simple as simple as adding a security transparent annotation.

You have to also make sure that the assembly actually works in partial trust, e.g. that  the security transparent code is not calling security critical code - both statically using [secanotate tool](http://blogs.msdn.com/b/shawnfa/archive/2009/11/18/using-secannotate-to-analyze-your-assemblies-for-transparency-violations-an-example.aspx), or dynamically (by running tests in partial trust environment).

Paying for this extra burden would be prohibitive across the board for corefx, paying for it in a few special cases should be ok.

AArnott: Thanks, @jkotas. I'll read up on that tool and try to run it against this particular library.

AArnott: The secannotate report is clean:

> Annotating 'System.Collections.Immutable'.
Beginning pass 1.
Pass complete, 0 new annotation(s) generated.

If I was more comfortable with the build process and knew it would work either cross platform or could constrain it to only happen on Windows, I could wire up secannotate to run as part of the build or test process. But as I think this library is unlikely to make any calls to Critical code, it's probably overkill to guard against introducing such unless the cost is low.

AArnott:
@stephentoub I was waiting for @NArnott to sign off on the effectiveness of this fix in a partial trust application prior to merging this.

stephentoub:
Oops, my bad. I can revert it out if you'd like.

AArnott:
Let's wait and see what @NArnott says before backing it out. I'm hopeful it will work. :)

PR #3848, opened 2015-10-14
Remove internal Validation namespace

### LINKED ISSUE #2278: Rename Validation to Microsoft.Internal.Validation
The C# (and possibly VB compiler) will match the top-level namespace `Validation` in `System.Collections.Immutable` even though it only contains internal types. This prevents customers from using a type named `Validation` in their code. Specifically, assume a customer has defined the type like this:

```C#
namespace Support
{
    public static class Validation
    {
	    public static bool IsValid(string x) { return true; }
    }
}
```
Consuming this type stops working when `System.Collections.Immutable` is referenced:

```C#
using System;
using System.Linq;
using System.Text;

using Support;

namespace ImmutableTest
{
    public class Experiment
    {
        // With out Immutable included, the below works just fine.
        // Once I include immutable, Validation, which was really
        // Support.Validation, but was found because of the using
        // statement above, it starts to fail showing that it can't
        // find IsValid in the namespace Validation.
	public void Test1() { Validation.IsValid("AS"); }
	public void Test2() { Support.Validation.IsValid("AS"); }
    }
}
```

According to @jaredpar this behavior is by-design:

> This behavior is by design for C# at least.  A simple name lookup will prefer matching namespaces over types (section 7.5.2).   I didn’t check the VB spec but it almost assuredly has the same design here.  

In order to avoid that conflict, we should rename the `Validation` namespace to something like `System.Validation`.

AArnott:
This moves the two internal types in the "Validation" namespace into the System.Collections.Immutable namespace so that we don't have any 'internal only' namespaces in the assembly.

Fix #2278

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/DictionaryEnumerator.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/DictionaryEnumerator.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
-using Validation;
 
 namespace System.Collections.Immutable
 {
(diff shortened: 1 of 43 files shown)


stephentoub:
LGTM.  Thanks.

PR #3851, opened 2015-10-14
Shim X509Ext methods.

eerhardt:
Adding shims for all methods in Interop.X509Ext.cs.

I also moved the decoding of X509 Basic Constraints into native code
so we don't have to marshal structs from native to managed.
This also saves on a few native calls.

I refactored some Basic Constraints tests to use a Theory, and split
the test into 2 - encoding and decoding. Encoding doesn't work on
Unix yet, so this allows the decoding test to run.


stephentoub:
LGTM

src/System.Security.Cryptography.X509Certificates/tests/ExtensionsTests.cs
```diff
@@ -213,39 +214,56 @@ public static void BasicConstraintsExtensionDefault()
...
+            X509BasicConstraintsExtension ext = new X509BasicConstraintsExtension(new AsnEncodedData(rawData), critical);
+            Assert.Equal(certificateAuthority, ext.CertificateAuthority);
+            Assert.Equal(hasPathLengthConstraint, ext.HasPathLengthConstraint);
+            Assert.Equal(pathLengthConstraint, ext.PathLengthConstraint);
         }
 
-        [Fact]
-        [ActiveIssue(1993, PlatformID.AnyUnix)]
-        public static void BasicConstraintsExtension_CA_559()
+        public static IEnumerable<object[]> BasicConstraintsData
```
nguerrera: InlineData would be nicer for a structure this simple IMHO,

nguerrera:
LGTM

sokket:
LGTM - please run the format-code.sh script before merging :)

src/System.Security.Cryptography.X509Certificates/tests/ExtensionsTests.cs
eerhardt: I thought about that, but decided to use MemberData so I could reuse the data and didn't have to duplicate it across the two Theories. Do you think it would be leaps and bounds better to use InlineData?

stephentoub: You could compromise and make it a field to trim down on some of the ceremony, e.g.
```C#
public static object[][] BasicConstraintsData
{
    new object[] { false, false, 0, false, "3000" },
    new object[] { true, false, 0, false, "30030101ff" },
    new object[] { false, true, 0, false, "3003020100" },
    new object[] { false, true, 7654321, false, "3005020374cbb1" },
    new object[] { true, true, 559, false, "30070101ff0202022f" }
}
```

nguerrera: Oh, I missed the repeated usage. I thought it was just used once. The field is a little nicer, but I'm fine with what you have.

eerhardt: Changed to a field. Thanks for the suggestion.

src/Native/System.Security.Cryptography.Native/pal_x509ext.cpp
```diff
@@ -3,14 +3,63 @@
 
 #include "pal_x509ext.h"
 
-extern "C" BASIC_CONSTRAINTS* DecodeBasicConstraints(const unsigned char* buf, int32_t len)
+#include <assert.h>
+
+extern "C" X509_EXTENSION* X509ExtensionCreateByObj(ASN1_OBJECT* obj, ASN1_OCTET_STRING* data)
```
bartonjs: Please leave the boolean isCritical exposed on the function.  I'll probably need it while implementing the various extension Encode methods.

src/Native/System.Security.Cryptography.Native/pal_x509ext.cpp
```diff
@@ -3,14 +3,63 @@
...
+}
+
+extern "C" int32_t DecodeX509BasicConstraints2Extension(
+    const unsigned char* encoded,
+    int32_t encodedLength,
+    int32_t* certificateAuthority,
+    int32_t* hasPathLengthConstraint,
+    int32_t* pathLengthConstraint)
+{
+    *certificateAuthority = false;
```
bartonjs: You should check that they weren't nullptr before dereferencing them.

PR #3852, opened 2015-10-14
Fix three issues in ZipArchiveEntry from #3820

stephentoub:
- AllowLargeZipArchiveEntriesInUpdateMode was returning whether the size of a pointer was >= 4.  That's going to be true on both 32-bit and 64-bit, when the intention was for it to only be true on 64-bit.
- AllowLargeZipArchiveEntriesInUpdateMode was a property.  Making it a readonly static bool allows the JIT compiler to essentially treat it as a constant, trim away unnecessary branches, etc.
- WriteLocalFileHeaderAndDataItemIfNeeded was unnecessarily creating a MemoryStream around a byte[] just to be able to CopyTo that stream to another one.  We can just write the bytes directly.

cc: @FiveTimesTheFun


stephentoub:
@dotnet-bot test this please (native mac build was aborted for some reason)

FiveTimesTheFun:
Thank you Stephen, looks good to me, I'm sick today hope to be in the office and online tomorrow.

stephentoub:
Thanks, Matt.  Feel better.

PR #3856, opened 2015-10-14
Fix exception documentation on System.Collections.Immutable

AArnott:
I noticed in making another change that a bunch of incorrect xml docs describing exceptions existed. This fixes those up by removing the invalid ones and adding explanations for the valid ones.

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1.cs
@@ -210,6 +210,7 @@ namespace System.Collections.Immutable
         /// </summary>
         /// <param name="index">The 0-based index of the element in the set to return.</param>
         /// <returns>The element at the given position.</returns>
+        /// <exception cref="IndexOutOfRangeException">Thrown from getter when <paramref name="index"/> is negative or not less than <see cref="Count"/>.</exception>
         public T this[int index]
         {
             get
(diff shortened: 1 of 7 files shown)


src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1.cs
```diff
@@ -1145,7 +1146,7 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
         /// </summary>
         /// <param name="index">The index.</param>
         /// <param name="item">The item.</param>
-        /// <exception cref="System.NotSupportedException"></exception>
+        /// <exception cref="NotSupportedException">Always thrown from the setter.</exception>
```
stephentoub: "setter" doesn't seem like the right description here

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1.cs
```diff
@@ -1155,7 +1156,7 @@ void IList<T>.Insert(int index, T item)
         /// Removes the value at the specified index.
         /// </summary>
         /// <param name="index">The index.</param>
-        /// <exception cref="System.NotSupportedException"></exception>
+        /// <exception cref="NotSupportedException">Always thrown from the setter.</exception>
```
stephentoub: Ditto

stephentoub:
Two minor nits, otherwise LGTM.

AArnott:
Thanks for the review. I'll merge after automated checks complete.

PR #3857, opened 2015-10-14
Shim ip_mreqn, ipv6_mreq, and linger.

pgavlin:
These types are used for getting and setting IP multicast options.


pgavlin:
@stephentoub @sokket @nguerrera PTAL.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -703,3 +708,119 @@ extern "C" int32_t TryGetIPPacketInformation(uint8_t* messageHeader, int32_t isI
...
+            return true;
+
+        default:
+            return false;
+    }
+}
+
+extern "C" int32_t GetIPv4MulticastOption(int32_t socket, int32_t multicastOption, IPv4MulticastOption* option)
+{
+    if (option == nullptr)
```
sokket: This should not happen, right? Should this be an assert then?

sokket: And everywhere below

src/Native/System.Native/pal_networking.h
```diff
@@ -58,6 +58,12 @@ enum AddressFamily : int32_t
     PAL_AF_INET6 = 23, // System.Net.AddressFamily.InterNetworkV6
 };
 
+enum MulticastOption : int32_t
+{
+    PAL_MULTICAST_ADD = 0, // IP{,V6}_ADD_MEMBERSHIP
```
sokket: Do these values map directly to native values? If so, please static_assert that they do

src/Native/System.Native/pal_networking.h
```diff
@@ -90,6 +96,21 @@ struct IPPacketInformation
...
+    uint32_t LocalAddress;     // Local address
+    int32_t InterfaceIndex;    // Interface index
+    int32_t Padding;           // Pad out to 8-byte alignment
+};
+
+struct IPv6MulticastOption
+{
+    IPAddress Address;      // Multicast address
+    int32_t InterfaceIndex; // Interface index
+    int32_t Padding;        // Pad out to 8-byte alignment
```
sokket: Rather than add the padding here, which is specific per-compilation-platform, can you pragma ignore the padding warning?

sokket:
A few comments but LGTM. Please run the format-code.sh script to make sure the C++ code adheres to the formatting rules (I saw a few things that would be changed) before merging

src/Native/System.Native/pal_networking.h
pgavlin: No: @stephentoub and I spoke about this and decided that the explicit padding is necessary. We're going to have a single definition of the type in IL, so we must be sure that the native definition matches, even on platforms where the padding is unnecessary. We will need to rethink this in the remote chance that we encounter a platform that requires > 8 byte alignment.

src/Native/System.Native/pal_networking.h
pgavlin: No, they do not.

src/Native/System.Native/pal_networking.cpp
pgavlin: No: this is effectively public surface area, so we should be coding defensively against upstream bugs. There are other portions of this file that need to be updated to do this as well.

pgavlin:
Re: code formatting: there are two items in this file that would be changed, but there are a number of other files that also have changes when the script is run.

I'd prefer not to run the formatter until we either a) add it to the native build, or b) add a CI leg to verify that the formatting is correct so these errors are caught before merging. Formatting changes to unrelated components make PRs too difficult to grok, esp. with GH's diff view.

src/Native/System.Native/pal_networking.h
sokket: Cool, works for me

src/Native/System.Native/pal_networking.cpp
sokket: Cool, sounds good

sokket:
@pgavlin If you rebase on master, a recent PR should have fixed the other files

jasonwilliams200OK:
> add a CI leg to verify that the formatting is correct so these errors are caught before merging

+1 for CI integration. This is a common practice in node.js world, where CI first runs a code linter (`jshint`) and code style cop (`jscs`) and fails early if they don't comply. Similarly, there exists [`cpplint`](https://github.com/tkruse/cpplint) and [c# lint](https://github.com/hach-que/cstools) too which can be consumed with CI. :)

src/Native/System.Native/pal_networking.cpp
```diff
@@ -13,6 +13,19 @@
 #include <sys/socket.h>
 #include <unistd.h>
 #include <vector>
+#include <errno.h>
+
+#if defined(__APPLE__)
```
jasonwilliams200OK: Out of curiosity, is this check required? The guard `define`s can be used by other kinds of Unices as well which lack `IPV6_[ADD/DROP]_MEMBERSHIP`.

pgavlin: The check is not required, it's simply being explicit about the target OS. I wondered if the BSDs would also be missing these definitions.

jasonwilliams200OK: Yup, without guard, FreeBSD with Clang37 complains:

```
/root/corefx/src/Native/System.Native/pal_networking.cpp:731:35: error: use of undeclared identifier 'IPV6_ADD_MEMBERSH$
            optionName = isIPv6 ? IPV6_ADD_MEMBERSHIP : IP_ADD_MEMBERSHIP;
                                  ^
```

There are other errors besides this, but I think they are out of this PR's scope. :)

jasonwilliams200OK:
Test this please (win-release build timed-out: [gist with link](https://gist.githubusercontent.com/jasonwilliams200OK/e3e3c920055db7df9b40/raw/9f7f69ecb0cef55f75320945351812e677288718/corefx-ci-failure.log)).

PR #3859, opened 2015-10-14
Add string.ToLower/UpperInvariant test for ASCII

stephentoub:
Validate a string with all of the ASCII characters converted to lower and upper case with invariant culture.

cc: @ellismg 
Related to https://github.com/dotnet/coreclr/pull/1747

### DIFF
--- a/src/System.Runtime/tests/System/String.cs
+++ b/src/System.Runtime/tests/System/String.cs
@@ -1250,7 +1250,7 @@ public static unsafe class StringTests
     }
 
     [Fact]
-    public static void TestToLowerToUpper()
+    public static void TestToLowerToUpper_Basic()
     {
         //@todo: Add tests for ToLower/Upper(CultureInfo) when we have better support for
         // getting more than just the invariant culture. In any case, ToLower() and ToUpper()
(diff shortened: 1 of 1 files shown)


src/System.Runtime/tests/System/String.cs
```diff
@@ -1273,6 +1273,37 @@ public static void TestToLowerToUpper()
...
     [Fact]
+    public static void TestToLowerToUpperInvariant_ASCII()
+    {
+        char[] asciiChars = new char[128];
+        char[] asciiCharsUpper = new char[128];
+        char[] asciiCharsLower = new char[128];
+
+        for (int i = 0; i < asciiChars.Length; i++)
+        {
+            asciiChars[i] = asciiCharsUpper[i] = asciiCharsLower[i] = (char)i;
```
eerhardt: Why assign asciiCharsUpper/Lower[i] here and then reassign it below?  What about something like:
```C#
for (int i = 0; i < asciiChars.Length; i++)  
{
    char c = (char)i;
    asciiChars[i] = c;  

    // Purposefully avoiding char.ToUpper/ToLower here so as not  
    // to use the same thing we're testing.  
    asciiCharsLower[i] = (c >= 'A' && c <= 'Z') ? (char)(c - 'A' + 'a') : c;
    asciiCharsUpper[i] = (c >= 'a' && c <= 'z') ? (char)(c - 'a' + 'A') : c;
}
```

eerhardt:
Just one minor comment.  LGTM either way.

src/System.Runtime/tests/System/String.cs
stephentoub: > Why assign asciiCharsUpper/Lower[i] here and then reassign it below?

Just for simplicity.  I'll change it to what you have.

PR #3860, opened 2015-10-14
Fix ImmutableList<T> IList implementation for incompatible element types

### LINKED ISSUE #1414: ImmutableList<T> (and Builder) IList.Contains, IndexOf, Remove methods throw instead of return -1 given unexpected argument types
A bit longer story

http://blog.rthand.com/post/2015/04/10/A-hidden-bug-in-Windows-Stores-ComboBox.aspx

Basically a Windows Store (8.1 and 10 at least) ComboBox first tries ComboBoxItem against its ItemsSource IList.IndexOf and if that gracefully fails (returns -1) then it tries with the actual type. This works with normal List<T> because this guy returns -1 when an incompatible type is used, where ImmutableList<T> throws an exception (doesn't do any check actually) and ComboBox doesn't understand that very well.
To make it worse ComboBox doesn't perform this stunt immediately but only after a while (when one scrolles some of its items).

Now, I'd say that this is ComboBox' fault but who knows.

AArnott:
Add 15 unit tests to cover the `ImmutableList<T>` and `ImmutableList<T>+Builder` class's implementation of `IList.Contains`, `IList.Remove`, and `IList.IndexOf` for cases where the `object` argument is not of a compatible type with the type argument for the collection.

All 15 of the added unit tests fail. Until the final commit, which fixes all the issues the tests demonstrated.

Fix #1414

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1+Builder.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList`1+Builder.cs
@@ -1051,7 +1051,12 @@ namespace System.Collections.Immutable
             /// <exception cref="System.NotImplementedException"></exception>
             bool IList.Contains(object value)
             {
-                return this.Contains((T)value);
+                if (IsCompatibleObject(value))
+                {
+                    return this.Contains((T)value);
+                }
+
+                return false;
             }
 
             /// <summary>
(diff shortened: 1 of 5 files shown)


jasonwilliams200OK:
Test this please.

AArnott:
@stephentoub can you take a look?

stephentoub:
LGTM

AArnott:
Thanks for the review.

Issue #2221, opened 2015-07-02
Replacement for unavailable Assembly methods


lleraromero:
Hi,

How is the supported way to deal with the Assembly class? For instance, I have some method calls like Assembly.Location or Assembly.GetCallingAssembly() which I couldn't find a way to bring those calls to CoreFX.

Thanks.

nguerrera:
Assembly.GetCallingAssembly() has very bad semantics in the face of inlining: see remarks in https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getcallingassembly(v=vs.110).aspx. If you're using this for logging/diagnostics, consider the much cheaper caller information attributes: https://msdn.microsoft.com/en-us/library/hh534540.aspx. If you're using this to make code-path decisions, I'd urge you to reconsider based on the poorly defined semantics of GetCallingAssembly.

Assembly.Location is problematic for two reasons: not all assemblies (consider loaded from byte[]) have locations and on some platforms (.NET Native), none do. What is your use case for Assembly.Location?

@jkotas @gkhanna79 how would you feel about adding Assembly.TryGetLocation()? It just keeps coming up and it seems a lot of libraries rely on being able to find other files that are meant to be placed alongside their dlls. (Just guessing for now that is the use-case here too.) There is AppContext.BaseDirectory, but that is not necessarily the library directory, and I think for DNX it never is.

jkotas:
What does `Assembly.TryGetLocation()` return?

jkotas:
> not all assemblies (consider loaded from byte[]) have locations 

Assembly.Location deals with this case just fine - by returning empty string.


>  Assembly.TryGetLocation()

I do not have a problem with it. The signature is different, but it would otherwise do the exact same thing as Assembly.Location on CoreCLR - correct?

jkotas:
> Assembly.TryGetLocation()

We may want to let AssemblyLoadContext override what it returns ... I guess DNX may need it to handle the in-memory compilation, etc.

nguerrera:
> I do not have a problem with it. The signature is different, but it would otherwise do the exact same thing as Assembly.Location on CoreCLR - correct?

Yes. I actually meant `bool TryGetLocation(out string location)`, just be to draw more attention to the fact that you have to deal with it not being available, and opening up the possibility of putting it elsewhere than core reflection as an extension.

However, on second thought, exposing Assembly.Location as-is would be much simpler and .NET Native could simply return String.Empty.  Maybe that is best. What do you think?

[...]

PR #3861, opened 2015-10-14
Expose Assembly.Location via public contract

nguerrera:
After some discussion, it seems we should just expose Assembly.Location after all. It already does not guarantee a path (string.Empty for load from stream/byte[], NotSupportedException for dynamic assembly). On .NET Native, we can simply return String.Empty.

This has been requested many times over: #2221, dotnet/coreclr#1367 (+ lots of requests offline).

Note that there are still scenarios not covered such as getting source path in DNX when compiled on the fly, but I think we can continue to discuss that separately. I believe those scenarios are better served by a separate API rather than an Assembly.Location override.

cc @weshaggard @jkotas


src/System.Reflection/tests/project.json
```diff
@@ -1,13 +1,16 @@
...
   "dependencies": {
+    "System.AppContext": "4.0.0",
     "System.Collections": "4.0.10",
     "System.Diagnostics.Debug": "4.0.10",
     "System.Globalization": "4.0.10",
     "System.IO": "4.0.10",
+    "System.IO.FileSystem": "4.0.0.0",
     "System.Linq": "4.0.0",
-    "System.Reflection": "4.0.10",
+    "System.Reflection.Emit": "4.0.0",
```
weshaggard: This + the System.Runtime.Loader dependency just caused these tests to not be runnable on .NET Native. We should consider splitting these out so we can still use these reflection tests on more of our platforms.

nguerrera: Pragmatically, all 3 tests are actually dependent on not being run on .NET Native at the moment. The .NET Native test will just be `Assert.NotNull(ThisAssembly().Location); Assert.Empty(ThisAssembly.Location())`. However, we don't currently share enough infrastructure to express that directly here in corefx repo.

src/System.Reflection/tests/project.json
```diff
@@ -1,13 +1,16 @@
 {
   "dependencies": {
+    "System.AppContext": "4.0.0",
     "System.Collections": "4.0.10",
     "System.Diagnostics.Debug": "4.0.10",
     "System.Globalization": "4.0.10",
     "System.IO": "4.0.10",
+    "System.IO.FileSystem": "4.0.0.0",
```
nguerrera: This should be "4.0.0" (no fourth part) for consistency. I'm surprised this even works. ;)

weshaggard: nuget/dnu just ignores the 4th part :)

src/System.Reflection/tests/project.json
weshaggard: All I want to be careful about is that when we start running these test on .NET Native that we don't loose all the reflection tests because of these new tests. If we think that might happen perhaps we should factor some of these out into another test project.

nguerrera: Fair point. Split out the tests in to a separate project.

jkotas:
Does it also need to be flipped to public in https://github.com/dotnet/coreclr/blob/master/src/mscorlib/model.xml ?

nguerrera:
@jkotas  Hmm, I imagined it was already there as tests are passing and I see public property and its getter in ildasm of coreclr mscorlib. I am now trying to understand what is making that happen without the entry where I'd expect it to be in model.xml...

src/System.Reflection/System.Reflection.sln
```diff
@@ -1,10 +1,14 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22609.0
+VisualStudioVersion = 14.0.23107.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Reflection.Tests", "tests\System.Reflection.Tests.csproj", "{B027C72E-F04E-42E0-A7F7-993AEF8400D2}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Reflection", "ref\System.Reflection.csproj", "{446DE6F0-47A0-4C2F-A918-768A76CCF461}"
```
stephentoub: Nice.  We should probably update other .sln fils to include the ref projects as well.

davidfowl:
golf clap :clap:

src/Common/tests/Data/TinyAssembly.il
```diff
@@ -0,0 +1,2 @@
+.assembly TinyAssembly {}
```
stephentoub: Can we somehow build TinyAssembly.dll as part of the build process rather than checking in the binary?

stephentoub:
One question, otherwise LGTM.

nguerrera:
@stephentoub I'm also using it in #3896 to get a known-guid -- i.e. depending on it not changing. (C#'s new deterministic feature would help with that eventually.)

The file is ~1.5KB uncompressed and ~500B compressed. The contents are deliberately as minimal as possible and I don't think it will change much if at all.

S.R.Loader also has source for a small assembly, but also checks in the dll to embed the resource. (That can probably be fixed by using ProjectReference with OutputItemType=EmbeddedResource, mind you.) We're embedding as resource in both cases to get around corerun putting everything in sight on the TPA list.

src/System.Reflection/tests/CoreCLR/System.Reflection.CoreCLR.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <Compile Include="Assembly\Assembly_LocationTests.CoreCLR.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="..\..\..\Common\tests\Data\TinyAssembly.dll">
+      <Link>Assembly\TinyAssembly.dll</Link>
+      <LogicalName>TinyAssembly.dll</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\ref\System.Reflection.csproj">
```
weshaggard: I assume you want to use a package reference once these changes get published? If so please file a tracking issue.

weshaggard:
LGTM

nguerrera:
I submitted https://github.com/dotnet/coreclr/pull/1783 to be explicit in the model file.

terrajobst:
@weshaggard, @KrzysztofCwalina and myself just took a look. Approved :+1:

nguerrera:
Added missing copyright headers and caught up to master in rebase. No other changes since LGTMs. I will merge when tests pass.

stephentoub:
@dotnet-bot test this please (https://github.com/dotnet/corefx/issues/3915)

PR #3864, opened 2015-10-14
Add ICustomMarshaler to the contract

### LINKED ISSUE #3863: Add System.Runtime.InteropServices.ICustomMarshaler to the contract
c.f. [coreclr #1767](https://github.com/dotnet/coreclr/issues/1767)

The ICustomMarshaler interface needs to be added to the contract so that people can use it.

jazzdelightsme:
The ICustomMarshaler machinery exists in mscorlib, but the actual
interface has been marked internal. Once coreclr has been updated to
expose it, we need to add the type to the contract.

Fix #3863

### DIFF
--- a/src/System.Runtime.InteropServices/ref/System.Runtime.InteropServices.cs
+++ b/src/System.Runtime.InteropServices/ref/System.Runtime.InteropServices.cs
@@ -305,6 +305,14 @@ namespace System.Runtime.InteropServices
     {
         object GetUnderlyingObject();
     }
+    public partial interface ICustomMarshaler
+    {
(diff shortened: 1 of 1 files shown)


weshaggard:
cc @yizhang82 

This change would also require you to bump the version of System.Runtime.InteropSerivices to 4.1.0.0.

stephentoub:
@jazzdelightsme, thanks for opening this and the corresponding coreclr issue.  As this depends on the coreclr change and a decision on whether to expose this mechanism, I'm going to close out this PR until such a decision is reached.  If the decision is to expose this, please do reopen this PR or submit it as a new one (ideally there'd also be a test included to validate that it's functional). Thanks!

PR #3868, opened 2015-10-14
Shim {send,recv}msg and related types.

pgavlin:
Just what it says on the tin.


pgavlin:
@stephentoub @nguerrera @sokket PTAL

src/Native/System.Native/pal_networking.h
```diff
@@ -117,6 +117,23 @@ struct LingerOption
...
 
+struct IOVector
+{
+    uint8_t* Base;
+    uintptr_t Count;
+};
+
+struct MessageHeader
+{
+    uint8_t* SocketAddress;
```
eerhardt: I've been using the native type 'unsigned char' to map to the managed 'byte' type.  I think it makes it more obvious that it is a byte and not an integer. Does that make sense?

src/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs
```diff
@@ -581,16 +597,14 @@ private static unsafe int Send(int fd, int flags, byte[] buffer, ref int offset,
...
             {
-                errno = Interop.Sys.GetLastError();
                 return -1;
             }
 
-            errno = Interop.Error.SUCCESS;
-            offset += sent;
-            count -= sent;
-            return sent;
+            offset += (int)sent;
```
eerhardt: Maybe only cast sent to an int once.

Also, can you add either an assert or a "checked" cast, so no data is silently lost?

eerhardt: Or have Sys.SendMessage have an out int parameter and have native do the cast.

src/Common/src/Interop/Unix/System.Native/Interop.MessageHeader.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        internal unsafe struct IOVector
+        {
+            public byte* Base;
+            public UIntPtr Count;
```
eerhardt: I don't understand why this is UIntPtr.  Shouldn't it be an int, uint, long, or ulong?  It's not really a pointer/handle, right?

src/Native/System.Native/pal_networking.h
pgavlin: Please do not use `unsigned char`: the `char` type is not guaranteed to be exactly 8 bits in length. Please use `uint8_t` (and other explicitly-sized integral types as appropriate).

src/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs
pgavlin: I'll change this. I meant for all of these to be `checked` casts.

src/Common/src/Interop/Unix/System.Native/Interop.MessageHeader.cs
pgavlin: The layout of this type is intended to exactly match the layout of the platform's `struct iovec` in order to avoid expensive allocation + copying on each vectored receive or send. The layout of `struct iovec` is always
```
struct iovec {
   void* iov_base;
   size_t iov_len;
}
```

Hence the `UIntPtr`. I'll add a comment to the native definition (and I should point out that there are already asserts in this change to verify that the layouts of an `IOVector` and a `struct iovec` are identical).

src/Native/System.Native/pal_networking.h
nguerrera: Agree style-wise on uint8_t, but fwiw sizeof(unsigned char) == 1 is guaranteed IIRC. I think the standard allows for bytes/sizeof units to be more than 8 bits, but we'd have other problems in that exotic case.

pgavlin: Practically speaking, you're correct--any reasonable system I'm aware of uses 8 bits for a `char`. As mentioned, though, it's better to be explicit.

pgavlin:
@dotnet-bot test this please.

stephentoub:
@dotnet-bot test this please (OS X native build was aborted for some reason)

stephentoub:
LGTM

src/Common/src/Interop/Unix/System.Native/Interop.MessageHeader.cs
eerhardt: I was told by @nguerrera that we decided not to use IntPtr for size_t conversions. Here are the quotes from him:

>1. ... we pick a fixed size in the shim. Either 32 or 64 bit. If it's just a scratch buffer, etc. We use 32-bit.
2. However, if we really need 64-bit (memory mapped files, then we use that.
3. ...< snip >...  You can do that. There's no hard rule other than no variable size_t in shim sig, and no silent truncation.

pgavlin: While that's the correct tradeoff for many/most situations, it is not so here: in this case, I have specifically chosen `uintptr_t`/`UIntPtr` to avoid the aforementioned allocation + copying. Throughout the rest of the shim, I've used `uint64_t`/`ulong` in place of `size_t` (or `int64_t`/`long` in place of `ssize_t`).

eerhardt: :+1: OK - now I understand.

src/Native/System.Native/pal_networking.h
eerhardt: For my education - what about when marshalling Ansi strings? Is it OK to use `const char*` then?

pgavlin: AFAIK we're marshalling strings as sequences of UTF-8 code units on *nix, so I think `uint8_t` is still preferable (esp. since UTF-8 code units may span multiple bytes). At boundaries between shim and platform code, however, you may need to cast between `uint8_t*` and `char*` as per the platform signatures. @sokket has some ~~paranoid~~ good code [here](https://github.com/dotnet/corefx/pull/3868/files#diff-a533f7324e00c23d36788e9084520906R33) to ensure the legality of the cast.

nguerrera: AFAICT, that assert cannot fail on any standard-compliant system. sizeof(char) must be 1. uint8_t is exactly 8 bits but is only defined on systems that can support it. If CHAR_BIT were more than 8, what would sizeof(uint8_t) be? I think such a system would not be able to provide uint8_t so nothing using it would compile.

http://en.cppreference.com/w/cpp/language/sizeof
http://en.cppreference.com/w/cpp/types/integer

I would prefer const char* for strings and uint8_t* for buffers. I think the string casts will just clutter the code.

src/Native/System.Native/pal_networking.cpp
```diff
@@ -858,3 +891,60 @@ extern "C" Error SetLingerOption(int32_t socket, LingerOption* option)
     int err = setsockopt(socket, SOL_SOCKET, SO_LINGER, &opt, sizeof(opt));
     return err == 0 ? PAL_SUCCESS : ConvertErrorPlatformToPal(errno);
 }
+
+// TODO: shim flags
```
nguerrera: Is there a bug for this todo?

pgavlin: This TODO has already been addressed.

Issue #3066, opened 2015-09-03
Some Vector<SByte> tests overflow their test values on AVX2


mellinoe:
Some of the tests for Vector\<SByte\> inadvertently overflow some of the array data they generate, because they do some arithmetic on the index of each element. For example:
````CSharp
T[] values2 = new T[Vector<T>.Count];
for (int g = 0; g < Vector<T>.Count; g++)
{
    values2[g] = (T)(dynamic)(g * 5);
}
Vector<T> vec2 = new Vector<T>(values2);
````

On AVX2, there are 32 bytes in a Vector\<T\>. 32 * 5 = 160, which has far overflowed a signed byte. Anything over 25 elements here will overflow, which is breaking the tests, as they are doing GreaterThan/LessThan/etc. comparisons between the three arrays generated.

Actually, the (unsigned) Byte versions of these tests will also break once they run on AVX512, since 64-byte vectors will easily overflow an unsigned byte with this calculation, as well.

These should be rewritten to be more robust to this sort of issue.

Note: this issue doesn't show up on the builds here; we don't run the tests in a configuration that would hit this (AVX2, x64, Release)

mellinoe:
cc @CarolEidt 
Making sure I keep track of this to fix.

mellinoe:
This should be fairly trivial to fix, but I haven't had time yet, moving to future milestone.

PR #3869, opened 2015-10-14
Disable Vector<SByte> tests failing on AVX2

mellinoe:
This is failing because on AVX2, the number of elements in these SByte vectors is so large (32) that the vector construction logic is overflowing its inputs, because it is based on the index of the vector element.

See #3066


mellinoe:
cc: @weshaggard

weshaggard:
LGTM

PR #3871, opened 2015-10-15
Accept null comparers and update docs around these parameters

### LINKED ISSUE #2036: Exposed Immutable Collections APIs should allow `null` comparers
Several immutable collections APIs throw an `ArgumentNullException` if a null `IComparer<T>` or `IEqualityComparer<T>` is provided. This is not intuitive, and deviates from a very common pattern established by the .NET framework. All publicly-exposed API methods in this assembly should be updated to ensure the following:

1. If a null value is provided for `IComparer<T>`, `Comparer<T>.Default` should be used instead.
2. If a null value is provided for `IEqualityComparer<T>`, `EqualityComparer<T>.Default` should be used instead.

AArnott:
Follow .NET BCL patterns more closely with regard to accepting `IEqualityComparer<T>` and `IComparer<T>` parameters that have null arguments.

Fixes #2036

### DIFF
--- a/src/System.Collections.Immutable/tests/ImmutableListTestBase.cs
+++ b/src/System.Collections.Immutable/tests/ImmutableListTestBase.cs
@@ -282,6 +282,12 @@ namespace System.Collections.Immutable.Tests
         }
 
         [Fact]
+        public void Sort_NullComparison_Throws()
+        {
+            Assert.Throws<ArgumentNullException>(() => this.SortTestHelper(ImmutableList<int>.Empty, (Comparison<int>)null));
+        }
+
+        [Fact]
         public void SortTest()
         {
             var scenarios = new[] {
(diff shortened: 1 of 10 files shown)


src/System.Collections.Immutable/tests/ImmutableListTestBase.cs
```diff
@@ -419,5 +431,20 @@ private void TrueForAllTestHelper<T>(ImmutableList<T> list, Predicate<T> test)
             var actual = this.GetListQuery(list).TrueForAll(test);
             Assert.Equal(expected, actual);
         }
+
+        private class CustomComparer<T> : IComparer<T>
```
stephentoub: Could you use ```Comparer<T>.Create(comparison)``` instead of defining a custom comparer?

stephentoub:
One question, otherwise LGTM.

src/System.Collections.Immutable/tests/ImmutableListTestBase.cs
AArnott: Cool. I didn't know about that API. :)

stephentoub:
LGTM

Issue #1198, opened 2015-03-21
Improve code coverage for System.Linq.Expressions


stephentoub:
According to the current code coverage report, this library is only at ~40%. Our general goal is closer to 90%, though this will depend on the library. We should investigate to see whether the current coverage level is appropriate, and if not, improve our test suites to provide better coverage.

rafidka:
Since I am working on improving the coverage for System.Linq, can I get this one as well?

stephentoub:
@rafidka, this library is fairly different, so I'd be surprised if there was a lot of reuse possible.  You're still interested?  I love the enthusiasm, but do you want to hold off until you're further along with System.Linq?

rafidka:
Yeah, I know it is different actually, but since I have experience working on Compilers and Computer Algebra, I found this task interesting to work on.

I don't mind holding off until I make progress with System.Linq.

stephentoub:
Thanks to the efforts of multiple folks, this is now at ~84%.

PR #3876, opened 2015-10-15
Exclude unreachable members in System.Linq.Expressions from coverage.

JonHanna:
All such members are entirely unreachable, and coverage including them is not
informative.

### DIFF
--- a/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
+++ b/src/System.Linq.Expressions/src/System.Linq.Expressions.csproj
@@ -17,6 +17,9 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


JonHanna:
The smallest of steps on #1198, but worth doing and doesn't fit in well with any change focusing on a particular part of the assembly.

stephentoub:
I don't love using ExcludeFromCodeCoverage in this way, but I'm ok with it.  I did just open a PR #3879 to help ensure these attributes don't show up in the release builds.

cc: @VSadov

JonHanna:
I obviously enough do like using it, but that change is a great idea. There is value sometimes in running tests against release builds, and sometimes on running coverage on those, but since this `internal` version of the attribute isn't used that way, and doesn't depend on another assembly the way the System.dll version does, that's a great ida.

stephentoub:
@VSadov, up to you whether you want to take this.

stephentoub:
@VSadov, do you want to merge this or should we close it?

VSadov:
LGTM

Issue #2647, opened 2015-08-06
Eliminate managed -> native transition cost of errno conversion


nguerrera:
Currently, we P/Invoke whenever we need to go from raw errno to PAL value or vice versa. We could instead retrieve a shared table pointer once from native code and do the lookup from managed code.

**EDIT**: I think we should probably just return converted errors from shims: https://github.com/dotnet/corefx/pull/3877#issuecomment-148484741

In #2606, @stephentoub and I had the following conversation about it (can't manage to link to it):

@stephentoub 
> What would you think of exposing a function to retrieve the whole conversion table once, and then doing all of these conversions purely in managed without having to make a managed->native call each time? I realize these are on an error path, so performance isn't as critical, but there are cases where things can fail that aren't actually failures, e.g. EINTR for I/O operations.

@nguerrera  
>>I thought about this and was concerned about duplicating the table in each library (since we just compile the Interop.* in to each), but there aren't really that many errors so maybe it's OK after all.

 @stephentoub 
>That's a fair point. Up to you. Either way, we'll want to re-evaluate as we do more performance testing/tuning.
>
>(Note that it could probably be done in a way that has the table shared across libraries, e.g. if the native code created the table and returned a pointer to it to managed in some format the managed code could easily use as-is.)

@nguerrera 
>> I'll file an issue to investigate, but stick to the current approach for the initial commit. You're right that it could be shared that way -- I hadn't thought of it. It is slightly tricky because we'd need to share something that is efficient to lookup in directly (whereas I was imagining building the lookup from an unordered list in managed code). The Pal -> Platform case can just be an index into an array so that's easy. However, the more interesting and common Platform -> Pal case can be sparse. I'd be inclined to use a sorted array of (int, Error) + binary search to keep the shared data structure trivial. It'll take me some time to measure, etc., and it's just complex enough to warrant its own review IMHO.

kapilash:
Would it be against architecture/design of corefx to use a pre-processing tool for solving this problem?
Say, we have a program that would take , as input, a file containing c# code interspersed with  c constructs and generates the corresponding C# file with the values populated for the target architecture:


      [C# + C ]  -> tool (On target architecture)   -> C# 
                                    
 
 
Such a  tool would need to run at the corefx build time on the target architecture.
 

This is analogous to the use of [hsc2hs in Haskell ecosystem](https://hackage.haskell.org/package/hsc2hs).

stephentoub:
[...]

PR #3877, opened 2015-10-15
Shim various socket methods.

pgavlin:
This change adds shims for most of the remaining straightforward methods
in sockets:
- accept
- bind
- connect
- getpeername
- getsockname
- getsockopt
- listen
- setsockopt
- shutdown


pgavlin:
@stephentoub @nguerrera @sokket PTAL

nguerrera:
LGTM. 

Side discussion: I notice we've taken a different direction on converting error codes, but I do like what you have. The original pattern of GetLastError() doing conversion made it easier to bootstrap with most of libc still unshimmed. We had various helpers like SysCall(...) that did error handling and it made it such that they could be used uniformly acrosss shimmed and unhsimmed calls. However, it had the obvious perf drawback of extra P/Invoke on error. Now that we're approaching shimming completion that advantage is basically moot and I think we can easily switch over to returning Error throughout. That would probably be the simplest way to close out #2647.

PR #3878, opened 2015-10-15
Add Sort(Comparison<T>) methods to ImmutableArray and Builder

### LINKED ISSUE #2167: Add ImmutableArray.Builder.Sort(Comparison<T> comparison)
List<T> has an overload of Sort that takes a delegate. It's useful when writing an inline comparer using a lambda:

```C#
list.Sort((x, y) => <comparison>);
```

AArnott:
ImmutableList and Builder has these Sort overloads, as does Array itself. This makes this same functionality easily accessible from ImmutableArray.

Fix #2167

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
@@ -636,6 +636,25 @@ namespace System.Collections.Immutable
             }
 
             /// <summary>
+            /// Sorts the elements in the entire array using
+            /// the specified <see cref="Comparison{T}"/>.
+            /// </summary>
+            /// <param name="comparison">
+            /// The <see cref="Comparison{T}"/> to use when comparing elements.
+            /// </param>
+            /// <exception cref="ArgumentNullException"><paramref name="comparison"/> is null.</exception>
+            [Pure]
+            public void Sort(Comparison<T> comparison)
+            {
+                Requires.NotNull(comparison, "comparison");
+
+                if (Count > 1)
+                {
+                    Array.Sort(_elements, comparison);
+                }
+            }
+
+            /// <summary>
             /// Sorts the array.
             /// </summary>
             /// <param name="comparer">The comparer to use in sorting. If <c>null</c>, the default comparer is used.</param>
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -922,6 +922,24 @@ namespace System.Collections.Immutable
         }
 
         /// <summary>
+        /// Sorts the elements in the entire <see cref="ImmutableArray{T}"/> using
+        /// the specified <see cref="Comparison{T}"/>.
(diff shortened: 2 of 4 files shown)


AArnott:
@richlander @terrajobst @stephentoub  I'm not sure what the current process is for the API review that may need to happen for this. I think it's very straightforward but I want to follow whatever process is necessary.

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
```diff
@@ -922,6 +922,24 @@ public ImmutableArray<T> Sort()
...
+        /// </param>
+        /// <returns>The sorted list.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="comparison"/> is null.</exception>
+        [Pure]
+        public ImmutableArray<T> Sort(Comparison<T> comparison)
+        {
+            Requires.NotNull(comparison, "comparison");
+
+            var self = this;
+            return self.Sort(Comparer<T>.Create(comparison));
```
stephentoub: While this is convenient, it does mean a new IComparer is being allocated on each call to Sort(Comparison).

stephentoub:
> I'm not sure what the current process is for the API review that may need to happen for this.

@terrajobst should comment.  Immo?  I'm also not clear on the process as it stands for libraries like System.Collections.Immutable that don't have separate contract assemblies vs libraries that do, whether the changes are ok to go to master or whether they need to go to future, etc. @weshaggard?

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
AArnott: True. I considered the alternative, but looking at the Sort(IComparer) method, it's non-trivial to replicate and then we have twice the maintenance burden. I ended with deciding that adding the API to enable customers is enabled by this smaller change, and we can decide to pay the cost of doubling the implementation if we later decide saving the allocation is worthwhile.

stephentoub:
@terrajobst, I believe the decision was that changes like this for immutable collections can go into master, correct?  And did we make a decision on this API specifically?

stephentoub:
@terrajobst, this is marked as "api-needs-work".  What is the work that's needed?

terrajobst:
LGTM

src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1+Builder.cs
```diff
@@ -636,6 +636,25 @@ public void Sort()
...
+            /// </param>
+            /// <exception cref="ArgumentNullException"><paramref name="comparison"/> is null.</exception>
+            [Pure]
+            public void Sort(Comparison<T> comparison)
+            {
+                Requires.NotNull(comparison, "comparison");
+
+                if (Count > 1)
+                {
+                    Array.Sort(_elements, comparison);
```
stephentoub: @AArnott, this is accessing both Count and _elements... do we need to do the whole ```var self = this;``` thing?

AArnott: No. That pattern is only necessary for the struct. This is the builder.

PR #3879, opened 2015-10-15
Make ExcludeFromCodeCoverage attribution conditional on DEBUG

stephentoub:
Some of our assemblies (e.g. System.Collections.Immutable) use ExcludeFromCodeCoverage to exclude from code coverage various members.  Since we only do code coverage runs against debug builds, there's no need to bloat such assemblies with ExcludeFromCodeCoverage on these members in release builds.

cc: @AArnott

### DIFF
--- a/src/Common/src/System/Diagnostics/CodeAnalysis/ExcludeFromCodeCoverageAttribute.cs
+++ b/src/Common/src/System/Diagnostics/CodeAnalysis/ExcludeFromCodeCoverageAttribute.cs
@@ -3,6 +3,7 @@
 
 namespace System.Diagnostics.CodeAnalysis
 {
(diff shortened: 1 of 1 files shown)


JonHanna:
Shiney.

AArnott:
LGTM, assuming this actually works. I haven't tried applying the `ConditionalAttribute` to another attribute.

stephentoub:
> assuming this actually works

It does.  I verified the behavior in builds locally, and it's documented:
"Applying ConditionalAttribute to an attribute indicates that the attribute should not be emitted to metadata unless the conditional compilation symbol is defined."
https://msdn.microsoft.com/en-us/library/system.diagnostics.conditionalattribute(v=vs.110).aspx

PR #3880, opened 2015-10-15
Do not throw from ImmutableArray<T>.RemoveRange when index==length

### LINKED ISSUE #2028: Inconsistent, unexpected bounds check in RemoveRange
* `ImmutableArray<T>.RemoveRange` requires 0<=index<size
* `ImmutableList<T>.RemoveRange` requires 0<=index<size *unless* size is 0, in which case it allows index=0

These bounds checks are not only inconsistent with each other, but they do not behave the same way as other range-based methods. In both of the cases above, the allowed range for index should be 0<=index<=size, regardless of the current size of the collection.

AArnott:
ImmutableArray<T>.RemoveRange(int, int) would throw when the first argument equals the length of the array, whereas ImmutableList<T>.RemoveRange(int, int) allows the first argument to equal the length of the list. This makes them consistent by slightly widening the allowed input to ImmutableArray.

ImmutableList is changed slightly to simplify the logic (but keep the meaning equivalent). The test is enhanced and a bug in the test fixed.

Fix #2028

### DIFF
--- a/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
+++ b/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray`1.cs
@@ -757,7 +757,7 @@ namespace System.Collections.Immutable
         public ImmutableArray<T> RemoveRange(int index, int length)
         {
             var self = this;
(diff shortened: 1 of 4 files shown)


stephentoub:
LGTM

PR #3885, opened 2015-10-15
Change how/when Console writes escape strings on Unix

stephentoub:
ForegroundColor, BackgroundColor, ResetColo, and CursorVisible all need to write out ANSI escape strings to perform their work, and we only want to do such writing if we expect the target will be able to correctly understand the inputs.

Previously, we were doing this by checking whether the file descriptor wrapped by the current output stream was a character device.  However, this behavior differs from the Windows implementation, which doesn't factor Console.SetOut into its decisions.  The heuristic is also susceptible to non-terminal character devices being the target of redirection, e.g. /dev/null.

This change updates the Unix implementation to be more consistent with Windows, simply checking whether stdout has been redirected rather than factoring in SetOut.  It also now uses isatty to determine whether to write out to the terminal.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Native/Interop.IsATty.cs
+++ b/src/Common/src/Interop/Unix/System.Native/Interop.IsATty.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @pallavit, @ellismg

stephentoub:
Thanks, Pallavi.

PR #3896, opened 2015-10-15
Add mvid and metadata token API to S.R.TypeExtensions

### LINKED ISSUE #2375: Add metadata tokens/module versions back to reflection libraries in CORE CLR.
Recent versions of the CORE clr have removed metadata from the reflection libraries. We are writing .NET diagnostics tools, and these APIs are important for supporting our diagnostics scenarios. We would like the APIs to be added back.

nguerrera:
*NOTE*: Marking as WIP because build and tests will fail without dotnet/coreclr#1781 and #3861

This adds ModuleExtensions and MemberInfoExtensions to the the System.Reflection.TypeExtensions library that provides extension methods to bridge reflection gaps. (It already has PropertyInfoExtensions, etc. etc.)

* Classic MemberInfo.MetadataToken is re-surfaced as `MemberInfo.TryGetMetadataToken(out int token)`. 
* Class Module.ModuleVersionId is re-surfaced as `Module.TryGetModuleVersionId(out Guid mvid)`

The TryGet pattern is used to draw attention to the fact these will not be supported everywhere (.NET Native will simply return false). 

**Update**: Changed to HasXxx and GetXxx from TryGet after API review feedback.

Other options such as nullables or using default(T) for failure were considered, but one thing that folks do with these is compare the return value from one instance to another which would give silent false positives. I think the scenarios for using these are limited enough for the more cumbersome signature.

Fix #2375 

cc @weshaggard @delmyers


src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs
```diff
@@ -279,6 +314,18 @@ public static MethodInfo GetBaseDefinition(MethodInfo method)
         }
     }
 
+    public static class ModuleExtensions
+    {
+        public static bool TryGetModuleVersionId(Module module, out Guid mvid)
```
weshaggard: Isn't this supposed to be an extension method? I don't see the this keyword.

src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs
```diff
@@ -270,6 +270,41 @@ public static MethodInfo GetRemoveMethod(EventInfo eventInfo, bool nonPublic)
...
+            int token;
+            try
+            {
+                token = member.MetadataToken;
+            }
+            catch (InvalidOperationException)
+            {
+                // Thrown for unbaked ref-emit members/types. 
+                // Surface as failed TryGet.
+                metadataToken = 0;
```
weshaggard: I guess it is these special cases that made you stick with the TryGet pattern instead of using Nullable<int>? I don't have a strong opinion either way, just curious.

src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs
nguerrera: Yes, it didn't affect testing because everything goes through the reference assembly. 

I actually left off the 'this' in implementation for things that have an exact instance counterpart in mscorlib (otherwise code looks like infinite recursion,  but that doesn't apply here, and in fact on desktop you reference the impl and need to see this new method as an extension.

Will fix.

src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs
nguerrera: It is more about the reason stated in the description. If (member1.GetMetadaToken() == member2.GetMetadataToken()) yields false positives when both fail (ditto for Mvid).

nguerrera:
Changed to account for API review feedback. @KrzysztofCwalina PTAL.

nguerrera:
(This will still fail CI as the CoreCLR package needs to be updated.)

KrzysztofCwalina:
looks good.

stephentoub:
@dotnet-bot test this please

nguerrera:
I've had to update the test-runtime to include Module.ModuleVersionId, which triggered #4040.

nguerrera:
@dotnet-bot test this please

nguerrera:
@dotnet-bot test this please

PR #3899, opened 2015-10-16
Make X500DistinguishedName.Decode be the same on Unix and Windows.

### LINKED ISSUE #1985: X509DistinguishedName::Format returns different answers on Windows and Unix
The most concrete example is that OID(2.5.4.8) displays in Windows as 'S', but on Unix (OpenSSL) as 'ST'.

This is related to https://www.github.com/dotnet/corefx/issues/1813, but a targeted fix for that issue would not fix this, and vice versa.

bartonjs:
The Windows and Unix implementations differed in a few cases:
* Key 2.5.4.8: Windows: "S", Unix: "ST"
* Key 1.2.840.113549.1.9.1: Windows "E", Unix "emailAddress"
* Value quoting: Windows wrapped the value in quotes for several special characters (such as comma).  Unix did not.

Now they're substantially closer to agreeing.  There's at least one known difference, and that is a single RDN with multiple components gets treated as separate RDNs on Unix, but is a special case on Windows.

Fixes #1985.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.Print.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.Print.cs
@@ -16,7 +16,10 @@ internal static partial class Interop
         [DllImport(Libraries.CryptoNative)]
         private static extern int Asn1StringPrintEx(SafeBioHandle bio, SafeAsn1StringHandle str, Asn1StringPrintFlags flags);
 
-        internal static unsafe string DerStringToManagedString(byte[] derString)
+        [DllImport(Libraries.CryptoNative)]
+        private static extern int Asn1StringPrintEx(SafeBioHandle bio, SafeSharedAsn1StringHandle str, Asn1StringPrintFlags flags);
+
+        internal static string DerStringToManagedString(byte[] derString)
         {
             SafeAsn1StringHandle asn1String = DecodeAsn1TypeBytes(derString, derString.Length, AnyTextStringType);
 
@@ -25,16 +28,36 @@ internal static partial class Interop
                 return null;
             }
 
+            using (asn1String)
+            {
+                return Asn1StringToManagedString(
+                    asn1String,
+                    (bio, str, flags) => Asn1StringPrintEx(bio, str, flags));
+            }
+        }
+
+        internal static string Asn1StringToManagedString(SafeSharedAsn1StringHandle asn1String)
+        {
+            CheckValidOpenSslHandle(asn1String);
+
+            return Asn1StringToManagedString(
+                asn1String,
+                (bio, str, flags) => Asn1StringPrintEx(bio, str, flags));
+        }
+
+        private static string Asn1StringToManagedString<THandle>(
+            THandle asn1String,
+            Func<SafeBioHandle, THandle, Asn1StringPrintFlags, int> asn1StringPrintEx)
+        {
             byte[] utf8Bytes;
 
-            using (asn1String)
             using (SafeBioHandle bio = CreateMemoryBio())
             {
-                int len = Asn1StringPrintEx(bio, asn1String, Asn1StringPrintFlags.ASN1_STRFLGS_UTF8_CONVERT);
+                int len = asn1StringPrintEx(bio, asn1String, Asn1StringPrintFlags.ASN1_STRFLGS_UTF8_CONVERT);
 
                 if (len < 0)
                 {
-                    throw Crypto.CreateOpenSslCryptographicException();
+                    throw CreateOpenSslCryptographicException();
                 }
 
                 int bioSize = GetMemoryBioSize(bio);
@@ -44,7 +67,7 @@ internal static partial class Interop
 
                 if (read < 0)
                 {
-                    throw Crypto.CreateOpenSslCryptographicException();
+                    throw CreateOpenSslCryptographicException();
                 }
             }
 
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
@@ -61,6 +61,22 @@ internal static partial class Interop
         [DllImport(Libraries.CryptoNative)]
         internal static extern void Asn1StringFree(IntPtr o);
 
+        internal static string GetOidValue(SafeSharedAsn1ObjectHandle asn1Object)
+        {
+            Debug.Assert(asn1Object != null);
+
+            bool added = false;
+            asn1Object.DangerousAddRef(ref added);
+            try
+            {
+                return GetOidValue(asn1Object.DangerousGetHandle());
+            }
+            finally
+            {
+                asn1Object.DangerousRelease();
+            }
+        }
+
         internal static string GetOidValue(IntPtr asn1ObjectPtr)
         {
             // OBJ_obj2txt returns the number of bytes that should have been in the answer, but it does not accept
@@ -104,6 +120,17 @@ internal static partial class Interop
             }
 
             return buf.ToString();
+        }
+    }
+}
+
+namespace Microsoft.Win32.SafeHandles
+{
+    internal class SafeSharedAsn1ObjectHandle : SafeInteriorHandle
+    {
+        private SafeSharedAsn1ObjectHandle() :
+            base(IntPtr.Zero, ownsHandle: true)
+        {
         }
     }
 }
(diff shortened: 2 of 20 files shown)


src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -61,6 +61,22 @@ internal static partial class Crypto
...
 
+        internal static string GetOidValue(SafeSharedAsn1ObjectHandle asn1Object)
+        {
+            Debug.Assert(asn1Object != null);
+
+            bool added = false;
+            asn1Object.DangerousAddRef(ref added);
+            try
+            {
+                return GetOidValue(asn1Object.DangerousGetHandle());
```
stephentoub: Can you remind me why the other GetOidValue doesn't just take a SafeHandle and pass it to a P/Invoke declared to take a SafeHandle rather than an IntPtr?

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
```diff
@@ -61,6 +61,22 @@ internal static partial class Crypto
         [DllImport(Libraries.CryptoNative)]
         internal static extern void Asn1StringFree(IntPtr o);
 
+        internal static string GetOidValue(SafeSharedAsn1ObjectHandle asn1Object)
+        {
+            Debug.Assert(asn1Object != null);
+
+            bool added = false;
+            asn1Object.DangerousAddRef(ref added);
```
stephentoub: Technically the pattern is to do this inside the try block and then have the finally conditionally do the release based on the value of added.  However, without thread aborts to worry about, this isn't necessary.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
```diff
@@ -0,0 +1,192 @@
...
+    internal static class X500NameEncoder
+    {
+        private static readonly char[] s_quoteNeedingChars =
+       {
+            ',',
+            '+',
+            '=',
+            '\"',
+            '\n',
+            // \r is NOT in this list.
```
stephentoub: Is it worth clarifying why?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
```diff
@@ -0,0 +1,192 @@
...
+        internal static string X500DistinguishedNameDecode(
+            byte[] encodedDistinguishedName,
+            bool printOid,
+            X500DistinguishedNameFlags flags)
+        {
+            bool reverse = flags.HasFlag(X500DistinguishedNameFlags.Reversed);
+            bool quoteIfNeeded = !flags.HasFlag(X500DistinguishedNameFlags.DoNotUseQuotes);
+            string dnSeparator;
+
+            if (flags.HasFlag(X500DistinguishedNameFlags.UseSemicolons))
```
stephentoub: Last I checked, HasFlag was unfortunately fairly expensive.  It'd be better to do the check yourself with something like:
```C#
if ((flags & X500DistinguishedNameFlags.UseSemicolons) == X500DistinguishedNameFlags.UseSemicolons)
```

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            using (SafeX509NameHandle x509Name = Interop.Crypto.DecodeX509Name(encodedDistinguishedName, encodedDistinguishedName.Length))
+            {
+                if (x509Name.IsInvalid)
+                {
+                    return "";
+                }
+
+                // GitHub.com's SSL cert has a "pretty long" subject (partially due to the unknown OIDs):
```
stephentoub: Maybe clarify earlier on that this is an example being used for length approximation.  Wasn't clear to me why we were talking about GitHub's cert until the very end of the comment.

stephentoub:
LGTM

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
```diff
@@ -29,40 +27,7 @@ public AsymmetricAlgorithm DecodePublicKey(Oid oid, byte[] encodedKeyValue, byte
 
         public string X500DistinguishedNameDecode(byte[] encodedDistinguishedName, X500DistinguishedNameFlags flags)
         {
-            OpenSslX09NameFormatFlags nativeFlags = ConvertFormatFlags(flags);
```
eerhardt: Is ConvertFormatFlags still needed after this change?

eerhardt: And OpenSslX09NameFormatFlags in general?

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.cs
bartonjs: Just that it was written before I'd started partying with SafeInteriorHandles.  I want to clean up the rest of the usages, but didn't want to conflict with Eric's changes for shimming; so kept this as a style adapter.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
bartonjs: > Is it worth clarifying why?

Just 'cuz it isn't in Windows; but I can add that to the comment.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509Encoder.cs
bartonjs: No, I thought I removed them; but maybe I forgot to stage it. :(

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
bartonjs: Updated the comment to give a reasonable lead-in.  It's practically a short story at this point :)

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/X500NameEncoder.cs
bartonjs: Just made them all be the inline and-and-test.

bartonjs:
@dotnet-bot test this please (System.Net.Sockets.Tests reported failure with no output?)

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.Print.cs
```diff
@@ -25,16 +28,36 @@ internal static unsafe string DerStringToManagedString(byte[] derString)
                 return null;
             }
 
+            using (asn1String)
+            {
+                return Asn1StringToManagedString(
+                    asn1String,
+                    (bio, str, flags) => Asn1StringPrintEx(bio, str, flags));
```
eerhardt: I think you can just make this:

```C#
return Asn1StringToManagedString(  
  asn1String,  
  Asn1StringPrintEx);  
```
And it will save on the creation of a new anonymous inline method.

Same comment for the method below.

bartonjs: Yep, but apparently there's a language requirement that the bare `Asn1StringPrintEx` create a new delegate object each call, and the lambda is allowed to be cached.   So using the lambda on a static method is "better".

eerhardt: So have I been doing it "worse"?

https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/CollectionBackedStoreProvider.cs#L125

bartonjs: Yep, according to the discussion at https://github.com/bartonjs/corefx/commit/f5452f9adfc0fea0f0262075f32e901f56db5771#commitcomment-12070315.

PR #3901, opened 2015-10-16
Correct ArgumentOutOfRangeException on out of range block expression index

### LINKED ISSUE #3881: Block.Expressions sometimes throws incorrect exception on out-of-range
Given:

    BlockExpression block = Expression.Block(Expression.Empty(), Expression.Constant(3));

One would expect `block.Expressions[-1]`, `block.Expressions[2]`, etc. to throw `ArgumentOutOfRangeException`. Instead they throw `InvalidOperationException`.

This depends on the factory used, so some cases will throw the correct exception, which is perhaps worse as this is then inconsistent.

JonHanna:
Fixes #3881

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
@@ -197,7 +197,7 @@ namespace System.Linq.Expressions
             {
                 case 0: return ReturnObject<Expression>(_arg0);
                 case 1: return _arg1;
-                default: throw new InvalidOperationException();
+                default: throw Error.ArgumentOutOfRange("index");
             }
         }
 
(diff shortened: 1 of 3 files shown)


stephentoub:
@dotnet-bot test this please (FindPackagesById errors)

JonHanna:

JonHanna:
Test Innerloop Windows_NT Release Build and Test

stephentoub:
The mechanics of this change LGTM.

But, @VSadov, is this something we want to change, i.e. do we want to change the exception type used here?

JonHanna:
Aside from `ArgumentOutOfRangeException` seeming the natural choice in terms of how it is normally used, it currently is the exception thrown here in some cases, depending on which factory was used to produce the `BlockExpression`. I think that inconsistency is worse than whether the type thrown is considered incorrect.

VSadov:
I think dependencies on actual exception type in a case of an error is uncommon so this is unlikely to be breaking. There are some advantages in being consistent though.
LGTM

PR #3900, opened 2015-10-16
Add block expression tests.

JonHanna:
Includes removal of dead code in BlockExpression.cs; `GetVariable(int)` is
never hit.

PR #3903, opened 2015-10-16
[WIP] Return same instance for Block.Update() with same children.

### LINKED ISSUE #3883: Block.Update returns different expression for identical children in different collection.
According to the documentation:

> If all of the children are the same, it will return this expression.

However, only the enumerables themselves are compared, so a different enumeration of the same children will incorrectly return a new expression.

JonHanna:
Fixes #3883

Currently including commit of #3900 in this. On that being merged, will update to include only second commit, or as suitable based on decision on that PR.

### DIFF
--- a/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
+++ b/src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
@@ -114,11 +114,6 @@ namespace System.Linq.Expressions
             throw ContractUtils.Unreachable;
         }
 
-        internal virtual ParameterExpression GetVariable(int index)
-        {
-            throw ContractUtils.Unreachable;
-        }
-
         internal virtual int VariableCount
         {
             get
@@ -418,11 +413,6 @@ namespace System.Linq.Expressions
             {
                 return _variables.Count;
             }
-        }
-
-        internal override ParameterExpression GetVariable(int index)
-        {
-            return _variables[index];
         }
 
         internal override ReadOnlyCollection<ParameterExpression> GetOrMakeVariables()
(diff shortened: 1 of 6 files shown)


stephentoub:
@dotnet-bot test this please (FindPackagesById errors)

src/System.Linq.Expressions/src/System/Linq/Expressions/BlockExpression.cs
```diff
@@ -88,9 +88,37 @@ public override Type Type
...
+                IList<ParameterExpression> varList = variables as IList<ParameterExpression> ?? variables.ToReadOnly();
+                variablesMatch = Variables.Count == varList.Count && Variables.SequenceEqual(varList);
+            }
+
+            if(variablesMatch)
+            {
+                bool expressionsMatch;
+                if (expressions == Expressions)
+                    expressionsMatch = true;
+                else if (expressions == null)
```
Clockwork-Muse: You don't have the equivalent of 
```
else if (Variables == null || Variables.Count == 0)  
    variablesMatch = variables == null || !variables.Any(); 
```
for block expressions.  Is how the classes are used making that check unnecessary (always one block)?

I find myself twitching, wanting to consolidate the mostly identical checks of variables and expressions here (mostly to remove the nesting, because what I'm working though currently has too much of it).

JonHanna: Even worse, I jump through all these hoops to short circuit with as little allocation as possible, and then the final call to Expression.Block is incorrect.

JonHanna:
Closing, pending more input on #3958, perhaps to return with something covering all affected expression types.

PR #3911, opened 2015-10-16
Finish most of the sockets shims.

pgavlin:
- Shim SocketFlags
- Shim select()
- Fix some issues with error propagation in socket creation and accepted
  sockets

The only functions that have not been shimmed are those relied upon for event notification (epoll/kqueue).


pgavlin:
@stephentoub @sokket @nguerrera PTAL

sokket:
Looks like Interop.optlevel.cs and Interop.optname.cs don't have any changes but are part of the PR

src/Native/System.Native/pal_networking.cpp
```diff
@@ -6,15 +6,16 @@
...
 #include <netinet/in.h>
 #include <netinet/tcp.h>
 #include <sys/socket.h>
+#include <sys/ioctl.h>
 #include <unistd.h>
-#include <vector>
+#include <alloca.h>
 #include <errno.h>
+#include <assert.h>
+#include <vector>
```
sokket: I'd prefer to try and keep the included alphabetical, unless there are dependencies between headers

sokket:
Besides the minor NIT, and run clang-format on the native files that you changed, LGTM

src/Native/System.Native/pal_networking.cpp
pgavlin: Ah, so that's how they were organized before. Noted.

pgavlin:
Interop.opt{level,name}.cs have been deleted--since they were empty files, the diff looks a little strange.

PR #3912, opened 2015-10-16
Add a padding byte for "negative" cert serial numbers on Unix

### LINKED ISSUE #3893: Unix X509Certificate2.SerialNumber doesn't include leading zero-bytes
When the serial number's most significant byte has the most significant bit set, Windows includes a leading `0x00` to prevent it from being interpreted as a negative value.

Unix does not.

bartonjs:
Ensured that there's an explicit test for when this is required, and when it isn't.

Fixes #3893.

cc @stephentoub @AtsushiKan

### DIFF
--- a/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.GetIntegerBytes.cs
+++ b/src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.GetIntegerBytes.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text;
+
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
+{
+    internal static partial class Crypto
+    {
+        [DllImport(Libraries.CryptoNative)]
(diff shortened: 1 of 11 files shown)


bartonjs:
I think there might be more to this.  A different cert isn't reporting the leading "00" from Windows; it may matter how it was encoded in the body.

stephentoub:
> I think there might be more to this

If you determine that the core premise is correct, the change LGTM.

bartonjs:
Okay, took a completely different approach.  It turns out that not only were "maybe negative" being done wrong, but "actually negative" was done wrong.

There was a lot more wrong here than met the eye, luckily I ran into another wrong while testing the fix to a different issue.

@stephentoub you may want to take another look :)

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.ASN1.GetIntegerBytes.cs
```diff
@@ -0,0 +1,44 @@
...
+            // OpenSSL stores negative numbers in their two's complement (positive) form, but
+            // sets an internal negative bit.
+            //
+            // If the number was positive, but could sign-test as negative, DER puts in a leading
+            // 0x00 byte, which reading OpenSSL's data directly won't have.
+            //
+            // So to ensure we're getting a set of bytes compatible with BigInteger (though with the
+            // wrong endianness here), DER encode it, then use the DER reader to skip past the tag
+            // and length.
+            byte[] derEncoded = OpenSslEnocde(
```
stephentoub: Same comment as in the other PR about "OpenSslEnocde" typo :)

stephentoub:
LGTM

PR #3914, opened 2015-10-16
Add shims for EcKey, EvpPkey, and EcDsa files

eerhardt:
Add shims for all the remaining libcrypto calls, except Interop.Rsa.cs.


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
```diff
@@ -245,10 +245,7 @@ public RSA GetRSAPrivateKey()
                 return null;
             }
 
-            using (SafeRsaHandle rsaHandle = Interop.libcrypto.EVP_PKEY_get1_RSA(_privateKey))
```
eerhardt: @bartonjs - are the two changes to this file OK? Doing it this way allowed me to not have to include a bunch of interop files in the X509Certificates.csproj. My naïve thinking is that this approach is a little different than existing code, but should be fine.  The current code calls EVP_PKEY_get1_* here and then passes in the IntPtr into the constructor, which then just duplicates the handle and then this code disposes of the first handle.  The new way is to pass in the privateKey into the constructor, which will call EVP_PKEY_get1_* and just hold on to that handle.  So this way seems simpler/more performant to me.

Thoughts?

stephentoub:
LGTM

sokket:
LGTM 2

nguerrera:
LGTM

src/Native/System.Security.Cryptography.Native/pal_evp_pkey.h
```diff
@@ -0,0 +1,37 @@
...
+The given EVP_PKEY pointer is invalid after this call.
+Always succeeds.
+*/
+extern "C" void EvpPkeyDestroy(EVP_PKEY* pkey);
+
+/*
+Function:
+UpRefEvpPkey
+
+Used by System.Security.Cryptography.X509Certificates' OpenSslX509CertificateReader when
```
bartonjs: I'd alter these to match the style of the rest of the shims; and get rid of the CoreClrPal notion of who cares about it.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslX509CertificateReader.cs
bartonjs: Seems reasonable.  When this code was written there wasn't an SafeEvpPKeyHandle constructor :)

bartonjs:
One comment about a comment, otherwise LGTM.

src/Native/System.Security.Cryptography.Native/pal_ecdsa.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    if (!siglen)
+    {
+        return 0;
+    }
+
+    unsigned int unsignedSigLength = UnsignedCast(*siglen);
+    int ret = ECDSA_sign(0, dgst, dgstlen, sig, &unsignedSigLength, key);
+    assert(unsignedSigLength <= INT32_MAX);
+    *siglen = static_cast<int32_t>(unsignedSigLength);
```
nguerrera: We should add SignedCast for this.

src/Native/System.Security.Cryptography.Native/pal_evp_pkey.h
eerhardt: Cleaned up.

src/Native/System.Security.Cryptography.Native/pal_ecdsa.cpp
eerhardt: Done.

PR #3916, opened 2015-10-16
Populate DataHash key size in SslConnectionInfo

shrutigarg:
Populating DataHash key size in SslConnectionInfo Structure.
Found out that opnssl doesn't provide a way to return the keyExchange size .. Internally it does some calculations for key size depending upon which keyEx Algo is used.
So that thing is still left .. For now would be tracked under same issue.

### DIFF
--- a/src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
+++ b/src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
@@ -31,6 +31,13 @@ namespace System.Net
         internal const int SSL_GOST89 = 229411;
         internal const int SSL_AEAD = 229412;
 
+        //Data Hash Key Sizes
+        internal const int MD5_HashKeySize = 128;
+        internal const int SHA1_HashKeySize = 160;
+        internal const int SHA256_HashKeySize = 256;
+        internal const int SHA384_HashKeySize = 384;
+        internal const int GOST_HashKeySize = 256;
+
         public readonly int Protocol;
         public readonly int DataCipherAlg;
         public readonly int DataKeySize = 0;
@@ -46,7 +53,12 @@ namespace System.Net
             KeyExchangeAlg = (int) MapExchangeAlgorithmType((Interop.libssl.KeyExchangeAlgorithm) cipher.algorithm_mkey);
             DataHashAlg = (int) MapHashAlgorithmType((Interop.libssl.DataHashAlgorithm) cipher.algorithm_mac);
             DataKeySize = cipher.alg_bits;
-            // TODO (Issue #3362) map key sizes
+            DataHashKeySize = GetHashAlgorithmBitSize((Interop.libssl.DataHashAlgorithm) cipher.algorithm_mac);
+
+            // TODO (Issue #3362) Map keyExchKeySize.
+            //There is no way we can get the key exchange size from the Openssl.
+            //It internally generates the key size in *_send_client_key_exchange method on the basis of the algorithm used.
+
         }
 
         private SslProtocols MapProtocolVersion(string protocolVersion)
@@ -188,6 +200,33 @@ namespace System.Net
 
                 default:
                     return HashAlgorithmType.None;
+            }
+        }
+        
+        private static int GetHashAlgorithmBitSize(Interop.libssl.DataHashAlgorithm mac)
+        {
+            switch (mac)
+            {
+                case Interop.libssl.DataHashAlgorithm.SSL_MD5:
+                    return MD5_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_SHA1:
+                    return SHA1_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_GOST94:
+                    return GOST_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_GOST89MAC:
+                    return GOST_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_SHA256:
+                    return SHA256_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_SHA384:
+                    return SHA384_HashKeySize;
+
+                default:
+                    return 0;
             }
         }
     }


shrutigarg:
@stephentoub, @CIPop, @SidharthNabar @bartonjs @vijaykota

src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
```diff
@@ -46,7 +53,12 @@ internal SslConnectionInfo(Interop.libssl.SSL_CIPHER cipher, string protocol)
             KeyExchangeAlg = (int) MapExchangeAlgorithmType((Interop.libssl.KeyExchangeAlgorithm) cipher.algorithm_mkey);
             DataHashAlg = (int) MapHashAlgorithmType((Interop.libssl.DataHashAlgorithm) cipher.algorithm_mac);
             DataKeySize = cipher.alg_bits;
-            // TODO (Issue #3362) map key sizes
+            DataHashKeySize = GetHashAlgorithmBitSize((Interop.libssl.DataHashAlgorithm) cipher.algorithm_mac);
```
stephentoub: We can remove the ```= 0``` from the declaration of ```DataHashKeySize``` earlier.

src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
```diff
@@ -190,5 +202,32 @@ private static HashAlgorithmType MapHashAlgorithmType(Interop.libssl.DataHashAlg
...
+                    return GOST_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_SHA256:
+                    return SHA256_HashKeySize;
+
+                case Interop.libssl.DataHashAlgorithm.SSL_SHA384:
+                    return SHA384_HashKeySize;
+
+                default:
+                    return 0;
```
stephentoub: What happens if another algorithm ends up being used and we return 0 here?

shrutigarg: apart from the macs mentioned there is one more mac which is AEAD ..  which is as per the explanation in Openssl is not is real mac .
AEAD is only used with *GCM Cipher Algorithm.. if *GCM encryption algorthim (cipher algo) is used, it typically does both the work encryption and hashing.

 "TLSv1.2 adds a new class of ciphers using AEAD modes such as GCM, which is a single crypto operation that does both encryption and integrity protection, instead of having separate encryption and HMAC e.g. AES128-CBC plus HMAC-SHA1. Mac=AEAD means really no HMAC is needed because of AEAD."

So I think returning some default thing would be the better idea.

src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
shrutigarg: yep , will fix

src/Common/src/Interop/Unix/libssl/SslConnectionInfo.cs
stephentoub: @bartonjs, opinion?

bartonjs: @CIPop wrote a little tool for printing the details of an SSL connection.

On the server side I ran `openssl s_server -cert eccombined.pem -cipher ECDHE-ECDSA-AES128-GCM-SHA256 -accept 8000`, forcing it into a GCM connection.  The Windows connection information says
```
        PROTOCOL=Tls12 , KEYEXCHG=( 256bit      44550), CYPHER=( 128bit     Aes128), HASH=(   0bit      32780)
```

I boosted it a bit:
```
 KeyEx: Class=5 Type=7 Sid=6
Cipher: Class=3 Type=3 Sid=14
  Hash: Class=4 Type=0 Sid=12
```

The hash appears to be reported as a 0-bit strong hash (class=4) any (type=0) SHA-256 (sid=12).   The relevant part for this function is that it agrees that the hash size should be 0 for AEAD.

What's probably better is to report 0 for AEAD (`# define SSL_AEAD 0x00000040L`) and -1 or throw, or something on otherwise unknown inputs.

bartonjs: We should be sure this is doing something sensible on OSX.

[Our struct](https://github.com/dotnet/corefx/blob/2c6a6dc47ce711ced3705d09641a816d4903baa5/src/Common/src/Interop/Unix/libssl/Interop.libssl.cs) looks like [OpenSSL Master](https://github.com/openssl/openssl/blob/master/ssl/ssl_locl.h#L538-L555), but might be different than [OpenSSL 0.9.8zg](https://github.com/openssl/openssl/blob/OpenSSL_0_9_8zg/ssl/ssl.h#L354-L365) (which is what OSX has).  I'm not sure that we're doing anything sensible on OSX.

bartonjs: And, after more digging, I'm pretty confident that we're reading something other than what we expect on OSX.

In the 0.9.8 days, the MAC was a few bits of the `SSL_CIPHER.algorithms` field `0x00c00000L`.  So... yeah.  This is bad.

bartonjs: @shrutigarg Given that this has a not-understood behavior on OSX due to the struct being copied into C# (vs shimmed), I think it's best if you shim this up.

Modern libssl exposes what you want in  `int SSL_CIPHER_get_digest_nid(const SSL_CIPHER *c);` (didn't exist in 0.9.8).

You probably want something like
```C++
extern "C" int32_t GetSslCipherMacInfo(SSL* ssl, int32_t** macNid, int32_t** macSizeBytes)
{
    // Test ssl, algNid, algSizeBytes for validity.  This is just a sample.

    SSL_CIPHER* cipher = SSL_get_current_cipher(ssl);

    if (!cipher)
    {
        // do something
    }

    int cipherMac;
    
#if IMNOTSUREWHAT // OpenSSL 0.9.8
    switch (cipher->algorithms & SSL_MAC_MASK)
    {
        case SSL_MD5:
            cipherMac = NID_md5;
            break;
        case SSL_SHA1:
            cipherMac = NID_sha1;
            break;
        default:
            // Handle this unknown case
            break;
    }
#else
    cipherMac = SSL_CIPHER_get_digest_nid(cipher);
#endif

    if (cipherMac == NID_unknown)
    {
        assert(cipher->algorithm_mac == SSL_AEAD);
        *macSizeBytes = 0;
    }
    else
    {
        *macSizeBytes = EVP_MD_size(EVP_get_digestbynid(cipherMac));
    }

    *macNId = cipherMac;
    return 1;
}
```

eerhardt: How about keep it real simple and `#if IMNOTSUREWHAT` becomes `#if OPENSSL_0_9_8` ?

vijaykota: @eerhardt, is there a similar macro to indicate "modern libssl" so that #if can be inverted?

shrutigarg: @bartonjs : yes, to make it work in veriosn0.9.8 we would have to change the logic for other algorithms too . keyEx and DataCipher.. ..
Btw I don't find  "SSL_CIPHER_get_digest_nid " in 1.0.* existing versions. instead we can use 

#if IMNOTSUREWHAT // OpenSSL 0.9.8
    switch (cipher->algorithms & SSL_MAC_MASK)
    {
        case SSL_MD5:
            cipherMac = NID_md5;
            break;
        case SSL_SHA1:
            cipherMac = NID_sha1;
            break;
        default:
            // Handle this unknown case
            break;
    }
#else
      switch (cipher->algorithms_mac)
    {
      //do mapping the way it is done currently.
    }
#endif
..

eerhardt: > is there a similar macro to indicate "modern libssl" so that #if can be inverted?

There aren't any macros/defines at all yet that I'm aware of, I was suggesting we create one named OPENSSL_0_9_8 to use on OSX.
If you wanted it inverted, you could use `#ifndef OPENSSL_0_9_8`.  I'm not sure how many versions of OpenSsl we are planning on supporting at one time. I would assume we are supporting 0.9.8 for use on OSX, and the last known good version for use on Linux. On my Ubuntu machine, it looks like I have 1.0.1.  @bartonjs - can you confirm?

> Btw I don't find "SSL_CIPHER_get_digest_nid " in 1.0.* existing versions. Which version is this present?

According to https://www.openssl.org/docs/manmaster/ssl/SSL_CIPHER_get_name.html, it exists in the latest, i.e. "master", version but not in 1.0.2 or 1.0.1.

eerhardt: @vijaykota and @shrutigarg - Please see the changes I am proposing at https://github.com/eerhardt/corefx/commit/4f48f85b582df14d0209dd6226f8fe03a5ac0b97 to support differing APIs in OpenSSL.

eerhardt: FYI - see https://github.com/dotnet/corefx/pull/4026 for how I shimmed the SSL_CIPHER struct.

shrutigarg: closing this and sending a new PR for this after shim update

Issue #674, opened 2015-02-06
FileVersionInfo tests' native assemblies and exes do not have sources


nslottow:
We should at least provide the sources for these native binaries so that we can tell what's going on if we hit issues.

The next time we need to make a change to these tests, it would also be worth getting the native binaries to build as vcxprojs, relying on the VS2013 prerequisite for now (see [System.ServiceProcess.ServiceController.Tests](https://github.com/dotnet/corefx/tree/v1.0/src/System.ServiceProcess.ServiceController/tests/System.ServiceProcess.ServiceController.Tests) for an example).  When we start to light up tests cross-platform, we will be forced to clean this up and use CMake or some other solution for native binaries.

cc @weshaggard

Alexx999:
I believe that sources are not *really* needed here since that binaries are never loaded and/or run.
They only serve purpose of "test data".
I even suppose that it's not that great to rebuild them since older legacy binaries built with previous versions of compilers should work fine too - that's the point of that test. It should test PE format in general.
And this will actually break cross-platform compatibility of this test since native PE executable can't be built on Linux (and we still should be testing PE on Linux since managed assemblies are PE executables even there)

nslottow:
All good points.  I should've examined these tests more closely before filing this.  I was coming from looking at the System.ServiceProcess.ServiceController tests which got initially committed with a binary that turns out to be unnecessary.

I'll leave it up to @mellinoe to decide whether to provide source.

weshaggard:
I agree with @alexx999 that in general we should just make the source available in case we need to examine or rebuild but it shouldn't be part of the build for these tests.

ianhays:
In solving #3919 I was unable to find the source for these assemblies and so fixed it by manually editing the dlls. I agree with @Alexx999 that in this particular case, the source for the binaries isn't really what we care about and I would say that including it may potentially be confusing because it may imply that that source should be built.

If we can find it then it may be worth including just as an effort towards transparency in "open sourcing all the things" but I don't think it will provide much value.

mellinoe:
TBH, I don't think this is actually important. If we need to change some of the metadata in the test files, we could just use a binary editor, or just recreate the binaries easily.

PR #3919, opened 2015-10-16
Fixed FileVersionInfo typo

### LINKED ISSUE #636: Possible typo in FileVersionInfoTest
This is low priority, but logging here so it doesn't get lost.

In [FileVersionInfoTest.cs line 239](https://github.com/dotnet/corefx/blob/0252409de5d77d4a104903a68d9ee34d54992e54/src/System.Diagnostics.FileVersionInfo/tests/FileVersionInfoTest.cs#L239):

    s_fviSecondNativeLibrary.ProductName = "Unkown_Product_Name";

"Unkown" should be "Unknown", but correcting it fails the test.  It should probably be first corrected wherever it originates from.

Thanks.

ianhays:
The test dll for FileVersionInfo SecondNativeLibrary.dll contained a typo for the word "unknown". I was unable to find the source for this dll anywhere in TFS/ToF, so opted instead to modify the binary info directly and merge that as a fix.

resolves #636

@FiveTimesTheFun @stephentoub


FiveTimesTheFun:
Thanks Ian, lgtm.  I recommend making the same change in ToF for consistency.

akoeplinger:
Should https://github.com/dotnet/corefx/issues/674 be closed then since the source can't be found? Or should we eventually replace SecondNativeLibrary.dll with a new file that has sources?

ianhays:
I think this is one of the rare cases where the source is pretty much irrelevant - so much so that <i>we can't even find it</i> and it doesn't much affect the tests. I think it may be more confusing than helpful to add source for the tests because of the points mentioned in #674, but I can also see the transparency benefit of open sourcing all the things.

I'll leave that decision up to @mellinoe and merge this PR as is for now. Thanks for bringing that issue to my attention @akoeplinger :)

PR #3923, opened 2015-10-16
Modified Collections perf tests to use ints

ianhays:
The perf results for the Collections tests were being affected by the performance of the String Equals/GetHashCode methods. I've modified the tests to use ints so that the overhead of string comparison is taken out of the measurement.

@stephentoub

### DIFF
--- a/src/System.Collections/tests/Performance/Perf.Dictionary.cs
+++ b/src/System.Collections/tests/Performance/Perf.Dictionary.cs
@@ -10,26 +10,9 @@ namespace System.Collections.Tests
     public class Perf_Dictionary
     {
         /// <summary>
-        /// Creates a Dictionary of string-string with the specified number of pairs
-        /// </summary>
-        public static Dictionary<string, string> CreateDictionary(int size)
-        {
-            PerfUtils utils = new PerfUtils(837322);
-            Dictionary<string, string> dict = new Dictionary<string, string>();
-            while (dict.Count < size)
-            {
-                string key = utils.CreateString(50);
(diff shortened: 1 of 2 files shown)


src/System.Collections/tests/Performance/Perf.Dictionary.cs
```diff
@@ -111,9 +94,9 @@ public void GetItem(int size)
                 using (iteration.StartMeasurement())
                     for (int i = 0; i <= 10000; i++)
                     {
-                        retrieved = dict["key1"]; retrieved = dict["key2"]; retrieved = dict["key3"];
-                        retrieved = dict["key4"]; retrieved = dict["key5"]; retrieved = dict["key6"];
-                        retrieved = dict["key7"]; retrieved = dict["key8"]; retrieved = dict["key9"];
+                        retrieved = dict[1]; retrieved = dict[2]; retrieved = dict[3];
```
ianhays: I wonder if it might be worthwhile to make these not be sequential i.e. make them occur at different places in the dictionary. The usefulness of that would depend on GetHashCode for ints, I think.

ellismg: It is probably a good idea.  When there are collisions insertion and access is slower as we have to walk the buckets array.

stephentoub: Seems like it would depend on what we're trying to test... we might want both a test when there aren't many collisions and a test for when everything hashes to the same bucket.

stephentoub:
LGTM

PR #3925, opened 2015-10-16
Enable System.Console.ReadKey

pallavit:
This change tries to enable #1153.

On Windows the implementation is pretty straight-forward. We have APIs to capture the standard input and APIs that tell us the kind of key and modifier is pressed. 

On Unix, there is no concept of virtual key codes and so, the implementation is mostly best case effort. 
The current implementation works as follows.
1. Capture stdin in the raw mode and reads as much of buffer as possible (max = 255 bytes).
2. Try to match the buffer with any of the special keys which it maintains in the db from the static string capabilities from TermInfo.
3. If not, it simply tries to use the ASCII key chart and map it to an equivalent ConsoleKey.

It also tries to second guess the modifier keys as follows
1. Alt - This is captured as esc key, so, we try to check if at the time of processing we can a combination of ESC + known key and if so, we assume alt was pressed.
2. Ctrl - Other than the known keys like \b \t and \n, the rest of the ASCII key charts from 1 to 26 are considered typed with Ctrl + the alphabet.
3. Shift - Any capital A to Z are assumed to be pressed with Shift key.

Not all keys are captured by this logic. For example, on my keyboard, NumLock + No. key results in a byte sequence, that can't be mapped to the no. itself. So, Unix implementation can never return ConsoleKey.NumPadX, Similarly combinations of Alt+FunctionKeys can't be interpreted correctly and are interpreted as Esc + the individual keys. Since not all key presses can be mapped to the correct ConsoleKey, the implementation tries to ignore 1 byte in case it can't be mapped to anything and continue to interpret the next byte. This is done because (ConsoleKey)0 is invalid. So, even if we can get the int representation of the byte being processed, we can't really create a valid ConsoleKeyInfo from it.

### DIFF
--- a/src/Common/src/Interop/Unix/System.Native/Interop.ReadStdinUnbuffered.cs
+++ b/src/Common/src/Interop/Unix/System.Native/Interop.ReadStdinUnbuffered.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Sys
+    {
+        [DllImport(Libraries.SystemNative, SetLastError = true)]
+        internal unsafe static extern int ReadStdinUnbuffered(byte* buffer, int bufferSize);
+    }
+}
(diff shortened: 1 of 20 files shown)


pallavit:
@weshaggard , @stephentoub Could you please take a look.

Issue #2523, opened 2015-07-25
RSAOpenSsl should support OAEP padding using a SHA-2 algorithm


bartonjs:
Currently it only supports Pkcs1 and OaepSha1.

Other hashing algorithms would work, but it requires doing the OAEP manually, since OpenSSL doesn't expose a convenient wrapper function for this.

PR #3928, opened 2015-10-17
Add shims for Interop.Rsa methods.

eerhardt:
Finishing up the last of the libcrypto native shims. There are a handful of cleanup tasks that need to be done once all the changes are merged, (ex. remove Libraries.LibCrypto from managed code). I'll send that out early next week.


src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
```diff
@@ -12,3 +31,118 @@ extern "C" RSA* DecodeRsaPublicKey(const unsigned char* buf, int32_t len)
...
+                                 uint8_t* q,
+                                 int32_t qLength,
+                                 uint8_t* dmq1,
+                                 int32_t dmq1Length,
+                                 uint8_t* iqmp,
+                                 int32_t iqmpLength)
+{
+    if (!rsa)
+    {
+        return;
```
nguerrera: Should this assert? There's no error code. Do we ever expect to let this no-op?

nguerrera: Same comment applies across the board.

src/Native/System.Security.Cryptography.Native/pal_rsa.h
```diff
@@ -6,6 +6,107 @@
...
+
+Returns 1 upon success, otherwise 0.
+*/
+extern "C" int32_t RsaVerify(int32_t type, const uint8_t* m, int32_t m_len, uint8_t* sigbuf, int32_t siglen, RSA* rsa);
+
+/*
+Gets all the parameters from the RSA instance.
+*/
+extern "C" void GetRsaParameters(const RSA* rsa,
+                                 BIGNUM** n,
```
nguerrera: Why not have PAL structs instead of so many parameters?

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
eerhardt: I'll add the asserts. We never expect this to no-op. I was just being defensive with these checks.

src/Native/System.Security.Cryptography.Native/pal_rsa.h
eerhardt: Since these two methods were only called in 1 place, it felt like overkill to introduce 2 structs for them. Plus I wouldn't expect the parameters to grow/shrink on these methods. So it felt easier to define the methods with all these parameters.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
```diff
@@ -12,3 +31,118 @@ extern "C" RSA* DecodeRsaPublicKey(const unsigned char* buf, int32_t len)
...
+{
+    return RSA_size(rsa);
+}
+
+extern "C" int32_t RsaGenerateKeyEx(RSA* rsa, int32_t bits, BIGNUM* e)
+{
+    return RSA_generate_key_ex(rsa, bits, e, nullptr);
+}
+
+extern "C" int32_t RsaSign(int32_t type, const uint8_t* m, int32_t m_len, uint8_t* sigret, int32_t* siglen, RSA* rsa)
```
nguerrera: m_len looks like a member variable to me.

eerhardt: Agreed. These are the parameter names of the underlying RSA_sign method. I'll change it to mlen, since it is currently inconsistent with siglen.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
```diff
@@ -155,38 +155,31 @@ public override byte[] Decrypt(byte[] data, RSAEncryptionPadding padding)
             if (padding == null)
                 throw new ArgumentNullException("padding");
 
-            Interop.libcrypto.OpenSslRsaPadding openSslPadding;
-
+            bool useOaepPadding;
             if (padding == RSAEncryptionPadding.Pkcs1)
             {
-                openSslPadding = Interop.libcrypto.OpenSslRsaPadding.RSA_PKCS1_PADDING;
+                useOaepPadding = false;
```
nguerrera: We should have a PAL enum for the padding, we're going to want to add more like OaepSHA256.

eerhardt: My thinking was that could be added when it was needed. YAGNI-style.

bartonjs:
The Import/Export methods are both how I once thought of doing it, and a little messy looking.  But all in all, LGTM :).

src/Native/System.Security.Cryptography.Native/pal_rsa.h
nguerrera: OK.

src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs
nguerrera: But we are going to need it: #2523 :)

bartonjs: @nguerrera Well, if it was easy to do in OpenSSL, I'd've done it already :)  It's probably going to be a hassle when we get there.

nguerrera: Sure, but when we get there, we might very well do it here, no? Is the argument that the manual stuff we have to do will require different shimmed functions rather than some extra C++ here? Even then, enums are easier to read than bools and there's always the possibility that we get a future OpenSSL version that does make it easy. It would be unfortunate IMHO, to add a new function just to account for more than two values. As a compromise, you could error out if neither 0 nor 1 is passed and compatibly switch bool -> int32_t marshalling to enum -> enum marshalling. I'd just make the enum, though, there's no need for any static_asserts or conversions, just switch over the input here. I'm not understanding the downside.

eerhardt: A concrete example of why to do things as simple as needed happened just yesterday.

I added a PAL enum for X509Name formatting:
https://github.com/eerhardt/corefx/commit/6c4956feef3e3e672906469dcf87bb68adecca39#diff-443fd6abce16893b65154a678b2e0e23R11

The very same day Jeremy removed it because he took a completely different approach - no longer using X509_NAME_print_ex, but doing it all in managed code:
https://github.com/dotnet/corefx/commit/2f8b7e293c003685a77690e1242fd1a4b5c82c73#diff-443fd6abce16893b65154a678b2e0e23L11

I'll add the PAL enum here now, but I really have no idea how the extra padding options are going to be implemented (i.e. native or managed) until we actually support them.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Rsa.cs
```diff
@@ -9,6 +12,117 @@ internal static partial class Interop
...
+        [return: MarshalAs(UnmanagedType.Bool)]
+        internal static extern bool RsaVerify(int type, byte[] m, int m_len, byte[] sigbuf, int siglen, SafeRsaHandle rsa);
+
+        internal static RSAParameters ExportRsaParameters(SafeRsaHandle key, bool includePrivateParameters)
+        {
+            Debug.Assert(
+                key != null && !key.IsInvalid,
+                "Callers should check the key is invalid and throw an exception with a message");
+
+            if (key == null || key.IsInvalid)
```
stephentoub: Is this condition ever possible if we've implemented the higher layers correctly?  I'm wondering why we have both the assert and the exception.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
```diff
@@ -12,3 +31,122 @@ extern "C" RSA* DecodeRsaPublicKey(const unsigned char* buf, int32_t len)
...
+{
+    return RSA_generate_key_ex(rsa, bits, e, nullptr);
+}
+
+extern "C" int32_t RsaSign(int32_t type, const uint8_t* m, int32_t mlen, uint8_t* sigret, int32_t* siglen, RSA* rsa)
+{
+    if (!siglen)
+    {
+        assert(false);
+        return 0;
```
stephentoub: We should initialize siglen and rsa here, as they're both used as out arguments in the managed declaration.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
```diff
@@ -12,3 +31,122 @@ extern "C" RSA* DecodeRsaPublicKey(const unsigned char* buf, int32_t len)
...
+                                 BIGNUM** p,
+                                 BIGNUM** dmp1,
+                                 BIGNUM** q,
+                                 BIGNUM** dmq1,
+                                 BIGNUM** iqmp)
+{
+    if (!rsa || !n || !e || !d || !p || !dmp1 || !q || !dmq1 || !iqmp)
+    {
+        assert(false);
+        return;
```
stephentoub: Same comment here about initializing the out arguments (assuming the pointers to them aren't null).

stephentoub:

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Rsa.cs
bartonjs: I think I was trying to avoid having an SR reference in the common code.  So the caller would hopefully throw the exception with a meaningful message, the assert is to report that the calling code was naughty, and the runtime if was to prevent bad things from happening.

eerhardt:
I added the PAL enum and rebased on the current code.  Let me know what you think.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
eerhardt: rsa isn't an 'out' argument.  It is just an 'in' pointer.  'siglen' is the only out argument.

if (!siglen), then siglen can't be initialized, right?

and if siglen is a valid pointer, then it will be set below unconditionally.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
eerhardt: They will get set below, right?

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
stephentoub: You're right, I misread the managed signature.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
stephentoub: Not if you return here.

src/Common/src/Interop/Unix/System.Security.Cryptography.Native/Interop.Rsa.cs
eerhardt: Do we want a change here?  I can change it or leave it, I was just pulling the existing code over.

src/Native/System.Security.Cryptography.Native/pal_rsa.cpp
eerhardt: If we return here, there was an invalid parameter passed in. If this was invoked by managed code, none of the 'out' parameters can be NULL, can they? And the managed code already has a check for an invalid 'rsa' pointer before calling this.
So really this guard is for someone else invoking this code, and because it's a good practice to always check pointers before de-referencing them.

Is what you are looking for, something like this?

```C++
if (!rsa || !n || !e || !d || !p || !dmp1 || !q || !dmq1 || !iqmp)
{
    assert(false);
    if (n)
        *n = nullptr;
    if (e)
        *e = nullptr;
    if (d)
        *d = nullptr;
    if (p)
        *p = nullptr;
    if (dmp1)
        *dmp1 = nullptr;
    if (q)
        *q = nullptr;
    if (dmq1)
        *dmq1 = nullptr;
    if (iqmp)
        *iqmp = nullptr;
    return;
}
*n = rsa->n;
*e = rsa->e;
*d = rsa->d;
*p = rsa->p;
*dmp1 = rsa->dmp1;
*q = rsa->q;
*dmq1 = rsa->dmq1;
*iqmp = rsa->iqmp;
```

stephentoub: > Is what you are looking for, something like this?

Yes, either that, or declaring the managed signature to use ref for these rather than out.

The problem here is exacerbated because this function doesn't return any error code, so imagine the managed code invoked it like:
```C#
SafeRsaHandle rsa = GetHandle();
IntPtr n, e, d, p, dmp1, q, dmq1, iqmp;
GetRsaParameters(rsa, out n, out e, out d, out p, out dmp1, out q, out dmq1, out iqmp);
```
and imagine that GetHandle() returned an IsInvalid handle. Without this null setting, those IntPtrs are going to contain garbage values and the caller will be none-the-wiser.  The one call site we have to the managed GetRsaParameters does an explicit check for IsInvalid, but we could easily find ourselves missing that in a future call to this.

stephentoub:
@dotnet-bot test this please (FindPackagesById failures)

eerhardt:
@stephentoub - how about that?

stephentoub:
>  how about that?

I still prefer the initialize-to-null-if-error-in-the-shim approach, because the managed signature then makes it clear what the use case for these arguments is, but I'm ok with this if you don't like that approach for whatever reason.

eerhardt:
I've changed them back to 'out' and added the initialization-to-null code.  I left in the bool return value.

stephentoub:
LGTM.  Thanks, Eric.

PR #3937, opened 2015-10-17
Compute the fallback value for FindBySubjectKeyIdentifier when needed.

### LINKED ISSUE #3935: Unix FindBySubjectKeyIdentifier does not do the same fallback as Windows
It turns out that a lot of the test failing marked as #3099 aren't failing due to that problem.  The test failures coming from a PFX was a red herring.  In the case of the System.Security.Cryptography.X509Certificates.Tests.FindTests.TestBySubjectKeyIdentifier_MatchA test case, it was actually testing the fallback behavior of [CertGetCertificateContextProperty(CERT_KEY_IDENTIFIER_PROP_ID)](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376079%28v=vs.85%29.aspx#_CERT_KEY_IDENTIFIER_PROP_ID):

> If nonexistent, searches for the szOID_SUBJECT_KEY_IDENTIFIER extension. If that fails, a SHA1 hash is done on the certificate's SubjectPublicKeyInfo member to produce the identifier values.

So, yeah, we should do that.

bartonjs:
The Windows CryptoAPI method that FindBySubjectKeyIdentifier defers to has documented fallback logic that, happily, a test was testing.

Updated the names of the tests to better identify what they're accomplishing, and implemented the fallback logic described.

Fixes #3935.

### DIFF
--- a/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
+++ b/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Numerics;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
 
 namespace Internal.Cryptography.Pal
 {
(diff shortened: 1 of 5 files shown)


src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
```diff
@@ -243,15 +244,36 @@ public void FindBySubjectKeyIdentifier(byte[] keyIdentifier)
...
+                        // MSDN says that when there's no extension, this method takes the SHA-1 of the
+                        // SubjectPublicKeyInfo block, and returns that.
+                        //
+                        // https://msdn.microsoft.com/en-us/library/windows/desktop/aa376079%28v=vs.85%29.aspx
+
+                        OpenSslX509CertificateReader certPal = (OpenSslX509CertificateReader)cert.Pal;
+
+                        using (HashAlgorithm hash = SHA1.Create())
+                        {
+                            byte[] publicKeyInfoBytes = Interop.Crypto.OpenSslEnocde(
```
stephentoub: "OpenSslEnocde"?  I assume that's a typo in some existing code?

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
```diff
@@ -243,15 +244,36 @@ public void FindBySubjectKeyIdentifier(byte[] keyIdentifier)
...
+                    {
+                        // The Desktop/Windows version of this method use CertGetCertificateContextProperty
+                        // with a property ID of CERT_KEY_IDENTIFIER_PROP_ID.
+                        //
+                        // MSDN says that when there's no extension, this method takes the SHA-1 of the
+                        // SubjectPublicKeyInfo block, and returns that.
+                        //
+                        // https://msdn.microsoft.com/en-us/library/windows/desktop/aa376079%28v=vs.85%29.aspx
+
+                        OpenSslX509CertificateReader certPal = (OpenSslX509CertificateReader)cert.Pal;
```
stephentoub: Is there any kind of cleanup we need to do with the reader when we're done with it?

stephentoub:
LGTM

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
bartonjs: Yep, now that I don't think there will be any merge conflicts induced by it I'm gonna submit a PR to fix it.

src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs
bartonjs: > Is there any kind of cleanup we need to do with the reader when we're done with it?

Nah, the certificate's still live, this is just reaching through to the SafeHandle that was already in place.

PR #3939, opened 2015-10-17
Enable System.Console.ReadKey

pallavit:
This change tries to enable #1153.

On Windows the implementation is pretty straight-forward. We have APIs to capture the standard input and APIs that tell us the kind of key and modifier is pressed. 

On Unix, there is no concept of virtual key codes and so, the implementation is mostly best case effort. 
 The current implementation works as follows.
 1. Capture stdin in the raw mode and read as much of buffer as possible (max = 1024 bytes).
 2. Try to match the buffer with any of the special keys (which it maintains in the db from the static string capabilities from TermInfo).
 3. If not, try use the ASCII key chart and map it to an equivalent ConsoleKey.
4. Finally return the first char with a ConsoleKey mapped to default.

It also tries to second guess the modifier keys as follows
 1. Alt - This is captured as esc key, so, we try to check if at the time of processing we can get a combination of ESC + known key and if so, we assume alt was pressed.
 2. Ctrl - Other than the known keys like \b \t and \n, the rest of the ASCII key charts from 1 to 26 are considered typed with Ctrl + the alphabet.
 3. Shift - Any capital A to Z are assumed to be pressed with Shift key.

Not all keys are captured by this logic. For example, on my keyboard, NumLock + No. key results in a byte sequence, that can't be mapped to the no. itself. So, Unix implementation can never return ConsoleKey.NumPadX, Similarly combinations of Alt+FunctionKeys can't be interpreted correctly and are interpreted as Esc + the individual keys.


pallavit:
@weshaggard , @stephentoub Could you please take a look.

src/Common/src/Interop/Unix/System.Native/Interop.ConsoleRawMode.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
```
stephentoub: Is this file still used anywhere?

src/Native/System.Native/pal_io.cpp
```diff
@@ -813,3 +814,36 @@ extern "C" int32_t GetWindowSize(WinSize* windowSize)
...
+    newtermios.c_lflag &= static_cast<uint32_t>(~ECHO);
+    newtermios.c_lflag &= static_cast<uint32_t>(~ICANON);
+    newtermios.c_cc[VMIN] = 1;
+    newtermios.c_cc[VTIME] = 0;
+    if (tcsetattr(0, TCSANOW, &newtermios) < 0) return -3;
+    ssize_t count = read(0, buffer, UnsignedCast(bufferSize));
+    tcsetattr(0, TCSANOW, &oldtermios);
+    return static_cast<int32_t>(count);
+#else
+    return -1;
```
stephentoub: We should probably set errno to ENOTSUP.

src/Native/System.Native/pal_io.cpp
```diff
@@ -813,3 +814,36 @@ extern "C" int32_t GetWindowSize(WinSize* windowSize)
...
+
+#if HAVE_TCGETATTR && HAVE_TCSETATTR && HAVE_ECHO && HAVE_ICANON && HAVE_TCSANOW
+    struct termios oldtermios = {};
+    struct termios newtermios = {};
+
+    if (tcgetattr(0, &oldtermios) < 0) return -1;
+
+    newtermios = oldtermios;
+    newtermios.c_lflag &= static_cast<uint32_t>(~ECHO);
+    newtermios.c_lflag &= static_cast<uint32_t>(~ICANON);
```
stephentoub: You could combine these two lines, e.g.
```C++
newtermios.c_lflag &= static_cast<uint32_t>(~(ECHO | ICANON));
```

src/Native/System.Native/pal_io.cpp
```diff
@@ -813,3 +814,36 @@ extern "C" int32_t GetWindowSize(WinSize* windowSize)
...
+    newtermios.c_cc[VMIN] = 1;
+    newtermios.c_cc[VTIME] = 0;
+    if (tcsetattr(0, TCSANOW, &newtermios) < 0) return -3;
```
stephentoub: Why -3 rather than -1?

src/Native/System.Native/pal_io.h
```diff
@@ -622,10 +622,11 @@ extern "C" void Sync();
...
-*
-* Returns 0 on success; otherwise, returns errorNo.
-*/
+ * Gets the windows size of the terminal
+ *
+ * Returns 0 on success; otherwise, returns errorNo.
+ */
 extern "C" int32_t GetWindowSize(WinSize* windowsSize);
+
+extern "C" int32_t ReadStdinUnbuffered(void* buffer, int32_t bufferSize)
```
stephentoub: This doesn't appear to be the latest version of this file, which has IsATty here.  The function also needs a comment.

src/System.Console/ref/System.Console.cs
```diff
@@ -95,6 +97,173 @@ public enum ConsoleColor
...
+        R = 0x52,
+        S = 0x53,
+        T = 0x54,
+        U = 0x55,
+        V = 0x56,
+        W = 0x57,
+        X = 0x58,
+        Y = 0x59,
+        Z = 0x5A,
+#if TODO
```
stephentoub: We shouldn't check this in with these TODOs.

src/System.Console/src/Resources/Strings.resx
```diff
@@ -204,10 +204,16 @@
   <data name="InvalidOperation_PrintF" xml:space="preserve">
     <value>The printf operation failed.</value>
   </data>
+  <data name="InvalidOperation_ConsoleReadKeyOnFile" xml:space="preserve">
+    <value>Cannot read keys when either application does not have a console or when console input has been redirected from a file. Try Console.Read.</value>
+  </data>
```
stephentoub: Can we remove "from a file"?

src/System.Console/src/Resources/Strings.resx
```diff
@@ -204,10 +204,16 @@
...
   <data name="PlatformNotSupported_GettingColor" xml:space="preserve">
     <value>This platform does not support getting the current color.</value>
   </data>
   <data name="PersistedFiles_NoHomeDirectory" xml:space="preserve">
     <value>The home directory of the current user could not be determined.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="ArgumentOutOfRange_ConsoleKey" xml:space="preserve">
+    <value>Console key values must be between 0 and 255.</value>
```
stephentoub: I'd suggest adding ", inclusive" to the end of the text.

src/System.Console/src/System/ConsoleKeyInfo.cs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ 
+namespace System
+{
+    public partial struct ConsoleKeyInfo
+    {
+        private char _keyChar;
+        private ConsoleKey _key;
+        private ConsoleModifiers _mods;
```
stephentoub: Can these be made readonly?

src/System.Console/ref/System.Console.cs
```diff
@@ -95,6 +97,173 @@ public enum ConsoleColor
         White = 15,
         Yellow = 14,
     }
+    public partial struct ConsoleKeyInfo
```
stephentoub: Why ```partial```?

src/System.Console/src/System/ConsoleKeyInfo.cs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ 
+namespace System
+{
+    public partial struct ConsoleKeyInfo
```
stephentoub: Why ```partial```?

src/System.Console/src/System/ConsoleKeyInfo.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            get { return _mods; }
+        }
+
+        public override bool Equals(Object value)
+        {
+            if (value is ConsoleKeyInfo)
+                return Equals((ConsoleKeyInfo)value);
+            else
+                return false;
```
stephentoub: Could just be:
```C#
return value is ConsoleKeyInfo && Equals((ConsoleKeyInfo)value);
```

src/System.Console/src/System/ConsoleKeyInfo.cs
```diff
@@ -0,0 +1,80 @@
...
+            return !(a == b);
+        }
+
+        public override int GetHashCode()
+        {
+            return (int)_keyChar | (int)_mods;
+        }
+    }
+}
+
```
stephentoub: Nit: a few unnecessary extra blank lines

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
             get { return GetConsoleEncoding(); }
         }
 
+        private static readonly object InternalSyncObject = new object();
```
stephentoub: Nit: InternalSyncObject => s_internalSyncObject
I'm also not clear on the "internal" part of the name and why that's necessary.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
...
+        private static SyncTextReader s_stdInReader;
+        private const int DefaultBufferSize = 255;
+
+        public static SyncTextReader StdInReader
```
stephentoub: private instead of public?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
...
+        {
+            get
+            {
+                if (Console.IsInputRedirected)
```
stephentoub: This check should probably be moved down to the result == null if block.  For cases where input isn't redirected, we don't need to be checking this every time we access StdInReader, as all but the first time it'll have been initialized and redirection won't change.  For cases where input is redirected, we're going to throw anyway, so it doesn't matter if we do a little bit of unnecessary work first.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
...
+
+                if (result == null)
+                {
+                    lock (InternalSyncObject)
+                    {
+                        result = Volatile.Read(ref s_stdInReader);
+                        if (result == null)
+                        {
+                            result = SyncTextReader.GetSynchronizedTextReader(
+                            new StdInStreamReader(
```
stephentoub: Nit: this block should be idented

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
...
+                                encoding: InputEncoding,
+                                detectEncodingFromByteOrderMarks: false,
+                                bufferSize: DefaultBufferSize,
+                                leaveOpen: true));
+                            Volatile.Write(ref s_stdInReader, result);
+                        }
+                    }
+                }
+
+                return s_stdInReader;
```
stephentoub: ```return result;```

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,75 @@ public static Encoding OutputEncoding
...
+            }
+        }
+
+        private const int DefaultConsoleBufferSize = 256; // default size of buffer used in stream readers/writers
+        public static TextReader In
```
stephentoub: Nit: I suggest this not be a property, since if input is redirected, every call to this is going to produce a new object.

src/Native/System.Native/pal_io.h
stephentoub: Also, this is missing a ```;``` at the end of the line and won't compile as is.

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -298,6 +371,34 @@ private static void ChangeColor(bool foreground, ConsoleColor color)
         /// <summary>Cache of the format strings for foreground/background and ConsoleColor.</summary>
         private static readonly string[,] s_fgbgAndColorStrings = new string[2, 16]; // 2 == fg vs bg, 16 == ConsoleColor values
 
+        public static bool TryGetSpecialConsoleKey(char[] givenChars, int startIndex, int endIndex, out ConsoleKey key, out int keyLength)
+        {
+            key = (ConsoleKey)0;
+            keyLength = 0;
+            int unprocessedCharCount = endIndex - startIndex + 1;
+
+            if (unprocessedCharCount < TerminalKeyInfo.Instance.MinKeyLength)
```
stephentoub: What happens if we previously read the max buffer size but only got part of the last key?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -298,6 +371,34 @@ private static void ChangeColor(bool foreground, ConsoleColor color)
...
+                return false;
+
+            int minRange = TerminalKeyInfo.Instance.MinKeyLength;
+            int maxRange = unprocessedCharCount > TerminalKeyInfo.Instance.MaxKeyLength ? TerminalKeyInfo.Instance.MaxKeyLength : unprocessedCharCount;
```
stephentoub: ```int maxRange = Math.Min(unprocessedCharCount, TerminalKeyInfoInstance.MaxKeyLength);```

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -298,6 +371,34 @@ private static void ChangeColor(bool foreground, ConsoleColor color)
...
+
+            for (int i = maxRange; i >= minRange; i--)
+            {
+                string currentString = new string(givenChars, startIndex, i);
+
+                // Check if the string prefix matches.
+                if (TerminalKeyInfo.Instance.KeyFormatToConsoleKey.ContainsKey(currentString))
+                {
+                    key = TerminalKeyInfo.Instance.KeyFormatToConsoleKey[currentString];
```
stephentoub:
```C#
if (TerminalKeyInfo.Instance.KeyFormatToConsoleKey.TryGetValue(currentString, out key))
```

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -364,6 +465,201 @@ private TerminalBasicInfo(TermInfo.Database db)
...
+            /// <summary>The format string for KeyEnter</summary>
+            public string KeyEnterFormat;
+            /// <summary>The format string for KeyHelp</summary>
+            public string KeyHelpFormat;
+            /// <summary>The format string for KeyPrint</summary>
+            public string KeyPrintFormat;
+            /// <summary>The format string for KeyInsert</summary>
+            public string KeyInsertFormat;
+            /// <summary>The format string for KeyDelete</summary>
+            public string KeyDeleteFormat;
```
stephentoub: Why do we need all of these fields?  Seems like they're only being used in the TerminalKeyInfo ctor.  Could you just change your AddKey helper to be:
```C#
private void AddKey(TermInfo.Database db, int keyId, ConsoleKey key)
{
    string format = db.GetString(keyId);
    if (!string.IsNullOrEmpty(format))
    {
        KeyFormatToConsoleKey[format] = key;
    }
}
```
and then change the TerminalKeyInfo ctor to be something like:
```C#
private TerminalKeyInfo(TermInfo.Database db)
{
    KeyFormatToConsoleKey = new Dictionary<string, ConsoleKey);
    MaxKeyLength = MinKeyLength = 0;
    if (db != null)
    {
        AddKey(db, TermInfo.Database.KeyF1, ConsoleKey.F1);
        AddKey(db, TermInfo.Database.KeyF2, ConsoleKey.F2);
        ... // etc.
    }
}
```
and then get rid of all of those fields, all of the:
```
KeyF1Format = db != null ? db.GetString(TermInfo.Database.KeyF1) : string.Empty;
```
lines, etc.?

src/System.Console/src/System/ConsolePal.Windows.cs
```diff
@@ -139,6 +141,180 @@ private static bool IsHandleRedirected(IntPtr handle)
             return (!Interop.mincore.IsGetConsoleModeCallSuccessful(handle));
         }
 
+        public static TextReader In
+        {
+            get
+            {
+                    Stream inputStream = OpenStandardInput();
```
stephentoub: Nit: too much indentation

src/System.Console/src/System/ConsolePal.Windows.cs
```diff
@@ -139,6 +141,180 @@ private static bool IsHandleRedirected(IntPtr handle)
...
+                            encoding: InputEncoding,
+                            detectEncodingFromByteOrderMarks: false,
+                            bufferSize: DefaultConsoleBufferSize,
+                            leaveOpen: true));
+            }
+        }
+
+        // Use this for blocking in Console.ReadKey, which needs to protect itself in case multiple threads call it simultaneously.
+        // Use a ReadKey-specific lock though, to allow other fields to be initialized on this type.
+        private static readonly Object ReadKeySyncObject = new object();
```
stephentoub: Nit: 
ReadKeySyncObject => s_readKeySyncObject
Object => object

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+     * which is then used for all the Read operations
+     */
+    internal class StdInStreamReader : StreamReader
+    {
+        char[] unprocessedBufferToBeRead; // Buffer that might have already been read from stdin but not yet processed.
+        const int bytesToBeRead = 255; // No. of bytes to be read from the stream at a time.
+        int startIndex; // First unprocessed index in the buffer;
+        int endIndex; // Last unprocessed index in the buffer;
+
+        internal StdInStreamReader(Stream stream, Encoding encoding, bool detectEncodingFromByteOrderMarks, int bufferSize, bool leaveOpen) : base(stream, encoding, detectEncodingFromByteOrderMarks, bufferSize, leaveOpen)
```
stephentoub: Do we ever construct one of these with varied values for leaveOpen or detectEncodingFromByteOrderMarks?  If not, they can be removed from this new ctor, and the constant values can be used when calling to base.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+            return false;
+        }
+
+        /// <summary> Reads the buffer in the raw mode. </summary>
+        unsafe void ReadExtraBuffer()
+        {
+            Debug.Assert(IsExtraBufferEmpty());
+
+            byte[] buffer = new byte[bytesToBeRead];
```
stephentoub: Does this need to be constructed new every time, or could it just be a field on the reader that we reuse each time?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+            fixed (byte* bufPtr = buffer)
```
stephentoub: If you declare ReadStdinUnbuffered to take a ```byte[]``` instead of a ```byte*```, you won't need to do this manually.  However (contrary to my previous question/comment), since it's a relatively small buffer and since we're not returning it or passing it to anyone else, you could avoid the heap allocation entirely by using a stackalloc here instead.  We should probably do one or the other, either reusing the buffer across calls or using stackalloc.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+     * It is designed to read stdin in raw mode for interpreting
+     * key press events and maintain its own buffer for the same.
+     * which is then used for all the Read operations
+     */
+    internal class StdInStreamReader : StreamReader
+    {
+        char[] unprocessedBufferToBeRead; // Buffer that might have already been read from stdin but not yet processed.
+        const int bytesToBeRead = 255; // No. of bytes to be read from the stream at a time.
+        int startIndex; // First unprocessed index in the buffer;
+        int endIndex; // Last unprocessed index in the buffer;
```
stephentoub: Nit:
The fields should all be declared private explicitly and should all begin with an underscore.
The const should begin with a capital letter.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+                int result;
+                Interop.CheckIo(result = Interop.Sys.ReadStdinUnbuffered((byte*)bufPtr, bytesToBeRead));
+                Debug.Assert(result <= bytesToBeRead);
+
+                // Now we need to convert the byte buffer to char buffer.
+                int charCount = CurrentEncoding.GetMaxCharCount(result);
+                Debug.Assert(charCount < unprocessedBufferToBeRead.Length);
+
+                CurrentEncoding.GetChars(buffer, 0, buffer.Length, unprocessedBufferToBeRead, ++startIndex);
+                endIndex += charCount;
```
stephentoub: This doesn't look right.  GetChars returns the number of characters converted, but this is ignoring that value.  Instead it's using GetMaxCharCount, which will likely produce the wrong result.  There shouldn't be a need to call GetMaxCharCount at all, and instead these four lines can just be:
```
endIndex += CurrentEncoding.GetChars(buffer, 0, buffer.Length, unprocessedBufferToRead, ++startIndex);
```
It's probably also worth a comment explaining the math here, that startIndex and endIndex are both -1, and you need endIndex to point to (not beyond) the last character and you need startIndex to be 0. Though if it were me, I'd skip the increment on startIndex and just be explicit about it:
```
startIndex = 0;
int charCount = CurrentEncoding.GetChars(buffer, 0, buffer.Length, unprocessedBufferToBeRead, startIndex);
endIndex = charCount - 1;
```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+            }
+        }
+
+        public override string ReadLine()
+        {
+            StringBuilder sb = new StringBuilder();
```
stephentoub: How about caching this on the StdInReader so we're not reallocating it and its internal buffer each time?  Then this would just be a Clear().

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                ConsoleKeyInfo keyInfo = ReadKeyInternal();
+
+                if (keyInfo.Key == ConsoleKey.Enter)
+                    return sb.ToString();
+
+                else if (keyInfo.Key == ConsoleKey.Backspace && sb.Length > 0)
+                    sb.Length--;
+
+                if (keyInfo.KeyChar != '\0') // We ignore all non-printable keys
```
stephentoub: Should this be ```else``` as well?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+            //Check if there is anything left in the unprocessedBufferToBeRead.
+            while (!IsExtraBufferEmpty())
+            {
+                ConsoleKeyInfo keyInfo = ReadKeyInternal();
+
+                if (keyInfo.Key == ConsoleKey.Enter)
+                    return sb.ToString();
+
+                else if (keyInfo.Key == ConsoleKey.Backspace && sb.Length > 0)
+                    sb.Length--;
```
stephentoub: Seems like this would be slightly more efficient as:
```C#
else if (keyInfo.Key == ConsoleKey.Backspace)
{
    int len = sb.Length;
    if (len > 0)
        sb.Length = len-1;
}
```
That way in the case of a backspace we're not checking the ```else if``` below in the case where the StringBuilder was empty.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x2A): return ConsoleKey.Multiply;
+
+                case (char)(int)(0x2B): return ConsoleKey.Add;
+
+                case (char)(int)(0x2D): return ConsoleKey.Subtract;
+
+                case (char)(int)(0x2F): return ConsoleKey.Divide;
+
+                case (char)(int)(0x2E): return ConsoleKey.Decimal;
```
stephentoub: Nit: I find this more difficult to read when some of the lines have return on the same line as the case and some don't... I think it'd be better to always have the case be on its own line if any of them are.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x0A):
+                    x = '\0'; // Non-printable character.
+                    return ConsoleKey.Enter;
+
+                case (char)(int)(0x1B):
+                    x = '\0'; // Non-printable character.
+                    return ConsoleKey.Escape;
+
+                case (char)(int)(0x2A): return ConsoleKey.Multiply;
```
stephentoub: ```'*'```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x2B): return ConsoleKey.Add;
```
stephentoub: ``` '+' ```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x2D): return ConsoleKey.Subtract;
```
stephentoub: ``` '-' ```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x2F): return ConsoleKey.Divide;
```
stephentoub: ``` '/' ```

src/System.Console/src/System/IO/StdInStreamReader.cs
stephentoub: ``` '.' ```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x2E): return ConsoleKey.Decimal;
+
+                case (char)(int)(0x7F):
+                    x = '\0'; // Non-printable character.
+                    return ConsoleKey.Delete;
+
+                case (char)(int)(0x20): return ConsoleKey.Spacebar;
```
stephentoub: ``` ' ' ```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                case (char)(int)(0x21):
+                    isShift = true;
+                    return ConsoleKey.D1;
+                    
+                case (char)(int)(0x40):
+                    isShift = true;
+                    return ConsoleKey.D2;
```
stephentoub: Do these mappings need to change based on culture or more generally on what keyboard is being used?  e.g. is '@' always shift-2 as it is typically in the US, or might it be shift-' as I've seen it be in GB?  Or is that normalized away somehow?

stephentoub: Or maybe for situations where we're not fairly certain (e.g. letters, numbers, etc.) we return a ConsoleKeyInfo that has some unknown sentinel for the ConsoleKey and the original value for the char?  That way at least we wouldn't be getting it "wrong", even if we weren't always returning as much information as we could in some situations?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                ReadExtraBuffer();
+            }
+
+            if(TryMapBufferToConsoleKey(out key, out ch, out isShift, out isAlt, out isCtrl))
+            {
+                return new ConsoleKeyInfo(ch, key, isShift, isAlt, isCtrl);
+            }
+
+            return default(ConsoleKeyInfo);
```
stephentoub: Shouldn't we at least return the information we do have? i.e. if a char came in that we couldn't map to anything, should we return a ConsoleKeyInfo that contains that char, even if we couldn't figure out what ConsoleKey to assign to it?  Otherwise we're losing user input, right?

src/System.Console/src/System/ConsoleKeyInfo.cs
```diff
@@ -0,0 +1,80 @@
...
+        public ConsoleKeyInfo(char keyChar, ConsoleKey key, bool shift, bool alt, bool control)
+        {
+            // Limit ConsoleKey values to 0 to 255, but don't check whether the
+            // key is a valid value in our ConsoleKey enum.  There are a few 
+            // values in that enum that we didn't define, and reserved keys 
+            // that might start showing up on keyboards in a few years.
+            if (((int)key) < 0 || ((int)key) > 255)
+            {
+                throw new ArgumentOutOfRangeException("key", SR.ArgumentOutOfRange_ConsoleKey);
```
stephentoub: I realize some things in this change are hard to test, but we should have tests for the things we can verify, e.g. that we're throwing where we expect to be throwing.

src/System.Console/src/System/IO/StdInStreamReader.cs
pallavit: Yes, in the current design we are certainly loosing information. We could implement it as you suggest, however in that case should we not expose a state 0 which is ConsoleKey.Undefined or something similar, so users can specifically code against it. @weshaggard what do you think?

stephentoub: I'd be fine using default(ConsoleKey) / 0 even if we don't have a name for it in the enum (though it'd be nice to have a name for it).  The only place I've seen validation for the value of ConsoleKey is in ConsoleKeyInfo's constructor, and there we're validating that it's between 0 and 255, inclusive, so we're already allowing 0 even though there's no named value for it in the enum.  Using default(ConsoleKey) / 0 is I believe much preferable than simply ignoring key strokes.

weshaggard: Losing data is never good so I agree we should figure out how to capture that somehow. What do we do on full framework if we get something that doesn't map?

pallavit: On windows for every key we always get a virtual key code, so even if we do not have a value for it in ConsoleKey we simply return the virtual key code. Since, adding a new state to the ConsoleKey enum would require update to the other platforms as well, I will do as you suggested use  default(ConsoleKey).

In fact since now we have a way to not lose information, I will remove ConsoleKey.OEM* keys from Unix implementation as well i.e. all the special characters like !,@,# because there is no real guarantee that they will always map to the same OEM key and the current implementation was best effort.

src/Native/System.Native/pal_io.h
pallavit: > This doesn't appear to be the latest version of this file, which has IsATty here. The function also needs a comment.

yes, I need to merge with the latest. However, since I prefer rebasing over merging, I would do it once I address all the comments otherwise the check-in history will change and I will need to force push the changes to the remote branch, causing the diffs and comments to misalign (I am not sure if there is a way to prevent it, let me know if there is :)).

src/System.Console/ref/System.Console.cs
pallavit: Yes, I added the TODO as an open question to get your's and @weshaggard feedback on this. We have two options here.
1. We could expose window's specific keys and ensure windows experience is not compromised and never return these values on Unix.
2. Or we could remove these keys all together.

I was inclining towards exposing them but would like to know your opinion.

src/System.Console/src/Resources/Strings.resx
pallavit: Sure. I don't think we need to be consistent with the desktop in the exception message string.

src/System.Console/src/System/ConsoleKeyInfo.cs
pallavit: Sorry, this was an artifact of an earlier design.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: > and why that's necessary.

Sorry I did not understand this? Do you mean why name it this way? How about s_stdInReaderSyncObject, to correctly depict the usage.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: But that is the intention right? This property will be called only when Console._in is null in which case, we would want to initialize a new TextReader.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: We always assume that we need to first exhaust the current unprocessed buffer and guess are keys based on the available buffer, before reading anything from the stdin (even if stdin has something available) . So, if we ended up capturing a key partially, we interpret the bytes we have to whatever key we can. Does that answer your question? If not, could you share an example?

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: Yeah, I was just referring to the naming.  That said, can we just use Console.EnsureInitialized instead?

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: Sure, but every time someone calls this, it's potentially going to return a new instance.  Typically properties aren't supposed to return new instances every time you call them.  So it seems this would be better as a method named something like GetOrCreateReader.

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: It answers my question, but also confirms my concern :smile:  I realize it's very unlikely that we'll hit such a situation, but if we do we'll end up potentially interpreting the data incorrectly.  I'm wondering if, when we have more than 0 but fewer than MaxKeyLength characters, we should poll to see if there's any more data available from stdin, and if there is, populate the buffer with it?

pallavit: Actually, I am not sure if we will hit this. with vmin = 1 and vtime=0, non-canonical mode and stdin read being atomic in nature, I would expect a key stroke to either be completely captured or completely ignored. The only time I think this case is possible is if we had more than 255 bytes in the buffer at the time we read it. 
However, if this case is possible, how will we differentiate between something getting added to stdin after our read and before our poll and the case where something just got missed from our read.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: Ahh I get your point. I will change it :).

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: We will be synchronizing against things not really required. So, I would prefer keeping a separate one if you don't feel strongly about it.

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: > how will we differentiate

Why would we need to?  If we poll and we can populate more data into the buffer, then we do and we continue processing.  If we poll and we can't get more data, then we know we didn't get a partial key due to previously having too full a buffer.

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: At worst wouldn't we just be synchronizing the first access that initializes any of the .Out/.In/.Error properties?  If so, doesn't seem worth an additional object, both in terms of runtime overhead and in terms of understanding what synchronization object to use for what.  But it's up to you.

src/System.Console/ref/System.Console.cs
stephentoub: I don't have a strong opinion either way.  I'm fine with them being exposed.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: > Why would we need to?

Our motivation for this change is accuracy. What if the last thing we captured was Esc and the next bit is J(when we poll), we will interpret this as enter with our new implementation and in our old one it would have been Esc and J. Any of this could have been true. My point is both implementations could give wrong results, since we are second guessing here. I am open to changing our implementation to what you suggest, I am just not sure if it is more accurate than the current one.

src/System.Console/ref/System.Console.cs
pallavit: That makes it two votes for exposing them :). @weshaggard let me know if you feel strongly about not exposing them? If not, I will just expose them.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: Umm. You are right. I can't think of any other issue with reusing the same object. I will change it.

src/System.Console/src/System/IO/StdInStreamReader.cs
pallavit: > Or maybe for situations where we're not fairly certain (e.g. letters, numbers, etc.) we return a ConsoleKeyInfo that has some unknown sentinel for the ConsoleKey and the original value for the char

As discussed earlier, I have removed all special chars from here (!, @, #) and made ConsoleKey return default here.

akoeplinger: @pallavit is the decision to not handle culture here then? E.g. `Decimal` is `,` in German instead of `.`.

pallavit: @akoeplinger Other than the information that we can get from TermInfo static information, w/0 virtual key code support I could not think of a way to cleanly support it across cultures. If you have any suggestions let me know and I will be happy to take a look at it.

akoeplinger: @pallavit I actually can't think of a way to do it as well, I just wanted to highlight it and make sure we're fine with this limitation :)

Btw. Mono seems to never return `Decimal` (it does `Multiply`, `Add`, etc.) so that may be another option.

stephentoub: > Btw. Mono seems to never return Decimal (it does Multiply, Add, etc.) so that may be another option.

This is essentially what I was suggesting earlier: any place where there could be doubt or a reasonable difference due to keyboard layouts, we just use default(ConsoleKey) rather than trying to guess and getting it right for some users and wrong for others.  It sounds like that's what Pallavi's doing, and we may just need to verify the set we end up with is actually the right set.

src/System.Console/ref/System.Console.cs
pallavit: All across the ref files we follow this pattern

src/System.Console/src/System/ConsoleKeyInfo.cs
pallavit: I am working on adding some tests now :)

src/System.Console/ref/System.Console.cs
pallavit: Since wes is OOF, I am just waiting a little for him to response, if we are close to merging I will expose all the disabled settings for now and we can change it, if needed later :)

src/System.Console/src/System/ConsoleKeyInfo.cs
pallavit: Actually other than testing that we throw InvalidOperationException in redirected mode, there is not much we can do :)

src/Native/System.Native/pal_io.cpp
```diff
@@ -813,3 +814,36 @@ extern "C" int32_t GetWindowSize(WinSize* windowSize)
...
+    struct termios newtermios = {};
+
+    if (tcgetattr(0, &oldtermios) < 0) return -1;
+
+    newtermios = oldtermios;
+    newtermios.c_lflag &= static_cast<uint32_t>(~(ECHO | ICANON));
+    newtermios.c_cc[VMIN] = 1;
+    newtermios.c_cc[VTIME] = 0;
+    if (tcsetattr(0, TCSANOW, &newtermios) < 0) return -1;
+    ssize_t count = read(0, buffer, UnsignedCast(bufferSize));
```
stephentoub: Instead of ```0``` can we use ```STDIN_FILENO```?

src/System.Console/src/System.Console.csproj
```diff
@@ -176,6 +184,9 @@
     <Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.GetWindowWidth.cs">
       <Link>Common\Interop\Unix\Interop.GetWindowWidth.cs"</Link>
     </Compile>
+	<Compile Include="$(CommonPath)\Interop\Unix\System.Native\Interop.ReadStdinUnbuffered.cs">
```
stephentoub: Nit: spacing here and above

src/System.Console/src/System/ConsoleKeyInfo.cs
stephentoub: You can't add a test for this ArgumentOutOfRangeException?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -49,6 +52,54 @@ public static Encoding OutputEncoding
...
+        private static SyncTextReader s_stdInReader;
+        private const int DefaultBufferSize = 255;
+
+        private static SyncTextReader StdInReader
+        {
+            get
+            {
+                return Volatile.Read(ref s_stdInReader) ??
+                    Console.EnsureInitialized(ref s_stdInReader,
+                    () => SyncTextReader.GetSynchronizedTextReader(
```
stephentoub: Nit: spacing

src/System.Console/src/System/ConsolePal.Unix.cs
stephentoub: That's fair.  What about making the buffer bigger to make it less likely we'd hit this?

src/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -364,6 +443,84 @@ private TerminalBasicInfo(TermInfo.Database db)
...
+                    AddKey(db, TermInfo.Database.KeyLeft , ConsoleKey.LeftArrow);
+                    AddKey(db, TermInfo.Database.KeyPageDown , ConsoleKey.PageDown);
+                    AddKey(db, TermInfo.Database.KeyPageUp , ConsoleKey.PageUp);
+                    AddKey(db, TermInfo.Database.KeyRight , ConsoleKey.RightArrow);
+                    AddKey(db, TermInfo.Database.KeyEnd , ConsoleKey.End);
+                    AddKey(db, TermInfo.Database.KeyEnter , ConsoleKey.Enter);
+                    AddKey(db, TermInfo.Database.KeyHelp, ConsoleKey.Help);
+                    AddKey(db, TermInfo.Database.KeyPrint, ConsoleKey.Print);
+                    AddKey(db, TermInfo.Database.KeyInsert, ConsoleKey.Insert);
+                    AddKey(db, TermInfo.Database.KeyDelete , ConsoleKey.Delete);
```
stephentoub: Nit: we don't need the space in each line between the database name and the console key.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+     * key press events and maintain its own buffer for the same.
+     * which is then used for all the Read operations
+     */
+    internal class StdInStreamReader : StreamReader
+    {
+        private char[] _unprocessedBufferToBeRead; // Buffer that might have already been read from stdin but not yet processed.
+        private const int BytesToBeRead = 255; // No. of bytes to be read from the stream at a time.
+        private int _startIndex; // First unprocessed index in the buffer;
+        private int _endIndex; // Last unprocessed index in the buffer;
+        private StringBuilder _readLineSB;
```
stephentoub: readonly?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+namespace System.IO
+{
+    /* This class is used by for reading from the stdin.
+     * It is designed to read stdin in raw mode for interpreting
+     * key press events and maintain its own buffer for the same.
+     * which is then used for all the Read operations
+     */
+    internal class StdInStreamReader : StreamReader
+    {
+        private char[] _unprocessedBufferToBeRead; // Buffer that might have already been read from stdin but not yet processed.
```
stephentoub: readonly?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+            if (_startIndex == -1 && _endIndex == -1) return true;
+
+            if (_startIndex > _endIndex) // Everything has been processed;
+            {
+                _startIndex = -1;
+                _endIndex = -1;
+                return true;
+            }
+
+            return false;
```
stephentoub: If we were to initialize _startIndex to be greater than _endIndex, then it seems this method could just be ```return _startIndex > _endIndex;```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// <summary> Reads the buffer in the raw mode. </summary>
+        unsafe void ReadExtraBuffer()
+        {
+            Debug.Assert(IsExtraBufferEmpty());
+
+            byte* bufPtr = stackalloc byte[BytesToBeRead];
+            int result;
+            Interop.CheckIo(result = Interop.Sys.ReadStdinUnbuffered(bufPtr, BytesToBeRead));
+
+            Debug.Assert(result <= BytesToBeRead);
```
stephentoub: Should we also assert that result > 0?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        private char[] _unprocessedBufferToBeRead; // Buffer that might have already been read from stdin but not yet processed.
+        private const int BytesToBeRead = 255; // No. of bytes to be read from the stream at a time.
+        private int _startIndex; // First unprocessed index in the buffer;
+        private int _endIndex; // Last unprocessed index in the buffer;
+        private StringBuilder _readLineSB;
+
+        internal StdInStreamReader(Stream stream, Encoding encoding, int bufferSize) : base(stream: stream, encoding: encoding, detectEncodingFromByteOrderMarks: false, bufferSize: bufferSize, leaveOpen: true)
+        {
+            _unprocessedBufferToBeRead = new char[encoding.GetMaxCharCount(bufferSize)];
```
stephentoub: Should this be encoding.GetMaxCharCount(BytesToBeRead) rather than encoding.GetMaxCharCount(bufferSize)?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+            //Check if there is anything left in the _unprocessedBufferToBeRead.
+            while (!IsExtraBufferEmpty())
+            {
+                ConsoleKeyInfo keyInfo = ReadKey();
+
+                if (keyInfo.Key == ConsoleKey.Enter)
+                    return _readLineSB.ToString();
+
+                else if (keyInfo.Key == ConsoleKey.Backspace)
+                {
```
stephentoub: Nit: if we have braces on any of the bodies in an if / else, I suggest we have them on all parts of it.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        public override string ReadLine()
+        {
+            //Check if there is anything left in the _unprocessedBufferToBeRead.
+            while (!IsExtraBufferEmpty())
+            {
+                ConsoleKeyInfo keyInfo = ReadKey();
+
+                if (keyInfo.Key == ConsoleKey.Enter)
+                    return _readLineSB.ToString();
```
stephentoub: Don't we also need to clear it? e.g.
```C#
string line = _readLineSB.ToString();
_readLineSB.Clear();
return line;
```
?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+            {
+                x = base.Read();
+            }
+            else
+            {
+                x = (int)_unprocessedBufferToBeRead[_startIndex++];
+            }
+
+            return x;
+        }
```
stephentoub: This is fine, but out of curiosity, why not something more concise like:
```C#
return IsExtraBufferEmpty() ?
    base.Read() :
    (int)_unprocessedBufferToBeRead[_startIndex++];
```

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+            isCtrl = false;
+
+            switch (x)
+            {
+                case '\b':
+                    x = '\0'; // Non-printable character.
+                    return ConsoleKey.Backspace;
+
+                case '\t':
+                    x = '\0'; // Non-printable character.
```
stephentoub: Are we sure this is correct?  If I do Console.ReadLine() on Windows and type a tab as part of my string, the string comes back with a tab in it.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+                    return ConsoleKey.Tab;
+
+                case '\n':
+                    x = '\0'; // Non-printable character.
+                    return ConsoleKey.Enter;
+
+                case (char)(int)(0x1B):
+                    x = '\0'; // Non-printable character.
```
stephentoub: Similarly on Windows the KeyChar when I press Escape comes back from ReadKey as 27; won't setting it to '\0' here make that behave differently on Unix?

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+
+            // 1. Try to get the special key match from the TermInfo static information.
+            if (ConsolePal.TryGetSpecialConsoleKey(_unprocessedBufferToBeRead, _startIndex, _endIndex, out key, out keyLength))
+            {
+                _startIndex += keyLength;
+                ch = '\0'; //Special keys are non-printable.
+                return true;
+            }
+
+            // 1. Check if we can match Esc + combination and guess if alt was pressed.
```
stephentoub: Nit: The above section begins with "1. " and then this does, too.

src/System.Console/src/System/IO/StdInStreamReader.cs
```diff
@@ -0,0 +1,259 @@
...
+            if (isAlt == false && _unprocessedBufferToBeRead[_startIndex] == (char)(int)(0x1B) /*Alt is send as an escape character*/ && _endIndex - _startIndex + 1 >= 2 /*We have at least two characters to read.*/)
```
stephentoub: Nit: could you split this condition onto multiple lines? e.g.
```C#
if (!isAlt && 
    _unprocessedBufferToBeRead[_startIndex] == (char)(int)(0x1B) && // Alt is send as an escape character
    _endIndex - _startIndex + 1 >= 2) //We have at least two characters to read.
{
```

src/System.Console/ref/System.Console.cs
weshaggard: I'm fine with leaving these out for now. I would however remove the #ifdef's and leave the entries there commented out as placeholders so folks coming along later don't put in entries that conflict with those. 

By the way please don't block on me for merging this, I would like to see it get merged today if possible.

src/System.Console/src/System/ConsolePal.Unix.cs
pallavit: We could do that. How about 1024 bytes?

stephentoub: Sounds good.

src/System.Console/src/System/IO/StdInStreamReader.cs
pallavit: Windows and Unix do behave differently here. I was using Console.WriteLine() in my testing so these chars were getting masked, however, I also found that Console.Write() behaves differently for '\0' on windows and Unix. My guess is its because of encoding but I am looking at it more closely.

src/System.Console/src/System/IO/StdInStreamReader.cs
pallavit: Sure, the only reason I prefer not to split these is because once split VS auto-formatting becomes a no-op and we get a ton of formatting issues in general :).

src/System.Console/src/System/IO/StdInStreamReader.cs
pallavit: CheckIO will fail and throw an exception for result < 0 unless the error is EINTR. I could use a while(Interop.CheckIO) but I thought it is unnecessary. Thoughts?

stephentoub: > CheckIO will fail and throw an exception for result < 0

Yes, but not for result == 0.

pallavit: Ahh, that should never happen, since our non-canonical read will block until at least 1 byte is read. However, I will add this just to be safe :).

stephentoub: > Ahh, that should never happen

That's what asserts are for :smile:

pallavit:
@stephentoub Let me know if you have any other comments. I would like to check this in today if possible :).

stephentoub:
The fixes LGTM.

pallavit:
@stephentoub  Thanks :). I will rebase and merge then.

pallavit:
The Jenkins build seems to have failed with an intermittent failure.

> Unhandled Exception: System.InvalidCastException: Unable to cast object of type 'System.Object[]' to type 'System.Net.Http.WinHttpRequestState'.
     at System.Net.Http.WinHttpRequestState.FromIntPtr(IntPtr gcHandle)
     at System.Net.Http.WinHttpRequestCallback.WinHttpCallback(IntPtr handle, IntPtr context, UInt32 internetStatus, IntPtr statusInformation, UInt32 statusInformationLength)
d:\j\workspace\dotnet_corefx_windows_release_prtest\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00107\lib\tests.targets(129,5): warning MSB3073: The command "CoreRun.exe xunit.console.netcore.exe System.Net.Http.Functional.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests  -notrait category=OuterLoop -notrait category=failing " exited with code -532462766. [d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Net.Http\tests\FunctionalTests\System.Net.Http.Functional.Tests.csproj]
d:\j\workspace\dotnet_corefx_windows_release_prtest\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00107\lib\tests.targets(136,5): error : One or more tests failed while running tests from 'System.Net.Http.Functional.Tests' please check log for details! [d:\j\workspace\dotnet_corefx_windows_release_prtest\src\System.Net.Http\tests\FunctionalTests\System.Net.Http.Functional.Tests.csproj]

pallavit:
@dotnet-bot test this please

stephentoub:
> The Jenkins build...

Yeah, see my comment here:
https://github.com/dotnet/corefx/pull/4001#issuecomment-149760942

stephentoub:
@dotnet-bot test this please

PR #3940, opened 2015-10-17
Shim the socket async backend.

pgavlin:
The async backend is exposed via a collection of methods for managing
socket events. The actual mechanism in use varies by platform: on
Linux, the code uses epoll; on OS X, it uses kqueues.


pgavlin:
@stephentoub @sokket @nguerrera PTAL

src/Common/src/Interop/Unix/System.Native/Interop.SocketEvent.cs
```diff
@@ -0,0 +1,48 @@
...
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct SocketEvent
+        {
+            public IntPtr Data;
+            public SocketEvents Events;
+            private int _padding;
+        }
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern unsafe Error CreateSocketEventPort(int* port);
```
stephentoub: Consider making this ```out int port``` instead of ```int* port```

src/Common/src/Interop/Unix/System.Native/Interop.SocketEvent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern Error CloseSocketEventPort(int port);
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern unsafe Error CreateSocketEventBuffer(int count, SocketEvent** buffer);
```
stephentoub: Consider making this ```out SocketEvent* buffer```

src/Common/src/Interop/Unix/System.Native/Interop.SocketEvent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern unsafe Error FreeSocketEventBuffer(SocketEvent* buffer);
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern Error TryChangeSocketEventRegistration(int port, int socket, SocketEvents currentEvents, SocketEvents newEvents, IntPtr data);
+
+        [DllImport(Libraries.SystemNative)]
+        internal static extern unsafe Error WaitForSocketEvents(int port, SocketEvent* buffer, int* count);
```
stephentoub: Consider ```ref int count```

src/System.Net.Sockets/src/System.Net.Sockets.csproj
```diff
@@ -402,48 +402,20 @@
...
-    <Compile Include="$(CommonPath)\Interop\Linux\libc\Interop.epoll_event.cs">
-      <Link>Interop\Linux\libc\Interop.epoll_event.cs</Link>
-    </Compile>
-    <Compile Include="$(CommonPath)\Interop\Linux\libc\Interop.epoll_wait.cs">
-      <Link>Interop\Linux\libc\Interop.epoll_wait.cs</Link>
-    </Compile>
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetsOSX)' == 'true' ">
-    <Compile Include="System\Net\Sockets\SocketAsyncEngineBackend.OSX.cs" />
```
stephentoub: Looks like the only thing standing in the way of recombining the Linux and OSX implementations is this one ```SocketPal.*.cs```... do we think we'll be able to recombine that as well, or do we expect we'll still have Linux vs OS X managed binaries?

stephentoub:
Nice to see the two very-similar managed implementations get combined. LGTM.

src/System.Net.Sockets/src/System.Net.Sockets.csproj
pgavlin: We should be able to recombine that as well; it's just a matter of cost (probably just a static field, a P/Invoke, and some associated branches).

PR #3943, opened 2015-10-18
native: Disable unreachable rule for clang v3.4

jasonwilliams200OK:
The latest EPEL provisioned version of Clang on CentOS 7 is v3.4.2.
On compiling native binaries, I was getting the following error:

```bash
/home/root/corefx/src/Native/System.Native/pal_process.cpp:439:25: error: will never be executed
      [-Werror,-Wunreachable-code]
        for (size_t i = sizeof(native); i < sizeof(pal); i++)
                        ^~~~~~~~~~~~~~
1 error generated.
make[2]: *** [System.Native/CMakeFiles/System.Native.dir/pal_process.cpp.o] Error 1
make[1]: *** [System.Native/CMakeFiles/System.Native.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

The fix was to disable the rule for Clang version less thant 3.5.

This was the only stopper.

### DIFF
--- a/src/Native/CMakeLists.txt
+++ b/src/Native/CMakeLists.txt
@@ -20,6 +20,10 @@ add_compile_options(-I${CMAKE_CURRENT_SOURCE_DIR}/Common)
 add_compile_options(-I${CMAKE_CURRENT_BINARY_DIR}/Common)
 add_compile_options(-Wno-c99-extensions)
 
+if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.5)
+    add_compile_options(-Wno-unreachable-code)
+endif ()
+
 if (CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 OR CMAKE_SYSTEM_PROCESSOR STREQUAL amd64)
     add_definitions(-DBIT64=1)
 elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL armv7l)


OtherCrashOverride:
I don't think this would be the proper way to solve it:

The issue is that the following is never true.
```if (sizeof(pal) > sizeof(native))```

The size of both is known at compile time and it can be deduced that the statement will never be true as the sizes do not ever change after compilation.  Doing the comparison is just wasting CPU cycles each time.  Consider a conditional compile instead.

**Comment by jasonwilliams200OK** [comment]
Nice catch @OtherCrashOverride.
It's fixed.

**Comment by OtherCrashOverride** [comment]
Actually, its not fixed as much as its obscured from detection.  ``#define`` is a pre-processor operation while ``sizeof`` is a compilation operation.  That is why you can not ``#if sizeof(x) > sizeof(y)``.  To solve the issue you would need to use template meta programming.

As an alternative, consider disabling the warning for the block of code affected with a detailed comment explaining the rational for doing so.

As for the other commit "code: Adds UNUSED macro for improved semantics." you should leave that out entirely so as not to hide the underlying problems with the code as-written.

```static void ConvertByteArrayToSockAddrIn(sockaddr_in& addr, const uint8_t* buffer, UNUSED_RELEASE_ONLY int32_t bufferLength)```
That founction would imply that bufferLength is important to avoid buffer over/underflow conditions.  However, the length argument is never used except as an assert.  In the event this fails, it does so silently with no warnings or error with the horrific consequence of having the computer connect to a different machine than intended.

**Comment by jasonwilliams200OK** [comment]
@dotnet-bot test this please

**Comment by jasonwilliams200OK** [comment]
@nguerrera, PTAL.

**Comment by nguerrera** [comment]
Agree that bufferLength should be validated in release too. I was going to file a general issue about making sure we have consistent error checking throughout. Can we leave the UNUSED thing as a separate PR? Is it blocking you?

As for the core issue, as long as we're tricking the compiler out of warning, I'm curious what happens without the if. The loop guard should be enough AFAICT. There would still be dead code when the sizes are equal, so disabling the warning might be more honest, but I'd be fine if in practice it eliminates the warning. I'm less keen on the #define SIZE_* trick.

**Comment by jasonwilliams200OK** [comment]
@nguerrera, I have rebased with dotnet/master, fixed another instance and verified the changes by building on CentOS.

The clang disabling syntax is bit verbose, and it will probably look like:

```C
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wunreachable-code"

if(sizeof(blah) < sizeof(meow))
{
//..
}
else
{
//..
}
#pragma clang diagnostic pop
```

If that is preferred, I would switch the construct. :)

> bufferLength should be validated in release too

Ok. I have stashed that change and removed from this PR. I would love to, but I don't understand in what particular scenario it will actually break anything? I have seen such code used in production in some cross-platform C-libs before and apparently working out quite well. It wasn't breaking anything. This was supposed to be a minor improvement that's why I included in this ongoing PR.  Mostly to cure my own curiosity, pinging @equalsraf: going by the laws of cross-platform C/C++ code, as someone who better understands the dark arts of language than me, do you see any objection with this definition+usage of this `UNUSED` macro https://github.com/jasonwilliams200OK/corefx/commit/ab491f8ae55d82d83c9bdb791f95448386136112? (note `NDEBUG` is a custom flag [set via cmake](https://github.com/dotnet/corefx/blob/2cdff2f/src/Native/CMakeLists.txt#L40) for release configuration, the rest is hopefully self-explanatory)

equalsraf:
@jasonwilliams200OK looks ok as far as I can see without testing (have not checked the MSVC pragma). Might want to have a look at this [SO question](http://stackoverflow.com/questions/3599160/unused-parameter-warnings-in-c-code), with different options and syntax. Some rename the variable to prevent use, which should work even in other compilers.

As for NDEBUG is the idea is to supress warning is release builds? It could probably be achieved with compiler flags too.

jasonwilliams200OK:
Thanks for the info and the SO link @equalsraf. I verified the MSVCR `pragma` rule works. Regarding `NDEBUG`, in one scenario for release configuration `UNUSED` rule needs to be applied but not in debug (due to the assert statement). In other places, it is required for all configurations.

stephentoub:
@dotnet-bot test this please

src/Native/System.Native/pal_networking.cpp
```diff
@@ -2094,7 +2094,10 @@ static Error WaitForSocketEventsInner(int32_t port, SocketEvent* buffer, int32_t
     assert(numEvents != 0);
     assert(numEvents < *count);
 
-    if (sizeof(epoll_event) < sizeof(SocketEvent))
+    size_t sizeof_epoll_event = sizeof(epoll_event);
+    size_t sizeof_socket_event = sizeof(SocketEvent);
```
nguerrera: Hmm, someone is likely to put this back as it was without a comment. Since clang-3.5+ seems forgiving to the extent that we want, maybe we should just test for < 3.5 in CMakeLists.txt and add  -W-no-unreachable-code in that case.

jasonwilliams200OK: Good idea, makes sense, done.

stephentoub:
@dotnet-bot test this please (disabled failing process tests)

nguerrera:
LGTM

PR #3945, opened 2015-10-18
Make Unix X509Certificate2(PFX) prefer a cert with a private key.

### LINKED ISSUE #3099: Unix X509Certificate2(PfxBytes) doesn't prefer a private-key cert
The single-cert path for Unix returns pfx[0]; but for Windows it tries to find a cert with a private key.  Unix should do the same.

bartonjs:
The Unix implementation returned back essentially the value of X509Certificate2Collection(PFX)[0].  The Windows version was more nuanced, returning the first value with a private key, or the first value if nothing had a private key.  The Unix version now matches that behavior.

Fixes #3099.


stephentoub:
LGTM

Issue #3690, opened 2015-10-07
Consider what a LocalMachine\My X509Store might be on Unix


bartonjs:
Largely there's a permissions problem.

* File writes, without reaching into the PAL directly, are done as rwx------
* Windows has the flexible administrator concept, Unix has root.
    * There's sometimes the group wheel, but that's not universal.

There's a possibility of having someone create /etc/dotnet/corefx/cryptography/x509stores, setting the group to whatever they want, giving the group write permission, and then we could start making use of that... but it's not as obvious as how to do it on Windows.

stephentoub:
> File writes, without reaching into the PAL directly, are done as rwx

Do you think the default permissions for new files created by System.IO.FileSystem should be something else?

bartonjs:
@stephentoub The default default should assume user data, so I agree with it.  But for a machine-scoped thing it would likely need to be rw?rw?r-?; and chown'd to root.  I don't think those concepts exist in the public surface area.

If that was the only thing standing in the way it would really be an issue, that's what PALs are for :).  The initial directory creation is one big problem.  (Usually) only root has access to create directories within /etc.  Once /etc/dotnet/corefx/cryptography/x509stores was created and set to the right permissions (rwxrwxr-x, uid:root gid:<group to manage certificates>) the rest could be maintained by the X509Store PAL.

I've now remembered to add the hard bullet... Windows allows everyone to enumerate certificates in the LM\My store... that is, to see the public key/certificate.  The approach taken for user stores of using a PFX for either public-only or public-private wouldn't work.

natemcmaster:
Has there been any more discussion or progess on designing this feature? I hit this limitation while prototyping support for certificates + encrypted XML with ASP.NET Core's data protection libraries.

See also: https://github.com/dotnet/corefx/issues/19489

bartonjs:
@natemcmaster Nope, it remains a difficult to reason-out problem; because it's really an OS feature of Windows (and macOS).  But, as I commented in your other issue, the problem isn't in X509Store per-se, but in EncryptedXml not handling the failure as gracefully as it handles other failures.

the-black-wolf:
has this been given more though? Its kind of annoying, it forces us into writing if/then/else spaghetti code for our .net standard library, which is a complete antithesis of what .net standard is about. Whats the point of having an X509 class in .net **standard** if it throws below exceptions? Then its not a cross platform standard.
[...]

PR #3947, opened 2015-10-18
Throw PNSE instead of NIE on machine X509Store not-supported functionality

### LINKED ISSUE #2207: Add support for additional LocalMachine X509Stores on Unix, or throw PlatformNotSupportedException
LM\Root, and to a lesser degree LM\CA, are populated from a system directory.  The other stores could be put into another path area, if we wanted to support having writable system stores.

If writable system stores makes no sense, then other LM\* stores should throw a descriptive PlatformNotSupportedException, not a blank NotImplementedException.

bartonjs:
Supporting LocalMachine stores is complicated, and tracked by #3690.  Since that isn't something likely to be fixed by 1.0-rtm, replace the placeholder NotImplementedExceptions with PlatformNotSupportedExceptions.

Fixes #2207. (#2206 was fixed a long time ago, that comment was stale)


stephentoub:
LGTM

